PR #276, opened 2020-02-19
✨[RUMF-315] collect initial document timing

bcaudan:
Send an event resource for the initial document

Capture d’écran 2020-02-19 à 14 49 24

### DIFF
--- a/packages/rum/src/performanceCollection.ts
+++ b/packages/rum/src/performanceCollection.ts
@@ -54,11 +54,8 @@ interface FakePerformanceNavigationTiming {
 function retrieveNavigationTimingWhenLoaded(callback: (timing: PerformanceNavigationTiming) => void) {
   function sendFakeTiming() {
     const timing: FakePerformanceNavigationTiming = {
-      domComplete: getRelativeTime(performance.timing.domComplete),
-      domContentLoadedEventEnd: getRelativeTime(performance.timing.domContentLoadedEventEnd),
-      domInteractive: getRelativeTime(performance.timing.domInteractive),
+      ...computeRelativePerformanceTiming(),
       entryType: 'navigation',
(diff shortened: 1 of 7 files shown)


packages/rum/src/resourceUtils.ts
```diff
@@ -44,7 +47,7 @@ function formatTiming(start: number, end: number) {
 }
 
 function isValidTiming(start: number, end: number) {
-  return start > 0 && end > 0 && end >= start
+  return start >= 0 && end >= 0 && end >= start
```
BenoitZugmeyer: The 'strictly greater than' is needed here. `entry.redirectStart`, `entry.secureConnectionStart` and associated are equal to `0` when they don't apply (ex: when there is no redirect or secure connection)

packages/rum/src/performanceCollection.ts
```diff
@@ -76,6 +73,21 @@ function retrieveNavigationTimingWhenLoaded(callback: (timing: PerformanceNaviga
...
 
+interface IndexedPerformanceTiming extends PerformanceTiming {
+  [key: string]: any
+}
+
+function computeRelativePerformanceTiming() {
+  const result: Partial<IndexedPerformanceTiming> = {}
+  const timing = performance.timing as IndexedPerformanceTiming
+  for (const key in timing) {
+    if (Number.isInteger(timing[key] as any)) {
```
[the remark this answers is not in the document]
bcaudan: nope, number should be fine

packages/rum/src/resourceUtils.ts
bcaudan: hum, with initial doc some timings can rightfully be at 0.
I'll look into that and try to add some test to explain that then.

PR #281, opened 2020-02-26
🐛[RUMF-330] fix intake requests exclusion

bcaudan:
when several SDKs are setup on the same page, client tokens can differ.

### DIFF
--- a/packages/rum/src/resourceUtils.ts
+++ b/packages/rum/src/resourceUtils.ts
@@ -1,4 +1,4 @@
-import { addMonitoringMessage, Configuration, includes, msToNs, ResourceKind, startsWith } from '@datadog/browser-core'
+import { addMonitoringMessage, Configuration, includes, msToNs, ResourceKind } from '@datadog/browser-core'
 
 import { PerformanceResourceDetails } from './rum'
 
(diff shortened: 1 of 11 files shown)


packages/rum/src/resourceUtils.ts
```diff
@@ -90,9 +90,13 @@ export function isValidResource(url: string, configuration: Configuration) {
...
-    (configuration.internalMonitoringEndpoint && startsWith(url, configuration.internalMonitoringEndpoint))
+    haveSameOrigin(url, configuration.logsEndpoint) ||
+    haveSameOrigin(url, configuration.rumEndpoint) ||
+    haveSameOrigin(url, configuration.traceEndpoint) ||
+    (configuration.internalMonitoringEndpoint && haveSameOrigin(url, configuration.internalMonitoringEndpoint))
   )
 }
+
+function haveSameOrigin(url1: string, url2: string) {
+  return startsWith(url2, new URL(url1).origin)
```
BenoitZugmeyer: startsWith("https://rum-intake.com.zogzog.net", new URL('https://rum-intake.com/foo?bar=baz').origin)

bcaudan: nice catch, it could be an issue if we reuse that in another context

PR #292, opened 2020-03-05
✨[RUMF-375] do not collect irrelevant timings

bcaudan:
remove connect, dns timings when persistent connection or cache

cf.https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings

### DIFF
--- a/packages/rum/src/resourceUtils.ts
+++ b/packages/rum/src/resourceUtils.ts
@@ -70,7 +70,11 @@ export function computePerformanceResourceDetails(
     redirect: isRelevantTiming(entry.redirectStart, entry.redirectEnd, 0)
       ? formatTiming(entry.redirectStart, entry.redirectEnd)
       : undefined,
-    ssl: entry.secureConnectionStart !== 0 ? formatTiming(entry.secureConnectionStart, entry.connectEnd) : undefined,
+    ssl:
+      entry.secureConnectionStart !== 0 &&
+      isRelevantTiming(entry.secureConnectionStart, entry.connectEnd, entry.fetchStart)
+        ? formatTiming(entry.secureConnectionStart, entry.connectEnd)
+        : undefined,
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/rum/src/resourceUtils.ts
```diff
@@ -67,15 +59,37 @@ export function computePerformanceResourceDetails(
...
+    dns: isRelevantTiming(entry.domainLookupStart, entry.domainLookupEnd, entry.fetchStart)
+      ? formatTiming(entry.domainLookupStart, entry.domainLookupEnd)
+      : undefined,
     download: formatTiming(entry.responseStart, entry.responseEnd),
     firstByte: formatTiming(entry.requestStart, entry.responseStart),
-    redirect: entry.redirectStart !== 0 ? formatTiming(entry.redirectStart, entry.redirectEnd) : undefined,
+    redirect: isRelevantTiming(entry.redirectStart, entry.redirectEnd, 0)
+      ? formatTiming(entry.redirectStart, entry.redirectEnd)
+      : undefined,
     ssl: entry.secureConnectionStart !== 0 ? formatTiming(entry.secureConnectionStart, entry.connectEnd) : undefined,
```
BenoitZugmeyer: Why not using `isRelevantTiming` here?

bcaudan: `secureConnectionStart` and `connectEnd` do not have the same relationship as the other timings.
In most of the cases 
```js
isRelevantTiming(entry. secureConnectionStart, entry.connectEnd, 0)
```
 should be true since `connectEnd` should be > 0

BenoitZugmeyer: A cached resource coming from a SSL connection looks like this:
Screenshot 2020-03-05 at 18 15 30
I think that if we follow the same logic, we'd want
```ts
entry.secureConnectionStart !== 0 &&
   isRelevantTiming(entry.secureConnectionStart, entry.connectEnd, entry.fetchStart)
```

Also, I think we should use `isRelevantTiming` for `download` and `firstByte` too.

bcaudan: > A cached resource coming from a SSL connection 
Yes, I missed that case

> Also, I think we should use isRelevantTiming for download and firstByte too
What cases do you see here?

PR #294, opened 2020-03-11
🐛[RUMF-320] Remove url-polyfill dependency

bcaudan:
The dependency used to polyfill URL API is modifying the global scope which caused an issue for a customer.
Since we don't use a lot of URL features, replace this polyfill by a simple implementation that support only our current needs.

### DIFF
--- a/packages/core/src/urlPolyfill.ts
+++ b/packages/core/src/urlPolyfill.ts
@@ -1,5 +1,7 @@
+import { getLinkElementOrigin, getLocationOrigin } from './utils'
+
 export function normalizeUrl(url: string) {
-  return buildUrl(url, window.location.origin).href
+  return buildUrl(url, getLocationOrigin()).href
 }
 
 export function isValidUrl(url: string) {
@@ -15,11 +17,12 @@ export function haveSameOrigin(url1: string, url2: string) {
 }
 
 export function getOrigin(url: string) {
-  return buildUrl(url).origin
+  return getLinkElementOrigin(buildUrl(url))
 }
 
 export function getPathName(url: string) {
-  return buildUrl(url).pathname
+  const pathname = buildUrl(url).pathname
+  return pathname[0] === '/' ? pathname : `/${pathname}`
 }
 
 export function getSearch(url: string) {
--- a/packages/core/src/urlPolyfill.ts
+++ b/packages/core/src/urlPolyfill.ts
@@ -1,10 +1,10 @@
 export function normalizeUrl(url: string) {
-  return new URL(url, window.location.origin).href
+  return buildUrl(url, window.location.origin).href
 }
 
 export function isValidUrl(url: string) {
   try {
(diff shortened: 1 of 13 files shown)


packages/core/src/urlPolyfill.ts
```diff
@@ -15,17 +15,34 @@ export function haveSameOrigin(url1: string, url2: string) {
...
-  return new URL(url).search
+  return buildUrl(url).search
 }
 
 export function getHash(url: string) {
-  return new URL(url).hash
+  return buildUrl(url).hash
+}
+
+function buildUrl(url: string, base?: string) {
```
BenoitZugmeyer: I think `buildUrl` should use `URL` if available (it would improve performances)

msokk:
https://github.com/DataDog/browser-sdk/pull/294/files#diff-252c841276c8ed9e5cbdc5a3865065dfR47 - appending the `<base>` element causes CSP violations in stricter setups. Ours has `'none'` for `base-uri`.

https://docs.datadoghq.com/real_user_monitoring/faq/content_security_policy/?tab=us - CSP article should include `base-uri 'self'`

bcaudan:
Hi @msokk,
Thanks for raising the issue.
Could you confirm that:
- it was not happening before this release
- it only happens on IE browsers
Thanks

msokk:
IE does not support CSP (apart from `X-Content-Security-Policy: sandbox`). 
This didn't break anything, apart from datadog url not being normalized, just our Sentry got spammed. 

Started first seeing events March 12, ~13:30 GMT

bcaudan:
@msokk is this issue reported by sentry on every browsers?

msokk:
Well, our reports mostly (99%) consisted of Chrome 80, as we have a lot of Chrome users, but there are some Firefox and Opera. 

It's not a big issue, just some extra configuration might be needed from site side if they have set up security headers. We just have a very strict `Content-Security-Policy` by whitelisting third-party scripts/styles/fonts/etc...

So far no third-party piece of code touched the `<base>` element, so we had the https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/base-uri totally blocked :)

bcaudan:
OK, I'd prefer to not need extra csp rules if we can avoid it.

Could you provide me one url where there is the issue?

msokk:
I already adjusted our CSP rules on all environments to stop eating up Sentry quota.

Here is a codepen with `<meta http-equiv="Content-Security-Policy" content="base-uri 'none'">` - https://codepen.io/msokk/pen/eYNMWxy.

Violation is in console

bcaudan:
Thanks a lot for that, I'll have a look and keep you updated if there is any changes on this topic.

bcaudan:
Hi @msokk,
It was indeed an issue on our side, it should be fixed by #302 and available in v1.7.3.
Thanks again for the feedback.

PR #306, opened 2020-03-16
🔒 upgrade vulnerable packages

bcaudan:


package.json
```diff
@@ -25,6 +25,10 @@
     "test:compat:tsc": "yarn build && (cd test/app && rm -rf node_modules && yarn && yarn compat:tsc) || yarn fail 'typescript 3.0 compatibility broken'",
     "test:compat:ssr": "yarn build && (cd test/app && rm -rf node_modules && yarn && yarn compat:ssr) || yarn fail 'server side rendering compatibility broken'"
   },
+  "resolutions": {
+    "acorn": "6.4.1",
+    "minimist": "1.2.2"
+  },
```
BenoitZugmeyer: I don't think we should pin versions in package.json. The lockfile should be enough.

BenoitZugmeyer: Or maybe set `">6.4.1"` so the dependency can be upgraded

bcaudan: If I don't pin version, it will still use older version, if the main version does not match:
```
minimist@0.0.8:
  version "0.0.8"
  resolved "https://registry.yarnpkg.com/minimist/-/minimist-0.0.8.tgz#857fcabfc3397d2625b8228262e86aa7a011b05d"
  integrity sha1-hX/Kv8M5fSYluCKCYuhqp6ARsF0=

minimist@^1.1.0, minimist@^1.1.3, minimist@^1.2.0:
  version "1.2.2"
  resolved "https://registry.yarnpkg.com/minimist/-/minimist-1.2.2.tgz#b00a00230a1108c48c169e69a291aafda3aacd63"
  integrity sha512-rIqbOrKb8GJmx/5bc2M0QchhUouMXSpd1RTclXsB41JdL+VtnojfaJR+h7F9k18/4kHUsBFgk80Uk+q569vjPA==

minimist@~0.0.1:
  version "0.0.10"
  resolved "https://registry.yarnpkg.com/minimist/-/minimist-0.0.10.tgz#de3f98543dbf96082be48ad1a0c7cda836301dcf"
  integrity sha1-3j+YVD2/lggr5IrRoMfNqDYwHc8=

```
instead of
```
minimist@0.0.8, minimist@1.2.2, minimist@^1.1.0, minimist@^1.1.3, minimist@^1.2.0, minimist@~0.0.1:
  version "1.2.2"
  resolved "https://registry.yarnpkg.com/minimist/-/minimist-1.2.2.tgz#b00a00230a1108c48c169e69a291aafda3aacd63"
  integrity sha512-rIqbOrKb8GJmx/5bc2M0QchhUouMXSpd1RTclXsB41JdL+VtnojfaJR+h7F9k18/4kHUsBFgk80Uk+q569vjPA==
```
wdyt?

BenoitZugmeyer: Ah, you're right, some libs are still requiring outdated versions. I think it'll be OK for now, but we should upgrade our dependencies and reconsider. Maybe use more precise overrides like this:
```
    "**/gulp-sourcemaps/acorn": "6.4.1",
    "**/optimist/minimist": "1.2.2",
    "**/mkdirp/minimist": "1.2.2"
```

Issue #280, opened 2020-02-21
No way to check if init has already been called or to catch/ignore resulting console error.


adamond:
Please let me know if I'm missing something but:

- There doesn't appear to be a way to check if `datadogLogs.init()` has already been called, and `console.error()` is called if it had already been initialized. However, since we're sending our console errors to datadog, this leads to our logs getting spammed by these error messages that we don't care about. Ideally, there would be something on the datadogLogs API that we could check before calling `init` to see if it has already been called.

- There's no good workaround at the moment for suppressing these console errors, because they're using `console.error()` rather than `throw` - presumably this is because you don't want to halt execution, which is understandable - however, it'd be nice in this case if there was a way to configure the logger to filter out certain console errors. If such functionality already exists, that would be something that would be helpful to highlight in the documentation. 

Thanks!

bcaudan:
Hi @adamond! 
Indeed, there is no API to check if the sdk is already initialized.
We chose to send error because multiple calls to init was related to sdk setup mistakes.
In order to avoid these errors, I would suggest to add the datadog init call in some bootstrap phase of your page that happens early and only once.
If it is not possible in your case, could  you give me extra details on your setup?

Sangrene:
I must agree with @adamond on the usefulness of a `isInit` function.

Actually, a workaround I use is to add an abstraction layer on top of browser-logs, and then use a singleton pattern on this abstraction.

bcaudan:
Hi @Sangrene

Could you give me some details on your setup?
It would help me to understand your need and see if adding a `isInit` function is the best way to go or if we could find an alternate strategy.

Thanks!

donovanclarke:
+1 for this.

Sangrene:
> Hi @Sangrene
> 
> Could you give me some details on your setup?
> It would help me to understand your need and see if adding a `isInit` function is the best way to go or if we could find an alternate strategy.
> 
> Thanks!

Hi @bcaudan 

I have an application divided in multiples containers. Each containers has its own bootstrap function, sometimes initialising _browser-logs_. A user might (or might not) navigate between containers, triggering the associated bootstrap function. When navigating between two containers which have initialised _browser-logs_, the initialise function is therefore called twice.
As I said, I use an abstraction layer and a singleton pattern to workaround this.

bcaudan:
Thanks for the details, it makes sense.

Would an init option be more convenient to suppress the `already initialized` error?

something like:

```js
datadogLogs.init({
[...]

PR #310, opened 2020-03-18
✨add an option to silent multiple Init errors

mquentin:
What it does
Add a new and optional init option : `silentMultipleInit?: boolean`

Context
[https://github.com/DataDog/browser-sdk/issues/280](https://github.com/DataDog/browser-sdk/issues/280)

Changes
- Update README files
- Add the new `silentMultipleInit?: boolean` property to the `UserConfiguration` type
- Check the boolean while being already initialized

### DIFF
--- a/packages/rum/src/rum.entry.ts
+++ b/packages/rum/src/rum.entry.ts
@@ -82,8 +82,10 @@ datadogRum.init = monitor((userConfiguration: RumUserConfiguration) => {
 })
 
 function canInitRum(userConfiguration: RumUserConfiguration) {
-  if (isAlreadyInitialized && !userConfiguration.silentMultipleInit) {
-    console.error('DD_RUM is already initialized.')
+  if (isAlreadyInitialized) {
+    if (!userConfiguration.silentMultipleInit) {
+      console.error('DD_RUM is already initialized.')
+    }
     return false
   }
   if (!userConfiguration || (!userConfiguration.clientToken && !userConfiguration.publicApiKey)) {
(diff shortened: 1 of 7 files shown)


packages/rum/src/rum.entry.ts
```diff
@@ -82,7 +82,7 @@ datadogRum.init = monitor((userConfiguration: RumUserConfiguration) => {
 })
 
 function canInitRum(userConfiguration: RumUserConfiguration) {
-  if (isAlreadyInitialized) {
+  if (isAlreadyInitialized && !userConfiguration.silentMultipleInit) {
```
bcaudan: Can you add a unit test on it?

packages/rum/src/rum.entry.ts
```diff
@@ -82,7 +82,7 @@ datadogRum.init = monitor((userConfiguration: RumUserConfiguration) => {
...
     console.error('DD_RUM is already initialized.')
     return false
   }
```
bcaudan: we should still return false

glorieux:
@mquentin is the `UserConfiguration` updated if the `init` is called multiple times?

mquentin:
> @mquentin is the `UserConfiguration` updated if the `init` is called multiple times?

The init function is probably called with the same `UserConfiguration`. As I understand form the issue [https://github.com/DataDog/browser-sdk/issues/280#issuecomment-599478304](https://github.com/DataDog/browser-sdk/issues/280#issuecomment-599478304) it happens when several container run the same `init.

packages/rum/src/rum.entry.ts
mquentin: done

PR #314, opened 2020-03-23
🔧 improve CBT test names

BenoitZugmeyer:
Give more information on which test is running on CBT

Screenshot 2020-03-23 at 18 22 30

### DIFF
--- a/test/getTestName.js
+++ b/test/getTestName.js
@@ -1,13 +1,31 @@
-const { CI_JOB_ID, CI_PIPELINE_ID, CI_COMMIT_SHORT_SHA } = process.env
+const NAME = 'browser-sdk'
 
 function getCIInfos() {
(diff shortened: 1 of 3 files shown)


test/getTestName.js
```diff
@@ -0,0 +1,13 @@
+const { CI_JOB_ID, CI_PIPELINE_ID, CI_COMMIT_SHORT_SHA } = process.env
+
+function getCIInfos() {
+  if (CI_PIPELINE_ID && CI_JOB_ID && CI_COMMIT_SHORT_SHA) {
+    return `job: ${CI_JOB_ID} commit: ${CI_COMMIT_SHORT_SHA}`
+  }
```
bcaudan: Could it be useful to have `${process.env.USER} (local)` for local runs?

BenoitZugmeyer: good idea!

PR #315, opened 2020-03-23
[RUMF-407] improve resource timings collection

BenoitZugmeyer:
With this PR, we try to improve the resources timing collection in different ways:

* Better cross browser support:
  * If `timing.duration` isn't available or 0, use `responseEnd - startTime` instead, if available (Safari)
  * If `redirectStart` isn't available, use `startTime` instead (Firefox)
  * If `redirectEnd` isn't available, use `fetchStart` instead (Firefox)

* Change how the timings `start` are computed: use difference (in nanosecond) between the request `startTime` and the timing start, so we can build reliable timings diagram.
  * Discard any timing lower than `startTime` instead of 0 since it should not happen and would provide negative timings `start`.

* Report timings for all cached resources, as we've seen that we can't accurately know if a resource is cached or not (timings may not be 0 when cached)

### DIFF
--- a/packages/rum/src/resourceUtils.ts
+++ b/packages/rum/src/resourceUtils.ts
@@ -48,68 +48,122 @@ export function computeResourceKind(timing: PerformanceResourceTiming) {
   return ResourceKind.OTHER
 }
 
-export function computePerformanceResourceDetails(
-  entry?: PerformanceResourceTiming
-): PerformanceResourceDetails | undefined {
-  if (!entry || !hasTimingAllowedAttributes(entry) || isCached(entry)) {
-    return undefined
+function areInOrder(...numbers: number[]) {
+  for (let i = 1; i < numbers.length; i += 1) {
+    if (numbers[i - 1] > numbers[i]) {
+      return false
+    }
   }
+  return true
+}
+
+export function computePerformanceResourceDuration(entry: PerformanceResourceTiming): number {
+  const { duration, startTime, responseEnd } = entry
+
+  // Safari duration is always 0 on timings blocked by cross origin policies.
+  if (duration === 0 && startTime < responseEnd) {
+    return msToNs(responseEnd - startTime)
+  }
+
+  return msToNs(duration)
+}
+
+export function computePerformanceResourceDetails(
+  entry: PerformanceResourceTiming
+): PerformanceResourceDetails | undefined {
+  const {
+    startTime,
+    fetchStart,
+    domainLookupStart,
+    domainLookupEnd,
+    connectStart,
+    secureConnectionStart,
+    connectEnd,
+    requestStart,
+    responseStart,
+    responseEnd,
+  } = entry
+  let { redirectStart, redirectEnd } = entry
+
+  // Ensure timings are in the right order.  On top of filtering out potential invalid
+  // PerformanceResourceTiming, it will ignore entries from requests where timings cannot be
+  // collected, for example cross origin requests without a "Timing-Allow-Origin" header allowing
+  // it.
   if (
-    !isValidTiming(entry.connectStart, entry.connectEnd) ||
-    !isValidTiming(entry.domainLookupStart, entry.domainLookupEnd) ||
-    !isValidTiming(entry.responseStart, entry.responseEnd) ||
-    !isValidTiming(entry.requestStart, entry.responseStart) ||
-    !isValidTiming(entry.redirectStart, entry.redirectEnd) ||
-    !isValidTiming(entry.secureConnectionStart, entry.connectEnd)
+    !areInOrder(
+      startTime,
+      fetchStart,
+      domainLookupStart,
+      domainLookupEnd,
+      connectStart,
+      connectEnd,
+      requestStart,
+      responseStart,
+      responseEnd
+    )
   ) {
     return undefined
   }
+
+  // The only time fetchStart is different than startTime is if a redirection occured.
+  const hasRedirectionOccured = fetchStart !== startTime
+
+  if (hasRedirectionOccured) {
+    // Firefox doesn't provide redirect timings on cross origin requests.  Provide a default for
+    // those.
+    if (redirectStart < startTime) {
+      redirectStart = startTime
+    }
+    if (redirectEnd < startTime) {
+      redirectEnd = fetchStart
+    }
+
+    // Make sure redirect timings are in order
+    if (!areInOrder(startTime, redirectStart, redirectEnd, fetchStart)) {
+      return undefined
+    }
+  }
+
+  const details: PerformanceResourceDetails = {
+    download: formatTiming(startTime, responseStart, responseEnd),
+    firstByte: formatTiming(startTime, requestStart, responseStart),
+  }
+
+  // Make sure a connection occured
+  if (connectEnd !== fetchStart) {
+    details.connect = formatTiming(startTime, connectStart, connectEnd)
+
+    // Make sure a secure connection occured
+    if (areInOrder(connectStart, secureConnectionStart, connectEnd)) {
+      details.ssl = formatTiming(startTime, secureConnectionStart, connectEnd)
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


test/e2e/scenario/agents.scenario.ts
```diff
@@ -72,32 +73,11 @@ describe('rum', () => {
...
-    expect(timing as any).not.toBe(undefined)
-    expect(timing.http.method).toEqual('GET')
-    expect((timing.http as any).status_code).toEqual(200)
-    expect(timing.duration).toBeGreaterThan(0)
-    expect(timing.http.performance!.download!.start).toBeGreaterThan(0)
+    const timing = await makeXHRAndCollectEvent(`${browser.config.baseUrl}/ok`)
+    expect(timing).not.toBeUndefined()
+    expect(timing!.http.method).toEqual('GET')
+    expect((timing!.http as any).status_code).toEqual(200)
+    expectToHaveValidTimings(timing!)
```
bcaudan:
```suggestion
    const timing = await makeXHRAndCollectEvent(`${browser.config.baseUrl}/ok`)!
    expect(timing).not.toBeUndefined()
    expect(timing.http.method).toEqual('GET')
    expect((timing.http as any).status_code).toEqual(200)
    expectToHaveValidTimings(timing)
```
and for other occurences

test/e2e/scenario/agents.scenario.ts
```diff
@@ -73,7 +74,31 @@ describe('rum', () => {
...
+  })
+
+  it('should not track disallowed cross origin xhr timings', async () => {
+    const timing = await makeXHRAndCollectEvent(`${serverUrl.crossOrigin}/ok`)
+    expect(timing).not.toBeUndefined()
+    expect(timing!.http.method).toEqual('GET')
+    expect((timing!.http as any).status_code).toEqual(200)
+    expect(timing!.duration).toBeGreaterThan(0)
+
+    // Edge 18 seems to have valid timings even on cross browser requests ¯\_ツ_/¯ It doesn't matter
```
bcaudan:
```suggestion
    // Edge 18 seems to have valid timings even on cross origin requests ¯\_ツ_/¯ It doesn't matter
```

test/e2e/scenario/agents.scenario.ts
```diff
@@ -73,7 +74,31 @@ describe('rum', () => {
...
+    // too much.
+    if (browser.capabilities.browserName === 'MicrosoftEdge' && browser.capabilities.browserVersion === '18') {
```
bcaudan: Chrome based Edge does not seem to be rollout by microsoft yet, this behavior could be fixed with it.

test/e2e/scenario/agents.scenario.ts
```diff
@@ -73,7 +74,31 @@ describe('rum', () => {
...
+    expect((timing!.http as any).status_code).toEqual(200)
+    expectToHaveValidTimings(timing!)
+  })
+
+  it('should not track disallowed cross origin xhr timings', async () => {
+    const timing = await makeXHRAndCollectEvent(`${serverUrl.crossOrigin}/ok`)
+    expect(timing).not.toBeUndefined()
+    expect(timing!.http.method).toEqual('GET')
+    expect((timing!.http as any).status_code).toEqual(200)
+    expect(timing!.duration).toBeGreaterThan(0)
```
bcaudan: could it worth to add a duration check on other tests?

test/server/fake-backend.js
```diff
@@ -38,6 +38,15 @@ module.exports = (app) => {
     }
     res.send('ok')
   })
+
+  app.get('/redirect/:n', (req, res) => {
+    const n = Number(req.params.n)
+    if (!isNaN(n) && n > 1) {
+      res.redirect(`${n - 1}`)
```
bcaudan: Since it is not used, does it worth to keep this logic?

packages/rum/src/resourceUtils.ts
```diff
@@ -48,68 +48,122 @@ export function computeResourceKind(timing: PerformanceResourceTiming) {
...
+
+  // The only time fetchStart is different than startTime is if a redirection occured.
+  const hasRedirectionOccured = fetchStart !== startTime
+
+  if (hasRedirectionOccured) {
+    // Firefox doesn't provide redirect timings on cross origin requests.  Provide a default for
+    // those.
+    if (redirectStart < startTime) {
+      redirectStart = startTime
+    }
```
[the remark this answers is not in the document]
BenoitZugmeyer: Yes, I didn't include timings `< startTime` but they are indeed `0`

packages/rum/src/resourceUtils.ts
```diff
@@ -48,68 +48,122 @@ export function computeResourceKind(timing: PerformanceResourceTiming) {
...
+    if (redirectEnd < startTime) {
+      redirectEnd = fetchStart
+    }
+
+    // Make sure redirect timings are in order
+    if (!areInOrder(startTime, redirectStart, redirectEnd, fetchStart)) {
+      return undefined
+    }
```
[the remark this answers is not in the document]
BenoitZugmeyer: No, because if no redirection occurred, `redirectStart` and `redirectEnd` are expected to be 0, so the `hasRedirectionOccured` condition is needed before testing those.

test/e2e/scenario/agents.scenario.ts
BenoitZugmeyer: Ok. In TS 3.8, I'd like to make something like:
```ts
function expectDefined<T>(value: T | undefined): asserts value is T {
  expect(value).toBeDefined()
}
```

test/e2e/scenario/agents.scenario.ts
BenoitZugmeyer: The Chrome based Edge has been released and stable since 15 january... but you are right, it doesn't auto-update on all Windows setups... things will be better in the future :)

test/e2e/scenario/agents.scenario.ts
BenoitZugmeyer: `duration` is tested by `expectToHaveValidTimings`

packages/rum/src/resourceUtils.ts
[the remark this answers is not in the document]
bcaudan: Yes, what I had in mind was something like:
```js
...
let { redirectStart, redirectEnd } = entry
const hasRedirectionOccured = fetchStart !== startTime
if (hasRedirectionOccured) {
...
}
if (
  !areInOrder(
    startTime,
    redirectStart,
    redirectEnd, 
    fetchStart,
    domainLookupStart,
    ...
) {
  return undefined
}
...
```
In order to check the order once with all the timings. Is there anything preventing it?

test/e2e/scenario/agents.scenario.ts
bcaudan: Does it worth the trouble?
IMO, it does not bring much more than the `!` operator and we (or at least I 😄) could forget the existence of this helper easily since it is pretty specific.

BenoitZugmeyer: We'll see if it can improve the situation. This is great to link the TS type to the actual JS type at runtime :) `!` should be avoided if possible, in the same way as `as` casts.

packages/rum/src/resourceUtils.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Yes: if `hasRedirectionOccured == false`, `redirectStart` and `redirectEnd` will be `0` (as expected), so `< startTime`, so we would ignore all timings without redirection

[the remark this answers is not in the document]
bcaudan: ok, I get it

test/e2e/scenario/agents.scenario.ts
bcaudan: ok, let's try

PR #316, opened 2020-03-24
[RUMF-402]: Adds CHANGELOG

glorieux:


CHANGELOG.md
```diff
@@ -0,0 +1,413 @@
...
+
+## v1.7.4
+
+- 🐛 replace console.error by console.warn when cookies are not supported (#307)
+- :lock: upgrade vulnerable packages (#306)
+
+## v1.7.3
+
+- :bug:[RUMF-403] fix checkURLSupported (#302)
+- :white_check_mark: add cbt error case (#299)
```
bcaudan: it could be nice to automatically convert emoji name to emoji icon, something like https://www.npmjs.com/package/emoji-name-map could help

glorieux: It's weird. Some of the commit have the unicode value and others have the `:name:` representation. I wonder what's doing the conversion? 🤔

glorieux: Done 😄

bcaudan: IMO, it is done by us manually.
On my side, I some time reuse the converted message from github as the PR desciption to avoid :xxx: in the PR name.

bcaudan: It could also be added as a pre-commit hook at some point

package.json
```diff
@@ -12,6 +12,7 @@
     "lint": "tslint -c ./tslint.json -p ./tsconfig.json && tslint -c ./tslint.json -p ./test/e2e/scenario/tsconfig.json",
     "dev": "ENV=development node test/server/server.js",
     "release": "lerna version --exact",
+    "version": "./scripts/changelog.sh",
```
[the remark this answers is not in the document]
glorieux: Lerna is calling `version` as one of its lifecycle scripts. 
https://github.com/lerna/lerna/tree/master/commands/version#lifecycle-scripts

scripts/emojis.js
```diff
@@ -0,0 +1,45 @@
...
+
+const fs = require('fs')
+const path = require('path')
+const readline = require('readline')
+const name = require('emoji-name-map')
+const replace = require('replace-in-file')
+
+const CHANGELOG_FILE = 'CHANGELOG.md'
+
+async function main() {
```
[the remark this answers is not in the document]
glorieux: I could call it right between when the CHANGELOG.md is generated and when it's shown to the user. What do you think?

CHANGELOG.md
```diff
@@ -0,0 +1,179 @@
...
+## v1.2.4
+
+- 🐛[types] do not globally override ts types (#215)
+- [RUMF-201] add debug monitoring for navigation timing entries (#214)
+
+## v1.2.3
+
+- [RUMF-158] fix view id associated to different session id (#211)
+- 🔧[packages] add scripts to release & publish (#212)
+- :pencil:[packages] improve README.md (#209)
```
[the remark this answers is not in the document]
glorieux: Yes. It's not on it 😢
It's only 🖊 `:pen:` or ✏️ `:pencil2:`.

scripts/changelog.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
BenoitZugmeyer: I'd be in favor of rewriting this in JS/Node. It would be more verbose but more portable and readable...

scripts/emojis.js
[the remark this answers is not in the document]
bcaudan: Yes, it could be integrated in the changelog script then

CHANGELOG.md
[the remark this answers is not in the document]
bcaudan: actually, it is displayed correctly when looking at the rendered [CHANGELOG](https://github.com/DataDog/browser-sdk/blob/b7600a0f4169bb04d23c5238752bd07b23841747/CHANGELOG.md#v123).
Do we really need that parsing script then?

CHANGELOG.md
```diff
@@ -0,0 +1,179 @@
+# Changelog
+
+## v1.7.5
+
+- ✨add an option to silent multiple Init errors (#310)
```
bcaudan: it could be nice to be able add the link the PR number, it does not seem to be generated from github, cf [CHANGELOG](https://github.com/DataDog/browser-sdk/blob/b7600a0f4169bb04d23c5238752bd07b23841747/CHANGELOG.md#v175)

CHANGELOG.md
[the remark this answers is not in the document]
glorieux: Yep. For all cases where you're not looking at the CHANGELOG from Github.

CHANGELOG.md
glorieux: Github should be able to recognize them once it's on master. 
I don't think their PR parsing engine is running on every document on every branches.

bcaudan: It does not, I found this one https://github.com/angular/protractor/blob/master/CHANGELOG.md#features

CHANGELOG.md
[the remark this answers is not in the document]
bcaudan: OK, why not.
Then, what about submitting a PR on https://github.com/muan/emojilib to add it?

scripts/generate-changelog.js
```diff
@@ -0,0 +1,90 @@
...
+const spawn = util.promisify(require('child_process').spawn)
+
+const name = require('emoji-name-map')
+
+const lernaConfig = require('../lerna.json')
+
+const CHANGELOG_FILE = 'CHANGELOG.md'
+
+async function main() {
+  const lastHashCmd = await exec('git rev-list --tags --max-count=1')
```
bcaudan:
```suggestion
  const lastHashCmd = await exec('git rev-parse HEAD')
```
It seems a bit more simple to me

scripts/generate-changelog.js
```diff
@@ -0,0 +1,90 @@
...
+    console.error(`error: ${stderr}`)
+    return
+  }
+  const oldTag = oldTagCmd.stdout.replace(/\n$/, '')
+
+  const commitsCmd = await exec(`git log ${oldTag}..HEAD --pretty=format:"- %s"`)
+  if (commitsCmd.stderr) {
+    console.error(`error: ${stderr}`)
+    return
+  }
```
bcaudan: with all this command / if error / return, there is probably a pattern to extract

scripts/generate-changelog.js
```diff
@@ -0,0 +1,90 @@
...
+    await replace({
+      files: CHANGELOG_FILE,
+      from: new RegExp('# Changelog', 'g'),
+      to: `# Changelog\n\n## v${lernaConfig.version}\n\n${commitsCmd.stdout}`,
+    })
+  } catch (error) {
+    console.error('Error occurred:', error)
+  }
+
+  await emojiNameToUnicode()
```
bcaudan: Could we not just parse emoji in `commitsCmd.stdout` ?

it would avoid to reopen and reparse the file every time

scripts/generate-changelog.js
```diff
@@ -0,0 +1,90 @@
...
+  if (commitsCmd.stderr) {
+    console.error(`error: ${stderr}`)
+    return
+  }
+
+  try {
+    await replace({
+      files: CHANGELOG_FILE,
+      from: new RegExp('# Changelog', 'g'),
+      to: `# Changelog\n\n## v${lernaConfig.version}\n\n${commitsCmd.stdout}`,
```
bcaudan: IMO, we could simplify the problem by removing the `# Changelog` title and just append the new version to the start of the file.

scripts/generate-changelog.js
```diff
@@ -0,0 +1,90 @@
...
+
+const lernaConfig = require('../lerna.json')
+
+const CHANGELOG_FILE = 'CHANGELOG.md'
+
+async function main() {
+  const lastHashCmd = await exec('git rev-list --tags --max-count=1')
+  if (lastHashCmd.stderr) {
+    console.error(`error: ${stderr}`)
+    return
```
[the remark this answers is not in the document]
glorieux: Indeed

scripts/generate-changelog.js
glorieux: I've followed a pretty common CHANGELOG format that includes a title.

scripts/generate-changelog.js
glorieux: What I'm trying to get is the commit hash of the last previous tag not HEAD's commit hash.

CHANGELOG.md
glorieux: I've added a pull request links.

scripts/generate-changelog.js
bcaudan: oh ok, so some renaming could make that clearer, what about:
- lastHashCmd -> lastTagHashCmd
- oldTagCmd -> lastTagNameCmd

scripts/generate-changelog.js
```diff
@@ -0,0 +1,80 @@
...
+    throw commitsCmd.stderr
+  }
+
+  const changesWithEmojis = await emojiNameToUnicode(
+    `# Changelog\n\n## v${lernaConfig.version}\n\n${commitsCmd.stdout}`
+  )
+
+  const changesWhihPullRequestLinks = changesWithEmojis.replace(
+		/\(#(\d+)\)/gm,
+    (match, id) => `[(#${id})](https://github.com/DataDog/browser-sdk/pull/${id})`
```
bcaudan:
```suggestion
    (match, id) => `([#${id}](https://github.com/DataDog/browser-sdk/pull/${id}))`
```
it would be more consistent with usual github PR number link decoration

scripts/generate-changelog.js
```diff
@@ -0,0 +1,80 @@
...
+      const emoji = name.get(match)
+      if (!!emoji) {
+        changelogEmojisNames.set(match, emoji)
+      }
+    })
+  }
+
+  for await (const [emojiName, emojiUnicode] of changelogEmojisNames) {
+    changes = changes.replace(new RegExp(emojiName, 'g'), emojiUnicode)
+  }
```
bcaudan:
```suggestion
  const matches = emojiNameRegex.exec(changes)
  if (!!matches) {
    matches.map((match) => {
      changes.replace(match, name.get(match) || match)
    })
  }
```
Could it be simplified to something like that?

scripts/generate-changelog.js
```diff
@@ -0,0 +1,80 @@
...
+
+  try {
+    await replace({
+      files: CHANGELOG_FILE,
+      from: new RegExp('# Changelog', 'g'),
+      to: changesWhihPullRequestLinks,
+    })
+  } catch (error) {
+    throw error
+  }
```
bcaudan: do we need to catch it then?

scripts/generate-changelog.js
bcaudan: ok no big deal

CHANGELOG.md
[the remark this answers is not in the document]
glorieux: OK will do.

scripts/generate-changelog.js
glorieux: Done

scripts/generate-changelog.js
glorieux: Thanks I've actually found a small bug in my initial implementation.

scripts/generate-changelog.js
glorieux: Nope. Removing it.

scripts/generate-changelog.js
glorieux: Done

scripts/generate-changelog.js
```diff
@@ -0,0 +1,66 @@
...
+async function emojiNameToUnicode(changes) {
+  const emojiNameRegex = new RegExp(/:[^:\s]*(?:::[^:\s]*)*:/, 'gm')
+
+  let matches
+  while ((matches = emojiNameRegex.exec(changes))) {
+    if (!!matches) {
+      await matches.map((match) => {
+        changes = changes.replace(match, name.get(match) || match)
+      })
+    }
```
bcaudan: FMU, it should loop forever if there is a missing emoji in the map. Am I missing something?

scripts/generate-changelog.js
```diff
@@ -0,0 +1,66 @@
+'use strict'
+
+const replace = require('replace-in-file')
+const util = require('util')
+const exec = util.promisify(require('child_process').exec)
+const spawn = util.promisify(require('child_process').spawn)
+
+const name = require('emoji-name-map')
```
bcaudan:
```suggestion
const emojiNameMap = require('emoji-name-map')
```
in order to be clearer when using it

scripts/generate-changelog.js
```diff
@@ -0,0 +1,66 @@
...
+const lernaConfig = require('../lerna.json')
+
+const CHANGELOG_FILE = 'CHANGELOG.md'
+
+async function main() {
+  const lastTagHash = await executeCommand('git rev-list --tags --max-count=1')
+  const lastTagName = await executeCommand(`git describe --tags ${lastTagHash}`)
+
+  const oldTag = lastTagName.replace(/\n$/, '')
+  const commits = await executeCommand(`git log ${oldTag}..HEAD --pretty=format:"- %s"`)
```
bcaudan: I am not sure to understand why the `\n` is an issue for the log command but not for the describe 🤔 

If it is needed, could we inline `oldTag` variable or name it to be consistent with `lastTagXXX`?

glorieux: The `\n` is always an issue if you pass it to the next command.
The describe command returns the tagName with an extra newline.

BenoitZugmeyer: Also, just use `.trimEnd()` ^^

scripts/generate-changelog.js
BenoitZugmeyer: It will not loop forever, because `emojiNameRegex` is stateful and will start its execution after its `lastIndex` property. But still, this loop should be changed:
* No need to be async
* No need to check for `!!matches` since it wont execute the loop body if `matches` were falsy
* No need to use `.map` since we don't return anything (you could use `matches[0]` since there is only one element in matches)
* Mutating a string while iterating over it using a regex will probably lead to bugs

BenoitZugmeyer: Why not just using
```js
return changes.replace(emojiNameRegex, (emoji) => name.get(match) || match)
```

CHANGELOG.md
```diff
@@ -0,0 +1,188 @@
...
+## v1.7.5
+
+- 🔊[RUMF-408] add new session check logs ([#318](https://github.com/DataDog/browser-sdk/pull/318))
+- [RUMF-407] improve resource timings collection ([#315](https://github.com/DataDog/browser-sdk/pull/315))
+- 🔧 improve CBT test names ([#314](https://github.com/DataDog/browser-sdk/pull/314))
+- [RUMF-382] prepare support for multiple feature flags ([#312](https://github.com/DataDog/browser-sdk/pull/312))
+- 🔧 update cbt chrome mobile ([#313](https://github.com/DataDog/browser-sdk/pull/313))
+- v1.7.5 ([#311](https://github.com/DataDog/browser-sdk/pull/311))
+
+## v1.7.5
```
BenoitZugmeyer: This version is here twice

CHANGELOG.md
```diff
@@ -0,0 +1,188 @@
+# Changelog
+
+## v1.7.5
+
+- 🔊[RUMF-408] add new session check logs ([#318](https://github.com/DataDog/browser-sdk/pull/318))
+- [RUMF-407] improve resource timings collection ([#315](https://github.com/DataDog/browser-sdk/pull/315))
+- 🔧 improve CBT test names ([#314](https://github.com/DataDog/browser-sdk/pull/314))
+- [RUMF-382] prepare support for multiple feature flags ([#312](https://github.com/DataDog/browser-sdk/pull/312))
+- 🔧 update cbt chrome mobile ([#313](https://github.com/DataDog/browser-sdk/pull/313))
+- v1.7.5 ([#311](https://github.com/DataDog/browser-sdk/pull/311))
```
BenoitZugmeyer: This commit is probably not needed

glorieux: Yep but those will have to be manually removed.

scripts/generate-changelog.js
bcaudan: > It will not loop forever, because emojiNameRegex is stateful and will start its execution after its lastIndex property. 

What do you mean by lastIndex?
My point was not about the regex but the fact that we are iterating on emoji matches while we can fail to replace them all (`replace(match, name.get(match) || match)`)

BenoitZugmeyer: A loop like this will always finish at some point, because the regex object keeps an ever-increasing index and will return null at some point:

```
const s = "aa";
const re = /a/g;

re.lastIndex // === 0
re.exec(s) // ["a"]
re.lastIndex // === 1
re.exec(s) // ["a"]
re.lastIndex // === 2
re.exec(s) // null -> this will break the loop
re.lastIndex // === 0
```

bcaudan: I was not aware of this api, good to know 👍

scripts/generate-changelog.js
```diff
@@ -0,0 +1,55 @@
...
+
+async function main() {
+  const lastTagHash = await executeCommand('git rev-list --tags --max-count=1')
+  const lastTagName = await executeCommand(`git describe --tags ${lastTagHash}`)
+
+  const commits = await executeCommand(`git log ${lastTagName.trimEnd()}..HEAD --pretty=format:"- %s"`)
+
+  const changesWithEmojis = await emojiNameToUnicode(`# Changelog\n\n## v${lernaConfig.version}\n\n${commits}`)
+
+  const changesWhihPullRequestLinks = changesWithEmojis.replace(
```
bcaudan:
```suggestion
  const changesWithPullRequestLinks = changesWithEmojis.replace(
```

PR #320, opened 2020-03-26
🐛fix feature check

bcaudan:
- array.include not supported on IE
- ensure enableExperimentalFeatures is an array to avoid `Object doesn't support property or method 'includes'`

### DIFF
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -2,7 +2,6 @@ import { BuildEnv, Datacenter, Environment } from './init'
 import { includes, ONE_KILO_BYTE, ONE_SECOND } from './utils'
 
 export const DEFAULT_CONFIGURATION = {
-  enableExperimentalFeatures: [] as string[],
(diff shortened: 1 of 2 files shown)


packages/core/src/configuration.ts
```diff
@@ -74,6 +76,12 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
   }
 
   const configuration: Configuration = {
+    isEnabled: (feature: string) => {
+      return (
+        Array.isArray(configuration.enableExperimentalFeatures) &&
+        includes(configuration.enableExperimentalFeatures, feature)
```
BenoitZugmeyer: It could be cleaner to not expose `configuration.enableExperimentalFeatures` and use `userConfiguration` directly here
```suggestion
        Array.isArray(userConfiguration.enableExperimentalFeatures) &&
        includes(userConfiguration.enableExperimentalFeatures, feature)
```

PR #327, opened 2020-04-01
🚧[RUMF-430] new cookie format strategy

bcaudan:
This PR is meant to be split in two separate PRs (github messed up the commit order display):

1. handle potential rollback

Capture d’écran 2020-04-01 à 12 04 31

2. new session format

Capture d’écran 2020-04-01 à 12 04 36

All potential feedbacks on this PR should then be squashed with the above commits before opening the two target PRs.


packages/core/src/oldCookiesMigration.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  if (!isValidSessionString(sessionString)) {
+    const session: SessionState = {}
+    if (sessionString) {
+      session.id = sessionString
+    }
+    if (logsType && /^[01]$/.test(logsType)) {
+      session[LOGS_SESSION_KEY] = logsType
+    }
+    if (rumType && /^[012]$/.test(rumType)) {
```
glorieux, line 31: Is there a way you could explain what those regex are matching for?

bcaudan: It is the allowed session types for logs and rum, ex: for [rum](https://github.com/DataDog/browser-sdk/blob/master/packages/rum/src/rumSession.ts#L12)

PR #338, opened 2020-04-03
Automatic click user action collection

BenoitZugmeyer:
This PR implements the automatic click user action collection based on the [previous POC](https://github.com/DataDog/browser-sdk/tree/benoit/poc-click-user-action).

For now, it is behind the `collect-user-actions` feature flag.

Future steps will include:

* Count errors, long tasks and resources in the user action event (in `user_action.measures`)
* Implement the "load view" user action, and discard the current "click" user action if a new view is accessed.


packages/rum/test/userActionCollection.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+      }
+
+      clock.expire()
+    })
+
+    it('expires after a limit', (done) => {
+      const changeObservable = new Observable<ChangeEvent>()
+      let stop = false
+      newUserAction(changeObservable, (details) => {
+        expect(details!.duration).toBe(10_000)
```
bcaudan, line 115: What about reusing the constant?

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+  beforeEach(() => {
+    events.length = 0
+  })
+  return {
+    events,
+    pushEvent(event: T) {
+      events.push(event)
+    },
+  }
+}
```
bcaudan, line 39: What about replacing `event` by `userAction`?

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,160 @@
...
+    finishCallback(undefined)
+    return
+  }
+
+  let idleTimeoutId: ReturnType<typeof setTimeout>
+  const id = generateUUID()
+  const startTime = performance.now()
+  let hasFinished = false
+
+  const validationTimeoutId = setTimeout(() => finish(false), BUSY_DELAY)
```
bcaudan, line 30: We should `monitor` all the setTimeout callbacks

packages/core/src/requestCollection.ts
```diff
@@ -58,13 +78,20 @@ export function trackXhr(observable: RequestObservable) {
   const originalSend = XMLHttpRequest.prototype.send
   XMLHttpRequest.prototype.send = function(this: BrowserXHR, body: unknown) {
     const startTime = performance.now()
+    const requestId = getNextRequestId()
+
+    observables.start.notify({
```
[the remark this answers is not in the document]
BenoitZugmeyer: I don't really mind. If the `start` property seems not explicit enough to you, would you prefer something like this?
```ts
const [ requestStartObservable, requestEndObservable ] = ...
```

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    const content = getElementContent(event.target)
+
+    const { observable: changesObservable, stop: stopChangesTracking } = trackPageChanges(lifeCycle)
+
+    newUserAction(changesObservable, (userActionDetails) => {
+      stopChangesTracking()
+      if (userActionDetails) {
+        lifeCycle.notify(LifeCycleEventType.USER_ACTION_COLLECTED, {
```
bcaudan, line 142: > Future steps will include:
> - Count user actions in the view event (view.measures.user_action_count)

it should already be handled then, no?

packages/rum/test/domMutationCollection.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  it(
+    'collects DOM mutation on Element',
+    domMutationSpec((root) => {
+      root.appendChild(document.createElement('button'))
+    }, 1)
+  )
+
+  it(
+    'collects DOM mutation on Text creation',
```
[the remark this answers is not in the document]
BenoitZugmeyer: This is not an Element, it's a text node.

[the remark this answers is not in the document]
bcaudan: yes but both are nodes, does mutationObserver behave differently for a text node?

packages/core/src/requestCollection.ts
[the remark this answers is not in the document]
bcaudan: yes, it should be great 👍

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+function newUserAction(
+  changesObservable: Observable<ChangeEvent>,
+  finishCallback: (event: UserActionEnded | undefined) => void
+) {
+  if (currentUserAction) {
+    finishCallback(undefined)
+    return
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: You're right, at first, the POC was replacing user actions, but has been changed after discussion with the product, to be closer to the product brief:
> If a user does an action while another action is happening, the second action is ignored

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,160 @@
+import { generateUUID, Observable } from '@datadog/browser-core'
+import { getElementContent } from './getElementContent'
+import { LifeCycle, LifeCycleEventType, Subscription } from './lifeCycle'
+import { UserActionType } from './rum'
+
+const IDLE_DELAY = 100
+const BUSY_DELAY = 100
```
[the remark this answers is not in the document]
BenoitZugmeyer: (renamed to `USER_ACTION_VALIDATION_DELAY` to be in line with other constants)

packages/rum/src/userActionCollection.ts
BenoitZugmeyer: True. This is listed as a new attribute is the "User actions data format" document so I thought we needed to do something about it.

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,160 @@
...
+  changesObservable.subscribe(({ isBusy }) => {
+    clearTimeout(validationTimeoutId)
+    clearTimeout(idleTimeoutId)
+    const time = performance.now()
+    if (!isBusy) {
+      idleTimeoutId = setTimeout(() => finish(true, time), IDLE_DELAY)
+    }
+  })
+
+  function finish(end: boolean, time = performance.now()) {
```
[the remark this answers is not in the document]
BenoitZugmeyer: Making two functions would not be ideal because they would share most of the same code. Instead, I changed the `finish` function like this: `finish(details: UserActionDetails | undefined)` to call the `finishCallback` with it. The logic is a bit more straightforward.

packages/rum/test/userActionCollection.spec.ts
BenoitZugmeyer: That's not necessarily user actions, I use it for page activity events too.

packages/rum/src/userActionCollection.ts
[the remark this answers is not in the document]
bcaudan: oh ok, could you add the rule as a comment?

packages/rum/test/domMutationCollection.spec.ts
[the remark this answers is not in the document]
BenoitZugmeyer: No, but from the unit test point of view we don't know about the mutation observer :)

packages/core/src/errorCollection.ts
```diff
@@ -165,24 +164,24 @@ export function toStackTraceString(stack: StackTrace) {
 export function trackNetworkError(
   configuration: Configuration,
   errorObservable: ErrorObservable,
-  requestObservable: RequestObservable
+  requestObservable: Observable<RequestCompleteEvent>
 ) {
-  requestObservable.subscribe((request: RequestDetails) => {
-    if (isRejected(request) || isServerError(request)) {
+  requestObservable.subscribe((completeEvent: RequestCompleteEvent) => {
```
bcaudan, line 169: In all contexts where we use `completeEvent`, it would seem more appropriate to me to use `request`

packages/core/test/requestCollection.spec.ts
```diff
@@ -169,184 +177,177 @@ describe('fetch tracker', () => {
...
-            startTime: jasmine.any(Number),
-            traceId: undefined,
-            type: 'xhr',
-          })
-        )
-        expectations(xhr)
+        const completeEvent = (completeSpy.calls.allArgs() as RequestCompleteEvent[][])
+          .map((args) => args[0])
+          // IE10/11 doesn't have .find()
+          .filter((d) => includes(d.url, expectedURL))[0]
```
bcaudan, line 223: we probably don't need to add that comment everywhere we use some of the utils polyfills

packages/rum/test/domMutationCollection.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      document.body.appendChild(root)
+
+      const lifeCycle = new LifeCycle()
+
+      let counter = 0
+      lifeCycle.subscribe(LifeCycleEventType.DOM_MUTATED, () => {
+        counter += 1
+      })
+
+      startDOMMutationCollection(lifeCycle)
```
bcaudan, line 31: Should we stop the mutation observer after each test?

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,200 @@
...
+    currentUserAction = undefined
+    finishCallback(details)
+  }
+}
+
+export interface PageActivityEvent {
+  isBusy: boolean
+}
+
+function trackPagePageActivities(lifeCycle: LifeCycle): { observable: Observable<PageActivityEvent>; stop(): void } {
```
bcaudan, line 142:
```suggestion
function trackPageActivities(lifeCycle: LifeCycle): { observable: Observable<PageActivityEvent>; stop(): void } {
```

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,200 @@
...
+let currentUserAction: { id: string; startTime: number } | undefined
+
+export function getUserActionId(time: number): string | undefined {
+  if (currentUserAction && time >= currentUserAction.startTime) {
+    return currentUserAction.id
+  }
+  return undefined
+}
+
+interface UserActionDetails {
```
bcaudan, line 83: What about `UserActionCompleteEvent` in order to be more consistent with other namings?
then, we we could rename `createDetails` into something like `completeUserAction`.

packages/rum/src/userActionCollection.ts
```diff
@@ -0,0 +1,200 @@
...
+    finishCallback(details)
+  }
+}
+
+export interface PageActivityEvent {
+  isBusy: boolean
+}
+
+function trackPagePageActivities(lifeCycle: LifeCycle): { observable: Observable<PageActivityEvent>; stop(): void } {
+  const result = new Observable<PageActivityEvent>()
```
bcaudan, line 143:
```suggestion
  const observable = new Observable<PageActivityEvent>()
```

packages/rum/src/domMutationCollection.ts
```diff
@@ -0,0 +1,20 @@
...
+
+interface BrowserWindow extends Window {
+  MutationObserver?: MutationObserver
+}
+
+export function startDOMMutationCollection(lifeCycle: LifeCycle) {
+  if ((window as BrowserWindow).MutationObserver) {
+    const observer = new MutationObserver(() => {
+      lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    })
```
bcaudan, line 11: we should add a monitor on this callback

packages/rum/src/rum.ts
```diff
@@ -88,6 +106,7 @@ export interface RumResourceEvent {
     kind: ResourceKind
   }
   traceId?: number
+  userActionId?: string
```
bcaudan, line 109: Since this attribute does not exist yet, I wonder if it would not be preferable to store it as `userAction.id` in order to be aligned with future format

packages/core/test/requestCollection.spec.ts
BenoitZugmeyer: This is for the `find()` method, not `includes()`. Maybe I could add a util `find()` function?

packages/rum/src/rum.ts
BenoitZugmeyer: This attribute is the one specified in the "User actions data format". What do you mean by "future format"?

packages/core/test/requestCollection.spec.ts
bcaudan: oh ok, yes adding find coud be handy

packages/rum/src/rum.ts
bcaudan: actually, there is still some discussion on the naming, let's stick with this one for now

PR #343, opened 2020-04-07
✨[RUMF-430] new session strategy

bcaudan:
# Motivations

- Allow page with iframe displayed to continue to expand the session even if there is no interaction with the main page
- Reduce long lasting sessions

# Changes

Behind the 'new-session' flag:
- Check page visibility every 1 minute to expand the session
- Add a session timeout at 4h
- Enforce the session expiration if last interaction was more than 15 minutes ago

### DIFF
--- a/packages/core/src/sessionManagement.ts
+++ b/packages/core/src/sessionManagement.ts
@@ -94,6 +94,7 @@ function retrieveActiveSession(sessionCookie: CookieCache, withNewSessionStrateg
   if (!withNewSessionStrategy || isActiveSession(session)) {
     return session
   }
+  // clear session
   const inactiveSession = {}
   persistSession(inactiveSession, sessionCookie, withNewSessionStrategy)
   return inactiveSession
@@ -124,7 +125,12 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
 }
 
 export function persistSession(session: SessionState, cookie: CookieCache, withNewSessionStrategy = false) {
-  if (withNewSessionStrategy && !utils.isEmptyObject(session)) {
+  if (utils.isEmptyObject(session)) {
+    // clear session
+    cookie.set('', 0)
+    return
+  }
+  if (withNewSessionStrategy) {
     session.expire = String(Date.now() + SESSION_EXPIRATION_DELAY)
   }
   const cookieString = utils
--- a/packages/core/src/sessionManagement.ts
+++ b/packages/core/src/sessionManagement.ts
@@ -1,4 +1,5 @@
 import { cacheCookieAccess, COOKIE_ACCESS_DELAY, CookieCache } from './cookie'
+import { monitor } from './internalMonitoring'
 import { Observable } from './observable'
 import { tryOldCookiesMigration } from './oldCookiesMigration'
 import * as utils from './utils'
@@ -105,9 +106,10 @@ export function stopSessionManagement() {
 let registeredActivityListeners: Array<() => void> = []
 
 export function trackActivity(expandOrRenewSession: () => void) {
+  const doExpandOrRenewSession = monitor(expandOrRenewSession)
   const options = { capture: true, passive: true }
   ;['click', 'touchstart', 'keydown', 'scroll'].forEach((event: string) => {
-    document.addEventListener(event, expandOrRenewSession, options)
-    registeredActivityListeners.push(() => document.removeEventListener(event, expandOrRenewSession, options))
+    document.addEventListener(event, doExpandOrRenewSession, options)
+    registeredActivityListeners.push(() => document.removeEventListener(event, doExpandOrRenewSession, options))
   })
 }
--- a/packages/core/src/sessionManagement.ts
+++ b/packages/core/src/sessionManagement.ts
@@ -147,10 +147,12 @@ let registeredIntervals: number[] = []
 export function trackActivity(expandOrRenewSession: () => void) {
   const doExpandOrRenewSession = monitor(expandOrRenewSession)
   const options = { capture: true, passive: true }
-  ;['click', 'touchstart', 'keydown', 'scroll'].forEach((event: string) => {
-    document.addEventListener(event, doExpandOrRenewSession, options)
-    registeredListeners.push(() => document.removeEventListener(event, doExpandOrRenewSession, options))
-  })
+  ;[utils.DOM_EVENT.CLICK, utils.DOM_EVENT.TOUCH_START, utils.DOM_EVENT.KEY_DOWN, utils.DOM_EVENT.SCROLL].forEach(
+    (event: string) => {
+      document.addEventListener(event, doExpandOrRenewSession, options)
+      registeredListeners.push(() => document.removeEventListener(event, doExpandOrRenewSession, options))
+    }
+  )
 }
 
 function trackVisibility(expandSession: () => void, visibilityStateProvider: () => VisibilityState) {
@@ -161,8 +163,10 @@ function trackVisibility(expandSession: () => void, visibilityStateProvider: () 
   })
 
   const visibilityCheckInterval = window.setInterval(expandSessionWhenVisible, VISIBILITY_CHECK_DELAY)
-  document.addEventListener('visibilitychange', expandSessionWhenVisible)
+  document.addEventListener(utils.DOM_EVENT.VISIBILITY_CHANGE, expandSessionWhenVisible)
 
   registeredIntervals.push(visibilityCheckInterval)
-  registeredListeners.push(() => document.removeEventListener('visibilitychange', expandSessionWhenVisible))
+  registeredListeners.push(() =>
+    document.removeEventListener(utils.DOM_EVENT.VISIBILITY_CHANGE, expandSessionWhenVisible)
+  )
 }
--- a/packages/core/src/sessionManagement.ts
+++ b/packages/core/src/sessionManagement.ts
@@ -93,10 +93,8 @@ function retrieveActiveSession(sessionCookie: CookieCache, withNewSessionStrateg
   if (!withNewSessionStrategy || isActiveSession(session)) {
     return session
   }
-  // clear session
-  const inactiveSession = {}
-  persistSession(inactiveSession, sessionCookie, withNewSessionStrategy)
-  return inactiveSession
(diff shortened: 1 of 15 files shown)


packages/core/src/sessionManagement.ts
```diff
@@ -1,22 +1,24 @@
 import { cacheCookieAccess, COOKIE_ACCESS_DELAY, CookieCache } from './cookie'
+import { monitor } from './internalMonitoring'
 import { Observable } from './observable'
 import { tryOldCookiesMigration } from './oldCookiesMigration'
 import * as utils from './utils'
 
 export const SESSION_COOKIE_NAME = '_dd_s'
-export const EXPIRATION_DELAY = 15 * utils.ONE_MINUTE
+export const SESSION_EXPIRATION_DELAY = 15 * utils.ONE_MINUTE
```
glorieux, line 8: Wouldn't it make sense to move those time related constants to a module called `time` instead of `utils`?

packages/core/src/sessionManagement.ts
```diff
@@ -89,25 +123,46 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
...
+    if (visibilityStateProvider() === 'visible') {
+      expandSession()
+    }
   })
+
+  const visibilityCheckInterval = window.setInterval(expandSessionWhenVisible, VISIBILITY_CHECK_DELAY)
+  document.addEventListener('visibilitychange', expandSessionWhenVisible)
+
+  registeredIntervals.push(visibilityCheckInterval)
+  registeredListeners.push(() => document.removeEventListener('visibilitychange', expandSessionWhenVisible))
```
glorieux, line 167: Could `'visibilitychange'` be a constant?

packages/core/src/sessionManagement.ts
bcaudan: IMO, as long as we don't duplicate these values everywhere it is fine.
utils is not that big for now, if we feel the need to split this file or to extract some of the functions at some point, we could do it.

packages/core/src/sessionManagement.ts
bcaudan: sure, what is your reasoning?
- avoid to duplicate this value inside this function
- have constant for all events we are listening

glorieux: I'd say both 😄

glorieux: But in the scope of that PR only inside that function.

packages/core/src/sessionManagement.ts
```diff
@@ -89,25 +123,46 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
...
   const options = { capture: true, passive: true }
   ;['click', 'touchstart', 'keydown', 'scroll'].forEach((event: string) => {
-    document.addEventListener(event, expandOrRenewSession, options)
-    registeredActivityListeners.push(() => document.removeEventListener(event, expandOrRenewSession, options))
+    document.addEventListener(event, doExpandOrRenewSession, options)
+    registeredListeners.push(() => document.removeEventListener(event, doExpandOrRenewSession, options))
+  })
+}
+
+function trackVisibility(expandSession: () => void, visibilityStateProvider: () => VisibilityState) {
```
mquentin, line 156: Could we have a `isVisible()` kind of function instead ? If we somehow need to  adapt the `Document.visibilityState` API for some specific browser, maybe it will be easier to have this logic in a separate function. Also I find it easier to read.

bcaudan: > if we somehow need to adapt the Document.visibilityState API for some specific browser, maybe it will be easier to have this logic in a separate function.

Let's worry about that if we encounter this kind of issue. We would then have more insight on how to solve it.

> I find it easier to read

Update the function to use a isVisible method instead would also mean to move the `'visible'`check outside of this function which seems really appropriate to me here.

Do you have another suggestion that could keep this check and make that more legible?

mquentin: all good then

packages/core/src/sessionManagement.ts
```diff
@@ -49,15 +56,23 @@ export function startSessionManagement<Type extends string>(
     }
   }, COOKIE_ACCESS_DELAY)
 
+  const expandSession = () => {
```
BenoitZugmeyer, line 59: This function could be in the scope where it's used (inside the `if`) or even inlined.

packages/core/src/sessionManagement.ts
```diff
@@ -74,6 +89,25 @@ export function isValidSessionString(sessionString: string | undefined): session
   )
 }
 
+function retrieveActiveSession(sessionCookie: CookieCache, withNewSessionStrategy: boolean): SessionState {
+  const session = retrieveSession(sessionCookie)
+  if (!withNewSessionStrategy || isActiveSession(session)) {
+    return session
+  }
+  const inactiveSession = {}
+  persistSession(inactiveSession, sessionCookie, withNewSessionStrategy)
```
BenoitZugmeyer, line 98: I don't think we should persist empty sessions. Else, we should document it

packages/core/src/sessionManagement.ts
```diff
@@ -89,25 +123,50 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
   return session
 }
 
-export function persistSession(session: SessionState, cookie: CookieCache) {
+export function persistSession(session: SessionState, cookie: CookieCache, withNewSessionStrategy = false) {
+  if (withNewSessionStrategy && !utils.isEmptyObject(session)) {
```
BenoitZugmeyer, line 127: Document why we don't store `expire` in an empty session

packages/core/src/sessionManagement.ts
```diff
@@ -89,25 +123,50 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
...
-    registeredActivityListeners.push(() => document.removeEventListener(event, expandOrRenewSession, options))
+  ;[utils.DOM_EVENT.CLICK, utils.DOM_EVENT.TOUCH_START, utils.DOM_EVENT.KEY_DOWN, utils.DOM_EVENT.SCROLL].forEach(
+    (event: string) => {
+      document.addEventListener(event, doExpandOrRenewSession, options)
+      registeredListeners.push(() => document.removeEventListener(event, doExpandOrRenewSession, options))
+    }
+  )
+}
+
+function trackVisibility(expandSession: () => void, visibilityStateProvider: () => VisibilityState) {
```
BenoitZugmeyer, line 158: IMO
* `trackVisibility` should expose an `Observable` instead of a callback function
* be self contained: no `visibilityStateProvider`, use `document.visibilityState` directly. We can mock `document.visibilityState` in tests

packages/core/src/sessionManagement.ts
```diff
@@ -89,25 +123,50 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
...
+  const expandSessionWhenVisible = monitor(() => {
+    if (visibilityStateProvider() === 'visible') {
+      expandSession()
+    }
   })
+
+  const visibilityCheckInterval = window.setInterval(expandSessionWhenVisible, VISIBILITY_CHECK_DELAY)
+  document.addEventListener(utils.DOM_EVENT.VISIBILITY_CHANGE, expandSessionWhenVisible)
+
+  registeredIntervals.push(visibilityCheckInterval)
```
BenoitZugmeyer, line 168: I would rename `registeredListeners` to `stopCallbacks` or `stopSessionManagementCallbacks`, and use only one array:
```suggestion
  stopSessionManagementCallbacks.push(() => clearInterval(visibilityCheckInterval))
```

packages/core/src/sessionManagement.ts
bcaudan: By inlining it, we would lost the intention (renew session) and keeping the function instead of adding a comment make it more consistent with the `expandOrRenewSession` function.
When we will remove the if, it would end up at the same position as it is actually.
wdyt?

BenoitZugmeyer: All right

packages/core/src/sessionManagement.ts
bcaudan: > trackVisibility should expose an Observable instead of a callback function

what advantage do you see with it?

> We can mock document.visibilityState in tests

How would you do that?

BenoitZugmeyer: > what advantage do you see with it?

I think having a common strategy to pass data streams around would make things more future-proof. For now, our Observable implementation is quite limited, but in the future it could be extendend with a "destructor" when unused ([example](https://github.com/zenparsing/zen-observable/blob/master/README.md#new-observablesubscribe)), and we would have a common way to remove listeners and stuff. This would make functions like this more self-contained and easier to test.

But if you feel that this is thinking too far ahead, I'm okay to keep a callback for now :)

> How would you do that?

```js
Object.defineProperty(document, 'visibilityState', { get() { return "zog" }, configurable: true })
// then
delete document.visibilityState
```

packages/core/src/sessionManagement.ts
```diff
@@ -124,7 +125,12 @@ function retrieveSession(sessionCookie: CookieCache): SessionState {
 }
 
 export function persistSession(session: SessionState, cookie: CookieCache, withNewSessionStrategy = false) {
-  if (withNewSessionStrategy && !utils.isEmptyObject(session)) {
+  if (utils.isEmptyObject(session)) {
+    // clear session
+    cookie.set('', 0)
+    return
+  }
```
BenoitZugmeyer: Thanks, now it's clearer to me :) But I think it would make things even clearer to have a separate function `clearSession`.

packages/core/src/sessionManagement.ts
bcaudan: About observable instead of callbacks, I am not sure that observable pattern should be enforce in every situation.
I think it:
- allow us to easily reuse a data source in another part of the code
- can help us to decouple some of the browser API to the domain logic 
But it also adds an extra layer of abstraction.
About future-proof arguments, let's see when we will have the need to do more complicated stuff with our observable and address it at this time.

In this case, I am not sure to see much benefit to switch to observable.

PR #345, opened 2020-04-07
🐛[RUMF-431] fix CSP issue with global object strategy

### LINKED ISSUE #328: getGlobalObject causes CSP violations on older browsers
Due to the use of `Function` in https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/utils.ts#L224, the browser agent causes errors when `unsafe-eval` is disallowed by CSP for older browsers without `globalThis` (e.g. Chrome <71, Firefox <65).

Suggest also checking for the existence of `window` (and possibly `self` for webworkers) in addition to `globalThis` to avoid invoking the Function constructor in the vast majority of cases.

bcaudan:
Unless adding `unsafe-eval`, Function is blocked by CSP restriction on older browsers.
Use black magic instead 🙈

fixes #328

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -237,9 +237,20 @@ export function getGlobalObject<T>(): T {
     configurable: true,
   })
   // @ts-ignore
-  const globalObject = _dd_temp_
+  let globalObject: unknown = _dd_temp_
   // @ts-ignore
   delete Object.prototype._dd_temp_
+  if (typeof globalObject !== 'object') {
+    // on safari _dd_temp_ is available on window but not globally
+    // fallback on other browser globals check
+    if (typeof self === 'object') {
+      globalObject = self
+    } else if (typeof window === 'object') {
+      globalObject = window
+    } else {
+      globalObject = {}
+    }
+  }
   return globalObject as T
 }
 
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -223,9 +223,24 @@ export function objectEntries(object: { [key: string]: unknown }) {
   return Object.keys(object).map((key) => [key, object[key]])
 }
 
+/**
+ * inspired by https://mathiasbynens.be/notes/globalthis
+ */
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer:
I wonder why it fails on Mobile Safari! I think we could just check for `globalThis` `window` `self` and `global` for now?

bcaudan:
> I wonder why it fails on Mobile Safari! I think we could just check for `globalThis` `window` `self` and `global` for now?

I am looking into it 🙂

packages/core/src/utils.ts
```diff
@@ -223,9 +223,35 @@ export function objectEntries(object: { [key: string]: unknown }) {
...
+    },
+    configurable: true,
+  })
+  // @ts-ignore
+  let globalObject: unknown = _dd_temp_
+  // @ts-ignore
+  delete Object.prototype._dd_temp_
+  if (typeof globalObject !== 'object') {
+    // on safari _dd_temp_ is available on window but not globally
+    // fallback on other browser globals check
```
bcaudan, line 245: I've preferred to add this fallback instead of only checking all the possible globals, since the above method should be more robust.

BenoitZugmeyer:
Code looks OK but I'd still prefer not to use the "black magic" strategy. I mean, even the popular `core-js` polyfill is way simpler: https://github.com/zloirock/core-js/blob/a431ac800cebcef071ff4636c45eb4fcab361d1a/packages/core-js/internals/global.js#L6

bcaudan:
> Code looks OK but I'd still prefer not to use the "black magic" strategy. I mean, even the popular `core-js` polyfill is way simpler: https://github.com/zloirock/core-js/blob/a431ac800cebcef071ff4636c45eb4fcab361d1a/packages/core-js/internals/global.js#L6

I agree that it is simpler but I would prefer to avoid falling into another global trap later and need to rework on it.
Wdyt?

BenoitZugmeyer:
Ok :) IMO we will be fine in either case anyway.

PR #346, opened 2020-04-08
[RUM] User action collection adjustements

BenoitZugmeyer:
After a few tests on staging, user actions needs two adjustments:

* change the data model so `user_action_id` becomes `user_action.id` on all RUM events
* convert the duration in nanoseconds to be in line with other rum events

### DIFF
--- a/packages/rum/src/rum.ts
+++ b/packages/rum/src/rum.ts
@@ -135,6 +135,7 @@ export interface RumViewEvent {
 }
 
 export interface RumLongTaskEvent {
+  date: number
   duration: number
   evt: {
     category: RumEventCategory.LONG_TASK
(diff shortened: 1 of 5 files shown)


packages/rum/src/rum.ts
```diff
@@ -106,17 +106,18 @@ export interface RumResourceEvent {
     kind: ResourceKind
   }
   traceId?: number
-  userActionId?: string
+  userAction?: UserActionReference
 }
 
 export interface RumErrorEvent {
+  date: number
```
bcaudan, line 113: Then, could you also make this change:
```ts
function trackErrors(lifeCycle: LifeCycle, addRumEvent: (event: RumErrorEvent) => void) {
```

PR #351, opened 2020-04-09
👷 add a typecheck job to the CI

BenoitZugmeyer:
Other jobs are not typechecking all TS source files, or at least not failing when there is a TS error:

* `yarn run build`: will fail if there is an typecheck error on files included in the bundle, but not on test files
* `yarn run test` (and any test executed by karma): karma will not fail if there is a typecheck error in files, and I didn't find a way to configure it otherwise
* `yarn run lint`: tslint does not typecheck the code

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,12 +39,21 @@ lint:
     - yarn
     - yarn lint
 
+typecheck:
+  stage: test
+  tags: ['runner:main', 'size:large']
+  image: $CI_IMAGE
+  script:
+    - yarn
+    - yarn typecheck
+
 build:
   stage: test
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -39,12 +39,21 @@ lint:
...
+    - yarn
+    - yarn typecheck
+
 build:
   stage: test
   tags: ['runner:main', 'size:large']
   image: $CI_IMAGE
   script:
     - yarn
+    - yarn typecheck
```
bcaudan, line 56: since there is a dedicated typecheck step, do we need to also add it here?

BenoitZugmeyer: Wops that was unintended. Removing it.

PR #352, opened 2020-04-09
✨ [RUMF-438] add user action reference to the internal context

BenoitZugmeyer:
Logs and traces will automatically pick up and associate this to their sent data.

### DIFF
--- a/packages/rum/src/rum.ts
+++ b/packages/rum/src/rum.ts
@@ -29,7 +29,7 @@ import {
   computeSize,
   isValidResource,
 } from './resourceUtils'
-import { RumGlobal } from './rum.entry'
+import { InternalContext, RumGlobal } from './rum.entry'
 import { RumSession } from './rumSession'
 import { getUserActionReference, UserActionReference } from './userActionCollection'
 import { trackView, viewContext, ViewMeasures } from './viewTracker'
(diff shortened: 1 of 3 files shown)


packages/rum/src/userActionCollection.ts
```diff
@@ -74,11 +74,16 @@ let currentUserAction: { id: string; startTime: number } | undefined
...
-    return { id: currentUserAction.id }
+export function getUserActionReference(time?: number): UserActionReference | undefined {
+  if (!currentUserAction) {
+    return undefined
   }
-  return undefined
+
+  if (time !== undefined && time < currentUserAction.startTime) {
+    return undefined
+  }
```
bcaudan, line 84: what about grouping these checks?

BenoitZugmeyer: I did at first, but I think this makes the function more readable. WDYT?

bcaudan: IMO:
```ts
if (!currentUserAction || (time !== undefined && time < currentUserAction.startTime)) {
```
has the same effect, you need understand both parts separately.
But having two if statements following each other with the same body bugged me a little, it does not seem very natural.
if you think that we should make the condition clearer, maybe extracting a function for the time predicate could clarify the intent.

BenoitZugmeyer: All right :)

packages/rum/src/rum.ts
```diff
@@ -223,6 +223,7 @@ export function startRum(
       return {
         application_id: applicationId,
         session_id: viewContext.sessionId,
+        user_action: getUserActionReference(),
```
bcaudan, line 226: On my PR (😆), I missed the `InternalContext` type update, what about enforcing it with:
```ts
getInternalContext: monitor((): InternalContext => {
```

PR #354, opened 2020-04-09
✨[RUMF-43] add proxyHost init option

bcaudan:
Send data through a proxy before sending it to datadog servers.
Correct datadog host to forward the data is added to requests in the `ddhost` query parameter

Ex:
By setting `proxyHost` init option to `my.proxy.io`
Requests URL will look like:

https://my.proxy.io/some/path/token?ddhost=service.datadog-domain.com&some=parameters

And must be forwarded to:

https://service.datadog-domain.com/some/path/token?some=parameters

### DIFF
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -43,6 +43,7 @@ export interface UserConfiguration {
   datacenter?: Datacenter
   enableExperimentalFeatures?: string[]
   silentMultipleInit?: boolean
+  proxyHost?: string
 
   // Below is only taken into account for e2e-test bundle.
   internalMonitoringEndpoint?: string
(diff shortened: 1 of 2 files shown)


packages/core/src/configuration.ts
```diff
@@ -126,6 +129,9 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
   const tld = conf.datacenter === 'us' ? 'com' : 'eu'
   const domain = conf.env === 'production' ? `datadoghq.${tld}` : `datad0g.${tld}`
   const tags = `version:${conf.version}`
-  return `https://${type}-http-intake.logs.${domain}/v1/input/${conf.clientToken}?ddsource=${source ||
-    'browser'}&ddtags=${tags}`
+  const datadogHost = `${type}-http-intake.logs.${domain}`
+  const host = conf.proxyHost ? conf.proxyHost : datadogHost
+  const proxyParameter = conf.proxyHost ? `dd-host=${datadogHost}&` : ''
```
glorieux, line 134: All the other query params don't contain an hyphen like `dd-host`.
We should maybe rename it to `ddhost`?

PR #356, opened 2020-04-10
🐛[RUMF-383] fix custom user action type case

bcaudan:
userAction was not snake cased

### DIFF
--- a/packages/rum/src/rum.ts
+++ b/packages/rum/src/rum.ts
@@ -209,7 +209,7 @@ export function startRum(
   trackErrors(lifeCycle, batch.addRumEvent)
   trackRequests(configuration, lifeCycle, session, batch.addRumEvent)
   trackPerformanceTiming(configuration, lifeCycle, batch.addRumEvent)
-  trackCustomUserAction(lifeCycle, batch.addUserEvent)
+  trackCustomUserAction(lifeCycle, batch.addRumEvent)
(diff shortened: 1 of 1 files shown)


packages/rum/src/rum.ts
```diff
@@ -257,9 +257,9 @@ function startRumBatch(
         batch.add(withSnakeCaseKeys(event as Context))
       }
     },
-    addUserEvent: (event: RumUserActionEvent) => {
+    addUserEvent: (event: RumUserActionEvent, context?: Context) => {
```
BenoitZugmeyer, line 260: What about removing `addUserAction` and add this new parameter to `addRumEvent`?

packages/rum/src/rum.ts
```diff
@@ -285,19 +280,24 @@ function trackErrors(lifeCycle: LifeCycle, addRumEvent: (event: RumErrorEvent) =
   })
 }
 
-function trackCustomUserAction(lifeCycle: LifeCycle, addUserEvent: (event: RumUserActionEvent) => void) {
+function trackCustomUserAction(
+  lifeCycle: LifeCycle,
+  addUserEvent: (event: RumUserActionEvent, context?: Context) => void
```
BenoitZugmeyer, line 285: NIT: rename `addUserEvent` to `addRumEvent`?

PR #357, opened 2020-04-10
👷 include the e2e tests in the typecheck job

BenoitZugmeyer:
The e2e scenario isn't typechecked by the typecheck command.


package.json
```diff
@@ -10,7 +10,7 @@
     "build:bundle": "lerna run build:bundle --stream",
     "format": "prettier --check \"**/*.{ts,js,json,md,yml,html}\"",
     "lint": "tslint -c ./tslint.json -p ./tsconfig.json && tslint -c ./tslint.json -p ./test/e2e/scenario/tsconfig.json",
-    "typecheck": "tsc -p . --noEmit true",
+    "typecheck": "tsc -p . --noEmit true && tsc -p test/e2e/scenario --noEmit true",
```
bcaudan, line 13: should we also add test/app also?

bcaudan, line 13: maybe test/app in lint also

PR #361, opened 2020-04-15
[RUMF-447]: Only collect first-contentful-paint if page is visible

glorieux:


packages/rum/test/viewTracker.spec.ts
```diff
@@ -214,5 +216,30 @@ describe('rum view measures', () => {
       resourceCount: 0,
       userActionCount: 0,
     })
+    restorePageVisibility()
+  })
+  it('should not collect firstContentfulPaint if page is not visible', () => {
```
bcaudan, line 221:
```suggestion
  })
  
  it('should not collect firstContentfulPaint if page is not visible', () => {
```

packages/rum/test/viewTracker.spec.ts
```diff
@@ -214,5 +216,30 @@ describe('rum view measures', () => {
...
+    setup({ addRumEvent, lifeCycle })
+    lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED, FAKE_PAINT_ENTRY as PerformancePaintTiming)
+    lifeCycle.notify(
+      LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
+      FAKE_NAVIGATION_ENTRY as PerformanceNavigationTiming
+    )
+    history.pushState({}, '', '/bar')
+
+    expect(getEventCount()).toEqual(3)
+    expect(getViewEvent(1).view.measures).toEqual({
```
bcaudan, line 233:
```suggestion
    expect(getViewEvent(1).view.measures. firstContentfulPaint).toBeUndefined()
```
it would make clearer what we expect

packages/rum/test/viewTracker.spec.ts
```diff
@@ -214,5 +216,30 @@ describe('rum view measures', () => {
...
+  })
+  it('should not collect firstContentfulPaint if page is not visible', () => {
+    setPageVisibility('hidden')
+    const lifeCycle = new LifeCycle()
+    setup({ addRumEvent, lifeCycle })
+    lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED, FAKE_PAINT_ENTRY as PerformancePaintTiming)
+    lifeCycle.notify(
+      LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
+      FAKE_NAVIGATION_ENTRY as PerformanceNavigationTiming
+    )
```
bcaudan, line 229: do we need to check other timings?

packages/rum/src/viewTracker.ts
```diff
@@ -131,7 +131,11 @@ function trackMeasures(lifeCycle: LifeCycle, scheduleViewUpdate: () => void) {
         loadEventEnd: msToNs(navigationEntry.loadEventEnd),
       }
       scheduleViewUpdate()
-    } else if (entry.entryType === 'paint' && entry.name === 'first-contentful-paint') {
+    } else if (
+      entry.entryType === 'paint' &&
+      entry.name === 'first-contentful-paint' &&
+      document.visibilityState === 'visible'
```
bcaudan, line 137: IMO, this check is not enough to prevent first contentful paint to be collected when page become visible
Have you tested it?

packages/rum/test/viewTracker.spec.ts
glorieux: Nope 😄

packages/rum/src/viewTracker.ts
glorieux: Yes. 
I used the scenario described in the "First Contentful Paint in RUM" document.
That check prevented the `first_contentful_paint` from being added the the `measures` when the page is not visible.

packages/rum/src/performanceCollection.ts
```diff
@@ -31,7 +31,12 @@ export function startPerformanceCollection(lifeCycle: LifeCycle, session: RumSes
     const observer = new PerformanceObserver(
       monitor((entries) => handlePerformanceEntries(session, lifeCycle, entries.getEntries()))
     )
-    observer.observe({ entryTypes: ['resource', 'navigation', 'paint', 'longtask'] })
+    const entryTypes = ['resource', 'navigation', 'longtask']
+
+    if (document.visibilityState === 'visible') {
+      entryTypes.push('paint')
+    }
```
bcaudan, line 38:
```suggestion
    // cf https://github.com/w3c/paint-timing/issues/40
    if (document.visibilityState === 'visible') {
      entryTypes.push('paint')
    }
```

packages/rum/test/viewTracker.spec.ts
```diff
@@ -214,5 +216,21 @@ describe('rum view measures', () => {
...
+    setup({ addRumEvent, lifeCycle })
+    lifeCycle.notify(
+      LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
+      FAKE_NAVIGATION_ENTRY as PerformanceNavigationTiming
+    )
+    history.pushState({}, '', '/bar')
+
+    expect(getEventCount()).toEqual(3)
+    expect(getViewEvent(1).view.measures.firstContentfulPaint).toBeUndefined()
+    restorePageVisibility()
```
bcaudan, line 234: is this test still suited to the strategy?
removing visibility hidden would not fail the test

glorieux: No. I'll rework it.

packages/rum/test/rum.spec.ts
```diff
@@ -520,3 +521,96 @@ describe('rum user action', () => {
     expect((getRumMessage(server, 0) as any).fooBar).toEqual('foo')
   })
 })
+
+describe('RUM hidden page', () => {
```
bcaudan, line 525:
```suggestion
describe('rum first_contentful_paint', () => {
```

packages/rum/test/rum.spec.ts
```diff
@@ -520,3 +521,96 @@ describe('rum user action', () => {
...
+    lifeCycle = new LifeCycle()
+  })
+
+  afterEach(() => {
+    browserWindow.PerformanceObserver = original
+    restorePageVisibility()
+    server.restore()
+  })
+
+  it('should not collect first_contentful_paint if page is not visible', () => {
```
bcaudan, line 561:
```suggestion
  it('should not be collected when page starts not visible', () => {
```

packages/rum/test/rum.spec.ts
```diff
@@ -520,3 +521,96 @@ describe('rum user action', () => {
...
+    expect(initialRequests.length).toBeGreaterThan(0)
+
+    initialRequests.map((request) => {
+      if (request.evt.category === 'view') {
+        expect(request.view.measures.first_contentful_paint).toBeUndefined()
+      }
+    })
+  })
+
+  it('should collect first_contentful_paint if page is visible', () => {
```
bcaudan, line 589:
```suggestion
  it('should be collected when page starts visible', () => {
```

packages/rum/test/rum.spec.ts
```diff
@@ -520,3 +521,96 @@ describe('rum user action', () => {
...
+  })
+
+  it('should not collect first_contentful_paint if page is not visible', () => {
+    const session = {
+      getId: () => undefined,
+      isTracked: () => true,
+      isTrackedWithResource: () => true,
+    }
+    const FAKE_RESOURCE: Partial<PerformanceEntry> = { name: 'http://foo.com', entryType: 'resource' }
+    const performanceEntry = { entryType: 'view', name: 'first-contentful-paint' }
```
bcaudan, line 568: it could be moved to the constant section at the beginning of the describe

packages/rum/test/rum.spec.ts
```diff
@@ -520,3 +521,96 @@ describe('rum user action', () => {
...
+    lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED, performanceEntry as PerformanceEntry)
+
+    const initialRequests = getServerRequestBodies<ExpectedRequestBody>(server)
+    expect(initialRequests.length).toBeGreaterThan(0)
+
+    initialRequests.map((request) => {
+      if (request.evt.category === 'view') {
+        expect(request.view.measures.first_contentful_paint).toBeUndefined()
+      }
+    })
```
bcaudan, line 586: this check is never performed since initialRequests contain only a `resource` request

packages/rum/test/rum.spec.ts
```diff
@@ -520,3 +521,96 @@ describe('rum user action', () => {
...
+    setPageVisibility('visible')
+    RUM = startRum('appId', lifeCycle, configuration as Configuration, session, internalMonitoring) as RumApi
+    startViewCollection(location, lifeCycle, session)
+    startPerformanceCollection(lifeCycle, session)
+    server.requests = []
+
+    stubBuilder.fakeEntry(FAKE_RESOURCE as PerformanceEntry, 'resource')
+    lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED, performanceEntry as PerformanceEntry)
+
+    const initialRequests = getServerRequestBodies<ExpectedRequestBody>(server)
```
BenoitZugmeyer, line 607: Could some of this setup be refactored with the other test case of the suite?

Also, this is a lot of setup, and maybe we could have tests a bit more unitary (I'm opening a subject here, this is more like something to think about for future development)

packages/rum/test/rum.spec.ts
BenoitZugmeyer: +1. Using something like `const viewEvents = initialRequests.filter(request => request.evt.category === 'view'); then using `viewEvents` for assertions would be better.

PR #363, opened 2020-04-16
👷[RUMF-449] switch cross browser tests to browserstack

bcaudan:
Iso functionality for now except: 
- unit tests in parallel
- link to browserstack UI from cli after starting a test

Future possible improvements:
- e2e in parallel
- rework build stages to allow unit and e2e in parallel


test/browsers.conf.js
```diff
@@ -0,0 +1,57 @@
...
+    browser_version: '11.0',
+    os: 'Windows',
+    os_version: '10',
+  },
+  IE_10: {
+    base: 'BrowserStack',
+    browser: 'IE',
+    browser_version: '10.0',
+    os: 'Windows',
+    os_version: '7',
```
glorieux, line 55: Windows 7 support has ended at the beginning of this year. 
Do we still have a reason to support it?

scripts/ci-bs.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+export BS_USERNAME=$(aws ssm get-parameter --region us-east-1 --name ci.browser-sdk.bs_username --with-decryption --query "Parameter.Value" --out text)
+export BS_ACCESS_KEY=$(aws ssm get-parameter --region us-east-1 --name ci.browser-sdk.bs_access_key --with-decryption --query "Parameter.Value" --out text)
+export IP_ADDRESS=$(ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1' | head -n 1)
```
BenoitZugmeyer, line 7: It this needed while running tests from a local machine? If it is, I'd be in favor of moving this in karma.bs.conf.js (maybe implement the grep part in JS)

test/browsers.conf.js
bcaudan: Since IE 10 support has been dropped at the end of january, it could worth challenging the support with product team.

test/getBuildInfos.js
```diff
@@ -0,0 +1,21 @@
...
+  if (CI_PIPELINE_ID && CI_JOB_ID && CI_COMMIT_SHORT_SHA) {
+    return `job: ${CI_JOB_ID} commit: ${CI_COMMIT_SHORT_SHA}`
+  }
+}
+
+function getLocalInfos() {
+  return `${process.env.USER} ${new Date().toLocaleString()}`
+}
+
+module.exports = function getBuildInfos() {
```
BenoitZugmeyer, line 12: You removed the LOCAL/CI and the unit/e2e tags. Is this intended?

test/e2e/scenario/agents.scenario.ts
```diff
@@ -100,7 +100,7 @@ describe('rum', () => {
 
     // Edge 18 seems to have valid timings even on cross origin requests ¯\_ツ_/¯ It doesn't matter
     // too much.
-    if (browser.capabilities.browserName === 'MicrosoftEdge' && browser.capabilities.browserVersion === '18') {
+    if (browser.capabilities.browserName === 'MicrosoftEdge') {
```
BenoitZugmeyer, line 103: I'd be in favor of having a condition taking the chromium-based Edge versions into account. Having a fixed version works: if the "browserVersion" isn't avaibable anymore, in any other capabilities property?
Else, does chromium-based Edge have a different "browserName"?

test/server/fake-backend.js
```diff
@@ -36,7 +36,7 @@ module.exports = (app) => {
     if (req.query['timing-allow-origin'] === 'true') {
       res.set('Timing-Allow-Origin', '*')
     }
-    res.send('ok')
+    setTimeout(() => res.send('ok'), 10)
```
BenoitZugmeyer, line 39: What was the issue here?

test/getBuildInfos.js
bcaudan: yes, with bs project/build concept it is pretty clear already IMO

test/e2e/scenario/agents.scenario.ts
bcaudan: I removed the check because version is handled differently by bs.
My take on that is that when we will upgrade edge version, this test will fail and we could remove this check instead of forgetting it.

test/server/fake-backend.js
bcaudan: tests too ⚡
duration was at 0 when we expected it to be greater than 0, so I introduced a small delay to ensure that the duration is really > 0

test/e2e/scenario/agents.scenario.ts
BenoitZugmeyer: IMO, we'll have to support both Chromium and non-Chromium Edge versions before remoing the non-Chromium Edge support, so we'll need to adjust the check anyway. But I'm fine to wait until then :)

test/server/fake-backend.js
BenoitZugmeyer: Ooh nice :)

test/e2e/wdio.bs.conf.js
```diff
@@ -1,7 +1,12 @@
+const execSync = require('child_process').execSync
+const path = require('path')
+
 const baseConf = require('./wdio.base.conf')
 const browsers = require('./browsers.conf')
 const getBuildInfos = require('../getBuildInfos')
 
+const ipAddress = execSync(path.join(__dirname, '../..', 'scripts/get-ip.sh'))
```
BenoitZugmeyer, line 8: What about:
```js
const ipAddress = Array.from(
  execSync("ifconfig", { encoding: "utf-8" })
    .matchAll(/inet (?:addr:)?((?:[0-9]*\.){3}[0-9]*)/g),
  match => match[1]
).find(ip => ip !== '127.0.0.1')
```

bcaudan: any advantage to use node instead of bash?

BenoitZugmeyer: Readability and portability. BTW, this may be even simpler:
```js
const ipAddress = Object.values(os.networkInterfaces())
  .flat()
  .find(({ family, internal }) => family === 'IPv4' && !internal)
  .address
```

Dockerfile
```diff
@@ -44,3 +44,6 @@ RUN mkdir -p /usr/share/man/man1
 RUN apt-get install -y -q --no-install-recommends default-jdk
 
 RUN apt-get -y install git
+
+RUN apt-get -y install net-tools
```
BenoitZugmeyer, line 48: Maybe remove this, since ifconfig isn't used anymore?

PR #366, opened 2020-04-17
Change view logic to emit LifeCycle events

BenoitZugmeyer:
Motivation: use lifecycle events to create new user actions.


packages/rum/src/rumSession.ts
```diff
@@ -15,19 +15,53 @@ export enum RumSessionType {
   TRACKED_WITHOUT_RESOURCES = '2',
 }
 
+class StoredSession {
```
glorieux, line 18: What's the purpose of the `StoredSession`? I don't get it 🤔

packages/rum/src/rumSession.ts
```diff
@@ -15,19 +15,53 @@ export enum RumSessionType {
...
   const session = startSessionManagement(RUM_SESSION_KEY, (rawType) => computeSessionState(configuration, rawType))
+  const storedSession = new StoredSession(() => session.getId() !== undefined)
+
+  storedSession.store(session.getId(), session.getType())
 
   session.renewObservable.subscribe(() => {
+    lifeCycle.notify(LifeCycleEventType.SESSION_WILL_RENEW)
+    storedSession.store(session.getId(), session.getType())
     lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
   })
```
bcaudan, line 62: Since the currentSessionId is already saved in session management and since we already have an object for session, why not only add something like:
```ts 
session.beforeRenewObservable.subscribe(() => {
  lifeCycle.notify(LifeCycleEventType.BEFORE_SESSION_RENEWED)
})
```

packages/core/src/transport.ts
```diff
@@ -160,12 +161,7 @@ export class Batch<T> {
        * register first to be sure to be called before flush on beforeunload
        * caveat: unload can still be canceled by another listener
        */
-      window.addEventListener(
-        DOM_EVENT.BEFORE_UNLOAD,
-        monitor(() => {
-          this.beforeFlushOnUnloadHandlers.forEach((handler) => handler())
```
bcaudan, line 166: Should we also remove `beforeFlushOnUnloadHandlers ` logic?

packages/core/src/transport.ts
```diff
@@ -160,12 +161,7 @@ export class Batch<T> {
...
-        })
-      )
+      window.addEventListener(DOM_EVENT.BEFORE_UNLOAD, monitor(this.willUnloadCallback))
```
bcaudan, line 164: IMO, it would be clearer and more consistent to use beforeXXX instead of willXXX

packages/rum/src/viewCollection.ts
```diff
@@ -16,156 +25,141 @@ export interface ViewMeasures {
...
-interface ViewContext {
-  id: string
-  location: Location
-}
+const THROTTLE_VIEW_UPDATE_PERIOD = 3000
 
-export let viewContext: ViewContext
+export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
+  let currentLocation = { ...location }
+  let endCurrentView = newView(lifeCycle, currentLocation, 0).end
```
bcaudan, line 32:
```suggestion
  const startOrigin = 0
  let endCurrentView = newView(lifeCycle, currentLocation, startOrigin).end
```

bcaudan, line 32: IMO, it would be clearer to use it this way:
```ts
  let currentView = newView(lifeCycle, currentLocation, 0)
  currentView.end()
```

packages/rum/src/viewCollection.ts
```diff
@@ -16,156 +25,141 @@ export interface ViewMeasures {
...
-  )
+  }
 }
 
-function trackHistory(
-  location: Location,
-  resetEventCounts: () => void,
-  upsertRumEvent: (event: RumEvent, key: string) => void
-) {
+function trackHistory(onChange: () => void) {
```
bcaudan, line 113:
```suggestion
function trackHistory(onHistoryChange: () => void) {
```

packages/rum/src/rumSession.ts
BenoitZugmeyer: This was my first thought, but I found that it's not that simple.

We can move the logic to sessionManagement, but because `session.getId()` is directly accessing the cookie value, we have no guarantee that the id right before a renew event is the one from the previously active session. It can already by the new session id or another session id. We'd still need to get rid of those direct cookie accesses, with a similar logic.

WDYT? Should I move this logic into sessionManagement?

packages/rum/src/rumSession.ts
BenoitZugmeyer: I agree this is non trivial and definitely should have write more doc on the subject. Currently, the most straightfoward way to get the session id is reading it directly from the cookie (there is a cache, but it doesn't help with the issue).

In RUM, we want to end the current view and start a new one when the session is renewed (= a new session is created when the previous session is expired). To end the current view, we need to send a final view event to the intake, to update the "duration" (and potentialy other metrics). Of course we need to send this event with the id of the ending session, but since the cookie may already be changed (ex: from another tab), we can't be sure we still have this id.

This is why we previously stored the current session id in the `viewContext` variable, to be sure to have a consistent id while the view is alive.

I wanted to move this storage elsewhere, because its weird that the `viewCollection` logic is in charge of the session id. This is where the `StoredSession` class takes place: an object to access the session id, updated only in a controled way.

packages/rum/src/rumSession.ts
bcaudan: FMU, there is two use cases:
- either we want the active session
- either we want the cached session
Could we expose from sessionManagement:
```ts
getCachedId() {
  return currentSessionId
}
```
And use that where we used viewContext.sessionId?

BenoitZugmeyer: This can be a good solution, yes. But we have to make sure that the session is still alive before using the cached id, else [this test](https://github.com/DataDog/browser-sdk/blob/master/test/e2e/scenario/agents.scenario.ts#L165-L171) fails.

Where would we want the active session outside of the `sessionManagement` module?

bcaudan: > This can be a good solution, yes. But we have to make sure that the session is still alive before using the cached id, else this test fails.

this rumSession check should still work, right?
```ts
isTracked: () => session.getId() !== undefined && isTracked(session.getType()),
```

> Also, where would we want the active session outside of the sessionManagement module?

in rumSession and loggerSession then.
Maybe using only cached session for rumSession.getId could work. renaming it to getCachedId could make this behavior more explicit

BenoitZugmeyer: > > This can be a good solution, yes. But we have to make sure that the session is still alive before using the cached id, else this test fails.
> 
> this rumSession check should still work, right?
> 
> ```ts
> isTracked: () => session.getId() !== undefined && isTracked(session.getType()),
> ```
IMO, the session type should be cached alongside the id, else we'll have incoherences.
> 
> > Also, where would we want the active session outside of the sessionManagement module?
> 
> in rumSession and loggerSession then.
> Maybe using only cached session for rumSession.getId could work. renaming it to getCachedId could make this behavior more explicit
* Like you say, in rumSession we could use only cached session informations.
* IMO, in loggerSession, nothing prevents us to use cached session informations as well. In fact, it should already use the session id from RUM since it is exposed in internal context
* Thus, having a single function to retrieve session information would make sense to me (whether this information is cached or not).

bcaudan: > IMO, the session type should be cached alongside the id, else we'll have incoherences.

session type is only computed once at startRum but we could move that to rumSession if it makes more sense

> Like you say, in rumSession we could use only cached session informations.

What I had in mind was return cached value for getId, use cookie value for isTracked/isTrackedWithResource

> IMO, in loggerSession, nothing prevents us to use cached session informations as well. In fact, it should already use the session id from RUM since it is exposed in internal context

I disagree, there is no reason to use cached value for logs since we don't keep local state like view for logs.
Since we need to compute session id when rum is not present, why should we change the current behavior?

> Thus, having a single function to retrieve session information would make sense to me (whether this information is cached or not).

not sure to see what you have in mind here

BenoitZugmeyer:
OK I reverted all session changes, and went back to the `viewContext.sessionId` for now. Things may be changed in a future PR.

packages/rum/src/lifeCycle.ts
```diff
@@ -48,11 +53,13 @@ export class LifeCycle {
...
   ): Subscription
   subscribe(eventType: LifeCycleEventType.USER_ACTION_COLLECTED, callback: (data: UserAction) => void): Subscription
+  subscribe(eventType: LifeCycleEventType.VIEW_COLLECTED, callback: (data: View) => void): Subscription
   subscribe(
     eventType:
       | LifeCycleEventType.SESSION_RENEWED
       | LifeCycleEventType.RESOURCE_ADDED_TO_BATCH
-      | LifeCycleEventType.DOM_MUTATED,
+      | LifeCycleEventType.DOM_MUTATED
+      | LifeCycleEventType.WILL_UNLOAD,
```
bcaudan, line 62: Could we rename it to `BEFORE_UNLOAD`?

packages/rum/src/viewCollection.ts
```diff
@@ -0,0 +1,171 @@
...
+  // Renew view on history changes
+  trackHistory(() => {
+    if (areDifferentViews(currentLocation, location)) {
+      currentLocation = { ...location }
+      currentView.end()
+      currentView = newView(lifeCycle, currentLocation, session)
+    }
+  })
+
+  // Renew view on session changes
```
bcaudan, line 45:
```suggestion
  // Renew view on session renewal
```

packages/rum/src/viewCollection.ts
```diff
@@ -0,0 +1,171 @@
...
+      startTime: startOrigin,
+    })
+  }
+
+  return {
+    end() {
+      stopTimingsTracking()
+      stopEventCountsTracking()
+      // Make a final view update
+      updateView()
```
bcaudan, line 114: is there a test that ensure that this last update is sent with the latest timings / event counts?

packages/rum/src/rumSession.ts
BenoitZugmeyer: This change have been reverted for now, let's consider this conversation resolved :p

PR #367, opened 2020-04-20
✅ improve browserstack e2e

bcaudan:
- fix flaky tests
- fix/improve server logging
- isolate tests
- run tests in parallel
- update wdio packages


test/e2e/wdio.base.conf.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   },
   e2eMode: process.env.E2E_MODE || 'bundle',
   onPrepare: function() {
+    exec('rm test/server/test-server.log')
```
BenoitZugmeyer, line 23: should be `execSync` (or better, `fs.unlinkSync`)

test/e2e/scenario/agents.scenario.ts
```diff
@@ -94,19 +94,15 @@ describe('rum', () => {
   })
 
   it('should not track disallowed cross origin xhr timings', async () => {
+    if (browser.capabilities.browserName === 'MicrosoftEdge') {
+      pending('Edge 18 seems to have valid timings even on cross origin requests')
```
BenoitZugmeyer, line 93: This is unfortunate. Anyway, this comment isn't really relevent. Maybe
```suggestion
      pending("Edge isn't always sending the timings")
```

test/e2e/scenario/helpers.ts
```diff
@@ -177,7 +177,10 @@ export function expectToHaveValidTimings(resourceEvent: ServerRumResourceEvent)
   expect(resourceEvent.date).toBeGreaterThan(0)
   expect(resourceEvent.duration).toBeGreaterThan(0)
   const performance = resourceEvent.http.performance!
```
BenoitZugmeyer, line 192:
```suggestion
  const performance = resourceEvent.http.performance
```

test/e2e/scenario/agents.scenario.ts
```diff
@@ -155,9 +155,13 @@ describe('rum', () => {
...
-    expect(viewEvents.length).toBe(2)
-    expect(viewEvents[0].session_id).not.toBe(viewEvents[1].session_id)
-    expect(viewEvents[0].view.id).not.toBe(viewEvents[1].view.id)
+    const firstViewEvent = viewEvents[0]
+    const lastViewEvent = viewEvents[viewEvents.length - 1]
+    expect(firstViewEvent.session_id).not.toBe(lastViewEvent.session_id)
+    expect(firstViewEvent.view.id).not.toBe(lastViewEvent.view.id)
+
+    const distinctIds = Array.from(new Set(viewEvents.map((viewEvent) => viewEvent.view.id)))
+    expect(distinctIds.length).toBe(2)
```
BenoitZugmeyer, line 159: why not using `set.size` ?

test/e2e/scenario/helpers.ts
```diff
@@ -10,6 +12,18 @@ export const serverUrl = {
 }
 
 const intakeRequest = request.defaults({ baseUrl: 'http://localhost:4000' })
+let specId: any
```
glorieux, line 15: Why not type it as `string`?

test/e2e/scenario/helpers.ts
```diff
@@ -1,5 +1,7 @@
 import { MonitoringMessage } from '@datadog/browser-core'
 import * as request from 'request'
+// @ts-ignore
```
BenoitZugmeyer, line 3: You should write a `currentSpecReporter.d.ts` instead of ignoring this.

test/e2e/wdio.local.conf.js
```diff
@@ -1,7 +1,7 @@
 const baseConf = require('./wdio.base.conf')
 
 // https://sites.google.com/a/chromium.org/chromedriver/downloads
-const CHROME_DRIVER_VERSION = '79.0.3945.36'
+const CHROME_DRIVER_VERSION = '81.0.4044.69'
```
glorieux, line 4: It seems that version is also defined on the Dockerfile. 
Could we define it once and then replace it in all the files?

.gitlab-ci.yml
```diff
@@ -7,8 +7,7 @@ variables:
 stages:
   - ci-image
   - test
-  - unit-bs
-  - e2e-bs
+  - browserstack
```
BenoitZugmeyer, line 10: Great!

test/e2e/wdio.local.conf.js
bcaudan: Driver and chrome version are not the same.
Not sure how we could mutualize that if it was.

glorieux: OK then 😄

test/e2e/currentSpecReporter.js
```diff
@@ -0,0 +1,16 @@
+let currentSpec = undefined
+
+class CurrentSpecReporter {
```
glorieux, line 3: Would it be possible to type that file to express the fact that it implements the `jasmine.CustomReporter`?

test/server/fake-backend.js
```diff
@@ -1,30 +1,31 @@
 module.exports = (app) => {
-  let logs = []
-  let rumEvents = []
-  let monitoring = []
+  let logs = {}
```
BenoitZugmeyer, line 2: Maybe you could have a middleware like this:
```js
const specContexts = new Map()

app.use((req, res, next) => {
  const specId = req.query['spec-id'] || 'unknown'
  if (!specContexts.has(specId)) {
    specContexts.set(specId, { id: specId, logs: [], rumEvents: [], monitoring: [] })
  }
  req.specContext = specContexts.get(specId)
})

// then

req.specContext.logs.push(JSON.parse(log))
```

test/e2e/currentSpecReporter.js
bcaudan: That would require to type the configuration which probably don't worth the trouble.
I'll go with https://github.com/DataDog/browser-sdk/pull/367#discussion_r411231844 to be cleaner anyway

PR #373, opened 2020-04-23
🐛 [RUM] fix view update after its end

BenoitZugmeyer:
With the last release (v1.11.0), we have seen a lot of "duplicated" views, meaning view events with the same id but a different date. This is likely caused by the throttled `updateView` function, sending a view update after the view ends when a timing or a event count have changed right before the view end.

This fixes the issue by ensuring that the view sending an update is still the current view.

### DIFF
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -82,7 +82,7 @@ function newView(
   viewContext = { id, location, sessionId: session.getId() }
 
   // Update the view every time the measures are changing
-  const scheduleViewUpdate = throttle(
+  const { throttled: scheduleViewUpdate } = throttle(
     monitor(() => {
       if (shouldIgnoreScheduledViewUpdate) {
         // The view update may come after the view has ended when the throttled scheduleViewUpdate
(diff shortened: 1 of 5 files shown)


packages/rum/src/viewCollection.ts
```diff
@@ -95,6 +95,12 @@ function newView(
   updateView()
 
   function updateView() {
+    if (viewContext.id !== id) {
+      // The view update may come after the view has ended when the throttled scheduleViewUpdate
+      // function is called right before ending the view.  In this case, we should not send the
+      // updated view event.
+      return
+    }
```
bcaudan, line 103: It could be clearer to introduce something like `cancelScheduledViewUpdate`
wdyt?

BenoitZugmeyer: A was going for it at first, then I thought this was safer, because it prevents *any* view update when another view is occuring. But indeed it shouldn't happen (we should not call startViewCollection twice for the same lifecycle), so I guess it's fine. Let's make things clearer instead!

bcaudan: instead of handling it with a flag, wdyt of having:
```ts
const {throttled: scheduleViewUpdate, cancel: cancelScheduledViewUpdate} = throttle(
```
it would more reflect the intent

BenoitZugmeyer:
LGTM! thanks. (PR authors can't approve their PR, so let's say it's ✅ )

PR #386, opened 2020-04-30
🐛 prevent runtime error in static test files

mquentin:
Fix runtime error occurring in `async-e2e-page.html` and `bundle-e2e-page.html` static test files

### DIFF
--- a/test/static/async-e2e-page.html
+++ b/test/static/async-e2e-page.html
@@ -12,7 +12,7 @@
     <script type="text/javascript">
       window.addEventListener('load', () => {
         const intakeOrigin = `http://${window.location.hostname}:4000`
-        const specIdParam = /spec-id=\d+/.exec(window.location.search)[0]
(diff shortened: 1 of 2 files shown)


test/static/async-e2e-page.html
```diff
@@ -12,7 +12,7 @@
     <script type="text/javascript">
       window.addEventListener('load', () => {
         const intakeOrigin = `http://${window.location.hostname}:4000`
-        const specIdParam = /spec-id=\d+/.exec(window.location.search)[0]
+        const specIdParam = window.location.search ? /spec-id=\d+/.exec(window.location.search)[0] : ''
```
bcaudan, line 15:
```suggestion
        const specIdParam = /spec-id=\d+/.exec(window.location.search) && /spec-id=\d+/.exec(window.location.search)[0]
```
it would avoid unexpected behavior if at some point there is something in search that is not spec-id

PR #388, opened 2020-05-04
✨[RUMF-373] Add View load duration and load type

mquentin:
### What it does
Implement a load duration for each collected View.
This duration starts at the View start time and end when there is no loading activity for a given time.
As the view load occurs before any possible user actions,  the collection of user action starts right after.

In addition, a view loading type is collected in view to identify `initial page load` and `route change` as any new view is loaded.

### Context
View load and User actions need to have separated lifecycle. The view loading mode needs to be identify.

### Change
- Prevent UA collection during the view load
- Collect the loading mode
- Test UA collection and View loading lifecycles
- Test view loading mode scenarios


packages/rum/test/userActionCollection.spec.ts
```diff
@@ -92,6 +103,54 @@ describe('startUserActionCollection', () => {
     button.click()
 
     clock.expire()
+  }
+
+  it('cancels user action on view loading', () => {
```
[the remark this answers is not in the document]
mquentin: I think that is not what have been decided. Until the end of the view load, no user action can be collected. I have to double check but that is my understanding.

packages/rum/src/viewCollection.ts
```diff
@@ -12,6 +12,13 @@ export interface View {
   documentVersion: number
   startTime: number
   duration: number
+  loadDuration?: number
+  loadType: ViewLoadType
+}
+
+export interface ViewLoad {
+  startTime: number
```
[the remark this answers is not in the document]
mquentin: no it is not used yet but I thought that a duration without a starting time is an incomplete information. But you right in the meantime I will remove it

packages/rum/src/viewCollection.ts
```diff
@@ -26,26 +33,33 @@ export interface ViewMeasures {
   userActionCount: number
 }
 
+export enum ViewLoadType {
+  INITIAL_PAGE_LOAD = 'initial page load',
+  ROUTE_CHANGE = 'route change',
```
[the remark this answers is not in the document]
mquentin: cf [RUMF-473](https://datadoghq.atlassian.net/browse/RUMF-473) (my ticket does specify these values so I guess its a product choice)

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
+interface ViewLoadingState {
+  pathname: string
+  startTime: number
+  stopPageActivitiesTracking: () => void
+}
+let currentViewLoadingState: ViewLoadingState | undefined
+
+function newViewLoading(lifeCycle: LifeCycle, pathname: string, startTime: number) {
+  // Cancel current user action
+  currentUserAction = undefined
```
bcaudan, line 113: could it be safer to also clear all waitUserActionComplete timeouts?

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
+  }
+
+  const { observable: pageActivitiesObservable, stop: stopPageActivitiesTracking } = trackPageActivities(lifeCycle)
+  currentViewLoadingState = {
+    pathname,
+    startTime,
+    stopPageActivitiesTracking,
+  }
+
+  waitUserActionCompletion(pageActivitiesObservable, (endTime) => {
```
bcaudan, line 126: since we are not using that for user action, we should rename this function and the used constants to something more appropriate.
like wait waitPageActivitiesEnded, PAGE_ACTIVITY_VALIDATION_DELAY, ...

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
+  stopPageActivitiesTracking: () => void
+}
+let currentViewLoadingState: ViewLoadingState | undefined
+
+function newViewLoading(lifeCycle: LifeCycle, pathname: string, startTime: number) {
+  // Cancel current user action
+  currentUserAction = undefined
+
+  if (currentViewLoadingState) {
+    currentViewLoadingState.stopPageActivitiesTracking()
```
bcaudan, line 116: could it be safer to also clear all waitUserActionComplete timeouts?

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
+  const { observable: pageActivitiesObservable, stop: stopPageActivitiesTracking } = trackPageActivities(lifeCycle)
+  currentViewLoadingState = {
+    pathname,
+    startTime,
+    stopPageActivitiesTracking,
+  }
+
+  waitUserActionCompletion(pageActivitiesObservable, (endTime) => {
+    stopPageActivitiesTracking()
+    if (currentViewLoadingState !== undefined) {
```
bcaudan, line 128: when can it be undefined?

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
+      // Validation timeout completion does not return an end time
+      const loadingEndTime = endTime || performance.now()
```
bcaudan, line 130: what does that mean? there is no test case on that

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
+  currentUserAction = { id, startTime, type, name }
 
   const { observable: pageActivitiesObservable, stop: stopPageActivitiesTracking } = trackPageActivities(lifeCycle)
   const { eventCounts, stop: stopEventCountsTracking } = trackEventCounts(lifeCycle)
 
   waitUserActionCompletion(pageActivitiesObservable, (endTime) => {
     stopPageActivitiesTracking()
     stopEventCountsTracking()
-    if (endTime !== undefined) {
+    if (endTime !== undefined && currentUserAction !== undefined) {
```
bcaudan: currentUserAction was supposed to be set to undefined before reaching this point by waitForUserActionCompletion.
Could it be done before this check and could we remove this undefined check?
If not, there is a test case missing

packages/rum/src/viewCollection.ts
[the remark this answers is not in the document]
bcaudan: yes, approved by product don't hesitate to comment on the brainstorm document if needed

packages/rum/src/viewCollection.ts
```diff
@@ -26,26 +33,33 @@ export interface ViewMeasures {
...
+}
+
 export const THROTTLE_VIEW_UPDATE_PERIOD = 3000
 
 export function startViewCollection(location: Location, lifeCycle: LifeCycle, session: RumSession) {
   let currentLocation = { ...location }
   const startOrigin = 0
-  let currentView = newView(lifeCycle, currentLocation, session, startOrigin)
+  let currentViewLoadType: ViewLoadType = ViewLoadType.INITIAL_PAGE_LOAD
```
bcaudan, line 45: why do we need to keep a reference to the currentViewLoadType?
Could we just call newView with the according type depending on the case?

packages/rum/src/viewCollection.ts
```diff
@@ -77,8 +93,9 @@ function newView(
     userActionCount: 0,
   }
   let documentVersion = 0
+  let loadDuration: number
 
-  viewContext = { id, location, sessionId: session.getId() }
+  viewContext = { id, location, loadType, sessionId: session.getId() }
```
bcaudan, line 97: why do we need that loadType in the view context?

packages/rum/src/viewCollection.ts
```diff
@@ -95,6 +112,12 @@ function newView(
   const { stop: stopTimingsTracking } = trackTimings(lifeCycle, updateMeasures)
   const { stop: stopEventCountsTracking } = trackEventCounts(lifeCycle, updateMeasures)
 
+  function updateLoadDuration(loadDurationValue: number) {
+    loadDuration = loadDurationValue
+    scheduleViewUpdate()
+  }
+  const { stop: stopLoadDurationTracking } = trackLoadDuration(lifeCycle, updateLoadDuration)
```
bcaudan, line 118: stopLoadDurationTracking is never called

packages/rum/test/userActionCollection.spec.ts
bcaudan, line 108: we should have two cases:
- cancel pending user action on new view
- dismiss new user action during view loading

packages/rum/test/viewCollection.spec.ts
```diff
@@ -32,11 +38,16 @@ describe('rum track url change', () => {
     initialLocation = viewContext.location
   })
 
-  it('should update view id on path change', () => {
+  it('should be a hard load before any path change', () => {
```
bcaudan, line 41:
```suggestion
  it('should collect initial view type as "initial page load"', () => {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -32,11 +38,16 @@ describe('rum track url change', () => {
...
+    expect(viewContext.loadType).toEqual(ViewLoadType.INITIAL_PAGE_LOAD)
+  })
+
+  it('should update view id and be a soft load on path change', () => {
```
bcaudan, line 45:
```suggestion
  it('should update view id and type on path change', () => {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -99,6 +110,32 @@ describe('rum track renew session', () => {
...
+      duration: loadingEndTime,
+      startTime: 0,
+    })
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getViewEvent(1).loadDuration).toEqual(loadingEndTime)
+
+    jasmine.clock().uninstall()
+  })
+})
```
bcaudan: you can probably merge this two cases,

```
rum load duration
   should be set when the load is complete
```

packages/rum/src/userActionCollection.ts
```diff
@@ -1,7 +1,8 @@
 import { Context, DOM_EVENT, generateUUID, monitor, Observable } from '@datadog/browser-core'
 import { getElementContent } from './getElementContent'
 import { LifeCycle, LifeCycleEventType, Subscription } from './lifeCycle'
-import { trackEventCounts } from './trackEventCounts'
+import { EventCounts, trackEventCounts } from './trackEventCounts'
```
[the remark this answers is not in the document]
mquentin: removed already

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
       return
     }
-
     newUserAction(lifeCycle, UserActionType.CLICK, getElementContent(event.target))
   }
 
-  addEventListener(DOM_EVENT.CLICK, processClick, { capture: true })
+  subscriptions.push(lifeCycle.subscribe(LifeCycleEventType.VIEW_COLLECTED, processViewLoading))
+  function processViewLoading(loadedView: View) {
+    if (!loadedView || loadedView.id === currentViewId) {
```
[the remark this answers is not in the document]
mquentin: nice point ! I will

packages/rum/src/userActionCollection.ts
```diff
@@ -66,55 +67,110 @@ export interface AutoUserAction {
...
   function processClick(event: Event) {
     if (!(event.target instanceof Element)) {
       return
     }
-
     newUserAction(lifeCycle, UserActionType.CLICK, getElementContent(event.target))
   }
 
-  addEventListener(DOM_EVENT.CLICK, processClick, { capture: true })
+  subscriptions.push(lifeCycle.subscribe(LifeCycleEventType.VIEW_COLLECTED, processViewLoading))
```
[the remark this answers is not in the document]
mquentin: I did that as we might subscribe to other event in the future if we add other UA types and it is cleaner to unsubscribe like this

packages/rum/src/viewCollection.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Thanks for the link :)

packages/rum/src/userActionCollection.ts
```diff
@@ -1,24 +1,50 @@
...
+//        |             |                  '------------'   |
+//        |             |                                   |
+//        '-------------'----------. ,----------------------'
+//                                  v
+//                         (View load complete)
+//                                  |
+//                                  |
+//                                  |
+//                                  v
+//                        (Start new user action)
```
BenoitZugmeyer, line 34: This schema can be moved to the "trackPageActivities" module (also, since it is the same for user action and view load duration, no need to put it twice)

packages/rum/src/trackPageActivities.ts
```diff
@@ -0,0 +1,111 @@
+import { monitor, Observable } from '@datadog/browser-core'
+import { LifeCycle, LifeCycleEventType, Subscription } from './lifeCycle'
+
+// Delay to wait for a page activity to validate the user action
+export const USER_ACTION_VALIDATION_DELAY = 100
```
BenoitZugmeyer, line 5: You should remove the "user action" warding in comments and constant names

packages/rum/src/trackPageActivities.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export function waitIdlePageActivity(
+  lifeCycle: LifeCycle,
+  completionCallback: (endTime: number | undefined) => void
+): { stop(): void } {
+  const { observable: pageActivitiesObservable, stop: stopPageActivitiesTracking } = trackPageActivities(lifeCycle)
+
+  waitPageActivitiesCompletion(pageActivitiesObservable, stopPageActivitiesTracking, (endTime) => {
+    completionCallback(endTime)
```
BenoitZugmeyer, line 22: I think we should not call `completionCallback` if `stop` has been called.

packages/rum/src/viewCollection.ts
```diff
@@ -175,3 +201,15 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
   )
   return { stop: stopPerformanceTracking }
 }
+
+function trackLoadDuration(lifeCycle: LifeCycle, id: string, callback: (loadDurationValue: number) => void) {
```
BenoitZugmeyer, line 205: This function should use `waitIdlePageActivity` to compute the load duration. You should remove the `VIEW_LOAD_COMPLETED` event, and any 'view load' related code in the user action collection.

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -303,11 +450,11 @@ describe('trackPagePageActivities', () => {
   })
 })
 
-describe('waitUserActionCompletion', () => {
+describe('waitPageActivitiesCompletion', () => {
```
BenoitZugmeyer, line 453: Could you move this test suite in a `trackPageActivities.spec.ts`?

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -92,6 +99,146 @@ describe('startUserActionCollection', () => {
...
+
+    clock.tick(SOME_ARBITRARY_DELAY)
+    button.click()
+    lifeCycle.notify(LifeCycleEventType.VIEW_COLLECTED, mockView as View)
+    clock.expire()
+
+    expect(events).toEqual([])
+  })
+
+  it('starts a user action when clicking on an element during a view loading', () => {
```
BenoitZugmeyer, line 125: I don't think this test is relevent anymore, since we don't do anything special related to user actions after a new view is created.

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -92,6 +99,146 @@ describe('startUserActionCollection', () => {
...
+          resourceCount: 0,
+        },
+        name: 'Click me',
+        startTime: jasmine.any(Number),
+        type: UserActionType.CLICK,
+      },
+    ])
+  })
+
+  it('starts a user action when clicking on an element after a view loading', () => {
```
BenoitZugmeyer, line 152: Same, I don't think this is relevent..

packages/rum/src/userActionCollection.ts
mquentin: In one case the [No page activity]  leads to discard the UA, in the other, it leads to the end of the view load so they are not exactly similar. But I get you point and moving it to the trackPageActivities makes sense

packages/rum/src/viewCollection.ts
mquentin: Then how do you know if you need to cancel the current UA while starting to load a new view ? Having the two process totally separated might hide the functional link between UA and page view. I am not super fan of that

BenoitZugmeyer: The collectUserAction module should expose a `cancelCurrentUserAction` function (as stated [here](https://github.com/DataDog/browser-sdk/pull/388#pullrequestreview-405046055)). AFAIK, this is the only link between the two processes.

packages/rum/src/userActionCollection.ts
mquentin: when there is no loading state ongoing. `currentViewLoadingState` is set to undefined

packages/rum/src/userActionCollection.ts
mquentin: That means that, if there is no activity at all during the `waitIdlePageActivity` and the `PAGE_ACTIVITY_VALIDATION_DELAY` occurs, the tracking process is not validated which leads to no endtime of the waiting process.
In the case of a UA, this scenario means that the UA is not valid
In the case of a Paga load, this scenario means that the loading has finished before the validation of the tracking process, as a result, the endtime of the process is now

packages/rum/src/viewCollection.ts
mquentin: this exported view context is used for the test and allows to expect the proper loadType for each scenarios

packages/rum/src/userActionCollection.ts
```diff
@@ -65,57 +32,71 @@ export interface AutoUserAction {
 
 export type UserAction = CustomUserAction | AutoUserAction
 
+export let currentUserActionProcess = { stop: noop }
+
 export function startUserActionCollection(lifeCycle: LifeCycle) {
+  const subscriptions: Subscription[] = []
```
BenoitZugmeyer, line 38: This is not needed anymore

packages/rum/src/viewCollection.ts
```diff
@@ -175,3 +198,35 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
   )
   return { stop: stopPerformanceTracking }
 }
+
+interface ViewLoadingState {
+  stop: () => void
+}
+let currentViewLoadingState: ViewLoadingState | undefined
```
BenoitZugmeyer, line 205: This shouldn't be needed, because:
* we always call `currentView.end()` (and thus stopping the `trackLoadDuration`) before starting a new view
* even if we wouldn't, it doesn't matter too much if two 'trackLoadDuration' calls are running simultaneously.

packages/rum/src/userActionCollection.ts
```diff
@@ -65,57 +32,71 @@ export interface AutoUserAction {
...
     stopEventCountsTracking()
-    if (endTime !== undefined) {
+    if (endTime !== undefined && currentUserAction !== undefined && currentUserAction.id === id) {
       lifeCycle.notify(LifeCycleEventType.USER_ACTION_COLLECTED, {
         id,
         name,
-        startTime,
         type,
-        duration: endTime - startTime,
+        duration: endTime - currentUserAction.startTime,
```
BenoitZugmeyer, line 82: Same comment as https://github.com/DataDog/browser-sdk/pull/388#discussion_r419483686 , I don't see why we use `currentDuration.startTime` instead of `startTime` directly.

packages/rum/src/userActionCollection.ts
```diff
@@ -65,57 +32,71 @@ export interface AutoUserAction {
 
 export type UserAction = CustomUserAction | AutoUserAction
 
+export let currentUserActionProcess = { stop: noop }
```
BenoitZugmeyer, line 35: Maybe we can add this `stop(): void` callback in `interface PendingAutoUserAction`, so we don't have too many global variables to manage?
To be clear, `newUserAction` would init the `currentUserAction` like this:
```ts
currentUserAction = {
  id,
  startTime,
  stop() {
    stopEventCountsTracking()
    currentUserAction = undefined
  }
}
```

And we could expose a function like that:

```ts
export function stopCurrentUserAction() {
    if (currentUserAction) currentUserAction.stop()
}
```

packages/rum/src/userActionCollection.ts
```diff
@@ -65,43 +32,51 @@ export interface AutoUserAction {
...
-
-    newUserAction(lifeCycle, UserActionType.CLICK, getElementContent(event.target))
+    stopNewUserAction = newUserAction(lifeCycle, UserActionType.CLICK, getElementContent(event.target))
   }
 
-  addEventListener(DOM_EVENT.CLICK, processClick, { capture: true })
-
   return {
     stop() {
+      stopNewUserAction.stop()
```
BenoitZugmeyer, line 53: `stopNewUserAction` is not necessarily defined (ex: if no click occured)

packages/rum/src/viewCollection.ts
```diff
@@ -175,3 +198,18 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
   )
   return { stop: stopPerformanceTracking }
 }
+
+export function trackLoadDuration(lifeCycle: LifeCycle, callback: (loadDurationValue: number) => void) {
+  // Stop the current user action
+  stopCurrentUserAction()
```
BenoitZugmeyer, line 204: NIT: This has little to do with `loadDuration`, I think it is safe to move in `newView` directly

packages/rum/src/userActionCollection.ts
mquentin: ok I initialized it too noop for this scenario

packages/rum/src/viewCollection.ts
mquentin: moved at the top of the `function newView(` which makes more sense

packages/rum/src/userActionCollection.ts
mquentin: I can remove the `currentUserAction !== undefined`from the updated version of the call but, as we are a new 'waitIdlePageActivity' module that replace `waitForUserActionCompletion` we implement that 'currentUserAction was supposed to be set to undefined before reaching this point by waitForUserActionCompletion' anymore

packages/rum/src/viewCollection.ts
mquentin: I guess I could use `currentView.loadType` to renew the view with the proper loading type instead of keeping this reference

packages/rum/src/userActionCollection.ts
```diff
@@ -65,42 +32,56 @@ export interface AutoUserAction {
...
 
-  waitUserActionCompletion(pageActivitiesObservable, (endTime) => {
-    stopPageActivitiesTracking()
+  function closeUserAction() {
     stopEventCountsTracking()
+    currentUserAction = undefined
+  }
+
+  const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (endTime) => {
     if (endTime !== undefined) {
```
bcaudan, line 85: either for page or user action, it is not explicit what has happened when `endTime === undefined`.
What about updating the callback signature to something like`(hadActivity, endTime) => void`?
it would allow us to do:
```
if (hadActivity) {
  // notify user action
}
```
```
const loadingEndTime = hadActivity ? endTime : performance.now()
```
which would be way more declarative and could probably spare the need to add the comment on trackLoadDuration

packages/rum/src/viewCollection.ts
bcaudan: What about using lifecycle events like the other tests to avoid leaking unnecessary data in this context?

mquentin: We already have a VIEW_COLLECTED event that notify the View updates but using this event would require to mock the view notification update process. The existing path change tests would not be straightforward anymore. Also creating a new event for these tests only would not be ideal as we would have duplicates.
For the understanding of the tests, keeping viewContext seems better.

packages/rum/src/viewCollection.ts
```diff
@@ -26,26 +30,31 @@ export interface ViewMeasures {
   userActionCount: number
 }
 
+export enum ViewLoadType {
+  INITIAL_LOAD = 'initial load',
+  ROUTE_CHANGE = 'route change',
```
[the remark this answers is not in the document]
mquentin: we can check with the product team but I like the load mention as it makes the initial load concept clearer that initial view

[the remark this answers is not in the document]
bcaudan: IMO, loadType: 'initial load' makes load a bit redundant

packages/rum/src/viewCollection.ts
bcaudan: I would be in favor of adapting the tests to the implementation rather than the other way around.
I am not in favor of adding direct dependency between viewCollection and userActionCollection when lifecycle purpose it to avoid that.

packages/rum/src/viewCollection.ts
bcaudan, line 35: are spaces well supported in the search? should not use _ to be more consistent with other types?

packages/rum/src/userActionCollection.ts
```diff
@@ -65,43 +32,56 @@ export interface AutoUserAction {
...
 
+interface PendingAutoUserAction {
+  id: string
+  startTime: number
+}
+let pendingAutoUserAction: PendingAutoUserAction | undefined
+
+let stopPendingAutoUserAction: { stop(): void } = {
+  stop: noop,
+}
```
bcaudan, line 43: what about merging these two concepts?
```ts
pendingAutoUserAction.stop()
```

packages/rum/src/viewCollection.ts
```diff
@@ -62,10 +70,11 @@ interface ViewContext {
 
 export let viewContext: ViewContext
 
-function newView(
+export function newView(
```
bcaudan, line 73: This function should only be used internally, you should not export it.
You should find another way to trigger new view in test based on conditions external to `startViewCollection`

packages/rum/test/viewCollection.spec.ts
```diff
@@ -99,6 +115,74 @@ describe('rum track renew session', () => {
...
+
+  beforeEach(() => {
+    let isFirstPerformanceNow = true
+    spyOn(performance, 'now').and.callFake(() => {
+      if (isFirstPerformanceNow) {
+        isFirstPerformanceNow = false
+        return mockDateNowT1
+      }
+      return mockDateNowT2
+    })
```
bcaudan, line 160: could it work with something like:
```ts
spyOn(performance, 'now').and.returnValues(mockDateNowT1, mockDateNowT2);
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -99,6 +115,74 @@ describe('rum track renew session', () => {
...
+    expect(getViewEvent(1).loadingTime).toEqual(mockDateNowT2 - mockDateNowT1)
+  })
+
+  it('should set a loadingTime once the load is complete after having some activity', () => {
+    jasmine.clock().tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+    expect(getViewEvent(1).loadingTime).toEqual(mockDateNowT2 - mockDateNowT1)
+  })
```
bcaudan, line 183: For me, it is confusing that both tests have the same expectation:
```
.toEqual(mockDateNowT2 - mockDateNowT1)
```
Could the difference between the two behaviors be expressed more clearly in the expectation?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -99,6 +115,74 @@ describe('rum track renew session', () => {
...
+
+  it('should set a loadingTime once the load is complete without having any activity', () => {
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_MAX_DURATION)
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getViewEvent(1).loadingTime).toEqual(mockDateNowT2 - mockDateNowT1)
+  })
+
+  it('should set a loadingTime once the load is complete after having some activity', () => {
```
bcaudan, line 177: IMO, `should set a loadingTime once ...` is not precise enought.
It should be `should set loadingTime to XXX once ...` to understand the difference of behavior between the two tests.

packages/rum/test/viewCollection.spec.ts
mquentin: Unfortunately the two scenarios uses the performance.now at different moments but they still need to be mocked as arbitrary values. Here we just want to make sure that both tests call performance.now twice and return the difference.

packages/rum/test/viewCollection.spec.ts
```diff
@@ -99,6 +115,65 @@ describe('rum track renew session', () => {
...
+
+  it('should have a loading time equal to the activity time if the load is complete with a unique activity', () => {
+    jasmine.clock().tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getViewEvent(1).loadingTime).toEqual(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+  })
+})
```
bcaudan, line 175: nice!

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -92,9 +95,35 @@ describe('startUserActionCollection', () => {
...
+
+    button.addEventListener(DOM_EVENT.CLICK, () => {
+      clock.tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+      // Since we don't collect dom mutations for this test, manually dispatch one
+      lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    })
+
+    clock.tick(SOME_ARBITRARY_DELAY)
+    button.click()
+    startViewCollection(fakeLocation as Location, lifeCycle, fakeSession as RumSession)
```
bcaudan, line 116: You are doing a unit test, FMU the main goal of unit test is to test the behavior of a component by looking at its outputs by giving it different inputs.
A test that verify the behavior of multiple components together is an integration test, it can be valuable but a downside is that its behavior is coupled to the behavior of all the tested components.

Here, you are testing `startUserActionCollection`.
By invoking `startViewCollection`, your test is now dependent of this component behavior too.
It seems unlikely to me that `startViewCollection` can be called after that `startUserActionCollection` has started to track some clicks.

IMO, we should stick to unit testing `startUserActionCollection` here.
If you really want to do an integration test, I would be curious why and you should have a more realistic setup.

packages/rum/src/userActionCollection.ts
```diff
@@ -115,113 +96,33 @@ function newUserAction(lifeCycle: LifeCycle, type: UserActionType, name: string)
...
   })
+
+  pendingAutoUserAction = {
+    id,
+    startTime,
+    stop() {
+      stopUserAction()
+      stopWaitIdlePageActivity()
+    },
+  }
```
bcaudan, line 109:
```suggestion
  pendingAutoUserAction = {
    id,
    startTime,
    stop() {
      pendingAutoUserAction = undefined
      stopEventCountsTracking()
      stopWaitIdlePageActivity()
    },
  }
```
To remove `stopUserAction` function, in order to have a single way to stop the pending user action, wdyt?

mquentin: I am not fan of than as it will rm the `stopUserAction` code factorisation that is used in the `waitIdlePageActivity` callback. Also this function is not exported and cannot be called outside of the `newUserAction` scope so there is still only one way to stop a pending action.

bcaudan: >  am not fan of than as it will rm the stopUserAction code factorisation that is used in the waitIdlePageActivity callback.

by calling `pendingAutoUserAction.stop()` in `waitIdlePageActivity`, there will be no duplication.

> Also this function is not exported and cannot be called outside of the newUserAction scope so there is still only one way to stop a pending action.

yes but inside `userActionCollection` there is two ways which is maybe not necessary

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -17,7 +10,7 @@ import {
   UserActionType,
 } from '../src/userActionCollection'
 const { resetUserAction, newUserAction } = $$tests
-import { startViewCollection, ViewLoadingType } from '../src/viewCollection'
+import { View, ViewLoadingType } from '../src/viewCollection'
```
bcaudan, line 13: 1. View is not used
1. this import should be above the const

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -92,9 +88,48 @@ describe('startUserActionCollection', () => {
...
     clock.expire()
+  }
+
+  it('cancels pending user action on view loading', () => {
+    const fakeLocation: Partial<Location> = { pathname: '/foo' }
+    const fakeSession = {
+      getId() {
+        return '42'
+      },
+    }
```
bcaudan, line 99: unused

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -92,9 +88,48 @@ describe('startUserActionCollection', () => {
...
+      loadingType: ViewLoadingType.INITIAL_LOAD,
+      location: fakeLocation as Location,
+      measures: {
+        errorCount: 0,
+        longTaskCount: 0,
+        resourceCount: 0,
+        userActionCount: 0,
+      },
+      startTime: 0,
+    })
```
bcaudan, line 122: Since we don't care about the content of the view in this test, what about using a stub:
```ts
const fakeView = {}
lifeCycle.notify(LifeCycleEventType.VIEW_COLLECTED, fakeView as View)
```
It would remove unnecessary noise and avoid that this test break when there is a change in the view event structure

packages/rum/src/userActionCollection.ts
mquentin: to define `pendingAutoUserAction.stop()` you need `stopWaitIdlePageActivity`. Calling `pendingAutoUserAction.stop()` and therefore `stopWaitIdlePageActivity` inside the callback of the same `waitIdlePageActivity` is redundant and makes the stop process way more ambiguous.

Also `pendingAutoUserAction` can be set to undefined while waiting for the  `waitIdlePageActivity` callback which, once more, would make the `pendingAutoUserAction.stop()` process ambiguous as we would need to call `pendingAutoUserAction?.stop()`

bcaudan: > to define pendingAutoUserAction.stop() you need stopWaitIdlePageActivity. Calling pendingAutoUserAction.stop() and therefore stopWaitIdlePageActivity inside the callback of the same waitIdlePageActivity is redundant and makes the stop process way more ambiguous.

Since it is already done for inside `trackPageActivities`, `stopPageActivitiesTracking` is called multiple times in `waitIdlePageActivity`, is it really an issue?

> Also pendingAutoUserAction can be set to undefined while waiting for the waitIdlePageActivity callback which, once more, would make the pendingAutoUserAction.stop() process ambiguous as we would need to call pendingAutoUserAction?.stop()

there is already several occurrences of:
```ts
if (pendingAutoUserAction) {
    pendingAutoUserAction.stop()
}
```
This should be the preferred approach instead of `pendingAutoUserAction?.stop()`
How would it be ambiguous then? since it is already how it is used elsewhere

mquentin: >Since it is already done inside trackPageActivities, stopPageActivitiesTracking is called multiple times in waitIdlePageActivity, is it really an issue?

`stopPageActivitiesTracking` is not called in the direct callback of `trackPageActivities`.
Here we are talking of calling `stopWaitIdlePageActivity` inside the callback of its own `waitIdlePageActivity` when the `waitIdlePageActivity` process has already managed its own lifecycle.

I am not fan of this:
waitIdlePageActivity -> stop myself -> send the callback -> stop myself a second time

bcaudan: ok fair enough

> I am not fan of this:
waitIdlePageActivity -> stop myself -> send the callback -> stop myself a second time

What issue do you see with this behavior?
IMO, we could either:
- allow this behavior 
- find a better naming for `stopUserAction`, stopping the user action should be `pendingAutoUserAction.stop` and `waitIdlePageActivity.stop` is a part of this process, so `stopUserAction` name should be more specific

PR #392, opened 2020-05-15
✨[RUMF-465] collect client service, env and version

mquentin:
## Motivation
Allows users to configure optional fields that tag their RUM events and logs with a specific service, environment and version.

## Changes
New optional configuration options have been added to the log, rum init process. If define, these new fields will then create new tags for each of the logs and RUM events send to the intake.

## Testing
1. add the optional `clientService, clientEnv, clientVersion` fields to the `window.DD_RUM.init and window.DD_LOGS.init`
2. check in the network tab that the intake URL carries the new tags ex: `&ddtags=sdk_version:dev,env:envaaa,service:serviceaaa,version:versionaaa`

### DIFF
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -141,10 +141,9 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
 function getEndpoint(type: string, conf: TransportConfiguration, source?: string) {
   const tld = conf.datacenter === 'us' ? 'com' : 'eu'
   const domain = conf.sdkEnv === 'production' ? `datadoghq.${tld}` : `datad0g.${tld}`
-  const tags = `sdk_version:${conf.sdkVersion}
-    ${conf.clientEnv ? `,env:${conf.clientEnv}` : ''}
-    ${conf.clientService ? `,service:${conf.clientService}` : ''}
-    ${conf.clientVersion ? `,version:${conf.clientVersion}` : ''}`
+  const tags = `sdk_version:${conf.sdkVersion}${conf.clientEnv ? `,env:${conf.clientEnv}` : ''}${
+    conf.clientService ? `,service:${conf.clientService}` : ''
+  }${conf.clientVersion ? `,version:${conf.clientVersion}` : ''}`
   const datadogHost = `${type}-http-intake.logs.${domain}`
   const host = conf.proxyHost ? conf.proxyHost : datadogHost
   const proxyParameter = conf.proxyHost ? `ddhost=${datadogHost}&` : ''
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -49,6 +49,10 @@ export interface UserConfiguration {
   internalMonitoringEndpoint?: string
   logsEndpoint?: string
   rumEndpoint?: string
+
+  clientService?: string
+  clientEnv?: string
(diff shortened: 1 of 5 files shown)


packages/core/src/configuration.ts
```diff
@@ -49,6 +49,10 @@ export interface UserConfiguration {
   internalMonitoringEndpoint?: string
   logsEndpoint?: string
   rumEndpoint?: string
+
+  clientService?: string
+  clientEnv?: string
+  clientVersion?: string
```
bcaudan, line 55: these fields should be above this line:
```ts
// Below is only taken into account for e2e-test build mode.
```

bcaudan, line 55: we should probably remove the `client` prefix for the API.
For example, `service` will be used as `service` and not `clientService`.
It will also be consistent with current tracer API: https://datadoghq.dev/dd-trace-js/

packages/core/src/configuration.ts
```diff
@@ -63,19 +67,26 @@ export type Configuration = typeof DEFAULT_CONFIGURATION & {
...
   const transportConfiguration: TransportConfiguration = {
     buildMode: buildEnv.buildMode,
+    clientEnv: userConfiguration.clientEnv,
+    clientService: userConfiguration.clientService,
     clientToken: userConfiguration.clientToken,
+    clientVersion: userConfiguration.clientVersion,
     datacenter: userConfiguration.datacenter || buildEnv.datacenter,
-    env: buildEnv.env,
     proxyHost: userConfiguration.proxyHost,
+    sdkEnv: buildEnv.sdkEnv,
```
bcaudan, line 89: 👍

packages/core/src/configuration.ts
```diff
@@ -129,8 +140,10 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
 
 function getEndpoint(type: string, conf: TransportConfiguration, source?: string) {
   const tld = conf.datacenter === 'us' ? 'com' : 'eu'
-  const domain = conf.env === 'production' ? `datadoghq.${tld}` : `datad0g.${tld}`
-  const tags = `sdk_version:${conf.sdkVersion}`
+  const domain = conf.sdkEnv === 'production' ? `datadoghq.${tld}` : `datad0g.${tld}`
+  const tags = `sdk_version:${conf.sdkVersion}${conf.clientEnv ? `,env:${conf.clientEnv}` : ''}${
+    conf.clientService ? `,service:${conf.clientService}` : ''
+  }${conf.clientVersion ? `,version:${conf.clientVersion}` : ''}`
```
bcaudan, line 146: What about making that a bit more readable/maintanable?

packages/core/test/configuration.spec.ts
```diff
@@ -80,4 +80,16 @@ describe('configuration', () => {
       expect(configuration.rumEndpoint).toContain('?ddhost=rum-http-intake.logs.datadoghq.com&')
     })
   })
+
+  describe('env, version and service', () => {
+    it('should set the env, version and services tags in the logs and rum endpoints', () => {
```
bcaudan, line 85:
```suggestion
    it('should be set as tags in the logs and rum endpoints', () => {
```

bcaudan, line 85: Could we add a test when they are not set?

PR #394, opened 2020-05-19
✨[RUMF-513] add a session keep alive mechanism

bcaudan:
## Motivation

Ensure that the backend consider active session as active

## Changes

- Stop sending a last view update when renewing the session
- Send a page view update every 5 minutes

## Testing

unit tests

---


bcaudan:
> Good to me. Just the we tend to have
> 
> `const { stop: stopViewCollection } = startViewCollection(location, lifeCycle, trackedWithResourcesSession)`
> 
> instead of
> 
> `viewCollection = startViewCollection(location, lifeCycle, trackedWithResourcesSession)`
> `let viewCollection: { stop(): void }`
> 
> elsewhere in the sdk

@mquentin I'm on the fence on that, keeping a reference to the object instead of reference to the function would allow us to not create multiple references if the object contains multiple functions, like `currentView` in this PR.
I am not sure that we need to enforce one way or the other for now.
Open to arguments though.

mquentin:
What about, using the unique function reference and then, if needed merge all the function / props into a unique object ref, see the other PR :
```
interface PendingAutoUserAction {
  id: string
  startTime: number
  stop(): void
}
```

packages/rum/test/rum.spec.ts
```diff
@@ -261,6 +262,7 @@ describe('rum session', () => {
   })
 
   afterEach(() => {
+    viewCollection.stop()
```
BenoitZugmeyer, line 265: viewCollection may not be defined here, because all test cases aren't defining it. Maybe test if it is defined before using it?

packages/rum/test/rum.spec.ts
```diff
@@ -519,11 +524,12 @@ describe('rum global context', () => {
     server = sinon.fakeServer.create()
     lifeCycle = new LifeCycle()
     RUM = startRum('appId', lifeCycle, configuration as Configuration, session, internalMonitoring) as RumApi
-    startViewCollection(location, lifeCycle, session)
+    viewCollection = startViewCollection(location, lifeCycle, session)
     server.requests = []
   })
 
   afterEach(() => {
+    viewCollection.stop()
```
BenoitZugmeyer, line 532: Same here, viewCollection may not be defined

packages/rum/src/viewCollection.ts
```diff
@@ -55,6 +56,20 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle, se
     currentView.triggerUpdate()
     currentView.end()
   })
+
+  // Session keep alive
+  const keepAliveInterval = setInterval(
```
BenoitZugmeyer, line 61: Unless there is a good reason I'm missing, I'd rather put this logic in the `newView` function where all other `updateView` are done.

packages/rum/src/viewCollection.ts
```diff
@@ -55,6 +56,20 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle, se
...
+    monitor(() => {
+      currentView.triggerUpdate()
+    }),
+    SESSION_KEEP_ALIVE_INTERVAL
+  )
+
+  return {
+    stop() {
+      clearInterval(keepAliveInterval)
```
BenoitZugmeyer, line 70: Maybe we should end the current view here

packages/rum/src/viewCollection.ts
```diff
@@ -37,19 +37,22 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle, se
...
       currentLocation = { ...location }
+      currentView.triggerUpdate()
       currentView.end()
       currentView = newView(lifeCycle, currentLocation, session)
     }
   })
 
   // Renew view on session renewal
   lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
+    // do not trigger view update to avoid wrong data
```
BenoitZugmeyer, line 49: Instead of introducing a `triggerUpdate` function, I would rather introduce a `endWithoutUpdate` (or add an option `end({ update: false })` or something like that). This makes things clearer : `startViewCollection` is only responsible of creating and ending views, and `newView` is responsible to update the view during its lifecycle.

packages/rum/test/rum.spec.ts
bcaudan: it is defined in the beforeEach, when can it not be defined?

packages/rum/test/rum.spec.ts
bcaudan: Currently, all tests are defining it but you are right, it could not always be the case.
I would be in favor of trying to move the startViewCollection in the beforeEach if it is possible, it could factorize some test setup logic.
wdyt?

packages/rum/src/viewCollection.ts
bcaudan: Since it is global to the viewCollection mechanism and not specific to specific views, what would be the pros/cons for that?

packages/rum/src/viewCollection.ts
bcaudan: cf first commit

packages/rum/src/viewCollection.ts
bcaudan: both could work, I find it more clear to separate the end from the update instead of having two ways to end a view. For me the two concerns are separated since there is other conditions where we want to trigger an update, cf keep alive.

bcaudan:
> What about, using the unique function reference and then, if you need to create other functions or fields merge them all into a unique object ref, like we did in the other PR :
> 
> ```
> interface PendingAutoUserAction {
>   id: string
>   startTime: number
>   stop(): void
> }
> ```
> 
> Advantages:
> Case 1 - a unique stop function is required - `const { stop: stopViewCollection } = startViewCollection(...)`
> 
> * typing is straightforward,
>   - no need to create an object `viewCollection` interface
>   - no need to question the naming of this `viewCollection` object
> * Clarify the return value of `startViewCollection` as a single stop function
> * code reading is easier
>   - no need to set objectA at one point and then call objectA.stop at another
>   - no questioning about the potential other fields of this objectA and where are they set
> 
> Case 2 - several fields and functions need to be referenced - `let viewCollection: { stop(): void, field1: number... }`
> 
> * typing describe the non trivial structure of the object and help understanding its usage
> * merge all the reference into one object

OK, I am not against it, @BenoitZugmeyer, @glorieux any thoughts?

packages/rum/src/viewCollection.ts
BenoitZugmeyer: We are sending RUM events to keep the session alive. IMO this is not related to the viewCollection mechanism either, but since we chose to send RUM view events for this, we are free to chose where we put them.

By updating the view in the `newView` function:
* all updates are done at the same place: this makes things self contained simpler to follow
* we could imagine a more elaborated logic in the future, to update the view only if hasn't been updated recently (a bit like a debounce)

packages/rum/src/viewCollection.ts
BenoitZugmeyer: Mmh the first commit is related to not updating the view on session renew.. I don't see how it's related? We could end the view without updating it.

BenoitZugmeyer:
If I understand correctly, you're in favor of using global variables instead of using the returned value of the function? I would not advise it :) I know we are doing this kind of stuff in user actions, and I think this is legitimate because we we need to be able to stop the user action from an external module, but it should not be the norm and should be limited as much as possible.

The current solution is fine by me.

I don't want to reopen the topic here, but I still think using real Observables would make those patterns more straightforward.

packages/rum/src/viewCollection.ts
bcaudan: > all updates are done at the same place: this makes things self contained and simpler to follow

I think it depends on what we choose for current `triggerUpdate` method

> we could imagine a more elaborated logic in the future, to update the view only if hasn't been updated recently (a bit like a debounce)

let's see when we get there

One potential drawback of having the keep alive strategy inside the `newView` function is that it would then technically make more sense to start/stop a specific interval for each view which is not really needed for our current problem.

packages/rum/src/viewCollection.ts
bcaudan: my bad, I was thinking about `view.end`

the view collection stop is only called in unit test, what would be the goal of stopping it here? how would it work for session renewed, path change?

bcaudan:
> If I understand correctly, you're in favor of using global variables instead of using the returned value of the function? I would not advise it :) I know we are doing this kind of stuff in user actions, and I think this is legitimate because we we need to be able to stop the user action from an external module, but it should not be the norm and should be limited as much as possible.

@BenoitZugmeyer not sure to follow you on the global variables part. It is more using the returned object instead of deconstructing it and use its functions. What are you referring to in user actions?

> I don't want to reopen the topic here, but I still think using real Observables would make those patterns more straightforward.

If you have a clear vision on that, don't hesitate to put some thoughts on a doc or a task.

BenoitZugmeyer:
All right, so I didn't understand correctly :) I was refering to the global PendingAutoUserAction in the user action collection mechanism, holding a `stop()` function.

So the debate is between (I use tests as an example, but it could be applied in other situation):
```js
let viewCollection
beforeEach(() => {
  viewCollection = startViewCollection()
})
afterEach(() => {
  viewCollection.stop()
})
```
and
```js
let stopViewCollection
beforeEach(() => {
  ;({ stop: stopViewCollection } = startViewCollection())
})
afterEach(() => {
  stopViewCollection()
})
```
?

I don' thave a strong opinion on this, but I find the former case more readable.

> If you have a clear vision on that, don't hesitate to put some thoughts on a doc or a task.

I'll think about it

packages/rum/src/viewCollection.ts
BenoitZugmeyer: You are right, this would be only for tests. Also yes, we should probably stop listening for session renewed events and path change.

Unsubscribing from all listeners and ending the current view would have the same goal as clearing the keepalive interval: make sure no event is sent after stoping view collection. This is optional (even stoping the keepalive interval is not needed), because we never use the same Lifecycle object for two tests. But it seems weird to me to stop one thing and not the other.

packages/rum/src/viewCollection.ts
BenoitZugmeyer: Ok

packages/rum/src/viewCollection.ts
bcaudan: oh, did you mean "Maybe we should **also** end the current view here"?

BenoitZugmeyer: Yes :)

PR #396, opened 2020-05-20
👷 Removing lodash dependencies

thibthib:
## Motivation

The SDK code being quite large – as you can see on [bundlephobia](https://bundlephobia.com/result?p=@datadog/browser-rum@1.11.4)  – I wanted to remove lodash as it's quite big and can easily be replaced with something smaller.

## Changes

Two dependencies got replaced:
- `lodash.merge` got replaced with a function that does the same thing, just without the ability to merge objects with circular references
- `lodash.assign` got replaced with simple loop re-assigning keys

## Testing

Everything should work _exactly the same_, but happy to write more tests if needed.

---

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -73,6 +73,54 @@ export function throttle(
 }
 
 /**
+ * Performs a deep merge of objects and arrays
+ * - arrays values are merged index by index
+ * - objects are merged by keys
+ * - values get replaced, unless undefined
+ *
+ * ⚠️ this method does not prevent infinite loops while merging circular references ⚠️
+ *
+ * @param {destination, ...toMerge} any - values to merge
+ * @returns {any} New merged value
+ */
+const isContextArray = (value: ContextValue): value is ContextArray => Array.isArray(value)
+const isContext = (value: ContextValue): value is Context => !Array.isArray(value) && typeof value === 'object'
+
+export function deepMerge(destination: ContextValue, ...toMerge: ContextValue[]): ContextValue {
+  return toMerge.reduce((value1: ContextValue, value2: ContextValue): ContextValue => {
+    if (isContextArray(value1) && isContextArray(value2)) {
+      return [...Array(Math.max(value1.length, value2.length))].map((_, index) =>
+        deepMerge(value1[index], value2[index])
+      )
+    }
+    if (isContext(value1) && isContext(value2)) {
+      return Object.keys(value2).reduce(
+        (merged, key) => ({
+          ...merged,
+          [key]: deepMerge(value1[key], value2[key]),
+        }),
+        value1
+      )
+    }
+    return value2 === undefined ? value1 : value2
+  }, destination)
+}
+
+interface Assignable {
+  [key: string]: any
+}
+
+export function assign(target: Assignable, ...toAssign: Assignable[]) {
+  toAssign.forEach((source: Assignable) => {
+    for (const key in source) {
+      if (Object.prototype.hasOwnProperty.call(source, key)) {
+        target[key] = source[key]
+      }
+    }
+  })
+}
+
+/**
  * UUID v4
  * from https://gist.github.com/jed/982883
  */
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -72,6 +72,9 @@ export function throttle(
   }
 }
 
+const isContextArray = (value: ContextValue): value is ContextArray => Array.isArray(value)
+const isContext = (value: ContextValue): value is Context => !Array.isArray(value) && typeof value === 'object'
+
 /**
  * Performs a deep merge of objects and arrays
  * - arrays values are merged index by index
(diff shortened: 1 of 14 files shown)


packages/logs/src/logger.ts
```diff
@@ -82,7 +82,7 @@ export function startLogger(
           },
         },
         globalContext,
-        getRUMInternalContext()
+        getRUMInternalContext() as ContextValue
```
bcaudan, line 85:
```suggestion
        getRUMInternalContext() as Context
```

packages/rum/src/rum.entry.ts
```diff
@@ -57,7 +57,7 @@ const STUBBED_RUM = {
 
 export type RumGlobal = typeof STUBBED_RUM
 
-export const datadogRum = makeGlobal(STUBBED_RUM)
+export let datadogRum = makeGlobal(STUBBED_RUM)
```
bcaudan, line 60:
```suggestion
export const datadogRum = makeGlobal(STUBBED_RUM)
```
it does not seem necessary

packages/core/src/utils.ts
```diff
@@ -72,6 +72,54 @@ export function throttle(
...
+ *
+ * ⚠️ this method does not prevent infinite loops while merging circular references ⚠️
+ *
+ * @param {destination, ...toMerge} any - values to merge
+ * @returns {any} New merged value
+ */
+const isContextArray = (value: ContextValue): value is ContextArray => Array.isArray(value)
+const isContext = (value: ContextValue): value is Context => !Array.isArray(value) && typeof value === 'object'
+
+export function deepMerge(destination: ContextValue, ...toMerge: ContextValue[]): ContextValue {
```
bcaudan, line 89: Since we were relying on lodash, I am not sure that we have test cases that ensure the behavior of this function.
It would be nice to have a set of test cases to validate it and ease future maintenance.

bcaudan, line 89: Could you colocate the comment with the related function?

packages/core/src/utils.ts
```diff
@@ -72,6 +72,54 @@ export function throttle(
...
+/**
+ * Performs a deep merge of objects and arrays
+ * - arrays values are merged index by index
+ * - objects are merged by keys
+ * - values get replaced, unless undefined
+ *
+ * ⚠️ this method does not prevent infinite loops while merging circular references ⚠️
+ *
+ * @param {destination, ...toMerge} any - values to merge
+ * @returns {any} New merged value
```
bcaudan, line 84: I would be in favor of not document `params` and `returns` since naming is pretty explicit and we would need to maintain types here too (which seems already outdated).

packages/rum/src/rum.entry.ts
thibthib: Indeed, I changed the assign implementation and forgot to revert this change !

packages/core/src/utils.ts
thibthib: I'll add unit tests indeed

packages/core/src/utils.ts
```diff
@@ -72,6 +72,52 @@ export function throttle(
...
+ */
+export function deepMerge(destination: ContextValue, ...toMerge: ContextValue[]): ContextValue {
+  return toMerge.reduce((value1: ContextValue, value2: ContextValue): ContextValue => {
+    if (isContextArray(value1) && isContextArray(value2)) {
+      return [...Array(Math.max(value1.length, value2.length))].map((_, index) =>
+        deepMerge(value1[index], value2[index])
+      )
+    }
+    if (isContext(value1) && isContext(value2)) {
+      return Object.keys(value2).reduce(
```
BenoitZugmeyer, line 95: A for/in loop or a (polyfilled) usage of `Object.fromEntries` would be easier to understand and faster here.

yarn.lock
```diff
@@ -6122,7 +6103,7 @@ lodash._root@^3.0.0:
   resolved "https://registry.yarnpkg.com/lodash._root/-/lodash._root-3.0.1.tgz#fba1c4524c19ee9a5f8136b4609f017cf4ded692"
   integrity sha1-+6HEUkwZ7ppfgTa0YJ8BfPTe1pI=
 
-lodash.assign@4.2.0, lodash.assign@^4.2.0:
+lodash.assign@^4.2.0:
```
glorieux, line 6106: I'm wondering why those are still here?

thibthib: `yarn why lodash.assign` tells us that it's a dependency of `_project_#webdriverio#archiver#archiver-utils`

packages/core/src/utils.ts
thibthib: It's not really clear what's polyfilled and what is not 😅 And I'm not sure `ObjectEntries` would be faster as it'll be doing an extra `map` on the array

BenoitZugmeyer: It's not polyfilled. It would have a lower complexity, since you recreate the whole object for each key. Anyway, a for/in could do the trick :)

yarn.lock
glorieux: OK so just for tests. 
Sounds good to me.

PR #397, opened 2020-05-22
✅ [RUM] introduce test setup builder

bcaudan:
## Motivation

Ease bootstrapping of RUM unit tests by:

- avoid to wire test subjects with all their dependencies
- abstract mock setup (fake server, clock, ...) 
- mutualize cleanup

```ts
beforeEach(() => {
  setupBuilder = setup()
    .withFakeServer()
    .withRum()
    .withViewCollection()
})

afterEach(() => {
  setupBuilder.cleanup()
})

it('should ...', () => {
  const { server, lifeCycle } = setupBuilder.build()
  server.requests = []

  lifeCycle.notify(...)

  ...
})


```

## Changes

- remove a useless test
- introduce test setup builder
- use it in existing tests when relevant

---

### DIFF
--- a/packages/rum/test/rum.spec.ts
+++ b/packages/rum/test/rum.spec.ts
@@ -2,25 +2,16 @@ import {
   Configuration,
   DEFAULT_CONFIGURATION,
   ErrorMessage,
-  InternalMonitoring,
   isIE,
-  Omit,
-  PerformanceObserverStubBuilder,
   RequestCompleteEvent,
   SPEC_ENDPOINTS,
 } from '@datadog/browser-core'
 import sinon from 'sinon'
 
 import { LifeCycle, LifeCycleEventType } from '../src/lifeCycle'
-import { startPerformanceCollection } from '../src/performanceCollection'
-import { handleResourceEntry, RumEvent, RumResourceEvent, startRum } from '../src/rum'
-import { RumGlobal } from '../src/rum.entry'
+import { handleResourceEntry, RumEvent, RumResourceEvent } from '../src/rum'
 import { UserAction, UserActionType } from '../src/userActionCollection'
-import { startViewCollection } from '../src/viewCollection'
-
-interface BrowserWindow extends Window {
-  PerformanceObserver?: PerformanceObserver
-}
+import { setup, TestSetupBuilder } from './specHelper'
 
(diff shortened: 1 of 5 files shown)


packages/rum/test/viewCollection.spec.ts
```diff
@@ -116,23 +119,32 @@ describe('rum track renew session', () => {
...
+  let addRumEvent: jasmine.Spy
   let getViewEvent: (index: number) => View
 
   beforeEach(() => {
-    jasmine.clock().install()
-    ;({ lifeCycle, getViewEvent } = spyOnViews())
+    ;({ addRumEvent, getViewEvent } = spyOnViews())
+
+    const fakeLocation: Partial<Location> = { pathname: '/foo' }
+    mockHistory(fakeLocation)
```
BenoitZugmeyer, line 130: Maybe `mockHistory` could be part of the setup builder?

packages/rum/test/rum.spec.ts
```diff
@@ -452,36 +440,34 @@ describe('rum global context', () => {
   })
 
   it('should not be automatically snake cased', () => {
-    RUM.setRumGlobalContext({ fooBar: 'foo' })
+    const { server, lifeCycle, rumApi } = setupBuilder.build()
+    server.requests = []
```
BenoitZugmeyer, line 444: I've seen this multiple times.. The server can receive requests between its creation and the test execution? I'd be in favor of reseting the server requests in a setup builder "build task".

packages/rum/test/viewCollection.spec.ts
bcaudan: I think we can still have helper specific to spec files when it is not used elsewhere.
I'd be in favor of moving it there if we need to reuse it in another context.

packages/rum/test/rum.spec.ts
bcaudan: it depends if you are interested or not in the events generated when starting rum.

packages/rum/test/viewCollection.spec.ts
BenoitZugmeyer: Ok!

packages/rum/test/rum.spec.ts
BenoitZugmeyer: Ah, I see! All right.

PR #401, opened 2020-05-25
✨[RUMF-473] collect view loading time in ns and integrate the load event timing in the loading time calculation

mquentin:
## Motivation
- Sync the view loading time and UA duration units to all other collected duration
- Make sure to have a view loading time equal to the load event when this last one occurs after the completion of all activities

## Changes
- Convert the ms duration in ns
- compare the value of the load event and the activity completion timings and set the loading time accordingly 

## Testing
Create new views and check that the collected loading time is not in ms anymore

### DIFF
--- a/packages/rum/src/trackPageActivities.ts
+++ b/packages/rum/src/trackPageActivities.ts
@@ -1,11 +1,11 @@
 import { monitor, Observable } from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType, Subscription } from './lifeCycle'
 
-// Delay to wait for a page activity to validate the tracking process
+// Delay to wait for a page activity to validate the tracking process in ms
 export const PAGE_ACTIVITY_VALIDATION_DELAY = 100
-// Delay to wait after a page activity to end the tracking process
+// Delay to wait after a page activity to end the tracking process in ms
 export const PAGE_ACTIVITY_END_DELAY = 100
-// Maximum duration of the tracking process
+// Maximum duration of the tracking process in ms
 export const PAGE_ACTIVITY_MAX_DURATION = 10_000
 
 export interface PageActivityEvent {
--- a/packages/rum/src/trackPageActivities.ts
+++ b/packages/rum/src/trackPageActivities.ts
@@ -1,11 +1,11 @@
 import { monitor, Observable } from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType, Subscription } from './lifeCycle'
 
-// Delay to wait for a page activity to validate the tracking process in ms
+// Delay to wait for a page activity to validate the tracking process
 export const PAGE_ACTIVITY_VALIDATION_DELAY = 100
-// Delay to wait after a page activity to end the tracking process in ms
+// Delay to wait after a page activity to end the tracking process
 export const PAGE_ACTIVITY_END_DELAY = 100
-// Maximum duration of the tracking process in ms
+// Maximum duration of the tracking process
 export const PAGE_ACTIVITY_MAX_DURATION = 10_000
 
 export interface PageActivityEvent {
--- a/packages/rum/src/userActionCollection.ts
+++ b/packages/rum/src/userActionCollection.ts
@@ -1,4 +1,4 @@
-import { Context, DOM_EVENT, generateUUID, msToNs } from '@datadog/browser-core'
+import { Context, DOM_EVENT, generateUUID } from '@datadog/browser-core'
 import { getElementContent } from './getElementContent'
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { trackEventCounts } from './trackEventCounts'
--- a/packages/rum/src/userActionCollection.ts
+++ b/packages/rum/src/userActionCollection.ts
@@ -83,7 +83,7 @@ function newUserAction(lifeCycle: LifeCycle, type: UserActionType, name: string)
         name,
         startTime,
         type,
-        duration: msToNs(endTime - startTime),
+        duration: endTime - startTime,
         measures: {
           errorCount: eventCounts.errorCount,
           longTaskCount: eventCounts.longTaskCount,
--- a/packages/rum/src/userActionCollection.ts
+++ b/packages/rum/src/userActionCollection.ts
@@ -1,4 +1,4 @@
-import { Context, DOM_EVENT, generateUUID, noop } from '@datadog/browser-core'
+import { Context, DOM_EVENT, generateUUID, msToNs } from '@datadog/browser-core'
 import { getElementContent } from './getElementContent'
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { trackEventCounts } from './trackEventCounts'
@@ -83,7 +83,7 @@ function newUserAction(lifeCycle: LifeCycle, type: UserActionType, name: string)
         name,
         startTime,
         type,
-        duration: endTime - startTime,
+        duration: msToNs(endTime - startTime),
         measures: {
           errorCount: eventCounts.errorCount,
           longTaskCount: eventCounts.longTaskCount,
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -196,10 +196,10 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
 }
 
 function trackLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number | undefined) => void) {
-  const startTime: number = performance.now()
+  const startTime = performance.now()
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
     if (hadActivity) {
-      callback(endTime - startTime)
+      callback(msToNs(endTime - startTime))
     } else {
       callback(undefined)
     }
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -236,7 +236,7 @@ function trackLoadEventLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeV
     (entry) => {
       if (entry.entryType === 'navigation') {
         const navigationEntry = entry as PerformanceNavigationTiming
-        callback(msToNs(navigationEntry.loadEventEnd))
+        callback(navigationEntry.loadEventEnd)
       }
     }
   )
@@ -248,7 +248,7 @@ function trackActivityLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeVa
   const startTime = performance.now()
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
     if (hadActivity) {
-      callback(msToNs(endTime - startTime))
+      callback(endTime - startTime)
     } else {
       callback(undefined)
     }
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -106,17 +106,10 @@ function newView(
   const { stop: stopEventCountsTracking } = trackEventCounts(lifeCycle, updateMeasures)
 
   function updateLoadingTime(loadingTimeValue: number) {
-    if (loadingTime === undefined || loadingTimeValue > loadingTime) {
-      loadingTime = loadingTimeValue
-      scheduleViewUpdate()
-    }
+    loadingTime = loadingTimeValue
+    scheduleViewUpdate()
   }
-  const { stop: stopActivityLoadingTimeTracking } = trackActivityLoadingTime(lifeCycle, updateLoadingTime)
-
-  let stopLoadEventLoadingTime = noop
-  if (loadingType === ViewLoadingType.INITIAL_LOAD) {
-    ;({ stop: stopLoadEventLoadingTime } = trackLoadEventLoadingTime(lifeCycle, updateLoadingTime))
-  }
+  const { stop: stopLoadingTimeTracking } = trackLoadingTime(lifeCycle, loadingType, updateLoadingTime)
 
   // Initial view update
   updateView()
@@ -139,8 +132,7 @@ function newView(
     end() {
       stopTimingsTracking()
       stopEventCountsTracking()
-      stopActivityLoadingTimeTracking()
-      stopLoadEventLoadingTime()
+      stopLoadingTimeTracking()
       // prevent pending view updates execution
       stopScheduleViewUpdate()
       // Make a final view update
@@ -203,6 +195,41 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
   return { stop: stopPerformanceTracking }
 }
 
+function trackLoadingTime(
+  lifeCycle: LifeCycle,
+  loadingType: ViewLoadingType,
+  callback: (loadingTimeValue: number) => void
+) {
+  let expectedTiming = 1
+  const receivedTimings: number[] = []
+
+  let stopLoadEventLoadingTime = noop
+  if (loadingType === ViewLoadingType.INITIAL_LOAD) {
+    expectedTiming += 1
+    ;({ stop: stopLoadEventLoadingTime } = trackLoadEventLoadingTime(lifeCycle, onTimingValue))
+  }
+
+  const { stop: stopActivityLoadingTimeTracking } = trackActivityLoadingTime(lifeCycle, onTimingValue)
+
+  function onTimingValue(timingValue: number | undefined) {
+    expectedTiming -= 1
+    if (timingValue) {
+      receivedTimings.push(timingValue)
+    }
+
+    if (expectedTiming === 0 && receivedTimings.length) {
+      callback(Math.max(...receivedTimings))
+    }
+  }
+
+  return {
+    stop() {
+      stopActivityLoadingTimeTracking()
+      stopLoadEventLoadingTime()
+    },
+  }
+}
+
 function trackLoadEventLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number) => void) {
   const { unsubscribe: stopPerformanceTracking } = lifeCycle.subscribe(
     LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
@@ -217,11 +244,13 @@ function trackLoadEventLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeV
   return { stop: stopPerformanceTracking }
 }
 
-function trackActivityLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number) => void) {
+function trackActivityLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number | undefined) => void) {
   const startTime = performance.now()
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
     if (hadActivity) {
       callback(msToNs(endTime - startTime))
+    } else {
+      callback(undefined)
     }
   })
 
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -1,4 +1,4 @@
-import { DOM_EVENT, generateUUID, monitor, msToNs, throttle } from '@datadog/browser-core'
+import { DOM_EVENT, generateUUID, monitor, msToNs, throttle, noop } from '@datadog/browser-core'
 
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { PerformancePaintTiming } from './rum'
@@ -105,20 +105,23 @@ function newView(
   const { stop: stopTimingsTracking } = trackTimings(lifeCycle, updateMeasures)
   const { stop: stopEventCountsTracking } = trackEventCounts(lifeCycle, updateMeasures)
 
-  function updateLoadingTime(loadingTimeValue: number | undefined) {
-    loadingTime = loadingTimeValue
-    scheduleViewUpdate()
+  function updateLoadingTime(loadingTimeValue: number) {
+    if (loadingTime === undefined || loadingTimeValue > loadingTime) {
+      loadingTime = loadingTimeValue
+      scheduleViewUpdate()
+    }
   }
-  const { stop: stopLoadingTimeTracking } = trackLoadingTime(lifeCycle, updateLoadingTime)
+  const { stop: stopActivityLoadingTimeTracking } = trackActivityLoadingTime(lifeCycle, updateLoadingTime)
+
+  let loadEventLoadingTime = { stop: noop }
+  if (loadingType === ViewLoadingType.INITIAL_LOAD) {
+    loadEventLoadingTime = trackLoadEventLoadingTime(lifeCycle, startOrigin, updateLoadingTime)
+  }
 
   // Initial view update
   updateView()
 
   function updateView() {
-    if (measures.loadEventEnd) {
-      loadingTime = Math.max(measures.loadEventEnd - startOrigin, loadingTime || 0)
-    }
-
     documentVersion += 1
     lifeCycle.notify(LifeCycleEventType.VIEW_COLLECTED, {
       documentVersion,
@@ -136,7 +139,8 @@ function newView(
     end() {
       stopTimingsTracking()
       stopEventCountsTracking()
-      stopLoadingTimeTracking()
+      stopActivityLoadingTimeTracking()
+      loadEventLoadingTime.stop()
       // prevent pending view updates execution
       stopScheduleViewUpdate()
       // Make a final view update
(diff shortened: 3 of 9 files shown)


packages/rum/src/userActionCollection.ts
```diff
@@ -83,7 +83,7 @@ function newUserAction(lifeCycle: LifeCycle, type: UserActionType, name: string)
         name,
         startTime,
         type,
-        duration: endTime - startTime,
+        duration: msToNs(endTime - startTime),
```
BenoitZugmeyer, line 86: This seems problematic to me, because the `msToNs` convertion will be done here too: https://github.com/DataDog/browser-sdk/blob/e03888639ab60f1b55cb91043958af281a9329ea/packages/rum/src/rum.ts#L317

Like other lifecycle events `duration` properties are in milliseconds, I don't think we should change this here.

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,65 @@ describe('rum view measures', () => {
...
+
+    expect(getRumEventCount()).toEqual(1)
+
+    jasmine.clock().tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(getViewEvent(1).loadingTime).toBeLessThan(567e6)
```
BenoitZugmeyer, line 355: Would this be right?
```suggestion
    expect(getViewEvent(1).loadingTime).toBe(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
```

mquentin: no it is not. I started with this approach but I decided to go for the `toBeLessThan(567e6)` which is what we want to test

packages/rum/src/userActionCollection.ts
mquentin: my bad then

packages/rum/test/viewCollection.spec.ts
bcaudan, line 355: what does 567e6 mean?

packages/rum/src/trackPageActivities.ts
```diff
@@ -1,11 +1,11 @@
...
 import { LifeCycle, LifeCycleEventType, Subscription } from './lifeCycle'
 
-// Delay to wait for a page activity to validate the tracking process
+// Delay to wait for a page activity to validate the tracking process in ms
 export const PAGE_ACTIVITY_VALIDATION_DELAY = 100
-// Delay to wait after a page activity to end the tracking process
+// Delay to wait after a page activity to end the tracking process in ms
 export const PAGE_ACTIVITY_END_DELAY = 100
-// Maximum duration of the tracking process
+// Maximum duration of the tracking process in ms
```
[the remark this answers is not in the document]
mquentin: It clarifies the unit of the constant as we have some outputs in ns and some in ms and helps to see if the variables are homogeneous in the durations calculation.

packages/rum/src/viewCollection.ts
```diff
@@ -115,6 +115,10 @@ function newView(
   updateView()
 
   function updateView() {
+    if (measures.loadEventEnd) {
+      loadingTime = Math.max(measures.loadEventEnd - startOrigin, loadingTime || 0)
+    }
```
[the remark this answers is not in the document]
mquentin: at any time, if you update the view ID, you check the value of `measures.loadEventEnd - startOrigin` vs loadingTime and you take the highest value. 
So in you example, and I all other scenarios, the loadevent timing will be set as a load time is it is higher that the load time.

packages/rum/src/trackPageActivities.ts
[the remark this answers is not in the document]
bcaudan: OK IMO it is maybe not the right approach since all the constants should be in ms and time usage in the code base should be in ms.
Conversion to ns should be done as late as possible because it is only needed for the intake format.

packages/rum/src/viewCollection.ts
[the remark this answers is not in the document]
bcaudan: > at any time, if you update the view ID, you check the value of measures.loadEventEnd - startOrigin vs loadingTime and you take the highest value.

you should not, load event should only be considered for initial view

> So in you example, the same view id will first have the calculated activity loading time, then some view updates occurs and finally the load event will trigger a last update that will change the measures as well as the loading time

in my example, the last view update should happen before the load event so the I am not sure what the loading time should be.

packages/rum/src/viewCollection.ts
```diff
@@ -196,10 +200,10 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
 }
 
 function trackLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number | undefined) => void) {
-  const startTime: number = performance.now()
+  const startTime = performance.now()
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
     if (hadActivity) {
-      callback(endTime - startTime)
+      callback(msToNs(endTime - startTime))
```
bcaudan, line 206: what about doing this conversion inside `rum.ts`, like duration?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
     jasmine.clock().uninstall()
   })
 
+  it('should have a loadEventEnd loading time when having no activity', () => {
```
bcaudan, line 311: no, for initial view, it should use loadEvent when it is bigger than loading time

bcaudan, line 311:
```suggestion
  it('should use loadEventEnd for initial view when having no activity', () => {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
+      FAKE_NAVIGATION_ENTRY as PerformanceNavigationTiming
+    )
+
+    expect(getRumEventCount()).toEqual(1)
+
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(getViewEvent(1).measures.loadEventEnd).toEqual(
+      msToNs(FAKE_NAVIGATION_ENTRY.loadEventEnd - getViewEvent(1).startTime)
```
bcaudan, line 332: For initial view, startTime should be 0, no?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
+    )
+    expect(getViewEvent(1).loadingTime).toEqual(getViewEvent(1).measures.loadEventEnd)
+
+    jasmine.clock().uninstall()
+  })
+
+  it('should have a loadEventEnd loading time when load event happens after all activity completions', () => {
```
bcaudan, line 339:
```suggestion
  it('should use loadEventEnd for initial view load event is bigger than computed loading time', () => {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
 
+  it('should have a loadEventEnd loading time when having no activity', () => {
+    jasmine.clock().install()
+    expect(getRumEventCount()).toEqual(1)
+    expect(getViewEvent(0).measures).toEqual({
+      errorCount: 0,
+      longTaskCount: 0,
+      resourceCount: 0,
+      userActionCount: 0,
+    })
```
bcaudan, line 319: why do we care about measures here?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
+
+    lifeCycle.notify(
+      LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
+      FAKE_NAVIGATION_ENTRY as PerformanceNavigationTiming
+    )
+
+    expect(getRumEventCount()).toEqual(1)
```
bcaudan, line 326: why do we need to recheck that?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
+
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(getViewEvent(1).measures.loadEventEnd).toEqual(
```
bcaudan, line 331: why should we check view load event value in this test?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
+
+    jasmine.clock().tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(getViewEvent(1).loadingTime).toBeLessThan(
+      msToNs(FAKE_NAVIGATION_ENTRY.loadEventEnd - getViewEvent(1).startTime)
+    )
```
bcaudan, line 359: I would be in favor of not setting the loading time of the initial view before receiving the load event, so I would prefer to not enforce this behavior in a test

packages/rum/test/viewCollection.spec.ts
```diff
@@ -308,6 +308,71 @@ describe('rum view measures', () => {
...
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(3)
+    expect(getViewEvent(2).measures.loadEventEnd).toEqual(
+      msToNs(FAKE_NAVIGATION_ENTRY.loadEventEnd - getViewEvent(2).startTime)
+    )
+    expect(getViewEvent(2).loadingTime).toEqual(getViewEvent(2).measures.loadEventEnd)
+
+    jasmine.clock().uninstall()
+  })
```
bcaudan, line 374: Could you add a test where loading time is bigger than load event for the initial view?

packages/rum/src/viewCollection.ts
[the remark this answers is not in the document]
mquentin: > you should not, load event should only be considered for initial view

Then we need to rethink all the trackTimings` and view update process as `loadEventEnd` is updated as a regular view measure like errorCount see [updateMeasures](https://github.com/DataDog/browser-sdk/blob/07555f0ab27888bb0b2875e59807b3e7a93a103f/packages/rum/src/viewCollection.ts#L101)
Also, `trackTimings`, `trackEventCounts` and `trackLoadingTime` are all run simultaneously and have unrelated ending processes which makes my solution the safest one

>in my example, the last view update should happen before the load event so the I am not sure what the loading time should be.

 The `load event` is updated, through the `trackTimings`, like any view measure and would trigger a n + 1 update which would use this timing as a loading time.

packages/rum/src/userActionCollection.ts
```diff
@@ -1,4 +1,4 @@
-import { Context, DOM_EVENT, generateUUID, noop } from '@datadog/browser-core'
+import { Context, DOM_EVENT, generateUUID, msToNs } from '@datadog/browser-core'
```
bcaudan, line 1: unused import

packages/rum/src/viewCollection.ts
```diff
@@ -195,13 +203,30 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
...
 }
 
-function trackLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number | undefined) => void) {
-  const startTime: number = performance.now()
+function trackLoadEventLoadingTime(
+  lifeCycle: LifeCycle,
+  startOrigin: number,
+  callback: (loadingTimeValue: number) => void
+) {
+  const startTime = performance.now()
```
bcaudan, line 211: unused

packages/rum/src/viewCollection.ts
```diff
@@ -195,13 +203,30 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
...
+  const { unsubscribe: stopPerformanceTracking } = lifeCycle.subscribe(
+    LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
+    (entry) => {
+      if (entry.entryType === 'navigation') {
+        const navigationEntry = entry as PerformanceNavigationTiming
+        callback(msToNs(navigationEntry.loadEventEnd - startOrigin))
```
bcaudan, line 217:
```suggestion
        callback(msToNs(navigationEntry.loadEventEnd))
```
startOrigin should be 0 for initial view

packages/rum/src/viewCollection.ts
```diff
@@ -105,11 +105,18 @@ function newView(
   const { stop: stopTimingsTracking } = trackTimings(lifeCycle, updateMeasures)
   const { stop: stopEventCountsTracking } = trackEventCounts(lifeCycle, updateMeasures)
 
-  function updateLoadingTime(loadingTimeValue: number | undefined) {
-    loadingTime = loadingTimeValue
-    scheduleViewUpdate()
+  function updateLoadingTime(loadingTimeValue: number) {
+    if (loadingTime === undefined || loadingTimeValue > loadingTime) {
```
bcaudan, line 109: IMO, it would be easier to maintain if all the logic related to loading time is encapsulated. 
It would also be easier to understand the intent by expressing that:
- we are waiting several timings
- we want the max of them
It would avoid to send an update with an incorrect or changing loadingTime value .

We could then have something like:

```
{stop: stopLoadingTime} = trackLoadingTime(lifecycle, loadingType, callback) {
  let expectedTiming = 1
  const receivedTimings = []
  if (initial view) {
    expectedTiming++
    // load event logic
  }

  // activity logic

  function onTimingValue(timingValue) {
    receivedTimings.push(timingValue)
    expectedTiming--
    if (expectedTiming === 0) {
      callback(Math.max(receivedTimings))
    }
  }

  return {
    stop() {
      // stop load event and activity
    }
  }
}
```

What do you think?

packages/rum/test/userActionCollection.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { DOM_EVENT, ErrorMessage } from '@datadog/browser-core'
+import { DOM_EVENT, ErrorMessage, msToNs } from '@datadog/browser-core'
```
bcaudan, line 1: unused

packages/rum/src/viewCollection.ts
```diff
@@ -105,11 +105,18 @@ function newView(
...
+      loadingTime = loadingTimeValue
+      scheduleViewUpdate()
+    }
+  }
+  const { stop: stopActivityLoadingTimeTracking } = trackActivityLoadingTime(lifeCycle, updateLoadingTime)
+
+  let loadEventLoadingTime = { stop: noop }
+  if (loadingType === ViewLoadingType.INITIAL_LOAD) {
+    loadEventLoadingTime = trackLoadEventLoadingTime(lifeCycle, startOrigin, updateLoadingTime)
```
bcaudan, line 118:
```suggestion
    ;({stop: stopLoadEventLoadingTime} = trackLoadEventLoadingTime(lifeCycle, startOrigin, updateLoadingTime))
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -170,7 +172,81 @@ describe('rum track loading time', () => {
...
+  }
+  const FAKE_NAVIGATION_ENTRY_2 = {
+    domComplete: 2,
+    domContentLoadedEventEnd: 1,
+    domInteractive: 1,
+    entryType: 'navigation',
+    loadEventEnd: 1,
+  }
+
+  it('should use loadEventEnd for initial view when having no activity', () => {
```
bcaudan, line 193: can you keep these constants inside the corresponding tests or move them with the other constants at the top of the file?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -170,7 +172,81 @@ describe('rum track loading time', () => {
...
+
+    expect(getRumEventCount()).toEqual(1)
+
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(getViewEvent(1).loadingTime).toEqual(msToNs(FAKE_NAVIGATION_ENTRY.loadEventEnd))
+  })
+
+  it('should use loadEventEnd for initial view when load event is bigger than computed loading time  ', () => {
```
bcaudan, line 215:
```suggestion
  it('should use loadEventEnd for initial view when load event is bigger than computed loading time', () => {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -170,7 +172,81 @@ describe('rum track loading time', () => {
...
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(FAKE_NAVIGATION_ENTRY.loadEventEnd).toBeGreaterThan(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    expect(getViewEvent(1).loadingTime).toEqual(msToNs(FAKE_NAVIGATION_ENTRY.loadEventEnd))
+  })
+
+  it('should not use loadEventEnd for initial view when load event is smaller than computed loading time  ', () => {
```
bcaudan, line 235:
```suggestion
  it('should not use loadEventEnd for initial view when load event is smaller than computed loading time', () => {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -170,7 +172,81 @@ describe('rum track loading time', () => {
...
+    entryType: 'navigation',
+    loadEventEnd: 567,
+  }
+  const FAKE_NAVIGATION_ENTRY_2 = {
+    domComplete: 2,
+    domContentLoadedEventEnd: 1,
+    domInteractive: 1,
+    entryType: 'navigation',
+    loadEventEnd: 1,
+  }
```
bcaudan, line 191: if they are not random navigation entries, can you improve their names?

packages/rum/src/trackPageActivities.ts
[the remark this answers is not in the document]
bcaudan: so?

packages/rum/src/viewCollection.ts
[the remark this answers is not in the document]
bcaudan: seen together

packages/rum/test/viewCollection.spec.ts
bcaudan: so?

packages/rum/test/viewCollection.spec.ts
bcaudan: ?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -170,7 +172,81 @@ describe('rum track loading time', () => {
...
+      FAKE_NAVIGATION_ENTRY as PerformanceNavigationTiming
+    )
+
+    lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(FAKE_NAVIGATION_ENTRY.loadEventEnd).toBeGreaterThan(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
```
bcaudan, line 231: `FAKE_NAVIGATION_ENTRY.loadEventEnd` value does not depend on the behavior of the implementation, no need to check that you have correctly set your tests inputs

packages/rum/test/viewCollection.spec.ts
bcaudan: phrase without negation are easier to grasp, let's limit negation when possible

packages/rum/test/viewCollection.spec.ts
```diff
@@ -170,7 +172,81 @@ describe('rum track loading time', () => {
...
+    lifeCycle.notify(
+      LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED,
+      FAKE_NAVIGATION_ENTRY_2 as PerformanceNavigationTiming
+    )
+    lifeCycle.notify(LifeCycleEventType.DOM_MUTATED)
+    jasmine.clock().tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+    jasmine.clock().tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getRumEventCount()).toEqual(2)
+    expect(FAKE_NAVIGATION_ENTRY_2.loadEventEnd).toBeLessThan(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
```
bcaudan, line 248: FAKE_NAVIGATION_ENTRY_2.loadEventEnd value does not depend on the behavior of the implementation, no need to check that you have correctly set your tests inputs

packages/rum/src/trackPageActivities.ts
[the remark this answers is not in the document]
mquentin: so I think it is still relevant to specify that theses constants are in ms as it gives more context and helps understanding their purpose. And you are right, as ms is the standard, I will keep the `msToNs` outside of the tracking timing logic.

packages/rum/test/viewCollection.spec.ts
mquentin: for no specific reason. I rm the lines

packages/rum/test/viewCollection.spec.ts
mquentin: `expect(getRumEventCount()).toEqual(1)` and `expect(getRumEventCount()).toEqual(2)` test that the notified event triggers a view update. It allows to test the new value of `loadingTime` being sure that an update have been completed before.

packages/rum/test/viewCollection.spec.ts
mquentin: If someone modify this constant, this test will prevent any regression. Also it gives more context and make the test easier to understand.

packages/rum/test/viewCollection.spec.ts
mquentin: same comment, I think it gives some context to this test and validate the constant arbitrary values.

packages/rum/src/viewCollection.ts
mquentin: It would be easier to understand for sure. I will use this structure to encapsulate the two `trackLoadEventLoadingTime` and `trackActivityLoadingTime` calls as it will also simplify the implementation of your approach.

packages/rum/src/trackPageActivities.ts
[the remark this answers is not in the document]
bcaudan: I am not against it but if we do that here we should also do it for every other time constants in the code base

packages/rum/test/viewCollection.spec.ts
bcaudan: current state is ok, in previous state you were checking it twice

packages/rum/test/viewCollection.spec.ts
bcaudan: if `FAKE_NAVIGATION_ENTRY.loadEventEnd < BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY`, the next expectation should fail which should prevent the regression, no?
To make it easier easier to understand, you could rename your entry like in next test, something like `FAKE_NAVIGATION_ENTRY_WITH_SLOW_LOADEVENT_TIMING`

mquentin: I renamed it `FAKE_NAVIGATION_ENTRY_WITH_FAST_LOADEVENT_TIMING` to clarify the entry purpose and yes the next expectation would fail but given the no trivial concept of loadtiming, I though this two lines `toBeLessThan` and `toBeGreaterThan` help a lot.

packages/rum/src/trackPageActivities.ts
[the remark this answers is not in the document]
mquentin: ok, out of the scope of this PR then

packages/rum/src/viewCollection.ts
```diff
@@ -195,11 +195,60 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
...
+
+  return { stop: stopPerformanceTracking }
+}
+
+function trackActivityLoadingTime(lifeCycle: LifeCycle, callback: (loadingTimeValue: number | undefined) => void) {
+  const startTime = performance.now()
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
     if (hadActivity) {
-      callback(endTime - startTime)
+      callback(msToNs(endTime - startTime))
```
bcaudan, line 251: what about doing the conversion in rum.ts like for duration?
then we only have to do it once and don't need to add this extra noise in the test

packages/rum/test/viewCollection.spec.ts
bcaudan: What could we do to make that easy to understand but avoid to perform useless check?

mquentin: For me they are not useless as they validate the assumptions that make these two tests valid

packages/rum/test/viewCollection.spec.ts
```diff
@@ -40,12 +40,20 @@ const FAKE_NAVIGATION_ENTRY = {
...
 }
 
-const FAKE_NAVIGATION_ENTRY_WITH_FAST_LOADEVENT_TIMING = {
+const FAKE_NAVIGATION_ENTRY_WITH_LOADEVENT_BEFORE_ACTIVITY_TIMING = {
   domComplete: 2,
   domContentLoadedEventEnd: 1,
   domInteractive: 1,
   entryType: 'navigation',
-  loadEventEnd: 1,
+  loadEventEnd: BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY * 0.8,
```
bcaudan, line 48: nice

PR #406, opened 2020-05-28
✨ [RUMF-385] improve click action naming

BenoitZugmeyer:
## Motivation

Improve the name of click actions.

## Changes

* Improve the name extraction from elements by adding different strategies and adjusting priorities.
* Don't collect actions if a name cannot be found.

## Testing

This should be left on staging for a little while to verify that it does not break thing, and indeed improve the action names.

---

### DIFF
--- a/packages/rum/src/getActionNameFromElement.ts
+++ b/packages/rum/src/getActionNameFromElement.ts
@@ -1,14 +1,14 @@
-export function getElementContent(element: Element): string {
+export function getActionNameFromElement(element: Element): string {
   return (
-    getElementContentForGetters(element, priorityGetters) || getElementContentForGetters(element, fallbackGetters) || ''
+    getActionNameFromElementForGetters(element, priorityGetters) ||
+    getActionNameFromElementForGetters(element, fallbackGetters) ||
+    ''
   )
 }
 
-type ContentGetter = (
-  element: Element | HTMLElement | HTMLInputElement | HTMLSelectElement
-) => string | undefined | null
+type NameGetter = (element: Element | HTMLElement | HTMLInputElement | HTMLSelectElement) => string | undefined | null
 
-const priorityGetters: ContentGetter[] = [
+const priorityGetters: NameGetter[] = [
   // associated LABEL text
   (element) => {
     // IE does not support element.labels, so we fallback to a CSS selector based on the element id
@@ -69,14 +69,14 @@ const priorityGetters: ContentGetter[] = [
   },
 ]
 
-const fallbackGetters: ContentGetter[] = [
+const fallbackGetters: NameGetter[] = [
   // element text
   (element) => {
     return getTextualContent(element)
   },
 ]
 
-function getElementContentForGetters(targetElement: Element, getters: ContentGetter[]) {
+function getActionNameFromElementForGetters(targetElement: Element, getters: NameGetter[]) {
   let element: Element | null = targetElement
   let recursionCounter = 0
   while (
@@ -87,11 +87,11 @@ function getElementContentForGetters(targetElement: Element, getters: ContentGet
     element.nodeName !== 'HEAD'
   ) {
     for (const getter of getters) {
-      const content = getter(element)
-      if (typeof content === 'string') {
-        const trimedContent = content.trim()
-        if (trimedContent) {
-          return truncate(normalizeWhitespace(trimedContent))
+      const name = getter(element)
+      if (typeof name === 'string') {
+        const trimedName = name.trim()
+        if (trimedName) {
+          return truncate(normalizeWhitespace(trimedName))
         }
       }
     }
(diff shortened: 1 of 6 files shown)


packages/rum/src/getActionNameFromElement.ts
```diff
@@ -0,0 +1,176 @@
...
+  return s.replace(/\s+/g, ' ')
+}
+
+function truncate(s: string) {
+  return s.length > 100 ? `${s.slice(0, 100)} [...]` : s
+}
+
+function getElementById(refElement: Element, id: string) {
+  // tslint:disable-next-line: no-null-keyword
+  return refElement.ownerDocument ? refElement.ownerDocument.getElementById(id) : null
```
bcaudan, line 117: what is the reasoning for using `elt.ownerDocument` instead of `document`?

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -0,0 +1,176 @@
...
+  if ((element as HTMLElement).isContentEditable) {
+    return
+  }
+
+  if ('innerText' in element) {
+    let text = element.innerText
+    if (!supportsInnerTextScriptAndStyleRemoval()) {
+      // remove the inner text of SCRIPT and STYLES from the result. This is a bit dirty, but should
+      // be relatively fast and work in most cases.
+      const elementsToRemove: NodeListOf<HTMLElement> = element.querySelectorAll('script, style')
```
bcaudan, line 130:
```suggestion
      const elementsTextToRemove: NodeListOf<HTMLElement> = element.querySelectorAll('script, style')
```

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -0,0 +1,176 @@
...
+      return getTextualContent(element.options[0])
+    }
+  },
+]
+
+const fallbackGetters: NameGetter[] = [
+  // element text
+  (element) => {
+    return getTextualContent(element)
+  },
```
bcaudan, line 76: IMO, function name seems clear enough to not need a comment

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  if ('innerText' in element) {
+    let text = element.innerText
+    if (!supportsInnerTextScriptAndStyleRemoval()) {
+      // remove the inner text of SCRIPT and STYLES from the result. This is a bit dirty, but should
+      // be relatively fast and work in most cases.
+      const elementsToRemove: NodeListOf<HTMLElement> = element.querySelectorAll('script, style')
+      // tslint:disable-next-line: prefer-for-of
+      for (let i = 0; i < elementsToRemove.length; i += 1) {
```
bcaudan, line 132: according to MDN for-of should work correctly on node list, typing does not seem to allow that though.
it could worth to check if it is supported by newer ts version but not important now

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -0,0 +1,176 @@
...
+  // alt attribute value
+  (element) => element.getAttribute('alt'),
+  // name attribute value
+  (element) => element.getAttribute('name'),
+  // title attribute value
+  (element) => element.getAttribute('title'),
+  // placeholder attribute value
+  (element) => element.getAttribute('placeholder'),
+  // SELECT first OPTION text
+  (element) => {
```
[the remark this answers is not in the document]
BenoitZugmeyer: I thought this was nice to have comments for every getters so it's easier to scan. But you are right, attributes getters are very trivial so it may not be really needed. I can remove those if you think it's too much.

packages/rum/src/getActionNameFromElement.ts
BenoitZugmeyer: Tests are using iframes to emulate a full page context. The `document.getElementeById` doesn't work in this case. Also it may be useful sometime if we want to support iframes in some ways.

packages/rum/src/getActionNameFromElement.ts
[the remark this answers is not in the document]
bcaudan: I think this kind of comments can be useful to separate chunks of related behaviors in a given context.
Here, the functions are quite simple so IMO it adds more noise than it helps the understanding.

packages/rum/src/getActionNameFromElement.ts
bcaudan: iframe support could be quite challenging 😄 
since it is not the go to API for that, I think a small comment could be helpful

packages/rum/src/getActionNameFromElement.ts
BenoitZugmeyer: It appears to be working on two conditions:
* use the DOM.Iterable `lib`
* and use ES6+ as a `target` or the `downlevelIteration` option
I'll keep the traditional `for` loop here for now.

PR #412, opened 2020-06-02
✨ [RUMF-385] implement a declarative API to set the action names

BenoitZugmeyer:
## Motivation

We want to allow RUM library users to declare action names directly within the HTML markup (or DOM tree).

## Changes

Add a new, top priority strategy to the `getActionNameFromElement` function that look for a `data-dd-action-name` attribute on the element or any of its parent.

## Testing

Unit tests should be enough

---

### DIFF
--- a/packages/rum/README.md
+++ b/packages/rum/README.md
@@ -75,12 +75,20 @@ datadogRum.init({
 
 ### Click action naming
 
-The RUM library will use various strategies to get a name for click actions, but if you want more
+The RUM library is using various strategies to get a name for click actions, but if you want more
 control, you can define a `data-dd-action-name` attribute on clickable elements (or any of their
-parents) that will be used to name the action. Example:
+parents) that will be used to name the action. Examples:
 
 ```html
 <a class="btn btn-default" href="#" role="button" data-dd-action-name="Login button">Try it out!</a>
+```
+
+```html
+<div class="alert alert-danger" role="alert" data-dd-action-name="Dismiss alert">
+  <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span>
+  <span class="sr-only">Error:</span>
+  Enter a valid email address
+</div>
 ```
 
 ## TypeScript support
--- a/packages/rum/README.md
+++ b/packages/rum/README.md
@@ -71,6 +71,18 @@ datadogRum.init({
   addUserAction (name: string, context: Context)
   ```
 
+## Declarative API
+
+### Click action naming
+
+The RUM library will use various strategies to get a name for click actions, but if you want more
+control, you can define a `data-dd-action-name` attribute on clickable elements (or any of their
+parents) that will be used to name the action. Example:
+
+```html
+<a class="btn btn-default" href="#" role="button" data-dd-action-name="Login button">Try it out!</a>
+```
+
 ## TypeScript support
 
 Types are compatible with TypeScript >= 3.0.
--- a/packages/rum/src/getActionNameFromElement.ts
+++ b/packages/rum/src/getActionNameFromElement.ts
@@ -1,6 +1,6 @@
 export function getActionNameFromElement(element: Element): string {
   // Proceed to get the action name in two steps:
-  // * first, get the name programmatically, explicitely defined by the user.
+  // * first, get the name programmatically, explicitly defined by the user.
   // * then, use strategies that are known to return good results. Those strategies will be used on
   //   the element and a few parents, but it's likely that they won't succeed at all.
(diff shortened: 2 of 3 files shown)


packages/rum/README.md
```diff
@@ -71,6 +71,18 @@ datadogRum.init({
   addUserAction (name: string, context: Context)
   ```
 
+## Declarative API
+
+### Click action naming
+
+The RUM library will use various strategies to get a name for click actions, but if you want more
```
bcaudan, line 78:
```suggestion
The RUM library is using various strategies to get a name for click actions, but if you want more
```
future tense seems a bit weird to me

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -1,16 +1,48 @@
 export function getActionNameFromElement(element: Element): string {
   // Proceed to get the action name in two steps:
-  // * first, use strategies that are known to return good results. Those strategies will be used on
+  // * first, get the name programmatically, explicitely defined by the user.
+  // * then, use strategies that are known to return good results. Those strategies will be used on
```
bcaudan, line 4: what do you think of renaming "getters" to "strategies" in order to have a more consistent wording between comments and implementation?

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -1,16 +1,48 @@
...
   )
 }
 
+/**
+ * Get the action name from the attribute 'data-dd-action-name' on the element or any of its parent.
+ */
+const PROGRAMMATIC_ATTRIBUTE = 'data-dd-action-name'
+function getActionNameFromElementProgrammatically(targetElement: Element) {
+  let elementWithAttribute
+  // We don't use getActionNameFromElementForGetters here, because we want to consider all parents,
```
bcaudan, line 22: What is the reasoning for considering all parents and not some parents? or only the target element?

BenoitZugmeyer: > What is the reasoning for considering all parents and not some parents?
Contrary to other strategies, where we try our best to be smart by using heuristics based on how interfaces are built in general, the programmatic one should be dumb, trivial and unsurprising. This allows users to experiment how they want, without limit.

> or only the target element?
I find it less convenient, as a button (or other control) is often composed of multiple elements, so every elements would need to have the attribute to consider the button as a whole.

packages/rum/src/getActionNameFromElement.ts
BenoitZugmeyer: Ok I'll rename this

packages/rum/src/getActionNameFromElement.ts
bcaudan: OK, we should probably also document this behavior in the API then.

BenoitZugmeyer: I already wrote
> on clickable elements (or any of their parents)
Do you think I should elaborate?

bcaudan: I quickly checked and missed it 🤦 
A second example with a parent case could be clearer

BenoitZugmeyer: Ok I'll add it!

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -1,19 +1,51 @@
 export function getActionNameFromElement(element: Element): string {
   // Proceed to get the action name in two steps:
-  // * first, use strategies that are known to return good results. Those strategies will be used on
+  // * first, get the name programmatically, explicitely defined by the user.
```
bcaudan, line 3:
```suggestion
  // * first, get the name programmatically, explicitly defined by the user.
```

packages/rum/src/getActionNameFromElement.ts
```diff
@@ -1,19 +1,51 @@
...
 
-type NameGetter = (element: Element | HTMLElement | HTMLInputElement | HTMLSelectElement) => string | undefined | null
+/**
+ * Get the action name from the attribute 'data-dd-action-name' on the element or any of its parent.
+ */
+const PROGRAMMATIC_ATTRIBUTE = 'data-dd-action-name'
+function getActionNameFromElementProgrammatically(targetElement: Element) {
+  let elementWithAttribute
+  // We don't use getActionNameFromElementForStrategies here, because we want to consider all parents,
+  // without limit. It is up to the user to declare a relevent naming strategy.
```
bcaudan, line 23:
```suggestion
  // without limit. It is up to the user to declare a relevant naming strategy.
```

PR #413, opened 2020-06-03
⚡️[RUMF-510] Improve sizeInByte calculation performance

mquentin:
### What it does
Improve RUM and Log SDK performance when numerous console.error occur.

## Before
1000 console error loaded in 15000ms 
Capture d’écran 2020-06-03 à 08 59 11

## After
1000 console error loaded in 7000ms 
Capture d’écran 2020-06-03 à 09 01 51

### Changes
- ⚡️ improve sizeInByte calculation

### DIFF
--- a/packages/core/src/transport.ts
+++ b/packages/core/src/transport.ts
@@ -133,7 +133,7 @@ export class Batch<T> {
 
   private maxSizeInBytes(candidate: string) {
     // the candidate byte size, once URI encoded can be maximized by 4 time its length
-    // Using this maximized value prevent any costly calculation while checking the byte size requirement
+    // Using this maximized value prevent any costly calculation while validating the byte size requirement
     return candidate.length * 4
   }
 
--- a/packages/core/src/transport.ts
+++ b/packages/core/src/transport.ts
@@ -89,7 +89,7 @@ export class Batch<T> {
   private process(message: T) {
     const contextualizedMessage = deepMerge({}, this.contextProvider(), (message as unknown) as Context) as Context
     const processedMessage = jsonStringify(contextualizedMessage)!
-    const messageBytesSize = this.sizeInBytes(processedMessage)
+    const messageBytesSize = this.maxSizeInBytes(processedMessage)
     return { processedMessage, messageBytesSize }
   }
 
@@ -110,7 +110,7 @@ export class Batch<T> {
   private remove(key: string) {
     const removedMessage = this.upsertBuffer[key]
     delete this.upsertBuffer[key]
-    const messageBytesSize = this.sizeInBytes(removedMessage)
+    const messageBytesSize = this.maxSizeInBytes(removedMessage)
     this.bufferBytesSize -= messageBytesSize
     this.bufferMessageCount -= 1
     if (this.bufferMessageCount > 0) {
@@ -131,9 +131,10 @@ export class Batch<T> {
     return this.bufferMessageCount === this.maxSize || this.bufferBytesSize >= this.bytesLimit
   }
 
-  private sizeInBytes(candidate: string) {
-    // tslint:disable-next-line no-bitwise
-    return ~-encodeURI(candidate).split(/%..|./).length
+  private maxSizeInBytes(candidate: string) {
+    // the candidate byte size, once URI encoded can be maximized by 4 time its length
+    // Using this maximized value prevent any costly calculation while checking the byte size requirement
+    return candidate.length * 4
   }
 
   private flushPeriodically() {
--- a/packages/core/src/transport.ts
+++ b/packages/core/src/transport.ts
@@ -2,7 +2,7 @@ import { monitor } from './internalMonitoring'
 import { Context, deepMerge, DOM_EVENT, jsonStringify, noop, objectValues } from './utils'
 
 // https://en.wikipedia.org/wiki/UTF-8
-const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/g
+const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/
 
 /**
  * Use POST request without content type to:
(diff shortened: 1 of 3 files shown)


packages/core/src/transport.ts
```diff
@@ -131,9 +131,10 @@ export class Batch<T> {
     return this.bufferMessageCount === this.maxSize || this.bufferBytesSize >= this.bytesLimit
   }
 
-  private sizeInBytes(candidate: string) {
-    // tslint:disable-next-line no-bitwise
-    return ~-encodeURI(candidate).split(/%..|./).length
+  private maxSizeInBytes(candidate: string) {
+    // the candidate byte size, once URI encoded can be maximized by 4 time its length
+    // Using this maximized value prevent any costly calculation while validating the byte size requirement
+    return candidate.length * 4
```
bcaudan, line 137: FMU, in general case bytes size = length, so with this change we are going to generate 4 times more requests to the intake, I am not sure they will be ok with that.
could a more optimist estimation work? like +10% of the size

mquentin: We would generate 4 times more requests if all the batch are 4000+ characters. For a 3000 long batch, it is still under the 16kb limit.

packages/core/src/transport.ts
```diff
@@ -132,8 +133,12 @@ export class Batch<T> {
...
 
   private sizeInBytes(candidate: string) {
-    // tslint:disable-next-line no-bitwise
-    return ~-encodeURI(candidate).split(/%..|./).length
+    // Using this approximate value prevent any costly calculation while validating the byte size requirement
+    if (!this.biggerThanOneByteRegex.test(candidate)) {
+      return candidate.length
+    }
+
+    return candidate.length + 4 * (candidate.match(this.biggerThanOneByteRegex) || '').length
```
BenoitZugmeyer, line 141:
```js
'é'.match(/([^\u0000-\u007F])/).length // 2
```
so if candidate = 'é', the size in byte will be `1 + 4 * 2 = 9` 

You probably want to have a regex like `/[^\u0000-\u007F]/g`.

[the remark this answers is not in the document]
mquentin: Capture d’écran 2020-06-03 à 17 27 11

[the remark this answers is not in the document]
bcaudan: so?

packages/core/test/transport.spec.ts
```diff
@@ -122,17 +122,20 @@ describe('batch', () => {
...
+    )
   })
 
   it('should call send one time when the size is too high and the batch is empty', () => {
-    const message = '101 bytes - xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
+    const message = '111 bytes - 🪐🪐xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
     batch.add({ message })
-    expect(transport.send).toHaveBeenCalledWith(`{"foo":"bar","message":"${message}"}`, 101)
+    expect(transport.send).toHaveBeenCalledWith(`{"foo":"bar","message":"${message}"}`, 111)
   })
```
bcaudan, line 139: These tests were dedicated to specific behaviors, can you create new ones for the behaviors you want to test?

packages/core/src/transport.ts
```diff
@@ -35,6 +35,7 @@ export class Batch<T> {
   private upsertBuffer: { [key: string]: string } = {}
   private bufferBytesSize = 0
   private bufferMessageCount = 0
+  private biggerThanOneByteRegex = /([^\u0000-\u007F])/
```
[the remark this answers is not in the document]
mquentin: I double checked and corrected the initial `var regex = /([^\u0000-\u00ff])/` [https://en.wikipedia.org/wiki/UTF-8](https://en.wikipedia.org/wiki/UTF-8) 
Capture d’écran 2020-06-03 à 17 29 59

Do you think. it would be suitable to link wikipedia ?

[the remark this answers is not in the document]
bcaudan: why u00ff then?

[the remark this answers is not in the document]
mquentin: I corrected u00ff to u007F

packages/core/src/transport.ts
```diff
@@ -68,6 +69,19 @@ export class Batch<T> {
     }
   }
 
+  sizeInBytes(candidate: string) {
+    // Accurate byte size computations can degrade performances when there is a lot of events to process
+    if (!HAS_MULTI_BYTES_CHARACTERS.test(candidate)) {
+      return candidate.length
+    }
+
+    if (isIE()) {
```
bcaudan, line 78: ⚠️  isIE is only used for unit test and must not be used in prod code because it relies on simple UA matching.
You should use feature detection instead.

packages/core/src/transport.ts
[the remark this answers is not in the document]
bcaudan: oh ok, link to wikipedia seems fine

packages/core/src/transport.ts
```diff
@@ -68,6 +71,24 @@ export class Batch<T> {
...
+    if (window.TextEncoder !== undefined) {
+      return new TextEncoder().encode(candidate).length
+    }
+
+    if (window.Blob !== undefined) {
+      return new Blob([candidate]).size
+    }
+
+    // tslint:disable-next-line no-bitwise
+    return ~-encodeURI(candidate).split(/%..|./).length
```
bcaudan, line 89: I would be in favor of keeping only one of these fallbacks since both should work on our current browser target.
On IE11 Blob is faster, so I would keep only this one.

packages/core/src/transport.ts
```diff
@@ -1,6 +1,9 @@
 import { monitor } from './internalMonitoring'
 import { Context, deepMerge, DOM_EVENT, jsonStringify, noop, objectValues } from './utils'
 
+// https://en.wikipedia.org/wiki/UTF-8
+const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/g
```
bcaudan, line 5:
```suggestion
const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/
```
since we don't use match anymore

PR #416, opened 2020-06-03
✨ adding ability to append custom tags to logs/rum

evahteev:
## What it does

Appends custom tags from user config

## Changes

packages/core/src/configuration.ts

### DIFF
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -48,6 +48,7 @@ export interface UserConfiguration {
   trackInteractions?: boolean
   proxyHost?: string
 
+  tags?: string
   service?: string
   env?: string
   version?: string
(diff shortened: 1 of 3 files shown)


bcaudan:
Hi @evahteev,

Thanks for submitting this PR!

However, there is an ongoing reflexion to merge the tags and attributes concepts, the specific tag behavior in the application would probably be limited for a fixed set of attributes in the future.

Considering that, we would prefer to promote the use of the `setGlobalContext` API and not introduce a new API that we should deprecate in a few months.

Do you have some specific use cases that would require the use of tags instead of global context attributes?

bcaudan:
I am closing this one then, feel free to keep commenting on it or open a new issue.

PR #424, opened 2020-06-16
[RUM] Customize error collection

michaelmcchen:
## Motivation
Hello! As my organization begins to explore adopting Datadog RUM, we're realizing that we'd like to be able to better configure some of the data that is being sent to Datadog. In this PR, I expose `collectErrorMessage`, `collectErrorStack`, && `isCollectingError` as options when configuring the client. In our experience, PII can sometimes be surfaced in error messages and error stack traces, so we'd like to avoid collecting that. If other organizations want to eliminate error collection completely, they have the option of configuring `isCollectingError` to be `false`.

I'm also curious if there are any other recommended approaches for a similar result?  

## Changes

- Expose `collectErrorMessage`, `collectErrorStack`, `isCollectingError` as optional arguments to fine tune the error collection. 
- This change is backwards compatible. 

## Testing

`isCollectingError` is covered by tests in the core lib since it's already an exposed arg in the log lib. I'm happy to add a few more tests as well, but want to focus on soliciting feedback re: if the idea itself is something that's worth entertaining and can be merged. 

---

### DIFF
--- a/packages/rum/README.md
+++ b/packages/rum/README.md
@@ -44,6 +44,9 @@ datadogRum.init({
   - `service`: name of the corresponding service
   - `env`: environment of the service
   - `version`: version of the service
+  - `isCollectingError`: if RUM should collect errors. Default is `true`
+  - `collectErrorMessage`: when collecting errors, if RUM should collect error message. Default is `true`. Depends on `isCollectingError`.
+  - `collectErrorStack`: when collecting errors, if RUM should collect error stack trace. Default is `true`. Depends on `isCollectingError`.
 
   ```
   init(configuration: {
@@ -57,6 +60,9 @@ datadogRum.init({
       service?: string,
       env?: string,
       version?: string,
+      isCollectingError?: boolean,
+      collectErrorMessage?: boolean,
+      collectErrorStack?: boolean
   })
   ```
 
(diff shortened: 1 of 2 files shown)


michaelmcchen:
I'll take a moment to review the CLA and sign but in the meantime, I'd love to hear your initial thoughts on this idea!

bcaudan:
Hello @michaelmcchen,

Thanks for raising that!
We are planning on working on PII filtering on the SDK side but our strategy will probably to mask them wherever they can appear (error, url, click target content, ...) instead of disabling a feature or a part of a feature to avoid them.
So we probably won't want to go with the suggested approach.

However, it could be interesting to us to understand what kind of PII you observed on your errors. 
It would allow us to consider them when we will work on this PII filtering topic.

michaelmcchen:
Hi @bcaudan 

Thanks for the response! That's awesome - do y'all have more info on the expected timeline on releasing this? That'll be very helpful in my planning. 

Hmm, I don't have any specific examples for error logs at the moment - it was mainly a general precaution we wanted to take. The other source of observed PII is in view events via the browser URL (ie. IDs being passed as part of a path or query parameter).

bcaudan:
Thanks for the extra info!
We are planning to make progress on this topic before the end of the year.

You can be notified of any progress on that front by either:
- subscribing to https://github.com/DataDog/browser-sdk/issues/426
- send an email to support@datadoghq.com asking for this PII masking feature

I am closing this PR then.

PR #427, opened 2020-06-19
👷[RUMF-528] allow dual shipping in staging

bcaudan:
## Motivation

Allow to analyze staging events in prod environment

## Changes

**Constraint**
For RUM, we need to have different application id when sending events.

**Strategy**
For each event source (rum, logs, internal monitoring), instantiate a primary and a replica batch. Each event is added to both.

**Extra configuration**
```
{
  replica?: {
    applicationId?: string
    clientToken: string
  }
}
```

## Testing

Manual test

---

### DIFF
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -53,7 +53,7 @@ export interface UserConfiguration {
   version?: string
 
   // only on staging build mode
-  slave?: SlaveUserConfiguration
+  replica?: ReplicaUserConfiguration
 
   // only on e2e-test build mode
   internalMonitoringEndpoint?: string
@@ -61,7 +61,7 @@ export interface UserConfiguration {
   rumEndpoint?: string
 }
 
-interface SlaveUserConfiguration {
+interface ReplicaUserConfiguration {
   applicationId?: string
   clientToken: string
 }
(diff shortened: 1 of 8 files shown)


packages/core/src/configuration.ts
```diff
@@ -144,6 +162,27 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
     }
   }
 
+  if (transportConfiguration.buildMode === 'staging') {
```
mquentin, line 165: You shouldn't we have an enum out of `export type BuildMode = 'release' | 'staging' | 'e2e-test'` instead of having twice the same 'staging' ?

bcaudan:
> I think we should add unit tests to validate that both masterBatch slaveBatch states are equivalent after pushing a message.

It is not a prod feature, so I don't think it worth the trouble.

packages/core/src/configuration.ts
bcaudan: it is not an enum but let's use some for this kind of type, it would also avoid the casting

packages/core/src/internalMonitoring.ts
```diff
@@ -34,8 +34,35 @@ let externalContextProvider: () => utils.Context
...
+
+function startMonitoringBatch(configuration: Configuration) {
+  const masterBatch = createMonitoringBatch(configuration.internalMonitoringEndpoint!)
+  let slaveBatch: Batch<MonitoringMessage>
+  if (configuration.slave !== undefined) {
+    slaveBatch = createMonitoringBatch(configuration.slave.internalMonitoringEndpoint)
+  } else {
+    slaveBatch = ({
+      add: utils.noop,
+    } as unknown) as Batch<MonitoringMessage>
```
BenoitZugmeyer, line 60: NIT: to avoid this cast, maybe we could just type `slaveBatch` as `Batch<MonitoringMessage> | undefined`

PR #438, opened 2020-07-02
✨[RUMF-603] Introduce and use new lifecycle events

bcaudan:
## Motivation

Prepare future work on associate events with its parents based on start time
Fix discard user action on view created instead of view update

## Changes

Introduce new events:
- VIEW_CREATED: when a view is created
- VIEW_UPDATED: when a view is updated
- ACTION_CREATED: when an action is created
- ACTION_COMPLETED: when an action is completed
- ACTION_DISCARDED: when an action is discarded

## Testing

automated tests

---

### DIFF
--- a/packages/rum/src/userActionCollection.ts
+++ b/packages/rum/src/userActionCollection.ts
@@ -53,7 +53,7 @@ export function startUserActionCollection(lifeCycle: LifeCycle) {
   }
 
   // New views trigger the cancellation of the current pending User Action
-  lifeCycle.subscribe(LifeCycleEventType.VIEW_COLLECTED, () => {
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, () => {
     if (pendingAutoUserAction) {
       pendingAutoUserAction.stop()
     }
@@ -78,11 +78,13 @@ function newUserAction(lifeCycle: LifeCycle, type: UserActionType, name: string)
   const id = generateUUID()
   const startTime = performance.now()
 
+  lifeCycle.notify(LifeCycleEventType.ACTION_CREATED)
+
   const { eventCounts, stop: stopEventCountsTracking } = trackEventCounts(lifeCycle)
 
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
     if (hadActivity) {
-      lifeCycle.notify(LifeCycleEventType.USER_ACTION_COLLECTED, {
+      lifeCycle.notify(LifeCycleEventType.ACTION_COMPLETED, {
         id,
         name,
         startTime,
@@ -94,6 +96,8 @@ function newUserAction(lifeCycle: LifeCycle, type: UserActionType, name: string)
           resourceCount: eventCounts.resourceCount,
         },
       })
+    } else {
+      lifeCycle.notify(LifeCycleEventType.ACTION_DISCARDED)
     }
 
     stopEventCountsTracking()
--- a/packages/rum/src/userActionCollection.ts
+++ b/packages/rum/src/userActionCollection.ts
@@ -35,6 +35,8 @@ export type UserAction = CustomUserAction | AutoUserAction
 interface PendingAutoUserAction {
   id: string
   startTime: number
+  complete(endTime: number): void
+  discard(): void
   stop(): void
 }
 let pendingAutoUserAction: PendingAutoUserAction | undefined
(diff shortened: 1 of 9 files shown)


packages/rum/src/userActionCollection.ts
```diff
@@ -71,18 +73,33 @@ export function startUserActionCollection(lifeCycle: LifeCycle) {
...
   const startTime = performance.now()
 
+  lifeCycle.notify(LifeCycleEventType.ACTION_CREATED)
+
   const { eventCounts, stop: stopEventCountsTracking } = trackEventCounts(lifeCycle)
 
   const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, (hadActivity, endTime) => {
+    if (!pendingAutoUserAction) {
+      return
+    }
```
bcaudan, line 90: not super happy with that, but I have some ideas that would be easier to implement once getUserActionReference will be replaced

packages/rum/test/viewCollection.spec.ts
```diff
@@ -81,47 +81,62 @@ function spyOnViews() {
...
-    history.pushState({}, '', '/bar')
+  it('should create new view on path change', () => {
+    const { lifeCycle } = setupBuilder.build()
 
-    expect(viewContext.id).not.toEqual(initialView)
-    expect(viewContext.location).not.toEqual(initialLocation)
+    lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (viewContext) => {
+      expect(viewContext.id).not.toEqual(initialViewId)
+      expect(viewContext.location).not.toEqual(initialLocation)
+    })
```
BenoitZugmeyer, line 111: This is a bit risky because if the event isn't triggered synchronously during the test, no assertion will be done. Maybe you can use a `done` callback to the test even if itsn't async?
```suggestion
    lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (viewContext) => {
      expect(viewContext.id).not.toEqual(initialViewId)
      expect(viewContext.location).not.toEqual(initialLocation)
      done()
    })
```

packages/rum/src/userActionCollection.ts
BenoitZugmeyer: I would be even more careful by making sure the `pendingAutoUserAction` is the same as the one created earlier by `newUserAction`:

When initializing pendingAutoUserAction, do something like this:
```js
const autoUserAction = {...}
pendingAutoUserAction = autoUserAction
```

and then you can replace the condition to:

```suggestion
    if (pendingAutoUserAction === autoUserAction) {
      return
    }
```

bcaudan: yes, I have experimented with something like that but I would prefer to rework the design to remove the need for the global variable. It should be easier after coming PRs so I preferred to keep it simple for now.
Do you see new risks introduce by these changes?

BenoitZugmeyer: It should be fine!

packages/rum/test/viewCollection.spec.ts
bcaudan: > Maybe you can use a done callback to the test even if itsn't async?

since the events are triggered synchronously, what is the reasoning for doing that?

BenoitZugmeyer: Because you are the author, you know this event will be triggered synchronously when doing a `pushState`. But this is not obvious.
* there is no test to show this. We could write a new test case to ensure this (with a spy for example)
* even then, it would help the reader to understand when the test case is ending

bcaudan: From my understanding, your issue is that the test is written like an asynchronous test but it is not asynchronous.
IMO, this approach is similar to something like:
```js
it('...', () => {
  const callback = (a) => expect(a).toBe(true)
  testSubject(callback)
})
```
I don't think that we should write tests to ensure that this kind of code is synchronous.

I agree that using a spy would make the test easier to read and would prevent expectations not being run if there is any issue.

packages/rum/test/viewCollection.spec.ts
```diff
@@ -132,19 +149,27 @@ describe('rum track renew session', () => {
...
 
-  it('should update page view id on renew session', () => {
+  it('should create new view on renew session', () => {
     const { lifeCycle } = setupBuilder.build()
-    const initialView = viewContext.id
-    lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
 
-    expect(viewContext.id).not.toEqual(initialView)
+    lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (viewContext) => {
+      expect(viewContext.id).not.toEqual(initialViewId)
```
BenoitZugmeyer, line 169: This could be a spy too

bcaudan: thx for the catch!

PR #440, opened 2020-07-06
♻️ [RUMF-604] introduce parentContexts to return current contexts

bcaudan:
## Motivation

Following #438, prepare future work on associate events with its parents based on start time.

## Changes

- introduce parentContexts
- use parentContexts
- handle view url & session in parentContexts
- return structured context
- prepare usage with startTime
- rework userActionCollection

## Testing

automated tests

---

### DIFF
--- a/packages/rum/src/rum.entry.ts
+++ b/packages/rum/src/rum.entry.ts
@@ -20,7 +20,7 @@ import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { startPerformanceCollection } from './performanceCollection'
 import { startRum } from './rum'
 import { startRumSession } from './rumSession'
-import { startUserActionCollection, UserActionReference } from './userActionCollection'
+import { startUserActionCollection } from './userActionCollection'
 import { startViewCollection } from './viewCollection'
 
 export interface RumUserConfiguration extends UserConfiguration {
@@ -33,7 +33,9 @@ export interface InternalContext {
   view: {
     id: string
   }
-  user_action?: UserActionReference
+  user_action?: {
+    id: string
+  }
 }
 
 const STUBBED_RUM = {
--- a/packages/rum/src/rum.entry.ts
+++ b/packages/rum/src/rum.entry.ts
@@ -73,9 +73,16 @@ datadogRum.init = monitor((userConfiguration: RumUserConfiguration) => {
 
   const { errorObservable, configuration, internalMonitoring } = commonInit(rumUserConfiguration, buildEnv)
   const session = startRumSession(configuration, lifeCycle)
-  const globalApi = startRum(rumUserConfiguration.applicationId, lifeCycle, configuration, session, internalMonitoring)
+  const globalApi = startRum(
+    rumUserConfiguration.applicationId,
+    location,
+    lifeCycle,
+    configuration,
+    session,
+    internalMonitoring
+  )
 
-  startViewCollection(location, lifeCycle, session)
+  startViewCollection(location, lifeCycle)
   const [requestStartObservable, requestCompleteObservable] = startRequestCollection()
   startPerformanceCollection(lifeCycle, session)
   startDOMMutationCollection(lifeCycle)
--- a/packages/rum/src/rum.entry.ts
+++ b/packages/rum/src/rum.entry.ts
@@ -21,7 +21,6 @@ import { startPerformanceCollection } from './performanceCollection'
 import { startRum } from './rum'
 import { startRumSession } from './rumSession'
 import { startUserActionCollection } from './userActionCollection'
-import { startViewCollection } from './viewCollection'
 
 export interface RumUserConfiguration extends UserConfiguration {
   applicationId: string
@@ -83,7 +82,6 @@ datadogRum.init = monitor((userConfiguration: RumUserConfiguration) => {
     internalMonitoring
   )
 
-  startViewCollection(location, lifeCycle)
   const [requestStartObservable, requestCompleteObservable] = startRequestCollection()
   startPerformanceCollection(lifeCycle, session)
   startDOMMutationCollection(lifeCycle)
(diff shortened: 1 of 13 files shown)


packages/rum/src/rum.entry.ts
```diff
@@ -71,9 +74,16 @@ datadogRum.init = monitor((userConfiguration: RumUserConfiguration) => {
...
+    rumUserConfiguration.applicationId,
+    location,
+    lifeCycle,
+    configuration,
+    session,
+    internalMonitoring
+  )
 
-  startViewCollection(location, lifeCycle, session)
+  startViewCollection(location, lifeCycle)
```
BenoitZugmeyer, line 86: 👏 🎉

packages/rum/src/rum.ts
```diff
@@ -204,15 +199,12 @@ export function startRum(
...
-          session_id: viewContext.sessionId,
-          user_action: getUserActionReference(),
-          view: {
-            id: viewContext.id,
-          },
-        }
+      (startTime?: number): InternalContext => {
+        return (withSnakeCaseKeys(deepMerge(
+          { applicationId },
+          parentContexts.findView(startTime),
```
BenoitZugmeyer, line 205: `findView` has type `ViewContext | undefined`, but it is required in the `InternalContext` type. Maybe change the `InternalContext view` prop to optional.

packages/rum/src/rum.ts
```diff
@@ -304,27 +319,27 @@ function trackView(lifeCycle: LifeCycle, upsertRumEvent: (event: RumViewEvent, k
...
+function trackErrors(lifeCycle: LifeCycle, addRumEvent: (startTime: number, event: RumErrorEvent) => void) {
   lifeCycle.subscribe(LifeCycleEventType.ERROR_COLLECTED, ({ message, startTime, context }: ErrorMessage) => {
-    addRumEvent({
+    addRumEvent(startTime, {
       message,
       date: getTimestamp(startTime),
       evt: {
         category: RumEventCategory.ERROR,
       },
-      userAction: getUserActionReference(startTime),
```
BenoitZugmeyer, line 315: I'm not a big fan of the implicit userAction (and view) references merge. I think the previous solution was better, and should have been generalized for the view reference. I think having something like
```ts
userAction: getUserActionReference(startTime),
view: getViewReference(startTime),
```
is easier to understand than to follow the various merges accross the library. Also, it is more strictly typed, and any change we do in the RumEvent types will be verified staticly.

bcaudan: I still think that we should decouple the assembly of attributes specific to the type from the assembly of context attributes, from parent or others.
Extra static typing could indeed be nice too.
I agree that multiple merge places make things a bit confusing, I had started something on that but this PR started to be quite big.
Let me see if I can find a satisfying approach.

bcaudan: I think we could go with an approach that:
- separate event assembly, add view or action to the event
- extract a bunch of non related logic from the rum batch management

I have started something looking like:
```ts
type RawRumEvent = RumErrorEvent | RumResourceEvent...
export type RumEvent =
  | RumErrorEvent & ActionContext & ViewContext
  | RumResourceEvent & ActionContext & ViewContext
  | ...

function makeHandler(parentContexts: ParentContexts, session: RumSession, globalContextProvider: () => Context) {
  return function handler<T extends RawRumEvent>(
    assemble: (event: T, { view, action }: { view: ViewContext; action?: ActionContext }) => RumEvent,
    callback: (event: Context) => void
  ) {
    return (startTime: number, event: T, customerContext?: Context) => {
      const view = parentContexts.findView(startTime)
      if (session.isTracked() && view && view.sessionId) {
        const action = parentContexts.findAction(startTime)
        const rumEvent = assemble(event, { action, view })
        const message = deepMerge(
          customerContext,
          withSnakeCaseKeys(rumEvent as Context),
          globalContextProvider()
        ) as Context
        callback(message)
      }
    }
  }
}

const handler = makeHandler(parentContexts, session, () => globalContext)

trackView(lifeCycle, handler((event: RumViewEvent) => event, batch.upsert))
trackErrors(lifeCycle, handler((event: RumErrorEvent, { view, action }) => combine(event, action, view), batch.add))

// deepMerge with types union
export function combine<A, B, C>(a: A, b: B, c: C): A & B & C;
```

wdyt?

BenoitZugmeyer: This looks better indeed! It will be even simpler once the context is in its own field.

I fear that even with improved types, the usage of `combine` won't allow TS to validate types strictly. I played a bit with TS types, and maybe we could express events like that: 

https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgGrAgdwMIHtwQAeYyA3gFDJXIBuGmAXGZda8ACZMDOYUoA5gBoWrKgFcoAG268Bw1gF9yS8qEixEKAMoQuXYPjwFizVl1378ASU7IefEP2XlVBDUmQBBBGAMgjkCYUrGLmUN6++EzBolQcMg5Ois7kYACeAA4oAApwUBDgAUQkALxo9EUmAGTIOnp+lSQ1EQ34gWAuatDwHgBKcJjoWACiNAUkMVR0WNEiovF2so7ysciSuHDsAgD6vgC2EEwgYnsARtArVEoqXe4o-ZgAMvj8ACpwXADWo+Om1OwSOCREBHE7nKApcgAeihyFeAAsUENMD9wMgQBAIOwuMh8JI0sg4HYLH5kAg2sVUpkkfRUaVkA9kXTkDU6pZ-BTiNDYQiUM9HO8vsyMVicUTzPV8IJCbR6ISQOxcRlgXB8fLCT4-FSssh+W8Pt8xmiyg89YLDb9WSTDJymuUsI0WchclBfKqADwtG3GMAAPhc5BgYhAmvwZNwZ1AEGRAAoIEawExGbSE9LSBL2TYFEwXeNGgBKJhMhN-Kj5MASEBkZAAOjr8fG0ozfhsyGuLiDIeB4cjGLNBrjCaTA37QtTZGb1nY0tC0C9IGl00w2edeTztsLupe5uZk2Q5cr1brNYb4Cb1pANkX9BnYXnbeUQA

Various flavors of `Context` types could be declared into the `parentsContext` module, and it could have a single `findContext` returning a `ParentContext` type instead of separate ones.

Anyway, this looks like a good improvement in any case.

PR #443, opened 2020-07-09
🐛 fix dual shipping

bcaudan:
## Motivation

Following #440, dual shipping was broken due to wrong application id being associated with replica events.
This was due to an incorrect merge order.

## Changes

Remove context provider from Batch and let the caller process him self the message before passing it to the batch.

## Testing

automated tests

---

### DIFF
--- a/packages/core/src/internalMonitoring.ts
+++ b/packages/core/src/internalMonitoring.ts
@@ -24,7 +24,7 @@ export interface MonitoringMessage extends utils.Context {
 }
 
 const monitoringConfiguration: {
-  batch?: Batch<MonitoringMessage>
+  batch?: Batch
   debugMode?: boolean
   maxMessagesPerPage: number
   sentMessageCount: number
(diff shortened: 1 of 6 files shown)


bcaudan:
> This change looks fine, but I wonder why we need this processor taking place into `Batch` instead of processing the message before `add` or `upsert`? I think it would be a bit simpler.

Indeed, I kept a similar strategy than the current one, I'll lokk into it

packages/core/src/internalMonitoring.ts
```diff
@@ -14,7 +14,7 @@ export interface InternalMonitoring {
   setExternalContextProvider: (provider: () => utils.Context) => void
 }
 
-export interface MonitoringMessage {
+export interface MonitoringMessage extends utils.Context {
```
glorieux, line 17: At some point we might want to extract Context from `utils`. I doesn't make for a descriptive name.

bcaudan: I am not sure to get your point, how extracting Context from utils would make the name more descriptive?

PR #444, opened 2020-07-09
🐛 fix current action context reset on custom action

bcaudan:
## Motivation

ACTION_COMPLETED is triggered by both auto and custom user actions.
Action context can then be reset by custom user action happening during an auto user action

## Changes

Split custom and auto life cycles which are pretty different.
Remove previous checks on action type on ACTION_COMPLETED subscription
View event counts track both completion events to update the user action count

## Testing

automated tests

---

### DIFF
--- a/packages/rum/src/userActionCollection.ts
+++ b/packages/rum/src/userActionCollection.ts
@@ -9,20 +9,22 @@ export enum UserActionType {
   CUSTOM = 'custom',
 }
 
+type AutoUserActionType = UserActionType.CLICK
+
 export interface UserActionMeasures {
   errorCount: number
   longTaskCount: number
   resourceCount: number
 }
 
-interface CustomUserAction {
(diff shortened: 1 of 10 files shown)


packages/rum/src/userActionCollection.ts
```diff
@@ -9,20 +9,22 @@ export enum UserActionType {
   CUSTOM = 'custom',
 }
 
+type AutoUserActionType = UserActionType.CLICK
```
mquentin, line 12: LGTM for me ! Just one thing : Considering Benoit's work to rename `UserAction` to `Action` in web-ui, shouldn't we do the same here and rename `AutoUserActionType` => `AutoActionType`... ?

bcaudan: Let's wait for v2 migration inside the SDK to do this cleaning

PR #445, opened 2020-07-10
✨ [RUMF-605] associate event to parent context by start date (behind flag)

bcaudan:
## Motivation

Events were attached to parent view or action when we received them. It caused some issues in the waterfall display where we could see events starting before the start of the view (or the action) which were very confusing.

## Changes

Keep an history of parent contexts to retrieve the one (if any) corresponding to the start time of the event.
Clear automatically old context on session renew or after a time out delay.

## Testing

automated tests

---

### DIFF
--- a/packages/rum/src/parentContexts.ts
+++ b/packages/rum/src/parentContexts.ts
@@ -1,6 +1,7 @@
 import { Context } from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { RumSession } from './rumSession'
+import { AutoUserAction } from './userActionCollection'
 
 export interface ViewContext extends Context {
   sessionId: string | undefined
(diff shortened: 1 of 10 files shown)


packages/rum/src/parentContexts.ts
```diff
@@ -16,45 +21,132 @@ export interface ActionContext extends Context {
...
+    currentContext?: CurrentContext,
+    startTime?: number
+  ) {
+    if (!startTime) {
+      return currentContext ? buildContext() : undefined
+    }
+    if (currentContext && startTime >= currentContext.startTime) {
+      return buildContext()
+    }
+    if (!withContextHistory) {
```
mquentin, line 127: Maybe should you extract this `if (!withContextHistory)` outside the scope of `findContext` as this function should not be called if `withContextHistory=false`.

I would rather :

```
  findView: (startTime) => {
      return withContextHistory ? findContext(buildCurrentViewContext, previousViews, currentView, startTime) : undefined
    },
```

Also by doing that, you could remove `findContext` from the body of the `startParentContexts`. That would reduce `startParentContexts` size and clarify its inner code.

bcaudan: Keep in mind that this `if (!withContextHistory)` will be removed when we will clean the feature flag.
By moving this outside of findContext, we should need two different functions:

1. `findContextWithoutHistory`
2. `findContextWithHistory`

the first one should be called by the second one to avoid logic duplication, is that what you mean?

mquentin: Why would you need two functions ? Isn't possible to call `findContext` only if there is an history. Otherwise, in `findView` and `findAction`, instead of calling `findContext` you return undefined.

When you remove the FF you will just have to clean `findView` and `findAction`.

bcaudan: I don't get what you are saying. In `findContext`, there is some logic before `if (!withContextHistory)`, returning undefined when the flag is false would break the current behavior.

mquentin: Yes I mean that this code, before the `findContext` does not really find the context in the history but build one instead (using the function you pass as a parameter).

A solution would have been to handle these case in the findView / findAction without passing a buildContext param and either findContext if there is an history or build one (using buildCurrentActionContext/buildCurrentViewContext) if there is none.

bcaudan: oh ok, findContext is not related only to history, like findAction/View.
I found convenient to have the same kind of naming but I am open other naming suggestion if you found it confusing.
However, I found that having the whole strategy at a single place is pretty convenient since it is not that big. IMO, moving some parts of it outside of startParentContext does not seem to bring much value.

PR #461, opened 2020-07-21
[RUMF-617] integrate tracing

bcaudan:
## Motivation

Instrument requests and generate spans from browser-sdk in order to avoid the overhead of dd-trace-js.

## Changes

- [x] send tracing headers with requests
- [x] send span to trace intake with all context + tags
- [x] associate trace id to rum requests
- [x] configuration to enable/disable
- [x] configuration to allow/disallow origins

## Testing

Automated tests

---


packages/core/src/traceCollection.ts
```diff
@@ -0,0 +1,143 @@
...
+
+function buildTrace(requestCompleteEvent: RequestCompleteEvent, userConfiguration: UserConfiguration) {
+  const traceId = requestCompleteEvent.traceId!.toString(16)
+  const globalMeta: SpanMetadata = {}
+
+  if (userConfiguration.env) {
+    globalMeta.env = userConfiguration.env
+  }
+  if (userConfiguration.service) {
+    globalMeta.service = `${userConfiguration.service}-http-client` // TODO what if not defined?
```
bcaudan, line 99: @rochdev is service mandatory for tracing?

rochdev: Yes, I'm not 100% sure if the trace will be entirely dropped, but in any case most of APM won't work if the trace doesn't have a service.

rochdev: In `dd-trace` we fallback to `browser` if not defined.

packages/rum/src/traceCollection.ts
```diff
@@ -0,0 +1,145 @@
...
+      _sampling_priority_v1: 1,
+      _top_level: 1,
+      'http.status': context.status,
+    }
+
+    const span: Span = {
+      meta,
+      metrics,
+      duration: msToNs(context.duration),
+      error: 0, // TODO How to capture the error?
```
bcaudan, line 127: @rochdev about error, what is the goal?

bcaudan: I've tried something, let me know what you think

packages/rum/src/tracer.ts
```diff
@@ -0,0 +1,153 @@
...
+  return 'ddtrace' in window && (window as BrowserWindow).ddtrace.tracer.scope().active()
+}
+
+function getTraceIdFromDdTraceJs(): TraceIdentifier | undefined {
+  // tslint:disable-next-line: no-unsafe-any
+  return (window as BrowserWindow).ddtrace.tracer
+    .scope()
+    .active()
+    .context()._traceId // internal trace identifier
+}
```
glorieux, line 99: Do we have a way to differentiate between traces sent from `ddtrace` and RUM to track there usage? And then remove that code.

packages/rum/src/tracer.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/* tslint:disable:no-bitwise */
+export class TraceIdentifier {
```
glorieux, line 102: Is it possible to create a unit test to validate those Identifiers?

packages/rum/src/tracer.ts
```diff
@@ -0,0 +1,153 @@
...
+export function toHexString(traceId: TraceIdentifier) {
+  return traceId.toString(16)
+}
+
+/**
+ * Format used elsewhere
+ */
+export function toDecimalString(traceId: TraceIdentifier) {
+  return traceId.toString(10)
+}
```
glorieux, line 153: Is there any documentation on those formats we could link to?

packages/rum/src/tracer.ts
```diff
@@ -0,0 +1,105 @@
...
+    traceXhr: (context, xhr) =>
+      onTracingAllowed(context.url!, (tracingHeaders: TracingHeaders) => {
+        Object.keys(tracingHeaders).forEach((name) => {
+          xhr.setRequestHeader(name, tracingHeaders[name])
+        })
+      }),
+  }
+}
+
+function onTracingAllowed(url: string, inject: (tracingHeaders: TracingHeaders) => void): TraceIdentifier | undefined {
```
BenoitZugmeyer: The `on` prefix is usually used for event handlers. Maybe this function could be renamed to something more accurate... `injectHeadersIfTracingAllowed`.

packages/rum/src/requestCollection.ts
```diff
@@ -14,64 +15,75 @@ export interface RequestCompleteEvent {
...
+    context.traceId = tracer.traceXhr(context, xhr)
+    context.requestId = getNextRequestId()
+
     requestStartObservable.notify({
-      requestId,
+      requestId: context.requestId as number,
     })
   })
   xhrProxy.onRequestComplete((context) => {
+    const traceId = context.traceId as TraceIdentifier | undefined
```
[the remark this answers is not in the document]
BenoitZugmeyer: I agree, the 'context' typing is not ideal here. Maybe we could add a parametric type like `startXhrProxy<{ traceId: TraceIdentifier | undefined, requestId: number }>()` ?

packages/rum/README.md
```diff
@@ -57,6 +59,8 @@ datadogRum.init({
       service?: string,
       env?: string,
       version?: string,
+      enableTracing?: boolean,
+      tracingAllowedUrls?: RegExp[],
```
BenoitZugmeyer, line 63: I think having a callback function for this (ex: `shouldTraceUrl(url: string): boolean`) would be better than a regex array. It would let the user to implement any strategy they want.

packages/rum/src/requestCollection.ts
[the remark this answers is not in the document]
bcaudan: FMU, there are two different topics:

1. `TraceIdentifier | undefined` usages
2. Context weak typing that require to explicitly cast extra properties

For 1, working on 2 should only remove 2 of 7 `TraceIdentifier | undefined` usages.
However, what would be the value of extracting a dedicated type for it? Similarly, should we also extract a dedicated type for `string | undefined`?

For 2, yes that could be helpful to improve the typing.

packages/rum/src/traceCollection.ts
```diff
@@ -0,0 +1,137 @@
...
+    const span: Span = {
+      meta,
+      metrics,
+      duration: msToNs(request.duration),
+      error: request.status === 0 || request.error ? 1 : 0,
+      name: 'browser.request',
+      parent_id: '0000000000000000',
+      resource: request.method,
+      service: globalMeta.service,
+      span_id: traceId,
```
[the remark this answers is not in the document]
bcaudan: yes, you can have a look at current trace intake requests

packages/rum/src/tracer.ts
bcaudan: We don't have easy way to track usage across customers.

packages/rum/src/tracer.ts
bcaudan: what would you want to validate?

packages/rum/src/tracer.ts
bcaudan: Not that I am aware of.
cc @rochdev?

packages/rum/src/requestCollection.ts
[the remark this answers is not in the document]
bcaudan: 2. done in 240af2c

packages/rum/src/tracer.ts
glorieux: The format of the identifier.

bcaudan: something like a2525e6?

glorieux: Perfect!

packages/core/src/xhrProxy.ts
```diff
@@ -98,7 +98,7 @@ function proxyXhr() {
 
     this.addEventListener('loadend', monitor(reportXhr))
 
-    beforeSendCallbacks.forEach((callback) => callback(this._datadog_xhr))
+    beforeSendCallbacks.forEach((callback) => callback(this._datadog_xhr, this))
```
rochdev, line 101: Why not just pass `this` at that point?

packages/rum/README.md
```diff
@@ -44,6 +44,8 @@ datadogRum.init({
   - `service`: name of the corresponding service
   - `env`: environment of the service
   - `version`: version of the service
+  - `enableTracing`: enable tracing of xhr and fetch requests
+  - `shouldTraceUrl`: function to determine if an URL should be traced (default: same domain only)
```
rochdev, line 48: If this is set, is the same domain still traced?

packages/rum/src/tracer.ts
rochdev: Not that I'm aware of either since this is considered an implementation detail for the most part.

packages/core/src/configuration.ts
```diff
@@ -40,6 +41,8 @@ export interface UserConfiguration {
   applicationId?: string
   internalMonitoringApiKey?: string
   isCollectingError?: boolean
+  enableTracing?: boolean
+  shouldTraceUrl?: (url: string) => boolean
```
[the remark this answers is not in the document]
bcaudan: What pros/cons do you see?

packages/core/src/xhrProxy.ts
bcaudan: clients don't need to know that the context is attached to the xhr object through a `_datadog_xhr` property.
We could more easily change this implementation details if needed

packages/rum/README.md
bcaudan: not in the current implementation, it allows customers to exclude some same domain endpoints if they want to

packages/core/src/configuration.ts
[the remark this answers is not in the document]
rochdev: To be clear, I'm not saying to replace the function, but to also add a list. In that sense, there is no con (other than having to implement it and adding tests). Pros are that it's much more friendly for a user to provide a list of domains for example, compared to having to implement a logic that they will need to test.

packages/rum/README.md
rochdev: That could be confusing if this is used to _add_ domains and then the default is no longer traced.

packages/core/src/configuration.ts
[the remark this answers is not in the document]
bcaudan: For both cases, we will need to add configuration examples in the docs. If we provide a simple example that they can copy/paste, IMO it should be as friendly as the list.
I'd prefer to avoid to maintain multiple options for the same feature.

packages/rum/README.md
bcaudan: a good doc could address that

packages/core/src/configuration.ts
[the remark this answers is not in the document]
bcaudan: something like `extraDomainsToTrace` could be simpler to understand and should match the need of customers.
any preference?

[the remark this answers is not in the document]
rochdev: I would prefer the same configuration option as otherwise there would be a conflict between the 2. I'm mostly trying to make onboarding as easy as possible, and a list of domains would definitely be easier to configure. For most cases there won't be a need to configure the rest of the URL, and even less so to programmatically detect what to trace.

packages/rum/README.md
rochdev: Something simple with good defaults should not need to be documented.

packages/core/src/configuration.ts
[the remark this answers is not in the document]
bcaudan: Do you mean the same configuration as in dd-trace-js? `distributedTracingOriginWhitelist?: (string|RegExp)[]`
If yes, I'd be in favor of `allowedTracingOrigins`. Let's not use whitelist and I don't get the value of distributed.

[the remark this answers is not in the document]
bcaudan: done in 82c8978

PR #463, opened 2020-07-22
SDK: better support for SPA routing - STEP 1/2 : add hash tracking

mquentin:
## Motivation
**Given**
  startViewCollection
**When**
  on window.location.hash changes
**Then**
  check if the views are different. 
  If the hash fragment has changed, then the views are different.

## Changes
- ✨    add a `trackHash` function
- ✨    modify `areDifferentViews` to handle hash changes

## Testing

 ✅    hash changes tests have been added both for pushStates and window.location.hash window.location.hash
Test with the Vue router without `mode: 'history'`:

`const router = new VueRouter({
  routes
})`

### SHOULD NOT BE MERGE UNTIL STEP 2 is done
SDK: better support for SPA routing - STEP 2/2 : filter out html anchor tags changes

### DIFF
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -177,10 +177,11 @@ function trackHash(onHashChange: () => void) {
 function areDifferentViews(previous: Location, current: Location): boolean {
   return (
     previous.pathname !== current.pathname ||
-    // hash can be empty or undefined
-    // switching from undefined to empty does not mean different views
-    (previous.hash && current.hash && previous.hash !== current.hash) ||
-    Boolean(previous.hash && !current.hash) ||
+    // the views are different when:
+    //  switching from a defined hash to no hash ('')
+    //  switching from a defined hash to a different and defined hash
+    //  switching from no hash ('') to a defined hash
+    (previous.hash && previous.hash !== current.hash) ||
     Boolean(!previous.hash && current.hash)
   )
 }
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -58,7 +58,6 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
   // Renew view on hash changes
   trackHash(() => {
     if (areDifferentViews(currentLocation, location)) {
-      console.log('trackHash areDifferentViews yes')
       onRouteChange()
     }
   })
@@ -110,8 +109,6 @@ function newView(
   let loadingTime: number | undefined
 
   lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, { id, startTime })
-
-  console.log('LifeCycleEventType.VIEW_CREATED notified', id, startTime)
 
   // Update the view every time the measures are changing
   const { throttled: scheduleViewUpdate, stop: stopScheduleViewUpdate } = throttle(
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -41,26 +41,20 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
   const startOrigin = 0
   let currentView = newView(lifeCycle, currentLocation, ViewLoadingType.INITIAL_LOAD, startOrigin)
 
-  function onRouteChange() {
-    currentLocation = { ...location }
-    currentView.triggerUpdate()
-    currentView.end()
-    currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
+  function renewViewOnChange() {
+    if (areDifferentViews(currentLocation, location)) {
+      currentLocation = { ...location }
+      currentView.triggerUpdate()
+      currentView.end()
+      currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
+    }
   }
 
   // Renew view on history changes
-  trackHistory(() => {
-    if (areDifferentViews(currentLocation, location)) {
-      onRouteChange()
-    }
-  })
+  trackHistory(renewViewOnChange)
 
   // Renew view on hash changes
-  trackHash(() => {
-    if (areDifferentViews(currentLocation, location)) {
-      onRouteChange()
-    }
-  })
+  trackHash(renewViewOnChange)
 
   // Renew view on session renewal
   lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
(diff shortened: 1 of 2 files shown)


packages/rum/src/viewCollection.ts
```diff
@@ -41,13 +41,24 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
...
+      onRouteChange()
+    }
+  })
+
+  // Renew view on hash changes
+  trackHash(() => {
+    if (areDifferentViews(currentLocation, location)) {
+      onRouteChange()
     }
   })
```
bcaudan, line 63: What about get rid of the comments and extract all the common logic with something like:
```ts
trackHistory(renewViewOnChange)
trackHash(renewViewOnChange)

function renewViewOnChange() {
  if (areDifferentViews(currentLocation, location)) {
   currentLocation = { ...location }
   ...
  }
}
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -119,23 +126,62 @@ describe('rum track url change', () => {
     expect(viewContext.id).not.toEqual(initialViewId)
   })
 
-  it('should not create new view on search change', () => {
+  it('should create a new view on pushState hash change', () => {
```
bcaudan, line 129: one of your tests is flaky:
```
  An error was thrown in afterAll
  Error: Expected spy create not to have been called.
      at <Jasmine>
      at eval (webpack:///./packages/rum/test/viewCollection.spec.ts?:142:35)
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -119,23 +126,62 @@ describe('rum track url change', () => {
...
+  })
+
+  it('should not create a new view when the hash has kept the same value', () => {
+    window.location.hash = '#bar'
+
+    const { lifeCycle } = setupBuilder.build()
+    lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, createSpy)
+
+    window.addEventListener('hashchange', () => {
+      expect(createSpy).not.toHaveBeenCalled()
```
bcaudan, line 171: Changing this condition does not fail the test

packages/rum/test/viewCollection.spec.ts
```diff
@@ -119,23 +126,62 @@ describe('rum track url change', () => {
...
   })
+
+  it('should create a new view on hash change', () => {
+    const { lifeCycle } = setupBuilder.build()
+    lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, createSpy)
+
+    window.addEventListener('hashchange', () => {
+      expect(createSpy).toHaveBeenCalled()
+      const viewContext = createSpy.calls.argsFor(0)[0] as ViewContext
+      expect(viewContext.id).not.toEqual(initialViewId)
```
bcaudan, line 158: changing this condition does not fail the test

packages/rum/src/viewCollection.ts
```diff
@@ -165,8 +176,19 @@ function trackHistory(onHistoryChange: () => void) {
...
-function areDifferentViews(previous: Location, current: Location) {
-  return previous.pathname !== current.pathname
+function trackHash(onHashChange: () => void) {
+  window.addEventListener('hashchange', monitor(onHashChange))
+}
+
+function areDifferentViews(previous: Location, current: Location): boolean {
+  return (
+    previous.pathname !== current.pathname ||
+    // hash can be empty or undefined
```
bcaudan, line 186: How so?
https://developer.mozilla.org/en-US/docs/Web/API/Location/hash

bcaudan:
This PR should not be merged on master while anchor hash are not excluded

packages/rum/src/viewCollection.ts
```diff
@@ -41,15 +41,20 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
...
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
     }
-  })
+  }
+
+  // Renew view on history changes
+  trackHistory(renewViewOnChange)
+
+  // Renew view on hash changes
+  trackHash(renewViewOnChange)
```
bcaudan, line 57:
```suggestion
  trackHistory(renewViewOnChange)
  trackHash(renewViewOnChange)
```

packages/rum/src/viewCollection.ts
```diff
@@ -165,8 +170,20 @@ function trackHistory(onHistoryChange: () => void) {
...
+
+function areDifferentViews(previous: Location, current: Location): boolean {
+  return (
+    previous.pathname !== current.pathname ||
+    // the views are different when:
+    //  switching from a defined hash to no hash ('')
+    //  switching from a defined hash to a different and defined hash
+    //  switching from no hash ('') to a defined hash
+    (previous.hash && previous.hash !== current.hash) ||
+    Boolean(!previous.hash && current.hash)
```
bcaudan, line 185:
```suggestion
   previous.hash !== current.hash
```
if it is more than that, tests are missing

packages/rum/test/viewCollection.spec.ts
```diff
@@ -66,10 +66,25 @@ function mockHistory(location: Partial<Location>) {
...
     location.pathname = getPathName(url)
     location.search = getSearch(url)
-    location.hash = getHash(url)
+    location.hash = getHash(url) || ''
   })
 }
 
+function mockHash(location: Partial<Location>) {
+  // Reset previous tests value
+  window.location.hash = ''
```
bcaudan, line 75: what about moving that to the cleanup method?
it would avoid conflict with other/future tests

packages/rum/test/viewCollection.spec.ts
```diff
@@ -119,21 +137,68 @@ describe('rum track url change', () => {
     expect(viewContext.id).not.toEqual(initialViewId)
   })
 
-  it('should not create new view on search change', () => {
+  it('should create a new view on pushState hash change', () => {
```
bcaudan, line 140:
```suggestion
  it('should create a new view on hash change from history', () => {
```

packages/rum/src/viewCollection.ts
mquentin: ok we can simplify as if `previous` is not defined, and `current.hash= ''` (we don't have an hash change) but we have a pathname change and the previous `previous.pathname !== current.pathname` would trigger a new view anyway

bcaudan: How previous can not be defined?

PR #466, opened 2020-07-27
SDK: better support for SPA routing - STEP 2/2 : filter out html anchor tags changes

mquentin:
## Motivation
**Given**
  startViewCollection
**When**
  on window.location.hash changes
**Then**
  check if the views are different. 
  If the hash fragment has changed and that it is not an Anchor nav, then the views are different.

## Changes
- ✨  add a `noAnchorNavHashChange` check in the `trackHash` function that check if there is an equivalent id on one of the HTML elements

## Testing
 ✅    test added that check if the hashchange is related to an Anchor nav

### DIFF
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -171,7 +171,13 @@ function trackHistory(onHistoryChange: () => void) {
 }
 
 function trackHash(onHashChange: () => void) {
-  window.addEventListener('hashchange', monitor(onHashChange))
+  function noAnchorNavHashChange() {
+    const cleanedHash = window.location.hash.replace(/#/g, '')
+    if (!document.getElementById(cleanedHash)) {
+      onHashChange()
+    }
+  }
+  window.addEventListener('hashchange', monitor(noAnchorNavHashChange))
 }
 
 function areDifferentViews(previous: Location, current: Location): boolean {
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -177,10 +177,11 @@ function trackHash(onHashChange: () => void) {
 function areDifferentViews(previous: Location, current: Location): boolean {
   return (
     previous.pathname !== current.pathname ||
-    // hash can be empty or undefined
-    // switching from undefined to empty does not mean different views
-    (previous.hash && current.hash && previous.hash !== current.hash) ||
-    Boolean(previous.hash && !current.hash) ||
+    // the views are different when:
+    //  switching from a defined hash to no hash ('')
+    //  switching from a defined hash to a different and defined hash
+    //  switching from no hash ('') to a defined hash
+    (previous.hash && previous.hash !== current.hash) ||
     Boolean(!previous.hash && current.hash)
   )
 }
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -49,7 +49,7 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
     } else {
       // Anchor navigations would modify the location without generating a new view.
-      // These changes need to be acknowledged so they don't interfer with the next change areDifferentViews call
+      // These changes need to be acknowledged so they don't interfer with the next areDifferentViews call
       currentLocation = { ...location }
     }
   }
--- a/packages/rum/src/viewCollection.ts
+++ b/packages/rum/src/viewCollection.ts
@@ -47,6 +47,8 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
       currentView.triggerUpdate()
       currentView.end()
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
+    } else {
+      currentLocation = { ...location }
     }
   }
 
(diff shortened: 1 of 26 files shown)


packages/rum/src/viewCollection.ts
```diff
@@ -171,7 +171,13 @@ function trackHistory(onHistoryChange: () => void) {
 }
 
 function trackHash(onHashChange: () => void) {
-  window.addEventListener('hashchange', monitor(onHashChange))
+  function noAnchorNavHashChange() {
+    const cleanedHash = window.location.hash.replace(/#/g, '')
```
bcaudan, line 175:
```suggestion
    const cleanedHash = window.location.hash.substr(1)
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -193,6 +211,21 @@ describe('rum track url change', () => {
     window.location.hash = '#bar'
   })
 
+  it('should not create a new view when it is an Anchor navigation', (done) => {
```
bcaudan, line 214: It could be nice to have e2e tests to validate the behavior with and without anchor

packages/rum/test/viewCollection.spec.ts
```diff
@@ -85,6 +85,23 @@ function mockHash(location: Partial<Location>) {
   }
 }
 
+function mockHTMLElement() {
```
bcaudan, line 88:
```suggestion
function mockGetElementById() {
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -85,6 +85,23 @@ function mockHash(location: Partial<Location>) {
...
+      dummyElement.id = 'testHashValue'
+      const linkELement = document.createElement('a')
+      linkELement.href = '#testHashValue'
+      dummyElement.appendChild(linkELement)
+
+      return dummyElement
+    }
+
+    // tslint:disable-next-line: no-null-keyword
+    return null
```
bcaudan, line 101:
```suggestion
    return elementId === 'testHashValue'
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -108,6 +125,7 @@ describe('rum track url change', () => {
   beforeEach(() => {
     const fakeLocation: Partial<Location> = { pathname: '/foo', hash: '' }
     mockHistory(fakeLocation)
+    mockHTMLElement()
```
bcaudan, line 128: since it is only used by a single test, you can probably call it or inline it there

packages/rum/test/viewCollection.spec.ts
mquentin: the jasmine Mock expect a `HTMLElement | null` result so `return elementId === 'testHashValue'` would not work here

bcaudan: do we care?

mquentin: From my understanding `spyOn(document, 'getElementById').and.callFake` does

bcaudan: Since we don't need the element that you create, I would rather go with something like `(elementId === 'testHashValue' as unknown) as any`

mquentin: Ok I thought that the cost of mocking a real HTMLElement was worth it but I will implement your version

packages/rum/src/viewCollection.ts
```diff
@@ -174,8 +176,13 @@ function trackHash(onHashChange: () => void) {
   window.addEventListener('hashchange', monitor(onHashChange))
 }
 
+function isAnchorHashInPage(hash: string) {
+  const cleanedHash = hash.substr(1)
```
bcaudan, line 180:
```suggestion
  const correspondingId = hash.substr(1)
```

packages/rum/src/viewCollection.ts
```diff
@@ -47,6 +47,8 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
       currentView.triggerUpdate()
       currentView.end()
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
+    } else {
+      currentLocation = { ...location }
```
bcaudan, line 51: A comment to explain why and a unit test could be helpful

packages/rum/src/viewCollection.ts
```diff
@@ -174,8 +176,13 @@ function trackHash(onHashChange: () => void) {
   window.addEventListener('hashchange', monitor(onHashChange))
 }
 
+function isAnchorHashInPage(hash: string) {
```
bcaudan, line 179:
```suggestion
function isHashAnAnchor(hash: string) {
```

test/e2e/scenario/agents.scenario.ts
```diff
@@ -270,3 +270,36 @@ describe('user action collection', () => {
...
   })
 })
+
+describe('anchor navigation', () => {
+  it('should not create a new view when it is an Anchor navigation', async () => {
+    await $('#test-anchor')
+
+    await browserExecute(() => {
+      window.location.hash = '#test-anchor'
+    })
```
bcaudan, line 280:
```suggestion
    await $('#test-anchor').click()
```
what about a real behavior?

test/e2e/scenario/agents.scenario.ts
```diff
@@ -270,3 +270,36 @@ describe('user action collection', () => {
...
+    await flushEvents()
+    const rumEvents = await waitServerRumEvents()
+    const viewEvents = rumEvents.filter(isRumViewEvent)
+
+    expect(viewEvents.length).toBe(1)
+    expect(viewEvents[0].view.loading_type).toBe(ServerRumViewLoadingType.INITIAL_LOAD)
+  })
+
+  it('should create a new view on hash change', async () => {
+    await $('#test-anchor')
```
bcaudan, line 291: why do we need that?

packages/rum/src/viewCollection.ts
```diff
@@ -48,6 +48,8 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
       currentView.end()
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
     } else {
+      // Anchor navigations would modify the location without generating a new view.
+      // These changes need to be acknowledged so they don't interfer with the next change areDifferentViews call
```
bcaudan, line 52:
```suggestion
      // These changes need to be acknowledged so they don't interfer with the next areDifferentViews call
```

packages/rum/src/viewCollection.ts
```diff
@@ -47,6 +47,10 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
       currentView.triggerUpdate()
       currentView.end()
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
+    } else {
+      // Anchor navigations would modify the location without generating a new view.
+      // These changes need to be acknowledged so they don't interfer with the next change areDifferentViews call
+      currentLocation = { ...location }
```
bcaudan, line 53: What about enforcing that by a unit test?

packages/rum/src/viewCollection.ts
```diff
@@ -47,6 +47,10 @@ export function startViewCollection(location: Location, lifeCycle: LifeCycle) {
       currentView.triggerUpdate()
       currentView.end()
       currentView = newView(lifeCycle, currentLocation, ViewLoadingType.ROUTE_CHANGE)
+    } else {
+      // Anchor navigations would modify the location without generating a new view.
+      // These changes need to be acknowledged so they don't interfer with the next areDifferentViews call
```
bcaudan, line 52:
```suggestion
      // These changes need to be acknowledged so they don't interfere with the next areDifferentViews call
```

packages/rum/test/viewCollection.spec.ts
```diff
@@ -191,6 +197,45 @@ describe('rum track url change', () => {
...
+    }
+
+    window.addEventListener('hashchange', hashchangeCallBack)
+
+    window.location.hash = '#testHashValue'
+  })
+
+  it('should acknowledge the view location hash change after an Anchor navigation', (done) => {
+    const { lifeCycle } = setupBuilder.build()
+    const spyObj = mockGetElementById()
```
bcaudan, line 218: what about something more expressive than `spyObj`?

mquentin: as we just "clear" it, I did looked for a better name. Also the comment helps understanding the usage of `spyObj.and.callThrough()`

bcaudan: it could have been an opportunity to be more expressive

PR #468, opened 2020-07-29
[RUMF-617] Extract XHR and Fetch proxies

bcaudan:
## Motivation

Current proxies of XHR and fetch were managed by request collection which was handled as a singleton to avoid to proxy XHR and fetch multiple times.
With the future integration of tracing inside the sdk (cf #461), we would want to collect request with trace id for rum but not for logs. 
In a npm setup this would be problematic since the request collection instance would be shared between rum and logs.

## Changes

1.  Extract XHR and Fetch proxies that:
    - instrument browser API only once
    - deal with request data, not with domain logic (request id, trace id, request type) 
    - provide hooks to attach/retrieve info during the request life cycle
2. Use proxies in request collection and error collection
3. Move request collection to rum package

## Testing

Automated tests

---

### DIFF
--- a/packages/core/src/fetchProxy.ts
+++ b/packages/core/src/fetchProxy.ts
@@ -1,0 +1,95 @@
+import { toStackTraceString } from './errorCollection'
+import { monitor } from './internalMonitoring'
+import { computeStackTrace } from './tracekit'
+import { normalizeUrl } from './urlPolyfill'
+
+export interface FetchProxy {
+  beforeSend: (callback: (context: Partial<FetchContext>) => void) => void
+  onRequestComplete: (callback: (context: FetchContext) => void) => void
+  reset: () => void
+}
+
+export interface FetchContext {
+  method: string
+  startTime: number
+  duration: number
+  url: string
+  status: number
+  response: string
+  responseType?: string
+  [key: string]: unknown
+}
+
+let originalFetch: typeof window.fetch
+let hasBeenStarted = false
+const beforeSendCallbacks: Array<(xhr: Partial<FetchContext>) => void> = []
+const onRequestCompleteCallbacks: Array<(xhr: FetchContext) => void> = []
+
+export function startFetchProxy(): FetchProxy {
+  if (!hasBeenStarted) {
+    hasBeenStarted = true
+    proxyFetch()
+  }
+  return {
+    beforeSend(callback: (context: Partial<FetchContext>) => void) {
+      beforeSendCallbacks.push(callback)
+    },
+    onRequestComplete(callback: (context: FetchContext) => void) {
+      onRequestCompleteCallbacks.push(callback)
+    },
+    reset() {
+      hasBeenStarted = false
+      beforeSendCallbacks.splice(0, beforeSendCallbacks.length)
+      onRequestCompleteCallbacks.splice(0, onRequestCompleteCallbacks.length)
+      window.fetch = originalFetch
+    },
+  }
+}
+
+function proxyFetch() {
+  if (!window.fetch) {
+    return
+  }
+
+  originalFetch = window.fetch
+
+  // tslint:disable promise-function-async
+  window.fetch = monitor(function(this: WindowOrWorkerGlobalScope['fetch'], input: RequestInfo, init?: RequestInit) {
+    const method = (init && init.method) || (typeof input === 'object' && input.method) || 'GET'
+    const startTime = performance.now()
+
+    const context: Partial<FetchContext> = {
+      method,
+      startTime,
+    }
+
+    const reportFetch = async (response: Response | Error) => {
+      context.duration = performance.now() - context.startTime!
+      context.url = normalizeUrl((typeof input === 'object' && input.url) || (input as string))
+
+      if ('stack' in response || response instanceof Error) {
+        context.status = 0
+        context.response = toStackTraceString(computeStackTrace(response))
+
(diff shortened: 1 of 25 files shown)


packages/logs/test/logs.entry.spec.ts
```diff
@@ -11,6 +12,7 @@ describe('logs entry', () => {
 
   afterEach(() => {
     stopSessionManagement()
+    startXhrProxy().reset()
```
BenoitZugmeyer, line 15: This is a bit confusing. Maybe we could have a standalone function `stopXhrProxy()` or `resetXhrProxy()`? Also this is not obvious where it is started in this file, maybe add a comment? (same for rum.entry.spec.ts)

packages/core/test/requestCollection.spec.ts
```diff
@@ -1,350 +1,153 @@
...
+describe('collect fetch', () => {
   const FAKE_URL = 'http://fake-url/'
-  let originalFetch: any
-  let fetchStubBuilder: FetchStubBuilder
-  let fetchStub: (input: RequestInfo, init?: RequestInit) => FetchStubPromise
+  let fetchStub: FetchStub
+  let fetchStubManager: FetchStubManager
+  let fetchProxy: FetchProxy
+  let startSpy: jasmine.Spy
+  let completeSpy: jasmine.Spy
```
BenoitZugmeyer: this could be properly typed, so we wouldn't need as much casts
```suggestion
  let startSpy: jasmine.Spy<(requestStartEvent: RequestStartEvent) => void>
  let completeSpy: jasmine.Spy<(requestCompleteEvent: RequestCompleteEvent) => void>
```

packages/rum/src/rum.ts
```diff
@@ -205,29 +205,35 @@ export function startRum(
...
+            parentContexts.findAction(startTime)
+          ) as Context) as unknown) as InternalContext
+        }
+      ),
+      setRumGlobalContext: monitor((context: Context) => {
+        globalContext = context
+      }),
+    },
+    stop: () => {
+      batch.stop()
```
BenoitZugmeyer, line 229: I feel that we should rather stop collecting stuff (stop the view collection etc.) instead of telling the batch to ignore future events. It may be more work because we need to add stop functions for everything, but I think most of those already exist, and it'll make testing things separately simpler.

packages/core/src/fetchProxy.ts
```diff
@@ -0,0 +1,95 @@
...
+  responseType?: string
+  [key: string]: unknown
+}
+
+let originalFetch: typeof window.fetch
+let hasBeenStarted = false
+const beforeSendCallbacks: Array<(xhr: Partial<FetchContext>) => void> = []
+const onRequestCompleteCallbacks: Array<(xhr: FetchContext) => void> = []
+
+export function startFetchProxy(): FetchProxy {
```
BenoitZugmeyer, line 28: The API is a bit weird, because it returns a new instance, but the state is global... maybe expose standalone functions instead?

packages/rum/src/rum.ts
bcaudan: I agree but this PR is already quite large and a preliminary task for another large task.
I'll add a comment here

packages/core/src/fetchProxy.ts
bcaudan: Would you prefer something like:
```ts
let fetchProxySingleton: FetchProxy

export function startFetchProxy(): FetchProxy {
  if (!hasBeenStarted) {
    hasBeenStarted = true
    proxyFetch()
    fetchProxySingleton = {
      beforeSend...
    }
  }
  return fetchProxySingleton
}
```

packages/logs/test/logs.entry.spec.ts
bcaudan: yes, it could be nicer

packages/rum/src/rum.ts
BenoitZugmeyer: Seems reasonable :)

packages/core/src/fetchProxy.ts
BenoitZugmeyer: With this, you'll still need to do `startXhrProxy().reset()` which I find unintuitive.

What about something like:

```ts
export function startFetchProxy(): FetchProxy {
  if (!hasBeenStarted) {
    hasBeenStarted = true
    proxyFetch()
  }
}

export function beforeSend(callback: (context: Partial<FetchContext>) => void) {
      beforeSendCallbacks.push(callback)
}
exporn function onRequestComplete(callback: (context: FetchContext) => void) {
      onRequestCompleteCallbacks.push(callback)
 }
export function resetFetchProxy() {
      if (hasBeenStarted) {
            hasBeenStarted = false
            beforeSendCallbacks.splice(0, beforeSendCallbacks.length)
            onRequestCompleteCallbacks.splice(0, onRequestCompleteCallbacks.length)
            window.fetch = originalFetch
      }
}
```

bcaudan: I would prefer to enforce the call to start before adding callbacks, to avoid to rely on a proxy started elsewhere.

with my previous suggestion, we could have:

```ts
export function resetFetchProxy() {
  if (hasBeenStarted) {
    fetchProxySingleton.reset()
  }
}
```
wdyt?

bcaudan: cf 23b5be9

PR #469, opened 2020-07-30
[RUMF-625] make sure view url doesn't change

BenoitZugmeyer:
## Motivation

When a SPA changes its URL, the ending view gets updated with the new URL. It should keep its initial URL no matter what.

See https://datadoghq.atlassian.net/browse/RUMF-625

## Changes

When formating a RumViewEvent, use the one from the collected View object instead of the ViewContext which already contains the new URL

## Testing

---

### DIFF
--- a/packages/rum/src/parentContexts.ts
+++ b/packages/rum/src/parentContexts.ts
@@ -24,6 +24,7 @@ export interface ActionContext extends Context {
 interface CurrentContext {
   id: string
   startTime: number
+  location?: Location
 }
 
 interface PreviousContext<T> {
@@ -56,6 +57,10 @@ export function startParentContexts(location: Location, lifeCycle: LifeCycle, se
     }
     currentView = currentContext
     currentSessionId = session.getId()
+  })
+
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (currentContext) => {
+    currentView = currentContext
   })
 
   lifeCycle.subscribe(LifeCycleEventType.AUTO_ACTION_CREATED, (currentContext) => {
@@ -100,7 +105,13 @@ export function startParentContexts(location: Location, lifeCycle: LifeCycle, se
   }
 
   function buildCurrentViewContext() {
-    return { sessionId: currentSessionId, view: { id: currentView!.id, url: location.href } }
+    return {
+      sessionId: currentSessionId,
+      view: {
+        id: currentView!.id,
+        url: (currentView!.location || location).href,
+      },
+    }
   }
 
   function buildCurrentActionContext() {
--- a/packages/rum/src/parentContexts.ts
+++ b/packages/rum/src/parentContexts.ts
@@ -21,10 +21,15 @@ export interface ActionContext extends Context {
   }
 }
 
-interface CurrentContext {
+interface CurrentViewContext {
   id: string
   startTime: number
-  location?: Location
+  location: Location
+}
+
+interface CurrentActionContext {
+  id: string
+  startTime: number
 }
 
 interface PreviousContext<T> {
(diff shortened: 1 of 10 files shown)


packages/rum/src/rum.ts
```diff
@@ -359,6 +360,7 @@ function trackView(lifeCycle: LifeCycle, handler: (startTime: number, event: Rum
         loadingTime: view.loadingTime ? msToNs(view.loadingTime) : undefined,
         loadingType: view.loadingType,
         measures: view.measures,
+        url: view.location.href,
```
bcaudan, line 363: With this approach, url from ViewContext is still wrong so it can still be wrongly associated to other event types.
What about fixing it there?

BenoitZugmeyer: I wrote a new fix, let me know if it looks right :)

packages/rum/src/viewCollection.ts
```diff
@@ -39,35 +39,38 @@ export const SESSION_KEEP_ALIVE_INTERVAL = 5 * ONE_MINUTE
...
-  let currentView = newView(lifeCycle, currentLocation, ViewLoadingType.INITIAL_LOAD, startOrigin)
+  let currentView = newView(lifeCycle, ViewLoadingType.INITIAL_LOAD, startOrigin)
+  currentView.triggerUpdate(currentLocation)
 
   // Renew view on history changes
   trackHistory(() => {
     if (areDifferentViews(currentLocation, location)) {
-      currentLocation = { ...location }
-      currentView.triggerUpdate()
+      currentView.triggerUpdate(currentLocation)
```
bcaudan, line 48: Two things are bothering me:
- I find it a bit confusing to update location of the view from a view update
- It is a bit weird to maintain a current location on startViewCollection scope and a location on newView scope

What about:
- maintaining the view location only in the view scope, areDifferentViews could be moved in current view
- have a triggerUpdate and an updateLocation in order to separate the two concerns

wdyt?

BenoitZugmeyer: If `areDifferentViews` usage is done in the `newView` function, it would mean that the `newView` is responsible for ending itself, so we'd need another callback to create a new view after that. Does it seem ok to you?

packages/rum/src/parentContexts.ts
```diff
@@ -58,6 +59,10 @@ export function startParentContexts(location: Location, lifeCycle: LifeCycle, se
     currentSessionId = session.getId()
   })
 
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (currentContext) => {
+    currentView = currentContext
+  })
```
bcaudan, line 64: Instead of subscribing to view update, what about having a VIEW_COMPLETED event to update the location?

BenoitZugmeyer: The location changes during the view lifetime, not only when the view is completed.

packages/rum/src/viewCollection.ts
bcaudan: I was thinking about declaring `areDifferentViews` inside new view but using it outside, just for the view location access

BenoitZugmeyer: So you'd have something like:
```ts
trackHistory(() => {
  if (currentView.isDifferentView(newLocation)) {
    currentView.triggerUpdate()
    currentView.end()
    currentView = newView(...)
} else {
   currentView.updateLocation(newLocation)
}
```
right?

packages/rum/src/parentContexts.ts
bcaudan: yes, we could keep to look at the actual location while the view is active.
Subscribe to view update coule would avoid to have two sources of truth though.

packages/rum/src/viewCollection.ts
bcaudan: yes

packages/rum/src/parentContexts.ts
BenoitZugmeyer: I'd rather keep a single source of truth. If the VIEW_CREATED data was a View (containing the location), we could drop the 'location' usage altogether:
```diff
-url: (currentView!.location || location).href,
+url: currentView!.location.href,
```
WDYT?

bcaudan: we could just add the location with id and start time

BenoitZugmeyer: That would introduce a new data structure instead of just re-using the View one. But alright.

PR #476, opened 2020-08-07
✨ [RUMF-626] use site configuration and deprecate suffixed bundle

bcaudan:
## Motivation

Allow to easily support future datacenters and align configuration parameter with other libraries.

## Changes

- Support site configuration
- Build new bundle files without suffix: `datadog-rum.js` and `datadog-logs.js`
- Use new bundle files everywhere

## Testing

Automated tests

---

### DIFF
--- a/scripts/replace-build-env.js
+++ b/scripts/replace-build-env.js
@@ -18,9 +18,27 @@ try {
     from: Object.keys(buildEnv).map((entry) => `<<< ${entry} >>>`),
     to: Object.values(buildEnv),
   })
+  if (buildEnv.WITH_SUFFIX) {
+    replace.sync({
+      files: `${buildDirectory}/**/*.js`,
+      from: /.*/,
+      to: (...args) =>
+        `${makeDeprecatedSuffixComment(args[args.length - 1])}\n${args[0]}\n${makeDeprecatedSuffixComment(
+          args[args.length - 1]
+        )}`,
+    })
+  }
   console.log('Changed files:', results.filter((entry) => entry.hasChanged).map((entry) => entry.file))
   process.exit(0)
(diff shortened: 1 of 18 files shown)


scripts/replace-build-env.js
```diff
@@ -18,9 +18,29 @@ try {
...
   console.error('Error occurred:', error)
   process.exit(1)
 }
+
+function makeDeprecatedSuffixComment(filePath) {
+  const fileName = filePath.split('/').pop()
+  const suffixRegExp = /-(us|eu)/
+  const env = fileName.match(suffixRegExp)[1]
+  const newFileName = fileName.replace(suffixRegExp, '')
+  return `/**\n * ${fileName} IS DEPRECATED, USE ${newFileName} WITH { site: 'datadoghq.${
```
BenoitZugmeyer, line 43: I fear that a code comment is not enough. What about having a console.warn instead?

Also, https://webpack.js.org/plugins/banner-plugin/ could be used for this purpose.

bcaudan: IMO, adding extra warning is a bit too much, I would do that if we set a date for the removal of the API.

about banner-plugin, since it is not something that we want to maintain, I don't think it worth the effort to change the implementation, wdyt?

BenoitZugmeyer: Since we don't know how much time we'll need to maintain this, I'd be in favor of having a nicer implementation

bcaudan: I'd say that if we need to touch it, let's push to remove it instead

BenoitZugmeyer: Alright

Issue #409, opened 2020-05-29
Feature request: Context hierarchy


mboogerd:
**Context**
To start with, some context for this feature request. I would like to:

1. Incrementally add contextual information to my logging, as my code progresses. 
2. Remove contextual information when I leave the context for which that data is relevant. 
3. Have `console.log` statements include the context

**Provided solution**
The library doesn't make it impossible to do this, but it lacks a bit of convenience:

1. is currently possible by adding data either to the global logger or a local logger.
2. is currently possible only by overwriting the entire context. This method is inconvenient for removing just the bits that are no longer relevant, and is further constrained because I cannot query the current context. As such, the only real way to do this currently is set up our own context tracking mechanism next to the one datadog provides.
3. although we can fork local loggers from global loggers for a single extra layer of context, the `console.logs` made from this context wouldn't include that additional context.

**Proposed solution**
Include the following API and implementation:

1. `pushContext([key1: value1, ..., keyN: valueN}])` to push a bit of context into a global context array
2. `popContext()` to remove one layer of context from the global context stack.

For each log statement issued, assemble a context by merging context layers, overwriting older context values with newer ones on key conflicts.

For now I am doing the above myself with some React components in my app, but I think this would be a great addition for anybody using this logging package.

Thanks in advance for your consideration!

toefel18:
The idea is similar to MDC (Mapped Diagnostic Context) in Java applications and it would be a nice addition!

bcaudan:
Hi @mboogerd!


#### About context
 
> 1. Incrementally add contextual information to my logging, as my code progresses.

Would the `logger.addContext (key: string, value: any)` suit this need?

> 2. Remove contextual information when I leave the context for which that data is relevant.

We don't have a specific API to remove specific key/value of the context but you should achieve this behavior by setting a key to undefined.
```js
logger.addContext("foo", undefined)
```



Let me know if this approach could work for you or if I missed something on your use case.

 
#### About `console.logs`

> 3. Have console.log statements include the context

Yes, we don't automatically collect `console.logs` for now, there is an open issue to track this feature: https://github.com/DataDog/browser-sdk/issues/400

mboogerd:
Hi @bcaudan, thanks for your response!

[...]

PR #477, opened 2020-08-07
✨ Update context api to include removeContext method

PinPinIre:
## Motivation

In issue #409 it was suggested updating the logger API to allow users to remove keys from the logger. 

## Changes

This PR updates the logger API to have a removeContext method that removes a key from the context object. No existing methods have been modified as part of this work.

## Testing

To test this new functionality, the user must call removeContext with an existing key. A unit test has been added to test the existing addContext function and the new removeContext function. 

---


bcaudan:
Thanks for this PR @PinPinIre!
I've fixed the formatting and merged it in #478

PR #479, opened 2020-08-11
[RUMF-622] attach loading measures to initial view

BenoitZugmeyer:
## Motivation

Loading time measures should only be set on the initial view.

## Changes

This PR is mostly based on the RFC https://docs.google.com/document/d/1MNlX0pA_RFMJL5MHaDzRlHalA8I__LZM7rMCOaUHPpE/edit . Some additional adjustements have been done for the 'duration' and 'loadingTime' values.

Resulting change in web-ui:

Screenshot 2020-08-11 at 16 15 41

As we can see, when quickly navigating from "Explorer" to "Analytics", the "Load event end" measure is shown on the initial view ("Explorer"). The "Loading time" is adjusted to the "Load event end". The duration, though, is not adjusted and is smaller than "Load event end" (as expected).

## Testing

See unit tests. Controling the load event in a E2E environment might be tough, so no test has been added yet, but I could try if we feel there is a need for this.

We will be able to test it on staging: the query `@view.measures.load_event_end:* @view.loading_type:route_change` should not return any view event anymore.

---


packages/rum/src/viewCollection.ts
```diff
@@ -126,9 +126,11 @@ function newView(
     scheduleViewUpdate()
   })
 
-  const { stop: stopLoadingTimeTracking } = trackLoadingTime(lifeCycle, loadingType, (loadingTimeValue) => {
-    loadingTime = loadingTimeValue
-    scheduleViewUpdate()
+  const { stop: stopActivityLoadingTimeTracking } = trackActivityLoadingTime(lifeCycle, (newLoadingTime) => {
+    if (newLoadingTime !== undefined) {
+      updateLoadingTime(newLoadingTime)
```
bcaudan, line 131: So for initial view, we could send a wrong loading time from activity and then correct it with the load event.
I would prefer to avoid that if possible

packages/rum/src/viewCollection.ts
```diff
@@ -148,8 +150,17 @@ function newView(
...
+    if (loadingTime === undefined || newLoadingTime > loadingTime) {
+      loadingTime = newLoadingTime
+    }
+  }
+
   function updateMeasures(newMeasures: Partial<ViewMeasures>) {
     measures = { ...measures, ...newMeasures }
+    if (measures.loadEventEnd !== undefined) {
+      updateLoadingTime(nsToMs(measures.loadEventEnd))
+    }
```
bcaudan: Several things are bothering me with this strategy:

- `trackLoadingTime` was responsible to handle all the loading time logic and by looking at it we could easily understand the functional rule that the loading time should take event load into account on initial view. Now, it depends on the fact that measures are updated for the initial view which is not as clear.
- updateMeasures is then not only updating measures but also loading time
- having to check the loading time value before updating it and doing the `nsToMs` conversion let me believe that there is a design issue.

What about having a strategy allowing us to keep a design similar to the previous one?

- `trackLoadingTime` could expose `onTimingValue` instead of having `trackLoadEventLoadingTime` and it would be called when an update if the load event occurs

- splitting `updateMeasures` into `updateTimings` and `updateCounters` would allow to better follow what is relative to timings and what is relative to counters. It would also simplify the v2 migration.

- what about doing the `nsToMs` conversion later to avoid the opposite conversion?

packages/rum/test/viewCollection.spec.ts
```diff
@@ -606,4 +606,73 @@ describe('rum view measures', () => {
...
+      const { lifeCycle, clock } = setupBuilder.withFakeClock().build()
+      expect(getHandledCount()).toEqual(1)
+      expect(getViewEvent(0).measures).toEqual({
+        errorCount: 0,
+        longTaskCount: 0,
+        resourceCount: 0,
+        userActionCount: 0,
+      })
+
+      clock.tick(100)
```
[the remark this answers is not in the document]
BenoitZugmeyer: I set this to have a non-0 duration for the view. Any positive integer would do.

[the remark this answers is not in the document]
BenoitZugmeyer: I fixed a small issue in the performance.now() mock, and used constants for this.

packages/rum/test/viewCollection.spec.ts
```diff
@@ -606,4 +606,73 @@ describe('rum view measures', () => {
...
+        domContentLoaded: 345e6,
+        domInteractive: 234e6,
+        errorCount: 0,
+        firstContentfulPaint: 123e6,
+        loadEventEnd: 567e6,
+        longTaskCount: 0,
+        resourceCount: 0,
+        userActionCount: 0,
+      })
+    })
```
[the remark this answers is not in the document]
BenoitZugmeyer: Done

packages/rum/src/viewCollection.ts
BenoitZugmeyer: Done

packages/rum/src/viewCollection.ts
BenoitZugmeyer: I implemented a new version of `trackLoadingTime` which fixes the loadingTime computation (= only set it once). It does not embed the `trackActivityLoadingTime` call, because calling `stop` should be final, and `trackLoadingTime` should not be stoped at the view end since it is still useful after. LMKWYT

packages/rum/test/rum.spec.ts
```diff
@@ -665,4 +673,26 @@ describe('rum event assembly', () => {
...
     expect(getRumMessage(server, 1).view.url).toEqual('http://foo.com/bar')
   })
+
+  it('should keep the same URL when updating an ended view', () => {
+    const { server, lifeCycle, clock } = setupBuilder
+      .withFakeClock()
+      .withFakeLocation('http://foo.com/')
+      .build()
+
+    clock.tick(VIEW_DURATION)
```
bcaudan, line 683: Why do we need to set a view duration to make the test work?

packages/rum/src/viewCollection.ts
```diff
@@ -226,53 +230,35 @@ function trackTimings(lifeCycle: LifeCycle, callback: (timings: Timings) => void
...
-
-  function onTimingValue(timingValue: number | undefined) {
-    expectedTiming -= 1
-    if (timingValue) {
-      receivedTimings.push(timingValue)
-    }
-
-    if (expectedTiming === 0 && receivedTimings.length) {
-      callback(Math.max(...receivedTimings))
+  function maybeInvokeCallback() {
```
bcaudan, line 238:
```suggestion
  function invokeCallbackIfAllCandidatesAreReceived() {
```

packages/rum/test/rum.spec.ts
BenoitZugmeyer: If we don't advance the time, then the first view and the second view all starts at the same time. Since the `parentContext` is using the `startTime` to match an view event with its view context, the initial view update is incorrectly using the `currentContext` (second view context) https://github.com/DataDog/browser-sdk/blob/master/packages/rum/src/parentContexts.ts#L135

PR #492, opened 2020-08-14
[RUMF-636] initial document trace id

BenoitZugmeyer:
## Motivation

Retrieve the trace id generated by APM for the initial document load and associate it with the generated document resource event.

## Changes

* Adjust TS typings so we can set a traceId to the performance resource timing object
* Look for data set by the APM into the document, and parse it to extract a trace id and trace time
* Use this data if the trace time is recent enough to define the trace id on the initial document resource event

## Testing

---


mquentin (approved):
Looks good to me once your are done with the e2e tests

packages/rum/src/rum.ts
```diff
@@ -503,10 +504,14 @@ function trackPerformanceTiming(
...
   handler: (startTime: number, event: RumResourceEvent) => void,
-  lifeCycle: LifeCycle
+  entry: PerformanceResourceTiming
 ) {
+  if (!session.isTrackedWithResource()) {
+    return
+  }
   if (!isValidResource(entry.name, configuration)) {
     return
   }
```
bcaudan, line 509: why not a single if?

packages/rum/src/performanceCollection.ts
```diff
@@ -7,6 +7,52 @@ interface BrowserWindow extends Window {
   PerformanceObserver?: PerformanceObserver
 }
 
+export interface RumPerformanceResourceTiming {
```
bcaudan, line 11: What about extending existing interfaces?

packages/rum/src/performanceCollection.ts
```diff
@@ -130,18 +158,23 @@ function computeRelativePerformanceTiming() {
...
+    }
+  })
+}
 
-  entries
-    .filter((entry) => entry.entryType === 'navigation')
-    // Exclude incomplete navigation entries by filtering out those who have a loadEventEnd at 0
-    .filter((entry) => (entry as PerformanceNavigationTiming).loadEventEnd > 0)
-    .forEach(notify)
+function handleRumPerformanceEntry(lifeCycle: LifeCycle, entry: RumPerformanceEntry) {
```
bcaudan, line 185:
```suggestion
function handlePerformanceEntry(lifeCycle: LifeCycle, entry: RumPerformanceEntry) {
```
or rename parent to `handleRumPerformanceEntries` for consistency

packages/rum/src/resourceUtils.ts
```diff
@@ -57,7 +58,9 @@ function areInOrder(...numbers: number[]) {
   return true
 }
 
-export function computePerformanceResourceDuration(entry: PerformanceResourceTiming): number {
+export function computePerformanceResourceDuration(
+  entry: RumPerformanceResourceTiming | PerformanceResourceTiming
```
bcaudan, line 62: `RumPerformanceResourceTiming extends PerformanceResourceTiming` would avoid that

packages/core/src/utils.ts
```diff
@@ -364,3 +364,8 @@ export function getLinkElementOrigin(element: Location | HTMLAnchorElement | URL
   const sanitizedHost = element.host.replace(/(:80|:443)$/, '')
   return `${element.protocol}//${sanitizedHost}`
 }
+
+export function findCommaSeparatedValue(rawString: string, name: string) {
+  const matches = rawString.match(`(?:^|;)\\s*${name}\\s*=\\s*([^;]+)`)
+  return matches ? matches[1] : undefined
```
bcaudan, line 370: why changing the implementation? `matches[1]` VS `matches.pop()`

packages/rum/src/getAPMDocumentData.ts
```diff
@@ -0,0 +1,89 @@
+import { findCommaSeparatedValue } from '@datadog/browser-core'
+
+interface APMDocumentData {
```
bcaudan, line 3:
```suggestion
interface DocumentTraceData {
```
I would be in favor of replacing all APM references by trace for consistency

packages/rum/src/performanceCollection.ts
```diff
@@ -99,9 +99,20 @@ export function startPerformanceCollection(lifeCycle: LifeCycle) {
...
-function retrieveInitialDocumentResourceTiming(callback: (timing: RumPerformanceResourceTiming) => void) {
+export const INITIAL_DOCUMENT_OUTDATED_TRACE_ID_THRESHOLD = 2 * ONE_MINUTE
+
+export function retrieveInitialDocumentResourceTiming() {
   let timing: RumPerformanceResourceTiming
+
+  let traceId
+  const apmDocumentData = getAPMDocumentData(document)
+  const now = Date.now()
+  if (apmDocumentData && apmDocumentData.traceTime > now - INITIAL_DOCUMENT_OUTDATED_TRACE_ID_THRESHOLD) {
```
bcaudan, line 110: What about having this logic inside `getAPMDocumentData`?

packages/rum/src/rum.ts
BenoitZugmeyer: This is the same condition as `trackRequests`, and I find having two separate conditions for two different concerns easier to read, but I could:

* join them in a single condition
* factorize it with `trackRequests` into a separate `shouldTrackResource` function

WDYT?

packages/rum/src/performanceCollection.ts
BenoitZugmeyer: Having simpler interfaces describing what is actually used makes those data structure creation easier and type checked (no need to cast as much as before), for example the initial document and navigation timings creation are actually type-checked: https://github.com/DataDog/browser-sdk/blob/87c68d74f0c5be293a20b50bf9b6bb15ce2c7d73/packages/rum/src/performanceCollection.ts#L101-L143

packages/rum/src/performanceCollection.ts
BenoitZugmeyer: I'm not sure about this. The parent is accepting `PerformanceEntry[]`, and this function is accepting `RumPerformanceEntry`. The naming reflects this difference.

packages/core/src/utils.ts
BenoitZugmeyer: Array access is faster than mutating the array. I wonder why `.pop()` was used in the first place.

packages/rum/src/rum.ts
bcaudan: +1 for `shouldTrackResource`

packages/rum/src/performanceCollection.ts
bcaudan: OK make sense

packages/rum/src/performanceCollection.ts
bcaudan: OK, let's always cast early native type to RUM type to work with the same types.
cf resourceUtil `entry: RumPerformanceResourceTiming | PerformanceResourceTiming`

bcaudan (approved):
Great 👍

Issue #317, opened 2020-03-24
Browser library bundle size is very large.


threehams:
Hi! While looking into using Datadog RUM, I noticed the bundle size is much bigger than I'd expect: [@datadog/browser-rum bundle analysis](https://bundlephobia.com/result?p=@datadog/browser-rum@1.7.5). Currently it's about half the size of our entire UI component library, including styles.

Most of this comes from [@datadog/browser-core](https://bundlephobia.com/result?p=@datadog/browser-core@1.7.5) and its dependencies.

While there's a lot of code in the packages themselves, the size is mostly inefficiency in dependencies and bundling. This is good! It's the easiest to cut down.

Likely highest gains:
* Inline `merge` and `assign` functions in the project instead of using `lodash`. The two lodash imports account for almost 25% of the total size.
* Consider switching from `tsc` to [rollup](https://rollupjs.org/guide/en/). Rollup is specifically designed for building self-contained libraries, and will avoid the `tslib` dependency (the second largest after `lodash`).

bcaudan:
Hi @threehams, thanks a lot for your feedbacks.

We have not put too much effort on bundle size analysis for now but it is definitely something that we have planned to work on.
Thanks for pointing that up.

bcaudan:
Hello,

Thanks to @thibthib, lodash dependencies have been remove with #396.
It is available in `1.12.0` version.

isaachinman:
I've just pulled v1.25.2 into a project, and was also surprised at the size – [still 30kb of minified JS](https://bundlephobia.com/result?p=@datadog/browser-logs@1.25.2).

@bcaudan I would be happy to contribute if you would like to point me in the right direction.

fdc-viktor-luft:
Update: According to [Bundlephobia](https://bundlephobia.com/package/@datadog/browser-rum@3.9.0) and I also tested it in my projects version 3.9.0 is now at almost 40 kB gzipped. That makes a significant amount of all optimized apps. Roughly 25% my total bundle size.

<img width="473" alt="Bildschirmfoto 2021-12-14 um 14 56 52" src="https://user-images.githubusercontent.com/17571883/146012120-e313feba-eaa8-4822-a0fe-1c3f49f901e3.png">

Could you somehow put a little more focus to the bundle size?

threehams:
We've removed browser logging and RUM from our project due to the bundle size. It doesn't make sense to slow down a page in order to check if it's fast.

[...]

PR #493, opened 2020-08-14
⚡️ move tslib to peerDependencies

turadg:
## Motivation

https://github.com/DataDog/browser-sdk/issues/317

## Changes

Move `tslib` dep to `peerDependencies` so it's only bundled once.

## Testing

Package.json specifies.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -11,12 +11,12 @@
     "build:cjs": "rm -rf cjs && tsc -p tsconfig.cjs.json",
     "build:esm": "rm -rf esm && tsc -p tsconfig.esm.json"
   },
-  "dependencies": {
-    "tslib": "1.10.0"
-  },
   "devDependencies": {
     "@types/sinon": "7.0.13",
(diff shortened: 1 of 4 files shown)


BenoitZugmeyer:
Thank you for your contribution. This is a fair concern, but:
* We are using tslib to build the 'bundle' version of the SDKs, so I think it should still be in `devDependencies`
* I'm not a fan of the idea to ask consumers of this package to install `tslib` in their project, even if the project has nothing to do with TypeScript. I'd rather keep things simple, and maybe find a way to remove the need of tslib in `browser-sdk` altogether. As far as I can see, it is mainly used for 'object assign' and 'array spreading'. We could use our lib alternative for those (ex: `assign` from `core/utils`).

Your second commit to relax tslib version requirement could be useful to reduce tslib duplication though! I'll pick it in a separate PR.

BenoitZugmeyer:
Closing the PR. Extra steps will be discussed internally.

PR #498, opened 2020-08-20
[RUMF-642] do not return internal context if the session is untracked

BenoitZugmeyer:
## Motivation

When the session is untracked, we still create a new view object, but we don't send it to the intake. It should not be used by logs or trace because it would lead to an undefined view event.

## Changes

Make sure the session have an id and is tracked before exposing the internal context

## Testing

---


mquentin (approved):
Nice !

bcaudan (approved):

PR #506, opened 2020-08-25
🐛[RUMF-645] do not track intake request errors

bcaudan:
## Motivation

Intake request errors are reported which cause a loop of errors 

## Changes

Exclude intake request errors like intake resource events

## Testing

Automated tests

---

### DIFF
--- a/packages/core/src/configuration.ts
+++ b/packages/core/src/configuration.ts
@@ -1,4 +1,5 @@
 import { BuildEnv, BuildMode, Datacenter, INTAKE_SITE } from './init'
+import { haveSameOrigin } from './urlPolyfill'
 import { includes, ONE_KILO_BYTE, ONE_SECOND } from './utils'
 
 export const DEFAULT_CONFIGURATION = {
@@ -200,4 +201,13 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
   const parameters = `${applicationIdParameter}${proxyParameter}ddsource=${source || 'browser'}&ddtags=${tags}`
 
   return `https://${host}/v1/input/${conf.clientToken}?${parameters}`
+}
+
+export function isIntakeRequest(url: string, configuration: Configuration) {
+  return (
+    haveSameOrigin(url, configuration.logsEndpoint) ||
+    haveSameOrigin(url, configuration.rumEndpoint) ||
+    haveSameOrigin(url, configuration.traceEndpoint) ||
+    (configuration.internalMonitoringEndpoint && haveSameOrigin(url, configuration.internalMonitoringEndpoint))
+  )
 }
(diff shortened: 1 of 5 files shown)


glorieux (approved):

BenoitZugmeyer (commented):
Why not filtering requests at the proxy level directly?

* idealy, intake requests should not impact the `trackPageActivity`
* if we were to use those proxies for something else (ex: trace integration), we wouldn't need to think about checking if the request is an intake request.

bcaudan:
> Why not filtering requests at the proxy level directly?
> 
> * idealy, intake requests should not impact the `trackPageActivity`
> * if we were to use those proxies for something else (ex: trace integration), we wouldn't need to think about checking if the request is an intake request.

In this case, it would work but I would prefer to avoid putting any domain logic in these proxies.
On npm, they are reused between logs/rum instance so it is not a good idea to add a dependency on the configuration that could be different between products.

> idealy, intake requests should not impact the `trackPageActivity`

if we want to tackle that now, we could move the `shouldTrackResource` to `requestCollection` and `performanceCollection` (we can have intake request here too)

> * if we were to use those proxies for something else (ex: trace integration), we wouldn't need to think about checking if the request is an intake request.

for previous PR on trace integration, we used `requestCollection` so above strategy should work

wdyt?

BenoitZugmeyer:
> > Why not filtering requests at the proxy level directly?
> > 
> > * idealy, intake requests should not impact the `trackPageActivity`
> > * if we were to use those proxies for something else (ex: trace integration), we wouldn't need to think about checking if the request is an intake request.
> 
> In this case, it would work but I would prefer to avoid putting any domain logic in these proxies.
> On npm, they are reused between logs/rum instance so it is not a good idea to add a dependency on the configuration that could be different between products.

That's a very good point, which raises another issue: if the request to the RUM intake fails, the failed request will still be logged, since the logs SDK has no knowledge of the RUM intake configuration, isn't it?

> 
> > idealy, intake requests should not impact the `trackPageActivity`
> 
> if we want to tackle that now, we could move the `shouldTrackResource` to `requestCollection` and `performanceCollection` (we can have intake request here too)
> 
> > * if we were to use those proxies for something else (ex: trace integration), we wouldn't need to think about checking if the request is an intake request.
> 
> for previous PR on trace integration, we used `requestCollection` so above strategy should work
> 
> wdyt?

It seems to be a good strategy, to be done in a separate PR maybe?

bcaudan:
> That's a very good point, which raises another issue: if the request to the RUM intake fails, the failed request will still be logged, since the logs SDK has no knowledge of the RUM intake configuration, isn't it?

all intake endpoints are computed in `core/configuration` regardless of the SDK (logs or RUM)

> It seems to be a good strategy, to be done in a separate PR maybe?

sounds good to me

BenoitZugmeyer:
> > That's a very good point, which raises another issue: if the request to the RUM intake fails, the failed request will still be logged, since the logs SDK has no knowledge of the RUM intake configuration, isn't it?
> 
> all intake endpoints are computed in `core/configuration` regardless of the SDK (logs or RUM)

Ok so it will work as long as LOGS and RUM are using the same transport configuration (site, proxyHost, internalMonitoringKey, buildMode). Seems fair to me.

BenoitZugmeyer (approved):

PR #507, opened 2020-08-25
[RUMF-621] set view referrer to the previous view URL

BenoitZugmeyer:
## Motivation

For `route_change` views, use the previous view URL as referrer.

## Changes

For the sake of simplicity, lifecycle types have been improved a bit so we can re-use them easily. Then, the referrer has been set to the View context so it can be used everywher the View context is used. Finally, the view collections is retrieving the previous view URL when creating the new view on URL change or session renew.

## Testing

---

### DIFF
--- a/packages/rum/src/rum.ts
+++ b/packages/rum/src/rum.ts
@@ -202,14 +202,8 @@ export function startRum(
         const viewContext = parentContexts.findView(startTime)
         if (session.isTracked() && viewContext && viewContext.sessionId) {
           return (withSnakeCaseKeys(deepMerge(
-            {
-              applicationId,
-              sessionId: viewContext.sessionId,
-              view: {
-                id: viewContext.view.id,
-                url: viewContext.view.url,
-              },
-            },
+            { applicationId },
+            viewContext,
             parentContexts.findAction(startTime)
           ) as Context) as unknown) as InternalContext
         }
--- a/packages/rum/src/rum.ts
+++ b/packages/rum/src/rum.ts
@@ -134,9 +134,6 @@ interface RumContext {
   session: {
     type: string
   }
-  view: {
-    referrer: string
-  }
 }
 
 export type RawRumEvent = RumErrorEvent | RumResourceEvent | RumViewEvent | RumLongTaskEvent | RumUserActionEvent
@@ -186,9 +183,6 @@ export function startRum(
         // cf https://github.com/puppeteer/puppeteer/issues/3667
         type: getSessionType(),
       },
-      view: {
-        referrer: document.referrer,
-      },
     }),
     () => globalContext
   )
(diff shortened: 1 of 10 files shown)


packages/rum/test/parentContexts.spec.ts
```diff
@@ -17,6 +17,10 @@ describe('parentContexts', () => {
   const FAKE_ID = 'fake'
   const startTime = 10
 
+  function createViewCreatedEvent(partialViewCreatedEvent: Partial<ViewCreatedEvent> = {}): ViewCreatedEvent {
```
mquentin, line 20: Probably just a NIT but I think `ViewCreationEvent` would be easier to understand especially with this introduced function `createViewCreatedEvent`.

Can we also rename it something like `startViewCreationEvent` / `buildViewCreationEvent` ?

mquentin (approved):
with Small NIT

packages/rum/test/parentContexts.spec.ts
bcaudan: agree that it could be nice to do something with this function name 🙂

packages/rum/src/rum.ts
```diff
@@ -205,11 +199,17 @@ export function startRum(
...
+        if (session.isTracked() && viewContext && viewContext.sessionId) {
           return (withSnakeCaseKeys(deepMerge(
-            { applicationId },
-            view,
+            {
+              applicationId,
+              sessionId: viewContext.sessionId,
+              view: {
+                id: viewContext.view.id,
+                url: viewContext.view.url,
```
bcaudan, line 210: what about passing referrer in internal context too?

packages/rum/src/viewCollection.ts
```diff
@@ -184,7 +184,9 @@ function newView(
     },
     updateLocation(newLocation: Location) {
       location = { ...newLocation }
+      this.url = location.href
     },
+    url: location.href,
```
bcaudan, line 189: what about a getter instead of maintaining a new property?
```suggestion
    updateLocation(newLocation: Location) {
      location = { ...newLocation }
    },
    get url() { return location.href },
```

packages/rum/test/parentContexts.spec.ts
BenoitZugmeyer: For the interface name: the lifecycle event is called `VIEW_CREATED`, so naturally the event name should be called `ViewCreatedEvent`. If you think `VIEW_CREATED` should be changed to `VIEW_CREATION`, we should also rename all lifecycle event (completed/completion, mutated/mutation etc.)

For the function name, I don't mind changing it to `buildViewCreatedEvent`.

packages/rum/src/rum.ts
BenoitZugmeyer: I'm not sure this is valuable, but I don't mind adding it, it would save us a few lines of code

bcaudan: it could be nice to have the same info in related trace/log

bcaudan (approved):
Nice job!

PR #511, opened 2020-09-01
[RUMF-617] integrate tracing

bcaudan:
## Motivation

Instrument requests and generate spans from browser-sdk in order to avoid the overhead of dd-trace-js.
Rework from #461

## Changes 

- [x] remove trace collection
- [x] attach span id to resource event
- [x] use new format
- [x] make service required when tracing is enabled

## Testing

Automated tests

---


test/app/app.ts
```diff
@@ -16,11 +16,13 @@ datadogLogs.init({
 })
 
 datadogRum.init({
+  allowedTracingOrigins: [new RegExp(hostname)],
```
BenoitZugmeyer, line 19: `new RegExp(hostname)` is a bit scary as its behavior is unpredictable (it changes according to the domain). Maybe be more explicit?
```suggestion
  allowedTracingOrigins: typeof location === 'object' ? [location.origin] : undefined,
```

test/e2e/scenario/agents.scenario.ts
```diff
@@ -297,3 +305,38 @@ describe('anchor navigation', () => {
...
+    expect(headers['x-datadog-origin']).toBe('rum')
+  }
+
+  async function checkTraceAssociatedToRumEvent() {
+    const requests = (await waitServerRumEvents()).filter(
+      (rumEvent) =>
+        rumEvent.evt.category === 'resource' &&
+        ((rumEvent as ServerRumResourceEvent).resource.kind === 'xhr' ||
+          (rumEvent as ServerRumResourceEvent).resource.kind === 'fetch')
+    ) as ServerRumResourceEvent[]
```
BenoitZugmeyer, line 336: You should be able to remove all the castings with the use of `isRumResourceEvent`:

```suggestion
    const requests = (await waitServerRumEvents()).filter(
      (rumEvent) =>
        isRumResourceEvent(rumEvent) &&
        (rumEventresource.kind === 'xhr' ||
          rumEvent.resource.kind === 'fetch')
    )
```

packages/rum/src/tracer.ts
```diff
@@ -0,0 +1,136 @@
...
+export function toHexString(traceId: TraceIdentifier) {
+  return traceId.toString(16)
+}
+
+/**
+ * Format used elsewhere
+ */
+export function toDecimalString(traceId: TraceIdentifier) {
+  return traceId.toString(10)
+}
```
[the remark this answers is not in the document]
bcaudan: Leftover from when we were retrieving dd-trace-js TraceIdentifier, we could indeed simplify that now

test/app/app.ts
bcaudan: not sure to get your point, `location.origin` also changes according to the domain.
anyway, it is the npm test app configuration, so is it relevant?

BenoitZugmeyer: `new RegExp(hostname)` raises a lot of questions:

* what if hostname is not a valid regexp? Is it possible? should I care?
* what is the impact of dots in the hostname?
* because the regexp has no `^` and `$`, it matches any domain that contains the hostname, is it expected?

It will probably not have any impact on our tests, but nonetheless I don't think it should be shown as an acceptable solution anywhere in our code (even if this is not directly documentation it should reflect good practices).

bcaudan: I don't think it worth to discuss it further, I'll do the change

packages/rum/src/tracer.ts
```diff
@@ -110,6 +110,20 @@ export class TraceIdentifier {
     return str
   }
 
+  /**
+   * Format used by trace intake
+   */
+  toHexString() {
+    return this.toString(16)
+  }
```
BenoitZugmeyer, line 118: NIT: It seems that this method is not used, maybe remove it

BenoitZugmeyer (approved):

PR #517, opened 2020-09-04
🐛[RUMF-617] fix missing headers on traced requests

bcaudan:
## Motivation

some requests headers possible types were not copied with the tracing headers, resulting in some request failures due to missing headers.

## Changes

Handle other headers types

## Testing

Automated tests

---

### DIFF
--- a/packages/rum/src/tracer.ts
+++ b/packages/rum/src/tracer.ts
@@ -19,11 +19,19 @@ export function startTracer(configuration: Configuration): Tracer {
     traceFetch: (context) =>
       injectHeadersIfTracingAllowed(configuration, context.url!, (tracingHeaders: TracingHeaders) => {
         context.init = { ...context.init }
-        const headers: { [key: string]: string } = {}
 
+        const headers: { [key: string]: string } = {}
         if (context.init.headers instanceof Headers) {
           context.init.headers.forEach((value, key) => {
             headers[key] = value
+          })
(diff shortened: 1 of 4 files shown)


packages/rum/src/tracer.ts
```diff
@@ -19,12 +19,20 @@ export function startTracer(configuration: Configuration): Tracer {
...
+          context.init.headers.forEach(([key, value]) => {
+            headers[key] = value
+          })
+        } else if (context.init.headers) {
+          Object.keys(context.init.headers).forEach((key) => {
+            headers[key] = (context.init!.headers as Record<string, string>)[key]
+          })
         }
 
         context.init.headers = { ...headers, ...tracingHeaders }
```
BenoitZugmeyer, line 38: This should be simpler and safer:
```suggestion
        const headers = new Headers(context.init.headers)
        objectEntries(tracingHeaders).forEach(([key, value]) => headers.set(key, value))
        context.init.headers = headers
```

Headers can be multi-valued:
```
new Headers([[ 'X-Foo', 'bar' ], ['X-Foo', 'baz']]).get('x-foo')  // bar, baz
```

With your solution, when iterating over the array, we lose the previous value (`bar` in the example)

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):
LGTM

PR #518, opened 2020-09-04
 🐛 [CORE] fix error collection for simultaneous rum + logs

MaxSechz:
## Motivation

We are using both RUM and Logs in our application. We want to forward errors for ingestion using RUM, but not for Logs. When we switched from including the libs via a script tag to bundling them w/ webpack we noticed that we were getting errors submitted to our logs, even though we set `forwardErrorsToLogs: false`. What we determined was that since we were initializing RUM first, the ErrorObservable created during the config process was reused for configuring Logs during initialization, regardless of our config settings.

## Changes

This change essentially reverts https://github.com/DataDog/browser-sdk/pull/249/commits/b00054de3d9b2c3ee431e21d9dc6957d54ca89d9 to ensure new ErrorObservables are created on each `commonInit` call. The bug addressed by https://github.com/DataDog/browser-sdk/pull/249/commits/b00054de3d9b2c3ee431e21d9dc6957d54ca89d9 is then fixed by only forming a closure over the current `console.error` in `startConsoleTracking` as opposed to a variable in the module scope. This will prevent a subsequent ErrorObservable from causing an infinite recursion in the originally overloaded `console.error`

## Testing

Run both RUM and Logs using webpack. Set one package to forward errors, and the other one to not forward errors. The setting from the package that is initialized first will apply to both, meaning either both will forward no console errors, or both will forward all console errors.

## Addendum

While this may seem like a niche issue, I am concerned about the implications of library users not noticing this behavior. This could result in users sending no errors and believing that their code is actually not generating any because they configured one of the libs to forward them. Conversely, it may result in extra unwanted data ingestion/indexing due to users believing the have error forwarding switched off in one of the libs, which may end up costing them money.

---

### DIFF
--- a/packages/core/src/errorCollection.ts
+++ b/packages/core/src/errorCollection.ts
@@ -36,19 +36,15 @@ export enum ErrorOrigin {
 }
 
 export type ErrorObservable = Observable<ErrorMessage>
-let filteredErrorsObservable: ErrorObservable
 
 export function startErrorCollection(configuration: Configuration) {
-  if (!filteredErrorsObservable) {
-    const errorObservable = new Observable<ErrorMessage>()
-    if (configuration.isCollectingError) {
-      trackNetworkError(configuration, errorObservable)
-      startConsoleTracking(errorObservable)
-      startRuntimeErrorTracking(errorObservable)
-    }
-    filteredErrorsObservable = filterErrors(configuration, errorObservable)
+  const errorObservable = new Observable<ErrorMessage>()
+  if (configuration.isCollectingError) {
+    trackNetworkError(configuration, errorObservable)
+    startConsoleTracking(errorObservable)
+    startRuntimeErrorTracking(errorObservable)
   }
-  return filteredErrorsObservable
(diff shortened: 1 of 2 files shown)


bcaudan:
Hi @MaxSechz,

Thank you for raising that, we will look into it.

BenoitZugmeyer:
@MaxSechz thank you for your well-written PR. I used a different strategy to fix the issue (see PR #522):

* to avoid unnecessary performance cost and stack trace polution, we want to avoid applying the same patches on native objects multiple times if both logs and RUM are starting with error collection.
* while your tests make sense, the stuff started when calling startErrorCollection is not always stoped, and we'll talk what would be the best approach for this internally.

MaxSechz:
@BenoitZugmeyer I appreciate the quick response. Your changes make sense to me. Thanks!

BenoitZugmeyer:
This should be fixed in the latest release `v1.19.0`. Please let us know if you still have issuse on this!

PR #525, opened 2020-09-08
🐛 [RUMF-670] wait for the DOM to be ready before getting the trace id

BenoitZugmeyer:
## Motivation

Because the RUM SDK can be initialized synchronously when the document is loading, the DOM may not be fully accessible when getting the trace id set by APM. In particular, if meta tags are written after the SDK snippet, the SDK will not have access to them during init.

## Changes

Wait for DOM to be loaded before getting the trace id and generating the initial document resource.

## Testing

Tests have been written in the integration test POC PR https://github.com/DataDog/browser-sdk/pull/524

---

### DIFF
--- a/packages/rum/src/performanceCollection.ts
+++ b/packages/rum/src/performanceCollection.ts
@@ -68,7 +68,9 @@ function supportPerformanceNavigationTimingEvent() {
 }
 
 export function startPerformanceCollection(lifeCycle: LifeCycle) {
-  handleRumPerformanceEntry(lifeCycle, retrieveInitialDocumentResourceTiming())
+  retrieveInitialDocumentResourceTiming((timing) => {
+    handleRumPerformanceEntry(lifeCycle, timing)
+  })
 
   if (supportPerformanceObject()) {
     handlePerformanceEntries(lifeCycle, performance.getEntries())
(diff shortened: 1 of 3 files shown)


packages/rum/src/performanceCollection.ts
```diff
@@ -99,49 +101,58 @@ export function startPerformanceCollection(lifeCycle: LifeCycle) {
...
-    callback({
-      ...computeRelativePerformanceTiming(),
-      entryType: 'navigation',
-    })
-  }
+  runOnReadyState('complete', () => {
+    // Send it a bit after the actual load event, so the "loadEventEnd" timing is accurate
+    setTimeout(
+      monitor(() => {
+        callback({
```
mquentin, line 136: Good to go for me. 
I just think that having the named function sendFakeTiming was quite useful for the understanding of this code block.
```
function sendFakeTiming() {
    callback({
      ...computeRelativePerformanceTiming(),
      entryType: 'navigation',
    })
  }
```
I would be un favor of keeping it :


```
runOnReadyState('complete', () => {
    // Send it a bit after the actual load event, so the "loadEventEnd" timing is accurate
    setTimeout(monitor(sendFakeTiming))
})
```

mquentin (approved):
Good to go with a NIT comment

packages/rum/src/performanceCollection.ts
```diff
@@ -99,49 +101,58 @@ export function startPerformanceCollection(lifeCycle: LifeCycle) {
   }
 }
 
-export function retrieveInitialDocumentResourceTiming() {
-  let timing: RumPerformanceResourceTiming
+export function retrieveInitialDocumentResourceTiming(callback: (timing: RumPerformanceResourceTiming) => void) {
+  runOnReadyState('interactive', () => {
```
bcaudan, line 105: Other function name is `retrieveNavigationTimingWhenLoaded`, so I think it could be nice to also reflect in the function name that it is waiting for a specific page state.
However, it could be even clearer to move the `runOnReadyState` outside of these functions:

```
runOnReadyState('interactive', () => {
  retrieveInitialDocumentResourceTiming((timing) => {
    handleRumPerformanceEntry(lifeCycle, timing)
  })
}}

runOnReadyState('complete', () => {
  retrieveNavigationTiming((timing) => {
    handleRumPerformanceEntry(lifeCycle, timing)
  })
}}
```

BenoitZugmeyer: Waiting for the DOM to be ready or document to be loaded is an implementation detail and I think it's better to keep it in the function. In particular, `retrieveNavigationTiming` is waiting for the document load + an bit of time for the `loadEventEnd` to be defined: it would be hard to understand what's going on if those tasks were in two different places.

I'm fine with renaming the function to `retrieveInitialDocumentResourceTiming` to `retrieveInitialDocumentResourceTimingWhenDomReady` even if I don't see much value in it.

bcaudan: ok, so maybe rename `retrieveNavigationTimingWhenLoaded` to `retrieveNavigationTiming` to be consistent in considering it as an implementation detail

bcaudan (approved):

PR #528, opened 2020-09-10
🐛[RUMF-650] exclude intake request from performance/request collection

bcaudan:
## Motivation

avoid page activities to consider intake requests

## Changes

- move check on url before life cycle event generation
- keep session check before resource event generation

## Testing

Automated test

---


packages/rum/src/requestCollection.ts
```diff
@@ -45,33 +46,41 @@ interface CustomXhrContext extends XhrContext {
...
+) {
   const xhrProxy = startXhrProxy<CustomXhrContext>()
   xhrProxy.beforeSend((context, xhr) => {
-    const tracingResult = tracer.traceXhr(context, xhr)
-    if (tracingResult) {
-      context.traceId = tracingResult.traceId
-      context.spanId = tracingResult.spanId
-    }
-    context.requestIndex = getNextRequestIndex()
+    if (isAllowedRequestUrl(configuration, context.url!)) {
```
BenoitZugmeyer, line 56: We could specify a proper type for `context` instead of `Partial<XhrContext>` in this case: we now that we'll always have `startTime`/`url`/`method`, a no other property

packages/rum/src/requestCollection.ts
```diff
@@ -82,34 +91,42 @@ interface CustomFetchContext extends FetchContext {
...
-      response: context.response,
-      responseType: context.responseType,
-      spanId: context.spanId,
-      startTime: context.startTime,
-      status: context.status,
-      traceId: context.traceId,
-      type: RequestType.FETCH,
-      url: context.url,
-    })
+    if (isAllowedRequestUrl(configuration, context.url!)) {
```
BenoitZugmeyer, line 115:
```suggestion
    if (isAllowedRequestUrl(configuration, context.url)) {
```

BenoitZugmeyer (approved):
Really nice! I like the idea of filtering URLs at a lower level :) Some small typings suggestions but other than that LGTM

BenoitZugmeyer (approved):

PR #545, opened 2020-09-24
 ✨ [RUMF-530][Logs] allow using logs API before init 

BenoitZugmeyer:
## Motivation

We want to allow using log APIs before initialising the SDK.

## Changes

* Remove stubs and regroup the public API implementation
* Introduce a 'buffered function' to buffer pre-ini function calls
* Use a buffered function to send logs

## Testing

Unit tests have been updated. Manual testing can be done by running some `DD_LOGS.logger.log('foo')` before `DD_LOGS.init()`.

---


packages/core/src/cookie.ts
```diff
@@ -57,7 +57,7 @@ export function getCookie(name: string) {
   return findCommaSeparatedValue(document.cookie, name)
 }
 
-export function areCookiesAuthorized(useSecureCookie: boolean): boolean {
+export function areCookiesAuthorized(useSecureCookie?: boolean): boolean {
```
bcaudan, line 60: IMO, it means that we can perform this check as unsecure when we don't know if we should be secure or not, which is not great.
what about keeping the same strategy as before (`mustUseSecureCookie(userConfiguration)`) or proposing another strategy that avoid this optional boolean?

packages/logs/src/logs.entry.ts
```diff
@@ -1,119 +1,163 @@
...
+  const sendLogBuffered = createBufferedFunction((message: LogsMessage, currentContext: Context) => {
+    if (session.isTracked()) {
+      batch.add(message, currentContext)
     }
-    const { errorObservable, configuration, internalMonitoring } = commonInit(logsUserConfiguration, buildEnv)
-    const session = startLoggerSession(configuration, areCookiesAuthorized(mustUseSecureCookie(userConfiguration)))
-    const globalApi = startLogger(errorObservable, configuration, session, internalMonitoring)
-    assign(global, globalApi)
-    isAlreadyInitialized = true
+  })
```
bcaudan, line 79: This approach works but the design of this file is not so great, it is hard to follow what is going on.
It could be better to have a strategy that allow us to better separate the different responsibilities.

With the strategy suggested in the RFC, the buffer logic is separated from the after init logic.
That could allow us to move most of the initialization logic to a separate file which would reduce the complexity of this one.

wdyt?

BenoitZugmeyer: Do you have concrete examples of why my code is not great, and how do you think it could be improved?

bcaudan: makeLogsGlobal is too complex and has too much responsibilities: 
- it creates the public API
- it handles the init logic, but take the "base" init as a dependency, which make things harder to follow and I don't see the need for it.
- due to the createBufferedFunction strategy, it needs to have access to all the components that are required once logs is initialized (batch, internal monitoring, ...) which then couple the before init logic to the after init logic.

IMO, we could make clearer the different problematic by not mixing everything.
I've tried to illustrate what I have in mind with this commit: https://github.com/DataDog/browser-sdk/commit/7174ad1e0ffbf5c30a50b8b25d1cbc6b6cb82f81

BenoitZugmeyer: > take the "base" init as a dependency, which make things harder to follow and I don't see the need for it
This is needed for unit test, so we can inject a session and a configuration (`maxBatchSize`).

Thank you for your commit, I'll make something like this.

packages/logs/test/logs.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+    errorObservable = new Observable<ErrorMessage>()
+    mockModule('./packages/core/src/init.ts', () => ({
+      commonInit() {
+        return {
+          errorObservable,
+          configuration: { ...(configuration as Configuration), ...configurationOverrides },
+          internalMonitoring: { setExternalContextProvider: () => undefined },
+        }
+      },
+    }))
```
bcaudan, line 76: It seems to me that you are trying to change the behavior of inner components of your test subject.

What about updating the design in order that your test subject take these components as dependencies?
it would avoid to introduce the `mockModule` technique and would be more consistent with what we are doing elsewhere in the code base.

packages/logs/src/logger.ts
```diff
@@ -39,142 +22,37 @@ export interface LogsMessage {
   [key: string]: ContextValue
 }
 
-export interface LoggerConfiguration {
-  level?: StatusType
-  handler?: HandlerType
-  context?: Context
-}
```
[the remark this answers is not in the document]
BenoitZugmeyer: This is only used for the `createLogger` API, not the `Logger` class.

packages/logs/src/logger.ts
```diff
@@ -39,142 +22,37 @@ export interface LogsMessage {
...
-        session_id: session.getId(),
-        view: {
-          referrer: document.referrer,
-          url: window.location.href,
-        },
-      },
-      globalContextProvider(),
-      getRUMInternalContext(),
-      message
-    )
```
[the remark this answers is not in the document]
BenoitZugmeyer: I think we have? Or maybe I missed something?

packages/logs/test/logs.spec.ts
BenoitZugmeyer: I introduced module mocking *because* you didn't like injecting initialisation values for tests: https://github.com/DataDog/browser-sdk/pull/545#discussion_r494974626
> it handles the init logic, but take the "base" init as a dependency, which make things harder to follow and I don't see the need for it.

Anyway, I really like mocking modules. This is the way `jest` is working. It allows to test the code without having to inject everything. I think it is a superior design to the RUM specBuilder we have in the RUM side.

If you find it confusing I'll revert the `baseInit` function.

packages/logs/src/logs.ts
```diff
@@ -0,0 +1,99 @@
...
+export function startLogs(
+  userConfiguration: LogsUserConfiguration,
+  errorLogger: Logger,
+  getGlobalContext: () => Context
+) {
+  const isCollectingError = userConfiguration.forwardErrorsToLogs !== false
+  const logsUserConfiguration = {
+    ...userConfiguration,
+    isCollectingError,
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: I think it would be better to completly `isCollectingError` from the configuration. Each SDK are overriding their configuration to set this value before passing it to `commonInit`. It would be much simpler to pass it as a second argument to `commonInit`.
WDYT?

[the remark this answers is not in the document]
bcaudan: I am assuming it is

> to completely **remove** `isCollectingError`

Yeah, it could be nice

packages/logs/src/logger.ts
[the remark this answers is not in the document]
bcaudan: No, it is me 😄 it could be interesting to have a test for it though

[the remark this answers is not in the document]
BenoitZugmeyer: I did add tests for it https://github.com/DataDog/browser-sdk/pull/545/commits/e0d7b8ff5d17b1e8d856f970fb4cc1bc5a4c3b66#diff-dfe7564216dcb9cd4d677851033e77c4R114-R138

packages/logs/test/logs.spec.ts
bcaudan: My previous comment was probably not clear enough, in the previous approach:
```ts
export function makeLogsGlobal(
  baseInit: (
    configuration: LogsUserConfiguration
  ) => {
    session: LoggerSession
    configuration: Configuration
    errorObservable: ErrorObservable
    internalMonitoring: InternalMonitoring
  }
)
```
A Part of the init was made on this base function and another part in the makeLogsGlobal which was hard to follow.
Base init function returned 4 objects which could be hidden from makeLogsGlobal, providing a better encapsulation of the init logic from the evaluation logic.

Now, with something like `makeLogsGlobal(startLog)`, by stubbing `startLog`, we could easily unit test the pre-init logic, the conf check logic and the sendLog strategy switch without needing to be aware of session, errorObservable, ...

> I think it is a superior design to the RUM specBuilder we have in the RUM side.

From my point of view, it is not what we should compare. 
In the current code base, we inject dependencies as parameters. It allows us to easily stub these dependencies in unit test.
With the `mockModule` technique, I can see several drawbacks:
- unit tests are now coupled with webpack and its module cache management
- unit tests are now coupled to some file path, ex:`mockModule('./packages/logs/src/logs.ts', () =>` 
- no type check are performed on mocked method output which could easily lead to confusing errors

packages/logs/src/logger.ts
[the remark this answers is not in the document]
bcaudan: Changing the order between `currentContext` and `getRUMInternalContext` does not fail a test, so it should at least missed this case.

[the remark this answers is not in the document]
BenoitZugmeyer: Should be better now

packages/logs/test/logs.spec.ts
BenoitZugmeyer: I reverted the module mocks and brought back dependency injection.

packages/logs/README.md
```diff
@@ -40,7 +40,7 @@ What we call `Context` is a map `{key: value}` that will be added to the message
 
   - Configurable options:
 
-    - `isCollectingError`: when truthy, we'll automatically forward `console.error` logs, uncaught exceptions and network errors.
+    - `forwardErrorsToLogs`: when truthy, we'll automatically forward `console.error` logs, uncaught exceptions and network errors.
```
bcaudan, line 43: 😮

packages/logs/src/logs.ts
```diff
@@ -0,0 +1,120 @@
...
+import { LoggerSession, startLoggerSession } from './loggerSession'
+import { LogsUserConfiguration } from './logs.entry'
+
+export const startLogs = makeStartLogs((userConfiguration) => {
+  const isCollectingError = userConfiguration.forwardErrorsToLogs !== false
+  const { configuration, internalMonitoring, errorObservable } = commonInit(
+    userConfiguration,
+    buildEnv,
+    isCollectingError
+  )
```
bcaudan, line 26: IMO we could have an approach
```ts
export function startLogs(...) {
  // base init
  return doStartLogs(...)
}

export function doStartLogs(...) {
  ...
}
```
and unit test `doStartLogs`.
it would remove some complexity introduce by the factory pattern and make more clear what are the input at the different levels.

bcaudan (approved):
Nice job 👍

packages/core/src/configuration.ts
```diff
@@ -5,7 +5,6 @@ import { includes, ONE_KILO_BYTE, ONE_SECOND } from './utils'
 
 export const DEFAULT_CONFIGURATION = {
   allowedTracingOrigins: [] as Array<string | RegExp>,
-  isCollectingError: true,
```
tbillington, line 8: Just letting you know this is a breaking change but wasn't reported as such in your changelog.

bcaudan: 👋 @tbillington, it should not be.
There was an issue in the logs documentation, `isCollectingError` was not an init parameter for logs, `forwardErrorToLogs` was the right one: https://github.com/DataDog/browser-sdk/pull/545/files#diff-2e4bca1144de53c8c504edb040ac34aa7ea63ba413a4453277d88307278d8972R43.
It is still collecting error by default: https://github.com/DataDog/browser-sdk/pull/545/files#diff-c62c675364c11388aeae30517afdac97f57f2638e8fb3d019177ebfeb382f540R25

If you still see something wrong, would you mind opening a dedicated issue and provide more details?

PR #546, opened 2020-09-24
✨ [RUMF-709][core] support 'null' as a context value 

BenoitZugmeyer:
## Motivation

Ensure that nothing breaks when the user unexpectedly pass `null` for values that end up in a `Context`.

## Changes

Add proper support for `null` value in context.

## Testing

Unit tests have been upgraded. You can test it manually by calling `DD_RUM.init({ service: null, clientToken: 'xxx', applicationId: 'xxx' })`

---

### DIFF
--- a/packages/core/test/utils.spec.ts
+++ b/packages/core/test/utils.spec.ts
@@ -21,6 +21,11 @@ describe('utils', () => {
       expect(deepMerge({ a: 1 }, { a: undefined })).toEqual({ a: 1 })
     })
 
+    it('should support null values', () => {
+      // tslint:disable-next-line: no-null-keyword
+      expect(deepMerge({ a: {} }, { a: null })).toEqual({ a: null })
+    })
+
     it('should merge arrays', () => {
       const target = [{ a: 'target' }, 'extraString']
       const source = [{ b: 'source' }]
@@ -259,10 +264,14 @@ describe('utils', () => {
         withSnakeCaseKeys({
           camelCase: 1,
           nestedKey: { 'kebab-case': 'helloWorld', array: [{ camelCase: 1 }, { camelCase: 2 }] },
+          // tslint:disable-next-line: no-null-keyword
+          nullValue: null,
         })
       ).toEqual({
         camel_case: 1,
         nested_key: { kebab_case: 'helloWorld', array: [{ camel_case: 1 }, { camel_case: 2 }] },
+        // tslint:disable-next-line: no-null-keyword
+        null_value: null,
       })
     })
   })
--- a/packages/core/test/utils.spec.ts
+++ b/packages/core/test/utils.spec.ts
@@ -21,9 +21,14 @@ describe('utils', () => {
       expect(combine({ a: 1 }, { a: undefined as number | undefined })).toEqual({ a: 1 })
     })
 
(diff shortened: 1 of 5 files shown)


packages/logs/src/logger.ts
```diff
@@ -61,8 +61,8 @@ export function startLogger(
 ) {
   let globalContext: Context = {}
 
-  internalMonitoring.setExternalContextProvider(
-    () => deepMerge({ session_id: session.getId() }, globalContext, getRUMInternalContext() as Context) as Context
+  internalMonitoring.setExternalContextProvider(() =>
+    combine({ session_id: session.getId() }, globalContext, getRUMInternalContext())
```
bcaudan, line 65: Nice! What about keeping deepMerge private to only use combine?

BenoitZugmeyer: I kept the export for the unit tests. But maybe we could replace `deepMerge` by `combine` in unit tests as well?

packages/core/test/utils.spec.ts
```diff
@@ -21,6 +21,11 @@ describe('utils', () => {
       expect(deepMerge({ a: 1 }, { a: undefined })).toEqual({ a: 1 })
     })
 
+    it('should support null values', () => {
+      // tslint:disable-next-line: no-null-keyword
+      expect(deepMerge({ a: {} }, { a: null })).toEqual({ a: null })
```
bcaudan, line 26: There is a strange behavior:
```ts
expect(deepMerge({ a: 1 }, undefined)).toEqual({ a: 1 })
expect(deepMerge({ a: 1 }, null)).toEqual(null)
```
could it be better to have:
```ts
expect(deepMerge({ a: 1 }, null)).toEqual({ a: 1 })
```

packages/logs/src/logger.ts
bcaudan: sound good to me

packages/core/test/utils.spec.ts
BenoitZugmeyer: Mmh, if we ignore `null` values, that would make things like:

```ts
DD_LOGS.setLogsGlobalContext({ page: 'toto' })
DD_LOGS.logger.log('message', { page: null })
// context is { page: 'toto' }
```
I find this a bit confusing. I can understand that `undefined` gets ignored, as it's kind of idiomatic in JS:

```ts
function log(page = 'toto') {
   // [...]
}
log(undefined) // page is 'toto'
```
But it's not the case for null values

```ts
log(null) // page is null
```

But we can also consider the `??` operator which handle `null` and `undefined` values the same:

```js
null ?? 1 // 1
undefined ?? 1 // 1
```

I have mixed feelings about this, but I think it's fine if we handle `null` and `undefined` values differently and don't ignore `null` values, so the user can override its context with `null`. WDYT?

bcaudan: I would be in favor of having the same behavior as [lodash](https://lodash.com/docs/4.17.15#merge):
Capture d’écran 2020-09-25 à 11 53 36

BenoitZugmeyer: I see. I'll give it a shot

bcaudan (approved):

PR #549, opened 2020-10-02
[RUMF-553] add dummy events format validation

bcaudan:
## Motivation

Test link with rum-events-format and validation tool.

## Changes

- rum-events-format as git submodule
- simple validation test ran from ci

## Testing

cf ci step events-format

---

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "rum-events-format"]
+	path = rum-events-format
+	url = https://github.com/DataDog/rum-events-format
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -126,6 +126,14 @@ check-licenses:
   script:
     - node --no-warnings scripts/check-licenses.js
 
+events-format:
+  stage: test
+  tags: ['runner:main', 'size:large']
+  image: $CI_IMAGE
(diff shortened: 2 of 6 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "rum-events-format"]
+	path = rum-events-format
+	url = https://github.com/DataDog/rum-events-format
```
glorieux, line 3: Isn't it possible to add `rum-event-format` as a dev dependency in the package.json by using:
```
"rum-event-format": "DataDog/rum-event-format"
```
It would remove the need to run the `git submodule` command before every test.

bcaudan: yes, it is. However, the behavior are not the same:

as a *node dependency*:

- a rum-events-format specific commit is linked in the yarn.lock, so we need to manually update it at each rum-events-format modification
- repo will be downloaded for each ci job

as a *git submodule*:

- we can always test to the HEAD of the repo but also be tied to a specific commit if we want to
- we can pull repo only when we need to (before the validation test)

glorieux: 👌

glorieux (approved):

BenoitZugmeyer (approved):

PR #551, opened 2020-10-02
✨ [RUMF-530][RUM] allow using RUM API before init 

BenoitZugmeyer:
## Motivation

We want to allow using RUM APIs before initialising the SDK. (see #545 for the Logs side)

## Changes

* Remove stubs and regroup the public API implementation
* Use a bounded buffer to collect pre-init user actions and send them at init
* Make `getInternalContext` return `undefined` before init

## Testing

Unit tests have been updated. Manual testing can be done by running some `DD_RUM.addUserAction('foo')` or `DD_RUM.getInternalContext()` before `DD_RUM.init()`.

---

### DIFF
--- a/packages/rum/src/rum.entry.ts
+++ b/packages/rum/src/rum.entry.ts
@@ -49,7 +49,7 @@ export function makeRumGlobal(startRumImpl: StartRum) {
   let globalContext: Context = {}
 
   let getInternalContextStrategy: ReturnType<StartRum>['getInternalContext'] = () => {
-    throw new Error('TODO')
+    return undefined
   }
   let addUserActionStrategy: ReturnType<StartRum>['addUserAction'] = () => {
     throw new Error('TODO')
--- a/packages/rum/src/rum.entry.ts
+++ b/packages/rum/src/rum.entry.ts
@@ -1,27 +1,18 @@
 import {
-  assign,
   checkCookiesAuthorized,
   checkIsNotLocalFile,
-  commonInit,
   Context,
   ContextValue,
   getGlobalObject,
   isPercentage,
   makeGlobal,
-  makeStub,
   monitor,
   mustUseSecureCookie,
   UserConfiguration,
 } from '@datadog/browser-core'
 
-import { buildEnv } from './buildEnv'
-import { startDOMMutationCollection } from './domMutationCollection'
-import { LifeCycle, LifeCycleEventType } from './lifeCycle'
-import { startPerformanceCollection } from './performanceCollection'
-import { startRequestCollection } from './requestCollection'
 import { startRum } from './rum'
-import { startRumSession } from './rumSession'
-import { startUserActionCollection } from './userActionCollection'
+import { UserActionType } from './userActionCollection'
 
 export interface RumUserConfiguration extends UserConfiguration {
   applicationId: string
@@ -40,31 +31,9 @@ export interface InternalContext {
   }
 }
 
-const STUBBED_RUM = {
-  init(userConfiguration: RumUserConfiguration) {
-    makeStub('core.init')
-  },
-  addRumGlobalContext(key: string, value: ContextValue) {
-    makeStub('addRumGlobalContext')
-  },
-  removeRumGlobalContext(key: string) {
-    makeStub('removeRumGlobalContext')
-  },
-  setRumGlobalContext(context: Context) {
-    makeStub('setRumGlobalContext')
-  },
-  addUserAction(name: string, context: Context) {
-    makeStub('addUserAction')
-  },
-  getInternalContext(startTime?: number): InternalContext | undefined {
-    makeStub('getInternalContext')
-    return undefined
-  },
-}
+export type RumGlobal = ReturnType<typeof makeRumGlobal>
 
-export type RumGlobal = typeof STUBBED_RUM
-
-export const datadogRum = makeRumGlobal(STUBBED_RUM)
+export const datadogRum = makeRumGlobal(startRum)
 
 interface BrowserWindow extends Window {
   DD_RUM?: RumGlobal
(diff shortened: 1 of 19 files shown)


packages/rum/src/rum.ts
```diff
@@ -310,7 +346,7 @@ function makeRumEventHandler(
       if (session.isTracked() && view && view.sessionId) {
         const action = parentContexts.findAction(startTime)
         const rumEvent = assemble(event, { action, view, rum: rumContextProvider() })
-        const message = combine(globalContextProvider(), customerContext, withSnakeCaseKeys(rumEvent))
+        const message = combine(customerContext || globalContextProvider(), withSnakeCaseKeys(rumEvent))
```
bcaudan, line 349: What is the reasoning for that? reverting this change does not fail any unit test

packages/rum/src/rum.ts
```diff
@@ -344,8 +374,8 @@ function trackRumEvents(
     handler(assembleWithoutAction, (message, event: RumEvent) => batch.upsert(message, event.view.id))
   )
   trackErrors(lifeCycle, handler(assembleWithAction, batch.add))
-  trackRequests(configuration, lifeCycle, session, handler(assembleWithAction, batch.add))
-  trackPerformanceTiming(configuration, lifeCycle, session, handler(assembleWithAction, batch.add))
+  trackRequests(lifeCycle, session, handler(assembleWithAction, batch.add))
+  trackPerformanceTiming(lifeCycle, session, handler(assembleWithAction, batch.add))
```
bcaudan, line 378: noice!

packages/rum/test/rum.spec.ts
```diff
@@ -547,23 +508,24 @@ describe('rum global context', () => {
...
-    const { server, lifeCycle, rumApi } = setupBuilder.build()
+    const { server, lifeCycle, setGlobalContext } = setupBuilder.build()
     server.requests = []
 
-    rumApi.addRumGlobalContext('bar', 'foo')
+    const globalContext = { bar: 'foo' }
+    setGlobalContext(globalContext)
     lifeCycle.notify(LifeCycleEventType.ERROR_COLLECTED, FAKE_ERROR as ErrorMessage)
-    rumApi.removeRumGlobalContext('bar')
+    delete globalContext.bar
```
bcaudan, line 517: what about splitting these tests in two:
- unit test of global context CRUD in rum.entry.spec
- unit test of global context attached to request / snake case in rum.spec

packages/rum/src/rum.ts
BenoitZugmeyer: I added a unit test for it. The goal is to make sure the current global context is not used when formating a user action that was sent a while ago (same as logs). But this could be challenged... WDYT?

packages/rum/test/rum.spec.ts
BenoitZugmeyer: I thought about it, and I think we could benefit from a small common module that would implement context handling (add/set/remove), used in logs.entry, rum.entry and logger. This way, we would only need to test this common module.

In any case, since the context is (currently) mutable, I think it makes sense to keep a test like this here, to show that the current context is deeply merged into the generated event. The test name could be more accurate though (like: `should ignore subsequent context mutation`)

WDYT?

packages/rum/src/rum.ts
bcaudan: ok I missed the [early combine](https://github.com/DataDog/browser-sdk/pull/551/files#diff-fcb436c4fbf0511b343fcf0e467d005bR102) of action and global context.

FMU, it is not the same strategy as logs since in the logs case we are saving both [LogsMessage and CurrentContext](https://github.com/DataDog/browser-sdk/blob/master/packages/logs/src/logs.entry.ts#L115) to decide how to combine them later.

By doing the same for RUM, we could pass the global context in the handler and keep all the combine logic at a single place:

```ts
const message = combine(savedGlobalContext || globalContextProvider(), customerContext, withSnakeCaseKeys(rumEvent))
```

which IMO would be easier to follow/understand.

BenoitZugmeyer: Okay! Sure, it's not the same strategy, what I meant was the end result is the same (using the old global context instead of the current one).

In logs, we have to keep the message and current context separated because we have to set the current RUM context in between. This is not needed in RUM, but if you think it's easier to understand, I'll store the current context and customer context separated

packages/rum/src/userActionCollection.ts
```diff
@@ -21,7 +21,8 @@ export interface CustomUserAction {
   type: UserActionType.CUSTOM
   name: string
   startTime: number
-  context?: Context
+  context: Context
+  globalContext: Context
```
bcaudan, line 25: it seems weird to me that only this event type contains the global context.
Why not propagating the global context like on the logs side `[CustomUserAction, Context]`?

by having:
```ts
 addUserActionStrategy(action, context)
```

we could also only copy contexts before init and use current ones after init. 
Or is there a reason to also copy them after init?

BenoitZugmeyer: So, if we use `[CustomUserAction, Context]`, this means that we would have a `LifeCycle.subscribe` signature like:

```ts
  subscribe(eventType: LifeCycleEventType.CUSTOM_ACTION_COLLECTED, callback: (data: [CustomUserAction, Context]) => void): Subscription
```

Does it looks right to you? I'd prefer to keep having an object instead of a tuple.

packages/rum/src/rum.entry.ts
```diff
@@ -99,7 +94,8 @@ export function makeRumGlobal(startRumImpl: StartRum) {
     addUserAction: monitor((name: string, context?: Context) => {
       addUserActionStrategy({
         name,
-        context: combine({}, globalContext, context),
+        context: combine({}, context),
+        globalContext: combine({}, globalContextManager.get()),
```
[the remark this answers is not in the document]
BenoitZugmeyer: FTR: I'll do this in a separate PR as it has other implications (`combine` does not perform a deep clone at the moment)

packages/rum/src/userActionCollection.ts
bcaudan: I don't really mind but we could use `data: {action: CustomUserAction, context: Context}` if you prefer an object

packages/rum/src/rum.entry.ts
```diff
@@ -92,13 +92,15 @@ export function makeRumGlobal(startRumImpl: StartRum) {
...
-        type: UserActionType.CUSTOM,
-      })
+      addUserActionStrategy(
+        {
+          name,
+          context: combine({}, context),
+          startTime: performance.now(),
+          type: UserActionType.CUSTOM,
+        },
+        combine({}, globalContextManager.get())
```
bcaudan, line 102: as asked in https://github.com/DataDog/browser-sdk/pull/551#discussion_r501709021, do we need to copy the context in after init?

if not, what about passing global context only on before init and use the current global context as usual on after init?

I have a hard time legitimate the test `it('should ignore the current global context', () => {`

BenoitZugmeyer: You are right, we don't need to clone the global context after init. I can move this clone in the 'before init' `addUserActionStrategy`, everything should be fine.

packages/rum/test/rum.spec.ts
```diff
@@ -588,13 +539,37 @@ describe('rum user action', () => {
...
+        startTime: 123,
+        type: UserActionType.CUSTOM,
+      },
+      context: {},
     })
 
     expect((getRumMessage(server, 0) as any).fooBar).toEqual('foo')
   })
+
+  it('should ignore the current global context', () => {
```
bcaudan, line 554:
```suggestion
  it('should ignore the current global context when a saved global context is provided ', () => {
```

bcaudan (approved):

PR #553, opened 2020-10-07
[RUMF-708] rework e2e tests to allow more custom setup

BenoitZugmeyer:
## Motivation

Allow custom setups in E2E tests

## Testing

---


packages/core/test/configuration.spec.ts
```diff
@@ -20,31 +20,16 @@ describe('configuration', () => {
   })
 
   describe('endpoint overload', () => {
-    it('should not be available for production env', () => {
```
bcaudan, line 23: why not keeping this one?

BenoitZugmeyer: Because we can't override endpoints via the user configuration anymore

test/e2e/scenario/logs.scenario.ts
```diff
@@ -0,0 +1,74 @@
+import { createTest } from '../lib/createTest'
+import { browserExecute, browserExecuteAsync, flushBrowserLogs, flushEvents, withBrowserLogs } from '../lib/helpers'
+import { allSetups } from '../lib/setups'
+
+const UNREACHABLE_URL = 'http://localhost:9999/unreachable'
+
+describe('logs', () => {
+  createTest('send logs', allSetups({ logs: {} }), async ({ events }) => {
```
bcaudan, line 8: I am not super fan of the setup API:
- it seems that we always use all setups so we could hide that
- `logs: {}` could be less cryptic

what about having a setup options builder instead?
something like:

```
createTest('xxx', withRum().withLogs({ forwardErrorsToLogs: true }).withBody(...),
```

test/e2e/lib/servers.ts
```diff
@@ -0,0 +1,141 @@
...
+  base: Server
+  intake: Server
+  crossOrigin: Server
+}
+
+let memoizedServers: undefined | Servers
+
+export async function getTestServers() {
+  if (!memoizedServers) {
+    memoizedServers = {
```
bcaudan, line 30: how is it memoized?

test/e2e/lib/logger.ts
```diff
@@ -0,0 +1,15 @@
+import * as fs from 'fs'
+import * as path from 'path'
```
bcaudan, line 2: unused import

test/e2e/lib/builds.ts
```diff
@@ -0,0 +1,56 @@
...
+function replaceEndpoints(content: Buffer, endpoints: Endpoints) {
+  return bufferReplace(content, {
+    '<<< E2E INTERNAL MONITORING ENDPOINT >>>': endpoints.internalMonitoring,
+    '<<< E2E LOGS ENDPOINT >>>': endpoints.logs,
+    '<<< E2E RUM ENDPOINT >>>': endpoints.rum,
+  })
+}
+
+function bufferReplace(buffer: Buffer, replacements: { [needle: string]: string }): Buffer {
+  const replacementsArray = Object.entries(replacements).map(([needle, replacement]) => [
```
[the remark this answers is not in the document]
BenoitZugmeyer: Old habits https://www.php.net/manual/en/function.strpos :D I'll change it.

test/e2e/lib/eventsRegistry.ts
```diff
@@ -0,0 +1,68 @@
...
+export class EventRegistry {
+  private db: Entry[] = []
+
+  push(type: IntakeType, event: AnyServerEvent) {
+    this.db.push({ type, event })
+  }
+
+  get all() {
+    return this.db.map((entry) => entry.event)
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: This is usefull to debug, I use it sometimes to print the full list of events. I could add a `get length()` though

test/e2e/lib/builds.ts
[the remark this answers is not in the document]
bcaudan: half of the joke was missing 😆

test/e2e/lib/servers.ts
BenoitZugmeyer: Well, the result is cached so it'll return the same object every time. I can call it `serversSingleton` if you prefer?

test/e2e/scenario/logs.scenario.ts
BenoitZugmeyer: * I used `allSetups` on all existing scenarios so we have the same coverage as before. As we discussed, this can be chalenged and we may want to use all setups only for a few simple tests.

* I introduced a new test on `internalMonitoring` that uses only the `bundle` setup as it does not matter too much.

* The goal is allow very custom setups in tests (for example, tests on APM document trace will need very custom HTML). I don't want to support every setups through the `setups` module, its goal was to provide simple building blocks, and a few default setups.

I could rename `allSetups` to `defaultSetups`. Also, would `logs: true` be less cryptic?

test/e2e/lib/servers.ts
bcaudan: ok, [memoization](https://en.wikipedia.org/wiki/Memoization#Overview) has a different meaning, singleton would be more accurate here.

test/e2e/scenario/logs.scenario.ts
bcaudan: We could do subtle tweaks to improve it but I think the fluent API could do a better job to explain what is the setup while configuring it.

> I used allSetups on all existing scenarios so we have the same coverage as before. As we discussed, this can be chalenged and we may want to use all setups only for a few simple tests.
> I introduced a new test on internalMonitoring that uses only the bundle setup as it does not matter too much.

we could have either have:
- implicit all setups by defaut and a `onlyBundleSetup()` for internal monitoring
- explicit `onAllSetup()` for each test and `onBundleSetup()` for internal monitoring

since *all* does not bring much clarity, I am leaning toward implicit.

> The goal is allow very custom setups in tests (for example, tests on APM document trace will need very custom HTML). I don't want to support every setups through the setups module, its goal was to provide simple building blocks, and a few default setups.

I'm ok with the current setup capabilities, I don't suggest to support new ones

> Also, would logs: true be less cryptic?

I though about it but it would mess a bit the SetupOptions type

test/e2e/lib/eventsRegistry.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  get logs() {
+    return this.getEvents('logs')
+  }
+
+  get rum() {
+    return this.getEvents('rum')
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: I used a single list to keep the events order regardless of their type. By using multiple lists, we'll lose this information, but it should be fine since we don't have tests on this for now!

test/e2e/scenario/logs.scenario.ts
BenoitZugmeyer: Alright, if you want a fluent API, let's make a fluent API. What about a more general API then? This could allow some nice features in the future. Something like:

```ts
createTest('xxx')
  .withRum()
  .withSetup(bundleSetup) // optional, defaults to async/bundle/npm setups
  // .retries(2) // possible addition in the future
  // .only() // possible addition in the future, to use `fit` instead of `it`
  .run(async ({ events }) => {
  })
```

bcaudan: sounds good to me

test/e2e/lib/browserHelpers.ts
```diff
@@ -0,0 +1,107 @@
...
+    const cookies = document.cookie.split(';')
+    for (const cookie of cookies) {
+      const eqPos = cookie.indexOf('=')
+      const name = eqPos > -1 ? cookie.substr(0, eqPos) : cookie
+      document.cookie = `${name}=;expires=Thu, 01 Jan 1970 00:00:00 GMT;samesite=strict`
+    }
+  })
+}
+
+async function findSessionCookie() {
```
bcaudan, line 66: session methods should be in SDK helpers, no?

test/e2e/lib/servers.ts
```diff
@@ -0,0 +1,146 @@
+import * as http from 'http'
+import { AddressInfo } from 'net'
+import { getIp } from '../../utils'
+import { log } from './logger'
```
bcaudan, line 4: I have a got a reorganization proposal:

```
- scenario helpers
  - sdkHelpers
  - browserHelpers
  - globalTypes
- testSetup
  - createTest (+ merge declareTest)
  - pageSetups (rename of setups)
  - sdkBuilds
  - servers
    - servers
    - intakeServerApp
    - mockServerApp
  - common
    - logger
    - eventsRegistry
- serverTypes
```
does it make sense?

BenoitZugmeyer: This looks a bit weird to me.

* Why `globalTypes` is with the helpers? They are not helper functions.
* Modules inside `testSetup` are all the code used when setuping a test. But `servers` live accross tests, and the `waitServersIdle` function is used after test setup. I don't think it has its place in `testSetup`.
* The `logger` also have little to do with a single test setup, and it is used by the servers during the execution of all the spec.
* Why the "testSetup/common" folder?

test/e2e/lib/browserHelpers.ts
BenoitZugmeyer: 👍

test/e2e/lib/servers.ts
bcaudan: nice!

It still a bit hard to me to navigate across this code and I feel like regrouping similar things together could help:

> Why globalTypes is with the helpers? They are not helper functions.

maybe scenario helpers is not the right name, the idea is to regroup utility code for scenarios and isolate it from the "test framework" code 

> Modules inside testSetup are all the code used when setuping a test. But servers live accross tests, and the waitServersIdle function is used after test setup. I don't think it has its place in testSetup.

what about having testSetup (or other name) directory contain all the code related to our "test framework"?
We could then then better identify the main responsibilities:
- test API
- build test page
- build test sdk 
- test servers

> Why the "testSetup/common" folder?

to regroup some less important files used by several others, could be kept at the same level for now

BenoitZugmeyer: I moved things around again, now it looks like this:
```
test/e2e
├── browsers.conf.js
├── lib
│   ├── framework
│   │   ├── createTest.ts
│   │   ├── eventsRegistry.ts
│   │   ├── httpServers.ts
│   │   ├── index.ts
│   │   ├── logger.ts
│   │   ├── pageSetups.ts
│   │   ├── sdkBuilds.ts
│   │   └── serverApps
│   │       ├── intake.ts
│   │       └── mock.ts
│   ├── helpers
│   │   ├── browser.ts
│   │   └── sdk.ts
│   └── types
│       ├── global.ts
│       └── serverEvents.ts
├── scenario
│   ├── internalMonitoring.scenario.ts
│   ├── logs.scenario.ts
│   ├── rum
│   │   ├── actions.scenario.ts
│   │   ├── errors.scenario.ts
│   │   ├── resources.scenario.ts
│   │   └── views.scenario.ts
│   └── tracing.scenario.ts
├── ts-node.js
├── tsconfig.json
├── wdio.base.conf.js
├── wdio.bs.conf.js
└── wdio.local.conf.js

7 directories, 26 files
```
I hope it makes more sense :)

bcaudan (approved):
Awesome 👏

PR #555, opened 2020-10-09
Use the same options to test and set cookies

kichik:
Current code only uses some of the configured flags to set test cookies. This leads to cases where testing for cookies can fail even though setting cookies would eventually work.

## Motivation

Using RUM inside `<iframe>` didn't work, even though all the documented flags required for it were enabled.

## Changes

`areCookiesAuthorized()` uses `CookieOptions` instead of just `{ secure: useSecureCookie }`.

## Testing

Use RUM in `<iframe>` that's on a different domain than the main page.

---


bcaudan:
Hi @kichik,

Thanks for this PR, it is conflicting with some in progress work so we will try to integrate it in the coming days.

bcaudan (approved):

kichik:
Thanks! Any idea when 1.23.0 might be available on npm?

BenoitZugmeyer:
@kichik Thank you for your contribution! v1.23.0 has been published, and should be available on NPM right now.

PR #557, opened 2020-10-09
🐛[RUMF-716] fix invalid action name

bcaudan:
## Motivation

Some UTF-8 characters can be composed of a pair of "surrogate" characters.

more details:  [MDN#encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

## Changes

Ensure that our string truncate don't make invalid string by keeping only the first part of the pair.

## Testing

unit test

---

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -372,3 +372,16 @@ export function findCommaSeparatedValue(rawString: string, name: string) {
   const matches = rawString.match(`(?:^|;)\\s*${name}\\s*=\\s*([^;]+)`)
   return matches ? matches[1] : undefined
 }
+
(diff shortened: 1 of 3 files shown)


BenoitZugmeyer (commented):
Good catch!

I think that testing if the character at the ending position is part of a surrogate pair would be nicer.

Something like:
```ts
  // tslint:disable-next-line: no-bitwise
  if ((candidate.charCodeAt(length - 1) & 0xf800) === 0xd800) {
    return candidate.slice(0, length + 1)
  }
  return candidate.slice(0, length)
```

WDYT?

BenoitZugmeyer (approved):

PR #560, opened 2020-10-12
[Core] dereference `combine` sources recursively

BenoitZugmeyer:
## Motivation

Follow up of https://github.com/DataDog/browser-sdk/pull/551#discussion_r501556141
We figured that our `combine` implementation was not de-referencing all sub values. This could lead to unexpected behavior when some context is stored in an event to be pushed in the future (like: pre-init addUserAction or log) and a context sub-value is mutated in the meantime.

## Changes

* Change the `deepMerge` (and `combine`) implementation to always dereference their arguments
* Introduce a `deepClone` function to explicitly dereference a value.  

## Testing

Some unit tests have been added

---


packages/core/test/utils.spec.ts
```diff
@@ -33,9 +34,55 @@ describe('utils', () => {
...
+
+    it('should merge multiple objects', () => {
+      expect(combine({ a: 1 }, { b: 2 }, { c: 3 })).toEqual({ a: 1, b: 2, c: 3 })
+    })
+
+    it('should not keep references on objects', () => {
+      const source = { a: { b: 1 } }
+      const result = combine({}, source)
+      source.a.b = 2
+      expect(result.a.b).toBe(1)
```
bcaudan, line 50: In all the the tests,  where you want to test that references are different, what about actually test the references like:
```ts
expect(result.a).not.toBe(source.a)
```

instead of testing that a modification on the source is not propagated to the result.

packages/core/test/utils.spec.ts
```diff
@@ -33,9 +34,55 @@ describe('utils', () => {
...
+      source.a[0] = 2
+      expect(result.a[0]).toBe(1)
+    })
+  })
+
+  describe('deepClone', () => {
+    it('should return a similar value', () => {
+      const clonedValue = deepClone({ a: 1 })
+      expect(clonedValue).toEqual({ a: 1 })
+    })
```
bcaudan, line 65: what about `should return a result deeply equal to the source` instead?

packages/core/test/utils.spec.ts
```diff
@@ -33,9 +34,55 @@ describe('utils', () => {
...
+
+    it('should clone the value', () => {
+      const value = { a: 1 }
+      const clonedValue = deepClone(value)
+      value.a = 2
+      expect(clonedValue).toEqual({ a: 1 })
+    })
```
bcaudan, line 72: what about `should return a different reference` instead?
I think it could be clearer to be explicit about what clone mean in our tests description

packages/core/test/utils.spec.ts
```diff
@@ -33,9 +34,55 @@ describe('utils', () => {
...
+    it('should not keep references on objects', () => {
+      const source = { a: { b: 1 } }
+      const result = combine({}, source)
+      source.a.b = 2
+      expect(result.a.b).toBe(1)
+    })
+
+    it('should not keep references on arrays', () => {
+      const source = { a: [1] }
+      const result = combine({}, source)
```
bcaudan, line 55:
```suggestion
      const result = combine(source)
```
or is there a reason to keep this one?

BenoitZugmeyer: * `combine` expects two arguments, so we can't really remove the first `{}`. We could change its signature to accept only one argument but it looks weird to have a `combine` call with only one argument (how is it combining?)
* I think we should keep this test case: it highlights that both objects and arrays are cloned.

packages/core/test/utils.spec.ts
```diff
@@ -33,9 +34,50 @@ describe('utils', () => {
...
+    it('should return different references for objects sub values', () => {
+      const value = { a: { b: 1 } }
+      const clonedValue = deepClone(value)
+      expect(clonedValue.a).not.toBe(value.a)
+    })
+
+    it('should return different references for arrays items', () => {
+      const value = { a: [1] }
+      const clonedValue = deepClone(value)
+      expect(clonedValue.a).toEqual(value.a)
```
bcaudan, line 80:
```suggestion
      expect(clonedValue.a).not.toBe(value.a)
```

bcaudan (approved):

PR #564, opened 2020-10-14
 ✨ [RUMF-534] implement logs.onReady and rum.onReady 

BenoitZugmeyer:
## Motivation

Have a proper way to execute code when the SDK is ready.

## Changes

* Implement `logs.onReady` and `rum.onReady` to execute previously stored callbacks
* Change the e2e tests async setup to use the new official async snippet

## Testing

---

### DIFF
--- a/test/e2e/lib/framework/pageSetups.ts
+++ b/test/e2e/lib/framework/pageSetups.ts
@@ -31,22 +31,18 @@ export const DEFAULT_SETUPS = [
 export function asyncSetup(options: SetupOptions) {
   let body = options.body || ''
 
+  function formatSnippet(url: string, globalName: string) {
+    return `(function(r,a,w,l,s) {
+r=r[s]=r[s]||{q:[],onReady:function(c){r.q.push(c)}}
+s=a.createElement(w);s.async=1;s.src=l
+l=a.getElementsByTagName(w)[0];l.parentNode.insertBefore(s,l)
+})(window,document,'script','${url}','${globalName}')`
+  }
+
   if (options.logs) {
     body += html`
-      <script type="text/javascript">
-        window.addEventListener('load', () => {
-          const logs = document.createElement('script')
-          logs.src = './datadog-logs.js'
-          document.getElementsByTagName('head')[0].appendChild(logs)
-        })
-
-        window.DD_LOGS = window.DD_LOGS || {
-          onReady: function(c) {
-            DD_LOGS.q.push(c)
-          },
-          q: [],
-        }
-
+      <script>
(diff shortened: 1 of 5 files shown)


test/e2e/lib/framework/pageSetups.ts
```diff
@@ -37,11 +37,19 @@ export function asyncSetup(options: SetupOptions) {
...
+        window.DD_LOGS = window.DD_LOGS || {
+          onReady: function(c) {
+            DD_LOGS.q.push(c)
+          },
+          q: [],
+        }
+
+        DD_LOGS.onReady(function() {
+          DD_LOGS.init(${formatLogsOptions(options.logs)})
+        })
```
bcaudan, line 52: it could be nice to use a setup as close as possible to what we will provide to customers.
what do you think?

BenoitZugmeyer: I agree, but I thought this was what will be provided to customers? I took it from your document https://docs.google.com/document/d/1EsiF_TmXs0exzBWBeowUaupoAm8Ra5MlH6dLyHLkfNM/edit#heading=h.a5fyuuv435sa

bcaudan: oh ok, in this section of the doc only the buffer mechanism is described not the loading of the SDK.
It could be nice to have a compact snippet like in the other sections of the doc.

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -31,16 +31,20 @@ export const DEFAULT_SETUPS = [
 export function asyncSetup(options: SetupOptions) {
   let body = options.body || ''
 
+  function formatSnippet(url: string, globalName: string) {
+    return `(function(r,a,w,l,s) {
```
BenoitZugmeyer, line 35: "rawls" is a [The Wire reference](https://en.wikipedia.org/wiki/William_Rawls)

test/e2e/lib/framework/pageSetups.ts
BenoitZugmeyer: I changed the snippet, let me know what you think!

test/e2e/lib/framework/pageSetups.ts
bcaudan: nice one but I don't think that the wire references are publicly shared 🙂 
what about something dog related instead? `hound` would fit 😉

bcaudan (approved):

test/e2e/lib/framework/pageSetups.ts
BenoitZugmeyer: Ok for `hound` :)

PR #565, opened 2020-10-14
✅[RUMF-732] use stub instead or real xhr in unit test

bcaudan:
## Motivation

unit test on bs are brittle because requests to karma middleware are randomly not completing

## Changes

- use xhr stub to avoid any issue with real requests during unit tests
- add seed reporter to ease ci failure reproduction

## Testing

automated tests

---

### DIFF
--- a/packages/core/src/specHelper.ts
+++ b/packages/core/src/specHelper.ts
@@ -168,17 +168,28 @@ class StubXhr {
     this.readyState = XMLHttpRequest.DONE
     this.onreadystatechange()
     if (status >= 200 && status < 500) {
-      this.element.dispatchEvent(new Event('load'))
+      this.element.dispatchEvent(createNewEvent('load'))
     }
     if (status >= 500) {
-      this.element.dispatchEvent(new Event('error'))
+      this.element.dispatchEvent(createNewEvent('error'))
     }
-    this.element.dispatchEvent(new Event('loadend'))
+    this.element.dispatchEvent(createNewEvent('loadend'))
   }
 
   addEventListener(name: string, callback: () => void) {
     this.element.addEventListener(name, callback)
   }
+}
+
+function createNewEvent(eventName: string) {
+  let event
+  if (typeof Event === 'function') {
+    event = new Event(eventName)
+  } else {
+    event = document.createEvent('Event')
+    event.initEvent(eventName, true, true)
+  }
+  return event
 }
 
 export function stubXhr() {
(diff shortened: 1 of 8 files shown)


packages/core/test/xhrProxy.spec.ts
```diff
@@ -119,31 +127,13 @@ describe('xhr proxy', () => {
     })
   })
 
-  it('should track successful sync request', (done) => {
```
BenoitZugmeyer, line 122: This test isn't needed anymore?

packages/core/src/specHelper.ts
```diff
@@ -141,11 +141,74 @@ export class PerformanceObserverStubBuilder {
...
+class StubXhr {
+  public response: string | undefined = undefined
+  public status: number | undefined = undefined
+  public readyState: number = XMLHttpRequest.UNSENT
+  public onreadystatechange: () => void = noop
+
+  private element: HTMLDivElement
+
+  constructor() {
+    this.element = document.createElement('div')
```
BenoitZugmeyer, line 153: This looks a bit weird, what about using a small event emiter implementation?

BenoitZugmeyer (commented):
What about using dependency injection?

we could have something like this
```ts
// in xhrProxy.ts
function startXhrProxy() {
  if (!xhrProxySingleton) {
   xhrProxySingleton = proxyXhr(XMLHttpRequest);
  }
  return xhrProxySingleton
}

function proxyXhr(XHRClass: typeof XMLHttpRequest) {
  const beforeSendCallbacks = []
  const onRequestCompleteCallbacks = []

  XHRClass.prototype.send = ...

  [...]
  return {
      beforeSend(callback: (context: XhrStartContext, xhr: XMLHttpRequest) => void) {
        beforeSendCallbacks.push(callback)
      },
      onRequestComplete(callback: (context: XhrCompleteContext) => void) {
        onRequestCompleteCallbacks.push(callback)
      },
    }
}

// In tests
const proxy = proxyXhr(StubXhr)
...
```

I think it would be easier to follow.

packages/core/test/xhrProxy.spec.ts
bcaudan: I don't see how we could replicate this behavior with a stub.
However, FMU it does not change our proxy behavior, so it seemed fine to me to drop it.

packages/core/src/specHelper.ts
bcaudan: does https://github.com/DataDog/browser-sdk/pull/565/commits/9d11fec5ac398ec0164298d61b49ef697f2e4db4 clarify it a bit?
the idea was to delegate xhr event capabilities to an other object with such features to avoid redoing it ourself.
If you see any issue with that, I can switch implementation.

BenoitZugmeyer (approved):

PR #570, opened 2020-10-16
✨ [RUMF-727] introduce v2 format 

bcaudan:
## Motivation

Start using v2 format and introduce format validation

## Changes

- Introduce `v2_format` feature flag
- Duplicate types for V2
- Use dedicated events for v2
- Add validation for v2 events generated in unit tests
- Handle V2 long tasks

## Testing

Automated tests

---


BenoitZugmeyer (approved):
Changes look good! I think it would have been better to rename V1 functions/types/events to add `v1` in their name, instead of adding new ones with `v2` (like I did on web-ui). This way, the migration cleanup would be easier. But that's fine now.

PR #573, opened 2020-10-19
:bug: [RUMF-742] fix cookie creation domain when trackSessionAcrossSubdomains: true

BenoitZugmeyer:
## Motivation

The function `getCurrentSite` can yield invalid result when a SDK is initialized multiple times within 1 second.

## Changes

* Cache `getCurrentSite` result
* Use an unique cookie name for `getCurrentSite`

## Testing

Setup the RUM SDK on a test page, and print its internal context. Beforehand, clear any cookie on the current domain. Then, refresh the page quickly 2 times: the session should stay the same, and the cookie should be created with a valid domain.

---

### DIFF
--- a/packages/core/src/cookie.ts
+++ b/packages/core/src/cookie.ts
@@ -1,4 +1,4 @@
-import { findCommaSeparatedValue, ONE_SECOND } from './utils'
+import { findCommaSeparatedValue, generateUUID, ONE_SECOND } from './utils'
 
 export const COOKIE_ACCESS_DELAY = ONE_SECOND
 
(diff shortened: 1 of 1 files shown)


packages/core/src/cookie.ts
```diff
@@ -77,15 +79,21 @@ export function areCookiesAuthorized(options: CookieOptions): boolean {
  * strategy: find the minimal domain on which cookies are allowed to be set
  * https://web.dev/same-site-same-origin/#site
  */
+let getCurrentSiteCache: string | undefined
 export function getCurrentSite() {
-  const testCookieName = 'dd_site_test'
-  const testCookieValue = 'test'
+  if (getCurrentSiteCache === undefined) {
```
bcaudan, line 84: what about turning it into an early return?

bcaudan (approved):

packages/core/src/cookie.ts
bcaudan: not really important, not changing the indentation would have make your commit easier to review

BenoitZugmeyer: Hiding whitespace changes helps in this case https://github.com/DataDog/browser-sdk/pull/573/files?w=1

PR #574, opened 2020-10-19
✨ [RUMF-724] implement API to capture an error 

BenoitZugmeyer:
## Motivation

Add a way to manually capture errors

## Changes

Implement a new `DD_RUM.addError` to manually submit an error:

```ts
DD_RUM.addError(error: unknown, context?: Context, source: ErrorSource = ErrorSource.CUSTOM)
```

## Testing

Unit tests + manual testing on rum playground

## To specify

* [x] which error origin should we use? should it be specified by the user?
* [x] is the function name aligned with other SDKs?
* [x] should we be able to pass a 'context' argument? 

---


packages/core/src/errorCollection.ts
```diff
@@ -13,12 +13,14 @@ export interface ErrorMessage {
     error: ErrorContext
     http?: HttpContext
   }
+  customerContext?: Context
+  savedGlobalContext?: Context
```
BenoitZugmeyer, line 17: I didn't want to modify the `ErrorMessage` too much, but if we follow the same logic used in `addUserAction`, we should [separate the `savedGlobalContext` from other props](https://github.com/DataDog/browser-sdk/blob/master/packages/rum/src/domain/lifeCycle.ts#L42-L45) like:

```ts
interface ErrorMessageWithGlobalContext {
  errorMessage: ErrorMessage,
  context?: Context
}
```

IMO, storing the global context in the error message is fine, and multiplying structure definitions makes it harder to understand.

packages/core/src/errorCollection.ts
```diff
@@ -27,26 +29,45 @@ export interface HttpContext {
...
+  source: ErrorSource
+}
+
+export enum InternalErrorSource {
   AGENT = 'agent',
   CONSOLE = 'console',
+  LOGGER = 'logger',
+}
+
+export enum ErrorSource {
```
BenoitZugmeyer, line 45: I intentionaly kept the `ErrorSource` name instead of having a `PublicErrorSource` enum, because this name is exported publicly (as a JS variable but also as a TS type).

packages/core/src/errorCollection.ts
```diff
@@ -118,26 +150,39 @@ export function stopRuntimeErrorTracking() {
...
-  if (stackTrace.message === undefined && !(errorObject instanceof Error)) {
-    message = `Uncaught ${jsonStringify(errorObject)}`
-    stack = 'No stack, consider using an instance of Error'
-  } else {
-    message = stackTrace.message || 'Empty message'
-    stack = toStackTraceString(stackTrace)
+export function makeAddError(errorObservable: ErrorObservable) {
+  return (addedError: AddedError, savedGlobalContext?: Context) => {
+    const stackTrace = addedError.error instanceof Error ? computeStackTrace(addedError.error) : undefined
+    const { message, stack, kind } = formatUnknownError(stackTrace, addedError.error, 'Captured')
```
BenoitZugmeyer, line 156: I used 'Captured' here, but I'm not convinced about it. I don't really like 'Added', but I'm open to alternatives.

packages/core/src/domain/errorCollection.ts
```diff
@@ -118,26 +150,39 @@ export function stopRuntimeErrorTracking() {
...
+          kind,
+          stack,
+          origin: addedError.source,
+        },
+      },
+      customerContext: addedError.context,
+      startTime: addedError.startTime,
+    })
+  }
+}
```
bcaudan, line 171: two concerns about having this logic in shared error collection:
- since it is a rum feature, any reason to have it here? we could just export the logic that process the error and keep this component about automatic error collection only
- FMU, in a npm setup, error collection instance is shared between rum and logs, so rum added error could be retrieved by the default logger

BenoitZugmeyer: > since it is a rum feature, any reason to have it here? we could just export the logic that process the error and keep this component about automatic error collection only

It depends if we want to use the error frequency limit in the same way as other errors. If we move the code in RUM, it would be a bit more difficult to reuse the same limit strategy.

> FMU, in a npm setup, error collection instance is shared between rum and logs, so rum added error could be retrieved by the default logger

Isn't this an expected behavior if the log `forwardErrorsToLog` option is enabled? But you are right, this won't work on non-NPM setups.

bcaudan: interesting questions!
About limit strategy, I'd say that the similar manual usage on the logs side (logger.error) is not limited so I'd be fine to keep the limiting strategy to automatic error collection for now. Maybe at some point adapting the limitation strategy if needed.
About `forwardErrorToLogs`, IMO it only concerns automatic error collection so I'd keep the two separated. It could worth discussing this point with @hdelaby though.

hdelaby: My 2 cents: it's better to send the RUM errors with logs (after all they're front-end errors too) and then users will filter them in their log pipelines rather than having to use 2 APIs side by side if you want to have the log both in logs AND rum.

bcaudan: it sounds fair.
since we also have a task to send log errors in RUM (RUMF-741), I'd be in favor of either:

- not send RUM errors with logs for now but handle it in RUMF-741
- tackle RUMF-741 now

IMO, tackling both at the same time could help us to handle it properly

BenoitZugmeyer: Agreed. I'll move the addError code in RUM then!

bcaudan: Nice, I'll update the the backlog task accordingly

BenoitZugmeyer:
Superseded by https://github.com/DataDog/browser-sdk/pull/585

PR #585, opened 2020-10-22
✨ [RUMF-724] implement API to capture an error 

BenoitZugmeyer:
## Motivation

Add a way to manually capture errors

This new PR supersed https://github.com/DataDog/browser-sdk/pull/574 . Differences:

* the implementation is now in RUM instead of part of the Core `errorCollection`
* the `ErrorSource` is not split in Internal/Public enums. This was a bit confusing, and the full `ErrorSource` should be public anyway since the RUM types are [exported publicly](https://github.com/DataDog/browser-sdk/blob/master/packages/rum/src/index.ts#L5). The `addError` use an union of supported `ErrorSource` values instead.
* the implementation follows a pattern very similar to the `addUserAction`

## Changes

Implement a new `DD_RUM.addError` to manually submit an error:

```ts
DD_RUM.addError(error: unknown, context?: Context, source: ErrorSource = ErrorSource.CUSTOM)
```

## Testing

Unit tests + manual testing on rum playground

---


packages/rum/src/boot/rum.ts
```diff
@@ -198,6 +206,36 @@ function trackErrors(lifeCycle: LifeCycle, handler: (startTime: number, event: R
...
 
+function trackManualErrors(
+  lifeCycle: LifeCycle,
+  handler: (startTime: number, event: RumErrorEvent, savedGlobalContext?: Context, customerContext?: Context) => void
+) {
+  lifeCycle.subscribe(
+    LifeCycleEventType.MANUAL_ERROR_COLLECTED,
+    ({ error: { error, startTime, context: customerContext, source }, context: savedGlobalContext }) => {
+      const stackTrace = error instanceof Error ? computeStackTrace(error) : undefined
+      const { message, stack, kind } = formatUnknownError(stackTrace, error, 'Captured')
```
BenoitZugmeyer, line 217: I used 'Captured' here, but I'm not convinced about it. I don't really like 'Added', but I'm open to alternatives.

bcaudan: `Manual`, `Custom`?

packages/rum/src/domain/lifeCycle.ts
```diff
@@ -1,13 +1,14 @@
...
 import { RawRumEventV2, RumEventV2 } from '../typesV2'
 import { RequestCompleteEvent, RequestStartEvent } from './rumEventsCollection/requestCollection'
 import { AutoActionCreatedEvent, AutoUserAction, CustomUserAction } from './rumEventsCollection/userActionCollection'
 import { View, ViewCreatedEvent } from './rumEventsCollection/viewCollection'
 
 export enum LifeCycleEventType {
   ERROR_COLLECTED,
+  MANUAL_ERROR_COLLECTED,
   PERFORMANCE_ENTRY_COLLECTED,
   CUSTOM_ACTION_COLLECTED,
```
bcaudan, line 13: Why not `CUSTOM_ERROR_COLLECTED`? it would be more similar to action

BenoitZugmeyer: I think it could be confused with `ErrorSource.CUSTOM` (manual errors can have various sources)...

packages/rum/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { Datacenter } from '@datadog/browser-core'
+export { Datacenter, ErrorSource } from '@datadog/browser-core'
```
bcaudan, line 1: should not we also exposed the `ManuallyAddedErrorSource` type used by addError?
from a user perspective it is a mouthful, it could be nicer to either:
- rename it to ErrorSource and prefix the core one with internal
- rename it to RumErrorSource

packages/rum/src/types.ts
```diff
@@ -129,3 +129,12 @@ export interface InternalContext {
     id: string
   }
 }
+
+export type ManuallyAddedErrorSource = ErrorSource.CUSTOM | ErrorSource.SOURCE | ErrorSource.NETWORK
```
bcaudan, line 133: from a user perspective it is a mouthful, it could be nicer to either:
- rename it to ErrorSource and prefix the core one with internal
- rename it to RumErrorSource

packages/rum/src/boot/rum.ts
```diff
@@ -198,6 +206,36 @@ function trackErrors(lifeCycle: LifeCycle, handler: (startTime: number, event: R
   })
 }
 
+function trackManualErrors(
```
bcaudan, line 209: it could have been nice to put this logic directly in an errorCollection file, to avoid to move it and its test with RUMF-740

packages/rum/src/types.ts
BenoitZugmeyer: Since the ErrorSource is used by the RumEvent type exported publicly, I don't think we should rename it "internal" (it is user facing). Also, adding the `Rum` prefix would be confusing, because we would have two user facing types with a slightly different name.

Instead of having a type for this, we could simply change the addError signature to list the possible source values like:

```ts
addError(error: unkwown, context?: Context, source: ErrorSource.CUSTOM | ErrorSource.SOURCE | ErrorSource.NETWORK)
```

WDYT?

bcaudan: LGTM

packages/rum/src/boot/rum.ts
BenoitZugmeyer: Done

packages/rum/src/boot/rum.ts
BenoitZugmeyer: `Manual { foo: bar }` seems weird. Those errors may not be `Custom` and are not `RUM Error`... Maybe we could use `Caught { foo: bar }` (in contrast to `Uncaught { foo: bar }`)? Or  we could be a bit more verbose:`Manually caught ${errorSource} error { foo: bar }`

bcaudan: wdyt of `Provided { foo: bar }`?

bcaudan (approved):

PR #586, opened 2020-10-22
✨[RUMF-738] migrate view to v2 format (experimental)

bcaudan:
## Motivation

Following #570, migrating view to v2 format

## Changes

- introduce view collection
- move some rum spec to appropriate test files
- handle view in v2

## Testing

unit tests, staging

---


packages/rum/src/boot/rum.spec.ts
```diff
@@ -10,7 +10,7 @@ import {
   SESSION_KEEP_ALIVE_INTERVAL,
   THROTTLE_VIEW_UPDATE_PERIOD,
   View,
-} from '../domain/rumEventsCollection/view/viewCollection'
+} from '../domain/rumEventsCollection/view/trackViews'
```
bcaudan: This naming is not really satisfying since trackXxx and xxxCollection means pretty much the same thing 😕 
I'd be willing to go with that for now and wait the end of the migration to have a better overview and see what we can do.

packages/rum/src/boot/rum.spec.ts
```diff
@@ -114,6 +83,34 @@ describe('rum session', () => {
     expect(subsequentRequests[0].view.id).not.toEqual(initialRequests[0].view.id)
   })
 
+  it('when the session is renewed, a new view event should be sent', () => {
```
bcaudan, line 86: Some of the remaining tests are integration tests, it could be interesting to see what we want to do with them at some point.

packages/rum/src/domain/rumEventsCollection/view/viewCollection.ts
```diff
@@ -0,0 +1,80 @@
+import { Configuration, getTimestamp, msToNs } from '@datadog/browser-core'
+import { RumEventCategory, RumViewEvent } from '../../../types'
+import { RumEventType, RumViewEventV2 } from '../../../typesV2'
+import { LifeCycle, LifeCycleEventType } from '../../lifeCycle'
+import { trackViews, View } from './trackViews'
+
+export function startViewCollection(lifeCycle: LifeCycle, configuration: Configuration, location: Location) {
```
BenoitZugmeyer, line 7: I think the `location` argument is always `window.location` now. We need to keep the `location` argument on `trackViews`, but we could remove this one.

BenoitZugmeyer (approved):

packages/rum/src/domain/rumEventsCollection/view/viewCollection.ts
bcaudan: it is still used by some [rum.ts spec](https://github.com/DataDog/browser-sdk/blob/master/packages/rum/src/boot/rum.spec.ts#L273)

BenoitZugmeyer: Ah, right!

PR #592, opened 2020-10-27
✨[RUMF-740] migrate error to v2 format (experimental)

bcaudan:
## Motivation

Following #570, migrating error to v2 format

## Changes

- mutualize collection start for provided and auto error
- handle error in v2
- use internal unstructured error format to ease usage across logs, rum V1 and rum v2

## Testing

unit tests, staging

---

### DIFF
--- a/packages/core/src/domain/errorCollection.ts
+++ b/packages/core/src/domain/errorCollection.ts
@@ -12,9 +12,11 @@ export interface RawError {
   type?: string
   stack?: string
   source: ErrorSource
-  url?: string
-  statusCode?: number
-  method?: string
+  resource?: {
+    url: string
+    statusCode: number
+    method: string
+  }
 }
 
 export enum ErrorSource {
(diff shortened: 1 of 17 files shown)


packages/core/src/domain/errorCollection.ts
```diff
@@ -6,25 +6,15 @@ import { Configuration, isIntakeRequest } from './configuration'
...
-
-export interface HttpContext {
-  url: string
-  status_code: number
-  method: string
+  source: ErrorSource
+  url?: string
+  statusCode?: number
+  method?: string
 }
```
BenoitZugmeyer, line 18: I'm not sure to understand the motivation behind this change. `ErrorMessage` was convenient to assemble V1 events, now `RawError` is inconvenient for both V1 and V2 events. Also, `ErrorMessage` was a bit stricter on types (ex: the `HttpContext` made sure to be populated with all `url`, `status_code` and `method`).

bcaudan: having an internal format for errors allow to have at some point the same structure for provided and auto error to mutualize the building of the v1 or v2 format between these errors. 
I flatten ErrorContext because properties were not consistently in this object and did not make sense in a product perspective. Keeping a structure for resource related properties could still make sens though.
I can go back on that.

BenoitZugmeyer (approved):
LGTM

PR #593, opened 2020-10-28
✨[RUMF-518] migrate internal context to v2 format (experimental)

bcaudan:
## Motivation

Following #570, migrating internal context to v2 format

## Changes

- extract to dedicated file
- handle v2
- keep v1 properties for logs for now

## Testing

unit tests, staging

---

### DIFF
--- a/packages/logs/src/boot/logs.ts
+++ b/packages/logs/src/boot/logs.ts
@@ -128,5 +128,18 @@ interface Rum {
 
 function getRUMInternalContext(startTime?: number): Context | undefined {
   const rum = (window as any).DD_RUM as Rum
-  return rum && rum.getInternalContext ? rum.getInternalContext(startTime) : undefined
+  const context = rum && rum.getInternalContext ? rum.getInternalContext(startTime) : undefined
+  if (!context || !context.session) {
+    return context
+  }
+  // TODO settle on integration strategy
+  // tslint:disable:no-unsafe-any
+  const v2context = context as any
+  v2context.session_id = v2context.session.id
+  v2context.application_id = v2context.application.id
+  if (v2context.action) {
+    v2context.user_action = { id: v2context.action.id }
+  }
+  // tslint:enable:no-unsafe-any
+  return v2context as Context
 }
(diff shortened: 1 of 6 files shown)


packages/logs/src/boot/logs.ts
```diff
@@ -128,5 +128,18 @@ interface Rum {
...
+    return context
+  }
+  // TODO settle on integration strategy
+  // tslint:disable:no-unsafe-any
+  const v2context = context as any
+  v2context.session_id = v2context.session.id
+  v2context.application_id = v2context.application.id
+  if (v2context.action) {
+    v2context.user_action = { id: v2context.action.id }
+  }
```
BenoitZugmeyer, line 142: Are we sure that sending merged V1 & V2 format won't cause any issue in logs/trace?

packages/rum/test/specHelper.ts
```diff
@@ -174,6 +178,12 @@ export function setup(): TestSetupBuilder {
       })
       return setupBuilder
     },
+    withInternalContext() {
+      buildTasks.push(() => {
+        internalContext = startInternalContext(FAKE_APP_ID, session, parentContexts)
```
BenoitZugmeyer: What's the threshold between using `beforeBuild` and adding new domain-specific methods in the `specHelper`? I feel that if the function to test (here `startInternalContext`) is used in a single spec file, we shouldn't add it to the common `specHelper`.

packages/logs/src/boot/logs.ts
bcaudan: only view.id is used by app queries on logs or trace and it has not changed.
keeping v1 attributes is only to not break potential dashboards, monitors, etc... while we figure out what we want to do here. Adding v2 attributes should not break anything.
It will stay experimental as long as the strategy is not decided.

packages/rum/test/specHelper.ts
bcaudan: yes, that is not clear for now.
Here, I chose that way to avoid to add parentContexts to beforeBuild parameters.
I'd be in favor of having a look at that at the same time that we have a look at remaining rum.ts test, to see which behavior we want to keep.

BenoitZugmeyer (approved):

PR #596, opened 2020-10-29
[RUMF-748] use RAW_RUM_EVENT_COLLECTED to compute event counts

BenoitZugmeyer:
## Motivation

The recent introduction of RAW_RUM_EVENT_COLLECTED lifecycle events allows to simplify the event counts tracking.

## Changes

* Adjust view measures tests a bit
* Use RAW_RUM_EVENT_COLLECTED and RAW_RUM_EVENT_V2_COLLECTED instead of various events in trackEventCounts
* Remove the now unused RESOURCE_ADDED_TO_BATCH event
* Replace the ERROR_COLLECTED event by listening directly to the raw error observable.

## Testing

Unit and e2e tests

---

### DIFF
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -1,8 +1,6 @@
 import { areCookiesAuthorized, CookieOptions } from '../browser/cookie'
 import { buildConfiguration, UserConfiguration } from '../domain/configuration'
-import { RawError, startErrorCollection } from '../domain/errorCollection'
 import { setDebugMode, startInternalMonitoring } from '../domain/internalMonitoring'
-import { Observable } from '../tools/observable'
 
 export function makeGlobal<T>(stub: T): T & { onReady(callback: () => void): void } {
   const global = {
@@ -58,14 +56,12 @@ export interface BuildEnv {
   sdkVersion: string
 }
 
-export function commonInit(userConfiguration: UserConfiguration, buildEnv: BuildEnv, isCollectingError: boolean) {
+export function commonInit(userConfiguration: UserConfiguration, buildEnv: BuildEnv) {
   const configuration = buildConfiguration(userConfiguration, buildEnv)
   const internalMonitoring = startInternalMonitoring(configuration)
-  const errorObservable = isCollectingError ? startErrorCollection(configuration) : new Observable<RawError>()
 
   return {
     configuration,
-    errorObservable,
     internalMonitoring,
   }
 }
--- a/packages/rum/src/domain/rumEventsCollection/error/errorCollection.ts
+++ b/packages/rum/src/domain/rumEventsCollection/error/errorCollection.ts
@@ -8,7 +8,7 @@ import {
   getTimestamp,
   Observable,
   RawError,
-  startErrorCollection as startRawErrorCollection,
+  startAutomaticErrorCollection,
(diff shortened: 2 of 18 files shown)


packages/core/src/boot/init.ts
```diff
@@ -58,14 +56,12 @@ export interface BuildEnv {
   sdkVersion: string
 }
 
-export function commonInit(userConfiguration: UserConfiguration, buildEnv: BuildEnv, isCollectingError: boolean) {
+export function commonInit(userConfiguration: UserConfiguration, buildEnv: BuildEnv) {
```
bcaudan, line 59: any reason to remove it from it?
passing the errorObservable to rum error collection could have been enough

packages/rum/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -6,7 +6,9 @@ import {
   ErrorSource,
   formatUnknownError,
   getTimestamp,
+  Observable,
   RawError,
+  startErrorCollection as startRawErrorCollection,
```
bcaudan, line 11: the naming collision is not ideal.
If we want to settle on different names between rum and core, I'd propose automatic error collection for core, since it how we refer to it on the logs side

packages/core/src/boot/init.ts
BenoitZugmeyer: Since the Core error collection is only used by the RUM error collection (in RUM), there is no reason for exposing the error observable in `rum.ts`. It makes things a bit simpler to follow.

`doStartErrorCollection` is following the pattern we've been using to inject dependencies for unit tests, do you think it should be reconsidered?

bcaudan: > doStartErrorCollection is following the pattern we've been using to inject dependencies for unit tests

are you referring to `doStartLogs`?

BenoitZugmeyer: Yes

packages/rum/src/domain/rumEventsCollection/error/errorCollection.ts
BenoitZugmeyer: Ok to rename `core/domain/errorCollection` to `core/domain/automaticErrorCollection`. Should we rename `RawError` to `AutomaticError`? It would be more explicit.

packages/core/src/boot/init.ts
bcaudan: IMO, we would probably not need to do that for logs if we split logs.ts like we do with rum.ts.
Anyway, I don't think it qualifies as a pattern if it is used once.

I understand the idea of putting all the error collection stuff at the same place but since we are testing rum error collection without the core error collection, it seems that it is what we want to work with.

By starting the core error collection from rum, we would still pass RUM error collection dependencies as arguments instead of having one imported.

rum.ts responsibility seems already to wire everything together.

I don't have strong opinion on keeping core error collection start in common init or not but it seems more consistent to me if we pass dependencies as arguments as much as possible.

BenoitZugmeyer: > Anyway, I don't think it qualifies as a pattern if it is used once
This has been used multiple times in the past.

Alright, I'll revert my changes.

bcaudan (approved):

PR #605, opened 2020-11-04
✅ [RUMF-759] cleanup RUM unit tests specHelper

BenoitZugmeyer:
## Motivation

The line between what should be in specHelper and what should not is blurry. Most of domain-related `withXXX` methods are only used by the related domain spec file. It seems that `beforeBuild` is a good way to start a given function based on a testing context (session, configuration, etc.).

## Changes

This PR tries to simplify the unit test specHelper by reducing its responsabilities:
* the `withFakeServer` is replaced with a spy on `RUM_EVENT_COLLECTED` lifecycle events
* a `withConfiguration` method has been added to the spec helper to allow modifying the configuration before any `beforeBuild` call
* the `beforeBuild` method has been extended to expose more context to the callback, and allow to specify a cleanup function
* domain-specific `withXXX` methods have been inlined to their related spec by using `beforeBuild`.

## Testing

---

### DIFF
--- a/packages/rum/src/domain/rumEventsCollection/view/trackViews.spec.ts
+++ b/packages/rum/src/domain/rumEventsCollection/view/trackViews.spec.ts
@@ -78,8 +78,7 @@ describe('rum track url change', () => {
           initialViewId = id
           subscription.unsubscribe()
         })
-        const { stop } = trackViews(location, lifeCycle)
-        return stop
+        return trackViews(location, lifeCycle)
       })
     createSpy = jasmine.createSpy('create')
   })
@@ -217,8 +216,7 @@ describe('rum view referrer', () => {
           initialViewCreatedEvent = event
           subscription.unsubscribe()
         })
-        const { stop } = trackViews(location, lifeCycle)
-        return stop
+        return trackViews(location, lifeCycle)
       })
     createSpy = jasmine.createSpy('create')
   })
@@ -285,8 +283,7 @@ describe('rum track renew session', () => {
           initialViewId = id
           subscription.unsubscribe()
         })
-        const { stop } = trackViews(location, lifeCycle)
-        return stop
+        return trackViews(location, lifeCycle)
       })
   })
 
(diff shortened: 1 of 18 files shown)


packages/rum/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -451,8 +460,11 @@ describe('rum view measures', () => {
 
     setupBuilder = setup()
       .withFakeLocation('/foo')
-      .withViewCollection()
-      .beforeBuild((lifeCycle) => lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, handler))
+      .beforeBuild(({ location, lifeCycle }) => {
+        lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, handler)
+        const { stop } = trackViews(location, lifeCycle)
+        return stop
+      })
```
bcaudan, line 467: For the cleanup, I think could be more explicit or more implicit.

explicit: remove the logic from the helper and it makes more clear what we want to do with stop in the test
```js
      beforeBuild(({ location, lifeCycle, cleanupTasks }) => {
        lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, handler)
        const { stop } = trackViews(location, lifeCycle)
        cleanupTasks.push(stop)
      })
```

implicit: handle by default returned object with stop since it is seems to be what we mostly use
```js
      beforeBuild(({ location, lifeCycle, cleanupTasks }) => {
        lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, handler)
        return trackViews(location, lifeCycle)
      })
```

no strong opinion on it, wdyt?

BenoitZugmeyer: Returning an object with `stop` looks good to me

bcaudan (approved):

Issue #589, opened 2020-10-23
Feature request: print log context to console when the handler is "console"


dantheman39:
Hello! When developing locally I set the log handler to "console", to not spam the live site.

```typescript
datadogLogs.logger.setHandler(
  process.env.DD_ENABLED === "true"
    ? HandlerType.http
    : HandlerType.console
);
```

The downside to this is that I can't see the context values of the log, which makes this much less useful. This call to the logger:

```typescript
datadogLogs.logger.info("Switching viewed account", {
  from: oldAccount,
  to: newAccount,
});
```

Results in this log to the console:

<img width="253" alt="Screen Shot 2020-10-23 at 11 20 47 AM" src="https://user-images.githubusercontent.com/10103957/97022396-0f052980-1522-11eb-88f0-621f1ae09ab2.png">

I can't see the "from" and "to" values I'm sending along, which makes these logs much less useful for debugging (if the handler were "http", I could see these values in the network tab). I end up having to go into the code and setting a `console.log` if I want to debug things, whereas I would hope that datadog's logger would make that unnecessary.

Either way I hope you all are well, thank you!

bcaudan:
This is available in browser-sdk@1.25.4

PR #613, opened 2020-11-09
✨ [RUMF-762] include the context when using console handler

webNeat:
## Motivation

Fixing https://github.com/DataDog/browser-sdk/issues/589

## Changes

The context is now included when using the `console` handler. It's passed as the second parameter to `console.log` call instead of calling it twice.

## Testing

I updated the logs unit tests to check that the context is included.

---

### DIFF
--- a/packages/logs/src/domain/logger.ts
+++ b/packages/logs/src/domain/logger.ts
@@ -52,7 +52,7 @@ export class Logger {
           })
           break
         case HandlerType.console:
-          console.log(`${status}: ${message}`)
+          console.log(`${status}: ${message}`, combine(this.contextManager.get(), messageContext))
           break
         case HandlerType.silent:
           break
--- a/packages/logs/src/boot/logs.entry.spec.ts
+++ b/packages/logs/src/boot/logs.entry.spec.ts
@@ -271,7 +271,10 @@ describe('logs entry', () => {
         logger.error('message')
 
         expect(sendLogsSpy).not.toHaveBeenCalled()
-        expect(console.log).toHaveBeenCalledWith('error: message')
+        expect(console.log).toHaveBeenCalledWith('error: message', {
+          error: { origin: 'logger' },
+          logger: { name: 'foo' },
+        })
       })
 
       it('should have their name in their context', () => {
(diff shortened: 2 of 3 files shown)


packages/logs/src/domain/logger.ts
```diff
@@ -52,7 +52,7 @@ export class Logger {
           })
           break
         case HandlerType.console:
-          console.log(`${status}: ${message}`)
+          console.log(`${status}: ${message}`, combine(this.contextManager.get(), messageContext))
```
bcaudan, line 55: We could think about including the [other attributes that should end up in the log eventually](https://github.com/DataDog/browser-sdk/blob/master/packages/logs/src/boot/logs.entry.ts#L113).
However, this seems to be the simplest thing to do for now and having the other attributes could also unnecessarily bloat the console context.
@BenoitZugmeyer, @hdelaby any opinion on that?

BenoitZugmeyer: I think user-defined context is enough here.

* `date` is the current timestamp in millisecond: not very user friendly.
* `view.url` is (in general) the URL shown in the browser URL bar.
* `view.referrer` is (in general) an information known by the developer since it's their navigation, and this information coulb be useful in very particular situations.

hdelaby: I am not sure to understand what's at stake here. Can you elaborate a bit? 🙏

bcaudan: With this PR current implementation, when logs are send to the console, the displayed context:

- contains the log context
- contains the logger context
- does not contain global context
- does not contain page, session, date info
- does not contain RUM internal context

are we happy with this behavior or is there some part that we would want to be present in the displayed context?

bcaudan: as seen together ok for now, and we'll add APIs to retrieve current global context.

bcaudan (approved):
👍

trestletech:
Thanks for doing this! I was hunting through the repo looking for exactly this feature; glad to see it's already done!

Do you know when the next npm release will be cut so I can get my hands on the new functionality?

bcaudan:
@trestletech, it has been released in v1.25.4

PR #616, opened 2020-11-13
✨ [RUMF-764] Use new intake domain for US

webNeat:
## Motivation

Use the new public intake for US site

## Changes

Applies the new intake domain when the US site is used in the configuration.

## Testing

Unit tests of configuration generation have been updated.

---

### DIFF
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -44,6 +44,11 @@ export const INTAKE_SITE = {
   [Datacenter.US]: 'datadoghq.com',
 }
 
+export const STAGING_INTAKE_SITE = {
+  [Datacenter.EU]: 'datad0g.eu',
+  [Datacenter.US]: 'datad0g.com',
+}
+
 export enum BuildMode {
   RELEASE = 'release',
   STAGING = 'staging',
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -44,8 +44,6 @@ export const INTAKE_SITE = {
   [Datacenter.US]: 'datadoghq.com',
 }
 
-export const NEW_INTAKE_DOMAIN_ALLOWED_SITES = [INTAKE_SITE[Datacenter.US], 'datad0g.com']
-
 export enum BuildMode {
   RELEASE = 'release',
   STAGING = 'staging',
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -44,10 +44,7 @@ export const INTAKE_SITE = {
   [Datacenter.US]: 'datadoghq.com',
 }
 
-export const STAGING_INTAKE_SITE = {
-  [Datacenter.EU]: 'datad0g.eu',
-  [Datacenter.US]: 'datad0g.com',
-}
+export const NEW_INTAKE_DOMAIN_ALLOWED_SITES = [INTAKE_SITE[Datacenter.US], 'datad0g.com']
 
 export enum BuildMode {
   RELEASE = 'release',
--- a/packages/core/src/domain/configuration.spec.ts
+++ b/packages/core/src/domain/configuration.spec.ts
@@ -132,6 +132,16 @@ describe('configuration', () => {
       expect(configuration.isIntakeUrl('https://trace.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
     })
 
+    it('should handle sites with subdomains', () => {
+      const configuration = buildConfiguration(
+        { clientToken, site: 'us3.datadoghq.com', useAlternateIntakeDomains: true },
+        usEnv
+      )
+      expect(configuration.isIntakeUrl('https://rum.browser-intake-us3-datadoghq.com/v1/input/xxx')).toBe(true)
+      expect(configuration.isIntakeUrl('https://logs.browser-intake-us3-datadoghq.com/v1/input/xxx')).toBe(true)
+      expect(configuration.isIntakeUrl('https://trace.browser-intake-us3-datadoghq.com/v1/input/xxx')).toBe(true)
+    })
+
     it('should detect proxy intake request', () => {
       let configuration = buildConfiguration({ clientToken, proxyHost: 'www.proxy.com' }, usEnv)
       expect(configuration.isIntakeUrl('https://www.proxy.com/v1/input/xxx')).toBe(true)
--- a/packages/core/src/domain/configuration.spec.ts
+++ b/packages/core/src/domain/configuration.spec.ts
@@ -130,10 +130,6 @@ describe('configuration', () => {
       expect(configuration.isIntakeUrl('https://rum.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
       expect(configuration.isIntakeUrl('https://logs.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
       expect(configuration.isIntakeUrl('https://trace.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
-
-      expect(configuration.isIntakeUrl('https://rum-http-intake.logs.datadoghq.com/v1/input/xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://browser-http-intake.logs.datadoghq.com/v1/input/xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://public-trace-http-intake.logs.datadoghq.com/v1/input/xxx')).toBe(true)
     })
 
     it('should detect proxy intake request', () => {
@@ -177,8 +173,6 @@ describe('configuration', () => {
 
       expect(configuration.isIntakeUrl('https://rum.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
       expect(configuration.isIntakeUrl('https://logs.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://rum-http-intake.logs.datadoghq.com/v1/input/xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://browser-http-intake.logs.datadoghq.com/v1/input/xxx')).toBe(true)
     })
   })
 })
--- a/packages/core/src/domain/configuration.spec.ts
+++ b/packages/core/src/domain/configuration.spec.ts
@@ -122,8 +122,27 @@ describe('configuration', () => {
     it('should detect intake request for US site', () => {
       const configuration = buildConfiguration({ clientToken }, usEnv)
       expect(isIntakeRequest('https://rum.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
-      expect(isIntakeRequest('https://browser.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://logs.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
       expect(isIntakeRequest('https://trace.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
+
+      expect(isIntakeRequest('https://rum-http-intake.logs.datadoghq.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://browser-http-intake.logs.datadoghq.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://public-trace-http-intake.logs.datadoghq.com/v1/input/xxx', configuration)).toBe(
+        true
+      )
+    })
+
+    it('should detect intake request for staging US site', () => {
+      const configuration = buildConfiguration({ clientToken, site: 'datad0g.com' }, usEnv)
+      expect(isIntakeRequest('https://rum.browser-intake-datad0g.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://logs.browser-intake-datad0g.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://trace.browser-intake-datad0g.com/v1/input/xxx', configuration)).toBe(true)
+
+      expect(isIntakeRequest('https://rum-http-intake.logs.datad0g.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://browser-http-intake.logs.datad0g.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://public-trace-http-intake.logs.datad0g.com/v1/input/xxx', configuration)).toBe(
+        true
+      )
     })
 
     it('should detect proxy intake request', () => {
@@ -148,7 +167,7 @@ describe('configuration', () => {
       expect(isIntakeRequest('https://public-trace-http-intake.logs.foo.com/v1/input/xxx', configuration)).toBe(true)
 
       expect(isIntakeRequest('https://rum.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
-      expect(isIntakeRequest('https://browser.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
+      expect(isIntakeRequest('https://logs.browser-intake-datadoghq.com/v1/input/xxx', configuration)).toBe(true)
     })
   })
 })
--- a/packages/core/src/domain/configuration.spec.ts
+++ b/packages/core/src/domain/configuration.spec.ts
@@ -52,9 +52,9 @@ describe('configuration', () => {
 
   describe('proxyHost', () => {
     it('should replace endpoint host add set it as a query parameter', () => {
-      const configuration = buildConfiguration({ clientToken, proxyHost: 'proxy.io' }, usEnv)
+      const configuration = buildConfiguration({ clientToken, site: 'datadoghq.eu', proxyHost: 'proxy.io' }, usEnv)
       expect(configuration.rumEndpoint).toMatch(/^https:\/\/proxy\.io\//)
-      expect(configuration.rumEndpoint).toContain('?ddhost=rum-http-intake.logs.datadoghq.com&')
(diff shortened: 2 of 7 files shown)


bcaudan (changes requested):
- it is not the correct endpoint for logs
- we should also use new domain in staging
- we should probably handle new and old endpoints for `isIntakeRequest` if there is another sdk running with the old endpoints.

packages/core/src/domain/configuration.ts
```diff
@@ -104,6 +106,12 @@ interface TransportConfiguration {
   version?: string
 }
 
+enum EndpointType {
+  BROWSER = 'browser',
+  RUM = 'rum',
+  TRACE = 'public-trace',
+}
```
bcaudan, line 113: IMO, it would be better to use the new version instead of the legacy one

packages/core/src/boot/init.ts
```diff
@@ -44,6 +44,11 @@ export const INTAKE_SITE = {
   [Datacenter.US]: 'datadoghq.com',
 }
 
+export const STAGING_INTAKE_SITE = {
+  [Datacenter.EU]: 'datad0g.eu',
```
bcaudan, line 48: there is no staging eu

packages/core/src/domain/configuration.ts
```diff
@@ -200,33 +222,58 @@ export function buildCookieOptions(userConfiguration: UserConfiguration) {
...
+      [EndpointType.BROWSER]: 'logs',
+      [EndpointType.RUM]: 'rum',
+      [EndpointType.TRACE]: 'trace',
+    }
+    if (conf.site === INTAKE_SITE[Datacenter.US]) {
+      return `${subdomains[type]}.browser-intake-datadoghq.com`
+    }
+    if (conf.site === STAGING_INTAKE_SITE[Datacenter.US]) {
+      return `${subdomains[type]}.browser-intake-datad0g.com`
+    }
```
bcaudan, line 251: since we will need to enable other host when available, a check against something like `NEW_INTAKE_DOMAIN_ALLOWED_SITES = [...]` could be easier to maintain and could clarify the intent

packages/core/src/domain/configuration.ts
```diff
@@ -200,33 +222,58 @@ export function buildCookieOptions(userConfiguration: UserConfiguration) {
...
+    endpoints.push(configuration.internalMonitoringEndpoint)
+  }
+  if (!!configuration.replica) {
+    endpoints.push(
+      configuration.replica.logsEndpoint,
+      configuration.replica.rumEndpoint,
+      configuration.replica.internalMonitoringEndpoint
+    )
+  }
+  return endpoints.concat(configuration.legacyEndpoints || [])
```
bcaudan, line 276: I think we could simplify the problem by not checking explicitly every possible endpoints but only if host match:
- `-http-intake.logs.${conf.site}`
- `.browser-intake-${conf.site}`
- proxyHost

webNeat: In order to do that we need to access the `site` from `isIntakeRequest`
```
function isIntakeRequest(url: string, configuration: Configuration)
```
One way would be to add `site` to the `Configuration` type. What do you think?

bcaudan: I agree.

Similar to your current implementation, we could clarify the intent by computing and storing an `intakeDomains` array instead.

packages/core/src/domain/configuration.ts
```diff
@@ -174,12 +184,15 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
...
+          EndpointType.LOGS,
           replicaTransportConfiguration,
           'browser-agent-internal-monitoring'
         ),
-        logsEndpoint: getEndpoint('browser', replicaTransportConfiguration),
-        rumEndpoint: getEndpoint('rum', replicaTransportConfiguration),
+        logsEndpoint: getEndpoint(EndpointType.LOGS, replicaTransportConfiguration),
+        rumEndpoint: getEndpoint(EndpointType.RUM, replicaTransportConfiguration),
+      }
+      if (transportConfiguration.site !== INTAKE_SITE[Datacenter.US]) {
```
bcaudan, line 194: Do we need this check?

packages/core/src/domain/configuration.ts
```diff
@@ -215,18 +228,44 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
...
+    [EndpointType.TRACE]: 'public-trace',
+  }
+  return `${oldTypes[type]}-http-intake.logs.${conf.site}`
+}
+
+function getIntakeUrls(conf: TransportConfiguration) {
+  if (conf.proxyHost) {
+    return [`https://${conf.proxyHost}/v1/input/`]
+  }
+  const domains = [
```
bcaudan, line 247:
```suggestion
  const urls = [
```

packages/core/src/domain/configuration.ts
webNeat: if the site is `INTAKE_SITE[Datacenter.US]` we will end up pushing the same URLs twice. That's why I added this condition.

bcaudan: since it is not planned to have replica on a prod configuration and if that happens the impact should be negligible, I'd be in favor of removing this check for simplicity sake.

packages/core/src/domain/configuration.ts
```diff
@@ -215,18 +226,44 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
...
-        (haveSameOrigin(url, configuration.replica.logsEndpoint) ||
-          haveSameOrigin(url, configuration.replica.rumEndpoint) ||
-          haveSameOrigin(url, configuration.replica.internalMonitoringEndpoint))))
-  )
+  for (const intakeUrl of configuration.intakeUrls) {
+    if (url.indexOf(intakeUrl) === 0) {
+      return true
+    }
+  }
+  return false
```
BenoitZugmeyer, line 266: Could be:
```suggestion
  return configuration.intakeUrls.some((intakeUrl) => url.indexOf(intakeUrl) === 0)
```

packages/core/src/domain/configuration.ts
```diff
@@ -215,18 +226,44 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
...
+  ]
+  if (NEW_INTAKE_DOMAIN_ALLOWED_SITES.indexOf(conf.site) !== -1) {
+    urls.push(
+      `https://rum.browser-intake-${conf.site}/v1/input/`,
+      `https://logs.browser-intake-${conf.site}/v1/input/`,
+      `https://trace.browser-intake-${conf.site}/v1/input/`
+    )
+  }
+  return urls
+}
```
BenoitZugmeyer, line 258: This looks great, nice and simple. One question (and forgive me if it was already discussed beforehand): do we need to store it in the configuration? Could this list be inlined in the `isIntakeRequest`? From my understanding, storing it in the configuration is a way to cache it, but I'm not sure that we should use the configuration as a cache.

packages/core/src/domain/configuration.ts
webNeat: Would the `.some()` work on IE 11? because I tried `.includes()` before and it didn't.

BenoitZugmeyer: Yes, IE9 supports `.some`: https://caniuse.com/?search=some

packages/core/src/domain/configuration.ts
webNeat: I agree that this is not the best way to store the intake URLs. But the other solution would be to store the `userConfiguration.site` inside the configuration object and let `isIntakeRequest` compute the URLs on every call. Do you think we should go that way?

BenoitZugmeyer: A third solution may be to use a strategy similar to `configuration.isEnabled`: instead of storing `intakeUrls` in the configuration object, we would implement a `configuration.isIntakeUrl` method. WDYT @bcaudan ?

bcaudan: That make sense to me, this function is already in the configuration file so it was begging to be inside configuration 🙂

packages/core/src/domain/configuration.ts
webNeat: Done

packages/core/src/domain/configuration.ts
webNeat: I implemented `configuration.isIntakeUrl`

bcaudan (approved):
Nice job

packages/core/src/domain/configuration.ts
```diff
@@ -54,6 +54,7 @@ export interface UserConfiguration {
   env?: string
   version?: string
 
+  useNewIntakeDomains?: boolean
```
bcaudan, line 57: since we will keep both intakes, old and new is maybe not the best naming.
Maybe something like `useAlternateIntakeDomains`?
it could worth discussing that with @hdelaby

packages/core/src/domain/configuration.spec.ts
```diff
@@ -119,29 +119,31 @@ describe('configuration', () => {
 
     it('should detect intake request for US site', () => {
       const configuration = buildConfiguration({ clientToken }, usEnv)
-      expect(configuration.isIntakeUrl('https://rum.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://logs.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://trace.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
+      // expect(configuration.isIntakeUrl('https://rum.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
+      // expect(configuration.isIntakeUrl('https://logs.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
+      // expect(configuration.isIntakeUrl('https://trace.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
```
bcaudan, line 124: what is your intent?

packages/core/src/domain/configuration.spec.ts
```diff
@@ -160,6 +162,21 @@ describe('configuration', () => {
       expect(configuration.isIntakeUrl('https://browser-http-intake.logs.foo.com/v1/input/xxx')).toBe(true)
       expect(configuration.isIntakeUrl('https://public-trace-http-intake.logs.foo.com/v1/input/xxx')).toBe(true)
 
+      // expect(configuration.isIntakeUrl('https://rum.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
+      // expect(configuration.isIntakeUrl('https://logs.browser-intake-datadoghq.com/v1/input/xxx')).toBe(true)
```
bcaudan, line 166: What is your intent?

packages/core/src/domain/configuration.ts
```diff
@@ -215,18 +227,42 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
...
+    sites.push(INTAKE_SITE[Datacenter.US])
+  }
+  const urls = []
+  for (const site of sites) {
+    urls.push(
+      `https://rum-http-intake.logs.${site}/v1/input/`,
+      `https://browser-http-intake.logs.${site}/v1/input/`,
+      `https://public-trace-http-intake.logs.${site}/v1/input/`
+    )
+    if (conf.useNewIntakeDomains && NEW_INTAKE_DOMAIN_ALLOWED_SITES.indexOf(site) !== -1) {
```
bcaudan, line 257: Since this will be a customer configuration, it seems unlikely that they will perform it on one SDK but not on another.
IMO, we could then check only one domains set or the other depending on the conf.

packages/core/src/domain/configuration.spec.ts
webNeat: I forgot to remove these lines, will do now

packages/core/src/domain/configuration.spec.ts
webNeat: I forgot to remove these lines, will do now

packages/core/src/domain/configuration.ts
webNeat: Makes sense. But what if they used a site that doesn't support new intake domains yet (like EU right now), don't we want to fallback to the old ones? Should we show them a warning in that case?

bcaudan: > what if they used a site that doesn't support new intake domains yet (like EU right now), don't we want to fallback to the old ones?

Yes

> Should we show them a warning in that case?

why not, we do this kind of checks on the configuration already


I don't see how these questions are related to the initial comment though.

webNeat: Sorry, I misunderstood your initial comment. I thought that you suggested to only check the configuration option and choose domains based on it. But now I see that we also check if the configured site is supported. All good now.

packages/core/src/boot/init.ts
```diff
@@ -44,6 +44,8 @@ export const INTAKE_SITE = {
   [Datacenter.US]: 'datadoghq.com',
 }
 
+export const NEW_INTAKE_DOMAIN_ALLOWED_SITES = [INTAKE_SITE[Datacenter.US], 'datad0g.com']
```
bcaudan, line 47: Since it will now be enabled by customer, I'd be in favor of supporting all sites out of the box, in order to not have to updates the sdk when the intakes will be ready

webNeat: Makes sense. But if the user uses `useAlternateIntakeDomains: true` while the alternate intake for their site is not ready yet, we will not be able to fallback to the classic intake. is that ok?

bcaudan: I would say yes, we could specify in the doc in which data center it is available and only update the doc when a new domain is up

packages/core/src/domain/configuration.ts
```diff
@@ -98,12 +99,33 @@ interface TransportConfiguration {
   sdkVersion: string
   applicationId?: string
   proxyHost?: string
+  useAlternateIntakeDomains: boolean
```
bcaudan, line 102: Do we need it?

packages/core/src/domain/configuration.ts
```diff
@@ -215,18 +249,28 @@ function getEndpoint(type: string, conf: TransportConfiguration, source?: string
...
-      (!!configuration.replica &&
-        (haveSameOrigin(url, configuration.replica.logsEndpoint) ||
-          haveSameOrigin(url, configuration.replica.rumEndpoint) ||
-          haveSameOrigin(url, configuration.replica.internalMonitoringEndpoint))))
-  )
+function getHost(intakeType: IntakeType, endpointType: EndpointType, site: string) {
+  const middleDomain = INTAKE_MIDDLE_DOMAINS[intakeType]
+  const endpoint = ENDPOINTS[intakeType][endpointType]
+  return `${endpoint}${middleDomain}${site}`
+}
```
bcaudan, line 256: Could we make it support host for new datacenter as well?
cf us3 use case

packages/core/src/domain/configuration.ts
webNeat: Nice catch, this is not needed anymore.

packages/core/src/domain/configuration.ts
webNeat: I think all new datacenters hosts follow the pattern `${endpoint}.browser-intake-${domain}` so as long as the `userConfiguration.site` equals the `domain`, it should work. Am I missing something?

bcaudan: yes, us3 case

packages/core/src/domain/configuration.ts
```diff
@@ -250,7 +250,9 @@ function getEndpoint(
 function getHost(intakeType: IntakeType, endpointType: EndpointType, site: string) {
   const middleDomain = INTAKE_MIDDLE_DOMAINS[intakeType]
   const endpoint = ENDPOINTS[intakeType][endpointType]
-  return `${endpoint}${middleDomain}${site}`
+  const domainParts = site.split('.')
+  const ext = domainParts.pop()
+  return `${endpoint}${middleDomain}${domainParts.join('-')}.${ext}`
```
bcaudan, line 255:
```js
const configuration = buildConfiguration(
  { clientToken, site: 'us3.datadoghq.com', useAlternateIntakeDomains: false },
  usEnv
)
expect(configuration.isIntakeUrl('https://rum-http-intake.logs.us3.datadoghq.com/v1/input/xxx')).toBe(true)

```
This test must pass

packages/core/src/domain/configuration.ts
```diff
@@ -250,9 +250,13 @@ function getEndpoint(
 function getHost(intakeType: IntakeType, endpointType: EndpointType, site: string) {
   const middleDomain = INTAKE_MIDDLE_DOMAINS[intakeType]
   const endpoint = ENDPOINTS[intakeType][endpointType]
-  const domainParts = site.split('.')
-  const ext = domainParts.pop()
-  return `${endpoint}${middleDomain}${domainParts.join('-')}.${ext}`
+  let suffix = site
+  if (intakeType === 'alternate') {
+    const domainParts = site.split('.')
+    const ext = domainParts.pop()
```
bcaudan, line 256: I'd be in favor of avoiding abbreviation to be more explicit

packages/core/src/domain/configuration.ts
```diff
@@ -250,9 +250,13 @@ function getEndpoint(
...
+    suffix = `${domainParts.join('-')}.${ext}`
+  }
+  return `${endpoint}${middleDomain}${suffix}`
```
bcaudan, line 259: IMO, making an abstraction for middle domain seems unnecessary and implementation could be clearer without it:

```ts
function getHost(intakeType: IntakeType, endpointType: EndpointType, site: string) {
  const endpoint = ENDPOINTS[intakeType][endpointType]
  if (intakeType === 'classic') {
    return `${endpoint}-http-intake.logs.${site}`
  }
  const domainParts = site.split('.')
  const extension = domainParts.pop()
  const suffix = `${domainParts.join('-')}.${extension}`
  return `${endpoint}.browser-intake-${suffix}`
}
```

webNeat: Now that I see it, I agree.

bcaudan (approved):
Nice!

PR #620, opened 2020-11-16
✅ display the jasmine seed after executing tests

BenoitZugmeyer:
## Motivation

Jasmine tests execution order is random. This can be useful to spot issues that occurs because of tests side effects, but sometimes it makes issues hard to reproduce.

## Changes

Display the seed used to execute tests, so they can be re-executed in the same order by using the jasmine `seed` option

Inspired from https://github.com/ketan/karma-jasmine-seed-reporter and https://gist.github.com/odlp/e1586f4f198d157de50e8303b18acfd9/ , but for some reason when using `baseReporterDecorator(this);` in the reporter (so it can use `this.write`), `console.log()` displays the same message twice. Using a karma "logger" instead of `this.write` fix the issue.

## Testing

---


test/unit/jasmineSeedReporterPlugin.js
```diff
@@ -0,0 +1,14 @@
+function seedReporter(logger) {
+  const log = logger.create('jasmine-seed-reporter')
+  this.onBrowserComplete = function(browser, result) {
+    if (result.order && result.order.random && result.order.seed) {
+      log.info(`${browser}: Randomized with seed ${result.order.seed}\n`)
```
bcaudan, line 5: Way better to display browser for browser stack tests seed!

bcaudan (approved):
LGTM

PR #621, opened 2020-11-16
✨ [RUMF-780] implement track first hidden

BenoitZugmeyer:
## Motivation

Currently, we are tracking particular performance entries only if the page is not hidden when initializing the SDK (see https://github.com/DataDog/browser-sdk/pull/361 ). The issue with this strategy is: if the page is visible when initializing RUM, but it gets hidden before the `first-contentful-paint` is produced, the metric will be incorrect.

We want to replace this strategy with something similar to [`web-vitals`](https://github.com/GoogleChrome/web-vitals/blob/master/src/lib/getFirstHidden.ts). It'll provide stronger guarantees that the page was not hidden prior to the observed performance entry.

## Changes

* Add an utility function to add global event listeners
* Move and split the function to track view timings: while quite concise for now, it'll grow in the near future when we add more metrics to it.
* Add a few tests
* Implement `trackFirstHidden`
* Use `trackFirstHidden` when collecting FCP

## Testing

---


packages/core/src/tools/utils.ts
```diff
@@ -304,3 +307,52 @@ export function safeTruncate(candidate: string, length: number) {
   }
   return candidate.slice(0, length)
 }
+
+export function addGlobalEventListeners(
```
bcaudan, line 312: A comment on the goal of this wrapper could be helpful

packages/core/src/tools/utils.ts
```diff
@@ -304,3 +307,52 @@ export function safeTruncate(candidate: string, length: number) {
...
+
+  return {
+    stop,
+  }
+}
+
+let supportPassiveEventListenersCache: boolean | undefined
+// Inspired by
+// https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support
+function supportPassiveEventListeners() {
```
bcaudan, line 338: what is the behavior without this check?

packages/rum/src/domain/rumEventsCollection/view/trackTimings.ts
```diff
@@ -10,26 +10,51 @@ export interface Timings {
...
+  return {
+    stop() {
+      stopNavigationTracking()
+      stopFCPTracking()
+    },
+  }
+}
+
+function trackNavigationTimings(lifeCycle: LifeCycle, callback: (newTimings: Partial<Timings>) => void) {
+  const { unsubscribe: stop } = lifeCycle.subscribe(LifeCycleEventType.PERFORMANCE_ENTRY_COLLECTED, (entry) => {
```
bcaudan, line 33: 🤔 I am wondering if we should not rename `unsubscribe` into `stop` to be consistent with all the other stop methods

bcaudan, line 33: 🤔 I am wondering if we should not rename `unsubscribe` into `stop` to be consistent with all the other stop methods

packages/rum/src/domain/rumEventsCollection/view/trackTimings.ts
```diff
@@ -10,26 +10,51 @@ export interface Timings {
...
+        loadEventEnd: entry.loadEventEnd,
+      })
+    }
+  })
+
+  return { stop }
+}
+
+function trackFirstContentfulPaint(lifeCycle: LifeCycle, callback: (fcp: number) => void) {
+  const { unsubscribe: unsubscribeLifeCycle } = lifeCycle.subscribe(
```
bcaudan, line 49: it would be simpler to use `stop` here like the one above

packages/core/src/tools/utils.ts
BenoitZugmeyer: We have to have a check so event capture works on IE11, since it doesn't support addEventListener object options. Because we need to pass an option object only for passive events, I thought it would make sense to have this check.

bcaudan: FMU, we did not need this test before because we did not have a use case where `{capture: false}` would be evaluated to a truthy capture value on IE11.
Now, with the wrapper we need to handle this case since we try to be passive everywhere.
is that is?

packages/core/src/tools/utils.ts
```diff
@@ -304,3 +307,52 @@ export function safeTruncate(candidate: string, length: number) {
...
+        get passive() {
+          supportPassiveEventListenersCache = true
+          return false
+        },
+      }
+
+      // tslint:disable-next-line: no-null-keyword
+      addEventListener('test' as any, null as any, options)
+      // tslint:disable-next-line: no-null-keyword
+      removeEventListener('test' as any, null as any, options)
```
bcaudan, line 353: what about using `noop` instead of `null`?

packages/core/src/tools/utils.ts
BenoitZugmeyer: Ah! I didn't know using an object would evaluate to true on IE11, that's great. OK to remove this check then. I see two alternatives:
* I don't think we need non-capturing event listeners, so we could simply remove the option and pass `{ capture: true, passive: true }` to all listeners. This makes the API easier to use.
* If you think that's a change a bit to risky, I can restore the previous behavior: add a `passive` option and if `true`, always use an option object, so listeners would be capturing on IE11 but not on newer browsers.

packages/core/src/tools/utils.ts
BenoitZugmeyer: Ok, but let's see first if we should remove this check altogether

packages/core/src/tools/utils.ts
bcaudan: > I didn't know using an object would evaluate to true on IE11

just a guess, I have not tested it

> OK to remove this check then

I am not saying that we should remove it, I am just trying to figure things out for now 🙂  

trying to be passive everywhere seems good to me, since it is not the default behavior for event listeners a comment on the function would be nice
if capturing everything would work for our current use cases, why not, it could worth ensuring it though.
I am not against keeping this check if we think we need or can have `capture: false` use cases.

BenoitZugmeyer: I think restoring the previous behavior (second alternative) would be safer for now, and we could try using only capturing listeners in a future PR.

bcaudan: what make you choose this alternative over the current implementation?

BenoitZugmeyer: * We'd have the same behavior as in prod, giving us time to reconsider changing the behavior in a separate PR
* Shorter / simpler code

bcaudan: > We'd have the same behavior as in prod

FMU, some events would be capturing in IE 11 but were not before, no?

packages/core/src/tools/utils.ts
```diff
@@ -304,3 +308,52 @@ export function safeTruncate(candidate: string, length: number) {
...
+    once
+      ? (event: Event) => {
+          stop()
+          listener(event)
+        }
+      : listener
+  )
+
+  const options = supportPassiveEventListeners() ? { capture, passive: true } : capture
+  events.forEach((event) => addEventListener(event, wrapedListener, options))
```
bcaudan, line 327: should not we make the difference between `window` and `document` addEventListener?

packages/core/src/tools/utils.ts
BenoitZugmeyer: To be clear, I'd modify the function like this:
```ts
function addGlobalEventListeners(..., options: { once: boolean, capture: boolean, passive: boolean }) {
    [...]
    const options = passive ? { capture, passive } : capture
    [...]
}
```

The [only case](https://github.com/DataDog/browser-sdk/blob/5b4418b2446a02fc6b7c633b3af4a4da875bd30c/packages/core/src/domain/sessionManagement.ts#L146) of passive event listener we have is also capturing, so it doesn't change anything.

bcaudan: ok clearer, lgtm

packages/core/src/tools/utils.ts
BenoitZugmeyer: I don't think it matters too much but just to be safe, I restored the ability to add listeners on `document`

packages/rum/src/domain/rumEventsCollection/view/trackTimings.ts
BenoitZugmeyer: No strong opinion on this, I feel `unsubscribe` is a bit easier to understand (it's clear that it's the inverse of `subscribe`), but if we constantly rename it to something else we could name it `stop` to be a bit more consistent.

bcaudan: agree, let's wait to have more need on that

packages/core/src/tools/utils.ts
```diff
@@ -304,3 +308,52 @@ export function safeTruncate(candidate: string, length: number) {
...
+ *   * using an option object only if needed
+ *   * emulating the `once` option
+ *
+ * * wraps the listener with a `monitor` function
+ *
+ * * returns a `stop` function to unsubscribe the listener
+ */
+export function addEventListeners(
+  emitter: EventEmitter,
+  events: DOM_EVENT[],
```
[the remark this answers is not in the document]
BenoitZugmeyer: I have a few more cases coming right up!

packages/core/src/domain/sessionManagement.ts
```diff
@@ -142,14 +142,13 @@ export function stopSessionManagement() {
...
 export function trackActivity(expandOrRenewSession: () => void) {
-  const doExpandOrRenewSession = monitor(expandOrRenewSession)
-  const options = { capture: true, passive: true }
-  ;[utils.DOM_EVENT.CLICK, utils.DOM_EVENT.TOUCH_START, utils.DOM_EVENT.KEY_DOWN, utils.DOM_EVENT.SCROLL].forEach(
-    (event: string) => {
-      document.addEventListener(event, doExpandOrRenewSession, options)
-      stopCallbacks.push(() => document.removeEventListener(event, doExpandOrRenewSession, options))
-    }
+  const { stop } = utils.addEventListeners(
+    window,
```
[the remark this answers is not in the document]
BenoitZugmeyer: Uniformity: we listen to the click event on `window` for action collection, I thought that we could use `window` here too.

packages/rum/src/domain/rumEventsCollection/view/trackFirstHidden.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { createNewEvent, DOM_EVENT, restorePageVisibility, setPageVisibility } from '@datadog/browser-core'
+import { resetFirstHidden, trackFirstHidden } from './trackFirstHidden'
+
+describe('trackFirstHidden', () => {
+  beforeEach(() => {
+    resetFirstHidden()
```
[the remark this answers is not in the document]
bcaudan: As long as we are doing the same in potential future tests, it should be ok.
FMU, the first reset is useless and we don't reset the last one, so it just does not seem right.

packages/core/src/tools/utils.ts
```diff
@@ -322,17 +322,44 @@ export interface EventEmitter {
...
+ *
+ * * returns a `stop` function to remove the listener
+ */
+export function addEventListener(
+  emitter: EventEmitter,
+  event: DOM_EVENT,
+  listener: (event: Event) => void,
+  options?: AddEventListenerOptions
+) {
+  return addEventListeners(emitter, [event], listener, options)
```
bcaudan, line 349: nit: it could have been an opportunity to decouple wrapper + options logic from the logic that handle multiple events.

bcaudan (approved):
👍

packages/core/src/tools/utils.ts
BenoitZugmeyer: There is a subtlety that would make it harder to do if the implementation was in `addEventListener`: using the `once` option makes sure the listener is called at most once, even when listening to  multiple events.

bcaudan: oh, I would not expect that behavior.
Do we need it?

BenoitZugmeyer: Yes we'll need it. I'll add a documentation note about it.

bcaudan: maybe it could be an extra option, like `onceForAll`

BenoitZugmeyer: In my mind, this is the expected behavior: the listener will be called at most once. I don't think we need both options: I can't imagine a use-case where we'd want to subscribe multiple events and the callback would be invoked once for each event.

bcaudan: ok, let's go this way and we'll see how it goes

PR #630, opened 2020-11-23
💥 [RUMF-730] prefer object and type alias over enum in APIs

bcaudan:
## Motivation

make APIs easier to use with TypeScript by avoiding to import our enums

before
```ts
import { datadogRum, Datacenter } from '@datadog/browser-rum'

datadogRum.init({
  ...
  datacenter: Datacenter.US
  ...
})
```

after
```ts
import { datadogRum } from '@datadog/browser-rum'

datadogRum.init({
  ...
  datacenter: 'us'
  ...
})
```

## Changes

replace enum usage in APIs by object and type alias

before
```ts
export enum Datacenter {	
  US = 'us',	  
  EU = 'eu',	  
}
```

after
```ts
export const Datacenter = {
  US: 'us',	  
  EU: 'eu',	  
} as const

export type Datacenter = (typeof Datacenter)[keyof typeof Datacenter]
```

## Testing

ci

---


packages/logs/src/boot/logs.entry.ts
```diff
@@ -77,7 +75,7 @@ export function makeLogsGlobal(startLogsImpl: StartLogs) {
     removeLoggerGlobalContext: monitor(globalContextManager.remove),
 
     createLogger: monitor((name: string, conf: LoggerConfiguration = {}) => {
-      customLoggers[name] = new Logger(sendLog, conf.handler, conf.level, {
+      customLoggers[name] = new Logger(sendLog, conf.handler as HandlerType, conf.level as StatusType, {
```
BenoitZugmeyer, line 78: What about using constant unions internally too? to avoid those cast

BenoitZugmeyer:
IMO it would be simpler to completely remove `enum` (`ErrorSource`, `StatusType`, `HandlerType`...) and use the string instead. Or, we could also use plain constants like:
```ts
const SOURCE_AGENT = 'agent' as const
```
This would result in a smaller code output, and avoid having two "types" for the same values (ex: `ErrorSource` and `Source`, `HandlerType` and `Handler`...)

 WDYT?

BenoitZugmeyer (approved):
LGTM!

aleksandrlat:
Oh no why did you remove types? We need to duplicate type on our own now :(

bcaudan:
Hi @aleksandrlat,
What was your usage?

PR #635, opened 2020-11-24
✨ [RUMF-770] Disable tracing for cancelled requests

webNeat:
## Motivation

Do not trace cancelled requests because they lack backend information.

## Changes

Set `request.traceId` to `undefined` if the request was cancelled.

## Testing

Added unit tests to ensure that `traceId` is removed from the `LifeCycleEventType.REQUEST_COMPLETED` event when the `request.status == 0`.

---

### DIFF
--- a/packages/rum/src/domain/requestCollection.ts
+++ b/packages/rum/src/domain/requestCollection.ts
@@ -15,10 +15,6 @@ export interface RequestStartEvent {
   requestIndex: number
 }
 
-interface CustomContext {
-  requestIndex: number
-}
-
 export interface RequestCompleteEvent {
   requestIndex: number
   type: RequestType
@@ -44,7 +40,7 @@ export function startRequestCollection(lifeCycle: LifeCycle, configuration: Conf
 }
 
 export function trackXhr(lifeCycle: LifeCycle, configuration: Configuration, tracer: Tracer) {
-  const xhrProxy = startXhrProxy<CustomContext & TracedXhrStartContext, CustomContext & TracedXhrCompleteContext>()
+  const xhrProxy = startXhrProxy<TracedXhrStartContext, TracedXhrCompleteContext>()
   xhrProxy.beforeSend((context, xhr) => {
     if (isAllowedRequestUrl(configuration, context.url)) {
       tracer.traceXhr(context, xhr)
@@ -76,10 +72,7 @@ export function trackXhr(lifeCycle: LifeCycle, configuration: Configuration, tra
 }
 
 export function trackFetch(lifeCycle: LifeCycle, configuration: Configuration, tracer: Tracer) {
-  const fetchProxy = startFetchProxy<
-    CustomContext & TracedFetchStartContext,
-    CustomContext & TracedFetchCompleteContext
-  >()
+  const fetchProxy = startFetchProxy<TracedFetchStartContext, TracedFetchCompleteContext>()
   fetchProxy.beforeSend((context) => {
     if (isAllowedRequestUrl(configuration, context.url)) {
       tracer.traceFetch(context)
@@ -102,7 +95,7 @@ export function trackFetch(lifeCycle: LifeCycle, configuration: Configuration, t
         spanId: context.spanId,
         startTime: context.startTime,
         status: context.status,
-        traceId: context.status === 0 ? undefined : context.traceId,
+        traceId: context.traceId,
         type: RequestType.FETCH,
         url: context.url,
       })
--- a/packages/rum/src/domain/requestCollection.ts
+++ b/packages/rum/src/domain/requestCollection.ts
@@ -1,15 +1,27 @@
-import { Configuration, Observable, RequestType, startFetchProxy, startXhrProxy } from '@datadog/browser-core'
+import {
+  Configuration,
+  FetchCompleteContext,
+  FetchStartContext,
+  Observable,
+  RequestType,
+  startFetchProxy,
+  startXhrProxy,
+  XhrCompleteContext,
+  XhrStartContext,
+} from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { isAllowedRequestUrl } from './rumEventsCollection/resource/resourceUtils'
-import {
-  startTracer,
-  TracedFetchCompleteContext,
-  TracedFetchStartContext,
-  TracedXhrCompleteContext,
-  TracedXhrStartContext,
-  TraceIdentifier,
-  Tracer,
-} from './tracing/tracer'
+import { startTracer, TraceIdentifier, Tracer } from './tracing/tracer'
+
+export interface CustomContext {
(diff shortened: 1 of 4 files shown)


packages/rum/src/domain/requestCollection.ts
```diff
@@ -109,7 +109,7 @@ export function trackFetch(lifeCycle: LifeCycle, configuration: Configuration, t
         spanId: context.spanId,
         startTime: context.startTime,
         status: context.status,
-        traceId: context.traceId,
+        traceId: context.status === 0 ? undefined : context.traceId,
```
bcaudan, line 112: it could be nice to also clear spanId then.

bcaudan, line 112: What about having this logic in `tracer` since it is related to tracing?

webNeat: I agree, I will clear it.

webNeat: I see that we add the `traceId` and `spanId` to context in this file (in `xhrProxy.beforeSend` and `fetchProxy.beforeSend`), so if we want to unset them inside the tracer, maybe we also want to set them there?

bcaudan: good point, tracing methods already have access to context so moving the set logic here seems good to me

webNeat: ok, I will do that.

packages/rum/src/domain/tracing/tracer.ts
```diff
@@ -52,17 +67,16 @@ export function startTracer(configuration: Configuration): Tracer {
...
+    return
   }
 
-  const traceId = new TraceIdentifier()
-  const spanId = new TraceIdentifier()
-  inject(makeTracingHeaders(traceId, spanId))
-  return { traceId, spanId }
+  context.traceId = new TraceIdentifier()
+  context.spanId = new TraceIdentifier()
+  inject(makeTracingHeaders(context.traceId!, context.spanId))
```
bcaudan, line 79:
```suggestion
  inject(makeTracingHeaders(context.traceId, context.spanId))
```

packages/rum/src/domain/tracing/tracer.ts
```diff
@@ -52,17 +67,16 @@ export function startTracer(configuration: Configuration): Tracer {
 
 function injectHeadersIfTracingAllowed(
   configuration: Configuration,
-  url: string,
+  context: Partial<TracedFetchStartContext | TracedXhrCompleteContext>,
```
bcaudan, line 70:
```suggestion
  context: Partial<TracedFetchStartContext | TracedXhrStartContext>,
```

packages/rum/src/domain/requestCollection.ts
```diff
@@ -109,7 +102,7 @@ export function trackFetch(lifeCycle: LifeCycle, configuration: Configuration, t
...
```
bcaudan, line 105: this should not be needed anymore

packages/rum/src/domain/requestCollection.ts
```diff
@@ -48,14 +44,10 @@ export function startRequestCollection(lifeCycle: LifeCycle, configuration: Conf
 }
 
 export function trackXhr(lifeCycle: LifeCycle, configuration: Configuration, tracer: Tracer) {
-  const xhrProxy = startXhrProxy<CustomContext & XhrStartContext, CustomContext & XhrCompleteContext>()
+  const xhrProxy = startXhrProxy<CustomContext & TracedXhrStartContext, CustomContext & TracedXhrCompleteContext>()
```
bcaudan, line 47: `CustomContext & XXXContext` had the advantage of make it clear what was added to the context in this file.
IMO, with only `requestId` in `CustomContext`, `CustomContext` does not really make sense anymore.
What about merging it with `TracedXXXContext`?

packages/rum/src/domain/requestCollection.ts
webNeat: You are right, I will remove it.

packages/rum/src/domain/requestCollection.ts
webNeat: I agree, I will move it.

packages/rum/src/domain/tracing/tracer.ts
```diff
@@ -9,6 +9,7 @@ import {
 } from '@datadog/browser-core'
 
 export interface TracingContext {
+  requestIndex: number
```
bcaudan, line 12: requestIndex is not specific to tracing (iirc, used initially in page activity), so I would leave it in request collection

webNeat: Do you mean something like the following?

__requestCollection.ts__
```ts
import { TracingContext, ... } from './tracing/tracer'

interface CustomContext extends TracingContext {
  requestIndex: number
}
export interface TracedFetchStartContext extends FetchStartContext, CustomContext {}
export interface TracedFetchCompleteContext extends FetchCompleteContext, CustomContext {}
export interface TracedXhrStartContext extends XhrStartContext, CustomContext {}
export interface TracedXhrCompleteContext extends XhrCompleteContext, CustomContext {}
```

__tracer.ts__
```ts
import {
  TracedFetchCompleteContext,
  TracedFetchStartContext,
  TracedXhrCompleteContext,
  TracedXhrStartContext,
} from '../requestCollection'

export interface TracingContext {
  spanId?: TraceIdentifier
  traceId?: TraceIdentifier
}
```

webNeat: What about moving these shared types to the `types.ts` file?

bcaudan: Since TracingContext would then not be used in tracer, why keeping it there?

bcaudan (approved):
👍

buenjybar:
Hi guys,

I was trying out the latest version of today the lib ( `2.2.1` ) and, 
Unfortunately, I can still see some errors related to "canceled XHR requests" 

`XHR error GET https://URL/..`
It is happening on both `LOG` and `RUM` libs

attached some screenshots 
1. image
2. image

BenoitZugmeyer:
This is expected. This PR disables *traces* for those requests, but they are still considered as errors. Could you provide more information about your issue/use-case?

buenjybar:
Ok,
I understand you will still log them as an error, and it is ok since you are following HTTP error standards, [here](https://stackoverflow.com/questions/872206/what-does-it-mean-when-an-http-request-returns-status-code-0/14507670#:~:text=An%20HTTP%20response%20code%20of,execute%20within%20a%20specified%20time)
This can bring some confusion to the support team when investigating bugs, ( I could add a pipeline to downgrade this error code:0 to Warnings level ) 
People think it is a legit HTTP error but it is actually part of our code ( we cancel the request and trigger a new one)

PR #627, opened 2020-11-23
💥 [RUMF-730] use v2 events format

bcaudan:
## Motivation

Use the same event format as the one used to display the data.

## Changes

**Breaking:**
Remove RUM event types export 

**Other:**
- Remove v1 specific code + v2 references
- Remove v2_format feature flag
- fix loadEvent naming

## Testing

automated tests

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/master/CONTRIBUTING.md) documentation.

PR #638, opened 2020-12-07
✨ [RUM] implement the User API

BenoitZugmeyer:
## Motivation

Implement the "user api" to define the current user in all RUM events

## Changes

* Implement `DD_RUM.setUser(user: User)`
* add tests

## Testing

Unit tests + manual testing

---


packages/rum/src/typesV2.ts
```diff
@@ -148,3 +148,15 @@ export type RumEventV2 =
...
   | RumActionEventV2 & ViewContextV2 & RumContextV2
+
+export interface User {
+  id?: number | string
+  email?: string
+  name?: string
+  [key: string]: unknown
+}
+
+export interface GlobalAttributes {
```
bcaudan, line 159: naming: attributes seems inconsistent with context that we are using everywhere. To avoid collision with globalContext we could use something like customerContext or customerCommonContext.
wdyt?

packages/rum/src/domain/assemblyV2.ts
```diff
@@ -63,7 +54,9 @@ export function startRumAssemblyV2(
           ? combine(rumContext, viewContext, actionContext, rawRumEvent)
           : combine(rumContext, viewContext, rawRumEvent)
         const serverRumEvent = withSnakeCaseKeys(rumEvent)
-        serverRumEvent.context = combine(savedGlobalContext || getGlobalContext(), customerContext)
+        const globalAttributes = savedGlobalAttributes || getGlobalAttributes()
+        serverRumEvent.context = combine(globalAttributes.context, customerContext)
+        serverRumEvent.user = deepClone(globalAttributes.user as Context)
```
bcaudan, line 59: what is the reasoning for cloning the user but not the context?
it would probably be better to do it when setting the user to avoid to do it for each event.

packages/rum/src/typesV2.ts
```diff
@@ -148,3 +148,15 @@ export type RumEventV2 =
...
+  user: User
+  context: Context
```
bcaudan, line 161: what about handling undefined values?
it could be a small payload optimization, avoiding to attach empty values to all events

packages/rum/src/typesV2.ts
BenoitZugmeyer: What's bothering me with using `context` in the name is that it already contains a `context` name, so we would end up with `customerContext.context`, which is not ideal. Also, it may not be only customer-related data in the future. Maybe we can find something else, like `Environment`?

packages/rum/src/domain/assemblyV2.ts
BenoitZugmeyer: The context is cloned via `combine`. I had to do this to support [this test](https://github.com/DataDog/browser-sdk/blob/benoit/user-api/packages/rum/src/domain/assemblyV2.spec.ts#L194-L208) which is a copy of [this one about global context](https://github.com/DataDog/browser-sdk/blob/benoit/user-api/packages/rum/src/domain/assemblyV2.spec.ts#L140-L154), but I'm not sure why we need to dereference the global context in the first place, since it is serialized synchronously right after that. Maybe I could remove the `deepClone` and test?

packages/rum/src/typesV2.ts
BenoitZugmeyer: Ok for the `user`, but `context` would be a bit harder since it is expected to always be an object, no? (meaning: `DD_RUM.getRumGlobalContext()` is expected to return an object, not `undefined`)

packages/rum/src/typesV2.ts
bcaudan: > Maybe we can find something else, like Environment?

IMO we should have a similar naming for this concept and the existing *Context concepts, because the purpose is the same, only the content varies.
I'd prefer having `Context.context` than thinking of another name for this generic concept and make the update everywhere.

> Also, it may not be only customer-related data in the future
we could rename it at this point then or use only commonContext

packages/rum/src/domain/assemblyV2.ts
bcaudan: sounds good to me

packages/rum/src/typesV2.ts
bcaudan: Could we have both behavior?
`DD_RUM.getRumGlobalContext()` returns an object but if empty context is not sent with the payload

packages/rum/src/typesV2.ts
BenoitZugmeyer: Let's use `CommonContext`

packages/rum/src/typesV2.ts
BenoitZugmeyer: Ok! I'll change this during the assembly then.

packages/rum/src/domain/assemblyV2.spec.ts
```diff
@@ -172,6 +192,54 @@ describe('rum assembly v2', () => {
...
+      user = { fooBar: 'foo' }
+      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_V2_COLLECTED, {
+        rawRumEvent: createRawRumEvent(RumEventType.VIEW),
+        startTime: 0,
+      })
+
+      expect(serverRumEvents[0].user!.fooBar).toEqual('foo')
+    })
+
+    it('should ignore the current global context when a saved global context is provided', () => {
```
bcaudan, line 226:
```suggestion
    it('should ignore the current common context when a saved common context is provided', () => {
```

bcaudan (approved):

BenoitZugmeyer:
Waiting for #627 before merging

loick:
Hi there, where can I find the documentation about this? I didn't find it on the website, nor on the npm library. I've seen a lot of updates recently on this repository, and we would be curious about its new features :) Thanks

bcaudan:
Hi @loick,
Thanks for your enthusiasm 😉 
Yes, there are some new features coming, the documentation should be available pretty soon!

loick:
okay thanks!

PR #640, opened 2020-12-09
[RUMF-772] generate types from json schema

bcaudan:
## Motivation

Generate a types from rum-events-format, allowing to provide it in the future `beforeSend` API

## Changes

Introduce a command to update rum-events-format submodule and generate corresponding types
Use a fork of the lib to support readOnly property

## Testing

- update rum-events-format
- run `yarn rum-events-format:sync`

---


package.json
```diff
@@ -23,7 +23,10 @@
     "test:e2e": "BUILD_MODE=e2e-test yarn build && (cd test/app && rm -rf node_modules && yarn && yarn build) && wdio test/e2e/wdio.local.conf.js",
     "test:e2e:bs": "BUILD_MODE=e2e-test yarn build && (cd test/app && rm -rf node_modules && yarn && yarn build) && node ./scripts/bs-wrapper.js wdio test/e2e/wdio.bs.conf.js",
     "test:compat:tsc": "yarn build && (cd test/app && rm -rf node_modules && yarn && yarn compat:tsc) || yarn fail 'typescript 3.0 compatibility broken'",
-    "test:compat:ssr": "yarn build && (cd test/app && rm -rf node_modules && yarn && yarn compat:ssr) || yarn fail 'server side rendering compatibility broken'"
+    "test:compat:ssr": "yarn build && (cd test/app && rm -rf node_modules && yarn && yarn compat:ssr) || yarn fail 'server side rendering compatibility broken'",
+    "submodule:update": "git submodule update --init && git submodule update --remote",
+    "json2type:build": "cd ./node_modules/json-schema-to-typescript && rm -rf dist && npm i; npm run clean && npm run build:server; echo '>>> errors due to build inside node_modules, but it should still work fine'",
```
webNeat, line 28: I don't understand why do we need to rebuild this dependency. Could you please explain what is the problem with using the originally built files?

bcaudan: It is because I am using a forked version, the files are not built when installing the package from github instead of npm

BenoitZugmeyer: This is because the dependency is on a git branch, not a package published on NPM, and yarn don't build the package when installing. This is not pretty, but since it is temporary while the PR is not merged upstream, I guess it's fine!

BenoitZugmeyer (commented):
Instead of writing non-trivial scripts in the package.json, I'd be in favor of using `./scripts/cli` like the lint and typecheck commands. We would be able to use comments and multiple lines.

package.json
bcaudan: Yes, I prefer to have this ugly rebuild than having to publish a new package or commit the dist files on the fork

BenoitZugmeyer (approved):

Issue #568, opened 2020-10-15
DOM event objects provided as errors are hard to troubleshoot


dtinth:
Some code (not sure if ours or 3rd party) seems to be throwing an Event object.

![image](https://user-images.githubusercontent.com/193136/96103902-6bed5800-0f02-11eb-94a3-016922dbdece.png)

This is a commonly-misused pattern in JavaScript world.

```js
something.onerror = e => {
  throw e
}
```

It would help if the event name and event target class name is serialized in Datadog RUM.

elsmr:
Hi @dtinth, browsers intentionally hide errors thrown by script files from different origins for security reasons.

You could try tweaking your CORS settings (if you control the third-party origin), or set `crossorigin=anonymous` on the script tags. For more details check out [this blog post](https://blog.sentry.io/2016/05/17/what-is-script-error#the-fix-cors-attributes-and-headers).

Let me know if that works for you 😄

amortemousque:
Hello @elsmr,
We have improved our object sanitization and now handle DOM events, DOM nodes, errors, and more. For more details, refer to the [test code](https://github.com/DataDog/browser-sdk/blob/4266bcd42b3a3c020658ef7578c9b52e7b8def57/packages/core/src/tools/serialisation/sanitize.spec.ts).

PR #651, opened 2020-12-15
🐛 fix tracekit handling of exotic errors

bcaudan:
## Motivation

Object thrown with an object message instead of a string is forwarded as is, which is not compatible with our backend and our UI.
ex:
```js
throw { message: { foo: 'bar' }}
```
sent:
```json
{ "message": { "foo": "bar" }, "stack": "Error: { \"foo\": \"bar\" }"}
```
expected:
```json
{ "message": "Uncaught { \"foo\": \"bar\" }", "stack": "'No stack, consider using an instance of Error'"}
```

## Changes

- Check inside tracekit that StackTrace message and name are string
- allow `StackTrace.message = undefined` which was already checked in the code

## Testing

automated tests + sandbox

---

### DIFF
--- a/packages/core/src/domain/tracekit.ts
+++ b/packages/core/src/domain/tracekit.ts
@@ -43,7 +43,7 @@ export interface StackFrame {
  */
 export interface StackTrace {
   name?: string
-  message: string
+  message?: string
   url?: string
   stack: StackFrame[]
   incomplete?: boolean
@@ -931,14 +931,18 @@ export const computeStackTrace = (function computeStackTraceWrapper() {
     }
 
     return {
-      message: forceString((ex as any).message),
-      name: forceString((ex as any).name),
+      message: tryToGetString(ex, 'message'),
+      name: tryToGetString(ex, 'name'),
       stack: [],
     }
   }
 
-  function forceString(candidate: unknown) {
-    return typeof candidate !== 'string' ? JSON.stringify(candidate) : candidate
+  function tryToGetString(candidate: unknown, property: string) {
+    if (typeof candidate !== 'object' || !candidate || !(property in candidate)) {
+      return undefined
+    }
+    const value = (candidate as { [k: string]: unknown })[property]
+    return typeof value === 'string' ? value : undefined
   }
 
   /**
--- a/packages/core/src/domain/tracekit.ts
+++ b/packages/core/src/domain/tracekit.ts
@@ -879,7 +879,7 @@ export const computeStackTrace = (function computeStackTraceWrapper() {
    * @param {(string|number)=} depth
    * @memberof computeStackTrace
    */
-  function doComputeStackTrace(ex: Error, depth?: string | number): StackTrace {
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/tracekit.ts
```diff
@@ -931,12 +931,16 @@ export const computeStackTrace = (function computeStackTraceWrapper() {
...
-      message: ex.message,
-      name: ex.name,
+      message: forceString((ex as any).message),
+      name: forceString((ex as any).name),
       stack: [],
     }
   }
 
+  function forceString(candidate: unknown) {
+    return typeof candidate !== 'string' ? JSON.stringify(candidate) : candidate
```
BenoitZugmeyer, line 941: In this PR, I would do only `candidate === undefined ? undefined : String(candidate)`. `JSON.stringify` can throw, for example if `candidate` is an `Element` (related issue: https://github.com/DataDog/browser-sdk/issues/568) .

packages/core/src/domain/tracekit.ts
```diff
@@ -879,15 +879,15 @@ export const computeStackTrace = (function computeStackTraceWrapper() {
    * @param {(string|number)=} depth
    * @memberof computeStackTrace
    */
-  function doComputeStackTrace(ex: Error, depth?: string | number): StackTrace {
+  function doComputeStackTrace(ex: unknown, depth?: string | number): StackTrace {
```
BenoitZugmeyer, line 882: Because we are using `ex.name` and `ex.message` without checking if `ex` is an object, I would be a bit more restrictive on the typing:
```suggestion
  function doComputeStackTrace(ex: { name?: string, message?: string }, depth?: string | number): StackTrace {
```
This would avoid the cast to `any` below.

If we want `ex` to be really any value, then we should not access `ex.name` and `ex.message` directly.

packages/core/src/domain/tracekit.ts
bcaudan: Not sure to get your point, FMU in the linked issue, the case is about DOM event handling not about an exception due to `JSON.stringify` of an `Element` 😕 
`JSON.stringify` is already what we are using in similar cases without message property, cf [formatUnknownError](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/tools/error.ts#L31), any reason to have a different behavior here?

BenoitZugmeyer: My point is: `JSON.stringify` can throw. I think we should use something safer like `jsonStringify` or `String`.

bcaudan: OK, I get it.
not needed with the updated implementation

packages/core/src/domain/tracekit.ts
bcaudan: I've updated the strategy and kept the unknown.

BenoitZugmeyer (approved):
LGTM

PR #655, opened 2020-12-16
🐛 sanitize error properties even when there is a valid stack

bcaudan:
## Motivation

Following #651, we still observe some malformed errors when tracekit can compute a stacktrace but the message is malformed.

## Changes

Do not assume input exception type and try to sanitize properties before using them.

## Testing

automated tests

---


BenoitZugmeyer (approved):
LGTM

PR #657, opened 2020-12-16
[RUMF-801] create a new package for rum-recorder

BenoitZugmeyer:
This is the first PR for implementing a session-replay recorder in RUM. Next PR: #658 

## Motivation

Create a package for `rum-recorder`. For now, it exports the exact same things than `rum`.

## Testing

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/master/CONTRIBUTING.md) documentation.

PR #658, opened 2020-12-16
[RUMF-803] import RRWeb

BenoitZugmeyer:
Previous PR: #657 
Next PR: #670 

## Motivation

Import RRWeb in the rum-recorder codebase

## Testing

---

### DIFF
--- a/packages/rum-recorder/src/domain/rrweb/record.spec.ts
+++ b/packages/rum-recorder/src/domain/rrweb/record.spec.ts
@@ -1,0 +1,36 @@
+import { createNewEvent } from '@datadog/browser-core'
+import { record } from './record'
+import { Event, EventType } from './types'
+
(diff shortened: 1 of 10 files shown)


packages/rum-recorder/src/domain/rrweb/record.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+
+  afterEach(() => {
+    input.remove()
+    if (stop) {
+      stop()
+    }
+  })
+
+  it('will only have one full snapshot without checkout config', () => {
```
BenoitZugmeyer, line 21: As I said, I only included a single test, so we can discuss if we are comfortable with the current RRWeb integration before importing more tests.

bcaudan (approved):

PR #668, opened 2020-12-28
✨ Implement addTiming

webNeat:
## Motivation

Add `DD_RUM.addTiming()` API.

## Changes

Added the method `addTiming` to the global RUM object
```ts
function addTiming(name: string, inInitialView = false): void
```
- `name` is the name of the custom timing. If the view already has that timing, it will be updated with the given `time`.
- `inInitialView` indicates if we want to add the timing in the initial view. Otherwise the timing is added to the current view.

This new method is hidden behind the feature flag `custom-timings`.

## Testing

- Added automated tests
- Tested with the local sandbox
- Tested the facet list timing on the RUM Explorer. Got around 1.5 sec.

---


packages/rum/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -829,3 +829,88 @@ describe('rum view measures', () => {
...
+    setCustomTiming('foo', false, time1)
+
+    const time2 = performance.now()
+    setCustomTiming('bar', false, time2)
+
+    const event = getViewEvent(2)
+    expect(event.customTimings).toEqual({
+      bar: time2 - event.startTime,
+      foo: time1 - event.startTime,
+    })
```
bcaudan, line 890: You tests logic duplicate the implementation logic:

- time = performance.now() 
- customTime = time - startTime

It could be better to have control on start time and time and not reimplement the logic in the test to compute the output, something like:

```
setupBuilder.build()

clock.tick(20)
setCustomTiming('foo', false)

clock.tick(10)
setCustomTiming('bar', false)

expect(event.customTimings).toEqual({
      bar: 20,
      foo: 30,
})
```

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -83,6 +88,11 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
   )
 
   return {
+    setCustomTiming(name: string, inInitialView = false, time = performance.now()) {
```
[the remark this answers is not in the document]
webNeat: `addTiming` is a bit misleading since we set a new value for the timing if already exists. Wouldn't it be better to name it `setTiming`?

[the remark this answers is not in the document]
webNeat: A use case where the user wants to record the timing of an event A but only report it if a certain event B happens after. Since we don't offer a way to remove a timing after adding it, the user will have to save the time of event A and only call addTiming when event B happens.

packages/rum/src/domain/rumEventsCollection/view/viewCollection.ts
```diff
@@ -23,6 +23,7 @@ function processViewUpdate(view: View) {
         count: view.eventCounts.userActionCount,
       },
       cumulativeLayoutShift: view.cumulativeLayoutShift,
+      customTimings: view.customTimings,
```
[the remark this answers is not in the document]
webNeat: In the Google document, it says:
> Custom timings are stored in view.custom_timings.[timing-name]

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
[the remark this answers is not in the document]
bcaudan: Do you mean for the public API?
it is maybe a bit late for that since afaik it is implemented on the mobile side, you can launch a discussion on that if you think it is better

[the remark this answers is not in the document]
bcaudan: do we want to handle this case?

packages/rum/src/domain/rumEventsCollection/view/viewCollection.ts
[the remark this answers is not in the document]
bcaudan: ok, my bad

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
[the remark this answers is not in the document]
webNeat: if it's already on the public API of mobile SDK, it makes sense to use the same name here. I will replace all `setCustomTiming` with `addTiming`.

packages/rum/src/domain/rumEventsCollection/view/trackViews.spec.ts
webNeat: You are right, I updated the test.

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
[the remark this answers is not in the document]
bcaudan: The downside is that once it is live we need to maintain it.
Don't hesitate to discuss with @hdelaby about it.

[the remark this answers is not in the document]
webNeat: After discussion with @hdelaby, we decided to keep it simple for the moment and remove the `time` parameter.

packages/rum/src/boot/rum.entry.spec.ts
```diff
@@ -1,11 +1,15 @@
-import { ErrorSource, ONE_SECOND } from '@datadog/browser-core'
+import { Configuration, ErrorSource, ONE_SECOND } from '@datadog/browser-core'
 import { setup, TestSetupBuilder } from '../../test/specHelper'
 import { ActionType } from '../domain/rumEventsCollection/action/trackActions'
 import { makeRumGlobal, RumGlobal, RumUserConfiguration, StartRum } from './rum.entry'
 
 const noopStartRum = () => ({
   addAction: () => undefined,
   addError: () => undefined,
+  addTiming: () => undefined,
```
bcaudan, line 9: could we add some tests for it here?
it could have caught 740ffb1508c15b71c804ef80eb2c7b1409d64596

packages/rum/src/boot/rum.entry.spec.ts
```diff
@@ -1,11 +1,15 @@
...
+  configuration: ({
+    isEnabled: () => true,
+  } as any) as Configuration,
```
bcaudan, line 12: is it needed?

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -193,6 +214,9 @@ function newView(
         setLoadEvent(newTimings.loadEvent)
       }
     },
+    addTiming(name: string, time: number) {
+      customTimings[name] = msToNs(time - startTime)
```
bcaudan, line 218: It should be better to be consistent for where we convert timings to ns.
For other timings, it is done in viewCollection since it is here that we convert internal view representation to raw view event (intake format).
It allow us to internally work with convenient format/unit without depending on intake format.

packages/rum/src/boot/rum.entry.ts
```diff
@@ -25,7 +26,7 @@ export interface RumUserConfiguration extends UserConfiguration {
   beforeSend?: (event: RumEvent) => void
 }
 
-export type RumGlobal = ReturnType<typeof makeRumGlobal>
+export type RumGlobal = ReturnType<typeof makeRumGlobal> & { addTiming?: AddTiming }
```
bcaudan, line 29: What about not exposing it in RumGlobal type?
Since it will not necessarily be there

sandbox/index.html
```diff
@@ -12,6 +12,7 @@
       })
       DD_LOGS.init({
         clientToken: 'xxx',
+        enableExperimentalFeatures: ['custom-timings'],
```
bcaudan, line 15: What is the intent?

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -83,6 +97,11 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
   )
 
   return {
+    addTiming(name: string, inInitialView = false) {
+      const view = inInitialView ? initialView : currentView
+      view.addTiming(name, performance.now())
```
bcaudan, line 102:
```suggestion
      view.addTiming(name)
```
no need for a time parameter then

packages/rum/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -25,6 +25,7 @@ describe('viewCollection', () => {
     const { lifeCycle, rawRumEvents } = setupBuilder.build()
     const view = {
       cumulativeLayoutShift: 1,
+      customTimings: {},
```
bcaudan, line 28: it could be nice to have a test with some values for this file

packages/rum/src/boot/rum.entry.spec.ts
webNeat: Right, I will add a test for it.

packages/rum/src/boot/rum.entry.spec.ts
webNeat: Yes, otherwise Typescript will complain about it.

bcaudan: My bad not clear enough, I was more concerned about the `true` in `isEnabled: () => true`

webNeat: The return of `isEnabled` should be `true` for `addTiming` to be added to `RumGlobal`. Do you think we should make it return `false` by default and only mock it to return `true` when we actually need to test `addTiming`?

packages/rum/src/domain/rumEventsCollection/view/trackViews.ts
webNeat: Makes sense, I will update this.

packages/rum/src/boot/rum.entry.ts
webNeat: if I don't include it on `RumGlobal` type, Typescript will complain whenever I try to access `addTiming`. I see two ways to fix this:
1- Cast `rumGlobal` to `any`.
2- Add a `InternalRumGlobal` type that contains `addTiming` and cast `rumGlobal` to that type when needed.
Which approach sounds good to you? is there a better way of handling this?

sandbox/index.html
webNeat: This is a left over from when I was testing locally, I will remove it.

packages/rum/src/boot/rum.entry.spec.ts
bcaudan: Not critical here, but since the general case would be not enabled, I'd be in favor of that yes

packages/rum/src/boot/rum.entry.ts
bcaudan: > 1- Cast rumGlobal to any.

Could we do that only when we want to use addTiming?

```ts
(DD_RUM as any).addTiming(...)
```

packages/rum/src/domain/rumEventsCollection/view/viewCollection.spec.ts
webNeat: I agree, I will add them.

packages/rum/src/boot/rum.entry.ts
webNeat: Yes, I will do that.

packages/core/src/tools/utils.ts
```diff
@@ -252,6 +252,14 @@ export function isEmptyObject(object: object) {
   return Object.keys(object).length === 0
 }
 
+export function mapObject<A, B>(object: { [key: string]: A }, fn: (arg: A) => B) {
+  const newObject: { [key: string]: B } = {}
+  for (const key of Object.keys(object)) {
+    newObject[key] = fn(object[key])
+  }
+  return newObject
```
bcaudan, line 260: I'd be in favor of:
- naming it mapValues
- not mutate the original object

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -385,4 +389,35 @@ describe('rum entry', () => {
...
+      setupBuilder = setup()
+    })
+
+    afterEach(() => {
+      setupBuilder.cleanup()
+    })
+
+    it('should add custom timings', () => {
+      rumGlobal.init(DEFAULT_INIT_CONFIGURATION)
+      ;(rumGlobal as any).addTiming('foo') // tslint:disable-line
```
bcaudan, line 418:
```suggestion
      // tslint:disable-next-line: no-unsafe-any
      ;(rumGlobal as any).addTiming('foo')
```
to be more precise

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -385,4 +389,35 @@ describe('rum entry', () => {
...
+
+    beforeEach(() => {
+      addTimingSpy = jasmine.createSpy()
+      errorSpy = spyOn(console, 'error')
+      rumGlobal = makeRumPublicApi(() => ({
+        ...noopStartRum(),
+        addTiming: addTimingSpy,
+        configuration: ({
+          isEnabled: () => true,
+        } as any) as Configuration,
```
bcaudan, line 407:
```suggestion
      const configuration: Partial<Configuration> = {
        isEnabled: () => true,
      }
      rumGlobal = makeRumPublicApi(() => ({
        ...noopStartRum(),
        addTiming: addTimingSpy,
        configuration: configuration as Configuration,
```

To help catch potential type issue for fields use in the configuration stub.
Same strategy could be applied in similar cases.

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -50,7 +54,7 @@ describe('viewCollection', () => {
         loadEvent: 10,
       },
     }
-    lifeCycle.notify(LifeCycleEventType.VIEW_UPDATED, view as View)
+    lifeCycle.notify(LifeCycleEventType.VIEW_UPDATED, (view as any) as View)
```
bcaudan, line 57: We tend to limit as much as possible the usages of any
Could we do something better?

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -59,6 +59,7 @@ export interface RawRumViewEvent {
     firstContentfulPaint?: number
     firstInputDelay?: number
     cumulativeLayoutShift?: number
+    customTimings: ViewCustomTimings
```
bcaudan, line 62: This field should be added in rum-events-format as well to update rumEvent.types

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.ts
```diff
@@ -23,6 +23,7 @@ function processViewUpdate(view: View) {
         count: view.eventCounts.userActionCount,
       },
       cumulativeLayoutShift: view.cumulativeLayoutShift,
+      customTimings: mapObject(view.customTimings, msToNs),
```
bcaudan, line 26: small optimisation, could we not add customTimings when empty?

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
webNeat: I fixed this by casting the `location` attribute to `Partial<Location>`

webNeat: This actually didn't work so I ended up using `window.location` to have the proper type.

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -1,11 +1,15 @@
...
 import { ActionType } from '../domain/rumEventsCollection/action/trackActions'
 import { makeRumPublicApi, RumPublicApi, RumUserConfiguration, StartRum } from './rumPublicApi'
 
 const noopStartRum = () => ({
   addAction: () => undefined,
   addError: () => undefined,
+  addTiming: () => undefined,
+  configuration: ({
+    isEnabled: () => false,
+  } as any) as Configuration,
```
bcaudan, line 12: cf previous comment, any could be removed here too

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -23,8 +23,14 @@ describe('viewCollection', () => {
 
   it('should create view from view update', () => {
     const { lifeCycle, rawRumEvents } = setupBuilder.build()
+    const location: Partial<Location> = {}
     const view = {
+      location,
```
bcaudan, line 28:
```suggestion
      location: location as Location,
```
it would remove the need to cast view as View in the next statement

packages/rum/src/rumEvent.types.ts
```diff
@@ -0,0 +1,658 @@
+/* tslint:disable */
```
bcaudan, line 1: cf c822e93599360d59acce4a39362820074b8bea3b, it is not the right file

packages/rum-core/test/fixtures.ts
```diff
@@ -66,6 +66,7 @@ export function createRawRumEvent(type: RumEventType, overrides?: Context): RawR
           date: 0,
           view: {
             action: { count: 0 },
+            customTimings: {},
```
bcaudan, line 69: is it needed?

bcaudan (approved):

PR #674, opened 2021-01-07
👷 fix build bundle

BenoitZugmeyer:
## Motivation

The staging build is broken: `yarn build:bundle` fails because the rum-recorder webpack build can't find the `@datadog/browser-rum-core` package (transitive dependency).

For some reason, `yarn build` works though.

## Changes

* read paths from `tsconfig.base.json` when building with webpack: this avoids having to redeclare paths in each packages
* add a ci job to check for `yarn build:bundle`

## Testing

CI

---

### DIFF
--- a/packages/rum-recorder/tsconfig.json
+++ b/packages/rum-recorder/tsconfig.json
@@ -1,9 +1,0 @@
-{
-  "extends": "../../tsconfig.base.json",
-  "compilerOptions": {
-    "baseUrl": ".",
-    "outDir": "./bundle/",
-    "module": "es6"
-  },
-  "include": ["src", "test"]
-}
--- a/packages/rum-recorder/tsconfig.json
+++ b/packages/rum-recorder/tsconfig.json
@@ -3,11 +3,7 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./bundle/",
-    "module": "es6",
-    "paths": {
-      "@datadog/browser-core": ["../core/src"],
-      "@datadog/browser-rum": ["../rum/src"]
-    }
+    "module": "es6"
   },
   "include": ["src", "test"]
 }
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "baseUrl": ".",
     "experimentalDecorators": true,
     "esModuleInterop": true,
     "importHelpers": true,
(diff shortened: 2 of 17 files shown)


packages/rum-recorder/tsconfig.json
```diff
@@ -3,11 +3,7 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./bundle/",
-    "module": "es6",
-    "paths": {
-      "@datadog/browser-core": ["../core/src"],
-      "@datadog/browser-rum": ["../rum/src"]
```
bcaudan, line 9: The issue could have been fixed by adding
```
"@datadog/browser-rum-core": ["../rum-core/src"]
```

BenoitZugmeyer: Yes, but with my proposed solution, we won't have this kind of issue anymore in the future. I think it simplifies things a little.

bcaudan: agree, I think we could simplify a bit more by:
- removing rum-core tsconfig.json, since we don't build a bundle for this one (like core) 
- extract a tsconfig.bundle.json that tsconfig.json files would extend, since they are all the same now

BenoitZugmeyer: 👌, I went one step further by replacing all `tsconfig` files used by webpack with a direct usage of `tsconfig.base.json` + a config override, it seems even simpler. I checked, we get the exact same bundles.

tsconfig.base.json
```diff
@@ -9,15 +9,18 @@
...
         "name": "typescript-tslint-plugin"
       }
     ],
+
     "paths": {
       "@datadog/browser-core": ["./packages/core/src"],
       "@datadog/browser-rum-core": ["./packages/rum-core/src"],
-      "@datadog/browser-rum": ["./packages/rum/src"]
+      "@datadog/browser-rum": ["./packages/rum/src"],
+      "@datadog/browser-rum-recorder": ["./packages/rum-recorder/src"]
```
bcaudan, line 23: do we need recorder here?

BenoitZugmeyer: We don't, I just added it for completeness. Do you think I should remove it?

bcaudan: Not necessarily, but for completeness we should also add logs

bcaudan: On second thoughts, once rum-recorder will use rum-core, we would only need core-and rum-core here.
Maybe targeting that would be better, not sure if completeness add some value

BenoitZugmeyer: Agreed! I removed it.

bcaudan (approved):
Nice!

PR #675, opened 2021-01-07
👷 fix build bundle (continuation)

BenoitZugmeyer:
## Motivation

Since `rum-recorder` is a new bundle, we don't have to support the legacy bundle with a suffix, so I didn't add the extra logic to generate it in the webpack configuration. The deploy job expect to have it though, and tries to add a comment banner on a non-suffixed file, which is unexpected and fails.

## Changes

Don't build rum-recorder bundle with a suffix.

## Testing

---

### DIFF
--- a/packages/logs/webpack.config.js
+++ b/packages/logs/webpack.config.js
@@ -2,18 +2,10 @@ const path = require('path')
 
 const webpackBase = require('../../webpack.base')
 
-const datacenter = process.env.TARGET_DATACENTER || 'us'
-const withSuffix = process.env.WITH_SUFFIX || false
+const entry = path.resolve(__dirname, 'src/boot/logs.entry.ts')
 
-const suffix = withSuffix ? `-${datacenter}` : ''
-
-module.exports = (env, argv) => ({
-  entry: {
-    logs: path.resolve(__dirname, 'src/boot/logs.entry.ts'),
-  },
-  ...webpackBase(argv.mode),
-  output: {
-    filename: `datadog-logs${suffix}.js`,
-    path: path.resolve(__dirname, 'bundle'),
-  },
-})
+module.exports = (env, argv) => [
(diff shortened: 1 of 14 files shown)


packages/rum-recorder/package.json
```diff
@@ -7,7 +7,7 @@
   "types": "cjs/index.d.ts",
   "scripts": {
     "build": "run-p build:cjs build:esm build:bundle",
-    "build:bundle": "rm -rf bundle && webpack --config webpack.config.js --mode=production && yarn replace-build-env bundle",
+    "build:bundle": "if [[ $WITH_SUFFIX != true ]]; then rm -rf bundle && webpack --config webpack.config.js --mode=production && yarn replace-build-env bundle; fi",
```
bcaudan, line 10: what about moving `build:bundle` scripts to a cli command?
it could help to document this behavior

packages/logs/webpack.config.js
```diff
@@ -2,18 +2,10 @@ const path = require('path')
...
-  ...webpackBase(argv.mode),
-  output: {
-    filename: `datadog-logs${suffix}.js`,
-    path: path.resolve(__dirname, 'bundle'),
-  },
-})
+module.exports = (env, argv) => [
+  webpackBase({ mode: argv.mode, entry, datacenter: 'us', filename: 'datadog-logs.js' }),
+  webpackBase({ mode: argv.mode, entry, datacenter: 'us', filename: 'datadog-logs-us.js' }),
+  webpackBase({ mode: argv.mode, entry, datacenter: 'eu', filename: 'datadog-logs-eu.js' }),
```
bcaudan, line 10: Since us/eu files are now built every time, they will end up in rum and logs npm packages.
I think it could be nice to not build them in this case or at least exclude them from the npm package.

bcaudan (approved):
Great!

PR #677, opened 2021-01-08
[RUMF-810] force alternate intake for us3

bcaudan:
## Motivation

Only alternate intake will be available for new data centers.
us3 alternate intake is now available.

## Changes

Temporary strategy to force alternate intake for us3.
When new intake will be available for gov, we should only allow classic intake for us and eu.

## Testing

Automated tests + us3 test

---


packages/core/src/domain/configuration.ts
```diff
@@ -256,6 +256,11 @@ function getHost(intakeType: IntakeType, endpointType: EndpointType, site: strin
   return `${endpoint}.browser-intake-${suffix}`
 }
 
+function getIntakeType(site: string, userConfiguration: UserConfiguration) {
+  // TODO when new intake will be available for gov, only allow classic intake for us and eu
```
BenoitZugmeyer, line 260: Couldn't we do this already? something like:
```ts
if (!userConfiguration.useAlternateIntakeDomains && (site === 'datadoghq.com' || site === 'datadoghq.eu')) {
  return 'classic'
}
return 'alternate'
```

webNeat (approved):
LGTM

BenoitZugmeyer (commented):
It seems fine, but I want to point to a related commit I had to do in rum-recorder: https://github.com/DataDog/browser-sdk/pull/670/commits/00cd89ec01c96d2bf89513a552d4440cabf5d339

It seems that our intake host logic gets more complex, and I wonder if we shouldn't centralize/colocate this logic.

packages/core/src/domain/configuration.ts
bcaudan: this would not allow classic intake for gov or am I missing something?

BenoitZugmeyer: I'm not sure about the specific rules we should apply, but it seems that new datacenters won't have classic intake anymore. Thus, I think that it would be better if we had a list of sites that *support* classic intakes (since it is not likely to change in the future), instead of a list of sites that *don't support* classic intakes (since this list is likely to grow when new datacenters are added).

bcaudan: It is what I have in mind too but I'd prefer to not break rum on dd gov app with this PR.
We could implement this logic now and define gov as supporting classic but we would still need an extra PR to remove it. 
no strong opinion about one way or the other.

> It seems that our intake host logic gets more complex, and I wonder if we shouldn't centralize/colocate this logic.

agree, we could wait for the PR for gov since this part should not change too much in the mean time and it would avoid to deal with conflicts for now.

BenoitZugmeyer: I'm fine either way. We should not take conflicts into account: my PR doesn't change the same part of `configuration.ts` so we shouldn't have any conflict. We can think again how to improve on this once everything gets merged.

BenoitZugmeyer (approved):

PR #680, opened 2021-01-11
👷 [RUMF-818] fix unit tests code coverage

BenoitZugmeyer:
## Motivation

Code coverage have been broken since 8a55636ed3d8d8bb7e79a23a05696c60667db665.

## Changes

Invert ts and string-replace loaders in webpack configuration. Changing the loader order fixes the code coverage during unit tests.  I investigated deeply into istanbul and string-replace-loader,  I have a gut feeling that it is because string-replace-loader does not support source maps, but can't confirm it.

## Testing

CI, https://codecov.io/gh/DataDog/browser-sdk/tree/546c12eaf46d48f9f802dbf62458c249b34331f4/packages

---

### DIFF
--- a/scripts/codecov.sh
+++ b/scripts/codecov.sh
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
 
+if ! grep -q '"s":{[^}]\+}' coverage/coverage-final.json; then
+  echo "Error: empty code coverage"
+  exit 1
+fi
+
 export CODECOV_TOKEN=$(aws ssm get-parameter --region us-east-1 --name ci.browser-sdk.codecov_token --with-decryption --query "Parameter.Value" --out text)
 yarn codecov -t "${CODECOV_TOKEN}" -f coverage/coverage-final.json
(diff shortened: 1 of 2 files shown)


webpack.base.js
```diff
@@ -16,6 +16,18 @@ module.exports = ({ entry, mode, filename, datacenter }) => ({
   devtool: mode === 'development' ? 'inline-source-map' : 'false',
```
bcaudan, line 16: It is not the first time that we [break it](https://github.com/DataDog/browser-sdk/pulls?q=is%3Apr+fix+coverage+in%3Atitle+), I am wondering if we could have some kind of check to prevent a regression here again

bcaudan (approved):
LGTM

scripts/codecov.sh
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
 
+if ! grep -q '"s":{[^}]\+}' coverage/coverage-final.json; then
```
bcaudan, line 3: nice!
Any idea what is "s"?

BenoitZugmeyer: covered **S**tatements

bcaudan (approved):

PR #683, opened 2021-01-15
[RUMF-812] remove mousemove batching from RRWeb

BenoitZugmeyer:
## Motivation

Mouse positions occuring during a segment should not leak to a next segment.  Batching should be done at the segment level.

## Changes

Remove mousemove batching from RRWeb

## Testing

Manual testing: replays mouse position should be smoother

---

### DIFF
--- a/packages/rum-recorder/src/domain/rrweb/observer.ts
+++ b/packages/rum-recorder/src/domain/rrweb/observer.ts
@@ -22,7 +22,6 @@ import {
   MouseInteractionCallBack,
   MouseInteractions,
   MousemoveCallBack,
-  MousePosition,
   MutationCallBack,
   ObserverParam,
   SamplingStrategy,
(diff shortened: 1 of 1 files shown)


packages/rum-recorder/src/domain/rrweb/observer.ts
```diff
@@ -74,34 +73,17 @@ function initMoveObserver(cb: MousemoveCallBack, sampling: SamplingStrategy): Li
...
+      const position = {
         id: mirror.getId(target as INode),
-        timeOffset: Date.now() - timeBaseline,
+        timeOffset: 0,
         x: clientX,
         y: clientY,
-      })
-      wrappedCb(isTouchEvent(evt))
+      }
+      cb([position], isTouchEvent(evt) ? IncrementalSource.TouchMove : IncrementalSource.MouseMove)
```
BenoitZugmeyer, line 86: We don't want to change the event model (yet) since it is a bit experimental and would require changes to the player too. Let's still emit an array, containing a single mouse position.

webNeat (approved):
LGTM

bcaudan (approved):

PR #684, opened 2021-01-19
[RUMF-824] support Request instances in tracing

BenoitZugmeyer:
## Motivation

Support `Request` instances in tracing
```js
fetch(
  new Request('http://localhost:3000', {
    headers: {
      'x-foo': 'bar',
    },
  })
)
```

## Changes

Clone instance and add trace headers if the fetch init argument is a `Request`.

## Testing

Unit

---

### DIFF
--- a/packages/rum-core/src/domain/tracing/tracer.spec.ts
+++ b/packages/rum-core/src/domain/tracing/tracer.spec.ts
@@ -204,12 +204,12 @@ describe('tracer', () => {
       const tracer = startTracer(configuration as Configuration)
       tracer.traceFetch(context)
 
-      expect(context.init!.headers).toEqual([
+      expect(context.init).toBe(undefined)
+      expect(context.input).not.toBe(request)
+      expect(headersAsArray((context.input as Request).headers)).toEqual([
         ['foo', 'bar'],
         ...tracingHeadersAsArrayFor(context.traceId!, context.spanId!),
       ])
-
-      expect(context.input).toBe(request)
       expect(headersAsArray(request.headers)).toEqual([['foo', 'bar']])
     })
 
@@ -319,7 +319,7 @@ function tracingHeadersFor(traceId: TraceIdentifier, spanId: TraceIdentifier) {
 }
 
 function tracingHeadersAsArrayFor(traceId: TraceIdentifier, spanId: TraceIdentifier) {
-  return objectEntries(tracingHeadersFor(traceId, spanId)) as string[][]
+  return objectEntries(tracingHeadersFor(traceId, spanId)) as Array<[string, string]>
 }
 
 function headersAsArray(headers: Headers) {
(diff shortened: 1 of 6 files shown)


packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -28,22 +28,29 @@ export function startTracer(configuration: Configuration): Tracer {
...
-            headers.push([key, value])
-          })
-        } else if (Array.isArray(context.init.headers)) {
-          context.init.headers.forEach((header) => {
-            headers.push(header)
-          })
-        } else if (context.init.headers) {
-          Object.keys(context.init.headers).forEach((key) => {
-            headers.push([key, (context.init!.headers as Record<string, string>)[key]])
+        if (context.init instanceof Request) {
```
bcaudan, line 31: `context.init` match `fetch` optional second parameter:

```ts
fetch(input: RequestInfo, init?: RequestInit)

type RequestInfo = Request | string
```

FMU, this condition will never be true IRL

packages/rum-core/src/domain/tracing/tracer.spec.ts
```diff
@@ -189,6 +189,48 @@ describe('tracer', () => {
       ])
     })
 
+    it('should preserve original headers contained in a Request instance', () => {
```
bcaudan, line 192: it could be nice to have a test with a real fetch to ensure that this strategy works (cf previous implementation)

packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -42,6 +42,10 @@ export function startTracer(configuration: Configuration): Tracer {
           Object.keys(context.init.headers).forEach((key) => {
             headers.push([key, (context.init!.headers as Record<string, string>)[key]])
           })
+        } else if (context.input instanceof Request) {
+          context.input.headers.forEach((value, key) => {
+            headers.push([key, value])
+          })
         }
         context.init.headers = headers.concat(objectEntries(tracingHeaders) as string[][])
```
bcaudan, line 50: It seems a bit weird to me to not update input headers but to provide an init headers instead.
I am wondering if it could not be better to provide an updated input only:
```ts
context.input = new Request(request, {headers: headers.concat(objectEntries(tracingHeaders) as string[][])})
```
no strong argument for one or the other though.
wdyt?

BenoitZugmeyer: Mmh it seems to be equivalent, since calling [`fetch(input, init)` calls `new Request(input, init)` internally](https://fetch.spec.whatwg.org/#fetch-method). After a few tests, I think that it should be fine. I only spoted an issue with [streaming requests](https://web.dev/fetch-upload-streaming/) but it is likely to be a Chrome bug in the ongoing implementation:

```ts
const request = new Request('/post', {
  method: 'POST',
  body: new ReadableStream(),
  mode: 'cors'
});
// works:
await fetch(request, { mode: 'cors' })
// fails with: TypeError: Failed to construct 'Request': If request is made from ReadableStream, mode should be"same-origin" or "cors"
const requestCopy = new Request(request)
await fetch(requestCopy, { mode: 'cors' })
```

I think it's fine nonetheless :)

bcaudan (approved):
👍

PR #686, opened 2021-01-20
✨ Implement a developer extension

BenoitZugmeyer:
## Motivation

When working on the Browser SDK project or its integration in websites, it may be tough to see what is actually sent to the server or use local bundles on production websites.

## Changes

Implement a browser extension to address developer needs. A few features are implemented, but more could be added in the future.

## Future features

* configuration page to set a list of custom intakes or CDN URLs
* log events as soon as they are generated by the SDK (needs SDK change)
* inject the SDK on any page
* log Replay records
* devtools tab integration + activate the extension only on tabs where devtools is open

---


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 // This tsconfig is only used for tooling (ex: typecheck in code editors)
 {
   "extends": "./tsconfig.base.json",
-  "include": ["./packages/**/src/**/*.ts", "./packages/**/test/**/*.ts"]
+  "include": ["./packages/**/src", "./packages/**/test"],
+  "exclude": ["./packages/developer-extension"]
```
BenoitZugmeyer, line 5: The developer extension tsconfig is not entirely compatible with the base tsconfig, notably it doesn't support `strictNullChecks`, so it needs to be kept appart from the root tsconfig.

apigirl (commented):
no docs team review needed

packages/developer-extension/src/background/domain/flushEvents.ts
```diff
@@ -0,0 +1,13 @@
...
+import { evaluateCodeInActiveTab } from '../utils'
+
+listenAction('flushEvents', () =>
+  // Simulates a brief page visibility change (visible > hide > visible)
+  evaluateCodeInActiveTab(`
+    const descriptor = Object.getOwnPropertyDescriptor(Document.prototype, 'visibilityState')
+    Object.defineProperty(Document.prototype, 'visibilityState', { value: 'hidden' })
+    document.dispatchEvent(new Event('visibilitychange', { bubbles: true }))
+    Object.defineProperty(Document.prototype, 'visibilityState', descriptor)
+    document.dispatchEvent(new Event('visibilitychange', { bubbles: true }))
```
webNeat, line 11: I would prefer to not write code as string, what about the following
```ts
const flushEvents = () => {
    const descriptor = Object.getOwnPropertyDescriptor(Document.prototype, 'visibilityState')
    Object.defineProperty(Document.prototype, 'visibilityState', { value: 'hidden' })
    document.dispatchEvent(new Event('visibilitychange', { bubbles: true }))
    Object.defineProperty(Document.prototype, 'visibilityState', descriptor)
    document.dispatchEvent(new Event('visibilitychange', { bubbles: true }))
}
evaluateCodeInActiveTab(`;(${flushEvents})()`)
```

packages/developer-extension/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
webNeat, line 1: Since this extension will not be published as an npm package, I would move it out of the `packages` directory. What about `extensions/chrome`? That way if we have other browser extensions in the future we can store them into the `extensions` directory too.

packages/developer-extension/src/background/domain/replaceBundles.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+    return
+  },
+  {
+    types: ['script'],
+    urls: [
+      // TODO: implement a configuration page to add more URLs in this list.
+      'https://www.datadoghq-browser-agent.com/datadog-logs.js',
+      'https://www.datadoghq-browser-agent.com/datadog-rum.js',
+      'https://www.datadoghq-browser-agent.com/datadog-rum-recorder.js',
```
webNeat, line 30: What if the SDK is configured to use a different intake domain? Is it possible generate these URLs based on the actual RUM and logs configurations?

packages/developer-extension/src/popup/useStore.ts
```diff
@@ -0,0 +1,35 @@
...
+    resolve()
+  })
+})
+
+export function useStore(): [Store, (newState: Partial<Store>) => void] {
+  if (!store) {
+    throw storeLoadingPromise
+  }
+
+  const forceUpdate = useReducer(() => ({}), {})[1] as () => void
```
webNeat, line 21: This line is a bit hard to read, what about rewriting it as?
```ts
  const reducer = useReducer(() => ({}), {})
  const forceUpdate = reducer[1] as () => void
```

webNeat (changes requested):
Looking good, just some remarks.

packages/developer-extension/src/background/domain/flushEvents.ts
bcaudan: or even:
```ts
evaluateCodeInActiveTab(flushEvents)
```
and let the evaluate function deal with the stringification

packages/developer-extension/src/background/domain/logRequests.ts
```diff
@@ -0,0 +1,64 @@
+import { store } from '../store'
+
+const decoder = new TextDecoder('utf-8')
+chrome.webRequest.onBeforeRequest.addListener(
+  (info) => {
+    if (!store.logEvents) {
```
bcaudan, line 6: some inconsistencies between `logRequests` and `logEvents` namings, what about `logEventsFromRequests` instead?

bcaudan, line 6: It would be more convenient to log events as soon as they occur. In previous hackathon, we triggered events from sdk code, cf [branch](https://github.com/DataDog/browser-sdk/compare/hack-a-dog-chrome-extension#diff-c04d6d921a5dcb5a54e9ae53e0e79810a60da0ae380b288ee6526c8b75bf2b78R7).
It could be a future improvement though

packages/developer-extension/src/background/domain/logRequests.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+  },
+  {
+    urls: [
+      // TODO: implement a configuration page to add more URLs in this list.
+      'https://*.logs.datadoghq.com/*',
+      'https://*.browser-intake-datadoghq.com/*',
+      'https://*.logs.datadoghq.eu/*',
+      'https://*.browser-intake-datadoghq.eu/*',
```
bcaudan, line 42: what about listing other prod data centers?

bcaudan, line 42: *.logs.datadoghq.com is also intercepting non sdk requests (ex: livetail), what about being a bit more restrictive?

packages/developer-extension/src/background/domain/replaceBundles.ts
```diff
@@ -0,0 +1,51 @@
+import { listenAction } from '../actions'
+import { DEV_LOGS_URL, DEV_RUM_RECORDER_URL, DEV_RUM_URL } from '../constants'
+import { setStore, store } from '../store'
+
+chrome.webRequest.onBeforeRequest.addListener(
```
bcaudan, line 5: This does not work well when bundles are cached by browser.
what about either:
- adapt the strategy to work even when bundles are cached
- add a "full reload may be needed" on the UI

packages/developer-extension/src/background/store.ts
```diff
@@ -0,0 +1,30 @@
...
+  logEvents: true,
+  useDevBundles: false,
+  useRumRecorder: false,
+}
+
+export function setStore(newStore: Partial<Store>) {
+  if (wouldModifyStore(newStore)) {
+    Object.assign(store, newStore)
+    sendAction('newStore', store)
+    chrome.storage.local.set({ store })
```
bcaudan, line 15: Having a single store for all tabs could lead to extension state displayed nnot being consistent with tab state.
What about managing the state of each tab separately?

packages/developer-extension/src/popup/useStore.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  useEffect(() => {
+    storeListeners.add(forceUpdate)
+    return () => {
+      storeListeners.delete(forceUpdate)
+    }
+  }, [])
```
bcaudan, line 28: Why do we need to do that?

packages/developer-extension/src/background/domain/flushEvents.ts
BenoitZugmeyer: Ok. I tend to avoid such pattern as it is not obvious that they will be used as a string, but I'll make the change anyway

packages/developer-extension/package.json
BenoitZugmeyer: I'm fine with moving the extension in another folder (@bcaudan WDYT?), but I wouldn't call the folder `chrome` since we should not need a separate extension for each browsers (Firefox and Safari are using WebExtensions too, so they can have the same codebase)

packages/developer-extension/src/background/domain/replaceBundles.ts
BenoitZugmeyer: This list is not intake domains, but I agree it would be great to be able to configure bundle URLs and intake lists. I'll add a list of feature ideas somewhere.

I'm not sure how to discover such URLs based on page the RUM and Logs configurations though.

packages/developer-extension/package.json
bcaudan: I'm fine with both, I have a slight preference for having it outside of packages too, since it is not sdk code (a bit like e2e tests code).
I agree that chrome is not a suitable name.

packages/developer-extension/src/background/domain/logRequests.ts
BenoitZugmeyer: I like to see events when they are sent, from an external point of view, to observe what the SDK does instead of what the SDK tells us. Also printing events when they are generated could be misleading because not all events are sent (view updates could be discarded). This could be configured though.

packages/developer-extension/src/background/domain/logRequests.ts
BenoitZugmeyer: Do we have a list somewhere?

packages/developer-extension/src/background/domain/logRequests.ts
bcaudan: both can indeed be helpful

packages/developer-extension/src/background/domain/replaceBundles.ts
BenoitZugmeyer: I'll see what I can do, it seems that there is [an API for it](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browsingData/removeCache)

packages/developer-extension/src/background/store.ts
BenoitZugmeyer: Sure, I designed the state of the extension to be global, not local to the tab. But I can change this.

packages/developer-extension/src/popup/useStore.ts
BenoitZugmeyer: Yes, to update the component when the store changes

packages/developer-extension/src/background/store.ts
BenoitZugmeyer: So, I'd like to keep this as it is for now: I like to have a persistent configuration (accross browser restarts), and using tab ids to store the state won't allow this.

In the future we may reconsider this. It could be nice to use a devtools panel to display SDK related stuff and extension configuration. In this case, I think it would make sense to apply the extension when the devtools are open, and have its effect apply only on the related tab.

bcaudan: sure

bcaudan (approved):
Great!

webNeat (approved):

PR #689, opened 2021-01-20
🐛[RUMF-809] URL encode tags in intake requests

bcaudan:
## Motivation

Ensure that tags value are properly interpreted on backend side.
Note: most of the special characters are converted to underscore anyway, cf [tags doc](https://docs.datadoghq.com/getting_started/tagging/#defining-tags).

## Changes

URL encode value of `tags` query param in intake requests 

## Testing

Unit + manual test

---

### DIFF
--- a/packages/core/src/domain/configuration.spec.ts
+++ b/packages/core/src/domain/configuration.spec.ts
@@ -83,6 +83,13 @@ describe('configuration', () => {
     })
   })
 
+  describe('tags', () => {
+    it('should be encoded', () => {
+      const configuration = buildConfiguration({ clientToken, service: 'bar+foo' }, usEnv)
+      expect(configuration.rumEndpoint).toContain(`ddtags=sdk_version%3Asome_version%2Cservice%3Abar%2Bfoo`)
+    })
+  })
+
   describe('cookie options', () => {
     it('should not be secure nor crossSite by default', () => {
       const configuration = buildConfiguration({ clientToken }, usEnv)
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/configuration.spec.ts
```diff
@@ -62,22 +62,22 @@ describe('configuration', () => {
   describe('sdk_version, env, version and service', () => {
     it('should not modify the logs and rum endpoints tags when not defined', () => {
       const configuration = buildConfiguration({ clientToken }, usEnv)
-      expect(configuration.rumEndpoint).toContain(`&ddtags=sdk_version:${usEnv.sdkVersion}`)
+      expect(decodeURIComponent(configuration.rumEndpoint)).toContain(`&ddtags=sdk_version:${usEnv.sdkVersion}`)
```
BenoitZugmeyer, line 65: Decoding the whole URL with `decodeURIComponent` seems weird to me. Could we avoid it by looking for encoded strings instead?

```suggestion
      expect(configuration.rumEndpoint).toContain(`&ddtags=sdk_version%3A${usEnv.sdkVersion}`)
```

Else, we might want to add a test to check if tags are correctly encoded.

bcaudan: having the encoded strings make the assertions not so lisible, I'll add another test on tags correctly encoded

BenoitZugmeyer (approved):

PR #690, opened 2021-01-20
[RUMF-813] take full snapshots when the view changes or the session is renewed

BenoitZugmeyer:
## Motivation

* Replays are watched by session. If a session doesn’t start with a full snapshot, it cannot be played from the start.

* Records can happen in multiple tabs simultatenously. This causes an issue during replay, because the player gets incremental snapshot but doesn’t know on which fullsnapshot it should apply. By starting all views with a full snapshot, the player can drop segments with a different view from the previous one but without full snapshot. This an easy workaround that can be improved in the future. 

## Changes

* Add the ability to manually trigger full snapshots in RRWeb
* Use it via LifeCycle events

## Testing

Unit + manual
 
---

### DIFF
--- a/packages/rum-recorder/src/boot/recorder.spec.ts
+++ b/packages/rum-recorder/src/boot/recorder.spec.ts
@@ -135,6 +135,35 @@ describe('startRecording', () => {
       expectNoExtraRequest(done)
     })
   })
+
+  it('takes a full snapshot when the view changes', (done) => {
+    const { lifeCycle } = setupBuilder.build()
+
+    lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {} as any)
+
+    flushSegment(lifeCycle)
+
+    waitRequests(2, (requests) => {
+      expect(requests[0].data.get('has_full_snapshot')).toBe('true')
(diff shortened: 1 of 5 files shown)


packages/rum-recorder/src/boot/recorder.spec.ts
```diff
@@ -135,6 +135,35 @@ describe('startRecording', () => {
...
+
+  it('takes a full snapshot when the view changes', (done) => {
+    const { lifeCycle } = setupBuilder.build()
+
+    lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {} as any)
+
+    flushSegment(lifeCycle)
+
+    waitRequests(2, (requests) => {
+      expect(requests[0].data.get('has_full_snapshot')).toBe('true')
```
bcaudan, line 147: FMU, the first request corresponds to the segment that start at the beginning of the recording and end with the creation of the new view.
If that is it, do we need to ensure that it has a full snapshot in this test?

BenoitZugmeyer: No, I'll remove it.

bcaudan (approved):
LGTM

PR #692, opened 2021-01-21
replay: rename block & ignore classes

vlad-mh:
## Motivation

The defaults (`rr-block` and `rr-ignore`) don't really make sense for a user of the SDK. `rr`, for `rrweb`, is an implementation detail from the user's POV.

## Changes

Renames the Session Replay `blockClass` & `ignoreClass`.

I propose:
- `dd-hide`: elements that will be obfuscated (a placeholder is used).
- `dd-ignore-input`: elements what will have their input events ignored.


vlad-mh:
Closing in favor of a new PR to use data attributes

PR #697, opened 2021-01-22
[recorder] set 'hasReplay' to undefined by default

BenoitZugmeyer:
## Motivation

`session.has_replay` should not be set if the record has not been started, to be consistent with the RUM bundle without recorder.

This change only affects the recorder with flag `postpone_start_recording`.

## Changes

Set hasReplay to undefined by default.
Adjust typings to enforce this.

## Testing

Unit

---


bcaudan (approved):
LGTM

ThibautGeriz (approved):

PR #699, opened 2021-01-22
 ♻️ Remove automatic snake case

bcaudan:
## Motivation

Our properties need to be snake cased but not customer properties.
Automatic conversion + exception for customer properties can be error prone if we forget to add exception when we add new customer properties.

## Changes

- Declare snake cased properties in RawRumEvents
- Remove automatic snake case logic

## Testing

Unit

---


packages/rum-core/src/domain/internalContext.ts
```diff
@@ -9,17 +8,24 @@ import { RumSession } from './rumSession'
...
+          user_action: actionContext
+            ? {
+                id: actionContext.action.id,
+              }
+            : undefined,
+          view: {
+            id: viewContext.view.id,
+            referrer: viewContext.view.referrer,
+            url: viewContext.view.url,
+          },
```
BenoitZugmeyer, line 27: NIT:

```suggestion
          view: viewContext.view,
```

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
```diff
@@ -24,21 +24,22 @@ export function startActionCollection(lifeCycle: LifeCycle, configuration: Confi
...
+          loading_time: msToNs(action.duration),
+          long_task: {
             count: action.counts.longTaskCount,
           },
           resource: {
             count: action.counts.resourceCount,
           },
         },
-      }
+      } as Partial<RawRumActionEvent>)
```
BenoitZugmeyer, line 42: Do we need this cast?

BenoitZugmeyer (approved):
Looks great!

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
bcaudan: I've added it because ts was not complaining with the old format.
It compiles without it but then autoActionProperties could be anything.
wdyt?

BenoitZugmeyer: FMU, the issue was because loadingTime and longTask are optional properties. What about using an approach similar to https://medium.com/terria/typescript-transforming-optional-properties-to-required-properties-that-may-be-undefined-7482cb4e1585 ? I guess it could be useful at other places too.

```diff
diff --git packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
index 201f0f9f..08a712f7 100644
--- packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
+++ packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
@@ -22,10 +22,13 @@ export function startActionCollection(lifeCycle: LifeCycle, configuration: Confi
   }
 }

+type Complete<T> = {
+  [P in keyof Required<T>]: Pick<T, P> extends Required<Pick<T, P>> ? Complete<T[P]> : Complete<T[P]> | undefined
+}
+
 function processAction(action: AutoAction | CustomAction) {
   const autoActionProperties = isAutoAction(action)
-    ? // tslint:disable-next-line: no-object-literal-type-assertion
-      ({
+    ? {
         action: {
           error: {
             count: action.counts.errorCount,
@@ -39,10 +42,10 @@ function processAction(action: AutoAction | CustomAction) {
             count: action.counts.resourceCount,
           },
         },
-      } as Partial<RawRumActionEvent>)
+      }
     : undefined
   const customerContext = !isAutoAction(action) ? action.context : undefined
-  const actionEvent: RawRumActionEvent = combine(
+  const actionEvent: Complete<RawRumActionEvent> = combine(
     {
       action: {
         target: {
```

bcaudan: I don't think that the issue is related to optional properties, for ex with:
```ts
action: {
  foo: 'bar',
  error: {
    count: action.counts.errorCount,
  },
  ...
}
``` 
I would expect a ts issue since `foo` is not defined on `RawRumActionEvent` but it is not the case.

Anyway, since the automatic snake case has been removed, this error is as least caught by unit tests.
I've removed the explicit casting.

BenoitZugmeyer (approved):

PR #700, opened 2021-01-22
rum-recorder: import rrweb-snapshot code

vlad-mh:
This PR removes the dependency on rrweb-snapshot and instead adds it's code directly, changing the imports accordingly.


packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -0,0 +1,589 @@
+/* eslint-disable no-underscore-dangle */
```
vlad-mh, line 1: While I removed some (e.g: I renamed `_id` to `nextId`), others appeared to be a more dangerous change (e.g `attributes._cssText` here: https://github.com/DataDog/browser-sdk/pull/700/files#diff-1c7cef52f64a686ba523420f60390086b7a16d8f13b51ac8f0a3e79d90be827eR224)

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -0,0 +1,589 @@
...
+/* eslint-disable @typescript-eslint/prefer-regexp-exec */
```
vlad-mh, line 2: I wasn't sure on this.

Example: https://github.com/DataDog/browser-sdk/pull/700/files#diff-1c7cef52f64a686ba523420f60390086b7a16d8f13b51ac8f0a3e79d90be827eR349

While replacing a `'foo'.match()` to `/foo/.exec()` makes sense, here we're talking about creating a new regexp from the return value of `lowerIfExists`?

vlad-mh:
@BenoitZugmeyer this should be ready for review 🙏

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
BenoitZugmeyer: I think it simply means to change
```ts
lowerIfExists(sn.attributes.name).match(/^msapplication-tile(image|color)$/)
```
to
```ts
/^msapplication-tile(image|color)$/.exec(lowerIfExists(sn.attributes.name))
```
also we could take the opportunity to change `exec` to `test` when the result is used as a boolean

```ts
/^msapplication-tile(image|color)$/.test(lowerIfExists(sn.attributes.name))
```

LICENSE-3rdparty.csv
```diff
@@ -3,7 +3,7 @@ require,tslib,Apache-2.0,Copyright Microsoft Corporation
 file,pako,MIT,(C) 2014-2017 Vitaly Puzrin and Andrey Tupitsin 
 file,rrweb,MIT,Copyright (c) 2018 Contributors (https://github.com/rrweb-io/rrweb/graphs/contributors) and SmartX Inc.
 file,tracekit,MIT,Copyright 2013 Onur Can Cakmak and all TraceKit contributors
-prod,rrweb-snapshot,MIT,Copyright (c) 2018 Contributors (https://github.com/rrweb-io/rrweb-snapshot/graphs/contributors) and SmartX Inc.
+file,rrweb-snapshot,MIT,Copyright (c) 2018 Contributors (https://github.com/rrweb-io/rrweb-snapshot/graphs/contributors) and SmartX Inc.
```
BenoitZugmeyer, line 6: NIT: sort alphabetically

packages/rum-recorder/src/domain/rrweb-snapshot/types.ts
```diff
@@ -0,0 +1,97 @@
+export enum NodeType {
+  Document,
+  DocumentType,
+  Element,
+  Text,
+  CDATA,
+  Comment,
+}
+
+export type documentNode = {
```
BenoitZugmeyer, line 10: I find types starting with a lowercase letter to be kind of confusing. I changed the casing in `rrweb/types.ts`, could you do the same here?

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -0,0 +1,589 @@
...
+    origin = url.split('/')[0]
+  }
+  origin = origin.split('?')[0]
+  return origin
+}
+
+const URL_IN_CSS_REF = /url\((?:(')([^']*)'|(")([^"]*)"|([^)]*))\)/gm
+const RELATIVE_PATH = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/
+const DATA_URI = /^(data:)([^,]*),(.*)/i
+export function absoluteToStylesheet(cssText: string | null, href: string): string {
```
bcaudan, line 72: What about also importing unit tests?
cf https://github.com/rrweb-io/rrweb-snapshot/blob/master/test/snapshot.test.ts

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -0,0 +1,589 @@
...
+
+function getCssRuleString(rule: CSSRule): string {
+  return isCSSImportRule(rule) ? getCssRulesString(rule.styleSheet) || '' : rule.cssText
+}
+
+function isCSSImportRule(rule: CSSRule): rule is CSSImportRule {
+  return 'styleSheet' in rule
+}
+
+function extractOrigin(url: string): string {
```
bcaudan, line 58: could we replace it by [getOrigin(url: string)](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/tools/urlPolyfill.ts#L19)?

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -0,0 +1,589 @@
...
+        return `${absUrl}`
+      }
+      return ''
+    })
+    .join(', ')
+
+  return resultingSrcsetString
+}
+
+export function absoluteToDoc(doc: Document, attributeValue: string): string {
```
bcaudan, line 135: I am wondering if we could replace that by [normalizeUrl(url: string)](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/tools/urlPolyfill.ts#L3)

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
vlad-mh: `getOrigin` seem more complicated (builds an `a` element with `buildUrl`) compared to this implementation.
I'm not comfortable enough with JS/TS to guarantee the behavior is the same, so I'd rather not touch the code atm 🤔

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
vlad-mh: fair enough, although this adds a dependency on jsdom and chai, are we comfortable with that? I know @BenoitZugmeyer wanted to look at importing rrweb tests in another PR

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
vlad-mh: ah, I didn't get the string & regexp were inverted between match and exec. I'll make the change

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
BenoitZugmeyer: Yes, I'm working on it. We should not import chai or jsdom though:
* jsdom is used when running tests in Nodejs (through Jest). Since we use Karma, we can just use the real DOM
* chai is an assertion library, karma-jasmine provides one out of the box

BenoitZugmeyer: I think importing snapshot unit tests would be really straightforward as there is no snapshots or complex asynchronous cases. It could be worth it to import and adjust them (to use karma/jasmine) here.

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
bcaudan: sure, it could worth looking into that in a separate task though

BenoitZugmeyer: IMO for now it would make sense to keep the current implementation. As @vlad-mh stated, it is simpler and may be faster than `getOrigin`. Using it to transform big stylesheets seems a bit perf critical. Agreed to look into it in a future task.

packages/rum-recorder/src/domain/rrweb-snapshot/types.ts
```diff
@@ -0,0 +1,97 @@
...
+export type CommentNode = {
+  type: NodeType.Comment
+  textContent: string
+}
+
+export type SerializedNode = DocumentNode | DocumentTypeNode | ElementNode | TextNode | CDataNode | CommentNode
+
+export type SerializedNodeWithId = SerializedNode & { id: number }
+
+export type tagMap = {
```
BenoitZugmeyer, line 54: This type doesn't seem to be used. Could you remove it? else, change it to `TagMap`

BenoitZugmeyer (approved):

packages/rum-recorder/src/domain/rrweb-snapshot/types.ts
vlad-mh: good catch! latest commit removes it

BenoitZugmeyer (approved):

bcaudan (approved):
LGTM

PR #702, opened 2021-01-25
✨[RUMF-820] expose API to add custom timings to the current view

bcaudan:
## Motivation

Following [POC](https://github.com/DataDog/browser-sdk/pull/668), allow to add custom timing to the current view:
```ts
DD_RUM.addTiming('my_timing')
```

## Changes

- Rework API to only attach timing to current view
- Remove feature flag

## Testing

Unit + manual

---


webNeat (approved):

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -418,10 +411,19 @@ describe('rum entry', () => {
       setupBuilder.cleanup()
     })
 
+    it('should do nothing before init', () => {
```
BenoitZugmeyer, line 414: I expected to follow the same pattern as other APIs and set the timing even if it called before init. Is there a reason not to do this?

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.spec.ts
bcaudan: I thought it would not make sense but it could 🤔, I'll have a look

BenoitZugmeyer (approved):

PR #705, opened 2021-01-28
[RUMF-814] extract the "waitRequests" pattern as a util function

BenoitZugmeyer:
## Motivation

I want to abstract the pattern [introduced while working on the recorder tests](https://github.com/DataDog/browser-sdk/pull/670/commits/0dbceac30bdad343aced91277a9792280ad55877) to be used in other test specs. Since there is many ways of implementing this, I'd like to share this beforehand so we can agree on the best way of doing so.

## Changes

* define `expectNoExtraCalls` and `waitCalls` in the recorder `test/utils`
* use it in `recorder.spec.ts`. Because now only `jasmine.Calls` objects are exposed instead of transformed `{ data: FormData, size: number }` objects, a `getRequestData` helper  has been introduced locally.

## Testing

Unit

---

### DIFF
--- a/packages/rum-recorder/src/boot/recorder.spec.ts
+++ b/packages/rum-recorder/src/boot/recorder.spec.ts
@@ -3,7 +3,7 @@ import { LifeCycle, LifeCycleEventType } from '@datadog/browser-rum-core'
 
 import { setup, TestSetupBuilder } from '../../../rum-core/test/specHelper'
 
-import { expectNoExtraCall, waitCalls } from '../../test/utils'
+import { collectAsyncCalls } from '../../test/utils'
 import { startRecording } from './recorder'
 
 describe('startRecording', () => {
@@ -40,8 +40,10 @@ describe('startRecording', () => {
       )
 
     const requestSendSpy = spyOn(HttpRequest.prototype, 'send')
-    expectNoExtraRequestSendCalls = expectNoExtraCall.bind(null, requestSendSpy)
-    waitRequestSendCalls = waitCalls.bind(null, requestSendSpy)
+    ;({
+      waitAsyncCalls: waitRequestSendCalls,
+      expectNoExtraAsyncCall: expectNoExtraRequestSendCalls,
+    } = collectAsyncCalls(requestSendSpy))
   })
 
   afterEach(() => {
--- a/packages/rum-recorder/src/boot/recorder.spec.ts
+++ b/packages/rum-recorder/src/boot/recorder.spec.ts
@@ -3,16 +3,17 @@ import { LifeCycle, LifeCycleEventType } from '@datadog/browser-rum-core'
 
 import { setup, TestSetupBuilder } from '../../../rum-core/test/specHelper'
 
+import { expectNoExtraCall, waitCalls } from '../../test/utils'
 import { startRecording } from './recorder'
 
 describe('startRecording', () => {
   let setupBuilder: TestSetupBuilder
   let sessionId: string | undefined
-  let waitRequests: (
-    expectedRequestCount: number,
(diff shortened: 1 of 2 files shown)


packages/rum-recorder/test/utils.ts
```diff
@@ -51,3 +51,26 @@ export class MockWorker implements DeflateWorker {
     this.pendingMessages.length = 0
   }
 }
+
+export function waitCalls<F extends jasmine.Func>(
+  spy: jasmine.Spy<F>,
+  expectedCallsCount: number,
+  callback: (calls: jasmine.Calls<F>) => void
```
webNeat, line 58: I have a question: is it possible to make this return a `Promise<jasmine.Calls<F>>` instead of having a callback parameter?

BenoitZugmeyer: Our IE11 support forbids us to use promises, except in parts where IE11 is not supported anyway (like `fetch` proxy/tests)

webNeat (approved):

packages/rum-recorder/src/boot/recorder.spec.ts
```diff
@@ -39,22 +40,8 @@ describe('startRecording', () => {
...
-    }
-
-    expectNoExtraRequest = (done) => {
-      requestSendSpy.and.callFake(() => {
-        fail('Unexpected request received')
-      })
-      setTimeout(done, 300)
-    }
+    expectNoExtraRequestSendCalls = expectNoExtraCall.bind(null, requestSendSpy)
+    waitRequestSendCalls = waitCalls.bind(null, requestSendSpy)
```
bcaudan, line 44: Since these kinds of binds should be needed for every usage of this utility, what about having a single API that returns the two functions bounded?
something like:
```ts
const { waitCalls, expectNoExtraCall } = foo(spyOn(HttpRequest.prototype, 'send'))
```

packages/rum-recorder/test/utils.ts
```diff
@@ -51,3 +51,26 @@ export class MockWorker implements DeflateWorker {
     this.pendingMessages.length = 0
   }
 }
+
+export function waitCalls<F extends jasmine.Func>(
+  spy: jasmine.Spy<F>,
```
bcaudan, line 56:
```suggestion
  requestSpy: jasmine.Spy<F>,
```
it could make the implementation clearer

BenoitZugmeyer: It is not for "requests" anymore though. Any functions that are called asynchronously would be a good candidate for this.

bcaudan: fair enough

packages/rum-recorder/test/utils.ts
```diff
@@ -51,3 +51,26 @@ export class MockWorker implements DeflateWorker {
     this.pendingMessages.length = 0
   }
 }
+
+export function waitCalls<F extends jasmine.Func>(
```
bcaudan, line 55:
```suggestion
export function waitAsyncCalls<F extends jasmine.Func>(
```

packages/rum-recorder/test/utils.ts
```diff
@@ -51,3 +51,26 @@ export class MockWorker implements DeflateWorker {
...
+  } else {
+    spy.and.callFake((() => {
+      if (spy.calls.count() === expectedCallsCount) {
+        callback(spy.calls)
+      }
+    }) as F)
+  }
+}
+
+export function expectNoExtraCall(spy: jasmine.Spy<any>, done: () => void) {
```
bcaudan, line 71:
```suggestion
export function expectNoExtraAsyncCall(spy: jasmine.Spy<any>, done: () => void) {
```

bcaudan (approved):

PR #706, opened 2021-01-28
🐛[RUMF-836] sanitize unsupported characters in timing name

bcaudan:
## Motivation

Timing name is used as facet path, use of unsupported characters can then prevent the measure creation:

image-20210127-103109 (1)

## Changes

- Sanitize timing name by replacing unsupported characters by `_`
- Warn when a sanitization is performed

## Testing

unit + manual

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
@@ -205,7 +205,7 @@ function newView(
       }
     },
     addTiming(name: string, time: number) {
-      customTimings[name] = time - startTime
+      customTimings[sanitizeTiming(name)] = time - startTime
     },
     updateLocation(newLocation: Location) {
       location = { ...newLocation }
@@ -311,4 +311,15 @@ function trackLayoutShift(lifeCycle: LifeCycle, callback: (layoutShift: number) 
  */
 function isLayoutShiftSupported() {
   return supportPerformanceTimingEvent('layout-shift')
+}
+
+/**
+ * Timing name is used as facet path that must contain only letters, digits, or the characters - _ . @ $
+ */
+function sanitizeTiming(name: string) {
+  const sanitized = name.replace(/[^a-zA-Z0-9-_.@$]/g, '_')
+  if (sanitized !== name) {
+    console.warn(`Invalid timing name: ${name}, sanitized to: ${sanitized}`)
+  }
+  return sanitized
 }
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
@@ -911,4 +911,16 @@ describe('rum track custom timings', () => {
       foo: 1234,
     })
   })
+
+  it('should sanitized timing name', () => {
+    setupBuilder.build()
+    const warnSpy = spyOn(console, 'warn')
+
+    addTiming('foo bar-qux.@zip_21%$*€👋', 1234)
+
+    expect(getViewEvent(1).customTimings).toEqual({
+      'foo_bar-qux.@zip_21_$____': 1234,
+    })
+    expect(warnSpy).toHaveBeenCalled()
+  })
 })


packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -312,3 +312,14 @@ function trackLayoutShift(lifeCycle: LifeCycle, callback: (layoutShift: number)
 function isLayoutShiftSupported() {
   return supportPerformanceTimingEvent('layout-shift')
 }
+
+/**
+ * Timing name is used as facet path that must contain only letters, digits, or the characters - _ . @ $
+ */
+function sanitizeTiming(name: string) {
+  const sanitized = name.replace(/[^a-zA-Z0-9-_.@$]/g, '_')
```
BenoitZugmeyer, line 320: I'd remove `.` too, because sending `"foo.bar": 1234` results in `"foo": {"bar": 1234}`

bcaudan: what issue do you see with that?

BenoitZugmeyer: The timing value should be a number, no? If we want to support objects as timing value, we'll need to discuss how to do it in the UI

bcaudan: It seems to be handled so far:
Capture d’écran 2021-01-29 à 11 30 07

It could be helpful to organize timings.

@hdelaby any plan for the UI that could not allow that?

hdelaby: My 2 cents: let's keep it simple because then it adds too much complexity in the UI. We're already struggling when it comes to displaying all the timings we have... We might consider this in the future though

bcaudan: @nioufe any insight?

nioufe: In general I'd stay as close as possible to what the event platform does. 
In all other places in the sdk we can send `a.b` and it will do `a : {b} ` no?

bcaudan: Yes, this part works, the concern was more on how these custom timings will be displayed in the UI, cf Hugo's comment.

BenoitZugmeyer: When sending an action with `DD_RUM.addAction("foo.bar")`, the action name will be `foo.bar`.
When setting a timing with `DD_RUM.setTiming("foo.bar")`, it seems weird to me that the timing will be anything else than `foo.bar`.

Maybe we did not spec the schema correctly and we should have used something like `view.custom_timings = {name: string, value: number}` instead of `view.custom_timings = {[name: string]: number}` which seems a bit problematic in this case.

bcaudan: Indeed, it could be a bit weird to have different behavior for these two.
However, action name facet is predefined, we can't do the same for `custom_timings.*`.

FMU, it is either `foo: {bar}` or `foo_bar`

bcaudan: After chatting with everyone:
- let's keep `.` it will be nicely handled by the UI by displaying either facet name or facet path
- let's add some precision on action and timing API documentation to precise restrictions with name for each case

BenoitZugmeyer: Should we update the events-format schemas to illustrate that we support nested timings?

bcaudan: since the only noticeable place would be the event JSON, is it relevant?

BenoitZugmeyer: IMO the events-format can also be used as a decomumentation, and it could be useful to understand the timings type.

bcaudan: agree, but IMO the fact that timings can be nested is not related to the type but to how it is handled by the backend.
I'd be in favor of adding something on supported characters though

BenoitZugmeyer: Alright

BenoitZugmeyer (approved):

PR #707, opened 2021-01-29
[REPLAY-164] track Focus records

BenoitZugmeyer:
## Motivation

Add focus information to always show the currently focused view during replay

## Changes

* adjust types
* adjust rrweb architecture a bit to externalize records generation: the goal is to avoid implementing new records in the RRWeb codebase, to be more modular and easier to test. In a future PR, I'll remove some records from RRWeb, limiting its usage to recording DOM changes.
* add Focus records tracking

## Testing

Unit tests, manual tests

---

### DIFF
--- a/packages/rum-recorder/src/boot/recorder.ts
+++ b/packages/rum-recorder/src/boot/recorder.ts
@@ -1,10 +1,10 @@
-import { Configuration } from '@datadog/browser-core'
+import { Configuration, DOM_EVENT, addEventListeners } from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType, ParentContexts, RumSession } from '@datadog/browser-rum-core'
 
 import { record } from '../domain/rrweb'
 import { startSegmentCollection } from '../domain/segmentCollection'
 import { send } from '../transport/send'
-import { RawRecord } from '../types'
+import { RawRecord, RecordType } from '../types'
 
 export function startRecording(
   lifeCycle: LifeCycle,
@@ -31,11 +31,27 @@ export function startRecording(
 
   lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, takeFullSnapshot)
   lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, takeFullSnapshot)
+  const { stop: stopTrackingFocusRecords } = trackFocusRecords(lifeCycle, addRawRecord)
 
   return {
     stop() {
       stopRecording()
       stopSegmentCollection()
+      stopTrackingFocusRecords()
     },
   }
+}
+
+export function trackFocusRecords(lifeCycle: LifeCycle, addRawRecord: (record: RawRecord) => void) {
+  function addFocusRecord() {
+    addRawRecord({
+      type: RecordType.Focus,
+      data: {
+        has_focus: document.hasFocus(),
+      },
+    })
+  }
+  addFocusRecord()
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, addFocusRecord)
+  return addEventListeners(window, [DOM_EVENT.FOCUS, DOM_EVENT.BLUR], addFocusRecord)
 }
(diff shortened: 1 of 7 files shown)


vlad-mh (commented):
I like that we're formalizing the "Record" name instead of "Event".
I see we're sending an event when focus is gained, but not lost. Is this a limitation brought by the browser APIs?

BenoitZugmeyer:
@vlad-mh 
>  I see we're sending an event when focus is gained, but not lost. Is this a limitation brought by the browser APIs?
We are sending an event when the focus is lost too (blur). Did you experience this yourself?

packages/rum-recorder/src/types.ts
```diff
@@ -1,21 +1,5 @@
-// Alias EventWithTime to Record, to avoid naming clash between RRWeb events and RUM events
```
ThibautGeriz, line 1: Will we be able to reuse this file from web-ui instead of the duplicate type file?

BenoitZugmeyer: Not just yet, we'll have to figure out how to do so.

packages/rum-recorder/src/boot/recorder.ts
```diff
@@ -31,11 +31,27 @@ export function startRecording(
...
     },
   }
 }
+
+export function trackFocusRecords(lifeCycle: LifeCycle, addRawRecord: (record: RawRecord) => void) {
+  function addFocusRecord() {
+    addRawRecord({
+      type: RecordType.Focus,
+      data: {
+        has_focus: document.hasFocus(),
```
bcaudan, line 50: What was the reasoning to choose focus rather than visibility?

BenoitZugmeyer: When multiple browser windows are open, multiple replays can be visible, while a single replay is focused. This is closer of what we want.

bcaudan (approved):
LGTM

packages/rum-recorder/src/boot/recorder.ts
ThibautGeriz: The only thing I can think of is: if we detect that the user has two (or more) visible window, we might show them side by side (even though the only one will has focus and show stuff). You can always add `is_visible` flag later one if we decide this is the behaviour we want to track

packages/rum-recorder/src/domain/rrweb/record.ts
```diff
@@ -86,28 +80,28 @@ function record<T = EventWithTime>(options: RecordOptions<T> = {}): RecordAPI |
...
+    emit(((packFn ? packFn(record) : record) as unknown) as T, isCheckout)
+    if (record.type === RecordType.FullSnapshot) {
+      lastFullSnapshotRecordTimestamp = Date.now()
       incrementalSnapshotCount = 0
-    } else if (e.type === EventType.IncrementalSnapshot) {
+    } else if (record.type === RecordType.IncrementalSnapshot) {
       incrementalSnapshotCount += 1
       const exceedCount = checkoutEveryNth && incrementalSnapshotCount >= checkoutEveryNth
-      const exceedTime = checkoutEveryNms && e.timestamp - lastFullSnapshotEvent.timestamp > checkoutEveryNms
+      const exceedTime = checkoutEveryNms && Date.now() - lastFullSnapshotRecordTimestamp > checkoutEveryNms
```
ThibautGeriz, line 104: Why did you changed from `e.timestamp`to `Date.now()`?

ThibautGeriz (approved):

packages/rum-recorder/src/domain/rrweb/record.ts
BenoitZugmeyer: Because `rrweb/record` uses `RawRecords` instead of `Records` now, so records don't have the `timestamp` property anymore. They are "wraped" later in the recorder code. See the whole commit for context https://github.com/DataDog/browser-sdk/pull/707/commits/995f8635a49ba731e9c22e371a2f633225851071

vlad-mh:
> @vlad-mh
> 
> > I see we're sending an event when focus is gained, but not lost. Is this a limitation brought by the browser APIs?
> 
> We are sending an event when the focus is lost too (blur). Did you experience this yourself?

mb, I misunderstood. all good :)

PR #708, opened 2021-01-29
🔧 [RUMF-825] restrict which types are used depending on the context

bcaudan:
## Motivation

All types present in node_modules are used for every type of compilations (typecheck, bundle, ...) which can then lead to some type ambiguity.

ex: setTimeout defined by DOM and node

## Changes

- list needed types depending on the context
- used compiled sources in e2e compilation to avoid ambiguity, move typecheck e2e to build-and-lint accordingly
- simplify code used to avoid previous ambiguities

## Testing

ci / editor

---

### DIFF
--- a/test/e2e/tsconfig.json
+++ b/test/e2e/tsconfig.json
@@ -8,11 +8,11 @@
     "target": "es5",
     "types": ["node", "webdriverio", "jasmine", "ajv"],
     "paths": {
-      "@datadog/browser-logs": ["../../packages/logs/src"],
-      "@datadog/browser-rum-core": ["../../packages/rum-core/src"],
-      "@datadog/browser-rum": ["../../packages/rum/src"],
-      "@datadog/browser-rum-recorder": ["../../packages/rum-recorder/src"],
-      "@datadog/browser-core": ["../../packages/core/src"]
+      "@datadog/browser-logs": ["../../packages/logs/cjs"],
+      "@datadog/browser-rum-core": ["../../packages/rum-core/cjs"],
+      "@datadog/browser-rum": ["../../packages/rum/cjs"],
+      "@datadog/browser-rum-recorder": ["../../packages/rum-recorder/cjs"],
+      "@datadog/browser-core": ["../../packages/core/cjs"]
     }
   }
 }
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -47,6 +47,7 @@ build-and-lint:
     - yarn build
     - yarn lint
     - scripts/cli typecheck test/app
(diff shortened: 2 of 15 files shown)


test/e2e/tsconfig.json
```diff
@@ -8,11 +8,11 @@
...
-      "@datadog/browser-logs": ["../../packages/logs/src"],
-      "@datadog/browser-rum-core": ["../../packages/rum-core/src"],
-      "@datadog/browser-rum": ["../../packages/rum/src"],
-      "@datadog/browser-rum-recorder": ["../../packages/rum-recorder/src"],
-      "@datadog/browser-core": ["../../packages/core/src"]
+      "@datadog/browser-logs": ["../../packages/logs/cjs"],
+      "@datadog/browser-rum-core": ["../../packages/rum-core/cjs"],
+      "@datadog/browser-rum": ["../../packages/rum/cjs"],
+      "@datadog/browser-rum-recorder": ["../../packages/rum-recorder/cjs"],
+      "@datadog/browser-core": ["../../packages/core/cjs"]
```
BenoitZugmeyer, line 15: This is a bit annoying: it implies that we'll need to build everything while working on e2e test. I understand that it's already needed for the `e2e/test` application, but we are a lot less likely to work on it than e2e tests.

> used compiled sources in e2e compilation to avoid ambiguity

What ambiguity did you encounter?

I feel that someday we'll need to adopt [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) (see [Overall Structure](https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure)) instead of having those kind of paths.

bcaudan: >  it implies that we'll need to build everything while working on e2e test

it is already what we are doing when running e2e tests: https://github.com/DataDog/browser-sdk/blob/master/package.json#L23

> What ambiguity did you encounter?

setTimeout same as before, packages src were compiled using e2e tsconfig which require node

Project references seems interesting seems cleaner, I'll have a look

bcaudan: I've played a bit with it but that would require more thoughts.
Another way to avoid those paths in the tsconfig could be to have a real dependency between e2e and the packages instead of aliasing relative paths.
I'd be in favor of keeping it like this for now and see if we need to do more later on.
wdyt?

BenoitZugmeyer: Ok, I'm fine with the state of this PR.

> have a real dependency between e2e and the packages 

I like this idea!

BenoitZugmeyer (approved):

PR #709, opened 2021-01-29
✨ Allow logs collection on file:// URL

kcaffrey:
## Motivation

The motivation for this change is to allow logging from an Electron app, where it is standard to bundle the JS with webpack and serve the bundle from a file locally.

## Changes

This change remove the local file check to allow logs SDK initialization when the page is served from file:// protocol. 

## Testing

There are no automated tests for this change, as I did not see any pre-existing tests for the local file initialization check, but I tested by building the bundle and pointing my app's dependency to that version rather than the live version of the browser-sdk.

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -51,7 +51,6 @@ export interface UserConfiguration {
   trackInteractions?: boolean
   proxyHost?: string
   beforeSend?: (event: any) => void
-  allowLocalFile?: boolean
 
   service?: string
   env?: string
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -51,6 +51,7 @@ export interface UserConfiguration {
   trackInteractions?: boolean
   proxyHost?: string
   beforeSend?: (event: any) => void
+  allowLocalFile?: boolean
 
   service?: string
   env?: string
--- a/packages/logs/README.md
+++ b/packages/logs/README.md
@@ -127,6 +127,7 @@ The following parameters are available to configure the Datadog browser logs SDK
 | `forwardErrorsToLogs` | Boolean | No       | `true`          | Set to `false` to stop forwarding console.error logs, uncaught exceptions and network errors to Datadog. |
 | `sampleRate`          | Number  | No       | `100`           | The percentage of sessions to track: `100` for all, `0` for none. Only tracked sessions send logs.       |
 | `silentMultipleInit`  | Boolean | No       |                 | Prevent logging errors while having multiple init.                                                       |
+| `allowLocalFile`      | Boolean | No       |                 | Allow initialization even when served from a `file:` url.                                                |
 
 Options that must have a matching configuration when using the `RUM` SDK:
 
--- a/packages/logs/README.md
+++ b/packages/logs/README.md
@@ -127,7 +127,6 @@ The following parameters are available to configure the Datadog browser logs SDK
 | `forwardErrorsToLogs` | Boolean | No       | `true`          | Set to `false` to stop forwarding console.error logs, uncaught exceptions and network errors to Datadog. |
 | `sampleRate`          | Number  | No       | `100`           | The percentage of sessions to track: `100` for all, `0` for none. Only tracked sessions send logs.       |
 | `silentMultipleInit`  | Boolean | No       |                 | Prevent logging errors while having multiple init.                                                       |
-| `allowLocalFile`      | Boolean | No       |                 | Allow initialization even when served from a `file:` url.                                                |
 
 Options that must have a matching configuration when using the `RUM` SDK:
 
--- a/packages/logs/src/boot/logs.entry.ts
+++ b/packages/logs/src/boot/logs.entry.ts
@@ -53,7 +53,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
     logger,
 
     init: monitor((userConfiguration: LogsUserConfiguration) => {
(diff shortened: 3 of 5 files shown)


sarina-dd (approved):

packages/logs/src/boot/logs.entry.ts
```diff
@@ -53,7 +53,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
     logger,
 
     init: monitor((userConfiguration: LogsUserConfiguration) => {
-      if (!checkIsNotLocalFile() || !canInitLogs(userConfiguration)) {
+      if (!(userConfiguration.allowLocalFile || checkIsNotLocalFile()) || !canInitLogs(userConfiguration)) {
```
bcaudan, line 56: There is a failing test on the case where init is called without configuration.
Checking `canInitLogs` first should solve that

packages/rum/README.md
```diff
@@ -141,6 +141,7 @@ The following parameters are available:
 | `silentMultipleInit`    | Boolean | No       | `false`         | Initialization fails silently if Datadog's RUM is already initialized on the page.                       |
 | `proxyHost`             | String  | No       |                 | Optional proxy host (ex: www.proxy.com), see the full [proxy setup guide][7] for more information.       |
 | `allowedTracingOrigins` | List    | No       |                 | A list of request origins used to inject tracing headers.                                                |
+| `allowLocalFile`        | Boolean | No       |                 | Allow initialization even when served from a `file:` url.                                                |
```
bcaudan, line 144: Since cookies are not handled by browsers on `file` protocol and cookies are required by RUM to correctly handled session, I am not in favor of integrating this option as is for RUM.
I'd be in favor of limiting this PR for logs and we should come with a proper strategy to handle RUM sessions on `file` protocol.
Would that work for you?

BenoitZugmeyer: Since we support using logs without cookies, it could be easier to simply remove `checkIsNotLocalFile` in `logs.entry` instead of introducing a new option.

+1 to keep RUM as it is for now, and discuss how to use RUM without cookie further.

packages/core/src/domain/configuration.ts
```diff
@@ -51,6 +51,7 @@ export interface UserConfiguration {
   trackInteractions?: boolean
   proxyHost?: string
   beforeSend?: (event: any) => void
+  allowLocalFile?: boolean
```
kcaffrey, line 54:
```suggestion
```

packages/logs/README.md
```diff
@@ -127,6 +127,7 @@ The following parameters are available to configure the Datadog browser logs SDK
 | `forwardErrorsToLogs` | Boolean | No       | `true`          | Set to `false` to stop forwarding console.error logs, uncaught exceptions and network errors to Datadog. |
 | `sampleRate`          | Number  | No       | `100`           | The percentage of sessions to track: `100` for all, `0` for none. Only tracked sessions send logs.       |
 | `silentMultipleInit`  | Boolean | No       |                 | Prevent logging errors while having multiple init.                                                       |
+| `allowLocalFile`      | Boolean | No       |                 | Allow initialization even when served from a `file:` url.                                                |
```
kcaffrey, line 130:
```suggestion
```

packages/logs/src/boot/logs.entry.ts
kcaffrey, line 56:
```suggestion
      if (!canInitLogs(userConfiguration)) {
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -61,7 +61,7 @@ export function makeRumPublicApi(startRumImpl: StartRum) {
     init: monitor((userConfiguration: RumUserConfiguration) => {
       if (
         !checkCookiesAuthorized(buildCookieOptions(userConfiguration)) ||
-        !checkIsNotLocalFile() ||
+        !(userConfiguration.allowLocalFile || checkIsNotLocalFile()) ||
```
kcaffrey, line 64:
```suggestion
        !checkIsNotLocalFile() ||
```

packages/rum/README.md
kcaffrey, line 144:
```suggestion
```

packages/rum/README.md
```diff
@@ -170,6 +171,7 @@ init(configuration: {
     trackSessionAcrossSubdomains?: boolean,
     useSecureSessionCookie?: boolean,
     useCrossSiteSessionCookie?: boolean,
+    allowLocalFile?: boolean,
```
kcaffrey, line 174:
```suggestion
```

packages/rum/README.md
kcaffrey: I have updated the PR with these suggestions. We were only using logs, so this would definitely work for us.

bcaudan (approved):

BenoitZugmeyer (approved):

bcaudan:
Thanks for your contribution @kcaffrey, it has been released with 2.5.0!

johnpyp:
I'm also using electron, and this change allows it to run, but I'm still getting a `Cookies are not authorized, we will not send any data` when serving from the file.

I tried to enable `useCrossSiteSessionCookie` to fix it, however according to the docs that implies requiring https which obviously isn't met here.

Any suggestions for this specific usecase if anyone else has run into would be appreciated, and I'll update this comment if I figure it out.

bcaudan:
Hello @johnpyp,

Are you using only Logs SDK on 2.5.0 version?
If further help is needed, would you mind opening a new issue?

Thanks!

PR #711, opened 2021-02-02
[REPLAY-149] implement ViewEnd record

BenoitZugmeyer:
## Motivation

This is part of multitab tracking strategy. To let the player know a view has ended, a new record is added when the view ends or before unloading the page.

## Note

There is some limitations with this strategy:

* there may be multiple ViewEnd per view because a ViewEnd record is added even if the "beforeunload" event gets cancelled. We may add a way to detect "beforeunload" cancellations in the future
* there may be no ViewEnd for a view: if the browser fails to send the segment for whatever reason (sendBeacon limitations, connection issues, ...)
* there may be records or segments added after the ViewEnd record: the "beforeunload" event is triggered when the page starts to unload, and it can take a moment (the browser is waiting for the first byte of the next page response to actually leave the current page)

The player will need to be aware of this and handle it gracefully.

## Changes

* Adapt the recorder "flush" strategy to avoid sideeffects
* Add a ViewEnd record at the end of views

## Testing

Unit + Manual

---

### DIFF
--- a/packages/rum-recorder/package.json
+++ b/packages/rum-recorder/package.json
@@ -21,5 +21,9 @@
     "type": "git",
     "url": "https://github.com/DataDog/browser-sdk.git",
     "directory": "packages/rum-recorder"
+  },
+  "devDependencies": {
+    "@types/pako": "1.0.1",
+    "pako": "2.0.3"
   }
 }
--- a/packages/rum-recorder/src/boot/recorder.ts
+++ b/packages/rum-recorder/src/boot/recorder.ts
@@ -13,10 +13,6 @@ export function startRecording(
   session: RumSession,
   parentContexts: ParentContexts
 ) {
-  // This needs to be invoked before startSegmentCollection so lifeCycle events are subscribed
-  // first, and ViewEnd records gets added before the segment is flushed.
-  trackViewEndRecord(lifeCycle, (record) => addRawRecord(record))
-
   const { addRecord, stop: stopSegmentCollection } = startSegmentCollection(
     lifeCycle,
     applicationId,
@@ -36,6 +32,7 @@ export function startRecording(
   lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, takeFullSnapshot)
   lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, takeFullSnapshot)
   const { stop: stopTrackingFocusRecords } = trackFocusRecords(lifeCycle, addRawRecord)
+  trackViewEndRecord(lifeCycle, (record) => addRawRecord(record))
 
   return {
     stop() {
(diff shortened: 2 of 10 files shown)


packages/rum-recorder/package.json
```diff
@@ -21,5 +21,9 @@
     "type": "git",
     "url": "https://github.com/DataDog/browser-sdk.git",
     "directory": "packages/rum-recorder"
+  },
+  "devDependencies": {
+    "@types/pako": "1.0.1",
+    "pako": "2.0.3"
```
ThibautGeriz, line 27: Why do we need pako for a the ViewEnd record?

BenoitZugmeyer: This is for tests: I want to make sure ViewEnd are inserted at the end of the record, so I need to inflate the request data to parse what would be sent to the intake.

ThibautGeriz (approved):

packages/rum-recorder/src/boot/recorder.ts
```diff
@@ -13,6 +13,10 @@ export function startRecording(
   session: RumSession,
   parentContexts: ParentContexts
 ) {
+  // This needs to be invoked before startSegmentCollection so lifeCycle events are subscribed
+  // first, and ViewEnd records gets added before the segment is flushed.
```
bcaudan, line 17: We also rely on the fact that `startRum` is called before `startRecording` to not generate a `ViewEnd` record with the initial view, all that seems fishy to me 😕 
What about having a `VIEW_ENDED` lifecycle event?

BenoitZugmeyer: Ah right, it is probably better.

packages/rum-recorder/test/utils.ts
```diff
@@ -69,6 +69,8 @@ export function collectAsyncCalls<F extends jasmine.Func>(spy: jasmine.Spy<F>) {
       spy.and.callFake((() => {
         fail('Unexpected extra call')
       }) as F)
+      // make sure the "setTimeout" is not mocked
+      jasmine.clock().uninstall()
```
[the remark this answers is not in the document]
BenoitZugmeyer: In recorder.spec.ts, I now wait for the segment to expire instead of notifying a BEFORE_UNLOAD lifecycle event. Thus, I needed to mock the clock with jasmine.

But, the code tested is still asynchronous because it uses WebWorker communications, so I still need to use `collectAsyncCalls`, and  use an actual `setTimeout` to wait for WebWorker messages to be sent.

So I could ensure `setTimeout` to not be mocked and throw an error, but I'd still need to unmock it in all my tests before using `expectNoExtraAsyncCalls`. I thought that removing the mock here makes things a bit less verbose and a bit more fool proof.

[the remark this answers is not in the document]
BenoitZugmeyer: Anyway, I'll just revert this commit, since it is now unnecessary with the VIEW_ENDED lifecycle event :)

bcaudan (approved):
LGTM

PR #715, opened 2021-02-03
replay: add dd-privacy attribute for obfuscation & ignoring input

vlad-mh:
Following up on https://github.com/DataDog/browser-sdk/pull/692 we figured it was best to expose privacy controls leveraging both classes & data attributes. Data attributes are used by other places of the SDK, and it made sense to show consistency.

This PR introduces the "privacy.ts" file with various helpers used by rrweb & rrweb-snapshot. This results in the recorder:
- ignoring the content of an element (obfuscation) if it has the attribute `data-dd-privacy` set to `hidden` or a class of `dd-privacy-hidden` 
- ignoring input events of an element if it has the attribute `data-dd-privacy` set to `input-ignored` or a class of `dd-privacy-input-ignored`, or if it is an input of type "password", "email" or "tel"

Further work could be done to offer a third strategy for inputs, obfuscating with asterisks instead of completely ignoring the input. Another idea would be to make the set of input types configurable. Finally, we could investigate automatic obfuscation of text node if matching regexps.

### DIFF
--- a/packages/rum-recorder/src/domain/privacy.ts
+++ b/packages/rum-recorder/src/domain/privacy.ts
@@ -1,7 +1,15 @@
 const PRIVACY_ATTR_NAME = 'data-dd-privacy'
 const PRIVACY_ATTR_VALUE_HIDDEN = 'hidden'
+const PRIVACY_ATTR_VALUE_INPUT_IGNORED = 'input-ignored'
 
 const PRIVACY_CLASS_HIDDEN = 'dd-privacy-hidden'
+const PRIVACY_CLASS_INPUT_IGNORED = 'dd-privacy-input-ignored'
+
+// PRIVACY_INPUT_TYPES_TO_IGNORE defines the input types whose input
+// events we want to ignore by default, as they often contain PII.
+// TODO: We might want to differentiate types to fully ignore vs types
+// to obfuscate.
+const PRIVACY_INPUT_TYPES_TO_IGNORE = ['email', 'password', 'tel']
 
 // Returns true if the given DOM node should be hidden. Ancestors
 // are not checked.
@@ -27,6 +35,36 @@ export function nodeOrAncestorsAreHidden(node: Node | null): boolean {
   return nodeOrAncestorsAreHidden(node.parentNode)
 }
 
+// Returns true if the given DOM node should have it's input events
+// ignored. Ancestors are not checked.
+export function nodeHasInputIngored(node: Node): boolean {
+  return (
+    isElement(node) &&
+    (node.getAttribute(PRIVACY_ATTR_NAME) === PRIVACY_ATTR_VALUE_INPUT_IGNORED ||
+      node.classList.contains(PRIVACY_CLASS_INPUT_IGNORED) ||
+      // if element is an HTMLInputElement, check the type is not to be ignored by default
+      (isInputElement(node) && PRIVACY_INPUT_TYPES_TO_IGNORE.includes(node.type)))
+  )
+}
+
+// Returns true if the given DOM node should have it's input events
+// ignored, recursively checking its ancestors.
+export function nodeOrAncestorsHaveInputIngnored(node: Node | null): boolean {
+  if (!node) {
+    return false
+  }
+
+  if (nodeHasInputIngored(node)) {
+    return true
+  }
+
+  return nodeOrAncestorsHaveInputIngnored(node.parentNode)
+}
+
 function isElement(node: Node): node is Element {
   return node.nodeType === node.ELEMENT_NODE
+}
+
+function isInputElement(elem: Element): elem is HTMLInputElement {
+  return elem.tagName === 'INPUT'
 }
--- a/packages/rum-recorder/src/domain/privacy.ts
+++ b/packages/rum-recorder/src/domain/privacy.ts
@@ -13,7 +13,7 @@ const PRIVACY_INPUT_TYPES_TO_IGNORE = ['email', 'password', 'tel']
 
 // Returns true if the given DOM node should be hidden. Ancestors
 // are not checked.
-export function nodeIsHidden(node: Node): boolean {
+export function nodeShouldBeHidden(node: Node): boolean {
   return (
     isElement(node) &&
     (node.getAttribute(PRIVACY_ATTR_NAME) === PRIVACY_ATTR_VALUE_HIDDEN ||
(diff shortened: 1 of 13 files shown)


packages/rum-recorder/src/domain/privacy.ts
```diff
@@ -0,0 +1,31 @@
+const PRIVACY_ATTR_NAME = 'data-dd-privacy'
+const PRIVACY_ATTR_VALUE_HIDDEN = 'hidden'
+
+const PRIVACY_CLASS_HIDDEN = 'dd-privacy-hidden'
+
+// Returns true if the given DOM node should be hidden. Ancestors
+// are not checked.
+export function nodeIsHidden(node: Node): boolean {
```
bcaudan, line 16: naming: I found `nodeIsHidden` a bit confusing since it is not hidden yet, maybe `nodeShouldBeHidden` or `nodeIsMarkedHidden` instead?
same logic for the other function

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -207,7 +185,7 @@ function serializeNode(
         systemId: (n as DocumentType).systemId,
       }
     case n.ELEMENT_NODE:
-      const needBlock = isBlockedElement(n as HTMLElement, blockClass, blockSelector)
+      const needBlock = nodeIsHidden(n)
```
bcaudan, line 188: Would it make sense to rename all `block` occurrences to `hidden` as well?

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -56,4 +63,84 @@ describe('recorder', () => {
...
+
+        const pFooNode = bodyNode.childNodes[1] as ElementNode
+        expect(pFooNode.childNodes.length).toBe(1)
+        expect((pFooNode.childNodes[0] as TextNode).textContent).toBe('foo')
+
+        const pBarNode = bodyNode.childNodes[3] as ElementNode
+        expect(pBarNode.childNodes.length).toBe(0) // obfuscated
+
+        const pBazNode = bodyNode.childNodes[5] as ElementNode
+        expect(pBazNode.childNodes.length).toBe(0) // obfuscated
```
bcaudan, line 104: That is a lot of checks and details about the data structure.
could it be interesting/possible to have something like:
```ts
const { segment } = events.sessionReplay[0]
expect(findNodeWithId(segment, 'foo').textContent).toBe('foo')
expect(findNodeWithId(segment, 'bar').textContent).toBe('')
expect(findNodeWithId(segment, 'baz').textContent).toBe('')
```

packages/rum-recorder/src/domain/privacy.ts
vlad-mh: I see your point. I don't have strong feeling, so I'll go with `should`

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
vlad-mh: agreed, I'll add a commit with that rename

packages/rum-recorder/src/domain/privacy.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (nodeIsHidden(node)) {
+    return true
+  }
+
+  return nodeOrAncestorsAreHidden(node.parentNode)
+}
+
+// Returns true if the given DOM node should have it's input events
+// ignored. Ancestors are not checked.
+export function nodeHasInputIngored(node: Node): boolean {
```
BenoitZugmeyer, line 43:
```suggestion
export function nodeHasInputIgnored(node: Node): boolean {
```

packages/rum-recorder/src/domain/privacy.ts
```diff
@@ -0,0 +1,72 @@
...
+    return true
+  }
+
+  // if element is an HTMLInputElement, check the type is not to be ignored by default
+  return elem.tagName === 'INPUT' && PRIVACY_INPUT_TYPES_TO_IGNORE.includes((elem as HTMLInputElement).type)
+}
+
+// Returns true if the given DOM node should have it's input events
+// ignored, recursively checking its ancestors.
+export function nodeOrAncestorsHaveInputIngnored(node: Node | null): boolean {
```
BenoitZugmeyer, line 62:
```suggestion
export function nodeOrAncestorsHaveInputIgnored(node: Node | null): boolean {
```

packages/rum-recorder/src/domain/privacy.ts
```diff
@@ -0,0 +1,72 @@
...
+// TODO: We might want to differentiate types to fully ignore vs types
+// to obfuscate.
+const PRIVACY_INPUT_TYPES_TO_IGNORE = ['email', 'password', 'tel']
+
+// Returns true if the given DOM node should be hidden. Ancestors
+// are not checked.
+export function nodeIsHidden(node: Node): boolean {
+  if (node.nodeType !== node.ELEMENT_NODE) {
+    return false
+  }
```
BenoitZugmeyer, line 19: I would add a function like:
```ts
function isElement(node: Node): node is Element {
  return node.nodeType === node.ELEMENT_NODE
}
```

Then, you wouldn't need a cast:
```ts
export function nodeIsHidden(node: Node): boolean {
  if (!isElement(node)) {
    return false;
  }
  // here, node has type "Element"
```

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -56,4 +63,84 @@ describe('recorder', () => {
         )
       ).toBeTruthy('have a IncrementalSnapshot/MouseInteraction record')
     })
+
+  fdescribe('snapshot', () => {
```
BenoitZugmeyer, line 67:
```suggestion
  describe('snapshot', () => {
```

BenoitZugmeyer (commented):
It seems that `blockClass` and `blockSelector` could be cleaned up in the whole codebase?

test/e2e/lib/helpers/recorder.ts
```diff
@@ -0,0 +1,80 @@
...
+  return segment.records.find((record) => record.type === RecordType.FullSnapshot) as FullSnapshotRecord
+}
+
+// Returns the first IncrementalSnapshotRecord of a given source in a
+// Segment, if any.
+export function findIncrementalSnapshot(segment: Segment, source: IncrementalSource): IncrementalSnapshotRecord | null {
+  return segment.records.find(
+    (record) => record.type === RecordType.IncrementalSnapshot && record.data.source === source
+  ) as IncrementalSnapshotRecord
+}
```
BenoitZugmeyer, line 33: Here, the typing *could* be improved with type guards too, for example:

```ts
const allInputData = segment.records.find(isIncrementalSnapshot).map(record => record.data).filter(isInputData)
```

But it would be a bit more verbose, I'll let you decide if it's worth it :p

BenoitZugmeyer (commented):
Couple of NIT on the test helpers, but it looks nice

test/e2e/lib/helpers/recorder.ts
```diff
@@ -0,0 +1,80 @@
...
+
+// Returns this first MetaRecord in a Segment, if any.
+export function findMeta(segment: Segment): MetaRecord | null {
+  return segment.records.find((record) => record.type === RecordType.Meta) as MetaRecord
+}
+
+// Returns this first FullSnapshotRecord in a Segment, if any.
+export function findFullSnapshot(segment: Segment): FullSnapshotRecord | null {
+  return segment.records.find((record) => record.type === RecordType.FullSnapshot) as FullSnapshotRecord
+}
```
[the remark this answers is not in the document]
vlad-mh: right but here, it makes it quite a bit more verbose? I was thinking of adding isXX only if it's used (as for isElementNode)?

test/e2e/lib/helpers/recorder.ts
vlad-mh: i see your point but it feels quite verbose adding these isXX functions here as well. maybe we add them only if/when other functions start requiring it?

BenoitZugmeyer (approved):

PR #719, opened 2021-02-04
🔥 recorder: remove forEach polyfills

bcaudan:
## Motivation

Avoid to touch shared API if it is not needed

## Changes

Replace `forEach` polyfills by a forEach utility function

## Testing

ci

---


test/e2e/lib/types/serverEvents.ts
```diff
@@ -1,5 +1,5 @@
 import { RumActionEvent, RumErrorEvent, RumEvent, RumResourceEvent, RumViewEvent } from '@datadog/browser-rum'
-import { Segment } from '../../../../packages/rum-recorder/src/types'
+import { Segment } from '@datadog/browser-rum-recorder/cjs/types'
```
BenoitZugmeyer, line 2: I was surprised that it works, but it looks fine, TS seems to find the packages in the root node_modules. I think we can remove the `paths` in `test/e2e/tsconfig.json`

bcaudan: Indeed!

BenoitZugmeyer (approved):

PR #721, opened 2021-02-04
[RUMF-843] monitor rrweb codebase

BenoitZugmeyer:
## Motivation

Monitor RRWeb codebase to have better visibility on what goes wrong in the wild.

## Changes

Use the Browser SDK internal monitoring on all RRWeb "observer" asynchronous functions.

## Testing

---

### DIFF
--- a/packages/core/src/domain/internalMonitoring.ts
+++ b/packages/core/src/domain/internalMonitoring.ts
@@ -109,18 +109,33 @@ export function monitored<T extends (...params: any[]) => unknown>(
 
 export function monitor<T extends (...args: any[]) => any>(fn: T): T {
   return (function (this: any) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return callMonitored(fn, this, (arguments as unknown) as Parameters<T>)
+  } as unknown) as T // consider output type has input type
+}
+
+export function callMonitored<T extends (...args: any[]) => any>(
+  fn: T,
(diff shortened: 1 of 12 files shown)


packages/core/src/domain/internalMonitoring.ts
```diff
@@ -101,26 +101,41 @@ export function monitored<T extends (...params: any[]) => unknown>(
...
+  context: utils.ThisParameterType<T>,
+  args: utils.Parameters<T>
+): ReturnType<T> | undefined
+export function callMonitored<T extends (this: void) => any>(fn: T): ReturnType<T> | undefined
+export function callMonitored<T extends (...args: any[]) => any>(
+  fn: T,
+  context?: any,
+  args?: any
+): ReturnType<T> | undefined {
+  try {
```
bcaudan, line 128: > This commit introduces callMonitored function to allow a more granular
monitoring of our code.

what was wrong with:
```ts
monitor(fn)()
```

BenoitZugmeyer: It creates a useless intermediate function.

bcaudan: hum, that quite some ts boilerplate to avoid intermediate functions.
I'm also a bit bothered by the naming but I have nothing better to propose.

unless you have a better idea let's go with that.

bcaudan (approved):

PR #724, opened 2021-02-08
✨ [RUMF-847] Add onNewLocation to configuration

webNeat:
## Motivation

Make it possible to to rename and skip views using a configuration callback.

## Changes

- Add `onNewLocation` to RUM userConfiguration
```ts
onNewLocation?: (newLocation: Location, oldLocation?: Location) => undefined | { shouldCreateView?: boolean; viewName?: string }
```

- Skip creating the view if the returned `shouldCreateView` is false (except the initial view).
- Use the returned `viewName` as the view name.

## Testing

- unit tests
- manual
---


packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -41,9 +43,19 @@ export interface ViewCustomTimings {
 export const THROTTLE_VIEW_UPDATE_PERIOD = 3000
 export const SESSION_KEEP_ALIVE_INTERVAL = 5 * ONE_MINUTE
 
-export function trackViews(location: Location, lifeCycle: LifeCycle) {
+export function trackViews(location: Location, lifeCycle: LifeCycle, newLocationListener?: NewLocationListener) {
```
BenoitZugmeyer, line 46: You could set the default value directly in the signature, so you don't have two names for the same thing (`onNewLocation` / `newLocationListener`):
```suggestion
export function trackViews(location: Location, lifeCycle: LifeCycle, onNewLocation: NewLocationListener = noop) {
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -41,9 +43,19 @@ export interface ViewCustomTimings {
...
   const startOrigin = 0
-  const initialView = newView(lifeCycle, location, ViewLoadingType.INITIAL_LOAD, document.referrer, startOrigin)
+  const onNewLocation: NewLocationListener = newLocationListener || (() => undefined)
+
+  const { viewName } = onNewLocation(location) || {}
```
BenoitZugmeyer, line 50: You could also use the `?.` operator like this (this way, you don't even have to set a default value for the parameter):
```ts
const viewName = onNewLocation?.(location)?.viewName
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -56,14 +68,21 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
...
-      currentView.end()
-      currentView.triggerUpdate()
-      currentView = newView(lifeCycle, location, ViewLoadingType.ROUTE_CHANGE, currentView.url)
-    } else {
-      currentView.updateLocation(location)
-      currentView.triggerUpdate()
+      const { shouldCreateView, viewName } = {
+        shouldCreateView: true,
+        ...onNewLocation(location, currentView.getLocation()),
+      }
```
BenoitZugmeyer, line 74:
```suggestion
      const { shouldCreateView = true, viewName } = onNewLocation(location, currentView.getLocation()) || {}
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -208,6 +209,87 @@ describe('rum track url change', () => {
   })
 })
 
+describe('rum use onNewLocation callback to rename/ignore views', () => {
```
BenoitZugmeyer, line 212: Could you add a test for the `oldLocation` parameter? (check that it is undefined for initial view, and defined with a correct value for other views)

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan, line 50: Since `onNewLocation` is defined by customers, how should we behave if it throw?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -56,14 +68,21 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
...
   function onLocationChange() {
     if (currentView.isDifferentView(location)) {
-      // Renew view on location changes
-      currentView.end()
-      currentView.triggerUpdate()
-      currentView = newView(lifeCycle, location, ViewLoadingType.ROUTE_CHANGE, currentView.url)
-    } else {
-      currentView.updateLocation(location)
-      currentView.triggerUpdate()
+      const { shouldCreateView, viewName } = {
```
bcaudan, line 71: what about letting customers create view even for cases where we would not?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -21,8 +21,14 @@ import { startRum } from './rum'
 export interface RumUserConfiguration extends UserConfiguration {
   applicationId: string
   beforeSend?: (event: RumEvent) => void
+  onNewLocation?: NewLocationListener
```
bcaudan, line 24: What about putting it behind a flag first?
it could allow us to do some tests internally before releasing it

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
webNeat: You mean we don't use `isDifferentView` at all and let the user decide?

bcaudan: more use `isDifferentView` as a fallback, if the user does not return a value for `shouldCreateView`

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
webNeat: I think we should catch the error, create the view then add the error to it as custom error. WDYT?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
webNeat: Makes sense, I will do that.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: For other cases where we want to communicate a wrong configuration to the customer, we use a console.warn (ex: add timing). 
I'd be in favor of still using the console (easier to see during the integration), but promoting this one to error, it will still be collected.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -324,3 +352,15 @@ function sanitizeTiming(name: string) {
   }
   return sanitized
 }
+
+function wrapOnNewLocation(onNewLocation: NewLocationListener): NewLocationListener {
```
BenoitZugmeyer, line 356: This is nice, and I think it may be re-used for other user-provided callbacks, like `onReady`. No need to do it now though.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -324,3 +352,15 @@ function sanitizeTiming(name: string) {
...
+  return (newLocation, oldLocation) => {
+    let result
+    try {
+      result = onNewLocation(newLocation, oldLocation)
+    } catch (err) {
+      console.error('onNewLocation threw an error:', err)
```
BenoitZugmeyer, line 362: Just a thought: I think we should prefix all our console logs so it is easily identifiable for end users.

BenoitZugmeyer (approved):

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -208,6 +209,166 @@ describe('rum track url change', () => {
...
+      if (location.pathname === '/bar') {
+        throw barError
+      }
+      return undefined
+    }
+    const consoleErrorSpy = spyOn(console, 'error')
+    setupBuilder.build()
+    expect(consoleErrorSpy).toHaveBeenCalledWith('onNewLocation threw an error:', fooError)
+    history.pushState({}, '', '/bar')
+    expect(consoleErrorSpy).toHaveBeenCalledWith('onNewLocation threw an error:', barError)
```
bcaudan, line 368: what is the reasoning for having two error cases?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -41,29 +43,42 @@ export interface ViewCustomTimings {
...
 
   const { stop: stopTimingsTracking } = trackTimings(lifeCycle, (timings) => {
     initialView.updateTimings(timings)
     initialView.scheduleUpdate()
   })
 
-  trackHistory(onLocationChange)
-  trackHash(onLocationChange)
+  const { stop: stopHistoryTracking } = trackHistory(onLocationChange)
+  const { stop: stopHashTracking } = trackHash(onLocationChange)
```
bcaudan, line 69: was there a need to do that?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -208,6 +209,166 @@ describe('rum track url change', () => {
...
+      return { viewName: `Foo ${i}`, shouldCreateView: true }
+    }
+    setupBuilder.build()
+    history.pushState({}, '', '/foo')
+    history.pushState({}, '', '/foo')
+
+    expect(getViewEvent(0).name).toBe('Foo 1')
+    expect(getViewEvent(2).name).toBe('Foo 2')
+    expect(getViewEvent(4).name).toBe('Foo 3')
+  })
```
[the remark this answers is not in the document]
webNeat: I wanted to test that the view will be created even when isDifferentView() returns false. I used the counter as a simple way to differentiate between the different views created.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -208,6 +209,166 @@ describe('rum track url change', () => {
...
+    setupBuilder.build()
+    history.pushState({}, '', '/bar')
+    history.pushState({}, '', '/baz')
+    history.pushState({}, '', '/foo')
+
+    expect(getViewEvent(0).name).toBe('initial view')
+    expect(getViewEvent(2).name).toBe('Bar from /foo')
+    expect(getViewEvent(4).name).toBeUndefined()
+    expect(getViewEvent(6).name).toBe('Foo from /baz')
+  })
```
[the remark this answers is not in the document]
webNeat: You are right, I will refactor this test.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
webNeat: The first one happens on the initial view; the second on location change.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
webNeat: Yes. When `trackViews` is called, it adds the `onLocationChange` listener to history and hash changes. if we don't remove this listener and call `trackViews` again, we end up with the two listeners being executed on history/hash changes. This caused all the `onNewLocation` callbacks that I added during the tests, to be called from all following tests that triggered a history change. This caused `onNewLocation threw an error ...` to be logged multiple times to the console during tests due to the test case where I threw an error. @BenoitZugmeyer helped me to find the root cause which is that we don't stop history/hash tracking.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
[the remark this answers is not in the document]
bcaudan: ok, make sense.
what about rephrasing the case to something like:
```
should allow customer to consider other location changes as new views
```
and with locations like:
```
/foo?view=bar
/foo?view=qux
```
IMO, exact same URL case sound like an edge case

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -208,6 +209,166 @@ describe('rum track url change', () => {
...
+
+  it('should not create view for the same URL when shouldCreateView is undefined', () => {
```
bcaudan, line 289: what about something like:
```
should use our own new view detection rules when shouldCreateView is undefined
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: ok, make sense. 
Thought for future cases: It could be interesting to add this kind of info in the commit description to help the reviewer understand why you are doing the changes

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: I think it could be valuable to do it in a following PR though, with `beforeSend` also

webNeat: Ok, I will do it in a separate PR after merging this one.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
webNeat: This makes sense, what do you suggest as a prefix? what about `DD_RUM: ...`?

BenoitZugmeyer: Maybe `Datadog SDK:` would be better (more explicit, and we can use the same prefix for logs and core parts). Anyway, no need to do this in your PR :)

packages/rum-core/src/domain/parentContexts.ts
```diff
@@ -97,6 +97,7 @@ export function startParentContexts(lifeCycle: LifeCycle, session: RumSession):
       },
       view: {
         id: currentView!.id,
+        name: currentView!.name,
```
bcaudan, line 100: That seems a pretty key behavior, could it be interesting to have some tests to avoid regression for it?

webNeat: Done

bcaudan (approved):

PR #726, opened 2021-02-09
replay: set data-dd-privacy attribute on snapshot node if hidden

vlad-mh:
This PR is a follow up on https://github.com/DataDog/browser-sdk/pull/715

As a node can be hidden from a data attribute rather than a class,
only adding the class to the snapshot node attributes means we can't
properly style the node when rebuilding it.

This PR ensures the `data-dd-privacy=hidden` attribute is always
set, even if the node was hidden using the class.

Additionally, we set the ID as a/ it helps with testing, b/ we can
assume it doesn't contain PII and c/ it may be used when
rebuilding (CSS rules could be selecting by ID).

### DIFF
--- a/packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
+++ b/packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-underscore-dangle */
-import { nodeShouldBeHidden } from '../privacy'
+import { PRIVACY_ATTR_NAME, PRIVACY_ATTR_VALUE_HIDDEN, nodeShouldBeHidden } from '../privacy'
 import {
   SerializedNode,
   SerializedNodeWithId,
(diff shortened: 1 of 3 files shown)


packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -254,9 +254,11 @@ function serializeNode(
       if (shouldBeHidden) {
         const { width, height } = (n as HTMLElement).getBoundingClientRect()
         attributes = {
+          id: attributes.id,
```
BenoitZugmeyer, line 257: Is the `id` attribute needed?

vlad-mh: See my PR description:

> Additionally, we set the ID as a/ it helps with testing, b/ we can
assume it doesn't contain PII and c/ it may be used when
rebuilding (CSS rules could be selecting by ID).

Not strictly needed, but I figured I wouldn't harm?

BenoitZugmeyer: Ok it looks fine

BenoitZugmeyer:
NIT: I kind of think a unit test would be better for this change, but the e2e test may be enough

BenoitZugmeyer (approved):

PR #738, opened 2021-02-18
✅ [RUMF-815] import RRWeb integration tests

BenoitZugmeyer:
## Motivation

Import RRWeb [integration tests](https://github.com/rrweb-io/rrweb/blob/master/test/integration.test.ts) as E2E tests. Since we are running those tests in various browsers and snapshots are heavily dependent of the browse, we cannot really use snapshot testing like RRWeb does.

## Changes

* [x] can record form interactions
* [x] can record childList mutations
* [x] can record character data muatations
* [x] can record attribute mutation
* [ ] can record node mutations
    skipped, too big to test, already covered by other tests
* [ ] can freeze mutations
    skipped, feature have been removed
* [x] should not record input events on ignored elements
    was already implemented
* [ ] should not record input values if maskAllInputs is enabled
    skipped, feature not used
* [ ] can use maskInputOptions to configure which type of inputs should be masked
    skipped, feature not used
* [x] should not record blocked elements and its child nodes
* [x] should record DOM node movement 1
* [x] should record DOM node movement 2
* [x] should record dynamic CSS changes
* [ ] should record canvas mutations
    skipped, feature not used
* [x] will serialize node before record
* [ ] will defer missing next node mutation
    skipped, no relation with RRWeb
* [ ] can record log mutation
    skipped, feature not implemented
* [ ] should nest record iframe
    skipped, feature not implemented

## Testing

E2E

---


test/e2e/scenario/recorder.scenario.ts
```diff
@@ -79,25 +81,512 @@ describe('recorder', () => {
...
+        `
+      )
+      .run(async ({ events }) => {
+        await browserExecute(() => {
+          const li = document.createElement('li')
+          const ul = document.querySelector('ul') as HTMLUListElement
+          ul.appendChild(li)
+          document.body.removeChild(ul)
+          const p = document.querySelector('p') as HTMLParagraphElement
+          p.appendChild(document.createElement('span'))
```
bcaudan, line 120: FMU, you are adding the `li` to be sure that there is no mutations captured due to the parent removal.
If that is it, it could be nice to be more explicit about it.

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -79,25 +81,512 @@ describe('recorder', () => {
...
+            parentId: findElementWithTagName(fullSnapshot, 'p')!.id,
+            node: jasmine.objectContaining({ tagName: 'span' }),
+          },
+        ])
+        expect(mutations[0].data.removes).toEqual([
+          {
+            parentId: findElementWithTagName(fullSnapshot, 'body')!.id,
+            id: findElementWithTagName(fullSnapshot, 'ul')!.id,
+          },
+        ])
```
bcaudan, line 147: IMO, there are a lot of similarities across these tests, we could probably make something to reduce the boilerplate and make them more expressive.

what would you think of having a test API like:
```js
flushSessionReplayEvents(events)  // page object-like API to have acces to events, segment, full snapshot
  .singleRecord() // select record to check and do something similar for filterRecordsByIdAttribute usages
    .expectAdded ({  tag: 'span', parentTag: 'p' }) 
    .expectRemoved({ tag: 'ul', parentTag: 'body' })
    .expectNotAdded({ tag: 'li' })
    .expectNotRemoved({ tag: 'li' })
```

BenoitZugmeyer: I'll try something

BenoitZugmeyer: Ok so I did something to remove the boilerplate. Since I didn't write the original tests I wanted to keep some kind of exhaustive snapshot, so I did not use a `expectAdded`/`expectNotAdded` API, and instead add a helper to verify the whole mutations. Because nodes are referenced in the initial full snapshot or the previous mutations, it makes things a bit complex. Let me know what you think about it!

bcaudan:
LGTM except flaky tests

bcaudan (approved):

PR #742, opened 2021-02-22
🐛 [REPLAY-187] flush pending records before taking a fullsnapshot

BenoitZugmeyer:
## Motivation

Records emited by a MutationObserver are buffered and emited in batch. When taking a full snapshot, we should flush and process the pending mutation records so they aren't applied on the dom structure defined by the full snapshot.  

## Changes

Flush mutations before taking a full snapshot.

## Testing

Unit, manual

---


packages/rum-recorder/src/domain/rrweb/observer.ts
```diff
@@ -64,6 +64,7 @@ function initMutationObserver(
     childList: true,
     subtree: true,
   })
+  mutationController.onFreeze(() => mutationBuffer.processMutations(observer.takeRecords()))
```
bcaudan, line 67: IMO, the relationship between `mutationController`, `mutationBuffer`,  `mutationObserver` and this init function is becoming hard to follow.
FMU:
- mutation controller allow to freeze/unfreeze from other part of the lifecycle
- mutation buffer is not only a buffer but process mutations and emit them
- the onFreeze configuration happen here but the unFreeze configuration happen on the buffer

what about moving the observer to mutationBuffer?
it would allow to encapsulate the observe / process / emit logic, only the freeze would remain deported to the controller.
observer could be disconnected with a stop method on buffer.

With that, I'd be in favor of renaming:
- mutationBuffer to something like mutationObserverWrapper
- mutationController to something like mutationFreezer since it only handle this behavior

BenoitZugmeyer: I moved things like you suggested. I'm mitigated about renaming `MutationController`:

* it is still unclear to me why we need to freeze the mutations during a full snapshot, and I think we could remove parts of this logic in the future

* on top of freeze/unfreeze, upstream now have introduced [lock/unlock](https://github.com/rrweb-io/rrweb/commit/9187bec8148599164cb9e3bc2e6fce15973ecf6e) functions that we should consider adding

packages/rum-recorder/src/domain/rrweb/mutation.spec.ts
```diff
@@ -101,3 +109,40 @@ function createNodeList(nodes: Node[]): NodeList {
     },
   }) as unknown) as NodeList
 }
+
+class MockMutationObserver implements MutationObserver {
+  static instances: MockMutationObserver[] = []
```
bcaudan, line 115: when do we need to handle multiple instances?

packages/rum-recorder/src/domain/rrweb/observer.ts
```diff
@@ -537,7 +526,7 @@ export function initObservers(o: ObserverParam, hooks: HooksParam = {}): Listene
   const fontObserver = o.collectFonts ? initFontObserver(o.fontCb) : noop
 
   return () => {
-    mutationObserver.disconnect()
+    mutationBuffer.stop()
```
bcaudan: what about returning a handler to be consistent with the others?

packages/rum-recorder/src/domain/rrweb/observer.ts
bcaudan: ok, let's see how it goes

packages/rum-recorder/src/domain/rrweb/mutation.spec.ts
BenoitZugmeyer: No we don't have to. Do you see any issue with this?

packages/rum-recorder/src/domain/rrweb/observer.ts
BenoitZugmeyer: Sure

packages/rum-recorder/src/domain/rrweb/mutation.spec.ts
bcaudan: what about simplifying the implementation to only deal with a single instance?
it could ease understanding / maintenance

bcaudan (approved):
LGTM

PR #743, opened 2021-02-22
✨ [RUMF-854] Enable beforeSend to dismiss events

webNeat:
## Motivation

Allow customers to dismiss RUM events.

## Changes

- Use `RUM_EVENT_COLLECTED` event to update event counts.
- Dismiss the event if `beforeSend` returned `false` and it's not a view event.
- Show a warning if `beforeSend` returned `false` for a view event.

## Testing

- unit tests
- manual

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -75,7 +75,7 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
     cookieOptions: CookieOptions
 
     service?: string
-    beforeSend?: (event: any) => void | boolean
+    beforeSend?: (event: any) => unknown
 
     isEnabled: (feature: string) => boolean
   }
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -75,7 +75,7 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
     cookieOptions: CookieOptions
 
     service?: string
-    beforeSend?: (event: any) => void
+    beforeSend?: (event: any) => void | boolean
 
     isEnabled: (feature: string) => boolean
   }
--- a/packages/core/src/tools/limitModification.spec.ts
+++ b/packages/core/src/tools/limitModification.spec.ts
@@ -78,7 +78,7 @@ describe('limitModification', () => {
     })
   })
 
-  it('should return the return of the modifier', () => {
+  it('should return the result of the modifier', () => {
     const object = { foo: { bar: 'bar' } }
     const modifier = (candidate: any) => {
       candidate.foo.bar = 'qux'
--- a/packages/core/src/tools/limitModification.spec.ts
+++ b/packages/core/src/tools/limitModification.spec.ts
@@ -8,9 +8,9 @@ describe('limitModification', () => {
       candidate.qux = 'modified2'
     }
 
-    const result = limitModification(object, ['foo.bar', 'qux'], modifier)
+    limitModification(object, ['foo.bar', 'qux'], modifier)
 
-    expect(result).toEqual({
+    expect(object).toEqual({
       foo: { bar: 'modified1' },
       qux: 'modified2',
     })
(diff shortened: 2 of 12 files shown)


packages/core/src/tools/limitModification.ts
```diff
@@ -8,14 +8,15 @@ import { Context, deepClone } from './context'
 export function limitModification<T extends Context>(
   object: T,
   modifiableFieldPaths: string[],
-  modifier: (object: T) => void
-): T {
+  modifier: (object: T) => boolean | void
+): boolean | void {
```
bcaudan, line 12: is it safe to assume the return type of the modifier?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -78,8 +78,15 @@ export function startRumAssembly(
...
+        let beforeSendResult
         if (configuration.beforeSend) {
-          limitModification(serverRumEvent, FIELDS_WITH_SENSITIVE_DATA, configuration.beforeSend)
+          beforeSendResult = limitModification(serverRumEvent, FIELDS_WITH_SENSITIVE_DATA, configuration.beforeSend)
+        }
+        if (beforeSendResult === false && serverRumEvent.type !== RumEventType.VIEW) {
+          return
+        }
+        if (beforeSendResult === false) {
+          console.warn(`Can't dismiss view events using beforeSend, use onNewLocation instead!`)
```
bcaudan, line 89: Not sure why you put the beforeSendResult checks outside of the configuration.beforeSend if.
If the intention was to avoid to introduce another if level, what about extracting the before send logic into a function?

packages/core/src/tools/limitModification.ts
webNeat: We were assuming that it returns a `void` before, now it can also return a boolean. I don't see the problem with assuming that return type, am I missing something?

BenoitZugmeyer: What about using another parametric type to avoid chosing at this level?

```ts
function limitModification<T extends Context, Result>(
  object: T,
  modifiableFieldPaths: string[],
  modifier: (object: T) => Result
): Result | undefined
```

webNeat: This would be better, I will use this.

packages/core/src/tools/limitModification.spec.ts
```diff
@@ -70,14 +70,29 @@ describe('limitModification', () => {
...
+    limitModification(object, ['foo.bar', 'qux'], modifier)
 
-    expect(result).toEqual({
+    expect(object).toEqual({
       foo: { bar: 'bar' },
       qux: 'qux',
     })
   })
 
+  it('should return the return of the modifier', () => {
```
BenoitZugmeyer, line 81:
```suggestion
  it('should return the result of the modifier', () => {
```

packages/core/src/tools/limitModification.ts
bcaudan: nice.
however, my point is that since customer can return anything, it can be better to internally handle result value as unknown.

packages/logs/src/boot/logs.ts
```diff
@@ -130,11 +130,14 @@ export function buildAssemble(session: LoggerSession, configuration: Configurati
       message
     )
     if (configuration.beforeSend) {
-      limitModification(
+      const result = limitModification(
```
bcaudan, line 133: what about shouldSend instead of result?

packages/core/src/tools/limitModification.ts
webNeat: Makes sense. I will change it's return type to `unknown`

packages/core/src/domain/configuration.ts
```diff
@@ -75,7 +75,7 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
     cookieOptions: CookieOptions
 
     service?: string
-    beforeSend?: (event: any) => void
+    beforeSend?: (event: any) => unknown
```
bcaudan, line 78: this type is used internally, it could be nice to update signatures exposed to customer to add the `| boolean` return type

webNeat: Done

bcaudan (approved):
Great!

PR #745, opened 2021-02-23
[RUM] Catch errors thrown by user callbacks

webNeat:
## Motivation

Catch errors thrown by user-provided callbacks and log them using `console.error`.

## Changes

- Added a new tools function `catchErrors`
- Used this function to wrap `beforeSend`, `onNewLocation` and callbacks given to `onReady` before the SDK initialization.

## Testing

- unit tests
---


packages/core/src/boot/init.ts
```diff
@@ -48,6 +49,9 @@ export interface BuildEnv {
 }
 
 export function commonInit(userConfiguration: UserConfiguration, buildEnv: BuildEnv) {
+  if (userConfiguration.beforeSend) {
+    userConfiguration.beforeSend = catchErrors(userConfiguration.beforeSend, 'beforeSend threw an error:')
+  }
```
bcaudan, line 54: IMO, this change is not at the right level of abstraction here, I'd move it inside the buildConfiguration function

packages/core/src/boot/init.ts
```diff
@@ -31,7 +32,7 @@ export function defineGlobal<Global, Name extends keyof Global>(global: Global,
   const existingGlobalVariable: { q?: Array<() => void> } | undefined = global[name]
   global[name] = api
   if (existingGlobalVariable && existingGlobalVariable.q) {
-    existingGlobalVariable.q.forEach((fn) => fn())
+    existingGlobalVariable.q.forEach((fn) => catchErrors(fn, 'Error thrown during SDK initialization:')())
```
BenoitZugmeyer, line 35:
```suggestion
    existingGlobalVariable.q.forEach((fn) => catchErrors(fn, 'onReady callback threw an error:')())
```

packages/core/src/tools/catchErrors.ts
```diff
@@ -0,0 +1,11 @@
+export function catchErrors<Args extends any[], R>(fn: (...args: Args) => R, errorMsg: string) {
```
BenoitZugmeyer, line 1: I would rename it to `catchUserErrors` to make it a bit clearer

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -41,7 +50,7 @@ export function trackViews(
   lifeCycle: LifeCycle,
   onNewLocation: NewLocationListener = () => undefined
 ) {
-  onNewLocation = wrapOnNewLocation(onNewLocation)
+  onNewLocation = catchErrors(onNewLocation, 'onNewLocation threw an error:')
```
bcaudan, line 53: I'd be in favor of moving this wrapping closer to when we check the configuration.
It could be nice to consistently perform this kind of wrapping during the same phase of the lifecycle and earlier the better imo.

packages/core/src/boot/init.ts
bcaudan, line 35: Could we have tests for this behavior of exposed APIs?

packages/core/src/tools/catchUserErrors.ts
```diff
@@ -0,0 +1,11 @@
+export function catchUserErrors<Args extends any[], R>(fn: (...args: Args) => R, errorMsg: string) {
+  return (...args: Args) => {
+    let result
+    try {
+      result = fn(...args)
+    } catch (err) {
+      console.error(errorMsg, err)
+    }
+    return result
```
bcaudan, line 9:
```suggestion
    try {
      return fn(...args)
    } catch (err) {
      console.error(errorMsg, err)
    }
```
do we need `result`?

packages/core/src/boot/init.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    const value = 'my value'
+    defineGlobal(myGlobal, 'foo', value)
+    expect(myGlobal.foo).toBe(value)
+    expect(fn1).toHaveBeenCalled()
+    expect(fn2).toHaveBeenCalled()
+  })
+
+  it('catches the errors thrown by the queued callbacks', () => {
+    const myError = 'Ooops!'
+    const fn = () => {
```
bcaudan, line 35:
```suggestion
    const onReady = () => {
```
since it is already in the error message, let's also make the test clearer

packages/core/src/boot/init.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      },
+    }
+    const value = 'my value'
+    defineGlobal(myGlobal, 'foo', value)
+    expect(myGlobal.foo).toBe(value)
+    expect(fn1).toHaveBeenCalled()
+    expect(fn2).toHaveBeenCalled()
+  })
+
+  it('catches the errors thrown by the queued callbacks', () => {
```
bcaudan, line 33: could we have similar tests for other wrapped APIs?

BenoitZugmeyer (approved):
LGTM

bcaudan (approved):
Nice!

PR #747, opened 2021-02-24
📝 Doc: remove resourceSampleRate

bcaudan:
## Motivation

No reason to promote this configuration parameter since it is not aligned with billing.
It only confuses customers and generate support questions.

## Changes

Remove resourceSampleRate from doc + add deprecated note in configuration

## Testing

NA

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -43,7 +43,7 @@ export interface UserConfiguration {
   internalMonitoringApiKey?: string
   allowedTracingOrigins?: Array<string | RegExp>
   sampleRate?: number
-  resourceSampleRate?: number
+  resourceSampleRate?: number // deprecated
   datacenter?: Datacenter // deprecated
   site?: string
   enableExperimentalFeatures?: string[]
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/configuration.ts
```diff
@@ -43,7 +43,7 @@ export interface UserConfiguration {
   internalMonitoringApiKey?: string
   allowedTracingOrigins?: Array<string | RegExp>
   sampleRate?: number
-  resourceSampleRate?: number
+  resourceSampleRate?: number // deprecated
```
webNeat, line 46: Why are we only deprecating it instead of removing it directly?

bcaudan: it can currently be used by some customers, we should not break the behavior if they are relying on it

BenoitZugmeyer (approved):

packages/core/src/domain/configuration.ts
webNeat: Makes sense. Does that mean that we will always keep it? or we may remove it sometime in the future?

bcaudan: As long as we provide the SDK with a CDN file that updates itself, it can be tricky to remove deprecated features.
However, some ongoing reflexion on extra monitoring could help us to clean up deprecated features at some point.

webNeat (approved):

jtappa (approved):
👍🏻  from docs

PR #748, opened 2021-02-24
⚗ [RUMF-858] add monotonic batch time

bcaudan:
## Motivation

in order to monitor behavior of monotonic clock compared to system clock

## Changes

add current monotonic time when sending batch

## Testing

manual

---

### DIFF
--- a/packages/core/src/transport/transport.ts
+++ b/packages/core/src/transport/transport.ts
@@ -1,5 +1,5 @@
 import { Context } from '../tools/context'
-import { addEventListener, DOM_EVENT, jsonStringify, noop, objectValues } from '../tools/utils'
+import { addEventListener, DOM_EVENT, getTimestamp, jsonStringify, noop, objectValues } from '../tools/utils'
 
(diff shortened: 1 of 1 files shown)


packages/core/src/transport/transport.ts
```diff
@@ -30,7 +30,9 @@ export class HttpRequest {
 }
 
 function addBatchTime(url: string) {
-  return `${url}${url.indexOf('?') === -1 ? '?' : '&'}batch_time=${new Date().getTime()}`
+  return `${url}${url.indexOf('?') === -1 ? '?' : '&'}batch_time=${new Date().getTime()}&m_time=${getTimestamp(
```
BenoitZugmeyer, line 33: I guess `m` stands for `monotonic`?

BenoitZugmeyer (approved):

packages/core/src/transport/transport.ts
bcaudan: yep, not sure we need too bee too explicit about it

webNeat (approved):
LGTM!

PR #751, opened 2021-02-25
♻️ [RUMF-860] Improve timing types

bcaudan:
## Motivation

Avoid using time value with inappropriate meaning.

## Changes

Introduce new types:

- Duration
- ServerDuration
- RelativeTime
- Timestamp

Use them across the code base

## Testing

ci

---

### DIFF
--- a/packages/core/src/tools/timeUtils.ts
+++ b/packages/core/src/tools/timeUtils.ts
@@ -8,7 +8,7 @@ export type RelativeTime = number & { r: 'Time relative to navigation start' } &
 export function toServerDuration(duration: Duration): ServerDuration
 export function toServerDuration(duration: Duration | undefined): ServerDuration | undefined
 export function toServerDuration(duration: Duration | undefined) {
-  if (!isNumber<Duration>(duration)) {
+  if (!isNumber(duration)) {
     return duration
   }
   return round(duration * 1e6, 0) as ServerDuration
--- a/packages/core/src/tools/timeUtils.ts
+++ b/packages/core/src/tools/timeUtils.ts
@@ -1,0 +1,44 @@
+import { isNumber, round } from './utils'
+
+export type Duration = number & { d: 'Duration in ms' }
+export type ServerDuration = number & { s: 'Duration in ns' }
+export type TimeStamp = number & { t: 'Epoch time' }
+export type RelativeTime = number & { r: 'Time relative to navigation start' } & { d: 'Duration in ms' }
+
+export function toOptionalServerDuration(duration: Duration | undefined): ServerDuration | undefined {
+  if (!isNumber<Duration>(duration)) {
+    return duration
+  }
+  return toServerDuration(duration)
+}
+
+export function toServerDuration(duration: Duration) {
+  return round(duration * 1e6, 0) as ServerDuration
+}
+
+/**
+ * Get the time since the navigation was started.
+ *
+ * Note: this does not use `performance.timeOrigin` because it doesn't seem to reflect the actual
+ * time on which the navigation has started: it may be much farther in the past, at least in Firefox 71.
+ * Related issue in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1429926
+ */
+export function getRelativeTime(timestamp: TimeStamp) {
+  return (timestamp - getNavigationStart()) as RelativeTime
+}
+
+export function getTimeStamp(relativeTime: RelativeTime) {
+  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+  return Math.floor(getNavigationStart() + relativeTime) as TimeStamp
+}
+
+/**
+ * Navigation start slightly change on some rare cases
+ */
+let navigationStart: TimeStamp | undefined
+function getNavigationStart() {
+  if (navigationStart === undefined) {
+    navigationStart = performance.timing.navigationStart as TimeStamp
+  }
+  return navigationStart
+}
--- a/packages/core/src/tools/timeUtils.ts
+++ b/packages/core/src/tools/timeUtils.ts
@@ -14,6 +14,20 @@ export function toServerDuration(duration: Duration | undefined) {
   return round(duration * 1e6, 0) as ServerDuration
 }
 
+export function timeStampNow() {
+  return Date.now() as TimeStamp
+}
+
+export function relativeNow() {
+  return performance.now() as RelativeTime
+}
+
+export function elapsed(start: TimeStamp, end: TimeStamp): Duration
+export function elapsed(start: RelativeTime, end: RelativeTime): Duration
+export function elapsed(start: number, end: number) {
+  return (end - start) as Duration
+}
+
 /**
  * Get the time since the navigation was started.
  *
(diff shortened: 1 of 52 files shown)


packages/core/src/tools/timeUtils.ts
```diff
@@ -0,0 +1,44 @@
+import { isNumber, round } from './utils'
+
+export type Duration = number & { d: 'Duration in ms' }
+export type ServerDuration = number & { s: 'Duration in ns' }
+export type TimeStamp = number & { t: 'Epoch time' }
+export type RelativeTime = number & { r: 'Time relative to navigation start' } & { d: 'Duration in ms' }
```
bcaudan, line 6: Use different type keys to allow composition, we use RelativeTime as a Duration in ms

packages/core/src/tools/timeUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+export function toOptionalServerDuration(duration: Duration | undefined): ServerDuration | undefined {
+  if (!isNumber<Duration>(duration)) {
+    return duration
+  }
+  return toServerDuration(duration)
+}
+
+export function toServerDuration(duration: Duration) {
+  return round(duration * 1e6, 0) as ServerDuration
+}
```
bcaudan, line 17: Previous type strategy was not working since "normal" input and output types are now different
```ts
export function msToNs<T>(duration: number | T): number | T 
```
Do you see any trick to have a single function?

packages/core/src/tools/timeUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+ * time on which the navigation has started: it may be much farther in the past, at least in Firefox 71.
+ * Related issue in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1429926
+ */
+export function getRelativeTime(timestamp: TimeStamp) {
+  return (timestamp - getNavigationStart()) as RelativeTime
+}
+
+export function getTimeStamp(relativeTime: RelativeTime) {
+  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+  return Math.floor(getNavigationStart() + relativeTime) as TimeStamp
```
bcaudan, line 32: There are several disable like that since it is not string or number around the operand, no similar check on other operands though.
IMO, keeping the rule is still relevant

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -268,11 +268,11 @@ function computeRelativePerformanceTiming() {
   const result: Partial<IndexedPerformanceTiming> = {}
   const timing = performance.timing as IndexedPerformanceTiming
   for (const key in timing) {
-    if (isNumber(timing[key])) {
-      result[key] = timing[key] === 0 ? 0 : getRelativeTime(timing[key] as number)
+    if (isNumber<TimeStamp>(timing[key])) {
+      result[key] = timing[key] === 0 ? 0 : getRelativeTime(timing[key])
```
BenoitZugmeyer, line 272: I would keep the explicit cast here instead of using a parametric type: `getRelativeTime(timing[key] as TimeStamp)`

BenoitZugmeyer (commented):
I would add some helpers to avoid a few casts like:

```ts
function timestampNow() {
  return Date.now() as Timestamp
}
function relativeNow() {
  return performance.now() as RelativeTime
}

function ellapsed(start: Timestam, end: Timestamp): Duration
function ellapsed(start: RelativeTime, end: RelativeTime): Duration
function ellapsed(start: number, end: number) {
  return (end - start) as Duration
}
```

packages/core/src/tools/timeUtils.ts
BenoitZugmeyer: Maybe something like this?
```ts
export function toOptionalServerDuration(duration: Duration | undefined): ServerDuration | undefined {
  if (!isNumber<Duration>(duration)) {
    return duration
  }
  return toServerDuration(duration)
}
export function toServerDuration(duration: Duration): ServerDuration
export function toServerDuration(duration: undefined): undefined
export function toServerDuration(duration: Duration | undefined) {
  if (!isNumber(duration)) {
    return undefined
  }
  return round(duration * 1e6, 0) as ServerDuration
}
```

packages/core/src/tools/timeUtils.ts
```diff
@@ -0,0 +1,44 @@
+import { isNumber, round } from './utils'
+
+export type Duration = number & { d: 'Duration in ms' }
+export type ServerDuration = number & { s: 'Duration in ns' }
```
webNeat, line 4: Can we make this more explicit? it's not obvious that `ServerDuration` means a duration in ns. What about `NsDuration`? maybe we can also name the other `MsDuration`.

bcaudan: > it's not obvious that ServerDuration means a duration in ns.

IMO it is more interesting to be explicit about when we should use this type rather than its content.

IMO, since the js APIs return ms, prefixing with `ms` or something like `internal` would add some unnecessary noise to the Duration type.

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -268,11 +268,11 @@ function computeRelativePerformanceTiming() {
...
     }
   }
-  return result as PerformanceTiming
+  return result as Omit<RumPerformanceResourceTiming, 'entryType'> & Omit<RumPerformanceNavigationTiming, 'entryType'>
```
webNeat, line 266: I don't understand this cast, as `RumPerformanceResourceTiming` contains fields that are not in result
```
initiatorType: string
name: string
```

I think we want to return a type like
```
type TypedPerformanceTiming = {
  [key in keyof PerformanceTiming]: TimeStamp
}
```

packages/core/src/tools/timeUtils.ts
webNeat: Ok, makes sense.

packages/rum-core/src/browser/performanceCollection.ts
bcaudan: We are using performance.timing to manually generate:
- a resource entry, for the request timings of the initial doc
- a navigation entry, if it is not supported through the observer
But you are right, we can probably do something better here, I'll look into it

BenoitZugmeyer (approved):
LGTM, great work

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -250,19 +250,24 @@ function retrieveFirstInputTiming(callback: (timing: RumFirstInputTiming) => voi
...
+}
+
+export type RelativePerformanceTiming = {
+  -readonly [key in keyof Omit<PerformanceTiming, 'toJSON'>]: RelativeTime
 }
 
 function computeRelativePerformanceTiming() {
-  const result: Partial<IndexedPerformanceTiming> = {}
+  const result: Partial<RelativePerformanceTiming> = {}
   const timing = performance.timing as IndexedPerformanceTiming
```
webNeat, line 265: Do we really need this cast? to my understanding it only makes the type of `timing[numberKey]` bellow `any` instead of `number`.

bcaudan: Indeed, by being more explicit about used keys, we don't need to have this IndexedType anymore 👍

webNeat (approved):
Good job!

PR #753, opened 2021-03-02
[RUMF-845] remove slimDOM option and enable strategies to ignore elements

BenoitZugmeyer:
## Motivation

A bit less code (removed option) and lighter snapshots ("slim DOM" option always enabled)

## Changes

* Remove the "slim DOM" option
* Always apply all "slim DOM" strategies
* Add some unit testings

## Testing

Unit, manual

---

### DIFF
--- a/packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.spec.ts
+++ b/packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.spec.ts
@@ -1,4 +1,5 @@
 /* eslint-disable max-len */
+import { isIE } from '@datadog/browser-core'
 import { absoluteToStylesheet, IGNORED_NODE, serializeNodeWithId } from './snapshot'
 
 describe('absolute url to stylesheet', () => {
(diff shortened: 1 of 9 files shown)


packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.spec.ts
```diff
@@ -73,3 +73,52 @@ describe('absolute url to stylesheet', () => {
...
+
+describe('serializeNodeWithId', () => {
+  describe('ignores some nodes', () => {
+    const defaultOptions = {
+      doc: document,
+      skipChild: false,
+      map: {},
+    }
+
+    it('does not same ignored nodes in the map', () => {
```
webNeat, line 85: I think there is a typo here, should it be `does not save ignored nodes in the map` ?

BenoitZugmeyer: Good catch!

webNeat (approved):
Nice!

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.spec.ts
```diff
@@ -73,3 +73,52 @@ describe('absolute url to stylesheet', () => {
...
+      map: {},
+    }
+
+    it('does not save ignored nodes in the map', () => {
+      const map = {}
+      serializeNodeWithId(document.createElement('script'), { ...defaultOptions, map })
+      expect(map).toEqual({})
+    })
+
+    it('sets the serialized node id to IGNORED_NODE', () => {
```
bcaudan, line 91:
```suggestion
    it('sets ignored serialized node id to IGNORED_NODE', () => {
```

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.spec.ts
```diff
@@ -73,3 +73,52 @@ describe('absolute url to stylesheet', () => {
     expect(absoluteToStylesheet(`url('')`, href)).toEqual(`url('')`)
   })
 })
+
+describe('serializeNodeWithId', () => {
```
bcaudan, line 77: cf BS unit tests, adding those tests expose some IE incompatibility in snapshot code (array.from)

bcaudan (approved):

PR #755, opened 2021-03-03
⚗ performance impact summary tool

BenoitZugmeyer:
## Motivation

Evaluate the Browser SDK performance impact on a given scenario

## Changes

Implement a tool to run various profilings with puppeteer. You can try it with:

## Testing

```
git checkout benoit/performance-doc
yarn
cd performances
yarn start
```

Example output:

```
# Wikipedia

Illustrates a mostly static site scenario.

* Navigate on three Wikipedia articles
* Do a search (with dynamic autocompletion) and go the first result

Memory (median): 294 kB (4.83%)
CPU: 258 ms (4.24%)
Bandwidth:
  upload: 116 kB
  download: 20.5 kB


# Twitter

Illustrates a SPA scenario.

* Navigate to the top trending topics
* Click on the first trending topic
* Click on Top, Latest, People, Photos and Videos tabs
* Navigate to the Settings page
* Click on a few checkboxes

Memory (median): 225 kB (0.51%)
CPU: 391 ms (1.01%)
Bandwidth:
  upload: 186 kB
  download: 20.5 kB
```

From my observations, results are quite stable in general.

---

### DIFF
--- a/performances/README.md
+++ b/performances/README.md
@@ -1,0 +1,5 @@
+# Browser SDK performances impact estimation tool
+
+This tool will run various scenarios in a browser and profile the impact of the Browser SDK.
+
+Use `yarn start` to execute it.
--- a/performances/README.md
+++ b/performances/README.md
@@ -1,5 +1,5 @@
 # Browser SDK performances impact estimation tool
 
-This tool will run various scenarios in a browser and profile the impact of the Browser SDK.
+This tool runs various scenarios in a browser and profile the impact of the Browser SDK.
 
 Use `yarn start` to execute it.
--- a/performances/src/main.ts
+++ b/performances/src/main.ts
@@ -58,10 +58,20 @@ async function profileScenario(
 async function runWikipediaScenario(page: Page, takeMeasurements: () => Promise<void>) {
   await page.goto('https://en.wikipedia.org/wiki/Event_monitoring')
   await takeMeasurements()
+
   await page.goto('https://en.wikipedia.org/wiki/Datadog')
   await takeMeasurements()
+
+  await page.type('[type="search"]', 'median', {
+    // large delay to trigger the autocomplete menu at each key press
+    delay: 400,
+  })
+  await Promise.all([page.waitForNavigation(), page.keyboard.press('Enter')])
+  await takeMeasurements()
+
   await page.goto('https://en.wikipedia.org/wiki/Ubuntu')
   await takeMeasurements()
+
   await page.goto('about:blank')
 }
 
--- a/performances/src/main.ts
+++ b/performances/src/main.ts
@@ -5,7 +5,10 @@ import { startProfiling } from './profiling'
 import { trackNetwork } from './trackNetwork'
 import { ProfilingResults, ProfilingOptions } from './types'
 
-main().catch(console.error)
+main().catch((error) => {
+  console.error(error)
+  process.exitCode = 1
+})
 
 async function main() {
   const options: ProfilingOptions = {
--- a/performances/src/main.ts
+++ b/performances/src/main.ts
@@ -9,18 +9,21 @@ async function main() {
   try {
     const page = await browser.newPage()
     await setupSDK(page)
-    const stopProfiling = await startProfiling(page)
-    await runScenario(page)
+    const { stopProfiling, takeMeasurements } = await startProfiling(page)
+    await runScenario(page, takeMeasurements)
     await stopProfiling()
   } finally {
     await browser.close()
   }
 }
 
-async function runScenario(page: Page) {
-  await page.goto('https://en.wikipedia.org/wiki/Ubuntu')
-  await page.goto('https://en.wikipedia.org/wiki/Datadog')
+async function runScenario(page: Page, takeMeasurements: () => Promise<void>) {
   await page.goto('https://en.wikipedia.org/wiki/Event_monitoring')
+  await takeMeasurements()
+  await page.goto('https://en.wikipedia.org/wiki/Datadog')
+  await takeMeasurements()
+  await page.goto('https://en.wikipedia.org/wiki/Ubuntu')
+  await takeMeasurements()
   await page.goto('about:blank')
 }
 
--- a/performances/src/main.ts
+++ b/performances/src/main.ts
@@ -17,17 +17,31 @@ async function main() {
   }
 
   const wikipediaResults = await profileScenario(options, runWikipediaScenario)
-
-  console.log('# Wikipedia:')
-  console.log()
-  console.log(formatProfilingResults(wikipediaResults))
-  console.log()
-
   const twitterResults = await profileScenario(options, runTwitterScenario)
 
-  console.log('# Twitter:')
-  console.log()
-  console.log(formatProfilingResults(twitterResults))
+  console.log(`
+# Wikipedia
+
+Illustrates a mostly static site scenario.
+
+* Navigate on three Wikipedia articles
+* Do a search (with dynamic autocompletion) and go to the first result
+
+${formatProfilingResults(wikipediaResults)}
+
+
+# Twitter
+
+Illustrates a SPA scenario.
+
+* Navigate to the top trending topics
+* Click on the first trending topic
+* Click on Top, Latest, People, Photos and Videos tabs
+* Navigate to the Settings page
+* Click on a few checkboxes
+
+${formatProfilingResults(twitterResults)}
+`)
 }
 
 async function profileScenario(
--- a/performances/src/main.ts
+++ b/performances/src/main.ts
@@ -1,0 +1,39 @@
+import puppeteer, { Page } from 'puppeteer'
+
+import { startProfiling } from './profiling'
+
+main().catch(console.error)
+
+async function main() {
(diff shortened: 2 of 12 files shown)


performances/src/main.ts
```diff
@@ -0,0 +1,116 @@
...
+    result = await stopProfiling()
+  } finally {
+    await browser.close()
+  }
+  return result
+}
+
+async function runWikipediaScenario(page: Page, takeMeasurements: () => Promise<void>) {
+  await page.goto('https://en.wikipedia.org/wiki/Event_monitoring')
+  await takeMeasurements()
```
bcaudan, line 54: What about adding some activity on the page?
search or languages cog seems to trigger some requests

ThibautGeriz:
Regarding the stat I would add more than the median like 95 and 99 percentile.
The CPU is not super clear:

> CPU: 146,665 microseconds (6.455%)

Is it microsecond per something? seconds? session?

It would be super cool to add this to graph that to see the evolution in time.

performances/src/main.ts
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    const page = await browser.newPage()
+    await setupSDK(page, options)
+    const { stopProfiling, takeMeasurements } = await startProfiling(options, page)
+    await runScenario(page, takeMeasurements)
+    result = await stopProfiling()
+  } finally {
+    await browser.close()
+  }
+  return result
```
bcaudan, line 49: it could be nice to exit main with error if one scenario fails

BenoitZugmeyer:
> Regarding the stat I would add more than the median like 95 and 99 percentile.

Maybe p75 would make sense, but 95 or 99 would just show the max value, because we don't take that many measurements.

> The CPU is not super clear:
> 
> > CPU: 146,665 microseconds (6.455%)
> 
> Is it microsecond per something? seconds? session?

It is plain microseconds. How much CPU time Browser SDK-located code has taken during the scenario.

> 
> It would be super cool to add this to graph that to see the evolution in time.

Would be cool indeed :)

performances/src/profiling.ts
```diff
@@ -0,0 +1,159 @@
...
+export async function startProfiling(options: ProfilingOptions, page: Page) {
+  const client = await page.target().createCDPSession()
+  const stopCPUProfiling = await startCPUProfiling(options, client)
+  const { stopMemoryProfiling, takeMemoryMeasurements } = await startMemoryProfiling(options, client)
+  const stopNetworkProfiling = await startNetworkProfiling(options, client)
+
+  return {
+    takeMeasurements: async () => {
+      await takeMemoryMeasurements()
+    },
```
webNeat, line 13: It seems that this can be simply rewritten to 
```ts
takeMeasurements: takeMemoryMeasurements
```

performances/src/profiling.ts
```diff
@@ -0,0 +1,159 @@
...
+  })
+
+  const measurements: Array<{ sdkConsumption: number; totalConsumption: number }> = []
+
+  return {
+    takeMemoryMeasurements: async () => {
+      await client.send('HeapProfiler.collectGarbage')
+      const { profile } = await client.send('HeapProfiler.getSamplingProfile')
+
+      const sizeForNodeId = new Map<number, number>()
```
webNeat, line 64: I wonder why using a `Map` here instead of a simple object `{}`, is there a benefit?

BenoitZugmeyer: I'd like to ask, is there any benefit to use `{}` instead of `Map` ? 😃 

I tend to use `Map` wherever possible, because using `{}` has too many subtleties to worry about, like:
* if my key is not a string, is it safe to cast it as a string?
* what if my key clashes with object methods? should I use `map.hasOwnProperty(key)`, `key in map`, `Object.prototype.hasOwnProperty.call(map, key)`?
* what happen if the iteration order changes?

Even if those questions may not apply in this particular use-case, I find that it's good practice to always use `Map` instead of `{}` when indexing values in a map.

performances/src/profiling.ts
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+      measurements.push({ totalConsumption, sdkConsumption })
+    },
+
+    stopMemoryProfiling: async () => {
+      await client.send('HeapProfiler.stopSampling')
+
+      measurements.sort((a, b) => a.sdkConsumption - b.sdkConsumption)
+      const { sdkConsumption, totalConsumption } = measurements[Math.floor(measurements.length / 2)]
```
webNeat, line 86: FMU here we are sorting the measurements and taking the one in the middle. This means that if for example the measurements are:
```
[
  { sdkConsumption: 1, totalConsumption: 100 },
  { sdkConsumption: 2, totalConsumption: 100 },
  { sdkConsumption: 50, totalConsumption: 100 },
]
```
We will return `{ sdkConsumption: 2, totalConsumption: 100 }`.
This doesn't fail right to me, maybe computing the average will make more sense.

BenoitZugmeyer: Yes this is how we compute a median value! In performance analysis I felt like we tend to prefer medians over averages, so we can compare with various percentiles. But I may be wrong here, no strong opinion on the subject.

performances/src/main.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+async function runWikipediaScenario(page: Page, takeMeasurements: () => Promise<void>) {
+  await page.goto('https://en.wikipedia.org/wiki/Event_monitoring')
+  await takeMeasurements()
+  await page.goto('https://en.wikipedia.org/wiki/Datadog')
+  await takeMeasurements()
+  await page.goto('https://en.wikipedia.org/wiki/Ubuntu')
+  await takeMeasurements()
+  await page.goto('about:blank')
```
webNeat, line 59: Is this really needed?

BenoitZugmeyer: Yes, to flush the last page RUM events.

performances/src/profiling.ts
webNeat: Interesting, thanks for the explanation, I was aware of the `hasOwnProperty` thing, but not others.

performances/src/main.ts
webNeat: In that case, should we take additional measurements after it?

performances/src/main.ts
```diff
@@ -0,0 +1,116 @@
...
+    await browser.close()
+  }
+  return result
+}
+
+async function runWikipediaScenario(page: Page, takeMeasurements: () => Promise<void>) {
+  await page.goto('https://en.wikipedia.org/wiki/Event_monitoring')
+  await takeMeasurements()
+  await page.goto('https://en.wikipedia.org/wiki/Datadog')
+  await takeMeasurements()
```
webNeat, line 56: I am not sure if it would be statistically better to run `takeMeasurements` periodically (like every 100 ms or so) instead of triggering it manually during the scenario. WDYT?

performances/src/main.ts
BenoitZugmeyer: Measurements are computed for the current page, at one point in time. On `about:blank`, measurements would always be null, because RUM is not setup on this page.

ThibautGeriz:
> It is plain microseconds. How much CPU time Browser SDK-located code has taken during the scenario.

Cool then I would add the time it took to run the scenario as 146,665 microseconds or upload: 162,482 bytes is not the same if the scenario lasted 1h ou 2.5 seconds.

performances/src/main.ts
BenoitZugmeyer: I don't think it would matter too much. I like controling when to take measurements, when the page is idle after doing something. If we didn't:

* measurements could be taken while the page is loading, when the SDK is not yet loaded, yielding not applicable results for our use-case
* measurements could be taken while some JS is running, and may impact the CPU profiling.

performances/README.md
```diff
@@ -0,0 +1,5 @@
+# Browser SDK performances impact estimation tool
+
+This tool will run various scenarios in a browser and profile the impact of the Browser SDK.
```
sarina-dd, line 3:
```suggestion
This tool runs various scenarios in a browser and profile the impact of the Browser SDK.
```

BenoitZugmeyer:
> > It is plain microseconds. How much CPU time Browser SDK-located code has taken during the scenario.
> 
> Cool then I would add the time it took to run the scenario as 146,665 microseconds or upload: 162,482 bytes is not the same if the scenario lasted 1h ou 2.5 seconds.

Well, it could be the same if the scenario lasts 1h or 2.5 seconds. The impact is not proportional to the scenario duration, but more on which actions are being done during it (ex: number of new views, resources loaded, etc.) The scenario duration would not be indicative of real world SDK usage, because no human will read 5 Wikipedia articles in 2 seconds :)

performances/README.md
BenoitZugmeyer: Thank you!

performances/src/profiling.ts
bcaudan: I'd also choose median to avoid being too sensitive to outliers (if any), since it is mentioned on the report I'd say it fine

bcaudan (approved):
Nice 👍

webNeat (approved):
Good job!

PR #759, opened 2021-03-09
🔊 [RUMF-855] add negative FID monitoring

BenoitZugmeyer:
## Motivation

Help debug the negative FID, see comment 1 in https://bugs.chromium.org/p/chromium/issues/detail?id=1185815#c1

## Changes

Add internal monitoring log when the FID is negative

## Testing

Manual

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackTimings.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackTimings.ts
@@ -1,4 +1,12 @@
-import { addEventListeners, DOM_EVENT, Duration, elapsed, EventEmitter, RelativeTime } from '@datadog/browser-core'
+import {
+  addEventListeners,
+  addMonitoringMessage,
+  DOM_EVENT,
+  Duration,
+  elapsed,
+  EventEmitter,
+  RelativeTime,
+} from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType } from '../../lifeCycle'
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/domain/rumEventsCollection/view/trackTimings.ts
```diff
@@ -137,17 +145,19 @@ export function trackFirstInputTimings(
...
-        firstInputTime: entry.startTime as Duration,
-      })
+      if (entry.startTime <= entry.processingStart) {
+        callback({
+          firstInputDelay: elapsed(entry.startTime, entry.processingStart),
+          firstInputTime: entry.startTime as Duration,
+        })
+      } else if (entry.toJSON) {
+        addMonitoringMessage(`Negative FID for entry ${entry.name}`, {
+          entry: { ...entry.toJSON(), target: undefined },
```
bcaudan, line 158: could it be interesting to have the negative value in the log as well?

BenoitZugmeyer: So I don't think so, since we can compute it simply by doing `processingStart - startTime`, but I can add it if you want

bcaudan: it could be more convenient if we want to quickly see the value or perform some query on it
I'd be in favor of adding it, just in case

bcaudan (approved):

webNeat (approved):
LGTM

PR #761, opened 2021-03-09
[RUMF-862] fix export MediaInteractions enum

BenoitZugmeyer:
## Motivation

Exporting `const enum` values only works if the user builds their package with TypeScript. Other bundlers don't use type informations, so can't use `const enum` values.

## Changes

* Change `MediaInteractions` to a standard enum
* Add an ESLint rule to avoid this kind of issue in the future

## Testing

Manually

---

### DIFF
--- a/eslint-local-rules/disallowConstEnumExports.js
+++ b/eslint-local-rules/disallowConstEnumExports.js
@@ -1,0 +1,67 @@
+const { SymbolFlags } = require('typescript')
+
+/**
+ * This rule forbids exporting 'const enums'.
+ *
+ * 'const enums' are useful in TS to optimize the generated code, but they don't produce any JS code
+ * on their own[0].  So, exporting them in public packages entry points should be avoided, because
+ * we don't know how public packages are used: it could be used by other builders like babel which
+ * don't have access to the type definition, so won't be able to access the enum value.
+ *
+ * Exporting 'const enums' from internal modules to be consumed by internal modules is fine, because
+ * we know that everything will be handled by TypeScript.
+ *
+ * So, this rule should not be applied on all the source code, but only on entry points that will be
+ * used in the wild.
+ *
+ * In the future, this rule could be removed and the 'isolatedModules' TS option[1] could be used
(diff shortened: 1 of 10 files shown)


eslint-local-rules/disallowConstEnumExports.js
```diff
@@ -0,0 +1,67 @@
+const { SymbolFlags } = require('typescript')
+
+/**
+ * This rule forbids exporting 'const enums'.
+ *
+ * 'const enums' are useful in TS to optimize the generated code, but they don't produce any JS code
+ * on their own[0].  So, exporting them in public packages entry points should be avoided, because
+ * we don't know how public packages are used: it could be used by other builders like babel which
+ * don't have access to the type definition, so won't be able to access the enum value.
```
bcaudan, line 9: FMU, the goal of `const enum` is to have the type info but not the js structure, so the behavior kinda make sense but seems really confusing from a client perspective. 
Not sure what was the original intention for `MediaInteractions` though.

Since we already had an [issue with exported enums](https://github.com/DataDog/browser-sdk/pull/630), should we not prevent enum exports?

packages/rum-recorder/src/domain/rrweb/types.ts
```diff
@@ -147,18 +147,20 @@ export interface MousePosition {
...
+export const MouseInteractions = {
+  MouseUp: 0 as const,
+  MouseDown: 1 as const,
+  Click: 2 as const,
+  ContextMenu: 3 as const,
+  DblClick: 4 as const,
+  Focus: 5 as const,
+  Blur: 6 as const,
+  TouchStart: 7 as const,
+  TouchEnd: 9 as const,
```
bcaudan, line 159: FMU, it would behave the same with
```ts
export const xxx = {
  foo: 1,
  ...
} as const
```

and it could remove some `as const` noise

bcaudan (approved):

packages/rum-recorder/src/domain/rrweb/types.ts
BenoitZugmeyer: Nice!

PR #762, opened 2021-03-09
⚗ [RUMF-863] rework console error calls containing error instances

bcaudan:
## Motivation

Ease Error Tracking grouping

## Changes

behind `console-stack` FF:

- remove error stack from raw error message
- set raw error stack to first error instance stack

## Testing

unit tests + manual

---

### DIFF
--- a/packages/core/src/domain/automaticErrorCollection.spec.ts
+++ b/packages/core/src/domain/automaticErrorCollection.spec.ts
@@ -58,7 +58,9 @@ describe('console tracker (console-stack disabled)', () => {
 
   it('should format error instance', () => {
     console.error(new TypeError('hello'))
-    expect((notifyError.calls.mostRecent().args[0] as RawError).message).toContain('console error: TypeError: hello')
+    expect((notifyError.calls.mostRecent().args[0] as RawError).message).toMatch(
+      /^console error: TypeError: hello\s+at/
+    )
   })
 })
 
@@ -92,6 +94,7 @@ describe('console tracker (console-stack enabled)', () => {
     expect(notifyError).toHaveBeenCalledWith({
       ...CONSOLE_CONTEXT,
       message: 'console error: foo bar',
+      stack: undefined,
       startTime: jasmine.any(Number),
     })
   })
@@ -101,13 +104,19 @@ describe('console tracker (console-stack enabled)', () => {
     expect(notifyError).toHaveBeenCalledWith({
       ...CONSOLE_CONTEXT,
       message: 'console error: Hello {\n  "foo": "bar"\n}',
+      stack: undefined,
       startTime: jasmine.any(Number),
     })
   })
 
   it('should format error instance', () => {
     console.error(new TypeError('hello'))
-    expect((notifyError.calls.mostRecent().args[0] as RawError).message).toContain('console error: TypeError: hello')
+    expect((notifyError.calls.mostRecent().args[0] as RawError).message).toBe('console error: TypeError: hello')
+  })
+
+  it('should extract stack from first error', () => {
+    console.error(new TypeError('foo'), new TypeError('bar'))
+    expect((notifyError.calls.mostRecent().args[0] as RawError).stack).toMatch(/^TypeError: foo\s+at/)
   })
 })
 
--- a/packages/core/src/domain/automaticErrorCollection.spec.ts
+++ b/packages/core/src/domain/automaticErrorCollection.spec.ts
@@ -13,7 +13,7 @@ import {
 } from './automaticErrorCollection'
 import { Configuration } from './configuration'
 
-describe('console tracker', () => {
+describe('console tracker (console-stack disabled)', () => {
   let consoleErrorStub: jasmine.Spy
   let notifyError: jasmine.Spy
   const CONSOLE_CONTEXT = {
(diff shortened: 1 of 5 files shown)


packages/core/src/domain/automaticErrorCollection.spec.ts
```diff
@@ -58,7 +58,9 @@ describe('console tracker (console-stack disabled)', () => {
 
   it('should format error instance', () => {
     console.error(new TypeError('hello'))
-    expect((notifyError.calls.mostRecent().args[0] as RawError).message).toContain('console error: TypeError: hello')
+    expect((notifyError.calls.mostRecent().args[0] as RawError).message).toMatch(
+      /^console error: TypeError: hello\s+at/
```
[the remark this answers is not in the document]
bcaudan: FMU, in this case, the stacktrace is only one part of formatting process (like adding console.error, error type, error message), any reason to focus on the stacktrace in the test description?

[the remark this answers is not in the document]
webNeat: Because it's what makes the difference between error formatting with console-stack flag disabled vs when enabled. Having the same test description "should format error instance" for both cases sounds like they have the same behaviour.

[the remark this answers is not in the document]
bcaudan: yes, it is just that formatting an error instance don't have the same meaning in both cases.
Nb: only one of those tests will survive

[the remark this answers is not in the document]
webNeat: Makes sense.

webNeat (approved):

packages/core/test/forEach.spec.ts
```diff
@@ -7,8 +8,11 @@ beforeEach(() => {
   ;(window as any).DD_RUM = {}
   // prevent 'Some of your tests did a full page reload!' issue
   window.onbeforeunload = () => 'stop'
+  // useful to debug monitored functions spec
+  setDebugMode(true)
```
BenoitZugmeyer, line 12: FMU, this does not work as intended. Because each `.spec.ts` file creates a bundle with its own dependencies, a given `.spec.ts` file will evaluate its own copy of `internalMonitoring.ts`, with its own local `monitoringConfiguration` variable. Setting it in the `forEach.spec.ts` won't do anything useful since it doesn't run any test.

packages/core/src/domain/automaticErrorCollection.ts
```diff
@@ -42,32 +42,45 @@ export function filterErrors(configuration: Configuration, errorObservable: Obse
...
+      ...buildErrorFromParams(configuration, params),
       source: ErrorSource.CONSOLE,
       startTime: relativeNow(),
     })
   })
 }
 
+function buildErrorFromParams(configuration: Configuration, params: unknown[]) {
+  if (configuration.isEnabled('console-stack')) {
+    const firstErrorParam = find(params, (param: unknown) => param instanceof Error) as Error
```
BenoitZugmeyer, line 61:
```suggestion
    const firstErrorParam = find(params, (param: unknown): param is Error => param instanceof Error)
```

packages/core/src/domain/automaticErrorCollection.ts
```diff
@@ -42,32 +42,45 @@ export function filterErrors(configuration: Configuration, errorObservable: Obse
...
 }
 
-function formatConsoleParameters(param: unknown) {
-  if (typeof param === 'string') {
-    return param
-  }
-  if (param instanceof Error) {
-    return toStackTraceString(computeStackTrace(param))
+function formatConsoleParameters(stackTraceFormatter: (stack: StackTrace) => string) {
+  return (param: unknown) => {
```
BenoitZugmeyer, line 75: IMO returning a function here makes it more complicated than necessary. I would do something like:

```ts
['console error:', ...params].map(param => formatConsoleParameters(formatErrorMessage, param)).join(' ')

function formatConsoleParameters(stackTraceFormatter: (stack: StackTrace) => string, param: unknown) { ... }
```

packages/core/test/forEach.spec.ts
bcaudan: indeed, I have moved that here without retesting it 🤦 
I'look into it

packages/core/src/domain/automaticErrorCollection.ts
bcaudan: that does not type `firstErrorParam` though 😕

BenoitZugmeyer: I would add this signature to `find` then: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4b9438a6ae9df8e1bcaccf68d149d1643273c828/types/lodash/common/collection.d.ts#L291

```ts
export function find<T, S extends T>(
  array: T[],
  predicate: (value: T, index: number, array: T[]) => value is S
): S | undefined

```

bcaudan: good reflex to check lodash!

packages/core/test/forEach.spec.ts
bcaudan: I have tried with a webpack string replacer approach but it messes with the specs that actually test the setDebug behavior or the internal monitoring...
I have reverted all that.

BenoitZugmeyer (approved):

Issue #308, opened 2020-03-17
Add ability to configure the https statuses for which error should be logged


jbojcic1:
Currently only responses with statuses 0 or >= 500 are logged. I'd like to log all responses with 404 status too so it would be great if this would be configurable.

hdelaby:
Hi @jbojcic1 

We added your feature request on our roadmap. We first need to define how exactly we want this option to work. We will let you know whenever it's ready.

jbojcic1:
I guess some function that would take request with response as a param and return the boolean flag (log or not to log) would be the most powerful option.

matheo:
any news about this to filter `404` XHR error?

PinPinIre:
I can't tell from reading the documentation but is it expected that the browser log only includes the correct status code if it is >500 and all other status codes are 0? Or are 0 status codes all the aborted requests? 

I am trying to work out whether the huge number of XHR errors I am seeing are valid or are aborted requests like mentioned here https://github.com/DataDog/browser-sdk/issues/362 .

BenoitZugmeyer:
For now, the SDK considers requests ending with a status 0 or >= 500 to be errors.

A request with status code 0 is not necessarily an aborted request, it can be a request that failed because of a connection failure (loss of connectivity...).

bryik:
Mapbox-gl-js frequently aborts requests if a user quickly zooms around a map (tiles rapidly go in and out of view). Within a few minutes of testing I inadvertently generated over 15,000 network errors. 

It would be really nice if there was a way to tell the browser-sdk to ignore these "errors".

BenoitZugmeyer:
Thank you for your feedback, this looks definitly like a SDK issue. We will look into it to improve the situation.

wmaca:
We are facing a similar situation where aborted requests are still being logged. Is this fixed? 

We are using the CDN version, and we are assuming the CND is delivering the latest version. 
Shouldn't it include this fix?

wmaca:
Thanks, @BenoitZugmeyer. In the meantime, we figured out how to use the experimental flag and, so far, for our use case, it seems to be working fine.

During this process, when we saw that `beforeSend` could return a boolean. We thought that by returning `false` it would avoid sending the log. Maybe this is the way you are considering evolving the `beforeSend` option.

Other problem that we currently face is that some errors, even when handled, are logged on console. This ends up creating two erro logs for us. Not sure how to address this one and uncertain this is the right issue to discuss this.

[...]

PR #768, opened 2021-03-18
[RUMF-853] don't emit error for requests aborted by the application

BenoitZugmeyer:
## Motivation

Calling `xhr.abort()` or aborting a fetch request via a signal should not be considered as an error.

Related: https://github.com/DataDog/browser-sdk/issues/308#issuecomment-776968366

## Changes

* detect aborted requests in fetch and xhr proxies
* don't consider those requests as error

## Testing

Unit

---


packages/core/src/browser/xhrProxy.spec.ts
```diff
@@ -120,13 +124,36 @@ describe('xhr proxy', () => {
...
         expect(request.duration).toEqual(jasmine.any(Number))
+        expect(request.isAborted).toBe(false)
         expect(xhr.status).toBe(0)
         expect(onReadyStateChange).toHaveBeenCalled()
         done()
       },
     })
   })
 
+  it('should track aborted requests', (done) => {
```
bcaudan, line 135: it could be nice to test different abort scenarios, I think of:
- abort call before state DONE
- abort call after state DONE but before loadend
- abort call after loadend

packages/core/src/browser/xhrProxy.ts
```diff
@@ -91,33 +93,47 @@ function proxyXhr() {
...
+          if (xhrContext.isAborted === undefined) {
+            xhrContext.isAborted = false
+          }
+        }
 
+        const originalOnreadystatechange = this.onreadystatechange
         this.onreadystatechange = function () {
           if (this.readyState === XMLHttpRequest.DONE) {
-            callMonitored(reportXhr)
+            // Try to complete the context as soon as possible, because the XHR may be mutated by
```
bcaudan, line 111: it seems that it was already the case but the onreadystate override strategy could lead to unexpected behaviors for cases where the xhr is "recycled":
```
var xhr = new XMLHttpRequest()
xhr.open('GET', 'http://localhost:8080')
xhr.send()
xhr.open('GET', 'http://localhost:8080')
xhr.send()
```

it could be handled in a different task though. 
wdyt?

packages/core/src/browser/xhrProxy.ts
```diff
@@ -91,33 +93,47 @@ function proxyXhr() {
...
+            // the application during a future event. For example, Angular is calling .abort() on
+            // completed requests during a onreadystatechange event, so the status becomes '0'
+            // before the request is collected.
+            callMonitored(completeContext)
```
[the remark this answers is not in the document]
BenoitZugmeyer: This is extremely unintuitive! Thanks for pointing this out. Should be handled by the new implementation.

packages/core/src/browser/xhrProxy.spec.ts
BenoitZugmeyer: Added as e2e tests.

packages/core/src/browser/xhrProxy.ts
BenoitZugmeyer: I reverted most of the code to the previous implementation, so it shouldn't introduce any regression at this level. We shall investigate this further in a future task. I'll open a card about this.

bcaudan (approved):
Nice job!

dhrrgn:
Great timing.  We just deployed this yesterday and are seeing hundreds of these (cancelled search-as-you-type requests).

Any idea when the next version will get tagged?

BenoitZugmeyer:
@dhrrgn I'll make a release tomorrow! (april 1st)

BenoitZugmeyer:
Sorry to disappoint, but we chose not to release this publicly yet. See https://github.com/DataDog/browser-sdk/pull/777 description for more information.

PR #771, opened 2021-03-23
[RUMF-867] implement stop recording

BenoitZugmeyer:
## Motivation

Implement a method to stop session replay recording

## Changes

* Add the method
* Make sure the last segment is flushed when stop recording
* Make sure the view keeps its has_replay field to `true`.

## Testing

Unit, manual

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -58,7 +58,6 @@ export function startRumAssembly(
           date: timeStampNow(),
           service: configuration.service,
           session: {
-            has_replay: commonContext.hasReplay,
             // must be computed on each event because synthetics instrumentation can be done after sdk execution
             // cf https://github.com/puppeteer/puppeteer/issues/3667
             type: getSessionType(),
@@ -71,6 +70,10 @@ export function startRumAssembly(
         const context = combine(commonContext.context, customerContext)
         if (!isEmptyObject(context)) {
           serverRumEvent.context = context
+        }
+
+        if (!('has_replay' in serverRumEvent.session)) {
+          ;(serverRumEvent.session as { has_replay?: boolean }).has_replay = commonContext.hasReplay
         }
 
         if (!isEmptyObject(commonContext.user)) {
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
@@ -1040,3 +1040,72 @@ describe('rum track custom timings', () => {
     expect(warnSpy).toHaveBeenCalled()
   })
 })
+
+describe('track hasReplay', () => {
+  let setupBuilder: TestSetupBuilder
+  let handler: jasmine.Spy
+  let getViewEvent: (index: number) => View
+
+  beforeEach(() => {
+    ;({ handler, getViewEvent } = spyOnViews())
+
+    setupBuilder = setup()
+      .withFakeLocation('/foo')
+      .withFakeClock()
+      .beforeBuild(({ location, lifeCycle }) => {
+        lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, handler)
+        return trackViews(location, lifeCycle)
+      })
+  })
+
+  afterEach(() => {
+    setupBuilder.cleanup()
+  })
+
+  it('sets hasReplay to false by default', () => {
+    setupBuilder.build()
+    expect(getViewEvent(0).hasReplay).toBe(false)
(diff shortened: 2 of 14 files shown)


packages/rum-core/src/domain/assembly.ts
```diff
@@ -73,6 +72,10 @@ export function startRumAssembly(
           serverRumEvent.context = context
         }
 
+        if (!('has_replay' in serverRumEvent.session)) {
+          ;(serverRumEvent.session as { has_replay?: boolean }).has_replay = commonContext.hasReplay
```
ThibautGeriz, line 76: weird `;` Should be at the end not ?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -1040,3 +1040,72 @@ describe('rum track custom timings', () => {
...
   })
 })
+
+describe('track hasReplay', () => {
+  let setupBuilder: TestSetupBuilder
+  let handler: jasmine.Spy
+  let getViewEvent: (index: number) => View
+
+  beforeEach(() => {
+    ;({ handler, getViewEvent } = spyOnViews())
```
ThibautGeriz, line 1050: weird `;` Should be at the end not ?

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.ts
```diff
@@ -54,6 +54,9 @@ function processViewUpdate(view: View) {
       },
       time_spent: toServerDuration(view.duration),
     },
+    session: {
+      has_replay: view.hasReplay || undefined,
```
ThibautGeriz, line 58: use `??` instead of `||`?

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -79,6 +79,9 @@ export interface RawRumViewEvent {
     long_task: Count
     resource: Count
   }
+  session: {
+    has_replay: true | undefined
```
ThibautGeriz, line 83: Semantically it would be better to have a boolean?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
BenoitZugmeyer: No, prettier inserts a semicolon when the line starts with a paren to avoid ASI issues

See: https://github.com/prettier/prettier/issues/1935#issuecomment-305968119

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: Yes but in practice, this field should not exist for backward and other SDK compatibility. If I want to select all events that don't have replay, using `@session.has_replay:false` would lead to incorrect results since not all SDKs implement this flag. Using `-@session.has_replay:true` is the way to go.

Using this weird typing forces us to ensure `has_replay` is never `false`.

packages/rum-recorder/src/domain/segmentCollection.ts
```diff
@@ -91,7 +95,7 @@ export function doStartSegmentCollection(
     { capture: true }
   )
 
-  function flushSegment(creationReason: CreationReason) {
+  function flushSegment(creationReason?: CreationReason) {
```
bcaudan, line 98: what about using a `stop` reason?
even if it is not used, it could be more consistent

BenoitZugmeyer: I'm not fond of this, because `CreationReason` is exposed to web-ui. Adding unused variant would add some noise. I'll rework things a bit to be more explicit on what's going on, let me know if it'll look fine then.

packages/rum-recorder/src/domain/segmentCollection.ts
```diff
@@ -33,32 +33,56 @@ export const MAX_SEGMENT_DURATION = 30_000
...
+    workerSingleton
   )
 }
 
+const enum StateType {
+  WaitingForInitialRecord,
+  SegmentPending,
+  Stopped,
+}
+type State =
```
bcaudan, line 61: I'd be in favor of having more explicit names for State/StateType, what about something like SegmentCollectionState/SegmentCollectionStatus?

packages/rum-recorder/src/domain/segmentCollection.ts
bcaudan: It is much clearer indeed

bcaudan (approved):
LGTM!

PR #779, opened 2021-04-01
[RUMF-869] Ensure the "Focus" records are emited in the same segment as the "FullSnapshot"

BenoitZugmeyer:
## Motivation

Previously, the Focus record logic was uncorrelated with the FullSnapshot logic, and we have seen cases where Focus records where emited in a segment before the initial FullSnapshot record (because rrweb waits for `onload` before starting to generate records)

## Changes

This PR moves the Focus record logic to be closer to the FullSnapshot logic and uses the same guarantees as the Meta record to ensure all three events are correctly added in the same segment.

## Testing

Unit, manual

---

### DIFF
--- a/packages/rum-recorder/src/boot/recorder.spec.ts
+++ b/packages/rum-recorder/src/boot/recorder.spec.ts
@@ -164,7 +164,7 @@ describe('startRecording', () => {
 
   // eslint-disable-next-line max-len
   it('does not split Meta, Focus and FullSnapshot records between multiple segments when taking a full snapshot', (done) => {
-    setMaxSegmentSize(10)
+    setMaxSegmentSize(0)
     setupBuilder.build()
 
     waitRequestSendCalls(1, (calls) => {
--- a/packages/rum-recorder/src/boot/recorder.spec.ts
+++ b/packages/rum-recorder/src/boot/recorder.spec.ts
@@ -4,6 +4,7 @@ import { inflate } from 'pako'
 
 import { setup, TestSetupBuilder } from '../../../rum-core/test/specHelper'
 import { collectAsyncCalls } from '../../test/utils'
+import { setMaxSegmentSize } from '../domain/segmentCollection'
 
 import { Segment, RecordType } from '../types'
 import { startRecording } from './recorder'
@@ -51,6 +52,7 @@ describe('startRecording', () => {
   })
 
   afterEach(() => {
+    setMaxSegmentSize()
     setupBuilder.cleanup()
   })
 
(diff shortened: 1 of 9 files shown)


packages/rum-recorder/src/domain/segment.ts
```diff
@@ -20,34 +26,50 @@ export class Segment {
...
       ...this.context,
     }
     this.writer.flush(`],${JSON.stringify(meta).slice(1)}\n`, meta)
   }
 }
+
+function reduceFullSnapshotState(currentState: FullSnapshotState, record: Record): FullSnapshotState {
+  // Note: to be exploitable by the replay, we have to ensure that FullSnapshot record is
+  // preceded by a Meta and a Focus records.  Because the record logic is emitting both records
+  // synchronously and contiguously, it should always be the case, but check it nonetheless.
```
bcaudan, line 57: I don't get why we need this state logic if `takeFullSnapshot` always trigger these 3 records.
How can it not be the case?
What could be the impact if one is missing?
How does this logic help us in that case?

BenoitZugmeyer: > How can it not be the case?

We have different limits to a segments, like a size limit. In theory, we could chose to flush a segment after a Meta or Flush record, so the trio Meta-Flush-FullSnapshot would be broken. In practice, it can't happen because the size limit is checked asynchronously and, as you noted, the Meta-Flush-FullSnapshot is sent synchronously. But since the segment collection logic and record emission logic are independent, I chose to ensure this rule so we don't have bad issues when changing one part without the other.

> What could be the impact if one is missing?

The replay logic uses "FullSnapshot" records as a starting state for playing a session, either from the start of the session or when seeking to arbitrary timestamps. But a FullSnapshot record is not enough information and it needs the Meta and Focus too to be effective. When fetching segments, the API only have access to the metadata and fetches segments up until it finds one with `has_full_snapshot` to `true`. If a segment had `has_full_snapshot` to `true` but didn't include a Focus and Meta records, the replay would fail.

packages/rum-recorder/src/boot/recorder.spec.ts
```diff
@@ -159,59 +161,20 @@ describe('startRecording', () => {
...
-  })
 
-  it('set has_focus to true if the document has the focus', () => {
-    hasFocus = true
-    trackFocusRecords(lifeCycle, addRecordSpy)
-    expect((addRecordSpy.calls.mostRecent().args[0] as FocusRecord).data.has_focus).toBe(true)
-  })
+  // eslint-disable-next-line max-len
+  it('does not split Meta, Focus and FullSnapshot records between multiple segments when taking a full snapshot', (done) => {
+    setMaxSegmentSize(10)
```
bcaudan, line 167: idk what is the size for meta, focus and full snapshot so not sure if 10 means something specific.
could it be 0?
naming the 10 value could help understand what we are trying to do here

BenoitZugmeyer: Alright

bcaudan (approved):
LGTM

PR #688, opened 2021-01-20
[RUMF-819] postpone start recording

BenoitZugmeyer:
## Motivation

For dogfooding, we want to be able to conditionally start the recording according to a frontend feature flag. Currently, calling `DD_RUM.init()` systematically starts the recording.

## Changes

This PR adds an experimental flag to postpone the recording, and (if enabled) exposes a `startSessionRecord` method. This is behind a flag because we are not sure if we want to stabilize this API for now.

## Testing

Unit tests, manual testing

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/master/CONTRIBUTING.md) documentation.

PR #784, opened 2021-04-06
[RUMF-867] enable start/stop recording API

BenoitZugmeyer:
## Motivation

Enable #688 #769 and #771 

## Changes

* Remove exeperimental flag
* Adjust Rum Recorder configuration and API typings

## Testing

CI, manual

---


packages/rum-recorder/src/domain/rrweb/types.ts
```diff
@@ -1,5 +1,5 @@
 import { IdNodeMap, INode, SerializedNodeWithId } from '../rrweb-snapshot/types'
-import type { FocusRecord, RawRecord } from '../../types'
+import { FocusRecord, RawRecord } from '../../types'
```
bcaudan, line 2: Did the compatibility tests break on this one?
if not, should they?

packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
```diff
@@ -72,7 +67,33 @@ export function makeRumRecorderPublicApi(startRumImpl: StartRum, startRecordingI
...
+
+  let startSessionReplayRecordingImpl = () => {
+    onInit = () => startSessionReplayRecordingImpl()
+  }
+
+  let stopSessionReplayRecordingImpl = () => {
+    onInit = noop
+  }
+
+  const rumRecorderPublicApi = {
```
bcaudan, line 89: Should we name this one rumRecorderGlobal to be consistent with rumGlobal on the RUM side?

packages/rum-recorder/src/domain/rrweb/types.ts
BenoitZugmeyer: Yes it has. I'm unsure why it didn't break on master. I am assuming that `domain/rrweb/types.d.ts` was not imported by `index.d.ts`, but now it (indirectly).

packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
BenoitZugmeyer: IMO, we should not. How the returned value will be used is up to the function caller. It will be used as a global and a exported value by `recorder.entry`, but I don't think the name of this internal variable should reflect that. I think it would make sense only if the function was called `makeRumRecorderGlobal`

packages/rum-recorder/src/domain/rrweb/types.ts
bcaudan: oh yeah the addition of exports from `rumRecorderPublicApi` is indeed pulling some extra stuff.
it makes sense, it seems to still be a reasonable way to perform this compatibility check then.

packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
bcaudan: So it is maybe `rumGlobal` in `makeRumPublicApi` that we should rename `rumPublicApi` then, wdyt?

bcaudan (approved):

packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
BenoitZugmeyer: Agreed, I renamed a few variables to `rumPublicApi` / `rumRecorderPublicApi` for consistency

PR #787, opened 2021-04-07
🐛 [RUMF-882] Segments can be flushed because of the max_size limit even if the max_size isn't reached 

BenoitZugmeyer:
## Motivation

When a lot of small records are emited over time, segments can be prematurely flushed by the max_size logic.

This happens because the size limit is checked asynchronously, when the worker responds to the main thread. If multiple responses are over the size limit, and records are added in between responses, a segment can be flushed even if it didn’t reach the max size itself.

## Changes

* Adjust segment collection communication with the deflate worker to be able to access the related segment in response callbacks
* Make sure the related segment hasn't been flushed before creating a new one when the size reach its limit

## Testing

Unit + Manual

---


packages/rum-recorder/src/domain/deflateSegmentWriter.ts
```diff
@@ -1,54 +1,44 @@
...
-          if (pendingMeta.id !== data.id) {
-            let lostCount = 0
-            while (pendingMeta.id !== data.id) {
-              pendingMeta = this.pendingMeta.shift()!
-              lostCount += 1
-            }
-            addMonitoringMessage(`${lostCount} deflate worker responses have been lost`)
+  constructor(private worker: DeflateWorker, onWrote: (size: number) => void, onFlushed: (data: Uint8Array) => void) {
+    const listener: DeflateWorkerListener = monitor(({ data }) => {
+      if (!('error' in data)) {
```
vlad-mh, line 12: nit: maybe we do an early return here?
```if (error in data) { return; }```

**Review by vlad-mh — APPROVED** [review]
yes this makes sense.

could you give us some before/after comparison on pages we know Datadog was having the issue?

File: packages/rum-recorder/src/domain/deflateSegmentWriter.ts [hunk 2]
```diff
@@ -1,54 +1,44 @@
...
+        if (data.id > this.id) {
+          // Messages should be received in the same order as they are sent, so if we receive a
+          // message with an id superior to this DeflateSegmentWriter instance id, we know that
```
bcaudan, line 15: I am not sure to see what are the pros/cons of having separate writer instances compared to keep track of sent messages in the writer to prevent flushing.
Any insight?

BenoitZugmeyer:
> could you give us some before/after comparison on pages we know Datadog was having the issue?

In `/monitors/manage`:

Before: Screenshot 2021-04-09 at 16 07 43

After:  Screenshot 2021-04-09 at 16 08 09

bcaudan (approved):
Looking good!

vlad-mh (approved):

Issue #794, opened 2021-04-15
Illegal invocation on transport file


felipeplets:
Since April 9th I started seeing errors that I can't reproduce reaching my users and it seems be more frequently.

I'm using @datadog/browser-rum 2.7.1 which depends on @datadog/browser-core 2.7.1

The raw error is: 
```bash
TypeError: Illegal invocation
  at e.send(../../../src/transport/transport.ts:22:34)
  at e.flush(../../../src/transport/transport.ts:68:20)
  at apply(../../../src/transport/transport.ts:153:12)
  at n(../../src/helpers.ts:87:17)
```

Line 22 on the source map is: 
```typescript
    HttpRequest.prototype.send = function (data, size) {
        var url = this.withBatchTime ? addBatchTime(this.endpointUrl) : this.endpointUrl; // <== this one
        if (navigator.sendBeacon && size < this.bytesLimit) {
            var isQueued = navigator.sendBeacon(url, data);
            if (isQueued) {
                return;
            }
```
On file `@datadog/browser-core/esm/transport/transport.js`.

But this does not seem to make sense as the error is `e.send` so it is happening when send is called. 

It seems to happen in any browser (Chrome 70/79/80/89, Mobile Safari, Samsung Internet, Chrome Mobile are some examples). I could not reproduce it yet.

Any thoughts or ideas are welcome.

bcaudan:
Hello @felipeplets,

From the source map, since it references ts files, it should probably map to the original sources [transport.ts@22](https://github.com/DataDog/browser-sdk/blob/v2.7.1/packages/core/src/transport/transport.ts#L22):
```ts
const isQueued = navigator.sendBeacon(url, data)
```                                                
which would make more sense.

From the [spec](https://w3c.github.io/beacon/#sec-processing-model), sendBeacon can send TypeError if there is an issue during url parsing.
Not sure what could go wrong here.

Did you upgrade the SDK version on April 9th?
If yes, which one were you using before?

If not, I would suggest you to go through [datadog support](https://www.datadoghq.com/support/) in order to exchange a bit more info about your application to help us diagnose the issue.

felipeplets:
@bcaudan we have not updated it, we do daily deploys but DD was not updated for a couple of months. I've updated it this week to the latest version in order to try fixing the issue.

bcaudan:
It could also worth checking if `navigator.sendBeacon` is not patched by another dependency. 
[...]

PR #796, opened 2021-04-15
🐛 fallback to xhr when sendBeacon throws

bcaudan:
## Motivation

From the spec, sendBeacon can throw TypeError.
cf #794 

## Changes

Fallback to xhr when sendBeacon throws

## Testing

unit

---

### DIFF
--- a/packages/core/src/transport/transport.ts
+++ b/packages/core/src/transport/transport.ts
@@ -1,7 +1,7 @@
 import { Context } from '../tools/context'
 import { getTimeStamp, relativeNow } from '../tools/timeUtils'
 import { addEventListener, DOM_EVENT, jsonStringify, noop, objectValues } from '../tools/utils'
-import { monitor } from '../domain/internalMonitoring'
+import { monitor, addErrorToMonitoringBatch } from '../domain/internalMonitoring'
 
 // https://en.wikipedia.org/wiki/UTF-8
 const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/
@@ -20,9 +20,13 @@ export class HttpRequest {
   send(data: string | FormData, size: number) {
     const url = this.withBatchTime ? addBatchTime(this.endpointUrl) : this.endpointUrl
     if (navigator.sendBeacon && size < this.bytesLimit) {
-      const isQueued = navigator.sendBeacon(url, data)
-      if (isQueued) {
-        return
+      try {
+        const isQueued = navigator.sendBeacon(url, data)
+        if (isQueued) {
+          return
+        }
+      } catch (e) {
+        addErrorToMonitoringBatch(e)
       }
     }
     const request = new XMLHttpRequest()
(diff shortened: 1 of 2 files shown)


felipeplets (approved):
LGTM 🚀

packages/core/src/transport/transport.ts
```diff
@@ -20,9 +20,13 @@ export class HttpRequest {
...
-      const isQueued = navigator.sendBeacon(url, data)
-      if (isQueued) {
-        return
+      try {
+        const isQueued = navigator.sendBeacon(url, data)
+        if (isQueued) {
+          return
+        }
+      } catch (e) {
+        addErrorToMonitoringBatch(e)
```
BenoitZugmeyer, line 29: When `sendBeacon` keeps failing, we might end-up trying to send the error in a loop. This might not be a big issue since it would be throttled by the batch though (and anyway, it wont impact customers since they don't use the internal monitoring).

BenoitZugmeyer (approved):

packages/core/src/transport/transport.ts
bcaudan: Indeed! I'll update that to send it only once by SDK execution

felipeplets (approved):

packages/core/src/transport/transport.ts
```diff
@@ -37,6 +41,14 @@ function addBatchTime(url: string) {
   )}`
 }
 
+let hasReportedBeaconError = false
+function reportBeaconError(e: unknown) {
```
BenoitZugmeyer, line 45: Sorry to be nitpicking, but IMO if we want to handle this behavior, we may want to have a more global solution: a monitor introduced in [your previous PR](https://github.com/DataDog/browser-sdk/pull/795) is also subject to this, since it can be triggered while reporting a monitored error. WDYT? Maybe having a monitor limit or avoiding sending similar errors multiple times could be better.

bcaudan: No worries!
I am not sure to follow you in:
> a monitor introduced in your previous PR is also subject to this, since it can be triggered while reporting a monitored error.

Could you precise your thoughts here?

BenoitZugmeyer: Imagine we try to send a monitor log. The message is added to the monitoring `Batch`. Then, when it flushes itself via the [`flushPeriodically` callback](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/transport/transport.ts#L154-L157), if something goes wrong and trigger an unexpected error (`this.flush()` throws somehow), this error will be added to the monitoring `Batch`. And then we can loop: if the same thing goes wrong, we keep re-adding it. Maybe `this.flush()` fails before doing the request, so the impact is limited. But maybe it fails after doing the request, so we could potentially report issue indefinitly.

bcaudan: as seen together IRL, that seems a pretty low level risk, let's go with this for now.

PR #801, opened 2021-04-20
[RUMF-895] cleanup the global "mirror" variable

BenoitZugmeyer:
## Motivation

This is a preliminary step for [RUMF-895], to reduce the complexity of the mutations processing before improving its performance impact.

## Changes

Cleanup the seemingly unused `mirror` global variable. This PR introduce a few utility functions that mimic the `mirror.*` features, trying to stay functionally identical, except for the `mirror.has` logic, where the use of a global "map" has been replaced by testing the existence of a serialized node attached to the DOM Node. FMU, the only implication it could have is that the Node id is not reused when removing then inserting again a DOM Node in the document in two different mutation batches. It should not have any impact on the player logic, but we should try a bit more on staging to ensure this.

## Testing

CI, manual

---

[RUMF-895]: https://datadoghq.atlassian.net/browse/RUMF-895


packages/rum-recorder/src/domain/rrweb/mutation.ts
```diff
@@ -300,19 +296,24 @@ export class MutationObserverWrapper {
...
       texts: this.texts
         .map((text) => ({
-          id: mirror.getId(text.node as INode),
+          id: getSerializedNodeId(text.node),
           value: text.value,
         }))
-        // text mutation's id was not in the mirror map means the target node has been removed
-        .filter((text) => mirror.has(text.id)),
+        // ignore mutations whose target node has been removed
+        .filter((text) => text.id !== -1),
```
bcaudan, line 310: could it be interesting to introduce a helper for all the `-1` checks?

packages/rum-recorder/src/domain/rrweb/mutation.ts
```diff
@@ -300,19 +296,24 @@ export class MutationObserverWrapper {
...
     }
+
+    while (this.mapRemoves.length) {
+      recursivelyRemoveSerializedNodes(this.mapRemoves.shift()!)
+    }
```
bcaudan, line 315: For my understanding, why do we need to remove serialized nodes from removed nodes?

packages/rum-recorder/src/domain/rrweb/mutation.spec.ts
```diff
@@ -86,10 +86,9 @@ describe('MutationObserverWrapper', () => {
   }
 })
 
-function addNodeToMap(node: Node, map: IdNodeMap) {
+function addNodeToMap(node: Node) {
```
bcaudan, line 89: we should probably rename this function then

packages/rum-recorder/src/domain/rrweb/record.ts
```diff
@@ -33,7 +33,7 @@ export function record(options: RecordOptions = {}): RecordAPI {
 
   const takeFullSnapshot = (isCheckout = false) => {
     const wasFrozen = mutationController.isFrozen()
-    mutationController.freeze() // don't allow any mirror modifications during snapshotting
+    mutationController.freeze() // don't allow any node to be serialized or mutation to be emitted during sharpshooting
```
bcaudan, line 36: Could it be interesting to add some details about why freeze is needed?

bcaudan, line 36:
```suggestion
    mutationController.freeze() // don't allow any node to be serialized or mutation to be emitted during snapshooting
```

packages/rum-recorder/src/domain/rrweb-snapshot/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { forEach } from '../rrweb/utils'
+import { SerializedNodeWithId } from './types'
+
+interface NodeWithSerializedNode extends Node {
+  __sn: SerializedNodeWithId
```
bcaudan, line 5: 🙇

packages/rum-recorder/src/domain/rrweb-snapshot/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+export function getSerializedNodeId(n: Node) {
+  // if n doesn't have a serialized node, use -1 as its id.
+  if (!hasSerializedNode(n)) {
+    return -1
+  }
+  return n.__sn.id // eslint-disable-line no-underscore-dangle
+}
+
+export function setSerializedNode(n: Node, serializeNode: SerializedNodeWithId) {
+  // eslint-disable-next-line no-underscore-dangle
```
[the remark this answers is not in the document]
BenoitZugmeyer: I agree, let's disable the rule

packages/rum-recorder/src/domain/rrweb/record.ts
BenoitZugmeyer: Damn autocorrect :)

packages/rum-recorder/src/domain/rrweb/mutation.ts
BenoitZugmeyer: Like discussed, I'd like to iterate on the concept of returning `-1` ids. I think it's fine to keep it like this for now as it will be revisited later.

packages/rum-recorder/src/domain/rrweb/record.ts
bcaudan: I'd love to have a sharpshooting concept though 😄

vlad-mh (approved):
So indeed, looking at this diff & from my understanding of rrweb, it looks like the "mirror" only makes the code more complicated with little benefits.

give we address Bastien's points, this lgtm.

Have we done a bit of archeology to understand _why_ the mirror was introduced? did it serve a purpose in the past but no more? was it code in prevision of something that didn't happen or hasn't happen yet?

packages/rum-recorder/src/domain/rrweb/record.ts
BenoitZugmeyer: As I don't think 'freeze' is needed in the first place, I can't add details justifying its usage. Let's remove it in a future PR.

packages/rum-recorder/src/domain/rrweb-snapshot/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { forEach } from '../rrweb/utils'
+import { SerializedNodeWithId } from './types'
+
+interface NodeWithSerializedNode extends Node {
+  __sn: SerializedNodeWithId
+}
+
```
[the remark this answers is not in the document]
BenoitZugmeyer: IMO it is fine to keep it named "rrweb-snapshot/utils", as it is exactly that: snapshot utilities. But I don't mind rename it if you think it is more explicit to have `serialization` in the name.

BenoitZugmeyer:
> Have we done a bit of archeology to understand why the mirror was introduced? did it serve a purpose in the past but no more? was it code in prevision of something that didn't happen or hasn't happen yet?

I dig a bit, and here are my observations:

* `mirror.getId` has always been a standalone function to get the attached serialized node id. It doesn't use `mirror.map`. This function has been moved/renamed to `getSerializedNodeId` without much change.

* `mirror.getNode` has never been used in the recorder, only the player

* `mirror.removeNodeFromMap` has been introduced [very early](https://github.com/rrweb-io/rrweb/commit/46ab1d6e8057056bd043b0041a4a270716847160), but it was seemingly unused then.

* `mirror.has` has been introduced in https://github.com/rrweb-io/rrweb/commit/8acf670a99f56eb1708368481723c82284019000 and https://github.com/rrweb-io/rrweb/commit/cff1617959b41aceca7b4b5f84a5fedbb2dfb623 to filter out text data and attributes mutations occuring on removed nodes. By removing the attached serialized node instead, we are providing the same functionality.

packages/rum-recorder/src/domain/rrweb-snapshot/utils.ts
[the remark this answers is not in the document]
bcaudan: as the file grow and it seems to be only serialization logic, I'd be in favor of being more explicit to not put other unrelated code with it.

bcaudan (approved):
LGTM, I let you decide for utils renaming

PR #727, opened 2021-02-09
[RUMF-848] Expose internal types and constants to be used in Datadog app

BenoitZugmeyer:
## Motivation

Expose internal types to be used in Datadog app

## Testing

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/master/CONTRIBUTING.md) documentation.

PR #802, opened 2021-04-22
🔥 remove mutation freeze logic

BenoitZugmeyer:
## Motivation

This logic was introduced to allow RRWeb users to "pause" the recording
by using the public method `freezePage`. This method has been removed in
PR #727, and it appears that the `freeze` logic is now unneeded: since
the "fullsnapshot" logic is synchronous, no DOM mutation can happen or
be processed while the document snapshot is taken.

## Changes

Remove mutation 'freeze' logic. Because mutations still need to be flushed before a "fullsnapshot" is taken, the `MutationController` have been kept, and `flush` / `onFlush` methods have been introduced.

## Testing

CI, manual

---


bcaudan (approved):
Nice 🔥

PR #809, opened 2021-04-27
⚗[RUMF-889] use preferred clock

bcaudan:
## Motivation

Since we observed issues with monotonic clock values (mainly on chrome), experiment with using system clock instead.

## Changes

Prefer system clock behind a `system-clock` FF (RUM + logs).
Introduce a preferred time strategy to consistently use relative time or timestamp depending on the FF value.

Choices:
- replace relative time by timestamp to compute duration
- keep both clock internally to use one or the other in different cases
- for event dates, use corrected timestamp from relative time when we only have a relative time

Make tests compatible with both modes and run automated tests with and without the system clock preferred.

## Testing
Automated + Manual
- [x] check overall behavior disabled
- [x] check overall behavior enabled
- [x] check behavior when triggering a drift

---


BenoitZugmeyer (approved):
Appart from a small suggestion, LGTM! Great work

PR #810, opened 2021-04-27
 ⚗ [RUMF-902] implement a new mutation observer 

BenoitZugmeyer:
## Motivation

The current mutation observer logic is complicated and inefficient. Let's implement a new mutation observer.

## Changes

* Introduce a few utility functions that were partially reverted from #805. This makes getting serialized node ids from DOM node a bit safer thanks to distinct TS types.
* Implement a new mutation observer
* Use the new mutation observer behind a FF, and adjust existing tests

## Testing

Manual, unit/e2e

### Before:

Screenshot 2021-04-28 at 14 36 39

SDK time: about 1900ms

### After:

Screenshot 2021-04-28 at 14 21 54

SDK time: about 560ms

---


packages/rum-recorder/src/domain/rrweb/mutationObserver.spec.ts
```diff
@@ -0,0 +1,635 @@
...
+            node: expectNewNode({ type: NodeType.Element, tagName: 'a' }),
+          },
+        ],
+      })
+    })
+
+    describe('does not emit mutations', () => {
+      it('when adding an ignored node', () => {
+        ignoredElement.remove()
+        snapshot(document)[0]!
```
bcaudan, line 490:
```suggestion
        snapshot(document)
```
or am I missing something?
(there are other instances)

packages/rum-recorder/src/domain/rrweb/mutationObserver.spec.ts
```diff
@@ -0,0 +1,635 @@
...
+        adds: [
+          {
+            parent: expectInitialNode({ idAttribute: 'sandbox' }),
+            node: expectNewNode({ type: NodeType.Element, tagName: 'a' }),
+          },
+        ],
+      })
+    })
+
+    describe('does not emit mutations', () => {
```
bcaudan, line 487:
```suggestion
    describe('does not emit mutations occurring in ignored node', () => {
```
to be clearer and consistent with the approach for hidden node

packages/rum-recorder/src/domain/rrweb/record.spec.ts
```diff
@@ -52,7 +54,7 @@ describe('record', () => {
     expect(records.filter((record) => record.type === RecordType.FullSnapshot).length).toEqual(1)
   })
 
-  it('is safe to checkout during async callbacks', (done) => {
+  it('is safe to checkout during async callbacks (old mutation observer)', (done) => {
```
bcaudan, line 57: does checkout mean snapshot?
there is a isCheckout parameter for takeFullSnapshot but is it used?

packages/rum-recorder/src/domain/rrweb/record.spec.ts
```diff
@@ -123,6 +125,72 @@ describe('record', () => {
     })
   })
 
+  it('is safe to checkout during async callbacks (new mutation observer)', (done) => {
```
bcaudan, line 128: I know this is an adaptation of the old test however I am not sure to understand how the expectations relate to this test case 😕

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -341,7 +341,57 @@ describe('recorder', () => {
...
+          document.body.removeChild(span)
+          p.appendChild(span)
+          p.removeChild(span)
+          div.appendChild(span)
+        })
+
+        await flushEvents()
+
+        const { validate, expectInitialNode } = createMutationPayloadValidatorFromSegment(
+          events.sessionReplay[0].segment.data
```
bcaudan, line 368:
```suggestion
          getFirstSegment(events)
```

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -417,6 +467,56 @@ describe('recorder', () => {
...
+          const div = document.createElement('div')
+          const span = document.querySelector('span')!
+          document.body.appendChild(div)
+          div.appendChild(span)
+        })
+
+        await flushEvents()
+
+        const { validate, expectInitialNode, expectNewNode } = createMutationPayloadValidatorFromSegment(
+          events.sessionReplay[0].segment.data
```
bcaudan, line 490:
```suggestion
          getFirstSegment(events)
```

packages/rum-recorder/src/domain/rrweb/mutationObserver.spec.ts
BenoitZugmeyer: Not missing something, this is bad copy/pasta 😅

packages/rum-recorder/src/domain/rrweb/record.spec.ts
BenoitZugmeyer: Mmh the 'checkout' concept have been moved outside of our RRWeb codebase. It was a way to take full snapshots at regular intervals or when a given number of records have been emitted.

In the [upstream RRWeb test](https://github.com/rrweb-io/rrweb/blob/master/test/record.test.ts#L168) they specify a "checkout" every 2 records, so it occured exactly after `span.innerText = 'test'`. Since we removed this concept from the RRWeb codebase, I used a `takeFullSnapshot` call to emulate the behavior.

I think the test is here because the RRWeb author feared that mutation occured while taking a full snapshot (hence the "freeze" concept removed in https://github.com/DataDog/browser-sdk/pull/802).

Because we know that taking a full snapshot is synchronous, "async callbacks" will never be a problem. Also, we'll have a much better code coverage of mutations, so we may not need this test anymore. I'd be in favor of removing it. WDYT?

The `isCheckout` parameter is not used indeed, and can be removed.

packages/rum-recorder/src/domain/rrweb/record.spec.ts
BenoitZugmeyer: (see previous comment explanation)

packages/rum-recorder/src/domain/rrweb/utils.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    d = document.createElement('d')
+    parent = document.createElement('parent')
+    parent.appendChild(a)
+    a.appendChild(aa)
+    parent.appendChild(b)
+    parent.appendChild(c)
+    parent.appendChild(d)
+  })
+
+  it('sorts siblings in reverse order', () => {
```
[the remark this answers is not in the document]
BenoitZugmeyer: Yes, this is explained [here](https://github.com/DataDog/browser-sdk/blob/38635839746d62301795b9f174e307f3a4e030fc/packages/rum-recorder/src/domain/rrweb/mutationObserver.ts#L137-L140)

packages/rum-recorder/test/utils.ts
```diff
@@ -362,7 +364,11 @@ export function createMutationPayloadValidator(initialDocument: SerializedNodeWi
       }
 
       if ('childNodes' in node) {
-        return new ExpectedNode({ ...node, childNodes: [] })
+        const nodeCopy = { ...node, childNodes: [] }
+        if (!nodeCopy.isSVG) {
+          delete nodeCopy.isSVG
+        }
```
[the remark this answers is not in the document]
BenoitZugmeyer: [We remove any `undefined` value in the "actual" payload](https://github.com/DataDog/browser-sdk/blob/38635839746d62301795b9f174e307f3a4e030fc/packages/rum-recorder/test/utils.ts#L314-L319). When using nodes serialized in the initial document, some values may be `undefined`, and breaks our comparison. I'll try to make this clearer.

packages/rum-recorder/src/domain/rrweb-snapshot/serializationUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+export function getSerializedNodeId(n: Node) {
+  return hasSerializedNode(n) ? n.__sn.id : undefined
+}
+
+export function setSerializedNode(n: Node, serializeNode: SerializedNodeWithId) {
+  ;(n as Partial<NodeWithSerializedNode>).__sn = serializeNode
+}
+
+export function nodeIsIgnored(n: Node): boolean {
+  return getSerializedNodeId(n) === IGNORED_NODE_ID
```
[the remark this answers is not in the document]
BenoitZugmeyer: All nodes passed to this function should have a serialized node in the first place... I'll make it clearer.

packages/rum-recorder/src/domain/rrweb/mutationObserver.spec.ts
```diff
@@ -0,0 +1,635 @@
...
+        removes: [
+          {
+            parent: expectInitialNode({ idAttribute: 'sandbox' }),
+            node: expectInitialNode({ tag: 'a' }),
+          },
+        ],
+      })
+    })
+
+    it('uses the initial parent id when removing a node from multiple places', () => {
```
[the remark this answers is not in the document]
bcaudan: That is clearer!

packages/rum-recorder/src/domain/rrweb/utils.spec.ts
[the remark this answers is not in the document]
bcaudan: I am a bit bothered that this function looks like it is generic but encapsulate some mutation processing logic.
What about keep this function in mutationObserver and rename it sortAddedAndMovedNodes?

packages/rum-recorder/src/domain/rrweb/record.spec.ts
bcaudan: 🔥 🔥 🔥

BenoitZugmeyer: `isCheckout` argument removed in https://github.com/DataDog/browser-sdk/pull/820

bcaudan (approved):
Great work 🎉

packages/rum-recorder/src/domain/rrweb/mutationObserver.ts
```diff
@@ -0,0 +1,274 @@
...
+    return null
+  }
+}
+
+function processCharacterDataMutations(mutations: Array<WithSerializedTarget<CharacterDataMutationRecord>>) {
+  const textMutations: TextMutation[] = []
+
+  // Deduplicate mutations based on their target node
+  const handledNodes = new Set<Node>()
+  const filteredMutations = mutations.filter((mutation) => {
```
vlad-mh, line 192: can we live with just 1 loop here? 🤔 

```
const textMutations: TextMutation[] = []
const handledNodes = new Set<Node>()

for (n in mutations) {
   if (handledNodes.has(mutation.target)) {
       continue;
    }
    handledNodes.add(mutation.target)
    const value = mutation.target.textContent
    if (value === mutation.oldValue) {
        continue
    }
    textMutations.push({
        id: getSerializedNodeId(mutation.target),
        value,
    })

```

vlad-mh (approved):
nice, super exciting perf wins! let's test it out
some nits, but mostly ideas for further improvements

packages/rum-recorder/src/domain/rrweb/mutationObserver.ts
```diff
@@ -0,0 +1,274 @@
...
+  // Discard any mutation with a 'target' node that:
+  // * isn't injected in the current document or isn't known/serialized yet: those nodes are likely
+  // part of a mutation occurring in a parent Node
+  // * should be hidden or ignored
+  const filteredMutations = mutations.filter(
+    (mutation): mutation is WithSerializedTarget<MutationRecord> =>
+      document.contains(mutation.target) &&
+      hasSerializedNode(mutation.target) &&
+      !nodeOrAncestorsIsIgnored(mutation.target) &&
+      !nodeOrAncestorsShouldBeHidden(mutation.target)
```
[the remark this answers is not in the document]
BenoitZugmeyer: Yes I agree, this is still something I have in mind: optimizing this loop or using the 'cache' optimization could still be usefull here. In practice, I observed that this whole filter loop takes less than 1ms even when hundreds of mutations occurs.

`document.contains` is **fast**. It is not O(N), and the browser probably uses internal tricks to avoid walking on all parents. See for yourself, run this in a busy page:

```js
{ const list = document.querySelectorAll('*'); const start = performance.now(); list.forEach(e => { while (e) { e = e.parentNode };  }); console.log(performance.now() - start) }

{ const list = document.querySelectorAll('*'); const start = performance.now(); list.forEach(e => document.contains(e)); console.log(performance.now() - start) }
```

packages/rum-recorder/src/domain/rrweb/mutationObserver.ts
```diff
@@ -0,0 +1,274 @@
...
+  const removedNodes = new Map<Node, NodeWithSerializedNode>()
+  for (const mutation of mutations) {
+    forEach(mutation.addedNodes, (node) => {
+      addedAndMovedNodes.add(node)
+    })
+    forEach(mutation.removedNodes, (node) => {
+      if (!addedAndMovedNodes.has(node)) {
+        removedNodes.set(node, mutation.target)
+      }
+      addedAndMovedNodes.delete(node)
```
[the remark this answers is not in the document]
BenoitZugmeyer: A node cannot be both removed and added in a single mutation. Then, the process order doen't matter.

packages/rum-recorder/src/domain/rrweb/mutationObserver.ts
BenoitZugmeyer: This was my first version for both `processCharacterDataMutations` and `processAttributesMutations`. But, for attributes, the loop got a bit hairy... so I split the loop into "deduplication" and "emission". To have a consistent pattern between the two functions, I applied the same structure to character data mutations. I find it a bit easier to read, and it shouldn't matter performance wise.

PR #814, opened 2021-04-29
🐛[RUMF-900] prevent events to be sent from expired session

bcaudan:
## Motivation

We have noticed that a tab in an expired session state can send events if another tab as an active session.

## Changes

Prevent to send event when current session id is not the same that the view context session id.

## Testing

unit + manual

---

### DIFF
--- a/packages/rum-recorder/src/domain/segmentCollection.ts
+++ b/packages/rum-recorder/src/domain/segmentCollection.ts
@@ -189,7 +189,7 @@ export function computeSegmentContext(applicationId: string, session: RumSession
     return undefined
   }
   const viewContext = parentContexts.findView()
-  if (!viewContext?.session.id) {
+  if (!viewContext?.session.id || viewContext.session.id !== session.getId()) {
     return undefined
   }
   return {
--- a/packages/rum-core/src/domain/assembly.spec.ts
+++ b/packages/rum-core/src/domain/assembly.spec.ts
@@ -327,7 +327,7 @@ describe('rum assembly', () => {
       expect(serverRumEvents.length).toBe(0)
     })
 
-    it('when view context has session id, it should generate event', () => {
+    it('when view context has current session id, it should generate event', () => {
       viewSessionId = '1234'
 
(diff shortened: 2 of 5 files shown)


packages/rum-recorder/src/domain/segmentCollection.ts
```diff
@@ -189,7 +189,7 @@ export function computeSegmentContext(applicationId: string, session: RumSession
     return undefined
   }
   const viewContext = parentContexts.findView()
-  if (!viewContext?.session.id) {
+  if (!viewContext?.session.id || viewContext.session.id !== session.getId()) {
```
BenoitZugmeyer, line 192:
```suggestion
  if (!viewContext || viewContext.session.id !== session.getId()) {
```
Could be a bit simpler

BenoitZugmeyer (approved):

packages/rum-recorder/src/domain/segmentCollection.ts
bcaudan: Since SegmentContext require `session.id` to be `string`, we still need to check that session id is defined to make the compiler happy.

BenoitZugmeyer: Alright. The typechecking is less safe in `assembly` then, since we don't need this check. Probably masked by the `as RumEvent` cast.

bcaudan: We could enforce ViewContext session id to be defined, it should not hurt

BenoitZugmeyer: FMU currently a `viewContext` *may*  have an undefined session id, for example if the session expire and the URL changes without any user interaction. It is not an issue because we correctly check if the session is tracked before assembling the event, but it may require a bit more work that expected.

bcaudan: ok, it could worth to improve that at some point

PR #815, opened 2021-04-29
Track foreground (focus)

ThibautGeriz:
## Motivation

Add focus info into RUM

Today, it's not possible to know understand the user journey with the RUM data if the user open multiple tab in the session. This PR track if the user focus the document in order to offer a better view of the user journey within the target websites

## Changes

- [x] add `view.in_foreground` to error event
- [x] add `view.in_foreground` to action event
- [x] add `view.in_foreground_periods` to view event
- [x] Hide everything behind a feature flag

_I am not adding `view.in_foreground` to long tasks & resources & view because, it's not yet needed by replay._

## Testing

Unit test are good and I can see the data in the GUI

image

[GUI](https://dd.datad0g.com/rum/explorer?event=AQAAAXlHZfQRzirjoQAAAABBWGxIWmZQdE5EV1N2TTFOazBWUTVB&from_ts=1620399663537&live=true&query=%40usr.email%3Athibaut.gery%40datadoghq.com%20%40browser.name%3AChrome&tab=view&to_ts=1620400563537&p_tab=attributes)

---


packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -64,6 +64,9 @@ function processError(error: RawError) {
       stack: error.stack,
       type: error.type,
     },
+    focus: {
+      start_focused: document.hasFocus(),
```
bcaudan, line 68: For network errors, it means `end_focused` not `start_focused`

packages/rum-core/src/domain/rumEventsCollection/view/trackViewFocus.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    if (lastIndex >= 0 && viewFocus.focusedTimes[lastIndex].currently_focused) {
+      viewFocus.focusedTimes[lastIndex] = {
+        start: toServerRelativeTime(viewFocus.focusedTimes[lastIndex].start),
+        duration: toServerDuration(elapsed(viewFocus.focusedTimes[lastIndex].start, relativeNow())),
+      }
+      scheduleViewUpdate()
+    }
+  }
+}
```
ThibautGeriz, line 71: I did not do any defensive programming.
I always expect that there is one `currently_focused=true` and it's the last item of the array. Let me know if you want to add more safe guard here.

packages/rum-core/src/domain/rumEventsCollection/view/trackViewFocus.ts
```diff
@@ -0,0 +1,86 @@
...
+    })
+    scheduleViewUpdate()
+  }
+
+  function closeLastFocusTime() {
+    const lastIndex = viewFocus.inForegroundPeriods.length - 1
+
+    if (lastIndex >= 0 && viewFocus.inForegroundPeriods[lastIndex].currentlyFocused) {
+      viewFocus.inForegroundPeriods[lastIndex] = {
+        start: toServerRelativeTime(viewFocus.inForegroundPeriods[lastIndex].start),
```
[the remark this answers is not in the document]
BenoitZugmeyer: When removing the `toServerRelativeTime`, tests still passes. We probably need a test on this.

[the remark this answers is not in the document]
ThibautGeriz: I refactored this whole thing to rely on `PreferredTime` I misunderstood what `RelativeTime` was.

packages/core/src/domain/automaticErrorCollection.ts
```diff
@@ -97,10 +107,15 @@ export function trackNetworkError(configuration: Configuration, errorObservable:
         source: ErrorSource.NETWORK,
         stack: truncateResponse(request.response, configuration) || 'Failed to load',
         startClocks: request.startClocks,
+        inForeground: request.inForeground as boolean | undefined,
       })
     }
   }
 
+  function addFocusToContext(context: XhrStartContext | FetchStartContext) {
```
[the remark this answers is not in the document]
ThibautGeriz: If I do this, I should rename every `focus*` to `foreground*` What do you think?

packages/core/src/tools/timeUtils.ts
```diff
@@ -47,7 +47,16 @@ export function toServerDuration(duration: Duration | undefined) {
   if (!isNumber(duration)) {
     return duration
   }
-  return round(duration * 1e6, 0) as ServerDuration
+  return toNs(duration) as ServerDuration
+}
+
+export function toServerRelativeTime(relativeTime: RelativeTime): RelativeTime
```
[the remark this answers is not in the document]
ThibautGeriz: I ended up removing that.

packages/core/src/domain/automaticErrorCollection.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Not necessarily, for stuff that only deal with `document.hasFocus` or `focus` / `blur` events, using `focus` seems fine. But other than that, I would agree that using `foreground` would be better.

packages/core/src/tools/timeUtils.ts
```diff
@@ -47,7 +47,11 @@ export function toServerDuration(duration: Duration | undefined) {
   if (!isNumber(duration)) {
     return duration
   }
-  return round(duration * 1e6, 0) as ServerDuration
+  return toNs(duration) as ServerDuration
+}
+
+function toNs(time: number): number {
+  return round(time * 1e6, 0)
```
BenoitZugmeyer, line 54: This change could be reverted

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
```diff
@@ -67,7 +54,7 @@ describe('rum track view metrics', () => {
       .withFakeLocation('/foo')
       .beforeBuild(({ location, lifeCycle }) => {
         lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, handler)
-        return trackViews(location, lifeCycle)
+        return trackViews(location, lifeCycle, configuration as Configuration)
```
BenoitZugmeyer, line 57: The `.beforeBuild(({ configuration })` argument should be used instead of an external configuration constant. You can use `.withConfiguration({ isEnabled: () => false })`

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -26,19 +26,7 @@ const FAKE_NAVIGATION_ENTRY: RumPerformanceNavigationTiming = {
...
-    return handler.calls.argsFor(index)[0] as ViewEvent
-  }
-
-  function getHandledCount() {
-    return handler.calls.count()
-  }
-
-  return { handler, getViewEvent, getHandledCount }
-}
+const configuration: Partial<Configuration> = { isEnabled: () => true }
```
BenoitZugmeyer, line 29: Same here, the `configuration` provided by the "setup builder" should be used instead

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -160,20 +168,34 @@ function newView(
...
     lifeCycle,
     scheduleViewUpdate,
     loadingType
   )
 
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  let stopViewFocusTracking: () => void = noop
+  let viewFocus: Partial<ViewEvent> = {}
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  let updateCurrentFocusDuration: (endTime: PreferredTime) => void = noop
```
BenoitZugmeyer, line 181: No need for eslint-disable comments anymore

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -1,10 +1,11 @@
-import { Duration, RelativeTime, ServerDuration, TimeStamp } from '@datadog/browser-core'
+import { Duration, RelativeTime, ServerDuration, TimeStamp, Configuration } from '@datadog/browser-core'
 import { setup, TestSetupBuilder } from '../../../../test/specHelper'
 import { RumEventType, ViewLoadingType } from '../../../rawRumEvent.types'
 import { LifeCycleEventType } from '../../lifeCycle'
 import { ViewEvent } from './trackViews'
 import { startViewCollection } from './viewCollection'
 
+const configuration: Partial<Configuration> = { isEnabled: () => true }
```
BenoitZugmeyer, line 8: Same here, the configuration provided by the "setup builder" should be used instead

packages/rum-core/src/domain/rumEventsCollection/view/trackViewFocus.ts
```diff
@@ -0,0 +1,91 @@
...
+  preferredNow,
+  addEventListener,
+  DOM_EVENT,
+  Duration,
+  clocksNow,
+  preferredClock,
+  PreferredTime,
+} from '@datadog/browser-core'
+
+const MAX_NUMBER_OF_FOCUSED_TIME = 500
```
bcaudan, line 13: Could you add a comment to explain why 500?
"arbitrary value" is an acceptable explanation

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -34,6 +39,8 @@ export interface ViewEvent {
   loadingType: ViewLoadingType
   cumulativeLayoutShift?: number
   hasReplay: boolean
+  inForeground?: boolean
```
bcaudan, line 42: Do we need the foreground info on view event?

packages/rum-core/src/domain/rumEventsCollection/view/trackViewFocus.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+function trackFocus(onFocusChange: () => void) {
+  return addEventListener(window, DOM_EVENT.FOCUS, () => onFocusChange())
+}
+
+function trackBlur(onBlurChange: () => void) {
+  return addEventListener(window, DOM_EVENT.BLUR, () => onBlurChange())
+}
```
bcaudan, line 91: You should wrap the listener in our `monitor` utility in order to catch potential execution errors

packages/core/src/domain/automaticErrorCollection.ts
```diff
@@ -78,8 +83,13 @@ export function stopRuntimeErrorTracking() {
 }
 
 export function trackNetworkError(configuration: Configuration, errorObservable: ErrorObservable) {
-  startXhrProxy().onRequestComplete((context) => handleCompleteRequest(RequestType.XHR, context))
-  startFetchProxy().onRequestComplete((context) => handleCompleteRequest(RequestType.FETCH, context))
+  const xhrProxy = startXhrProxy()
+  xhrProxy.beforeSend(addFocusToContext)
+  xhrProxy.onRequestComplete((context) => handleCompleteRequest(RequestType.XHR, context))
```
bcaudan, line 88: I am not really fond of scattering the foreground logic everywhere.
Could we centralized the collection of foreground period and request what we need when building the event?
something like:
```
Foreground
  getPeriods(start, end)
  getStatus(time)

From view collection:
  view.foreground_periods = Foreground.getPeriods(view.start, view.end).map(toServerDuration)

From error/action collection:
  view.in_foreground = Foreground.getStatus(event.start)
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViewFocus.ts
BenoitZugmeyer: `addEventListener` does that for you

bcaudan: 🤦‍♂️ argh my bad, too transparent

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
ThibautGeriz: No it's only used internally, might be able to remove it if you feel like it's misleading.
It's not exposed in `viewCollections`.

bcaudan: if we don't need it, let's not add it

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -40,6 +40,9 @@ export interface PerformanceResourceDetailsElement {
 export interface RawRumErrorEvent {
```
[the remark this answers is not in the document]
ThibautGeriz: :ah: I did it  manually, did not know there was a command. Will run the command

packages/core/src/domain/automaticErrorCollection.ts
ThibautGeriz: Feels weird, if we were to follow that logic, we would need a URL manager used in view, resources & network error.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
ThibautGeriz: Actually removed it as it makes no sense

PR #816, opened 2021-04-29
🐛 [RUMF-900] clear parent view context when view end

bcaudan:
## Motivation

When a view is ended when changing page, some events can still be tracked and attached to the view after its end, which seems quite surprising.

## Changes

Clear current view context on view end

## Testing

unit + manual

---

### DIFF
--- a/packages/rum-core/src/domain/parentContexts.ts
+++ b/packages/rum-core/src/domain/parentContexts.ts
@@ -37,13 +37,6 @@ export function startParentContexts(lifeCycle: LifeCycle, session: RumSession): 
   let previousActions: Array<PreviousContext<ActionContext>> = []
 
   lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (currentContext) => {
-    if (currentView) {
-      previousViews.unshift({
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/domain/parentContexts.ts
```diff
@@ -37,25 +37,29 @@ export function startParentContexts(lifeCycle: LifeCycle, session: RumSession):
...
     // A view can be updated after its end.  We have to ensure that the view being updated is the
     // most recently created.
-    if (currentView!.id === currentContext.id) {
+    if (currentView && currentView.id === currentContext.id) {
       currentView = currentContext
     }
   })
 
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_ENDED, ({ id, endClocks }) => {
+    if (currentView && currentView.id === id) {
```
BenoitZugmeyer, line 53: Do we need to check the view id? If so, I think we should do the same in the recorder.

bcaudan: indeed, we should not

BenoitZugmeyer (approved):

PR #828, opened 2021-05-03
🐛 [RUMF-870] Max errors threshold should not take into account errors excluded by beforeSend

BenoitZugmeyer:
## Motivation

The user is able to discard errors (logs or rum) via the `beforeSend` callback. Those errors should not be taken into account for the error limitation process, because it could prevent some more legitimate errors from being sent.

## Changes

* Implement a generic 'error filter' utility to apply the error threshold on Logs and RUM codebases
* Use the 'error filter' in Logs and RUM
* Remove the previous error limitation logic

## Testing

Unit, manual

---


packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -43,6 +44,12 @@ export function doStartErrorCollection(lifeCycle: LifeCycle, observable: Observa
   }
 }
 
+export function createRumErrorFilter(lifeCycle: LifeCycle, configuration: Configuration) {
+  return createErrorFilter(configuration, (error) => {
+    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processError(error))
+  })
+}
```
bcaudan, line 51: From a design perspective, I think it could be more interesting to:
- create the error filter in assembly, since it is not really related to error collection
- introduce a RAW_ERROR_COLLECTED lifeCycle event to allow to generate errors from other places

bcaudan: hum, lifeCycle events seems quite consistent with what we are doing, so not sure why we should do something different here or what is the issue with introducing a new lifeCycle event.
I am also not a big fan of introducing an error collection dependency in assembly.

packages/core/src/tools/errorFilter.ts
```diff
@@ -0,0 +1,41 @@
+import { Configuration } from '../domain/configuration'
+import { ErrorSource, RawError } from './error'
+import { clocksNow } from './timeUtils'
+import { ONE_MINUTE } from './utils'
+
+export type ErrorFilter = ReturnType<typeof createErrorFilter>
+
+export function createErrorFilter(configuration: Configuration, onLimitReached: (limitError: RawError) => void) {
+  let errorCount = 0
+  let isSendingLimitError = false
```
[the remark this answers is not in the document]
BenoitZugmeyer: This variable allows to send errors only while the `onLimitReached` callback is invoked, so it will already forbid subsequent errors from being sent if the customers filters the 'limit error'. I can rename it though, `allowNextError` seems fine if we actually reset it to `false` after using it.

Having a test based on the error message would be a bit more complicated because the message is not stored at the same place in Logs and RUM events.

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
BenoitZugmeyer: > what is the issue with introducing a new lifeCycle event

LifeCycle events are global, so any new LifeCycle event is complexifying the overall design of RUM. Having only a few important events makes things easier to explain and could be represented as a graph if we want to. We chose to have a few more events to decouple modules allowing to write tests without having to mock them. Here, we don't need to listen to a potential `RAW_ERROR_COLLECTED` event in tests, so IMO introducing a new lifecycle event for this is not necessary.

>  I am also not a big fan of introducing an error collection dependency in assembly.

Here, the only requirement is to transform the 'raw error' into an 'raw rum error event'. It could be moved to another shared module.

Anyway, it's not a big deal, I'll add the RAW_ERROR_COLLECTED event.

bcaudan: it could worth to talk it through to be aligned on that though

packages/logs/src/boot/logs.ts
```diff
@@ -53,10 +55,11 @@ export function doStartLogs(
     combine({ session_id: session.getId() }, getGlobalContext(), getRUMInternalContext())
   )
 
-  const assemble = buildAssemble(session, configuration)
+  const errorFilter = createErrorFilter(configuration, reportError)
+  const assemble = buildAssemble(session, configuration, errorFilter)
```
bcaudan, line 59: nit: I would even put createErrorFilter inside buildAssemble but no strong argument for it

packages/core/src/tools/errorFilter.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  beforeEach(() => {
+    ;({ clock, stop: cleanupClock } = mockClock())
+    resetNavigationStart()
+  })
+
+  afterEach(() => {
+    cleanupClock()
+  })
+
+  it('allows to send an error', () => {
```
bcaudan, line 24: What about removing send occurrences?

packages/core/src/tools/errorFilter.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  it('allows to send errors again once one minute have passed', () => {
+    errorFilter = createErrorFilter(CONFIGURATION, noop)
+
+    errorFilter.isLimitReached()
+    errorFilter.isLimitReached()
+    clock.tick(ONE_MINUTE)
+    expect(errorFilter.isLimitReached()).toBe(false)
+  })
+
+  it('calls the "onLimitReached" callback with the raw "limit" error to send when the limit is reached', () => {
```
bcaudan, line 46:
```suggestion
  it('calls the "onLimitReached" callback with the raw "limit reached" error to send when the limit is reached', () => {
```
and other occurrences to increase naming consistency

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -26,18 +26,26 @@ export function startErrorCollection(lifeCycle: LifeCycle, configuration: Config
 }
 
 export function doStartErrorCollection(lifeCycle: LifeCycle, observable: Observable<RawError>) {
-  observable.subscribe((error) => lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processError(error)))
+  observable.subscribe((error) => lifeCycle.notify(LifeCycleEventType.RAW_ERROR_COLLECTED, { error }))
```
bcaudan, line 29: nit: by wiring the observable to the raw error event, we could now: 
- remove the observable from doStartErrorCollection signature
- wire the automaticErrorCollection to lifecycle in parent function

packages/logs/src/boot/logs.spec.ts
```diff
@@ -292,4 +298,73 @@ describe('logs', () => {
...
+    afterEach(() => {
+      cleanupClock()
+    })
+
+    it('stops sending error logs when reaching the limit', () => {
+      const sendLogSpy = jasmine.createSpy<(message: LogsMessage & { foo?: string }) => void>()
+      const sendLog = startLogs({ errorLogger: new Logger(sendLogSpy), configuration: { maxErrorsByMinute: 1 } })
+      sendLog({ message: 'foo', status: StatusType.error }, {})
+      sendLog({ message: 'bar', status: StatusType.error }, {})
+      sendLog({ message: 'baz', status: StatusType.error }, {})
```
bcaudan, line 319:
```suggestion
```
any reason to have another filtered log?

packages/core/src/tools/errorFilter.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    errorFilter.isLimitReached()
+    errorFilter.isLimitReached()
+    expect(onLimitReachedSpy).toHaveBeenCalledOnceWith({
+      message: 'Reached max number of errors by minute: 1',
+      source: 'agent',
+      startClocks: relativeToClocks(0 as RelativeTime),
+    })
+  })
+
+  it('returns false while calling the "onLimitReached" to allow sending the "limit reached" error', () => {
```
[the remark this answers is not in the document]
BenoitZugmeyer: FWIW, I think `shouldSendError` was a better name. Sure, `isLimitReached` reflects the internal state of the filter a bit better, but it is not always right (since when the callback is called, the limit *is* reached but `isLimitReached` returns `false`). `shouldSendError` is more abstract and related to the end-usage of the filter, but it is more accurate.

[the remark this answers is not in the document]
bcaudan: IMO, we could have an alternate naming that:
- is not tied to the limit
- is not tied to the end usage 
something like `errorFilter.shouldAllowError()`

bcaudan (approved):
Great!

packages/core/src/tools/errorFilter.spec.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Let's roll with `isLimitReached`, it's explicit enough.

PR #829, opened 2021-05-04
👷 add woke check

bcaudan:
## Motivation

Prevent non inclusive terms to be used in the code base.

## Changes

- Add a `yarn woke` check 
- Add a gitlab job to perform this check 
- Fix current violations (except the one from pako)

## Testing

local + ci

---


scripts/cli
```diff
@@ -58,4 +58,9 @@ cmd_version () {
   git add yarn.lock
 }
 
+cmd_woke () {
+  command -v woke || yarn fail 'woke not installed, see https://github.com/get-woke/woke#installation'
```
BenoitZugmeyer, line 62: Calling `yarn fail` from a bash script seems overkill

```suggestion
  if ! command -v woke; then
    echo 'woke not installed, see https://github.com/get-woke/woke#installation'
    exit 1
  fi
```

bcaudan: the fail script is exactly what I want to do though 😕 
would you prefer to call the bash script directly? (without yarn)

BenoitZugmeyer: What's wrong with my suggestion?

bcaudan: that should work but we loose a bit of ["display sugar"](https://github.com/DataDog/browser-sdk/blob/master/scripts/fail.sh) and duplicate an existing behavior

BenoitZugmeyer: what about replacing the `fail.sh` script with a `cmd_fail () {}` command in `cli` ?

bcaudan: lgtm
and remove `yarn fail` then?

BenoitZugmeyer: Yes

BenoitZugmeyer: In the future, I'd be in favor of moving complex 'script' like `test:compat:*` in `cli`, so we can rename `cmd_fail` to a simple function `fail`. I think it would be less weird :)

BenoitZugmeyer (approved):
Sweet!

PR #837, opened 2021-05-07
✅ add `yarn test:unit:watch` script

BenoitZugmeyer:
## Motivation

Small trick that I always use when writing unit tests

## Changes

Add a `yarn test:unit:watch` script

## Testing

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "release": "lerna version --exact",
     "version": "scripts/cli version",
     "publish:npm": "TARGET_DATACENTER=us BUILD_MODE=release yarn build && lerna publish from-package",
-    "test": "yarn test:unit",
+    "test": "yarn test:unit:watch",
     "test:unit": "karma start ./test/unit/karma.local.conf.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
     "publish:npm": "TARGET_DATACENTER=us BUILD_MODE=release yarn build && lerna publish from-package",
     "test": "yarn test:unit",
     "test:unit": "karma start ./test/unit/karma.local.conf.js",
+    "test:unit:watch": "yarn test:unit --no-single-run",
```
bcaudan, line 23: I also wanted to open a PR for something similar from some time.
What would you think of having this behavior for `yarn test` directly?

BenoitZugmeyer: It is somewhat unusual to have this as the default test command, so as a newcomer I wouldn't expect this behavior. But if it is discoverable as a separate test as well, I wouldn't mind :)

bcaudan: So, keep `test:unit:watch` and use it from `test`?

bcaudan (approved):
Sweet!

Issue #774, opened 2021-03-29
Clear a previously set user object


igrayson:
When a user logs out (and becomes an anonymous user), we'd like RUM data to reflect this.

`datadogRum.setUser()` currently does nothing when passed `null`, keeping any previously set value.

BenoitZugmeyer:
Hello @igrayson. You can reset individual user properties by setting them to `undefined`. Would it be fine for you? Exemple: `datadogRum.setUser({ id: undefined, email: undefined })`

igrayson:
Thanks! That might have the outcome we're after (need to read through the code), but it sounds like that would be accidental and likely to break in the future (say, after more attributes are added to the user).

Maybe a better outcome would be a change the datadog RUM SDK API to capture our actual intention: not to clear out specific user attributes, but clear out the entire user.

amortemousque:
This has been released in version 2.9.0.
You can now remove a previously set user like so `datadogRum.removeUser()`

PR #840, opened 2021-05-11
✨ Clear previously set user context

amortemousque:
## Motivation

Provide an API to clear a previously set user without having to clear all its attributes.
#774

## Changes

Allows you to clear a  previously set user with the API `removeUser()`

## Testing

Unit, manual

---

### DIFF
--- a/packages/rum-core/src/boot/rumPublicApi.spec.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.spec.ts
@@ -381,6 +381,18 @@ describe('rum public api', () => {
       expect(errorSpy).not.toHaveBeenCalled()
     })
 
+    it('should remove the user', () => {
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -381,6 +381,16 @@ describe('rum public api', () => {
       expect(errorSpy).not.toHaveBeenCalled()
     })
 
+    it('should remove the user', () => {
+      rumPublicApi.removeUser()
```
bcaudan, line 385: We should probably add a user first, otherwise this test would pass even if we don't call `removeUser`

BenoitZugmeyer (approved):

PR #842, opened 2021-05-11
⚡️ [RUMF-902] enable new mutation observer

BenoitZugmeyer:
## Motivation

Enable the mutation observer introduced in #810 

## Changes

* Remove the experimental flag usages
* Remove now unused related code
* Move the `MutationController` into `mutationObserver`

## Note

While working on this, I started thinking about a more in-depth cleaning in two ways:

* redesign the 'rrweb-snapshot' logic. I did not forget about [this comment](https://github.com/DataDog/browser-sdk/pull/810#discussion_r622359608), and I think we can drastically simplify the design of the whole file, for example to avoid serializing ignored nodes.

* move modules around. I'd like to:
  * move `rrweb-snapshot` (serialization logic) and the `privacy` module closer to the rest of the 'rrweb' codebase
  * rename `domain/rrweb` to `domain/record`
  * rename `domain/{segment,segmentCollection,deflateWorker}` to `domain/segmentCollection/*`

This will be done later, in separate PRs though.

## Testing

Unit, E2E

---


packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -399,7 +398,7 @@ export function serializeNodeWithId(
     return null
   }
   map[id] = true
-  let recordChild = !skipChild
+  let recordChild = true
   if (serializedNode.type === NodeType.Element) {
     recordChild = recordChild && !serializedNode.shouldBeHidden
```
bcaudan, line 398: `recordChild &&` should not be needed anymore then

packages/rum-recorder/src/domain/rrweb-snapshot/snapshot.ts
```diff
@@ -426,13 +426,10 @@ export function serializeNodeWithId(
...
-export function snapshot(n: Document): [SerializedNodeWithId | null, IdNodeMap] {
-  const idNodeMap: IdNodeMap = {}
-  return [
-    serializeNodeWithId(n, {
-      doc: n,
-      map: idNodeMap,
-    }),
-    idNodeMap,
-  ]
+export function serializeDocument(n: Document): SerializedNodeWithId {
```
bcaudan, line 424: should we rename this file then or is snapshot still relevant?
or do you plan to address that with later refactorings?

BenoitZugmeyer: Yes I plan to address it later, because the whole `rrweb-snapshot` is not really relevant anymore either.

bcaudan (approved):
LGTM 🔥

vlad-mh (approved):
makes sense to me, thanks for the cleanup Benoit!
the follow up plan seem good as well, I'll be happy to see those PRs

Issue #622, opened 2020-11-16
Handle logs via console AND http simultaneously


trestletech:
I'd like to have log entries routed both to the console AND submitted via HTTP.

https://docs.datadoghq.com/logs/log_collection/javascript/#change-the-destination documents how to handle the logs either via HTTP, console, or silent. But it doesn't appear that there's a way to have logs both printed to the console and reported to DataDog.

I have two  use cases where I'd like this:

1. In a staging environment, it'd be nice to allow our QA folks to be able to pull up the JS Console and take screenshots that show any logs that we emitted alongside whatever they're seeing in the browser.
2. We have a "debug mode" that we can enable in production. With our previous logger, when you switched in to debug mode the logs would be emitted to the console so that you could inspect them. I'd like to be able to do that while still having the logs forwarded to DataDog.

yaplusplus:
I have the exact same case

brunor97:
It's hard to believe that this is not already implemented. It seems like such a basic feature to have...

bcaudan:
👋 Sorry that we don't make progress on some of the requested improvements, unfortunately this feature is not on our top priorities for now 😕 
Don't hesitate to drop a PR if you have a chance, we'd be happy to review it and add it to our feature set.

brunor97:
Thanks for the answer! Sorry if that sounded a bit harsh, I understand that priorities have to be managed. If I have the time I might look into it :)

amortemousque:
This has been released in version 2.9.0.
You can now set a table of handler like so `datadogLogs.logger.setHandler(['http', 'console'])`

PR #844, opened 2021-05-12
[RUMF-910] ⚡️ handle logs console and http simultaneously

amortemousque:
## Motivation

Handle logs via console and http simultaneously
#622

## Changes

- The logger now accepts multiple handlers `logger.setHandler(['console','http'])`
- Create custom logger with mutliple handlers `LOGS.createLogger('foo', {
          handler: ['console','http']
        }`

## Testing

Unit, manual

---


packages/logs/src/domain/logger.spec.ts
```diff
@@ -94,6 +94,16 @@ describe('Logger', () => {
       })
     })
 
+    it('should be configurable to "console" and "http"', () => {
+      logger.setHandler(['console', 'http'])
```
BenoitZugmeyer, line 98: While using values directly is legitimate, we tend to prefer using the constant `HandlerType` internally instead

packages/logs/src/domain/logger.spec.ts
```diff
@@ -94,6 +94,16 @@ describe('Logger', () => {
       })
     })
 
+    it('should be configurable to "console" and "http"', () => {
```
bcaudan, line 97: I'd be in favor of adding a test for silent + console, just to document the expected behavior

BenoitZugmeyer (approved):

bcaudan (approved):
👍

PR #846, opened 2021-05-14
 🚚 move recorder modules

BenoitZugmeyer:
## Motivation

As a follow up of #842 , I wanted to move "recorder" source files a bit to have a clearer module hierarchy.

## Changes

Proposed new hierarchy (not showing spec files):
```
packages/rum-recorder/src
├── boot
│   ├── recorder.entry.ts
│   ├── recorder.ts
│   └── rumRecorderPublicApi.ts
├── domain
│   ├── record
│   │   ├── index.ts
│   │   ├── mutationBatch.ts
│   │   ├── mutationObserver.ts
│   │   ├── observer.ts
│   │   ├── privacy.ts
│   │   ├── record.ts
│   │   ├── serializationUtils.ts
│   │   ├── serialize.ts
│   │   ├── types.ts
│   │   └── utils.ts
│   └── segmentCollection
│       ├── deflateWorker.d.ts
│       ├── deflateWorker.js
│       ├── index.ts
│       ├── segment.ts
│       └── segmentCollection.ts
├── transport
│   └── send.ts
├── constants.ts
├── index.ts
├── internal.ts
└── types.ts
```

## Testing

CI

---


test/e2e/scenario/recorder.scenario.ts
```diff
@@ -63,7 +62,7 @@ describe('recorder', () => {
       )
     })
 
-  describe('snapshot', () => {
+  describe('full snapshot', () => {
```
bcaudan, line 65: is full snapshot equivalent to document serialization?
if yes, what about renaming these references as well?

BenoitZugmeyer: A full snapshot is a FullSnapshot record, which contain the serialized document and the document scroll offset. I think we should keep the same naming to keep consistency between player and recorder.

bcaudan: I think we should not hesitate to tweak the naming on player and recorder side if it makes more sense to us.
You have a better grasp on that so I'll let you decide what is making the more sense.

bcaudan (approved):

test/e2e/scenario/recorder.scenario.ts
BenoitZugmeyer: I would agree that we could improve the protocol namings, but I would not do this incrementally by renaming only FullSnapshot and keeping IncrementalSnapshot for example. I think it would be worth to discuss more about this with an RFC when we feel the need for it.

PR #847, opened 2021-05-14
✨[RUMF-907] Use unaltered console functions when displaying console messages

bcaudan:
## Motivation

Avoid to generate logs / events when we want to display something to the user

## Changes

- Keep early references on console functions
- Use them when we want to display messages to the user

## Testing

ci, manual

---


packages/core/src/tools/display.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Keep references on console methods to avoid triggering patched behaviors
+ */
+export const display: Pick<typeof console, 'log' | 'warn' | 'error'> = {
+  log: console.log,
+  warn: console.warn,
+  error: console.error,
```
BenoitZugmeyer, line 7: I don't think this is enough to ensure that we are using unpatched functions in this kind of setup:

```
    <script src="http://localhost:8080/datadog-rum-recorder.js"></script>
    <script>
      DD_RUM.init({
        clientToken: 'xxx',
        applicationId: 'xxx',
      })
    </script>
    <script src="http://localhost:8080/datadog-logs.js"></script>
    <script>
      DD_LOGS.init({
        clientToken: 'xxx',
      })
    </script>
```

BenoitZugmeyer (commented):
I like the approach! We should probably enable the ESLint rule `no-console` on most of the codebase then :)

packages/core/src/tools/display.ts
bcaudan: Agree, I am not sure that this is an issue though.
Wort case scenario, a rum displayed error generate a logs error or vice versa, so we should not end up in infinite loop.
wdyt?

BenoitZugmeyer: Ok, could you adjust the comment to make it clear that it can happen then?

BenoitZugmeyer (approved):

PR #850, opened 2021-05-17
⚗ [RUMF-878] add startView API

bcaudan:
## Motivation

First step in the manual view naming strategy

## Changes

behind `view-renaming` flag, introduce new API:

```ts
DD_RUM.startView(name?: string)
```

Allowing to manually start a new view with an optional name

## Testing

unit / manual

---

### DIFF
--- a/packages/logs/src/boot/logs.entry.ts
+++ b/packages/logs/src/boot/logs.entry.ts
@@ -38,9 +38,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
 
   const beforeInitSendLog = new BoundedBuffer()
   let sendLogStrategy = (message: LogsMessage, currentContext: Context) => {
-    beforeInitSendLog.add<[LogsMessage, Context]>([message, currentContext], ([message, context]) =>
-      sendLogStrategy(message, context)
-    )
+    beforeInitSendLog.add(() => sendLogStrategy(message, currentContext))
   }
   const logger = new Logger(sendLog)
 
--- a/packages/logs/src/boot/logs.entry.ts
+++ b/packages/logs/src/boot/logs.entry.ts
@@ -36,9 +36,11 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
   const globalContextManager = createContextManager()
   const customLoggers: { [name: string]: Logger | undefined } = {}
 
-  const beforeInitSendLog = new BoundedBuffer<[LogsMessage, Context]>()
+  const beforeInitSendLog = new BoundedBuffer()
   let sendLogStrategy = (message: LogsMessage, currentContext: Context) => {
-    beforeInitSendLog.add([message, currentContext])
+    beforeInitSendLog.add<[LogsMessage, Context]>([message, currentContext], ([message, context]) =>
+      sendLogStrategy(message, context)
+    )
   }
   const logger = new Logger(sendLog)
 
@@ -56,7 +58,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
       }
 
       sendLogStrategy = startLogsImpl(userConfiguration, logger, globalContextManager.get)
-      beforeInitSendLog.drain(([message, context]) => sendLogStrategy(message, context))
+      beforeInitSendLog.drain()
 
       isAlreadyInitialized = true
     }),
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
@@ -125,7 +125,7 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
       currentView.triggerUpdate()
     },
     startView: (name?: string, startClocks?: ClocksState) => {
-      currentView.end()
+      currentView.end(startClocks)
       currentView.triggerUpdate()
       currentView = trackViewChange(startClocks, name)
     },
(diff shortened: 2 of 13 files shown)


packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -115,15 +115,20 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
...
+    return newView(lifeCycle, location, isRecording, ViewLoadingType.ROUTE_CHANGE, currentView.url, startClocks, name)
   }
 
   return {
     addTiming: (name: string, endClocks = clocksNow()) => {
       currentView.addTiming(name, endClocks)
       currentView.triggerUpdate()
     },
+    startView: (name?: string, startClocks?: ClocksState) => {
+      currentView.end()
```
BenoitZugmeyer, line 128: The `end()` function will make the current view ends "now", but the newly started view could have a `startsClock` in the past, so the two views may overlap. I think we should use the `startsClock` value to set the end/duration of the ending view.

BenoitZugmeyer (commented):
This seems like a good start. Two notes:

* I think we should ensure that the initial document resource is attached to the initial view instead of the latest created view, since now views can be created before `init()`

* I'm mitigated about the function name. I kind of liked `newView`. What made you change your mind from the RFC proposal?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -115,15 +115,20 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
...
+  function trackViewChange(startClocks?: ClocksState, name?: string) {
+    return newView(lifeCycle, location, isRecording, ViewLoadingType.ROUTE_CHANGE, currentView.url, startClocks, name)
   }
 
   return {
     addTiming: (name: string, endClocks = clocksNow()) => {
       currentView.addTiming(name, endClocks)
       currentView.triggerUpdate()
     },
+    startView: (name?: string, startClocks?: ClocksState) => {
```
bcaudan, line 127: > I'm mitigated about the function name. I kind of liked newView. What made you change your mind from the RFC proposal?

A comment on the RFC about the naming.
[Mobile](https://docs.datadoghq.com/real_user_monitoring/android/advanced_configuration/#custom-views) already have a `startView` method on their side and it could be more natural if we need to introduce a `stopView` at some point.


cc @BenoitZugmeyer

BenoitZugmeyer: Ok it makes sense! ✅

test/e2e/lib/framework/createTest.ts
```diff
@@ -96,6 +102,11 @@ class TestBuilder {
       declareTest(this.title, setups[0].factory(setupOptions), runner)
     }
   }
+
+  private rumInit: (options: RumSetupOptions) => void = (options) => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    ;(window.DD_RUM as any).init(options)
```
BenoitZugmeyer, line 108: It seems that `test/e2e/lib/types/global.ts` is not used.  To avoid some type casts, what about importing it somewhere in E2E tests? Else, if we don't want to define those types globally, we could pass a correctly typed DD_RUM argument to `rumInit` (and remove `global.ts`):

```ts
private rumInit: (rumGlobal: RumGlobal, options: RumSetupOptions) => void = (options) => {
   rumGlobal.init(options)
}
```

test/e2e/scenario/rum/init.scenario.ts
```diff
@@ -0,0 +1,60 @@
...
+      setTimeout(() => {
+        ;(window.DD_RUM as any).addError('after manual view')
+        ;(window.DD_RUM as any).addAction('after manual view')
+        ;(window.DD_RUM as any).addTiming('after manual view')
+      }, 20)
+
+      setTimeout(() => (window.DD_RUM as any).init(options), 30)
+    })
+    .withSetup(bundleSetup, 'bundle')
+    .withSetup(asyncSetup, 'async')
```
BenoitZugmeyer, line 25: It could be nice to have somithing working with all setups. In the NPM setup, what about having something like `window.RUM_INIT = ${options.rumInit.toString()}`, that would be called from the app if it exists?

packages/logs/src/boot/logs.entry.ts
```diff
@@ -36,9 +36,11 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
...
   const customLoggers: { [name: string]: Logger | undefined } = {}
 
-  const beforeInitSendLog = new BoundedBuffer<[LogsMessage, Context]>()
+  const beforeInitSendLog = new BoundedBuffer()
   let sendLogStrategy = (message: LogsMessage, currentContext: Context) => {
-    beforeInitSendLog.add([message, currentContext])
+    beforeInitSendLog.add<[LogsMessage, Context]>([message, currentContext], ([message, context]) =>
+      sendLogStrategy(message, context)
+    )
   }
```
BenoitZugmeyer, line 44: Storing arguments in `BoundedBuffer` seems to be avoidable. What about simply passing a closure?

```ts
  let sendLogStrategy = (message: LogsMessage, currentContext: Context) => {
    beforeInitSendLog.add(() => sendLogStrategy(message, context))
  }
```

bcaudan: thanks, way simpler indeed!

test/e2e/lib/framework/createTest.ts
bcaudan: the issue was more on duplicated type between options and configuration, situation should be better now

BenoitZugmeyer (approved):
Nice !

PR #851, opened 2021-05-17
♻️ cleanup recorder DOM serialization

BenoitZugmeyer:
## Motivation

Follow up of #842 : 

This PR tries to improve the code quality and logic of the recorder DOM serialization.

## Changes

* A few shared functions have been moved into `serializationUtils`
* A few functions have been renamed to have a clearer name
* A few functions have been improved
  * to make use of the URL parsing utilities from core
  * simplify regular expressions and related
  * split big functions to smaller parts
* We don't serialize ignored nodes anymore
* Add more tests on the main serialization functions

## Testing

CI, manual

---


packages/rum-recorder/src/domain/record/serialize.spec.ts
```diff
@@ -21,10 +21,10 @@ describe('serializeNodeWithId', () => {
       expect(map).toEqual({})
     })
 
-    it('sets ignored serialized node id to IGNORED_NODE_ID', () => {
+    it('does not serialize ignored nodes', () => {
```
bcaudan, line 275: Was there a benefit for a feature we are not using?
do we have a test case covering the reordering of nodes with one ignored?
[node, ignored, node] -> [node, node, ignore]

packages/rum-recorder/src/domain/record/mutationObserver.ts
```diff
@@ -152,16 +151,16 @@ function processChildListMutations(mutations: Array<WithSerializedTarget<ChildLi
   sortAddedAndMovedNodes(sortedAddedAndMovedNodes)
 
   // Then, we iterate over our sorted node sets to emit mutations. We collect the newly serialized
-  // node ids in a map to be able to skip subsequent related mutations.
-  const serializedNodesIdMap: IdNodeMap = {}
+  // node ids in a set to be able to skip subsequent related mutations.
+  const serializedNodeIds = new Set<number>()
```
bcaudan, line 155: indeed 👍

packages/rum-recorder/src/domain/record/serializationUtils.ts
```diff
@@ -20,19 +30,46 @@ export function setSerializedNode(n: Node, serializeNode: SerializedNodeWithId)
...
+
+const SRCSET_URLS = /(^\s*|,\s*)([^\s,]+)/g
+export function makeSrcsetUrlsAbsolute(attributeValue: string, baseUrl: string) {
+  return attributeValue.replace(
+    SRCSET_URLS,
+    (_, prefix: string, url: string) => `${prefix}${makeUrlAbsolute(url, baseUrl)}`
+  )
+}
+
+export function makeUrlAbsolute(url: string, baseUrl: string): string {
```
bcaudan, line 73:
```suggestion
function makeUrlAbsolute(url: string, baseUrl: string): string {
```

packages/rum-recorder/src/domain/record/serializationUtils.ts
```diff
@@ -38,101 +39,45 @@ export function nodeOrAncestorsIsIgnored(n: NodeWithSerializedNode) {
...
   return value
 }
 
 const URL_IN_CSS_REF = /url\((?:(')([^']*)'|(")([^"]*)"|([^)]*))\)/gm
-const RELATIVE_PATH = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/
-const DATA_URI = /^(data:)([^,]*),(.*)/i
-export function makeStylesheetUrlsAbsolute(cssText: string | null, href: string): string {
-  return (cssText || '').replace(
+const ABSOLUTE_URL = /^[A-Za-z]:|^\/\//
+const DATA_URI = /^data:.*,/i
```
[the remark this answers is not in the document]
BenoitZugmeyer: For DATA_URI: the new regex in functionaly the same as the old one (`/^(data:)([^,]*),(.*)/i`): it matches a string starting with `data:`, with a comma later on. The capturing parenthesis are not used.

For ABSOLUTE_URL: the old regex to match strings starting with `www.`, `http://` and associated, single letter drive Windows absolute path (ex: `C:\foo`), and protocol-less URLs `//`.

* The first case (`www.`) was incorrect: URLs starting with `www.` are not absolute:
```js
new URL('www.foo.com', 'http://example.org').href
// "http://example.org/www.foo.com"
```
* By testing, I figured out that any URL starting with characters followed by a semicolon was considered as an absolute URL, so `http:` and `C:` checks can be factorized.
* The last case, protocol-less URLs, has been kept.

Please note that those regexes are not really necessary since `makeUrlAbsolute` handles absolute URLs and data URLs correctly. I think they could be kept to avoid parsing many absolute or DATA URLs unnecessarily (that could be a performance bottleneck).

packages/rum-recorder/src/domain/record/serialize.spec.ts
BenoitZugmeyer: > Was there a benefit for a feature we are not using?
I don't think there was a real benefit. The mutation processing was more complex and needed to differentiate between ignored nodes and not-yet-serialized nodes, but I think it was an implementation detail more than a feature.

>  do we have a test case covering the reordering of nodes with one ignored?

I think this is covered by other test (in particular, [this one](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-recorder/src/domain/record/mutationObserver.spec.ts#L492)), but it won't hurt to add a new one for this specific case. I'll add it.

packages/rum-recorder/src/domain/record/mutationObserver.ts
```diff
@@ -3,11 +3,11 @@ import { nodeOrAncestorsShouldBeHidden } from './privacy'
...
   getSerializedNodeId,
   hasSerializedNode,
+  nodeAndAncestorsHaveSerializedNode,
   NodeWithSerializedNode,
-  nodeIsIgnored,
-  nodeOrAncestorsIsIgnored,
+  transformAttribute,
 } from './serializationUtils'
-import { transformAttribute, serializeNodeWithId } from './serialize'
+import { serializeNodeWithId } from './serialize'
```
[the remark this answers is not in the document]
BenoitZugmeyer: Responsibilities:

* `serialize` is exporting functions to serialize DOM nodes
* `serializationUtils` is exporting utility functions used in various modules.

If a function is only used by `serialize` it should be in the `serialize` module, else it should be in `serializationUtils`. There is an exception, `setSeliarizedNode`, because it makes sens to collocate it with similar functions.

bcaudan (approved):
LGTM!

packages/rum-recorder/src/domain/record/serializationUtils.spec.ts
```diff
@@ -27,3 +34,110 @@ describe('serialized Node storage in DOM Nodes', () => {
...
+describe('makeUrlAbsolute', () => {
+  it('makes an absolute URL from a relative path', () => {
+    expect(makeUrlAbsolute('bar', 'http://example.org/foo/')).toBe('http://example.org/foo/bar')
+  })
+
+  it('makes an absolute URL from an absolute path', () => {
+    expect(makeUrlAbsolute('/bar', 'http://example.org/foo/')).toBe('http://example.org/bar')
+  })
+
+  it('do not change data URI', () => {
```
vlad-mh, line 140:
```suggestion
  it('does not change data URIs', () => {
```

vlad-mh (approved):
🚢

PR #854, opened 2021-05-18
Track foreground

ThibautGeriz:
## Motivation

Add foreground (focus) info into RUM

Today, it's not possible to understand the user journey with the RUM data if the user open multiple tab in the session. This PR track if the user focus the document in order to offer a better view of the user journey within the target websites.

This is a alternate implementation from https://github.com/DataDog/browser-sdk/pull/815 after comment https://github.com/DataDog/browser-sdk/pull/815#discussion_r631043821

## Changes

- [x] add type
- [x] add module
- [x] implement logic of focus

## Notes

- Still using `RelativeTime` even though `PreferredTime` makes more sense. It was a too big refactoring to do in the PR. If you feels that it needs to be done, I would rather do it into another PR
- Tried to simplify the assembly but typescript won & I cowardly conceded defeat, I can invest more time in it if you feel it's important
- I've added a fourth commit adding a few checks because I had a [view](https://app-491fd269e5f78f447a9037f3b324e786.datadoghq.com/rum/explorer?event=AQAAAXmEo5Ij1bg8SgAAAABBWG1FbzVIeU1UYXN0Z01mc2FYZ3FR&from_ts=1621421066948&live=true&query=%40usr.email%3Athibaut.gery%40datadoghq.com%20%40browser.name%3AChrome%20%40session.has_replay%3Atrue%20%40view.id%3A36936bd0-b817-475d-be01-eff2acf4c053&tab=view&to_ts=1621435466948&p_tab=attributes) with two periods starting at the same moment and one of them was never closed

image

## Testing

I've added unit test

---


packages/rum-core/src/boot/rum.ts
```diff
@@ -90,6 +100,8 @@ export function startRumEventCollection(
       // prevent batch from previous tests to keep running and send unwanted requests
       // could be replaced by stopping all the component when they will all have a stop method
       batch.stop()
+      parentContexts.stop()
```
ThibautGeriz, line 103: Was this missing from the current SDK?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -79,10 +84,22 @@ export function startRumAssembly(
...
+          serverRumEvent = combine(serverRumEvent, inForegroundContext)
+        }
+
+        if (isView(rawRumEvent)) {
+          const in_foreground_periods = foregroundContexts.getInForegroundPeriods(
+            startTime,
+            rawRumEvent.view.time_spent
+          )
+          serverRumEvent = combine(serverRumEvent, { view: { in_foreground_periods } })
+        }
```
ThibautGeriz, line 101: I tried to do
```javascript
serverRumEvent.view.in_foreground_periods = in_foreground_periods
```
I though it was cleaner but the property is written as `readonly`

packages/rum-core/src/boot/rum.ts
bcaudan: it is only used for test, nice catch though 😉

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+    .map((focusPeriod) => {
+      const startTime = eventStartTime > focusPeriod.start ? eventStartTime : focusPeriod.start
+      const startDuration = elapsed(eventStartTime, startTime)
+      const endTime = focusPeriod.end == null || eventEndTime < focusPeriod.end ? eventEndTime : focusPeriod.end
+      const endDuration = elapsed(startTime, endTime)
+      return {
+        start: toServerDuration(startDuration),
+        duration: toServerDuration(endDuration),
+      }
+    })
```
bcaudan, line 122: From a performance perspective, we could probably avoid to look at all periods by having an approach like: 
- start from the most recent period
- while event end <  period start, reject period
- while event start < period  start, take period

wdyt?

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+          it('should false', () => {
+            expect(foregroundContext.getInForeground(2_000 as RelativeTime)).toEqual({ view: { in_foreground: false } })
+          })
+        })
+
+        describe('after 10 second', () => {
+          it('should true', () => {
+            expect(foregroundContext.getInForeground(10_000 as RelativeTime)).toEqual({ view: { in_foreground: true } })
+          })
+        })
```
bcaudan, line 86: IMO, this test structure does not seem to bring much more value than having a single test with all the expectations.
describe and it text are duplicates of expectations values, except for the last one but a comment could do the job.

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+      describe('getInForeground', () => {
+        describe('after 2 second', () => {
+          it('should false', () => {
+            expect(foregroundContext.getInForeground(2_000 as RelativeTime)).toEqual({ view: { in_foreground: false } })
+          })
+        })
+
+        describe('after 10 second', () => {
+          it('should true', () => {
+            expect(foregroundContext.getInForeground(10_000 as RelativeTime)).toEqual({ view: { in_foreground: true } })
```
bcaudan, line 84: it could be more helpful to explain what case you are testing with these values, rather than saying that you are testing these values.
what make 2 second and 10 second special?

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,280 @@
+import { RelativeTime, relativeNow, ServerDuration, toServerDuration, Duration } from '@datadog/browser-core'
+import { setup, TestSetupBuilder } from '../../test/specHelper'
+import { createNewEvent } from '../../../core/test/specHelper'
+import { InForegroundPeriod } from '../rawRumEvent.types'
+import { startForegroundContexts, ForegroundContexts } from './foregroundContexts'
+
+const FIVE_SECONDS_MS = 5_000 as Duration
+const TEN_SECONDS_MS = 10_000 as Duration
```
bcaudan, line 8: It could be more valuable to name these variables with their purpose rather than their value.

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,280 @@
+import { RelativeTime, relativeNow, ServerDuration, toServerDuration, Duration } from '@datadog/browser-core'
+import { setup, TestSetupBuilder } from '../../test/specHelper'
+import { createNewEvent } from '../../../core/test/specHelper'
+import { InForegroundPeriod } from '../rawRumEvent.types'
+import { startForegroundContexts, ForegroundContexts } from './foregroundContexts'
+
+const FIVE_SECONDS_MS = 5_000 as Duration
```
bcaudan, line 7: using ms directly could spare you a bunch of `_000`, reducing a bit of noise

packages/rum-core/src/domain/assembly.ts
ThibautGeriz: Not sure what is the take away here.

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,110 @@
...
+  DOM_EVENT,
+  RelativeTime,
+  elapsed,
+  relativeNow,
+  toDuration,
+  toServerDuration,
+} from '@datadog/browser-core'
+import { ForegroundContext, InForegroundPeriod } from '../rawRumEvent.types'
+
+const MAX_NUMBER_OF_FOCUSED_TIME = 500
```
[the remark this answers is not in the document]
ThibautGeriz: Forgot to add the comment back. No particular reason, gotta start somewhere

packages/rum-core/src/domain/assembly.ts
bcaudan: you should add it to `rawRumEvent` instead of `serverRumEvent`

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export interface ForegroundContexts {
+  getInForeground: (startTime: RelativeTime) => ForegroundContext | undefined
+  getInForegroundPeriods: (startTime: RelativeTime, duration: ServerDuration) => InForegroundPeriod[] | undefined
```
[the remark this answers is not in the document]
ThibautGeriz: I though about it. But the `in_foreground_periods` is part of `RawRumViewEvent. I can extract it under its own `ForegroundPeriodsContext` but it feels weird. I will switch to not use the server context datastructure

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+      focusPeriods = []
+      stopFocusTracking()
+      stopBlurTracking()
+    },
+  }
+}
+
+function addNewFocusPeriod() {
+  const now = relativeNow()
+  if (focusPeriods.length > MAX_NUMBER_OF_FOCUSED_TIME) {
```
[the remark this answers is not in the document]
ThibautGeriz: Yes. How?

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+    return
+  }
+  const lastIndex = focusPeriods.length - 1
+  if (lastIndex >= 0 && focusPeriods[lastIndex].end == null) {
+    // already here
+    if (focusPeriods[lastIndex].start === now) {
```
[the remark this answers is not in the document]
ThibautGeriz: Yes 
image
https://app-491fd269e5f78f447a9037f3b324e786.datadoghq.com/rum/explorer?event=AQAAAXmEo5Ij1bg8SgAAAABBWG1FbzVIeU1UYXN0Z01mc2FYZ3FR&from_ts=1621421066948&live=true&query=%40usr.email%3Athibaut.gery%40datadoghq.com%20%40browser.name%3AChrome%20%40session.has_replay%3Atrue%20%40view.id%3A36936bd0-b817-475d-be01-eff2acf4c053&tab=view&to_ts=1621435466948&p_tab=attributes

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+    // close current periods if we miss a blur event
+    const { start } = focusPeriods[lastIndex]
+    focusPeriods[lastIndex] = {
+      start,
+      end: now,
+    }
+  }
+  focusPeriods.push({
+    start: now,
+  })
```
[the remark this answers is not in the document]
ThibautGeriz: I want to say yes because it means we never closed the previous one. And if we do not it. It will continue to increase. I would like to show that's its two different periods. I first wanted to `end: now - DELAY` but it feels artificial. We can also decide to discard to unclosed period because it's not normal.
Let's add some monitoring because it's not normal

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+  return addEventListener(window, DOM_EVENT.FOCUS, () => onFocusChange())
+}
+
+function trackBlur(onBlurChange: () => void) {
+  return addEventListener(window, DOM_EVENT.BLUR, () => onBlurChange())
+}
+
+function getInForeground(startTime: RelativeTime): ForegroundContext {
+  const in_foreground = focusPeriods.some(
+    (focus) => startTime > focus.start && (focus.end == null || startTime < focus.end)
```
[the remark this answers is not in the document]
ThibautGeriz: Not really, I do not have a strong opinion on that. Happy to chat about it.

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+  getInForegroundPeriods: (startTime: RelativeTime, duration: ServerDuration) => InForegroundPeriod[] | undefined
+  stop: () => void
+}
+
+export interface FocusPeriod {
+  start: RelativeTime
+  end?: RelativeTime
+}
+
+let focusPeriods: FocusPeriod[] = []
```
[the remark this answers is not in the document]
ThibautGeriz: Since it's a very different and internal data structure populated by `focus` & `blur` event I would call it like that. I can change

packages/rum-core/src/domain/foregroundContexts.spec.ts
ThibautGeriz: How? Do you have an example somewhere?

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
bcaudan: either from backend side or from the sdk I guess.
From the sdk, you can have a look at [internalMonitoring methods](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/domain/internalMonitoring.ts#L142) to trigger a log or an error.

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
bcaudan: Did you understand what happen?
if yes, it could nice to add a test explaining this case
if no, I'd be in favor of trying to

packages/rum-core/src/domain/foregroundContexts.spec.ts
bcaudan: my point is that your test don't seem to need to manipulate values in seconds.
So you could directly have:
```ts
const FOCUS_PERIOD_LENGTH = 5
clock.tick(1)
foregroundContext.getInForeground(2 as RelativeTime)
```

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
bcaudan: `FocusPeriod` is just a `start` and a `end`, nothing really specific to focus/blur events. 
`FocusPeriod` and `InForegroundPeriod` seems quite close maybe there is something to do here.
I would go with something like:
```ts
let foregroundPeriods: ForegroundPeriod[] = []
trackFocus(addNewForegroundPeriod)
trackBlur(closeForegroundPeriod)
```
to use browser terminology when using browser concepts and our naming for our own structure / logic.

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
bcaudan: >  it means we never closed the previous one. And if we do not it. It will continue to increase.

from a complexity perspective, I'd say let's ignore the new focus and keep the current one open

> I would like to show that's its two different periods.

why? would it be usable somehow?

> We can also decide to discard to unclosed period because it's not normal.

hum, do we prefer to miss focused periods or blur periods?

> Let's add some monitoring because it's not normal

sounds good to me, with the simplest approach in the mean time if we don't have good reason to add extra complexity

packages/rum-core/src/domain/assembly.ts
bcaudan: Using `foregroundContexts` in `viewCollection`, `actionCollection` and `errorCollection` instead of `assembly` could make that more natural.
It would also avoid the conversion from serverDuration to duration which feel fishy.
wdyt?

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,123 @@
...
+    .filter((focus) => {
+      const eventEndsBeforeFocusStart = eventEndTime < focus.start
+      const eventStartsAfterFocusEnds = focus.end == null || eventStartTime > focus.end
+
+      return !eventEndsBeforeFocusStart || !eventStartsAfterFocusEnds
+    })
+    .map((focusPeriod) => {
+      const startTime = eventStartTime > focusPeriod.start ? eventStartTime : focusPeriod.start
+      const startDuration = elapsed(eventStartTime, startTime)
+      const endTime = focusPeriod.end == null || eventEndTime < focusPeriod.end ? eventEndTime : focusPeriod.end
```
BenoitZugmeyer, line 116: `focusPeriod.end` cannot be null, could you use `focusePeriod.end === undefined` instead?

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
ThibautGeriz: I was planning to do so in the backend later on, or on web ui for opened replay.
This function only monitor it for DD or is it doing it too for our client?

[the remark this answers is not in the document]
bcaudan: only DD for now

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
ThibautGeriz: > hum, do we prefer to miss focused periods or blur periods?

From a replay point of view, I prefer to say to the user that we have a foreground period with nothing happening then saying nothing is happing while in reality there are stuff happening.

So I would prefer to miss blur periods.
I am merging the two periods

packages/rum-core/src/domain/foregroundContexts.ts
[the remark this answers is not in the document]
ThibautGeriz: > Did you understand what happen?

Not really, I've added a bunch of internalMonitoring to debug this

>  if yes, it could nice to add a test explaining this case

I've got test case but it's just very artificial and I do not get the exact root cause

packages/rum-core/src/domain/foregroundContexts.ts
ThibautGeriz: I like to use `!= null` && `==null` to cover all cases but I do not have a strong opinion, I guess with typescript, the typechecker will yell if null is a possibility and it will be okay.

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.ts
```diff
@@ -1,63 +1,65 @@
...
-      action: {
-        count: view.eventCounts.userActionCount,
-      },
-      cumulative_layout_shift: view.cumulativeLayoutShift,
-      dom_complete: toServerDuration(view.timings.domComplete),
-      dom_content_loaded: toServerDuration(view.timings.domContentLoaded),
-      dom_interactive: toServerDuration(view.timings.domInteractive),
-      error: {
-        count: view.eventCounts.errorCount,
+  function processViewUpdate(view: ViewEvent) {
```
bcaudan, line 12: any reason to move processViewUpdate inside startViewCollection?

passing foregroundContext as parameter instead could ease the review

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
```diff
@@ -51,9 +56,14 @@ function processAction(action: AutoAction | CustomAction) {
...
       date: action.startClocks.timeStamp,
       type: RumEventType.ACTION as const,
+      view: {},
     },
     autoActionProperties
   )
+  const inForeground = foregroundContexts.getInForeground(action.startClocks.relative)
+  if (inForeground !== undefined) {
+    actionEvent.view.in_foreground = inForeground
+  }
```
bcaudan, line 66:
```suggestion
    },
    autoActionProperties
  )
  const inForeground = foregroundContexts.getInForeground(action.startClocks.relative)
  if (inForeground !== undefined) {
    actionEvent.view = { in_foreground: inForeground }
  }
```
no need to create a view object if foreground context is not enabled

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,240 @@
+import { RelativeTime, relativeNow, toServerDuration, Duration } from '@datadog/browser-core'
+import { setup, TestSetupBuilder } from '../../test/specHelper'
+import { createNewEvent } from '../../../core/test/specHelper'
+import { InForegroundPeriod } from '../rawRumEvent.types'
+import { startForegroundContexts, ForegroundContexts } from './foregroundContexts'
+
+const FOCUS_PERIOD_LENGTH = 10 as Duration
+const BLUR_PERIOD_LENGTH = 5 as Duration
+
+describe('foreground context', () => {
```
bcaudan, line 10: From [codecov report](https://app.codecov.io/gh/DataDog/browser-sdk/compare/854/tree/packages/rum-core/src/domain/foregroundContexts.ts), some code paths are not covered.
Should they?

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+      })
+
+      describe('getInForeground', () => {
+        it('should return false in the first blurred period', () => {
+          expect(foregroundContext.getInForeground(2 as RelativeTime)).toEqual(false)
+        })
+
+        it('should return true in the first focused period', () => {
+          expect(foregroundContext.getInForeground(10 as RelativeTime)).toEqual(true)
+        })
```
bcaudan, line 82: cf https://github.com/DataDog/browser-sdk/pull/854#discussion_r635348898
what about replacing the test description by comments:
```suggestion
          // first blurred period
          expect(foregroundContext.getInForeground(2 as RelativeTime)).toEqual(false)
         // first focused period
          expect(foregroundContext.getInForeground(10 as RelativeTime)).toEqual(true)

```

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+    if (startTime > foregroundPeriod.start && (foregroundPeriod.end == null || startTime < foregroundPeriod.end)) {
+      return true
+    }
+  }
+  return false
+}
+
+function getInForegroundPeriods(eventStartTime: RelativeTime, duration: Duration): InForegroundPeriod[] {
+  const eventEndTime = ((eventStartTime as number) + (duration as number)) as RelativeTime
```
bcaudan, line 106: for lint issue about addition, for now we ignore it instead of casting every member to number.
I'd be in favor of being consistent for when we'll deal with it

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+          })
+
+          it('should have the whole second period', () => {
+            expect(periods![1]).toEqual({
+              start: toServerDuration(20 as Duration),
+              duration: toServerDuration(10 as Duration),
+            })
+          })
+
+          it('should have the thrid period finishing with the view end', () => {
```
bcaudan, line 125:
```suggestion
          it('should have the third period finishing with the view end', () => {
```

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+          beforeEach(() => {
+            periods = foregroundContext.getInForegroundPeriods(0 as RelativeTime, 50 as Duration)
+          })
+          it('should have 3 in foreground periods', () => {
+            expect(periods).toHaveSize(3)
+          })
+
+          it('should have the whole first period', () => {
+            expect(periods![0]).toEqual({
+              start: toServerDuration(5 as Duration),
```
bcaudan, line 113: let's not mirror the implementation in the test, `5 * 1e6 as ServerDuration` is what we expect
if we don't want to deal with that here we could also move the conversion to ServerDuration inside `viewCollection`.

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+function addNewForegroundPeriod() {
+  if (foregroundPeriods.length > MAX_NUMBER_OF_FOCUSED_TIME) {
+    addMonitoringMessage('Reached maximum of foreground time')
+    return
+  }
+  const currentForegroundPeriod = foregroundPeriods[foregroundPeriods.length - 1]
+  if (currentForegroundPeriod != null && currentForegroundPeriod.end === undefined) {
```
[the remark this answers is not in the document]
BenoitZugmeyer: IMO comparing `currentForegroundPeriod` to `null` is counter-intuitive, since `currentForegroundPeriod` is never `null`. So either using its falsysness or comparing it to `undefined` seems find to me.

I don't think there is a strict convention on this, but I've never used `==` in my career, this operator scares me :D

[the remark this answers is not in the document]
ThibautGeriz: > I don't think there is a strict convention on this, but I've never used == in my career, this operator scares me :D

Using it with `null` is fine. I always do that.
Since eslint is okay with that we should add a eslint rule to remove it so we don't have to check for that in review.

Falsy check scares me a lot more than `== null`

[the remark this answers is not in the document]
bcaudan: agree on adding an eslint rule to avoid those discussions.

the "convention" here is that we don't use `==` and `null`.

If you are not fond of falsy checks, that should work:
```suggestion
  if (currentForegroundPeriod !== undefined && currentForegroundPeriod.end === undefined) {
```

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.ts
bcaudan: nit: I'd be in favor to do it, it would also avoid to mess the history for those lines.

packages/rum-core/src/domain/foregroundContexts.spec.ts
bcaudan: not testing feature flag false is fine by me

packages/rum-core/src/domain/foregroundContexts.spec.ts
bcaudan: as explained before, with this approach, the test description is a duplication of the expectation.

> Label are shown in the run test providing a bit of documentation contrary to comment

IMO, it adds noise on the file and the label will only be shown when the test fails, so instead locating the expectation and the related comment in the source code should be fairly simple.  

I'd be in favor of either:

- having a single test by case you want to test, clearer than a scenario with a lot of expectations
- having a single test with a scenario and all the wanted expectations, less boilerplate and duplication

packages/rum-core/src/domain/foregroundContexts.ts
bcaudan: what I meant was:
```suggestion
  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
  const eventEndTime = (eventStartTime + duration) as RelativeTime
```

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -190,6 +214,49 @@ describe('foreground context', () => {
         })
       })
     })
+
+    describe(`after reaching the maximum number of focus periods: ${MAX_NUMBER_OF_FOCUSED_TIME}`, () => {
+      beforeEach(() => {
```
bcaudan, line 219: what is the need for a separate suite and a beforeEach since there is a single test in it?

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -190,6 +214,49 @@ describe('foreground context', () => {
...
+        // given
```
bcaudan, line 220: not super fond of this types of comments, IMO you can achieve the same with spacing between the different phases.

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -51,6 +51,9 @@ export interface RawRumErrorEvent {
     source: ErrorSource
     message: string
   }
+  view?: {
+    in_foreground?: boolean
```
bcaudan, line 55: can `in_foreground` be undefined now?

packages/rum-core/src/domain/foregroundContexts.spec.ts
ThibautGeriz: I like to extract the setup because it allow me to add more test in it leveraging the nested structure of jest but you seems to prefer less verbose tests. I will refactor that

packages/rum-core/src/rawRumEvent.types.ts
ThibautGeriz: No, I will remove it

packages/rum-core/src/domain/foregroundContexts.spec.ts
```diff
@@ -166,7 +166,8 @@ describe('foreground context', () => {
 
     it('should not record anything after reaching the maximum number of focus periods', () => {
       const { clock } = setupBuilder.build()
-      Array.from({ length: MAX_NUMBER_OF_FOCUSED_TIME + 1 }).forEach(() => {
+      // eslint-disable-next-line prefer-spread
+      Array.apply(null, Array(MAX_NUMBER_OF_FOCUSED_TIME + 1)).forEach(() => {
```
bcaudan, line 170: I like functional approach too but a for-loop seems simpler in this context

PR #858, opened 2021-05-19
⚡️ start mutation observer only when needed

amortemousque:
## Motivation

Start MutationObserver only when needed for performance optimization

## Changes

- start DOM Mutation observation only when we need to track the page activities

## Testing

Unit, local

---


packages/rum-core/src/domain/trackPageActivities.ts
```diff
@@ -66,6 +67,8 @@ export function trackPageActivities(
   let firstRequestIndex: undefined | number
   let pendingRequestsCount = 0
 
+  const DOMMutationCollection = startDOMMutationCollection(lifeCycle)
```
bcaudan, line 70: two design concerns with this approach:

- using lifeCycle does not seem to be needed anymore since the component that needs the mutation is the one that start the mutation observer
- if another component need to use a mutation observer, it will use the same instance but stop conditions could be different

bcaudan: In fact, we can already have several `waitIdlePageActivity` at the same time so stopping the shared mutation observer instance can already be an issue

amortemousque: A new instance of `MutationObserver` with its own `stop()` is created each time we call `startDOMMutationCollection` function. So we can have multiple call to `waitIdlePageActivity` without issues or maybe I miss something?

bcaudan: oh my bad, I thought it was a singleton but it is not. 
So then we can have several MutationObservers in parallel, meaning that one dom mutation will trigger several DOM_MUTATED lifecycle events 🤔
It could be nice to instead have a single observer and keep track of subscribers to stop it when nobody needs it anymore.


what about having something like:
```ts
export function trackPageActivities(
  lifeCycle: LifeCycle,
  mutations: Mutations
) {
  
  ...

  subscriptions.push(mutations.subscribe(() => notifyPageActivity()))

  ...

  return {
    stop: () => {
      subscriptions.forEach((s) => s.unsubscribe())
    }
  }
}
```
mutations could have a similar API than lifeCycle limiting the impact to the tests

packages/rum-core/src/browser/domMutationObserver.spec.ts
```diff
@@ -21,21 +20,17 @@ describe('domMutationCollection', () => {
       root.appendChild(document.createTextNode('foo'))
       document.body.appendChild(root)
 
-      const lifeCycle = new LifeCycle()
+      const DOMMutation = createDOMMutationObservable()
```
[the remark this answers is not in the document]
amortemousque: I agree that we need to enforce camelCase with the linter but the code includes also snake_case. 
So they are different strategies:

- Disable the linter in the .eslint for specific files or names.
- Disable the linter in the code with a comment.
- We can try to allow camelCase and snake_case at the same time, but I'm not comfortable with this one.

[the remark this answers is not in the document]
bcaudan: Would https://eslint.org/docs/rules/dot-notation#allowpattern fit our case?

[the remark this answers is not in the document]
amortemousque: Good idea. I can investigate it in another PR.

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -1,8 +1,64 @@
...
+    if (!observer) {
+      return
+    }
+    observer.observe(document, {
+      attributeOldValue: true,
+      attributes: true,
+      characterData: true,
+      characterDataOldValue: true,
+      childList: true,
+      subtree: true,
```
BenoitZugmeyer, line 27: Any reason to change the options here? I think we don't need the `*OldValue` ones

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -1,8 +1,64 @@
-import { monitor } from '@datadog/browser-core'
-import { LifeCycle, LifeCycleEventType } from '../domain/lifeCycle'
+import { monitor, Subscription } from '@datadog/browser-core'
 
-type MutationObserverConstructor = new (callback: MutationCallback) => MutationObserver
+export interface DOMMutation {
```
BenoitZugmeyer, line 3: What about renaming this interface to `DOMMutationObservable`, since it is the value returned by `createDOMMutationObservable` (as well as its usages `domMutation` -> `domMutationObservable`)? `DOMMutation` might be confused with the object emited by the `Observer` (`MutationRecord`).

packages/rum-core/test/createDomMutationStub.ts
```diff
@@ -0,0 +1,25 @@
...
+    subscribe: (callback: () => void) => {
+      callbacks.push(callback)
+      return {
+        unsubscribe: () => {
+          callbacks = callbacks.filter((other) => callback !== other)
+        },
+      }
+    },
+  }
+}
```
BenoitZugmeyer, line 25: this seems to match the `Observable` API, maybe we don't need it and we could use an `Observable` directly to stub it?

packages/rum-core/src/browser/domMutationObserver.ts
```diff
@@ -0,0 +1,55 @@
+import { Subscription } from '../domain/lifeCycle'
```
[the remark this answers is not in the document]
amortemousque: I've moved Subscription with core/src/tools/observable.ts

packages/rum-core/src/browser/domMutationObservable.ts
amortemousque: Yes that's a good point

packages/rum-core/test/createDomMutationStub.ts
amortemousque: Now that I've implemented the `unsubscribe()` in `Observable`, it seems reasonable.

packages/rum-core/src/browser/domMutationObserver.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface DOMMutation {
+  notify(): void
```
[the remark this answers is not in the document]
amortemousque: Yes, I've used the already existing Observable object has a mock

packages/rum-core/src/browser/domMutationObservable.ts
amortemousque: You right

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -1,8 +1,62 @@
...
+
+export function createDOMMutationObservable(): DOMMutationObservable {
+  let callbacks: Array<() => void> = []
+  const MutationObserver = getMutationObserverConstructor()
+  const observer = MutationObserver ? new MutationObserver(monitor(notify)) : undefined
+  let isDOMObserved = false
+
+  function notify() {
+    callbacks.forEach((callback) => callback())
+  }
```
bcaudan, line 15: Could we use an observable internally rather than reimplementing the observable logic here?

packages/rum-core/src/domain/rumEventsCollection/action/trackActions.spec.ts
```diff
@@ -149,7 +149,7 @@ describe('newAction', () => {
     document.body.appendChild(root)
     setupBuilder = setup()
       .withFakeClock()
-      .beforeBuild(({ lifeCycle }) => trackActions(lifeCycle))
+      .beforeBuild(({ lifeCycle, domMutationObservable: DOMMutation }) => trackActions(lifeCycle, DOMMutation))
```
bcaudan, line 152:
```suggestion
      .beforeBuild(({ lifeCycle, domMutationObservable }) => trackActions(lifeCycle, domMutationObservable))
```

packages/rum-core/src/browser/domMutationObservable.spec.ts
```diff
@@ -21,23 +20,19 @@ describe('domMutationCollection', () => {
       root.appendChild(document.createTextNode('foo'))
       document.body.appendChild(root)
 
-      const lifeCycle = new LifeCycle()
+      const DOMMutation = createDOMMutationObservable()
```
bcaudan, line 23:
```suggestion
      const domMutationObservable = createDOMMutationObservable()
```

packages/rum-core/src/browser/domMutationObservable.ts
amortemousque: I also thought about it, but I have to hook into the observable logic to start and stop the MutationObserver which is currently not straightforward with the Observable is.

packages/rum-core/src/browser/domMutationObservable.spec.ts
amortemousque: Okay

packages/rum-core/src/browser/domMutationObservable.ts
bcaudan: what about something like https://github.com/DataDog/browser-sdk/commit/f9afcc2b4ab5613deac42f373999d83da89e22d5?

bcaudan (approved):

PR #859, opened 2021-05-19
✨ add id on rum events

amortemousque:
## Motivation

Add id on rum events to better trace them in the backend

## Changes

- Add ids to views, resources, long tasks, errors, actions

## Testing

Unit, local
---


packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.ts
```diff
@@ -26,6 +26,7 @@ function processViewUpdate(view: ViewEvent) {
     date: preferredTimeStamp(view.startClocks),
     type: RumEventType.VIEW,
     view: {
+      id: view.id,
```
bcaudan, line 29: Adding the id here does not seem useful since it will be overridden later
what about not adding it?

amortemousque: I can't see where it is overriden?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -48,6 +48,7 @@ function processRequest(request: RequestCompleteEvent) {
     {
       date: preferredTimeStamp(startClocks),
       resource: {
+        id: generateUUID(),
```
BenoitZugmeyer, line 51: Since now all resources have an id, you could remove the `resource.id` from the `computeRequestTracingInfo` function below.

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
```diff
@@ -44,6 +44,7 @@ function processAction(action: AutoAction | CustomAction) {
   const actionEvent = combine(
     {
       action: {
+        id: generateUUID(),
```
bcaudan, line 47: What about generating/adding it only for custom actions?

packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -298,7 +298,7 @@ describe('rum assembly', () => {
         rawRumEvent: createRawRumEvent(RumEventType.ACTION),
         startTime: 0 as RelativeTime,
       })
-      expect((serverRumEvents[0] as RumActionEvent).action.id).not.toBeDefined()
+      expect((serverRumEvents[0] as RumActionEvent).action.id).toBeDefined()
```
bcaudan, line 301: what is the intent of this expectation?

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -39,7 +39,7 @@ describe('viewCollection', () => {
         resourceCount: 10,
         userActionCount: 10,
       },
-      id: 'xxx',
+      id: 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee',
```
bcaudan, line 42: what is the purpose of this change?

amortemousque: The view id has to respect a certain format compliant with rum-events-format

bcaudan: 🤦‍♂️ indeed

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
amortemousque: I'm probably missing a point. I don't really see why we only have to target custom actions?

packages/rum-core/src/domain/assembly.spec.ts
amortemousque: Shouldn't the id be required for all actions?

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
bcaudan: action id for auto action is generated earlier in the code and retrieved in `autoActionProperties`.
With the current code, for an auto action, an id is generated and overwritten immediately by the id from `autoActionProperties`.

IMO, we could make the intent clearer by having something like:
```ts
const actionTypeProperties = isAutoAction(action)
    ? {
        action: {
          error: {
            count: action.counts.errorCount,
          },
          id: action.id,
          loading_time: toServerDuration(action.duration),
          long_task: {
            count: action.counts.longTaskCount,
          },
          resource: {
          },
        },
      }
    : {
        action: {
          id: generateUUID()
        }
      }
```

packages/rum-core/src/domain/assembly.spec.ts
bcaudan: yes, but it is not the goal of this test.
In the context of this test, the idea is to check by event type is the action context is added to the event or not, meaning testing [this function](https://github.com/DataDog/browser-sdk/blob/abccf2c/packages/rum-core/src/domain/assembly.ts#L127)

amortemousque: okay done ;)

PR #860, opened 2021-05-19
✨ [RUMF-913] allow masking input values

BenoitZugmeyer:
## Motivation

Let the user chose to mask input values (replacing the characters with asterisks) instead of completely ignoring them via a new privacy attribute value `data-dd-privacy="input-masked"` / `class="dd-privacy-input-masked"`.

## Changes

This PR removes the collection of some input values (radio/checkbox). Those values are likely to be useless during replay (note: if you think of a usecase, let me know). This way, we don't have to chose whether we should ignore/mask them.

It introduces an "input privacy mode", either computed by iterating over a node parent or propagated while serializing nodes.

It factorizes the input elements "value" computation by introducing a common function that will respect the "input privacy mode" and will be used while:
* the value is changed by the user, triggering an "input" event (`IncrementalSnapshot/Input`)
* the value is read from the element while serializing the node (`FullSnapshot`, `IncrementalSnapshot/Mutation` on child list)
* the value is programmatically set via `setAttribute` (`IncrementalSnapshot/Mutation` on attribute)

Note: those last two points where missing from the upstream rrweb-snapshot implementation.

## Testing

Manual, unit, e2e

---

### DIFF
--- a/packages/rum-recorder/src/domain/record/observer.ts
+++ b/packages/rum-recorder/src/domain/record/observer.ts
@@ -1,7 +1,6 @@
 import { monitor, callMonitored, throttle, DOM_EVENT, addEventListeners, addEventListener } from '@datadog/browser-core'
-import { InputPrivacyMode } from '../../constants'
-import { nodeOrAncestorsShouldBeHidden, getNodeOrAncestorsInputPrivacyMode } from './privacy'
-import { getSerializedNodeId, hasSerializedNode } from './serializationUtils'
+import { nodeOrAncestorsShouldBeHidden } from './privacy'
+import { getElementInputValue, getSerializedNodeId, hasSerializedNode } from './serializationUtils'
 import {
   FocusCallback,
   HookResetter,
@@ -157,17 +156,11 @@ function initViewportResizeObserver(cb: ViewportResizeCallback): ListenerHandler
 
 export const INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT']
 const lastInputValueMap: WeakMap<EventTarget, InputValue> = new WeakMap()
-function initInputObserver(cb: InputCallback): ListenerHandler {
+export function initInputObserver(cb: InputCallback): ListenerHandler {
   function eventHandler(event: { target: EventTarget | null }) {
     const target = event.target as HTMLInputElement | HTMLTextAreaElement
 
-    if (
-      !target ||
-      !target.tagName ||
-      INPUT_TAGS.indexOf(target.tagName) < 0 ||
-      nodeOrAncestorsShouldBeHidden(target) ||
-      getNodeOrAncestorsInputPrivacyMode(target) === InputPrivacyMode.IGNORED
-    ) {
+    if (!target || !target.tagName || INPUT_TAGS.indexOf(target.tagName) < 0 || nodeOrAncestorsShouldBeHidden(target)) {
       return
     }
 
@@ -177,7 +170,11 @@ function initInputObserver(cb: InputCallback): ListenerHandler {
     if (type === 'radio' || type === 'checkbox') {
       inputValue = { isChecked: (target as HTMLInputElement).checked }
     } else {
-      inputValue = { text: target.value }
+      const value = getElementInputValue(target)
+      if (value === undefined) {
+        return
+      }
+      inputValue = { text: value }
     }
 
     cbWithDedup(target, inputValue)
--- a/packages/rum-recorder/src/domain/record/observer.ts
+++ b/packages/rum-recorder/src/domain/record/observer.ts
@@ -1,5 +1,6 @@
 import { monitor, callMonitored, throttle, DOM_EVENT, addEventListeners, addEventListener } from '@datadog/browser-core'
-import { nodeOrAncestorsShouldBeHidden, nodeOrAncestorsShouldHaveInputIgnored } from './privacy'
+import { InputPrivacyMode } from '../../constants'
+import { nodeOrAncestorsShouldBeHidden, getNodeOrAncestorsInputPrivacyMode } from './privacy'
 import { getSerializedNodeId, hasSerializedNode } from './serializationUtils'
 import {
   FocusCallback,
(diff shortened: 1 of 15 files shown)


packages/rum-recorder/src/domain/record/observer.ts
```diff
@@ -156,40 +156,35 @@ function initViewportResizeObserver(cb: ViewportResizeCallback): ListenerHandler
...
+    const target = event.target as HTMLInputElement | HTMLTextAreaElement
 
-    if (
-      !target ||
-      !(target as Element).tagName ||
-      INPUT_TAGS.indexOf((target as Element).tagName) < 0 ||
-      nodeOrAncestorsShouldBeHidden(target) ||
-      nodeOrAncestorsShouldHaveInputIgnored(target)
-    ) {
+    if (!target || !target.tagName || INPUT_TAGS.indexOf(target.tagName) < 0 || nodeOrAncestorsShouldBeHidden(target)) {
```
bcaudan, line 163: nit:
```suggestion
    if (!target || !target.tagName || !includes(INPUT_TAGS, target.tagName) || nodeOrAncestorsShouldBeHidden(target)) {
```

packages/rum-recorder/src/domain/record/types.ts
```diff
@@ -216,8 +219,8 @@ export interface ViewportResizeDimention {
 export type ViewportResizeCallback = (d: ViewportResizeDimention) => void
 
 export interface InputValue {
-  text: string
-  isChecked: boolean
+  text?: string
+  isChecked?: boolean
```
bcaudan, line 223: Could it be  `{ text: string } | { isChecked: boolean }`?

packages/rum-recorder/src/domain/record/serializationUtils.ts
```diff
@@ -74,3 +76,50 @@ export function makeSrcsetUrlsAbsolute(attributeValue: string, baseUrl: string)
...
+    type === 'checkbox'
+  ) {
+    return
+  }
+
+  return inputPrivacyMode === InputPrivacyMode.MASKED ? maskValue(value) : value
+}
+
+export function maskValue(value: string) {
+  return value.replace(/./g, () => '*')
```
bcaudan, line 124:
```suggestion
  return value.replace(/./g, '*')
```

packages/rum-recorder/src/domain/record/privacy.spec.ts
```diff
@@ -57,58 +58,60 @@ describe('privacy helpers', () => {
...
+      expect(getNodeInputPrivacyMode(node)).toBeUndefined()
     })
     it('considers a DOM Element with a data-dd-privacy="input-ignored" attribute to be ignored', () => {
       const node = document.createElement('input')
       node.setAttribute('data-dd-privacy', 'input-ignored')
-      expect(nodeShouldHaveInputIgnored(node)).toBeTruthy()
+      expect(getNodeInputPrivacyMode(node)).toBe(InputPrivacyMode.IGNORED)
     })
-    it('considers a DOM Element with a data-dd-privacy="foo" attribute as not to be ignored', () => {
+    it('cannot determine the input privacy mode for a DOM Element with a data-dd-privacy="foo" attribute', () => {
```
bcaudan, line 72: nit: what about using `unknown-mode` instead of `foo`?

vlad-mh (approved):

packages/rum-recorder/src/domain/record/privacy.ts
```diff
@@ -36,30 +39,53 @@ export function nodeOrAncestorsShouldBeHidden(node: Node | null): boolean {
...
-      node.classList.contains(PRIVACY_CLASS_INPUT_IGNORED) ||
-      // if element is an HTMLInputElement, check the type is not to be ignored by default
-      (isInputElement(node) && PRIVACY_INPUT_TYPES_TO_IGNORE.includes(node.type)))
-  )
-}
+/**
+ * Returns the given node input privacy mode without checking the privacy mode on the node
+ * ancestors. Returns 'undefined' if the privacy mode cannot be determined and should be taken from
+ * an ancestor.
+ */
```
bcaudan, line 46: FMU, it seems that we get the closest privacy mode to determine the privacy mode of the input.
so an input masked with an ignored parent would return masked.
Returning ignored would seem less surprising to me.
wdyt?

packages/rum-recorder/src/domain/record/serializationUtils.spec.ts
```diff
@@ -141,3 +150,101 @@ describe('makeUrlAbsolute', () => {
...
+  })
+
+  it('masks the value according to the InputPrivacyMode', () => {
+    const input = document.createElement('input')
+    input.value = 'foo'
+    input.setAttribute(PRIVACY_ATTR_NAME, PRIVACY_ATTR_VALUE_INPUT_MASKED)
+    expect(getElementInputValue(input)).toBe('***')
+  })
+
+  function wrapInInputIgnored(element: Element) {
```
bcaudan, line 239:
```suggestion
  function wrapInIgnoredParent(element: Element) {
```

packages/rum-recorder/src/domain/record/privacy.ts
BenoitZugmeyer: That's debatable! IMO using the closest privacy mode is simpler to understand, instead of establishing a prority between them. Let's make a quick poll to get the thoughts of other devs.

packages/rum-recorder/src/domain/record/types.ts
BenoitZugmeyer: I initially did that but it complexify the comparison here: https://github.com/DataDog/browser-sdk/blob/6c894daf216141129ed9008464ae37af5d318246/packages/rum-recorder/src/domain/record/observer.ts#L198 (we'd have to do a bunch of cast, or check that each property is set before accessing it). Since I don't think that it matters too much I chose to keep it simple, but if you prefer I can change that.

bcaudan: I'd be in favor of doing it, it would clarify that it is one or the other

bcaudan (approved):

PR #861, opened 2021-05-21
🐛 ignore full snapshots taken before "load" event

BenoitZugmeyer:
## Motivation

The recorder starts recording at the document "load" event.

If a new view is created before the "load" event, a full snapshot is taken, then another one is taken at the "load" event, resulting in two consecutive full snapshots on the same view.

## Changes

This commit fixes this issue by ignoring any full snapshot occurring before the actual recording start, and makes sure early events have the correct `has_replay` value.

## Testing

Unit

---

### DIFF
--- a/packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
+++ b/packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
@@ -1,4 +1,4 @@
-import { Configuration, monitor, noop } from '@datadog/browser-core'
+import { Configuration, monitor, noop, runOnReadyState } from '@datadog/browser-core'
 import { LifeCycleEventType, makeRumPublicApi, RumUserConfiguration, StartRum } from '@datadog/browser-rum-core'
 
 import { startRecording } from './recorder'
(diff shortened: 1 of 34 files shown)


packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
```diff
@@ -31,7 +32,7 @@ export function makeRumRecorderPublicApi(startRumImpl: StartRum, startRecordingI
 
     const startRumResult = startRumImpl(userConfiguration, () => ({
       ...getCommonContext(),
-      hasReplay: state.status === RecorderStatus.Started ? true : undefined,
+      hasReplay: state.status === RecorderStatus.Started && state.isRecording() ? true : undefined,
```
bcaudan, line 35: not clear to me what is the difference between state.status and isRecording 🤔
could we have a single source of truth? or make the difference clearer?

packages/rum-recorder/src/domain/record/record.ts
```diff
@@ -58,87 +57,82 @@ export function record(options: RecordOptions): RecordAPI {
...
-          emit({
-            type: RecordType.Focus,
-            data,
-          }),
-      })
-    )
-  }
-
-  runOnReadyState('complete', init)
+  const handlers = [
```
bcaudan, line 62: nit: handlers array does not seem needed anymore
```suggestion
  const stopObservers = initObservers(...)
```

bcaudan (approved):

PR #865, opened 2021-05-26
🐛  [RUMF-876] Improve proxy behavior for xhr reuse

amortemousque:
## Detected issues

- We can have a `max stack size error` when using a single instance of `XMLHTTPRequest`.

## Changes

- Ensure to attach our listeners only once per request.

## Testing

Unit, Locally

---


packages/core/src/browser/xhrProxy.spec.ts
```diff
@@ -209,4 +209,26 @@ describe('xhr proxy', () => {
       },
     })
   })
+
+  it('should track only the second request (the first request is canceled)', (done) => {
```
BenoitZugmeyer, line 213: Beside a unit test, it could be nice to have some E2E tests in `test/e2e/scenario/rum/resources.scenario.ts`. Because XHRs are mocked for unit tests, we might miss issues occuring in real browsers.

packages/core/src/browser/xhrProxy.ts
```diff
@@ -77,66 +78,72 @@ function proxyXhr() {
   // eslint-disable-next-line @typescript-eslint/unbound-method
   originalXhrAbort = XMLHttpRequest.prototype.abort
 
+  function reportXhr(this: BrowserXHR, clearListeners: () => void) {
```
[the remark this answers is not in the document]
amortemousque: Well, mostly for two reasons:
- Readability: first, that would lead to 3 levels of nested function declarations which I find harder to read
- Performance: like this, we don't recreate a new function each time the function sends is called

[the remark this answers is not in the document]
BenoitZugmeyer: > Readability

If we go this way, you might want to move it outside of `proxyXhr` function, since it doesn't use anything from it. Also, using `this` to pass the XHR instance is kind of unintuitive, and since `clearListeners` is called unconditionally at the start of the function, I don't see why it should be passed to this function. What about using a normal argument for `xhr` and remove the `clearListeners` argument?

> Performance

This should have a negligeable performance impact.

packages/core/src/browser/xhrProxy.ts
```diff
@@ -77,66 +78,72 @@ function proxyXhr() {
...
     callMonitored(() => {
       // WARN: since this data structure is tied to the instance, it is shared by both logs and rum
       // and can be used by different code versions depending on customer setup
       // so it should stay compatible with older versions
-      ;(this._datadog_xhr as Pick<XhrStartContext, 'method' | 'url'>) = {
+      this._datadog_xhr = {
         method,
         url: normalizeUrl(url),
-      }
+      } as XhrStartContext
```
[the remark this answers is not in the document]
amortemousque: I've added a third type and made BrowserXHR generic. Let me know if that okay for you.

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -201,6 +201,56 @@ describe('rum resources', () => {
...
+
+        await flushEvents()
+
+        const resourceEvents = events.rumResources.filter((event) => event.resource.type === 'xhr')
+        expect(resourceEvents.length).toEqual(1)
+        expect(events.rumErrors.length).toBe(0)
+        expect(resourceEvents[0]?.resource.status_code).toBe(200)
+      })
+
+    createTest('track both XHRs when the second wait for the respond of the first')
```
bcaudan, line 229: was there an issue?

amortemousque: I'm checking that the first request isn't canceled by the second one when waiting for it to be finished.

packages/core/src/browser/xhrProxy.spec.ts
bcaudan, line 213: FMU, there was a bit more to it.
The first one is cancel only if the second one start before its end.
Should not we reflect that in the test and test both cases? or only rely on e2e scenario?

bcaudan, line 213: Do we have test cases for the other issues/changes:
```
We can have a max stack size error when using a single instance of XMLHTTPRequest.
Some requests can be missed due to the override of onreadystatechange by the user
Attach our event listeners sooner in the XMLHTTPRequest lifecycle
Ensure to attach our listeners only once per request.
```

amortemousque: You'r right, because this behaviours are tied to the browser, I'd be in favour of relying only on the e2e tests. 
In my opinion, it's the only reliable way to know if our implementation comply with browsers behaviours.

bcaudan: let's remove this one then

packages/core/src/browser/xhrProxy.spec.ts
```diff
@@ -110,9 +110,9 @@ describe('xhr proxy', () => {
     const onReadyStateChange = jasmine.createSpy()
     withXhr({
       setup(xhr) {
-        xhr.onreadystatechange = onReadyStateChange
-        xhr.addEventListener('load', () => xhr.abort())
         xhr.open('GET', '/ok')
+        xhr.addEventListener('readystatechange', onReadyStateChange)
+        xhr.addEventListener('load', () => xhr.abort())
         xhr.send()
```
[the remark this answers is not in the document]
amortemousque: FMU, this test is related to the Angular use case. By looking at their implementation [here](https://github.com/angular/angular/blob/master/packages/common/http/src/xhr.ts).
I can't find any readystatechange handler in their code. What I found, is that they [abort in-flight requests in the unsubscribe on their observable](https://github.com/angular/angular/blob/3d1c905c9190f467cc6f8b0c3c612cdd203c3f75/packages/common/http/src/xhr.ts#L317) by looking at the xhr `readyState`. Both the previous implementation and the new one work because onreadystatechange is called before the xhr apply internally the DONE readyState.

That's why I changed the test to be closer to the Angular implementation.

packages/core/src/browser/xhrProxy.spec.ts
amortemousque: Not really, I'll make more unit tests to cover these cases.

packages/core/src/browser/xhrProxy.spec.ts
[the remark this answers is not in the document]
bcaudan: what about supporting both then?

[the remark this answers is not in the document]
amortemousque: I'm working on it

packages/core/src/browser/xhrProxy.spec.ts
amortemousque: The new version of the PR only focus on the first issue: We can have a max stack size error when using a single instance of XMLHTTPRequest.

packages/core/test/specHelper.ts
```diff
@@ -178,11 +184,28 @@ class StubXhr {
   }
 
   addEventListener(name: string, callback: () => void) {
-    this.fakeEventTarget.addEventListener(name, callback)
+    if (!this.listeners[name]) {
+      this.listeners[name] = []
+    }
```
BenoitZugmeyer, line 189: Did you need a custom event emitter implementation? We are using DOM-based event emitter in a few places because it is good enough for our usecase. If we have more particular requirements, I think we should implement a reusable event emitter class.

packages/core/test/specHelper.ts
```diff
@@ -133,16 +133,14 @@ class StubXhr {
   public status: number | undefined = undefined
   public readyState: number = XMLHttpRequest.UNSENT
   public onreadystatechange: () => void = noop
-
+  public listeners: { [k: string]: Array<() => void> } = {}
+  private isOnreadystatechangeAttributeCallFirst = true
```
BenoitZugmeyer, line 137: Maybe we could have something like this:

```
private _onreadystatechange: null | (() => void) = null

get onreadystatechange() {
  return this._onreadystatechange
}
set onreadystatechange(value) {
  if (!this._onreadystatechange) {
    this.addEventListener('readystatechange', () => this._onreadystatechange())
  }
  this._onreadystatechange = value
}
```

So we are sure that `readystatechange` listeners will be called in order, and we don't need to handle another state.

packages/core/test/specHelper.ts
```diff
@@ -220,17 +243,27 @@ export function stubXhr() {
...
   const xhr = new XMLHttpRequest()
-  xhr.addEventListener('loadend', () => {
+  const complete = (xhr: StubXhr | XMLHttpRequest) => {
     setTimeout(() => {
-      onComplete(xhr)
+      onComplete((xhr as unknown) as XMLHttpRequest)
     })
-  })
-  setup((xhr as unknown) as StubXhr)
+  }
```
BenoitZugmeyer, line 257: You wouldn't need to pass the `xhr` instance:

```ts
        const complete = () => {
          setTimeout(() => {
            onComplete(xhr)
          })
        })
```

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -201,6 +201,33 @@ describe('rum resources', () => {
...
+          xhr.open('GET', '/ok?duration=100&call=1')
+          xhr.send()
+        })
+
+        await flushEvents()
+
+        const resourceEvents = events.rumResources.filter((event) => event.resource.type === 'xhr')
+        expect(resourceEvents.length).toEqual(2)
+        expect(events.rumErrors.length).toBe(0)
+        expect(resourceEvents.map((resourceEvent) => resourceEvent.resource.status_code)).toEqual([200, 200])
```
BenoitZugmeyer, line 228: Maybe you could test for request URLs (`&call=1`/`&call=2`) to make sure the two resources are independent.

packages/core/src/browser/xhrProxy.spec.ts
```diff
@@ -154,7 +159,29 @@ describe('xhr proxy', () => {
     })
   })
 
-  it('should track request with onreadystatechange overridden', (done) => {
+  it('should track request with onreadystatechange overridden before open', (done) => {
+    withXhr({
+      setup(xhr) {
+        xhr.onreadystatechange = () => undefined
```
bcaudan, line 165: could it be interesting to ensure that this function is called?
(same for following test)

packages/core/test/specHelper.ts
```diff
@@ -167,7 +165,15 @@ class StubXhr {
     this.response = response
     this.status = status
     this.readyState = XMLHttpRequest.DONE
-    this.onreadystatechange()
+
+    if (this.isOnreadystatechangeAttributeCallFirst) {
+      this.onreadystatechange()
+      this.dispatchEvent('readystatechange')
```
bcaudan, line 171: do we listen for this event somewhere?
if not, what about not handling it?

packages/core/src/browser/xhrProxy.ts
```diff
@@ -69,85 +80,81 @@ export function resetXhrProxy() {
...
+      if (originalOnreadystatechange) {
+        originalOnreadystatechange.apply(this, arguments as any)
+      }
+    }
 
-  XMLHttpRequest.prototype.abort = function (this: BrowserXHR) {
-    callMonitored(() => {
-      if (this._datadog_xhr) {
-        this._datadog_xhr.isAborted = true
+    const onEnd = monitor(() => {
```
bcaudan, line 124: Does it change the behavior for the case where the xhr is aborted because the same instance reopen a request?

packages/core/test/specHelper.ts
amortemousque: I noticed with the DOM-based emitter (const eventEmitter = document.createElement('div')) that the "this" keyword is bound to the DIV element. In my case, I want it to be bound to StubXHR. In addition, DIVElement does not expose subscriptions, so we can't make any assertion about them. But you're right, I'm gonna isolate the implementation to make it reusable.

packages/core/src/browser/xhrProxy.ts
amortemousque: If you refer to the scenario 1 of the [RFC](https://docs.google.com/document/d/1nLE_EoAdqOYsYbtRej0qUruO63S6D-WUaWmQbj_RYnw/edit#heading=h.36igkj90d6fd), no the behaviour is still the same as the current version.

packages/core/test/specHelper.ts
amortemousque: Right, it's no logger needed

test/e2e/scenario/rum/resources.scenario.ts
amortemousque: Good point

BenoitZugmeyer (approved):
LGTM

packages/core/src/browser/xhrProxy.spec.ts
```diff
@@ -209,4 +237,50 @@ describe('xhr proxy', () => {
...
   })
+
+  it('should track multiple requests with the same xhr instance', (done) => {
+    let listeners: { [k: string]: Array<() => void> }
+    withXhr({
+      completionMode: 'manual',
+      setup(xhr, complete) {
+        const secondOnload = () => {
+          xhr.removeEventListener('load', secondOnload)
+          complete(xhr)
```
bcaudan, line 248:
```suggestion
      setup(xhr) {
        const secondOnload = () => {
          xhr.removeEventListener('load', secondOnload)
```
with this modification, the test still pass so I don't get what you are trying to achieve with manual mode.
could we discuss it?

packages/core/src/browser/xhrProxy.ts
```diff
@@ -2,8 +2,8 @@ import { monitor, callMonitored } from '../domain/internalMonitoring'
 import { Duration, elapsed, relativeNow, RelativeTime, ClocksState, clocksNow, timeStampNow } from '../tools/timeUtils'
 import { normalizeUrl } from '../tools/urlPolyfill'
 
-interface BrowserXHR extends XMLHttpRequest {
-  _datadog_xhr: XhrStartContext
+interface BrowserXHR<T extends XhrOpenContext = XhrStartContext> extends XMLHttpRequest {
```
bcaudan, line 5: `XhrStartContext` as default value seems a bit arbitrary to me, what about not having a default value?

packages/core/src/browser/xhrProxy.ts
```diff
@@ -69,85 +80,84 @@ export function resetXhrProxy() {
...
+  callMonitored(() => {
+    if (!this._datadog_xhr) {
+      return
+    }
+    this._datadog_xhr = {
+      ...this._datadog_xhr,
+      startTime: relativeNow(),
+      startClocks: clocksNow(),
+      isAborted: false,
+    }
```
bcaudan, line 106: Since it is available publicly, I'd be in favor of keeping the actual behavior which keep the same reference for this object during the lifecycle of the xhr.

packages/core/src/browser/xhrProxy.ts
amortemousque: yes, ok

packages/core/src/browser/xhrProxy.spec.ts
amortemousque: Yes, no longer needed

packages/core/src/browser/xhrProxy.ts
amortemousque: ok

packages/core/src/browser/xhrProxy.ts
```diff
@@ -151,10 +149,10 @@ function abortXhr(this: BrowserXHR) {
   return originalXhrAbort.apply(this, arguments as any)
 }
 
-function reportXhr(xhr: BrowserXHR) {
+function reportXhr(xhr: BrowserXHR<XhrStartContext>, pendingContext: XhrStartContext) {
   const xhrCompleteContext: XhrCompleteContext = {
-    ...xhr._datadog_xhr!,
-    duration: elapsed(xhr._datadog_xhr!.startClocks.timeStamp, timeStampNow()),
+    ...pendingContext,
+    duration: elapsed(pendingContext.startClocks.timeStamp, timeStampNow()),
```
bcaudan: what is the reasoning for not retrieving the context from xhr anymore?

packages/core/src/browser/xhrProxy.ts
```diff
@@ -69,85 +80,82 @@ export function resetXhrProxy() {
...
+function abortXhr(this: BrowserXHR<XhrStartContext>) {
+  callMonitored(() => {
+    if (this._datadog_xhr) {
+      this._datadog_xhr.isAborted = true
+    }
+  })
+  return originalXhrAbort.apply(this, arguments as any)
+}
+
+function reportXhr(xhr: BrowserXHR<XhrStartContext>, pendingContext: XhrStartContext) {
```
bcaudan, line 152: what is the reasoning for not retrieving the context from xhr anymore?

packages/core/src/browser/xhrProxy.ts
```diff
@@ -69,85 +80,82 @@ export function resetXhrProxy() {
...
+  const xhrCompleteContext: XhrCompleteContext = {
+    ...pendingContext,
+    duration: elapsed(pendingContext.startClocks.timeStamp, timeStampNow()),
+    response: xhr.response as string | undefined,
+    status: xhr.status,
   }
```
bcaudan, line 158: what about reinforcing the design by having something like:
```suggestion
function reportXhr(xhr: BrowserXHR<XhrCompleteContext>) {
  if (xhr._datadog_xhr) {
    xhr._datadog_xhr.duration = elapsed(pendingContext.startClocks.timeStamp, timeStampNow())
    ...
  }
```

amortemousque: Very good point. Seems much nicer !

packages/core/src/browser/xhrProxy.ts
amortemousque: Yes, I can simplify

packages/core/src/browser/xhrProxy.ts
bcaudan: it seems good enough to me

bcaudan (approved):

PR #867, opened 2021-05-27
⚗[RUMF-878] add trackViewsManually option (disabled)

bcaudan:
## Motivation

Following #850, another step for the manual view naming strategy, allowing to disable the automatic tracking of views on location change.

## Changes

- add `trackViewsManually` configuration (behind `view-renaming` flag)
- when `trackViewsManually` is enabled: 
  - rum start is postponed until the first `startView` call
  - new view is not created on session renew
  - current view location is still updated on location change
- rename `boot/[package].ts` to `boot/start[Package].ts` 

## Testing

unit, e2e, manual

---


packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -53,62 +53,43 @@ export interface ViewEndedEvent {
...
-export function trackViews(location: Location, lifeCycle: LifeCycle) {
+type View = ReturnType<typeof newView>
+
+export function trackViews(location: Location, lifeCycle: LifeCycle, areViewsTrackedAutomatically: boolean) {
   let isRecording = false
+  let hasInitialView = false
+  let currentView: View
+  let stopInitialViewTracking: () => void
+  let stopViewLifeCycle: () => void
+  let stopLocationChangesTracking: (() => void) | undefined
```
BenoitZugmeyer, line 64: We might want to explicit this as a state like:

```ts
type TrackViewsState =
{
  status: TrackViewStatus.WaitingForInitialView,
} |
{
  status: TrackViewStatus.Started,
  currentView: View,
  stop: () => void,
} |
{
  status: TrackViewStatus.Stopped
}
```

It would make clear that `currentView` is not always defined, and what should be called to stop the view tracking.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -120,21 +101,69 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
...
+    })
+
+    lifeCycle.subscribe(LifeCycleEventType.RECORD_STARTED, () => {
+      isRecording = true
+      currentView.updateHasReplay(true)
+    })
+
+    lifeCycle.subscribe(LifeCycleEventType.RECORD_STOPPED, () => {
+      isRecording = false
+    })
```
BenoitZugmeyer, line 127: We probably want to prevent recording to start if there is no view, like other public APIs, else this will not be accurate.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -120,21 +101,69 @@ export function trackViews(location: Location, lifeCycle: LifeCycle) {
...
 
+  function startViewLifeCycle() {
+    lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
+      // do not trigger view update to avoid wrong data
+      currentView.end()
+      if (areViewsTrackedAutomatically) {
+        // Renew view on session renewal
+        currentView = trackViewChange()
+      }
+    })
```
BenoitZugmeyer, line 112: Has it be discussed somewhere? Since the user can't know when the session is renewed, they won't be able to create a new view in this case.

bcaudan: > Has it be discussed somewhere?

No!
IMO, it seems more aligned with the manual tracking strategy to not take the responsibility to start a new view in this case. It can also be deceptive to have view without name in this case and I am not sure that we can make the assumption that we can reuse the previous view name. 
We could expose something like `DD_RUM.onSessionRestart(callback)` API to allow customer to deal with these cases.
wdyt?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: I look into it

degregar:
Hi!
I guess this new feature will help me using `addTiming`. Right now it counts time relative to the initial view timestamp. I've noticed, that when some resources load longer, `addTiming` is longer as well. I have a SPA app and I need to measure how long it's bootstrapping, and `addTiming` is unreliable to do this.

So I guess I could trigger `startView()` and from then on it would restart the clock? Or is there any other way to achieve this?

Do you have any idea when you will be ready with that feature?

bcaudan:
Hi @degregar,

This is a work in progress but we hope to have something usable publicly in the coming weeks.
Would you mind opening a separate issue to describe your expectations about `addTiming` and the behavior you observed?  
It would be easier to follow the conversation outside of this PR 🙂

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: I've updated the strategy to always have an initial view

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -67,38 +70,62 @@ export function makeRumPublicApi<C extends RumUserConfiguration>(startRumImpl: S
...
+      // drain beforeInitCalls by buffering them until we start RUM
+      // if we get a startView, drain re-buffered calls before continuing to drain beforeInitCalls
+      // in order to ensure that calls are processed in order
+      const beforeInitCalls = bufferApiCalls
+      bufferApiCalls = new BoundedBuffer()
+
+      startViewStrategy = (name) => {
+        doStartRum(name)
       }
+      beforeInitCalls.drain()
```
BenoitZugmeyer, line 98: I have a bit of hard time wrapping my head around this, but it seems to handle our use cases nicely with little effort, so great!

packages/rum-recorder/src/boot/rumRecorderPublicApi.spec.ts
```diff
@@ -28,41 +28,95 @@ describe('makeRumRecorderPublicApi', () => {
...
-        enableExperimentalFeatures: ['postpone_start_recording'],
+        enableExperimentalFeatures: ['view-renaming'],
+        trackViewsManually: true,
+      }
+      it('starts RUM when initial view is started', () => {
+        expect(startRumSpy).not.toHaveBeenCalled()
+        rumRecorderPublicApi.init(MANUAL_VIEWS_CONFIGURATION)
+        expect(startRumSpy).not.toHaveBeenCalled()
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+        ;(rumRecorderPublicApi as any).startView()
```
BenoitZugmeyer, line 81: What about tweaking the `rumRecorderPublicApi` variable type instead of casting to any / disabling eslint?
```ts
let rumRecorderPublicApi: RumRecorderPublicApi & { startView(name: string) }
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -142,10 +169,10 @@ export function trackViews(location: Location, lifeCycle: LifeCycle, domMutation
       currentView = trackViewChange(startClocks, name)
     },
     stop: () => {
+      stopAutomaticViewCollection && stopAutomaticViewCollection()
```
BenoitZugmeyer, line 172: Let the bundler do the minification for us:
```suggestion
      if (stopAutomaticViewCollection) stopAutomaticViewCollection()
```
(you could also initialize it to `noop` to simplify)

packages/rum-core/src/boot/rumPublicApi.ts
bcaudan: yes, it handles our case without the need to modify the existing APIs but I felt the need to add some explanation about what is going on 😕 
Are you confortable with it?

packages/rum-recorder/src/boot/rumRecorderPublicApi.spec.ts
bcaudan: I have not put too much though about it since we will remove the cast when we will remove the feature flag.
Do you think it still worth to improve it?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: the issue disappear with latest change 😉

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
BenoitZugmeyer: Ok it makes sense to avoid making assumption on the new view name. I would prefer to avoid adding new APIs if possible. Maybe ensuring that the session is renewed when using `DD_RUM.startView()` could be enough?

packages/rum-core/src/boot/rumPublicApi.ts
BenoitZugmeyer: Yes it LGTM. I can't think of an alternative without introducing a lot more tricky code..

packages/rum-recorder/src/boot/rumRecorderPublicApi.spec.ts
BenoitZugmeyer: Ok no problem

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
```diff
@@ -1,14 +1,15 @@
...
 import { RumPerformanceNavigationTiming } from '../../../browser/performanceCollection'
+import { RumEventType } from '../../../rawRumEvent.types'
 import {
   PAGE_ACTIVITY_END_DELAY,
   PAGE_ACTIVITY_MAX_DURATION,
   PAGE_ACTIVITY_VALIDATION_DELAY,
 } from '../../trackPageActivities'
-import { ViewEvent, trackViews, THROTTLE_VIEW_UPDATE_PERIOD } from './trackViews'
+import { THROTTLE_VIEW_UPDATE_PERIOD } from './trackViews'
+import { setupViewTest, ViewTest } from './trackViews.spec'
```
BenoitZugmeyer, line 12: `.spec` files should not be imported, because their tests will be executed twice

bcaudan: oh, I was not aware of that 🤔 
nice strategy to bump the number of tests 😆 

maybe we could add a lint check on that

BenoitZugmeyer: It is caused by our webpack configuration when running karma: https://github.com/DataDog/browser-sdk/blob/2f661f2e66b49224b627921cd6887096ce05c3b0/test/unit/karma.base.conf.js#L49-L53 (this was the default with the previous version of karma-webpack). We need this because some tests are not properly cleaned-up. We may investigate how to improve the situation one day :)

bcaudan: what do you think of the lint rule in the mean time?

BenoitZugmeyer: It wouldn't hurt!

eslint-local-rules/disallowSpecImportSpec.js
```diff
@@ -0,0 +1,36 @@
...
+  },
+  create(context) {
+    if (!isSpecFile(context.getFilename())) {
+      return {}
+    }
+    return {
+      Program(node) {
+        reportSpecImportSpec(context, node)
+      },
+    }
```
BenoitZugmeyer, line 19: NIT: You could probably do this directly (without having a recursive function):

```js
return {
  ImportDeclaration(node) {
    if (isSpecFile(node.source.value)) {
      context.report({ ... })
    }
  }
}
```

eslint-local-rules/disallowSpecImportSpec.js
```diff
@@ -0,0 +1,36 @@
...
+        // eslint-disable-next-line max-len
+        'Disallow importing spec file code inside a spec file, it would execute the tests from the imported spec file twice',
+      recommended: false,
+    },
+    schema: [],
+  },
+  create(context) {
+    if (!isSpecFile(context.getFilename())) {
+      return {}
+    }
```
BenoitZugmeyer, line 14: We could run this on all test files, to catch sub-imports (ex: foo.spec imports specHelper imports bar.spec)

BenoitZugmeyer (approved):

eslint-local-rules/disallowSpecImportSpec.js
bcaudan: Let's check that everywhere then, an import on prod code would have the same behavior

BenoitZugmeyer (approved):

PR #869, opened 2021-05-27
✨[RUMF-909] allow event context edition in beforeSend

BenoitZugmeyer:
## Motivation

Allow `event.context` edition in beforeSend, to pass extra data defined by the user on each event.

## Changes

* always set the `context` field before calling `beforeSend`
* in limitModifications, accept any value modification as long as it keeps the same type
* add "context" to modifiable fields for non-view events

## Testing

Unit, manual

---


packages/rum-core/src/domain/assembly.ts
```diff
@@ -97,6 +101,9 @@ export function startRumAssembly(
           ;(serverRumEvent.usr as RumEvent['usr']) = commonContext.user as User & Context
         }
         if (shouldSend(serverRumEvent, configuration.beforeSend, errorFilter)) {
+          if (isEmptyObject(serverRumEvent.context)) {
+            delete serverRumEvent.context
+          }
```
bcaudan, line 106: removing the context when empty, is it an optimisation or is there more to it?

packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -49,94 +49,132 @@ describe('rum assembly', () => {
...
-        rawRumEvent: createRawRumEvent(RumEventType.LONG_TASK, {
-          view: { id: 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee' },
-        }),
-        startTime: 0 as RelativeTime,
+        expect(serverRumEvents[0].view.id).toBe('aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee')
       })
 
-      expect(serverRumEvents[0].view.id).toBe('aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee')
-    })
+      it('should allow modification on context field for events other than views', () => {
```
bcaudan, line 86: We should also add test for modification of `event.context` directly.

notably setting it to null, which seems to crash the SDK currently 😕
```ts
beforeSend: (event) => {
  event.context = null
}

File: packages/core/src/tools/limitModification.ts [hunk 3]
```diff
@@ -52,3 +52,15 @@ function set(object: unknown, path: string, value: string) {
...
 }
+
+function areSameType(valueA: unknown, valueB: unknown) {
+  if (valueA === null) {
+    return valueB === null
+  }
+
+  if (Array.isArray(valueA)) {
+    return Array.isArray(valueB)
+  }
```
bcaudan, line 63: could we have tests for those cases?

packages/rum-core/src/domain/assembly.ts
BenoitZugmeyer: It's to get the same behavior as before (which was an optimization, to avoid sending an unused property)

packages/core/src/tools/limitModification.spec.ts
```diff
@@ -47,18 +48,33 @@ describe('limitModification', () => {
...
 
     expect(object).toEqual({
-      foo: { bar: 'bar' },
-      qux: 'qux',
+      string_to_undefined: 'bar',
+      string_to_number: 'qux',
+      null_to_object: null,
+      object_to_null: {},
+      array_to_object: [],
+      object_to_array: {},
```
bcaudan, line 77: with this strict approach, I am not super confortable with:
- forcing customers to set context to `{}` to clear it, allowing undefined or null to clear the context would seem less surprising to me
- relying on the fact the the commonContext is init with `{}` to allow customers to overwrite the context with an object, if for some reason we change this implementation detail and end up with an undefined context before beforeSend, it would silently break this use case 

wdyt?

BenoitZugmeyer: > with this strict approach, I am not super confortable with:
> 
>     * forcing customers to set context to `{}` to clear it, allowing undefined or null to clear the context would seem less surprising to me

I think this is an edge case, but even then, using an empty object to set an empty context does not seem surprising to me.

> 
>     * relying on the fact the the commonContext is init with `{}` to allow customers to overwrite the context with an object, if for some reason we change this implementation detail and end up with an undefined context before beforeSend, it would silently break this use case

I don't think this is an implementation detail:

* the `context` property *needs* to be defined in the original object, because the object passed to the `beforeSend` callback is expected to have a defined `context` property. Unit tests are already ensuring this, but I will add a proper unit test to make it more explicit.

* `commonContext.context` is, in fact, the "global context", which has to be initialized to an empty object (at least to expose it via `getRumGlobalContext`)

If the `context` ends up undefined somehow, it will break at least a few tests.

bcaudan: > I think this is an edge case, but even then, using an empty object to set an empty context does not seem surprising to me.

my point was more about the fact that we can't empty it with undefined or null

> If the context ends up undefined somehow, it will break at least a few tests.

indeed, I'd be a bit more confortable with some higher level tests though

test/e2e/scenario/rum/init.scenario.ts
```diff
@@ -55,3 +55,20 @@ describe('before init API calls', () => {
       expect(afterManualViewAction.view.id).toBe(manualView.view.id)
     })
 })
+
+describe('beforeSend', () => {
+  createTest('allows to edit non-view events context')
```
bcaudan, line 60: Could you add another with setting the context directly?

bcaudan (approved):

PR #873, opened 2021-05-28
✨ add id on rum events

amortemousque:
## Motivation

Add id on rum events to better trace them in the backend

## Changes

- Add ids to views, resources, long tasks, errors, actions

## Testing

Unit, local
---

### DIFF
--- a/packages/rum-core/src/rawRumEvent.types.ts
+++ b/packages/rum-core/src/rawRumEvent.types.ts
@@ -13,7 +13,7 @@ export interface RawRumResourceEvent {
   type: RumEventType.RESOURCE
   resource: {
     type: ResourceType
-    id?: string // only for traced requests
+    id: string
     duration: ServerDuration
     url: string
     method?: string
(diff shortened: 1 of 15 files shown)


packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -112,7 +114,7 @@ export interface RawRumActionEvent {
   date: TimeStamp
   type: RumEventType.ACTION
   action: {
-    id?: string
+    id: string
```
bcaudan, line 117: should not the ids be optional?

packages/rum-core/src/rumEvent.types.ts
```diff
@@ -254,7 +254,7 @@ export type RumResourceEvent = CommonProperties & {
     /**
      * UUID of the resource
      */
-    readonly id?: string
+    readonly id: string
```
bcaudan, line 257: should not the id be optional?

amortemousque: yes

bcaudan (approved):
Nice 👍

PR #880, opened 2021-06-04
Monitor potential invalid date

bcaudan:
## Motivation

Investigate some report of events with `date: {}`

## Changes

add monitoring log before adding event to batch when event date is not an integer

## Testing

manual

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -9,6 +9,8 @@ import {
   timeStampNow,
   currentDrift,
   display,
+  addMonitoringMessage,
+  relativeNow,
 } from '@datadog/browser-core'
 import {
   CommonContext,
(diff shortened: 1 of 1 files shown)


packages/rum-core/src/domain/assembly.ts
```diff
@@ -98,6 +100,18 @@ export function startRumAssembly(
           ;(serverRumEvent.usr as Mutable<RumEvent['usr']>) = commonContext.user as User & Context
         }
         if (shouldSend(serverRumEvent, configuration.beforeSend, errorFilter)) {
+          if (!Number.isInteger(serverRumEvent.date)) {
```
BenoitZugmeyer, line 103: IE11 does not support `Number.isInteger`... maybe use something a bit more oldschool ?

```ts
typeof serverRumEvent.date === 'number' && serverRumEvent.date === Math.floor(serverRumEvent.date)
```

bcaudan: Damn! nice catch 🙂

I think a typeof check should be enough here

BenoitZugmeyer (approved):

PR #882, opened 2021-06-04
👷 prevent release from main branch

bcaudan:
## Motivation

Add early check to prevent release from main branch by mistake.

## Changes

- introduce a `release` cli command
- check current branch with `git branch --show-current` (git >= 2.22)

## Testing

manual

---

### DIFF
--- a/scripts/cli
+++ b/scripts/cli
@@ -50,6 +50,11 @@ cmd_build_json2type () {
   set -e
 }
 
+cmd_release () {
+  [[ `git branch --show-current` != "main" ]] || fail 'please do not release from `main` branch'
+  yarn lerna version --exact
+}
+
 cmd_version () {
   node ./scripts/generate-changelog.js
   # keep test app lockfile up to date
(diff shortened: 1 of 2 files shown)


scripts/cli
```diff
@@ -50,6 +50,11 @@ cmd_build_json2type () {
   set -e
 }
 
+cmd_release () {
+  [[ `git branch --show-current` != "main" ]] || fail 'please do not release from `main` branch'
```
BenoitZugmeyer, line 54: Maybe we could be even more restrictive:

```suggestion
  ! [[ `git branch --show-current` =~ ^release/v ]] || fail 'please release from a `release/vX.Y.Z` branch'
```

bcaudan: hum, the idea here is more to avoid to push something to main since then we need to remove it and that is a bit tedious with protection rules.
I don't think there is a great value to check the naming of the branch for the release. If we want to enforce the naming, I'd be in favor of letting the command create the branch and probably perform some extra checks.
That could be an improvement for another day.

BenoitZugmeyer: Alright. I agree that having a command to automate a bit the release would be nice.

BenoitZugmeyer (approved):

PR #883, opened 2021-06-04
✨ [RUMF-909] add beforeSend context

BenoitZugmeyer:
## Motivation

Now that `beforeSend` can edit the event context, it would be useful to have access on a bit of information about the browser data that have been used to generate the event.

## Changes

For each kind of RUM event, expose a domain-related 'context' as a second argument of `beforeSend`.

## Testing

Manual, unit

---


packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -209,3 +209,31 @@ export interface CommonContext {
...
+
+export type RumEventDomainContext<E extends RawRumEvent> = E extends RawRumErrorEvent
+  ? { error?: unknown }
+  : E extends RawRumActionEvent
+  ? { event?: Event }
+  : E extends RawRumLongTaskEvent
+  ? { performanceEntry: PerformanceEntry }
+  : E extends RawRumResourceEvent
+  ? {
+      performanceEntry?: PerformanceEntry
```
BenoitZugmeyer, line 233: Some of them are not actual `PerformenceEntry` objects, for example on `initial_document` and XHR/fetch requests. I wonder if it would be worth it to implement a proper type for it?

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -209,3 +209,31 @@ export interface CommonContext {
...
+      xhr?: XMLHttpRequest
+      response?: Response
```
BenoitZugmeyer, line 235: I didn't expose the `Request` object in case of `fetch`, because it can't be accessed if it not passed as first argument of the `fetch` function, so this would be rather limited. Also, the most interesting infos of the `Request` object is already in the RUM event. But I could add it if we fill like it would be a good addition.

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -209,3 +209,31 @@ export interface CommonContext {
...
+  readonly pathname: string
+  readonly port: string
+  readonly protocol: string
+  readonly search: string
+}
+
+export type RumEventDomainContext<E extends RawRumEvent> = E extends RawRumErrorEvent
+  ? { error?: unknown }
+  : E extends RawRumActionEvent
+  ? { event?: Event }
```
BenoitZugmeyer, line 228: I didn't add the action name (as initially planned in the RFC) since it is already in the RUM event.

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -76,12 +84,18 @@ function processResourceEntry(entry: RumPerformanceResourceTiming) {
...
     },
     tracingInfo,
     entryMetrics
   )
-  return { startTime: startClocks.relative, rawRumEvent: resourceEvent as RawRumResourceEvent }
+  return {
+    startTime: startClocks.relative,
+    rawRumEvent: resourceEvent,
+    domainContext: {
+      performanceEntry: (entry as unknown) as PerformanceEntry,
```
BenoitZugmeyer, line 96: This is not a good cast, since for `initial_document`, `entry` is not a PerformanceEntry

packages/rum-core/src/domain/rumEventsCollection/longTask/longTaskCollection.ts
```diff
@@ -19,7 +19,7 @@ export function startLongTaskCollection(lifeCycle: LifeCycle) {
     lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, {
       rawRumEvent,
       startTime: startClocks.relative,
-      domainContext: {},
+      domainContext: { performanceEntry: (entry as unknown) as PerformanceEntry },
```
amortemousque, line 22: Is there a way to avoid this cast?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -54,15 +54,23 @@ function processRequest(request: RequestCompleteEvent) {
...
     },
     tracingInfo,
     correspondingTimingOverrides
   )
-  return { startTime: startClocks.relative, rawRumEvent: resourceEvent as RawRumResourceEvent }
+  return {
+    startTime: startClocks.relative,
+    rawRumEvent: resourceEvent,
+    domainContext: {
+      performanceEntry: matchingTiming as PerformanceEntry | undefined,
```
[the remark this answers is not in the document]
amortemousque: May I ask, why it's likely to be the `toJSON` representation?

packages/rum-core/src/rawRumEvent.types.ts
amortemousque: Why not create a request object out of the user fetch proxy params?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -256,3 +256,17 @@ function sanitizeTiming(name: string) {
   }
   return sanitized
 }
+
+function cloneLocation(location: Location): ReadonlyLocation {
```
amortemousque, line 260: I'm wondering if "clone" has the right meaning in this case.

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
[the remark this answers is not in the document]
BenoitZugmeyer: You can see the difference with the following on any page:

```ts
console.log(performance.getEntriesByType('resource')[0]):
console.log(performance.getEntriesByType('resource')[0].toJSON()):
```

I think the two noticeable things are:

* the `toJSON` method is not in the `toJSON` representation
* `entry instanceof PerformanceEntry` will return false

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: I think that could works, yeah. It could simplify the [tracer](https://github.com/DataDog/browser-sdk/blob/ca9e9d3998d990fbc0e5829b3318335fc855fd80/packages/rum-core/src/domain/tracing/tracer.ts#L31-L53) too, if we always have a `Request` object instead of various kind of `input/init`.

Maybe we should try this in a separate PR, since this one is wide enough.

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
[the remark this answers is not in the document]
amortemousque: Then I would agree on having another type called "JsonPerformanceEntry" or "PerformanceEntryRepresentation".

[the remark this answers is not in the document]
BenoitZugmeyer: I think it could do the trick (in fact, I tried that in a previous version of the PR), but I wanted to have your opinion first. I'm a bit mitigated because I don't think it is our job to provide native browser types (that's a bit cumbersome to maintain).

We could re-use our [custom `PerformanceEntry` types](https://github.com/DataDog/browser-sdk/blob/ca9e9d3998d990fbc0e5829b3318335fc855fd80/packages/rum-core/src/browser/performanceCollection.ts#L20-L85) types and expose them publicly (we introduced them to ease the mock of those objects, as well as adding the possibility to create custom ones like `initial_document` resources). But we'll need to consolidate those types (there is a few missing properties), and chose what to do for the unofficial `traceId` prop)

packages/rum-core/src/domain/rumEventsCollection/longTask/longTaskCollection.ts
BenoitZugmeyer: See my answer there: https://github.com/DataDog/browser-sdk/pull/883#discussion_r646412570
We could if we exposed the `RumPerformanceEntry` types publicly.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
BenoitZugmeyer: What do you think about `createReadonlyLocation` then?

amortemousque: yes, nice

packages/rum-core/src/rawRumEvent.types.ts
bcaudan: a couple of suggestions:
- reflect in the naming that this is related to fetch
- we should also handle the `response: Error` case
- what about exposing `{ input: RequestInfo, init?: RequestInit, response: Response | Error  }`?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -166,7 +166,7 @@ function newView(
   const customTimings: ViewCustomTimings = {}
   let documentVersion = 0
   let endClocks: ClocksState | undefined
-  let location: Location = { ...initialLocation }
+  let location: ReadonlyLocation = cloneLocation(initialLocation)
```
bcaudan, line 169: What was the issue with the previous approach?
Could we keep it and directly use `Readonly<Location>`
it would avoid to explicitly list properties

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: Ok to expose input/init. Ok to expose `error` but I'd like to use a different field if that's ok :)

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
BenoitZugmeyer: The issue I tried to avoid was exposing `location` methods, because they can't be used (they throw an error):

```js
({ ...location }).reload()
// Uncaught TypeError: Illegal invocation
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -16,13 +16,13 @@ import {
...
 import { ProvidedSource } from '../domain/rumEventsCollection/error/errorCollection'
-import { CommonContext, User, ActionType } from '../rawRumEvent.types'
+import { CommonContext, User, ActionType, RumEventDomainContext } from '../rawRumEvent.types'
 import { RumEvent } from '../rumEvent.types'
 import { startRum } from './rum'
 
 export interface RumUserConfiguration extends UserConfiguration {
   applicationId: string
-  beforeSend?: (event: RumEvent) => void | boolean
+  beforeSend?: (event: RumEvent, context: RumEventDomainContext<any>) => void | boolean
```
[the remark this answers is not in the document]
BenoitZugmeyer: I implemented your solution. After a few tests, this seems a bit less strict (TS is confused with large type unions), but maybe a bit easier to understand.

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: Implemented as `{ fetchResponse?: Response, fetchInput: RequestInfo, fetchInit?: RequestInit, fetchError?: Error }`

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: if we make clear that this is not the actual location, throwing an error seems an OK behavior to me

packages/rum-core/src/rawRumEvent.types.ts
bcaudan: What is your reasoning for splitting response and error? 
IMO keeping it close to browser behavior could be easier to understand where to get the data

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -209,3 +209,58 @@ export interface CommonContext {
...
+export interface RumActionEventDomainContext {
+  event?: Event
+}
+export interface RumFetchResourceEventDomainContext {
+  fetchInit?: RequestInit
+  fetchInput: RequestInfo
+  fetchResponse?: Response
+  fetchError?: Error
+  performanceEntry?: PerformanceEntryRepresentation
+}
```
bcaudan, line 234: IMO, as it is not mixed with xhr anymore, fetch prefix could be omitted

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: It is not the browser behavior. Our implementation receives a value that is either an error or a response, but that's because we use the same function to handle both successful and failed calls. I think having separate properties makes it easier to use, because we can simply ducktype instead of using `instanceof`.

bcaudan: oh indeed, great then!

packages/rum-core/src/boot/rumPublicApi.ts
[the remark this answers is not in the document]
bcaudan: what about something like:
```ts
export type RawRumEventCollectedData<
  Event = RawRumEvent, 
  Domain = RumEventDomainContext
> = {
  startTime: RelativeTime
  savedCommonContext?: CommonContext
  customerContext?: Context
  rawRumEvent: Event
  domainContext: Domain
}
```

It could allow us to specify the right types when needed and could have a better TS support.

packages/rum-core/src/domain/rumEventsCollection/longTask/longTaskCollection.ts
```diff
@@ -19,7 +19,7 @@ export function startLongTaskCollection(lifeCycle: LifeCycle) {
     lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, {
       rawRumEvent,
       startTime: startClocks.relative,
-      domainContext: { performanceEntry: (entry as unknown) as PerformanceEntry },
+      domainContext: { performanceEntry: entry instanceof PerformanceEntry ? entry.toJSON() : entry },
```
bcaudan, line 22: Could we receive different representation for long tasks?

packages/rum-core/src/boot/rumPublicApi.ts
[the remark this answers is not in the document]
BenoitZugmeyer: I think my initial solution was superior, then: it was the same, but with the domain context automatically inferred from the event type.

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: From a TS perspective, it is now clearer that we have a "fetch" context. But from a JS perspective, everything is still mixed together, so I think it's nice to have prefixes?

packages/rum-core/src/domain/rumEventsCollection/longTask/longTaskCollection.ts
BenoitZugmeyer: Yes, during tests we mock performence entries so they don't have a `toJSON` method. Do you think we should implement a `toJSON` method for `RumPerformanceLongTaskTiming`?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
BenoitZugmeyer: Alright, I changed it back

packages/rum-core/src/domain/rumEventsCollection/longTask/longTaskCollection.ts
bcaudan: oh ok, then I'd be in favor of having the test code closer to the prod behavior rather than supporting test behavior in prod code

packages/rum-core/src/boot/rumPublicApi.ts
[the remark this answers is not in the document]
bcaudan: In the initial approach, DomainContext types were not easy to discover or reuse for customers.

what I had in mind was for customers to be able to import those DomainContext types in their beforeSend implementation, something like:
```
beforeSend(event, domainContext) {
  if (event.type === 'error') {
    const errorContext = domainContext as RumErrorDomainContext
    ...
  } 
}
```

with the proposed approach, we could make types a bit more strict and declarative, like in `resourceCollection`:
```ts
function processRequest(request: RequestCompleteEvent): RawRumEventCollectedData<RawRumResourceEvent, RumFetchDomainContext | RumXhrDomainContext>
```
which would raise a type error with the current implementation

wdyt?

packages/rum-core/src/rawRumEvent.types.ts
bcaudan: could we avoid to have things mixed up?

packages/rum-core/src/boot/rumPublicApi.ts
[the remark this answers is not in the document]
BenoitZugmeyer: We can totally keep both the individual domain interfaces and the inference upon the RUM event. I'll make something.

packages/rum-core/src/domain/rumEventsCollection/longTask/longTaskCollection.ts
BenoitZugmeyer: Ok I'll add the `toJSON` method to `RumPerformanceLongTaskTiming` then

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: What do you have in mind? Something like this?

```ts
{
  longtask?: { performanceEntry: PerformanceEntryRepresentation },
  error?: { error: unknown },
  fetchResource?: { input: RequentInfo, init: RequestInit, response?: Response, error?: Error, performanceEntry?: PerformanceEntryRepresentation },
  xhrResource?: { xhr: XMLHttpRequest, performanceEntry?: PerformanceEntryRepresentation },
  otherResource?: { performanceEntry: PerformanceEntryRepresentation },
  autoAction?: { event: Event },
  view?: { location: Readonly<Location> }
}
```

bcaudan (approved):
Great work 🎉

PR #889, opened 2021-06-10
✨ [RUMF-922] stack trace on handled calls

amortemousque:
## Motivation

Some errors are reported without a proper Error object. In these cases, there is no stack trace reported, making them a lot less actionable by the user. In order to enrich these errors, RUM need to generate a stack trace internally.

## Changes

- Generate an handling stack on console.log
- Generate an handling stack on addError

## Testing

Unit, Locally

---


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -152,21 +154,25 @@ export function makeRumPublicApi<C extends RumUserConfiguration>(startRumImpl: S
...
-        display.error(`DD_RUM.addError: Invalid source '${source as string}'`)
-        checkedSource = ErrorSource.CUSTOM
-      }
-      addErrorStrategy({
-        error,
-        context: deepClone(context as Context),
-        source: checkedSource,
-        startClocks: clocksNow(),
+    addError: (error: unknown, context?: object, source: ProvidedSource = ErrorSource.CUSTOM) => {
+      const instrumentationError = createInstrumentationError()
```
amortemousque, line 158: I wanted to put as less domain logic as possible here. 
`createInstrumentationError` is only here to ensure IE compatibility. 
Could also be a util function called `createErrorWithCallstack`

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.spec.ts
```diff
@@ -107,12 +112,34 @@ describe('error collection', () => {
...
+      function publicApiAddError() {
+        const instrumentationError = createInstrumentationError()
+        addError({
+          error: new Error('foo'),
+          instrumentationError,
+          source: ErrorSource.CUSTOM,
+          startClocks: { relative: 1234 as RelativeTime, timeStamp: 123456789 as TimeStamp },
+        })
+      }
+      function triggerAddError() {
```
amortemousque, line 134: Could have use Immediately-invoked function or currying

amortemousque, line 134: Could have used Immediately-invoked function or currying.

packages/core/src/tools/error.spec.ts
```diff
@@ -71,3 +71,23 @@ describe('formatUnknownError', () => {
     expect(formatted.message).toEqual('Uncaught {"foo":"bar"}')
   })
 })
+
+describe('createHandlingStackTrace', () => {
```
bcaudan, line 75: Could we have a test with an actual handling stack expectation?

packages/core/src/tools/error.ts
```diff
@@ -34,18 +36,25 @@ export enum ErrorHandling {
...
+  stackTrace: StackTrace | undefined,
+  errorObject: any,
+  nonErrorPrefix: string,
+  handlingStack?: StackTrace
+) {
   if (!stackTrace || (stackTrace.message === undefined && !(errorObject instanceof Error))) {
     return {
       message: `${nonErrorPrefix} ${jsonStringify(errorObject)!}`,
       stack: 'No stack, consider using an instance of Error',
+      handlingStack: handlingStack ? toStackTraceString(handlingStack) : undefined,
```
bcaudan, line 49: From local test, it seems to be formatted as:
```
Error: 
  at func @ file:line:col
  ...
```
what about replacing `Error` by `Handled`?

packages/core/src/tools/error.ts
```diff
@@ -65,3 +74,26 @@ export function toStackTraceString(stack: StackTrace) {
 export function formatErrorMessage(stack: StackTrace) {
   return `${stack.name || 'Error'}: ${stack.message!}`
 }
+
+/** Has to be called at the utmost position of the RUM call stack */
```
bcaudan, line 78: what is the issue otherwise?

packages/core/src/tools/error.ts
```diff
@@ -65,3 +74,26 @@ export function toStackTraceString(stack: StackTrace) {
...
+export function createHandlingStackTrace(): StackTrace {
+  const instrumentationCallDepth = 2
```
bcaudan, line 80: why 2?

packages/core/src/tools/error.ts
```diff
@@ -65,3 +74,26 @@ export function toStackTraceString(stack: StackTrace) {
...
+      throw error
+    } catch (e) {
+      noop()
+    }
+  }
+
+  callMonitored(() => {
+    stackTrace = computeStackTrace(error)
+    stackTrace.stack = stackTrace.stack.slice(instrumentationCallDepth)
+  })
```
bcaudan, line 96: why do we need to wrap this part in a `callMonitored`?

packages/core/src/tools/error.ts
amortemousque: If we do that, should we also change "Error" by "Unhandled" for the other cases?

packages/core/src/tools/error.ts
amortemousque: 2 is number of calls we need to skip in order to reach the user stacktrace.
Here is the two scenarios to report the handling stack
```
addError  -> createHandlingStackTrace
```
or
```
console.log  -> createHandlingStackTrace
```

packages/core/src/tools/error.ts
amortemousque: We need to wrap it in `callMonitored` because `createHandlingStackTrace` is call outside of them in order for `const error = new Error()` to be at the upmost point of the RUM callstack.
https://github.com/DataDog/browser-sdk/blob/c7d3145240054d9dd115c205912e84c3f59f1339/packages/rum-core/src/boot/rumPublicApi.ts#L157-L159
and 
https://github.com/DataDog/browser-sdk/blob/c7d3145240054d9dd115c205912e84c3f59f1339/packages/core/src/domain/automaticErrorCollection.ts#L38-L40

packages/core/src/tools/error.spec.ts
amortemousque: What do you mean, by actual handling stack? Are these tests more what you're looking for?
https://github.com/DataDog/browser-sdk/blob/c7d3145240054d9dd115c205912e84c3f59f1339/packages/rum-core/src/boot/rumPublicApi.spec.ts#L283-L290


https://github.com/DataDog/browser-sdk/blob/c7d3145240054d9dd115c205912e84c3f59f1339/packages/core/src/domain/automaticErrorCollection.spec.ts#L51-L59

packages/core/src/tools/error.ts
amortemousque: `createHandlingStackTrace` go back in the stacktrace to only keep user calls. 
If we move the createHandlingStackTrace  function call, then the calculation will be wrong and we will have RUM internal calls inside the `handling stacktrace`.

packages/core/src/tools/error.spec.ts
bcaudan: I'd say something like:
https://github.com/DataDog/browser-sdk/blob/fb7e9a1e4549b1d2959413733a22c45cc2ecb3e8/packages/core/src/tools/error.spec.ts#L40-L43
but for handling stack

packages/core/src/tools/error.ts
bcaudan: For `Unhandled`, we don't have an `handling_stack`, right?
Not sure to get what you have in mind

packages/core/src/tools/error.ts
bcaudan: Could it be doable/interesting to be able to call `createHandlingStackTrace` from anywhere and remove RUM internal  calls ourself?

packages/core/src/tools/error.ts
bcaudan: if we stick with this strategy (cf https://github.com/DataDog/browser-sdk/pull/889#discussion_r651768404), I'd be in favor of:
- clarify the strategy in the function comment
- rename this variable in something like `internalFrameToSkip`

packages/core/src/tools/error.ts
bcaudan: OK makes sense, it could be nice to mention that in the function description

packages/core/src/tools/error.ts
amortemousque: I already tried it but I found no ways to identify if a call come from our code base or not.

- I tried looking at files name but they can change regarding how the user bundles is solution with RUM. 
- I tried looking at functions name, but they can be updated and/or have the same name twice in the stack. The dipper the handling stack is generated the more uncertainty we will have.
- Finally, the best solution I found, is to create the stack trace at the utmost point in our call stack and to remove the last call. This is the less fickle implementation which is also used in tracekit 
https://github.com/DataDog/browser-sdk/blob/c7d3145240054d9dd115c205912e84c3f59f1339/packages/core/src/domain/tracekit/computeStackTrace.ts#L573-L574

packages/core/src/tools/error.ts
amortemousque: I don't really understand the goal of this tag to make a decision.

packages/core/src/tools/error.ts
bcaudan: OK, I'd be in favor of adding some of this explanation in the function comment as well

packages/core/src/tools/error.ts
bcaudan: I thought that this is what will be displayed in the UI but it may worth to sync with error tracking to see what they need in this field.

packages/core/src/tools/error.spec.ts
amortemousque: The thing is, the handling stack includes karma files path and calls, that's why I used contains. I can try to use a regex for a more complete ​check if you think it worth it?

packages/core/src/tools/error.ts
amortemousque: Yes, I'll check with them

amortemousque: Maxime said that `Error:` is okay for him.

bcaudan: OK, did you check with someone from the error tracking app to ensure that the field structure is convenient for them?

packages/core/src/tools/error.spec.ts
```diff
@@ -71,3 +71,25 @@ describe('formatUnknownError', () => {
...
+    rumInstrumentationCall()
+  }
+  function userCallOne() {
+    userCallTwo()
+  }
+
+  it('should get handling stack trace without instrumental function calls', () => {
+    userCallOne()
+
+    const formattedStackTrace = toStackTraceString(stackTrace)
```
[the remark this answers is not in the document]
amortemousque: I'm not sure about it. I tried to be consistent with the way we handle the "normal" stack trace.
https://github.com/DataDog/browser-sdk/blob/1f8b7c6e258cd5cd14817c9cd55cd086f79a0af3/packages/core/src/domain/automaticErrorCollection.ts#L57-L65
Also, by looking at error.ts, it seems that the idea was to keep stack traces as  `StackTrace` type and to be able to make transformation on them.
https://github.com/DataDog/browser-sdk/blob/1f8b7c6e258cd5cd14817c9cd55cd086f79a0af3/packages/core/src/tools/error.ts#L39-L43

Maybe I should change the test to only test the stack trace as `StackTrace` type and not as string. It would also make sens because the `toStackTraceString` has already been tested.

packages/core/src/domain/automaticErrorCollection.spec.ts
```diff
@@ -42,17 +42,29 @@ describe('console tracker', () => {
...
   })
 
+  it('should generate an instrumentation stack', () => {
+    function triggerError() {
+      console.error('foo', 'bar')
+    }
+    triggerError()
+    const rawError = notifyError.calls.mostRecent().args[0] as RawError
+    expect(rawError.handlingStack).toContain('triggerError')
+    expect(rawError.handlingStack).not.toContain('console.error')
```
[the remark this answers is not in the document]
amortemousque: yes, sure

packages/core/src/tools/error.ts
```diff
@@ -65,3 +74,38 @@ export function toStackTraceString(stack: StackTrace) {
...
+ Example:
+  ```javascript
+  function utmostFunction() {
+      const handlingStack = createHandlingStackTrace()
+      callMonitored(() => {})
+  }
+  ```
+ */
+export function createHandlingStackTrace(): StackTrace {
+  const internalFrameToSkip = 2 // Skip the two RUM internal frames in order to keep only the user calls
```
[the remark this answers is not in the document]
amortemousque: Yes

packages/core/src/tools/error.ts
```diff
@@ -65,3 +74,38 @@ export function toStackTraceString(stack: StackTrace) {
 export function formatErrorMessage(stack: StackTrace) {
   return `${stack.name || 'Error'}: ${stack.message!}`
 }
+
+/**
+ Creates a stacktrace starting from the previous call and working down.
+ - Has to be called at the utmost position of the RUM call stack.
+ - No internal monitoring should encapsulate the function, that is why we need to use callMonitored inside of it.
```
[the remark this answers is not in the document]
amortemousque: Yes

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -278,6 +280,15 @@ describe('rum public api', () => {
...
     })
 
+    it('should generate an stacktrace at the upmost position of RUM call stack', () => {
+      rumPublicApi.init(DEFAULT_INIT_CONFIGURATION)
+      rumPublicApi.addError(new Error('message'))
+
+      expect(addErrorSpy).toHaveBeenCalledTimes(1)
+      const stacktrace = toStackTraceString(addErrorSpy.calls.argsFor(0)[0].handlingStack)
+      expect(stacktrace).not.toContain('addError')
+    })
```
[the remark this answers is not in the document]
amortemousque: yes

packages/core/src/tools/error.ts
amortemousque: I've check with Eloy and I quote " if its a JS stack trace all that matters is that it has the same format and keywords as normal stack traces"

packages/core/src/tools/error.spec.ts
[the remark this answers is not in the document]
bcaudan: `StackTrace` is a tracekit type, it seems ok to me to keep it for tracekit API output but I'd be in favor of avoiding to use it internally if we can.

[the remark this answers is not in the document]
amortemousque: Okay done

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.spec.ts
```diff
@@ -46,6 +48,7 @@ describe('error collection', () => {
             resource: undefined,
             source: ErrorSource.CUSTOM,
             stack: jasmine.stringMatching('Error: foo'),
+            handling_stack: jasmine.any(String),
```
bcaudan, line 51:
```suggestion
            handling_stack: 'Error: handling foo',
```

packages/core/src/tools/error.spec.ts
```diff
@@ -71,3 +71,24 @@ describe('formatUnknownError', () => {
     expect(formatted.message).toEqual('Uncaught {"foo":"bar"}')
   })
 })
+
+describe('createHandlingStackTrace', () => {
+  let stackTrace: string
+  function internalCall() {
+    stackTrace = createHandlingStackTrace()
```
bcaudan, line 78:
```suggestion
    handlingStack = createHandlingStack()
```

it could be nice to always refer to it as "handling stack" and remove all the "trace" references

bcaudan (approved):

PR #897, opened 2021-06-17
✨ [PROF-3535] Expose init configuration via `getInitConfiguration()` API

piotr-oles:
## Motivation
I'm developing the front-end profiler in synthetics. I want to tag profiles with `service`, `version`, and `env` tags. 
Currently, they are passed as a user configuration to the RUM and there is no API to read this configuration.

## Changes
 * Add `getInitConfiguration()` method to DD_RUM and DD_LOG to read init configuration outside of the RUM context.
 * Re-implement `mergeInto`, `cloneDeep` and `combine` methods to not be tied to `Context` so I can use it for cloning init configuration

## Testing
I added unit tests :) I performed manual tests too.


packages/core/src/domain/configuration.ts
```diff
@@ -68,6 +68,8 @@ export interface UserConfiguration {
   replica?: ReplicaUserConfiguration
 }
 
+export type PublicUserConfiguration = Pick<UserConfiguration, 'service' | 'env' | 'version'>
```
bcaudan, line 71: > I intentionally published only a subset of the user configuration - otherwise, a third-party script could mutate user configuration which could have some security implications 🤔

Accessing the whole user configuration could probably be helpful for other uses cases, what about returning a copy of it to avoid any mutation?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -84,4 +84,15 @@ describe('logs', () => {
       expect(unreachableRequest.http.status_code).toEqual(0)
       expect(unreachableRequest.error!.stack).toContain('TypeError')
     })
+
+  createTest('returns user configuration')
```
bcaudan, line 88: nice intention, however I'd say that for this kind of update unit test should be enough.
e2e tests (wdio) are more time expensive so we tend to only add some when unit/integration tests (karma) are not enough

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -140,6 +141,35 @@ describe('rum public api', () => {
     })
   })
 
+  describe('getUserConfiguration', () => {
```
bcaudan, line 144: could we add the same kind of test for logs?

packages/core/src/domain/configuration.ts
piotr-oles: Do you mean deep clone? Because there are some complex values, like arrays, regexp, objects.
In my opinion it make sense to limit API as much as possible and extend it when there is a request for it. Adding new fields to this type will not be a breaking change, but removing will be :)

bcaudan: > Do you mean deep clone? Because there are some complex values, like arrays, regexp, objects.

👍

> In my opinion it make sense to limit API as much as possible and extend it when there is a request for it. Adding new fields to this type will not be a breaking change, but removing will be :)

I get your point but accessing all values would already be helpful while debugging (easier than find where the SDK is init) or if we want to retrieve the whole configuration from a browser extension.

piotr-oles: I was trying to deepClone using built-in function from `context.ts` but unfortunately it doesn't support RegExp and functions which can be a part of the init object. Should we omit these values or just copy references to them?
It also reports a type-error even if InitConfiguration is compatible with `ContextValue`. This issue can be solved by upgrading TypeScript to at least 4.2.3.

bcaudan: >  Should we omit these values or just copy references to them?

no way to clone them? (IE11 compatibility)
otherwise, copy the reference should be ok IMO

> It also reports a type-error even if InitConfiguration is compatible with ContextValue.

Do you have more details on that?

piotr-oles: image
I was wrong - upgrading TypeScript didn't help :/ It seems that TypeScript is missing index signature from the `InitConfiguration` type

bcaudan: if we extend `deepClone` to handle (somehow) other types than the those supported by `Context`, it could worth removing the type limitation and move `deepClone` to `core/src/tools/utils`.

piotr-oles: I updated the PR with generic deepClone, mergeInto and combine :)

packages/core/src/index.ts
```diff
@@ -41,7 +42,7 @@ export * from './tools/timeUtils'
 export * from './tools/utils'
 export * from './tools/errorFilter'
 export { ErrorSource, ErrorHandling, formatUnknownError, createHandlingStack, RawError } from './tools/error'
-export { combine, Context, ContextArray, ContextValue, deepClone } from './tools/context'
```
piotr-oles, line 44: `combine` and `deepClone` are exported in `export * from './tools/utils'`

packages/core/src/tools/utils.spec.ts
```diff
@@ -317,3 +327,189 @@ describe('findCommaSeparatedValue', () => {
...
+    expect(deepClone('test')).toBe('test')
+    expect(deepClone(true)).toBe(true)
+    expect(deepClone(null)).toBe(null)
+    const fn = () => null
+    expect(deepClone(fn)).toBe(fn)
+    expect(deepClone(1)).toBe(1)
+    expect(deepClone(undefined)).toBe(undefined)
+  })
+
+  it('clones array', () => {
```
bcaudan, line 445: Could we keep some explanation of the behavior in the test description?
`clones array` VS `should return different references for arrays items`

packages/core/src/tools/utils.spec.ts
```diff
@@ -317,3 +327,189 @@ describe('findCommaSeparatedValue', () => {
...
+  })
+
+  it('should not keep references on arrays', () => {
+    const source = { a: [1] }
+    const result = combine({}, source)
+    expect(result.a).not.toBe(source.a)
+  })
+})
+
+describe('deepClone', () => {
```
bcaudan, line 434: how does it behave for other types of objects?
like function for example

piotr-oles: For functions, we don't clone - we just copy a reference. It's tested a few lines below (line 440). 
This implementation support all types that were supported by the previous implementation. There will be no breaking change.

That means:
 - we don't support `Set` and `Map` - I didn't add support for these types because we don't use them, and it would increase bundle size. 
 - we don't support custom classes - we don't clone prototype chain as it's non-trivial, and it seems we don't use it

bcaudan: ok, got it.
as mentioned in https://github.com/DataDog/browser-sdk/pull/897/files#r663744326, it could be nice to have a more expressive test for this one, maybe also splitting primitive handling from other object types.

piotr-oles: sure, I will update test descriptions :)

packages/core/src/tools/utils.ts
```diff
@@ -388,6 +388,148 @@ export function runOnReadyState(expectedReadyState: 'complete' | 'interactive',
...
+        source.sticky ? 'y' : '',
+        source.unicode ? 'u' : '',
+      ].join('')
+    return (new RegExp(source.source, flags) as unknown) as Merged<D, S>
+  }
+
+  const reference = references.get(source) as Merged<D, S> | undefined
+  if (reference) {
+    // handle circular references
+    return reference
```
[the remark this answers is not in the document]
piotr-oles: Good catch :) I will revert this behaviour change

packages/core/src/tools/utils.ts
```diff
@@ -388,6 +388,148 @@ export function runOnReadyState(expectedReadyState: 'complete' | 'interactive',
   }
 }
 
+type Merged<TDestination, TSource> =
```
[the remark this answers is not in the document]
piotr-oles: Yes, it should - but it will not be easy to do. I did a quick pass and it will also introduce some type errors that will require further changes in type annotations. Could we defer this change to a separate PR?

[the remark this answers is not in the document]
BenoitZugmeyer: Sure! That's pretty neat already :)

packages/core/src/tools/utils.spec.ts
```diff
@@ -317,3 +328,209 @@ describe('findCommaSeparatedValue', () => {
...
+  it('should pass-through non-objects and functions', () => {
```
bcaudan, line 458: as mentioned in [previous comment](https://github.com/DataDog/browser-sdk/pull/897/files#r663764155), what about splitting this test in:
- it should pass-through primitives
- it should pass-through other object types, could be at the end of the suite with function and a couple of other examples

bcaudan (approved):

BenoitZugmeyer (approved):

PR #902, opened 2021-06-22
🐛 [REPLAY-325] Add verbose error details on XHR transport failure

jagracey:
## Motivation

Occasionally replay segments are being dropped. We have very little understanding of the circumstances of why certain requests are failing. Here we collect more information to log the circumstance- and implicitly hope the next network request succeeds.

JavaScript in the wild can be a funny place, with many different vendor patches and frameworks fighting with each other. We take a few unconventional approaches to cover all our bases around what is going on:
1. Use of try/catch with XHR open + send. This code _should_ not throw, but this is critical code so let's add ~200 bytes and see what happens.
 2. `event.isTrusted` Maybe a vendor is dispatching poorly crafted error events on XHR, so let's detect if this is a native event.
 
We can run this code in production for a while, see if it provides any unique insight. If no value is provided for certain properties, we can remove some code.

 
**Try/Catch:** - 
 - error message + stack trace
 
**request.onError:**
 - `event.isTrusted`
 - `event.total` - how many bytes does the browser natively report. Maybe something patched our XHR request resulting in a failure.
 
 **Both methods:**
  - online: does the browser report internet connectivity?
  - `try_beacon` did we at least try to use the beacon first?
  - `url` who's servers might be dropping packets?

 
 **Other approaches if we need to go further:**
  - track request duration: slow timeout vs instant fail
  - check if XHR has been patched
  - track in sessionStorage or localStorage each failed internal sendBeacon + XHR, and pickup these stats as needed.

## Changes
Just adds a `addMonitoringMessage(..)` to `packages/core/src/transport/transport.ts`

## Testing

Passes tests locally.

---

### DIFF
--- a/packages/core/src/transport/transport.ts
+++ b/packages/core/src/transport/transport.ts
@@ -7,25 +7,16 @@ import { monitor, addErrorToMonitoringBatch, addMonitoringMessage } from '../dom
 const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/
 
 /* eslint-disable camelcase */
-type ErrorCtx =
-  | {
-      err_msg?: string
-      err_stack?: string
-    }
-  | undefined
-
 type TransportErrorContext = {
-  error?: ErrorCtx
   event?: {
     type: string
     is_trusted: boolean
     total: number
     loaded: number
   }
-  req?: {
+  request?: {
     status: number
     ready_state: number
-    timeout: number
     response_text: string
   }
   on_line: boolean
@@ -61,31 +52,33 @@ export class HttpRequest {
       }
     }
 
-    const transportErrorHandler = (ctx: Partial<TransportErrorContext>) => {
-      const transportErrCtx: TransportErrorContext = {
+    const transportIntrospection = (event: ProgressEvent) => {
+      const req = event?.currentTarget as XMLHttpRequest
+      const xhrTransportContext: TransportErrorContext = {
         on_line: navigator.onLine,
         size,
         url,
         try_beacon: tryBeacon,
         bytes_limit: this.bytesLimit,
-        ...ctx,
+        event: {
+          type: event.type,
+          is_trusted: event.isTrusted,
+          total: event.total,
+          loaded: event.loaded,
+        },
+        request: {
+          status: req.status,
+          ready_state: req.readyState,
+          response_text: req.responseText,
+        },
       }
-      addMonitoringMessage('Internal XHR failed', transportErrCtx)
+      addMonitoringMessage('XHR fallback failed', xhrTransportContext)
     }
 
-    try {
-      const request = new XMLHttpRequest()
-      request.addEventListener('error', (e) => transportErrorHandler(getTransportErrorEventDetails(e)))
-      request.open('POST', url, true)
-      request.send(data)
-    } catch (err) {
-      transportErrorHandler({
-        error: {
-          err_msg: err.message,
-          err_stack: err.stack,
-        },
-      })
-    }
+    const request = new XMLHttpRequest()
+    request.addEventListener('loadend', (event) => transportIntrospection(event))
+    request.open('POST', url, true)
+    request.send(data)
   }
 }
 
@@ -253,26 +246,4 @@ export class Batch {
       addEventListener(window, DOM_EVENT.BEFORE_UNLOAD, () => this.flush())
     }
   }
-}
-
-function getTransportErrorEventDetails(event: ProgressEvent) {
-  const ctx: Partial<TransportErrorContext> = {}
-  if (event) {
-    ctx.event = {
-      type: event.type,
-      is_trusted: event.isTrusted,
-      total: event.total,
-      loaded: event.loaded,
-    }
-    if (event?.currentTarget) {
-      const req = event?.currentTarget as XMLHttpRequest
-      ctx.req = {
-        status: req.status,
-        ready_state: req.readyState,
-        timeout: req.timeout,
-        response_text: req.responseText,
-      }
-    }
-  }
-  return ctx
 }
--- a/packages/core/src/transport/transport.ts
+++ b/packages/core/src/transport/transport.ts
@@ -1,10 +1,40 @@
 import { display } from '../tools/display'
 import { Context } from '../tools/context'
 import { addEventListener, DOM_EVENT, jsonStringify, noop, objectValues } from '../tools/utils'
-import { monitor, addErrorToMonitoringBatch } from '../domain/internalMonitoring'
+import { monitor, addErrorToMonitoringBatch, addMonitoringMessage } from '../domain/internalMonitoring'
 
 // https://en.wikipedia.org/wiki/UTF-8
 const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/
+
+/* eslint-disable camelcase */
+type ErrorCtx =
+  | {
+      err_msg?: string
+      err_stack?: string
+    }
+  | undefined
+
+type TransportErrorContext = {
+  error?: ErrorCtx
+  event?: {
+    type: string
+    is_trusted: boolean
+    total: number
+    loaded: number
+  }
+  req?: {
+    status: number
+    ready_state: number
+    timeout: number
+    response_text: string
+  }
+  on_line: boolean
+  size: number
+  url: string
+  try_beacon: boolean
+  bytes_limit: number
+}
+/* eslint-enable camelcase */
 
 /**
  * Use POST request without content type to:
@@ -19,7 +49,8 @@ export class HttpRequest {
 
   send(data: string | FormData, size: number) {
     const url = this.withBatchTime ? addBatchTime(this.endpointUrl) : this.endpointUrl
-    if (navigator.sendBeacon && size < this.bytesLimit) {
+    const tryBeacon = !!navigator.sendBeacon && size < this.bytesLimit
+    if (tryBeacon) {
       try {
         const isQueued = navigator.sendBeacon(url, data)
         if (isQueued) {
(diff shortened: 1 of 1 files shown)


packages/core/src/transport/transport.ts
```diff
@@ -29,9 +64,32 @@ export class HttpRequest {
         reportBeaconError(e)
       }
     }
-    const request = new XMLHttpRequest()
-    request.open('POST', url, true)
-    request.send(data)
+
+    const transportErrorHandler = (ctx: any) => {
```
BenoitZugmeyer, line 68:
```suggestion
    const transportErrorHandler = (ctx: Partial<TransportErrorContext>) => {
```

packages/core/src/transport/transport.ts
```diff
@@ -29,9 +64,32 @@ export class HttpRequest {
...
+          bytes_limit: this.bytesLimit,
+        },
+        ...ctx,
+      }
+      addMonitoringMessage('Internal XHR failed', transportErrCtx)
+    }
+
+    try {
+      const request = new XMLHttpRequest()
+      request.addEventListener('error', (e) => transportErrorHandler(getTransportErrorEventDetails(e)))
```
BenoitZugmeyer, line 84: Beside listening to the “error” event, you should also listening to “load” and send a message if the status is >= 400 or the response is not `{}` , something like that. It could be useful to include `request.responseText` in the context too

packages/core/src/transport/transport.ts
```diff
@@ -1,11 +1,45 @@
...
+      type: evt?.type,
+      is_trusted: evt?.isTrusted,
+      total: evt?.total,
+      loaded: evt?.loaded,
+    },
+    req: {
+      status: req?.status,
+      ready_state: req?.readyState,
+      timeout: req?.timeout,
+    },
```
BenoitZugmeyer, line 39: I don't think all those `?.` are needed. If they are, why not returning early if there is no `evt` or `req`?

FYI: `foo?.bar` is transpiled to `foo === null || foo === void 0 ? void 0 : foo.bar;`, so adding unnecessary operators like this may produce a bit of extra code in the final bundle.

packages/core/src/transport/transport.ts
```diff
@@ -1,11 +1,45 @@
...
+type TransportErrorCtx = {
+  beacon: Record<string, string | number | boolean>
+  event?: Record<string, string | number | boolean>
+  req?: Record<string, string | number | boolean>
+  error_ctx?: ErrorCtx
+  on_line: boolean
+}
+/* eslint-enable camelcase */
+
+const getTransportErrorEventDetails = (evt: ProgressEvent) => {
```
BenoitZugmeyer, line 26: We tend to put the most important functions (ex: the exported functions) at the top of modules. Could you move this one a bit below?

packages/core/src/transport/transport.ts
```diff
@@ -29,9 +64,32 @@ export class HttpRequest {
         reportBeaconError(e)
```
BenoitZugmeyer, line 64: It may be interesting to use the same "transport error context" for beacon errors too.

BenoitZugmeyer (commented):
Because adding monitoring logs also uses this piece of code to send the log, it could loop forever if it keeps failing. This is why we only send the [first log when sendBeacon is throwing](https://github.com/DataDog/browser-sdk/blob/ee5717a87aed6cd38184c23dedb51a8c7a04536a/packages/core/src/transport/transport.ts#L100-L106).

We should do the same for XHR errors. We may reuse the `reportBeaconError` function  (by giving it a proper name), so only a single transport error is sent per page load.

packages/core/src/transport/transport.ts
jagracey: Good idea. Let's see how this works out first?

packages/core/src/transport/transport.ts
jagracey: Good idea. Let's see how this works out first?

packages/core/src/transport/transport.ts
jagracey: As needed, we can add to the XHR (and something similar to sendBeacon)
- onTimeout
- onLoad
- onAbort

BenoitZugmeyer: According to the Jira task, we want to make it easier to investigate issues where the intake rejects the request we are sending because it is too large. So, your PR does not fit our current needs: when the intake rejects the request, the XHR succeeds but the response will have an invalid status code and/or an error message in the body I guess. So we’ll have to hook into the “onload” event.

`ontimeout`, `onabort`, `onerror` and `try/catch` are nice to have but a bit out of scope for this PR. I suggest to keep what you did (`onerror` and `try/catch`), add a check on the response via `onload`. That should be enough for now. Does it make sense?

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,32 @@ export class HttpRequest {
...
       }
     }
+
+    const transportIntrospection = (event: ProgressEvent) => {
+      const req = event?.currentTarget as XMLHttpRequest
```
BenoitZugmeyer, line 56: `transportIntrospection` is called unconditionally during `loadend`, even if the request did succeed. You should add a check here to skip sending the failure message if it succeeded (check on request status and maybe response text

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,32 @@ export class HttpRequest {
...
+          status: req.status,
+          ready_state: req.readyState,
+          response_text: req.responseText,
+        },
+      }
+      addMonitoringMessage('XHR fallback failed', xhrTransportContext)
+    }
+
     const request = new XMLHttpRequest()
+    request.addEventListener('loadend', (event) => transportIntrospection(event))
```
BenoitZugmeyer, line 79: We usually wrap all asynchronous functions in a `monitor`, but here it may cause more harm than good since it may lead to an infinite loop if sending the failure message keeps failing the same way.

So I think it's fine (@bcaudan wdyt?), but we might want to add a comment about it just in case

bcaudan: Should not we report only the first xhr error? like for sendBeacon

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,41 @@ export class HttpRequest {
...
+        return
+      }
+      const xhrTransportContext: TransportIssueContext = {
+        on_line: navigator.onLine,
+        size,
+        url,
+        try_beacon: tryBeacon,
+        bytes_limit: this.bytesLimit,
+        event: {
+          type: event.type,
```
BenoitZugmeyer, line 67: NIT: Since we listen to a signle event, the event type should always be the same (`loadend`). I think we could remove this.

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,41 @@ export class HttpRequest {
         reportBeaconError(e)
       }
     }
+
+    const transportIntrospection = (event: ProgressEvent) => {
+      const req = event?.currentTarget as XMLHttpRequest
+      if (req.status >= 200 && req.status < 300) {
+        return
+      }
+      const xhrTransportContext: TransportIssueContext = {
```
BenoitZugmeyer, line 60: NIT: Maybe we could compute the context only when actually sending the message?

BenoitZugmeyer (approved):
LGTM

packages/core/src/transport/transport.ts
```diff
@@ -47,6 +103,14 @@ function reportBeaconError(e: unknown) {
   }
 }
 
+let hasReportedXhrError = false
+function reportXhrIssue(xhrTransportContext: TransportIssueContext) {
+  if (!hasReportedXhrError) {
+    hasReportedXhrError = true
+    addMonitoringMessage('XHR fallback failed', xhrTransportContext)
```
bcaudan, line 110: was there any thoughts on log level?
if not, I'd be in favor of being consistent with beacon and use `error` instead of `info`.

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,41 @@ export class HttpRequest {
...
+      }
+      reportXhrIssue(xhrTransportContext)
+    }
+
     const request = new XMLHttpRequest()
+    /*
+      Successful responses (200–299)
+    */
+    request.addEventListener(
+      'loadend',
```
bcaudan, line 86: is it the right place for this comment? IMO, it would be more useful close to the status check
not sure that we even need it though

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,41 @@ export class HttpRequest {
...
+      const req = event?.currentTarget as XMLHttpRequest
+      if (req.status >= 200 && req.status < 300) {
+        return
+      }
+      const xhrTransportContext: TransportIssueContext = {
+        on_line: navigator.onLine,
+        size,
+        url,
+        try_beacon: tryBeacon,
+        bytes_limit: this.bytesLimit,
```
bcaudan, line 65: FMU this value is a constant so maybe logging `size > this.bytesLimit` could be more helpful

packages/core/src/transport/transport.ts
```diff
@@ -29,7 +51,41 @@ export class HttpRequest {
...
+        event: {
+          type: event.type,
+          is_trusted: event.isTrusted,
+          total: event.total,
+          loaded: event.loaded,
+        },
+        request: {
+          status: req.status,
+          ready_state: req.readyState,
+          response_text: req.responseText,
```
bcaudan, line 75: I'd be in favor of truncating the response to avoid large payloads

packages/core/src/transport/transport.ts
jagracey: Removed now since `try_beacon` already has `size > this.bytesLimit`.

packages/core/src/transport/transport.ts
jagracey: `req.responseText.slice(0, 64)`

packages/core/src/transport/transport.ts
jagracey: Erroneous comment now removed.

packages/core/src/transport/transport.ts
jagracey: Was thinking there is no error to capture with a stack trace (name, message, stack). It's more of a warning to let the team know network requests are dropping. No actions are necessarily involved.

bcaudan: ok, got it.

bcaudan (approved):

PR #903, opened 2021-06-23
✨ allow users to customize the attribute used to define the action name

niieani:
## Motivation

`data-test-id` and `data-testid` are the de-facto standard for consistent naming of actionable elements, used for both unit and integration testing. Being forced to duplicate these same names and introduce a vendor-specific attribute (`data-dd-action-name`) is not ideal. 

The reason why a custom attribute is preferred to the default fallback mechanisms is that localized applications will contain strings in different languages, splitting the same action into multiple ones. Additionally, this helps prevent leaking of personal data in case the action is personalized to the content only visible to the user.

## Changes

Adds the option to customize the attribute name.

## Testing

Uses the same code paths that already exist, but nonetheless added test cases for these new scenarios.

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -82,6 +82,8 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
     service?: string
     beforeSend?: BeforeSendCallback
 
+    actionNameAttribute?: string
+
     isEnabled: (feature: string) => boolean
   }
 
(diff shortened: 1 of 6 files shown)


bcaudan:
Hi @niieani, thanks for submitting this!

Being able to reuse existing attributes to avoid duplication is really interesting, two concerns with the suggested approach:

- current usage of `data-test-id` and `data-testid` could change the data currently collected, so it would be a breaking change
- other customers may use different attributes

So instead of hardcoding these two new attributes, it could be interesting to introduce a new configuration option to provide another attribute name to use.

What do you think?

niieani:
Hi @bcaudan, updated and rebased the code based on your review. The attribute is now customizable, but still defaults to the attribute that works today, so it is not a breaking change.

bcaudan:
Thanks @niieani!

I took over your PR in #919  to ease the review / merge process.
we'll ping here when it will be available.

bcaudan:
Available in 2.16.0

PR #908, opened 2021-06-25
🐛 fix new view segment starting with an IncrementalSnapshot

BenoitZugmeyer:
## Motivation

Pending mutations were flushed when a FullSnapshot was taken, but in this case the new view was already active, so pending mutations were incorrectly attributed to the new view.

## Changes

To fix this issue, we should flush pending mutations when the view ends.

## Testing

Manual testing, unit test.

---

### DIFF
--- a/packages/rum-recorder/src/boot/startRecording.ts
+++ b/packages/rum-recorder/src/boot/startRecording.ts
@@ -25,10 +25,11 @@ export function startRecording(
     addRecord({ ...rawRecord, timestamp: Date.now() })
   }
 
-  const { stop: stopRecording, takeFullSnapshot } = record({
+  const { stop: stopRecording, takeFullSnapshot, flush } = record({
     emit: addRawRecord,
   })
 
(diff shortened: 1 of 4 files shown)


ThibautGeriz (approved):
Looks good.

webNeat (approved):
LGTM

packages/rum-recorder/src/boot/startRecording.ts
```diff
@@ -25,10 +25,11 @@ export function startRecording(
     addRecord({ ...rawRecord, timestamp: Date.now() })
   }
 
-  const { stop: stopRecording, takeFullSnapshot } = record({
+  const { stop: stopRecording, takeFullSnapshot, flush } = record({
```
bcaudan, line 28: what about making explicit the fact the we are flushing mutations?
to avoid any confusions with flush of segment

bcaudan (approved):

PR #910, opened 2021-06-28
🐛  Fix suffixed packages deployment

amortemousque:
## Motivation

Fix malformed suffixed package names
Add suffixed packages invalidation

## Changes

Update deployment script

## Testing

Locally

---

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -49,9 +49,9 @@ main() {
 upload-to-s3() {
     assume-role "build-stable-browser-agent-artifacts-s3-write"
     for file_path in "${FILE_PATHS[@]}"; do
-      local file_name=$(basename "$file_path")
+      local file_name=$(suffixed-file-name "$file_path")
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -60,11 +60,17 @@ invalidate-cloudfront() {
     echo "Creating invalidation"
     local -a paths_to_invalidate
     for file_path in "${FILE_PATHS[@]}"; do
-      paths_to_invalidate+=("/$(basename "$file_path")")
+      paths_to_invalidate+=($(suffixed-file-name "$file_path"))
```
BenoitZugmeyer, line 63: There was a slash at the beginning of the path, did it matter?

BenoitZugmeyer (approved):

scripts/deploy.sh
amortemousque: I'll double check

PR #912, opened 2021-06-29
Add segment real size

ThibautGeriz:
## Motivation

In order to concatenate the segment zip, we need to get the size of the segment uncompressed

The intake already add `segment_size_in_byte` field which is the size of the compressed segment

_Worth reviewing by commit_

## Changes

- add field `raw_segment_size`, it will be rename `raw_segment_size_in_byte` in the intake
- remove string encoding from pako and rely on `TextEncoder`

## Testing

---


packages/rum-recorder/src/domain/segmentCollection/deflateWorker.js
```diff
@@ -14,28 +14,35 @@ function workerCodeFn() {
...
     let deflate = new Deflate()
+    let sizeInBytes = 0
     self.addEventListener(
       'message',
       monitor((event) => {
         const data = event.data
         switch (data.action) {
           case 'write':
-            deflate.push(data.data, constants.Z_SYNC_FLUSH)
+            const binaryData = new TextEncoder().encode(data.data)
```
ThibautGeriz, line 24: Use `TextEncoder` instead of `Blob` because better supported even though it's slower than `Blob`

packages/rum-recorder/src/domain/segmentCollection/deflateWorker.js
```diff
@@ -14,28 +14,35 @@ function workerCodeFn() {
...
+            deflate.push(binaryData, constants.Z_SYNC_FLUSH)
+            sizeInBytes += binaryData.length
             self.postMessage({
               id: data.id,
               size: deflate.chunks.reduce((total, chunk) => total + chunk.length, 0),
```
BenoitZugmeyer, line 29: Could you rename this to `compressedSize`?

packages/rum-recorder/src/domain/segmentCollection/deflateWorker.js
```diff
@@ -14,28 +14,35 @@ function workerCodeFn() {
...
-              deflate.push(data.data, constants.Z_SYNC_FLUSH)
+              const binaryData = new TextEncoder().encode(data.data)
+              deflate.push(binaryData, constants.Z_SYNC_FLUSH)
+              sizeInBytes += binaryData.length
             }
             deflate.push('', constants.Z_FINISH)
             self.postMessage({
               id: data.id,
               result: deflate.result,
+              sizeInBytes,
```
BenoitZugmeyer, line 42: Could you rename this to `rawSize`?

packages/rum-recorder/src/domain/segmentCollection/segment.ts
```diff
@@ -19,7 +19,7 @@ export class Segment {
     private creationReason: CreationReason,
     initialRecord: Record,
     onWrote: (size: number) => void,
-    onFlushed: (data: Uint8Array) => void
+    onFlushed: (data: Uint8Array, sizeInBytes: number) => void
```
BenoitZugmeyer, line 22:
```suggestion
    onFlushed: (data: Uint8Array, rawSize: number) => void
```

packages/rum-recorder/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -145,8 +145,8 @@ export function doStartSegmentCollection(
           flushSegment('max_size')
         }
       },
-      (data) => {
-        send(data, segment.meta)
+      (data, rawSegmentSize) => {
+        send(data, { ...segment.meta }, rawSegmentSize)
```
BenoitZugmeyer, line 149:
```suggestion
        send(data, segment.meta, rawSegmentSize)
```

packages/rum-recorder/src/transport/send.ts
```diff
@@ -15,6 +15,7 @@ export function send(endpointUrl: string, data: Uint8Array, meta: SegmentMeta):
   )
 
   toFormEntries(meta, (key, value) => formData.set(key, value))
+  formData.set('raw_segment_size', segmentSize.toString(10))
```
BenoitZugmeyer, line 18: This is the default, let save 2 bytes :)
```suggestion
  formData.set('raw_segment_size', segmentSize.toString())
```

packages/rum-recorder/src/domain/segmentCollection/deflateWorker.d.ts
```diff
@@ -24,10 +24,11 @@ export type DeflateWorkerAction =
 export type DeflateWorkerResponse =
   | {
       id: number
-      size: number
+      compressedSize: number
     }
   | {
       id: number
       result: Uint8Array
+      rawSize: number
```
bcaudan, line 32: what about using `uncompressed` instead of `raw`?
it could more explicit

packages/rum-recorder/src/boot/startRecording.spec.ts
```diff
@@ -86,6 +86,7 @@ describe('startRecording', () => {
         segment: jasmine.any(File),
         'session.id': 'session-id',
         start: jasmine.stringMatching(/^\d{13}$/),
+        raw_segment_size: jasmine.stringMatching(/^\d+$/),
```
bcaudan, line 89: > add field raw_segment_size, it will be rename raw_segment_size_in_byte in the intake

what is the reasoning for renaming the field at the intake? could we name it correctly from the SDK?

packages/rum-recorder/src/domain/segmentCollection/deflateWorker.d.ts
ThibautGeriz: I am okay with both but I would like to settle on a name because I also have to modify the intake of the event platform

packages/rum-recorder/src/boot/startRecording.spec.ts
ThibautGeriz: At the beginning it was not but @BenoitZugmeyer did not like having the unit in the variable name and I believe it's useful when used across systems so I rename it in the intake. That way in dogweb I know what I an dealing with

webNeat: I am just curious, do we often rename fields on the intake or is this a rare exception?
In my opinion, it's better to have the same names, but if we are already doing this for many other fields then I think it's ok for this one too.

packages/rum-recorder/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -145,8 +145,8 @@ export function doStartSegmentCollection(
           flushSegment('max_size')
         }
       },
-      (data) => {
-        send(data, segment.meta)
+      (data, rawSegmentSize) => {
```
webNeat, line 148: I see that we are calling this same argument `segmentSize` above, but `rawSegmentSize` here. I think using the same name in all places would be clearer.

packages/rum-recorder/src/boot/startRecording.spec.ts
ThibautGeriz: Following discussion it's not renamed anymore. We keep: `raw_segment_size` across systems

packages/rum-recorder/src/domain/segmentCollection/deflateWorker.d.ts
ThibautGeriz: Following discussion, keeping `raw`

packages/rum-recorder/src/boot/startRecording.spec.ts
ThibautGeriz: > I am just curious, do we often rename fields on the intake or is this a rare exception?

No, that was me trying to be explicit.
The convention is to have the sizes in bytes so I will follow convention, no need to add the unit.

packages/rum-recorder/src/domain/segmentCollection/segmentCollection.ts
ThibautGeriz: Indeed, well caught.

webNeat (approved):

BenoitZugmeyer (approved):

PR #914, opened 2021-06-29
✨[REPLAY-336] Privacy by Default: MVP scaffold

jagracey:
## Motivation

Part one of a series of PRs aimed at implementing privacy by default (REPLAY-336).
- This PR adds `censorshipLevel` config property to `rum-core` for use with `rum-recorder`
- Creates an importable reference to the global config object
- Exposes new methods `isFlagEnabled` and `getCensorshipLevel`
- Adds new helper method `censorText` for HTML text censorship that preserves whitespace
- Defines basic constants for all further PRs to build upon: `PRIVACY_INPUT_MASK` and `FORM_PRIVATE_TAG_NAMES`.

## Testing

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -13,7 +13,7 @@ export const DEFAULT_CONFIGURATION = {
   silentMultipleInit: false,
   trackInteractions: false,
   trackViewsManually: false,
-  censorshipLevel: 'PRIVATE',
+  censorshipLevel: 'PUBLIC',
 
   /**
    * arbitrary value, byte precision not needed
@@ -140,6 +140,10 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
 
   if ('trackViewsManually' in userConfiguration) {
     configuration.trackViewsManually = !!userConfiguration.trackViewsManually
+  }
+
+  if ('censorshipLevel' in userConfiguration) {
+    configuration.censorshipLevel = userConfiguration.censorshipLevel ?? configuration.censorshipLevel
   }
 
   return configuration
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -2,6 +2,7 @@ import { BuildEnv } from '../boot/init'
 import { CookieOptions, getCurrentSite } from '../browser/cookie'
 import { catchUserErrors } from '../tools/catchUserErrors'
 import { includes, ONE_KILO_BYTE, ONE_SECOND } from '../tools/utils'
+import { CensorshipLevel } from '../../../../packages/rum-recorder/src/constants'
 import { computeTransportConfiguration, Datacenter } from './transportConfiguration'
 
 export const DEFAULT_CONFIGURATION = {
@@ -13,6 +14,7 @@ export const DEFAULT_CONFIGURATION = {
   silentMultipleInit: false,
   trackInteractions: false,
   trackViewsManually: false,
+  censorshipLevel: CensorshipLevel.PRIVATE,
 
   /**
    * arbitrary value, byte precision not needed
@@ -54,6 +56,7 @@ export interface UserConfiguration {
   trackViewsManually?: boolean
   proxyHost?: string
   beforeSend?: BeforeSendCallback
+  censorshipLevel?: CensorshipLevel
 
   service?: string
   env?: string
@@ -81,6 +84,7 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
 
     service?: string
     beforeSend?: BeforeSendCallback
+    censorshipLevel?: CensorshipLevel
 
     isEnabled: (feature: string) => boolean
   }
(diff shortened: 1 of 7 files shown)


packages/core/src/domain/configuration.ts
```diff
@@ -13,6 +13,7 @@ export const DEFAULT_CONFIGURATION = {
   silentMultipleInit: false,
   trackInteractions: false,
   trackViewsManually: false,
+  censorshipLevel: 'PUBLIC',
```
BenoitZugmeyer, line 16: * It could be interesting to pass recording options to the `startSessionReplayRecording` function instead of `init`, so they are scoped and can be changed every time the recording is restarted
* I'm not found of `censorship`: I think it is a bit negatively connoted. What about something like `privacyLevel` or `maskMode`?
* We should have an object constants (ex: similar to [`StatusType`](https://github.com/DataDog/browser-sdk/blob/5331c1f33324458e983a4c8a689b53f7131c1bde/packages/logs/src/domain/logger.ts#L12-L20)) to avoid hardcoding values

packages/core/src/domain/configuration.ts
```diff
@@ -139,6 +142,10 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
     configuration.trackViewsManually = !!userConfiguration.trackViewsManually
   }
 
+  if ('censorshipLevel' in userConfiguration) {
+    configuration.censorshipLevel = userConfiguration.censorshipLevel ?? configuration.censorshipLevel
```
BenoitZugmeyer, line 146: I don't think we should diverge from the current configuration handling here. Just keep whatever value the user provides.

```suggestion
    configuration.censorshipLevel = userConfiguration.censorshipLevel
```

packages/rum-recorder/src/domain/record/privacy.ts
```diff
@@ -89,3 +91,5 @@ function isElement(node: Node): node is Element {
 function isInputElement(elem: Element): elem is HTMLInputElement {
   return elem.tagName === 'INPUT'
 }
+
+export const censorText = (text: string) => text.replace(/[^\s]/g, MASKING_CHAR)
```
BenoitZugmeyer, line 95: This is equivalent:
```suggestion
export const censorText = (text: string) => text.replace(/\S/g, MASKING_CHAR)
```

packages/rum-recorder/src/domain/record/serializationUtils.ts
```diff
@@ -125,5 +126,27 @@ export function getElementInputValue(element: Element, ancestorInputPrivacyMode?
 }
 
 export function maskValue(value: string) {
+  if (isFlagEnabled('privacy-by-default-poc')) {
+    return value.replace(/.+/, PRIVACY_INPUT_MASK)
```
BenoitZugmeyer, line 130: FMU, you are trying to replace the value with `PRIVACY_INPUT_MASK`, right? So the value length doesn't leak.

First, what about:
```suggestion
    return PRIVACY_INPUT_MASK
```

This is not enough to avoid leaking input length, because we are capturing every keystroke, so we will generate one `IncrementalSnapshot.Input` record for each character entered by the user. We might want to adjust things [around here](https://github.com/DataDog/browser-sdk/blob/90573dcabdefe68b6c1817d463c8b73f60820355/packages/rum-recorder/src/domain/record/observer.ts#L219).

But, I think we should not try to implement this in the same PR.

packages/rum-recorder/src/boot/startRecording.ts
```diff
@@ -47,3 +51,7 @@ export function trackViewEndRecord(lifeCycle: LifeCycle, addRawRecord: (record:
     })
   })
 }
+
+export function getRumRecorderConfig(): Configuration | undefined {
```
BenoitZugmeyer, line 55: We will speak about this a bit further. I don't think it will be that useful to have a singleton for the configuration, because the "censorship level" depends on the DOM tree, not only the user configuration.

packages/rum-recorder/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -153,7 +153,7 @@ export function doStartSegmentCollection(
     state = {
       status: SegmentCollectionStatus.SegmentPending,
       segment,
-      expirationTimeoutId: setTimeout(
+      expirationTimeoutId: window.setTimeout(
```
BenoitZugmeyer, line 156: This shouldn't be needed. We'll have to audit which files are included during e2e tests and avoid including too many files in nodejs.

packages/rum-recorder/src/boot/startRecording.ts
jagracey: Naming is going to need to change as we progress through the POC.
In this case, this variable represents the default censorship mode.
The name should reflect that (along with a few other vars + consts)

packages/rum-recorder/src/domain/record/serializationUtils.ts
jagracey: Agreed. More to discuss, probably making a distinction between `block` and `mask` (or other synonyms) whereby `mask` censors (or at least throttles) JS events. I imagine `block` makes more sense for simple stuff like text nodes (`textContent`)

packages/core/src/domain/configuration.ts
jagracey: `startSessionReplayRecording` is interesting. 🤔

PR #919, opened 2021-06-30
✨ [RUMF-945] allow users to customize the attribute used to define the action name

bcaudan:
## Motivation

It can be useful to reuse existing tagging in the html to retrieve the action name rather than forcing customer to duplicate their tagging for every third party that need to retrieve this info.

take over of #903 

## Changes

Introduce new `actionNameAttribute` init configuration option

Usage:

```html

<script>
  DD_RUM.init({
    ...
    trackInteractions: true,
    actionNameAttribute: 'data-custom-name'
    ...
  })
</script>

<a onclick="triggerSomeRequestOrDomChange()" data-custom-name="Login button">Try it out!</a>
```

Both `data-dd-action-name` and configured attribute can be used, `data-dd-action-name` is favored when both are present on an element.

## Testing

unit, locally

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -42,6 +42,7 @@ export interface UserConfiguration {
   publicApiKey?: string // deprecated
   clientToken: string
   applicationId?: string
+  actionNameAttribute?: string
   internalMonitoringApiKey?: string
   allowedTracingOrigins?: Array<string | RegExp>
   sampleRate?: number
@@ -139,6 +140,10 @@ export function buildConfiguration(userConfiguration: UserConfiguration, buildEn
 
   if ('trackViewsManually' in userConfiguration) {
     configuration.trackViewsManually = !!userConfiguration.trackViewsManually
+  }
+
+  if ('actionNameAttribute' in userConfiguration) {
+    configuration.actionNameAttribute = userConfiguration.actionNameAttribute
   }
 
   return configuration
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -82,6 +82,8 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
     service?: string
     beforeSend?: BeforeSendCallback
 
+    actionNameAttribute?: string
+
     isEnabled: (feature: string) => boolean
   }
 
--- a/packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
@@ -6,7 +6,7 @@ import { safeTruncate } from '@datadog/browser-core'
  */
 const PROGRAMMATIC_ATTRIBUTE = 'data-dd-action-name'
 
-export function getActionNameFromElement(element: Element, actionNameAttribute?: string): string {
+export function getActionNameFromElement(element: Element, actionNameAttribute = PROGRAMMATIC_ATTRIBUTE): string {
   // Proceed to get the action name in two steps:
   // * first, get the name programmatically, explicitly defined by the user.
   // * then, use strategies that are known to return good results. Those strategies will be used on
@@ -21,10 +21,7 @@ export function getActionNameFromElement(element: Element, actionNameAttribute?:
   )
 }
 
-function getActionNameFromElementProgrammatically(
-  targetElement: Element,
-  programmaticAttribute = PROGRAMMATIC_ATTRIBUTE
-) {
+function getActionNameFromElementProgrammatically(targetElement: Element, programmaticAttribute: string) {
   let elementWithAttribute
   // We don't use getActionNameFromElementForStrategies here, because we want to consider all parents,
   // without limit. It is up to the user to declare a relevant naming strategy.
--- a/packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
@@ -1,6 +1,11 @@
 import { safeTruncate } from '@datadog/browser-core'
 
-export function getActionNameFromElement(element: Element): string {
+/**
+ * Get the action name from the attribute 'data-dd-action-name' on the element or any of its parent. This can be customized by the setting the actionNameAttribute configuration option.
+ */
+const PROGRAMMATIC_ATTRIBUTE = 'data-dd-action-name'
+
+export function getActionNameFromElement(element: Element, actionNameAttribute?: string): string {
   // Proceed to get the action name in two steps:
   // * first, get the name programmatically, explicitly defined by the user.
   // * then, use strategies that are known to return good results. Those strategies will be used on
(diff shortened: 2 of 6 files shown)


packages/core/src/domain/configuration.ts
```diff
@@ -42,6 +42,7 @@ export interface UserConfiguration {
   publicApiKey?: string // deprecated
   clientToken: string
   applicationId?: string
+  actionNameAttribute?: string
```
amortemousque, line 45: I'm wondering if the name is not misleading.
It can be confused with the [name of the action object we collect](https://docs.datadoghq.com/real_user_monitoring/browser/data_collected/#action-attributes) 
Maybe something like this is clearer:
- actionNameSelectorAttribute
- actionNameHtmlAttribute
- actionNameTrackingAttribute (to refer to the [tracking section of the doc](https://docs.datadoghq.com/real_user_monitoring/browser/tracking_user_actions/?tab=npm))

packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
```diff
@@ -1,36 +1,38 @@
 import { safeTruncate } from '@datadog/browser-core'
 
-export function getActionNameFromElement(element: Element): string {
+/**
+ * Get the action name from the attribute 'data-dd-action-name' on the element or any of its parent.
+ * This can be customized by the setting the actionNameAttribute configuration option.
+ */
+const PROGRAMMATIC_ATTRIBUTE = 'data-dd-action-name'
```
amortemousque, line 7: Why not to put 'data-dd-action-name' as a default config in the buildConfiguration function like so:
```typescript
if ('actionNameAttribute' in userConfiguration) {
  configuration.actionNameAttribute = userConfiguration.actionNameAttribute
} else {
  configuration.actionNameAttribute = 'data-dd-action-name'
}
```

bcaudan: IMO the advantage of the current approach it that it keeps all the logic at the same place.

what benefit do you see with your suggestion?

packages/core/src/domain/configuration.ts
bcaudan: we will add some doc about it to explain the usage so IMO it would be clearer enough.

@hdelaby @BenoitZugmeyer @webNeat any thoughts?

packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
```diff
@@ -1,36 +1,38 @@
...
+export function getActionNameFromElement(element: Element, actionNameAttribute = PROGRAMMATIC_ATTRIBUTE): string {
   // Proceed to get the action name in two steps:
   // * first, get the name programmatically, explicitly defined by the user.
   // * then, use strategies that are known to return good results. Those strategies will be used on
   //   the element and a few parents, but it's likely that they won't succeed at all.
   // * if no name is found this way, use strategies returning less accurate names as a fallback.
   //   Those are much likely to succeed.
   return (
-    getActionNameFromElementProgrammatically(element) ||
+    getActionNameFromElementProgrammatically(element, actionNameAttribute) ||
```
BenoitZugmeyer, line 17: What about always checking `data-dd-action-name`, and checking *extra* attributes based on the configuration? This would allow library authors to use the constant attribute name without worrying that it could be changed by the application.

(Illustration: I'm a frontend dev, and I want to share my uikit library to the world. I want to support Datadog Browser SDK and provide action names on buttons out of the box. I want to use a constant attribute for it, and avoid asking the user for its "actionNameAttribute" value.)

packages/core/src/domain/configuration.ts
BenoitZugmeyer: About the wording: `actionNameAttribute` + doc LGTM, it's simple enough.

About the configuration: maybe we could have a more abstract `getActionName(element, event)` function instead, to allow the customer to implement their own logic to add beside our logic? That would be a bit more flexible.

packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
amortemousque: I like the idea of having the configuration defined in a single place.
Overriding the output of buildConfiguration(), would mean to lose the control of something we want to centralize.
IMO, it's easier to understand and debug.

packages/core/src/domain/configuration.ts
bcaudan: a more abstract strategy sounds interesting but I'd be in favor of waiting to have the need, in order to see what could be the most relevant approach to do that.

packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
bcaudan: obsolete with f3c7736

packages/core/src/domain/configuration.ts
BenoitZugmeyer: Sounds good to me!

BenoitZugmeyer (approved):
Nice!

packages/core/src/domain/configuration.ts
amortemousque: Yes, good to me too :)

packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
```diff
@@ -1,36 +1,39 @@
...
   // Proceed to get the action name in two steps:
   // * first, get the name programmatically, explicitly defined by the user.
   // * then, use strategies that are known to return good results. Those strategies will be used on
   //   the element and a few parents, but it's likely that they won't succeed at all.
   // * if no name is found this way, use strategies returning less accurate names as a fallback.
   //   Those are much likely to succeed.
   return (
-    getActionNameFromElementProgrammatically(element) ||
+    getActionNameFromElementProgrammatically(element, DEFAULT_PROGRAMMATIC_ATTRIBUTE) ||
+    (userProgrammaticAttribute && getActionNameFromElementProgrammatically(element, userProgrammaticAttribute)) ||
```
webNeat, line 18: Can you please explain why the default attribute has priority over the user custom attribute?

packages/core/src/domain/configuration.ts
webNeat: I was thinking that having an array of attributes to check for, instead of just one would offer more flexibility, but the function `getActionName` also is interesting.
The current solution sounds enough to me for now.

webNeat (approved):
LGTM, just added a small question.

packages/rum-core/src/domain/rumEventsCollection/action/getActionNameFromElement.ts
bcaudan: it seemed more convenient for reusing a previous tagging but allow to override it with our own attribute if needed

packages/core/src/domain/configuration.ts
niieani: The function idea is great! Love that concept.

PR #927, opened 2021-07-08
🐛 ensure that test cookie is correctly deleted after the test

### LINKED ISSUE #926: dd_cookie_test_{uuid} not manually deleted which makes it just stick around in firefox.
dd_cookie_test_{uuid} cookies produced in packages/core/src/browser/cookie.ts rely on the expiry time to delete the cookie, which firefox doesn't do. Combined w/ the uuid in the cookie name, this makes for a noisy cookie source producing one extra cookie per page load until you clear cookies. 

Would love for this to either be manually deleted (probably the simplest fix) or to be a session cookie to better signal the transient intent of this cookie ( though that doesn't work to confirm non-session cookies can be used )

bcaudan:
## Motivation

Firefox does not seem to quickly clean cookie based on expiration date, so test cookies can pile up quickly during a navigation.
fixes #926

## Changes

Explicitly delete test cookie after the test

## Testing

manual

---

### DIFF
--- a/packages/core/src/browser/cookie.ts
+++ b/packages/core/src/browser/cookie.ts
@@ -100,6 +100,7 @@ export function getCurrentSite() {
       candidateDomain = `${domainLevels.pop()!}.${candidateDomain}`
       setCookie(testCookieName, testCookieValue, ONE_SECOND, { domain: candidateDomain })
     }
+    deleteCookie(testCookieName, { domain: candidateDomain })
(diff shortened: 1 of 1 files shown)


BenoitZugmeyer (commented):
Maybe we should do the same for `getCurrentSite`?

BenoitZugmeyer (approved):

webNeat (approved):

PR #929, opened 2021-07-08
✨ [RUMF-938] add the session plan to RUM events

BenoitZugmeyer:
## Motivation

Send the session plan to the intake for each RUM events so they can apply specific rules to them.

## Changes

This commit introduces a new method to get the plan from a `RumSession`. For now, it is hardcoded to `REPLAY`, but in the future (RUMF-940) it will be defined by an `Configuration` option.

## Testing

Unit

---

### DIFF
--- a/packages/rum-core/src/domain/rumSession.ts
+++ b/packages/rum-core/src/domain/rumSession.ts
@@ -1,12 +1,18 @@
-import { Configuration, performDraw, startSessionManagement } from '@datadog/browser-core'
+import { Configuration, performDraw, Session, startSessionManagement } from '@datadog/browser-core'
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 
 export const RUM_SESSION_KEY = 'rum'
 
 export interface RumSession {
   getId: () => string | undefined
+  getPlan(): RumSessionPlan | undefined
   isTracked: () => boolean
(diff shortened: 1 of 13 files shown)


webNeat (approved):

packages/rum-core/src/domain/rumSession.ts
```diff
@@ -26,7 +32,8 @@ export function startRumSession(configuration: Configuration, lifeCycle: LifeCyc
 
   return {
     getId: session.getId,
-    isTracked: () => session.getId() !== undefined && isTracked(session.getTrackingType()),
+    getPlan: () => getSessionPlan(session),
+    isTracked: () => getSessionPlan(session) !== undefined,
```
bcaudan, line 36: nit: extracting a `isTracked` method directly could:
- be clearer here and in getSessionPlan method
- avoid to run future getSessionPlan code when only interested by isTracked info

bcaudan (approved):

packages/rum-core/src/domain/rumSession.ts
BenoitZugmeyer: That was my initial thought, but then I realized we already have an `isTracked` function and I couldn't think of a good name to differentiate the two :D Let me try something.

PR #930, opened 2021-07-08
✨ network error removal

amortemousque:
## Motivation

RUM errors are not the best way to trigger attention for network requests with a bad status code. We already collect that information within RUM resources. By stopping the generation of RUM errors for these resources, we simplify the value proposition of RUM errors.

## Changes

- Put console, network and runtime error tracking in separated files
- Stop RUM network error when `remove-network-errors` flag is enabled

## Testing

Unit, Locally

---


packages/core/src/tools/observable.ts
```diff
@@ -18,3 +20,5 @@ export class Observable<T> {
     this.observers.forEach((observer) => observer(data))
   }
 }
+
+export type ErrorObservable = Observable<RawError>
```
BenoitZugmeyer, line 24: I'm not sure this type has its place there. I think, either:
* put it in a new `domain/error/types.ts`
* put it in `tools/error`
* replace `ErrorObservable` with `Observable<RawError>` everywhere

packages/core/src/domain/error/trackNetworkError.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      fetchStubManager.whenAllComplete(() => {
+        const stack = (errorObservableSpy.calls.mostRecent().args[0] as RawError).stack
+        expect(stack).toEqual('Lorem ipsum dolor sit amet orci ...')
+        done()
+      })
+    })
+  })
+
+  describe('when trackAbortedRequests param is true', () => {
```
BenoitZugmeyer, line 137:
```suggestion
  describe('when trackAbortedRequests param is false', () => {
```

packages/logs/src/boot/startLogs.ts
```diff
@@ -37,10 +39,14 @@ export function startLogs(
...
-  const errorObservable =
-    initConfiguration.forwardErrorsToLogs !== false
-      ? startAutomaticErrorCollection(configuration)
-      : new Observable<RawError>()
+  const errorObservable = new Observable<RawError>()
+
+  if (initConfiguration.forwardErrorsToLogs !== false) {
+    trackConsoleError(errorObservable)
+    trackRuntimeError(errorObservable)
+    trackNetworkError(configuration, errorObservable, configuration.isEnabled('remove-network-errors'))
```
BenoitZugmeyer, line 47: When releasing V3, I don't think we'll remove errors for aborted requests for logs, so this extra parameter may be removed. Nonetheless, we can leave it here until we get confirmation from @hdelaby

webNeat (approved):
Good job!

packages/core/src/tools/observable.ts
webNeat: I would simply use `Observable<RawError>` because it's already used in some places

packages/logs/src/boot/startLogs.ts
amortemousque: Yes, I think this decision does not impact this PR since we are still in V2. Maybe it can be added to the v3 release plan when @hdelaby will confirm.

packages/core/src/tools/observable.ts
amortemousque: Yep, using Observable<RawError> is probably easier and still well express the intent.

BenoitZugmeyer (approved):

PR #931, opened 2021-07-09
✨ [RUMF-940] implement the replay sample rate option and remove resource sample rate option

BenoitZugmeyer:
## Motivation

Let the customer chose which session plan to use.

## Changes

We chose to use the same value as the deprecated "session without resource" to represent "sessions with lite plans" in session cookies. Because of this, replacing the deprecated `resourceSampleRate` option with the new `replaySampleRate` option in a single PR made more sense than doing it in separate PRs.

* remove "session without resource" support
* replace `resourceSampleRate` option with `replaySampleRate`
* adjust namings
* implement the "lite plan" restrictions: don't collect long tasks and don't allow recording

## Testing

Unit, manual

---


amortemousque (approved):

webNeat (approved):
Good job!

packages/rum-recorder/src/boot/rumRecorderPublicApi.ts
```diff
@@ -24,13 +25,17 @@ export type RumRecorderUserConfiguration = RumRecorderInitConfiguration
...
+  IntentToStart,
   Starting,
   Started,
 }
 type RecorderState =
   | {
       status: RecorderStatus.Stopped
     }
+  | {
+      status: RecorderStatus.IntentToStart
```
bcaudan, line 37: what about adding some comments explaining the difference between these start states?

packages/rum-core/src/domain/rumSession.ts
```diff
@@ -16,8 +16,11 @@ export enum RumSessionPlan {
 
 export enum RumTrackingType {
   NOT_TRACKED = '0',
-  TRACKED_WITH_RESOURCES = '1',
-  TRACKED_WITHOUT_RESOURCES = '2',
+  // Note: the "tracking type" value (stored in the session cookie) does not match the "session
+  // plan" value (sent in RUM events). This is expected, and was done to keep retrocompatibility
+  // with active sessions when upgrading the SDK.
+  TRACKED_REPLAY = '1',
+  TRACKED_LITE = '2',
```
bcaudan, line 23: As `TRACKED_WITHOUT_RESOURCES` is not strictly equivalent to `TRACKED_LITE`, we could also consider to perform a new draw or consider it as not tracked.
It sounds fine to me but it worth mentioning this behavior to @hdelaby.

packages/rum-core/src/domain/rumSession.ts
```diff
@@ -43,7 +46,9 @@ function isSessionTracked(session: Session<RumTrackingType>) {
 function getSessionPlan(session: Session<RumTrackingType>) {
   return isSessionTracked(session)
     ? // TODO: return correct plan based on tracking type
-      RumSessionPlan.REPLAY
+      session.getTrackingType() === RumTrackingType.TRACKED_LITE
```
bcaudan, line 49: Do we still need the above TODO?

packages/rum-core/src/boot/startRum.ts
```diff
@@ -46,7 +46,9 @@ export function startRum(
     getCommonContext
   )
 
-  startLongTaskCollection(lifeCycle)
+  if (session.getPlan() === RumSessionPlan.REPLAY) {
```
bcaudan, line 49: it could be interesting to have `session.hasReplayPlan()` / `session.hasLitePlan()` instead of letting callers do the check.
It seems that `getPlan` is also used in assembly but `isTracked` would be more appropriate here, right?

BenoitZugmeyer: > it could be interesting to have session.hasReplayPlan() / session.hasLitePlan() instead of letting callers do the check.

I don't really mind. I can add those methods if you want to, but then I'd like to introduce a helper to mock RumSession objects, because it's a bit inconvenient to explicitly every functions everywhere.

>  It seems that getPlan is also used in assembly but isTracked would be more appropriate here, right?

I mean, `getPlan` can return `undefined`, so I used it in the condition to avoid casting the result. But sure if we implement `hasReplayPlan` and `hasLitePlan`, we can just use `isTracked` instead.

bcaudan: In assembly, the intent seems better expressed to me by checking `session.isTracked` rather than retrieving the plan and checking that it is undefined.

Then, since it seems that we don't need to manipulate the plan but rather explicitly check the current plan, the `hasXXXPlan` API seems nicer to me but not a big deal.

packages/rum-core/src/boot/startRum.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { RelativeTime, Configuration } from '@datadog/browser-core'
 import { RumSession } from '@datadog/browser-rum-core'
+import { createRumSessionMock, RumSessionMock } from '../..//test/mockRumSession'
```
bcaudan, line 3:
```suggestion
import { createRumSessionMock, RumSessionMock } from '../../test/mockRumSession'
```

bcaudan (approved):

PR #932, opened 2021-07-09
👷 Versioned CDN files

amortemousque:
## Motivation

All CDN users are currently updated to the latest version at each deployment. It means that:

- Customers are exposed to bugs potentially introduced by a new version
- Each breaking change is critical and force users to update their code
- It adds lot of pressure on how to communicate these changes to customers

To tackle these issues, we decided to use versioned CDN URLs for all major changes.

## Changes

- Update gitlab deploy-prod-stable to send the version to the deploy script
- Update deployment script to handle suffix version

## Testing

Locally

---


scripts/deploy.sh
```diff
@@ -3,8 +3,9 @@
 set -ex
 
 env=$1
-suffix=${2+"-$2"}
+suffix=$2
 
+USAGE="Usage: ./deploy.sh staging|prod head|canary|version"
```
BenoitZugmeyer, line 8: NIT:

```suggestion
USAGE="Usage: ./deploy.sh staging|prod head|canary|vXXX"
```

BenoitZugmeyer (approved):

PR #940, opened 2021-07-15
💥 remove deprecated rum init options

amortemousque:
## Motivation

In the context of the RUM version 3, we are cleaning up deprecated options.
 
## Changes

- Remove deprecated options publicApiKey
- Remove deprecated options datacenter

## Testing

Unit, Locally

---


packages/core/src/domain/configuration.ts
```diff
@@ -69,6 +67,11 @@ export interface InitConfiguration {
   replica?: ReplicaUserConfiguration
 }
 
+export interface InitConfigurationWithDeprecated extends InitConfiguration {
```
bcaudan, line 70: Instead of handling that in core by adding a new interface with deprecated keys, what about making `RumInitConfiguration` extends `Omit<InitConfiguration, 'publicApiKey', 'datacenter'>`?

BenoitZugmeyer: We talked about it and I think it's a bit more explicit with a specific interface, but no strong opinion. After thinking about it, in the future we may want to deprecate more options and then we could have deprecated options in both interfaces, which would be unfortunate. Using the Omit method would be less explicit but more flexible.

amortemousque: Yes, personally, I found neither of these solutions, completely satisfying. But liked we discussed with Bastien, I'll go with the Omit before doing a more in depth analyses of our configuration strategy.

bcaudan: IMO for the future, the configuration code should be reworked to better address our current usages. I am not super happy with the current mix of product specific options with common options in a core package.
FMU, the current need is to hide those properties from RUM users, the Omit strategy would address this need without adding extra complexity in the common configuration code.

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -25,7 +25,7 @@ import { RumEvent } from '../rumEvent.types'
 import { buildEnv } from './buildEnv'
 import { startRum } from './startRum'
 
-export interface RumInitConfiguration extends InitConfiguration {
+export interface RumInitConfiguration extends Omit<InitConfiguration, 'publicApiKey' | 'datacenter'> {
```
bcaudan, line 28: You should import Omit from utils to not break compatibility with older ts version

packages/core/src/domain/configuration.ts
amortemousque: Do we agree that, with this approach, if a user use the package in plain JS, he will still be able to use the options ?

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: Yes

packages/core/src/domain/configuration.ts
bcaudan: what about preventing it?

amortemousque: Yes, done.

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -18,14 +18,18 @@ import {
...
 } from '@datadog/browser-core'
 import { RumEventDomainContext } from '../domainContext.types'
 import { CommonContext, User, ActionType } from '../rawRumEvent.types'
 import { RumEvent } from '../rumEvent.types'
 import { buildEnv } from './buildEnv'
 import { startRum } from './startRum'
 
-export interface RumInitConfiguration extends InitConfiguration {
+const ignoredConfigurationOptions = { publicApiKey: true, datacenter: true } as const
+type IgnoredConfigurationOptions = keyof typeof ignoredConfigurationOptions
```
BenoitZugmeyer, line 30: Nice! It could be better to express this as an array:

```suggestion
const ignoredConfigurationOptions = [ 'publicApiKey', 'datacenter' ] as const
type IgnoredConfigurationOptions = (typeof ignoredConfigurationOptions)[number]
```

amortemousque: Sure !

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -211,14 +215,22 @@ export function makeRumPublicApi<C extends RumInitConfiguration>(startRumImpl: S
     return result
   }
 
+  function ignoreConfigurationOptions(initConfiguration: C, options: readonly string[]) {
+    for (const option in initConfiguration) {
+      if (includes(options, option)) {
+        delete initConfiguration[option]
+      }
+    }
```
BenoitZugmeyer, line 223: NIT:
```suggestion
    for (const option of options) {
      delete initConfiguration[option]
    }
```

BenoitZugmeyer (dismissed):

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: If I do this, typescript complains `No index signature with a parameter of type 'string' was found on type RumInitConfiguration.ts`.
I found 4 options:
- `delete initConfiguration[option as keyof RumInitConfiguration]`: IMO it adds confusion as the key I'm trying to remove are not part of RumInitConfiguration
- `delete initConfiguration[option as any]`: I generally avoid it as it is error prone
- Ignore typescript check: Not a big fan, either
- Change the loop: What I did

But I've no strong opinion about it. 
Is there a better solution? or what do you prefer?

bcaudan: what about 
```ts
ignoredConfigurationOptions.forEach((option) => {
  delete (initConfiguration as InitConfiguration)[option]
})
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -18,14 +18,19 @@ import {
...
 } from '@datadog/browser-core'
 import { RumEventDomainContext } from '../domainContext.types'
 import { CommonContext, User, ActionType } from '../rawRumEvent.types'
 import { RumEvent } from '../rumEvent.types'
 import { buildEnv } from './buildEnv'
 import { startRum } from './startRum'
 
-export interface RumInitConfiguration extends InitConfiguration {
+const ignoredConfigurationOptions = ['publicApiKey', 'datacenter'] as const
+type IgnoredConfigurationOptions = typeof ignoredConfigurationOptions[number]
```
bcaudan, line 31:
```suggestion
const deprecatedConfigurationOptions = ['publicApiKey', 'datacenter'] as const
type DeprecatedConfigurationOptions = typeof deprecatedConfigurationOptions[number]
```

amortemousque: I've thought about it. But to me it's not really the `deprecated` options. It's the options we no logger support. 
At some point, we might have some other RUM deprecated options which won't be in this list. Won't it be a bit confusing?

bcaudan: what difference do you see between deprecated options and options that we no longer support?
Do you have an example scenario where a deprecated options should not be in this list?

packages/rum-core/src/boot/rumPublicApi.ts
BenoitZugmeyer: I think a TS cast is fine here. `any` should be fine IMO. I think it's better than iterating over all configuration options. We may imagine more complex typings, but I don't think it's worth it here, and I think we'll need an imprecise cast no matter what.

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: To me, an option marked as `deprecated` is here to warn the user that the option is going to be removed at some point.
It also indicate, what's the new way to do what we want. 
For example we might have 
```js
export interface RumInitConfiguration extends Omit<InitConfiguration, IgnoredConfigurationOptions> {
  applicationId: string
  /**
   * @deprecated use XXXX instead
   */
  beforeSend?: (event: RumEvent, context: RumEventDomainContext) => void | boolean
}
```

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: I go with the cast of Bastien. It's short and clear :)

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: Not compatible with Typescript 3.0
Only available for TypeScript 3.4

packages/rum-core/src/boot/rumPublicApi.ts
bcaudan: ok, so in our case they are removed from a user perspective but only deprecated in the code base since they are still handled by core. It depends from where we look.
IMO, `ignored` is not right because a reader can ask himself why are they ignored.
wdyt?

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: I used a tuple helper for compatibility with Typescript v3.0

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: I go with `droppedConfigurationOptions`

packages/core/src/tools/utils.ts
```diff
@@ -542,6 +542,15 @@ export function combine(...sources: any[]): unknown {
...
+/*
+ * Readonly tuple for TS 3.0 compatibility
+ * Example: `const t = tuple('foo', 'bar')`
+ * With TS 3.4 or latter: `const t = ['foo', 'bar'] as const`
+ */
+export function tuple<T extends string[]>(...args: T) {
+  return args
+}
+
 // Define those types for TS 3.0 compatibility
```
bcaudan, line 554:
```suggestion
// Define those utilities for TS 3.0 compatibility

/*
 * Example: `const t = tuple('foo', 'bar')`
 * With TS 3.4 or latter: `const t = ['foo', 'bar'] as const`
 */
export function tuple<T extends string[]>(...args: T) {
  return args
}
```

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { ONE_SECOND, RelativeTime, getTimeStamp, display, TimeStamp } from '@datadog/browser-core'
+import { ONE_SECOND, RelativeTime, getTimeStamp, display, TimeStamp, Datacenter } from '@datadog/browser-core'
```
bcaudan, line 1: what about not importing this type?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -18,14 +18,20 @@ import {
...
 } from '@datadog/browser-core'
 import { RumEventDomainContext } from '../domainContext.types'
 import { CommonContext, User, ActionType } from '../rawRumEvent.types'
 import { RumEvent } from '../rumEvent.types'
 import { buildEnv } from './buildEnv'
 import { startRum } from './startRum'
 
-export interface RumInitConfiguration extends InitConfiguration {
+// eslint-disable-next-line local-rules/disallow-side-effects
+const droppedConfigurationOptions = tuple('publicApiKey', 'datacenter')
```
BenoitZugmeyer, line 31: I think you can do:

```suggestion
const droppedConfigurationOptions = [ 'publicApiKey' as 'publicApiKey', 'datacenter' as 'datacenter'] 
```

which is a bit verbose but doesn't add anything at runtime.

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.spec.ts
amortemousque: What is  your reasonning behind it? 
I personally like to have a valid configuration on tests.

bcaudan: since we don't support Datacenter anymore on RUM, it could be nice to remove any dependency on it in RUM package.

amortemousque: yes, done

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: solution: `const droppedConfigurationOptions = ['publicApiKey' as const, 'datacenter' as const]
`

bcaudan (approved):

PR #944, opened 2021-07-15
💥 always use alternative domains for RUM

amortemousque:
## Motivation

In the context of the version 3 always use alternative domains.

## Changes

Remove useAlternateIntakeDomains option

## Testing

Unit, Locally

---


BenoitZugmeyer (commented):
This options defaults to `false`, so you should set it to `true` to always use the alternative domains

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -90,6 +91,8 @@ export function makeRumPublicApi<C extends RumInitConfiguration>(startRumImpl: S
       initConfiguration.clientToken = initConfiguration.publicApiKey
     }
 
+    ;(initConfiguration as InitConfiguration).useAlternateIntakeDomains = true
+
     const { configuration, internalMonitoring } = commonInit(initConfiguration, buildEnv)
```
BenoitZugmeyer, line 96: Suggestion: the cast could be avoided like that:

```suggestion
    const { configuration, internalMonitoring } = commonInit(
      { ...initConfiguration, useAlternateIntakeDomains: true },
      buildEnv
    )
```

WDYT?

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: Nice !

PR #954, opened 2021-07-22
Update computeStackTrace.ts

dickfickling:
The open parenthesis should be optional as it's not included in stack traces without a function name. The close parenthesis is already marked optional.

## Motivation

Formatting the stack trace is currently broken with stacks that don't include the function name.

Original stack trace:
```
Error: wow so broke
    at chrome-extension://<id>/content/index.js:85:37167
    at chrome-extension://<id>/content/index.js:85:37194
    at chrome-extension://<id>/content/index.js:85:37379
```

After parsing / without the fix (the function names here are wrong):
```
Error: wow so broke
at addError @ undefined
at <anonymous> @ undefined
at yt @ undefined
at addError @ undefined
```

After parsing / with the fix:
```
Error: wow so broke
at <anonymous> @ chrome-extension://<id>/content/index.js:85:37167
at <anonymous> @ chrome-extension://<id>/content/index.js:85:37194
at <anonymous> @ chrome-extension://<id>/content/index.js:85:37379
```

## Changes

Make the open parenthesis optional when parsing stack traces

## Testing

Throw an exception and see whether anonymous function calls are included correctly in the stack

---

### DIFF
--- a/packages/core/src/domain/tracekit/computeStackTrace.ts
+++ b/packages/core/src/domain/tracekit/computeStackTrace.ts
@@ -184,7 +184,7 @@ export function computeStackTraceFromStackProp(ex: unknown) {
   }
 
   // eslint-disable-next-line  max-len
-  const chrome = /^\s*at (.*?) ?\(((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\/).*?)(?::(\d+))?(?::(\d+))?\)?\s*$/i
+  const chrome = /^\s*at (.*?) ?\(?((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\/).*?)(?::(\d+))?(?::(\d+))?\)?\s*$/i
   // eslint-disable-next-line  max-len
   const gecko = /^\s*(.*?)(?:\((.*?)\))?(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle)(?::(\d+))?(?::(\d+))?\s*$/i
   // eslint-disable-next-line  max-len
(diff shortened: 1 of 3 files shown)


bcaudan:
Hello @dickfickling, thanks for submitting that.
Could you add a test for it?

dickfickling:
@bcaudan There's a test that _seems_ like it should be failing here: https://github.com/DataDog/browser-sdk/blob/22fdb27dfb276cd77382977e01cd0f23a92f49e8/packages/core/src/domain/tracekit/parser.spec.ts#L359 without visibility into your CI, i don't know why it's currently working.

bcaudan:
Can you test it locally with `yarn test:unit`?

dickfickling:
can't run tests locally because `yarn install` is failing:
```error /Users/dick/chums/browser-sdk/node_modules/puppeteer: Command failed.
Exit code: 1
Command: node install.js
Arguments: 
Directory: /Users/dick/chums/browser-sdk/node_modules/puppeteer
Output:
The chromium binary is not available for arm64.
If you are on Ubuntu, you can install with: 

 sudo apt install chromium


 sudo apt install chromium-browser

/Users/dick/chums/browser-sdk/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:115
                    throw new Error();
                    ^
```
i'm on an M1 mac.. I don't really have the time to get a dev env set up on an x86 machine just to run tests on a datadog library

dickfickling:
some more details: the current test is working somewhat by accident, as the regex matches the fourth line of the stack (the one without parentheses around the url) as a winjs line. Because if we change the url scheme from `http` to `chrome-extension`, it breaks.

However, my change of making the open paren optional breaks a bunch of other tests, because the chrome regex starts matching lines that _aren't_ chrome exceptions.

dickfickling:
Ok, as with all regex bugs, fixing one bug just causes another one to appear.

I'm pushing a new commit that updates a test to encompass the fix I've made. Without making the parenthesis optional, the test fails. While my change fixes my issue and causes the test to pass, it also causes a bunch of other tests to fail.

I don't have the time to fix the regexes bit by bit until all the tests pass, unfortunately. I can close this PR and turn it into a github issue, if you'd like.

bcaudan:
Yes, turning it into an issue could be better then.
Thanks.

dickfickling:
Closing this and moving discussion to #957 957

PR #955, opened 2021-07-22
🐛 fix recording session renewal

amortemousque:
## Motivation

Some LITE sessions have the replay enabled while they shouldn't. Using sample rates, it's possible to go from a REPLAY session to a LITE or an untracked session. In these cases, the recorder should stop recording.

## Changes

Stop recording when a new session with a LITE plan or untracked is started

## Testing

Unit

---

### DIFF
--- a/packages/rum/src/boot/recorderApi.spec.ts
+++ b/packages/rum/src/boot/recorderApi.spec.ts
@@ -179,6 +179,7 @@ describe('makeRecorderApi', () => {
         session.setNotTracked()
         lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
         expect(startRecordingSpy).not.toHaveBeenCalled()
+        expect(stopRecordingSpy).not.toHaveBeenCalled()
       })
     })
 
@@ -191,6 +192,7 @@ describe('makeRecorderApi', () => {
         recorderApi.start()
         lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
         expect(startRecordingSpy).not.toHaveBeenCalled()
+        expect(stopRecordingSpy).not.toHaveBeenCalled()
       })
     })
 
@@ -203,7 +205,8 @@ describe('makeRecorderApi', () => {
         recorderApi.start()
         session.setLitePlan()
         lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
-        expect(startRecordingSpy).toHaveBeenCalled()
+        expect(startRecordingSpy).toHaveBeenCalledTimes(1)
+        expect(stopRecordingSpy).toHaveBeenCalled()
       })
     })
 
@@ -216,6 +219,7 @@ describe('makeRecorderApi', () => {
         recorderApi.start()
         session.setNotTracked()
         lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
+        expect(startRecordingSpy).toHaveBeenCalledTimes(1)
         expect(stopRecordingSpy).toHaveBeenCalled()
       })
     })
--- a/packages/rum/src/boot/recorderApi.spec.ts
+++ b/packages/rum/src/boot/recorderApi.spec.ts
@@ -228,8 +228,8 @@ describe('makeRecorderApi', () => {
       it('keeps recording if startSessionReplayRecording was called', () => {
         recorderApi.start()
         lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
-        expect(stopRecordingSpy).not.toHaveBeenCalled()
-        expect(startRecordingSpy).toHaveBeenCalledTimes(1)
+        expect(stopRecordingSpy).toHaveBeenCalled()
+        expect(startRecordingSpy).toHaveBeenCalledTimes(2)
       })
 
       it('does not starts recording if stopSessionReplayRecording was called', () => {
--- a/packages/rum/src/boot/recorderApi.spec.ts
+++ b/packages/rum/src/boot/recorderApi.spec.ts
@@ -157,6 +157,7 @@ describe('makeRecorderApi', () => {
         session.setReplayPlan()
         lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
         expect(startRecordingSpy).toHaveBeenCalled()
+        expect(stopRecordingSpy).not.toHaveBeenCalled()
       })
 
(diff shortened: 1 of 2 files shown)


packages/rum/src/boot/recorderApi.ts
```diff
@@ -63,6 +63,11 @@ export function makeRecorderApi(startRecordingImpl: StartRecording): RecorderApi
         if (state.status === RecorderStatus.IntentToStart) {
           startStrategy()
         }
+
+        if (!session.hasReplayPlan() && state.status === RecorderStatus.Started) {
```
bcaudan, line 67:
```suggestion
        if (state.status === RecorderStatus.Started) {
```
that does not break any test

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -158,6 +137,85 @@ describe('makeRecorderApi', () => {
...
+    let session: RumSessionMock
+    let lifeCycle: LifeCycle
+    beforeEach(() => {
+      session = createRumSessionMock()
+      setupBuilder.withSession(session)
+      ;({ lifeCycle } = setupBuilder.build())
+      rumInit(DEFAULT_INIT_CONFIGURATION)
+    })
+
+    describe('when an untracked session expires and the renewed session plan is REPLAY', () => {
```
bcaudan, line 150: would it make sense to test all the possible combinations?
lite -> lite, replay -> replay, ...

packages/rum/src/boot/recorderApi.ts
```diff
@@ -63,6 +63,11 @@ export function makeRecorderApi(startRecordingImpl: StartRecording): RecorderApi
...
+          stopStrategy()
+          state = { status: RecorderStatus.IntentToStart }
+        }
```
bcaudan, line 70: I am wondering if it would make sense to always re-call stop/start at session renew.
Could the triggered lifecycle events be needed?

packages/rum/src/boot/recorderApi.ts
amortemousque: In the current implementation, we only want to stop when the session becomes untracked or lite.
I've updated the tests

packages/rum/src/boot/recorderApi.ts
amortemousque: From what I see,  RECORD_STARTED and RECORD_STOPPED are only subscribed in view tracking to update the hasReplay flag. If the plan doesn't change between two session there no need to trigger them. 
In addition, it does not really simplify the code. 
Maybe it could make sense in term of domain logic but for now, I don't see it.

packages/rum/src/boot/recorderApi.spec.ts
amortemousque: Yes, done

packages/rum/src/boot/recorderApi.ts
bcaudan: > RECORD_STARTED and RECORD_STOPPED are only subscribed in view tracking to update the hasReplay flag. 

yes and since this flag is not reset on renew session it works fine

>  it does not really simplify the code

it could remove the need to check the plan so one less concept to have in mind to understand what is happening


Also, could we end up with a segment mixing events from different sessions if we don't flush the current one on session renew?

amortemousque: > it could remove the need to check the plan so one less concept to have in mind to understand what is happening

Yes, we might end up with:       
```js
lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
  const status = state.status
  stopStrategy()

  if (status !== RecorderStatus.Stopped) {
    startStrategy()
  } 
})
```
> Also, could we end up with a segment mixing events from different sessions if we don't flush the current one on session renew?

FMU, the recorder only cares about view life cycle. SESSION_RENEWED event triggers VIEW_ENDED and VIEW_CREATED, which will be caught by session replay to flushSegments. 
https://github.com/DataDog/browser-sdk/blob/80d25788d40ccaffcb8377927e8bdd4a88b4f14a/packages/rum/src/domain/segmentCollection/segmentCollection.ts#L95-L97
New segments are flagged with the new session id found in the same parentContexts has RUM as you can see here.
https://github.com/DataDog/browser-sdk/blob/80d25788d40ccaffcb8377927e8bdd4a88b4f14a/packages/rum/src/domain/segmentCollection/segmentCollection.ts#L191-L200

The parent context is up to date because it also listen VIEW_CREATED
https://github.com/DataDog/browser-sdk/blob/80d25788d40ccaffcb8377927e8bdd4a88b4f14a/packages/rum-core/src/domain/parentContexts.ts#L39-L42

packages/rum/src/boot/recorderApi.ts
```diff
@@ -60,7 +60,10 @@ export function makeRecorderApi(startRecordingImpl: StartRecording): RecorderApi
       parentContexts: ParentContexts
     ) => {
       lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
-        if (state.status === RecorderStatus.IntentToStart) {
+        const status = state.status
+        stopStrategy()
+
+        if (status !== RecorderStatus.Stopped) {
           startStrategy()
         }
```
bcaudan, line 68:
```suggestion
        if (state.status === RecorderStatus.Started) {
          stopStrategy()
          state = { status: RecorderStatus.IntentToStart }
        }

        if (state.status === RecorderStatus.IntentToStart) {
          startStrategy()
        }
```
it seems to behave the same but I would find it easier to understand this way.
wdyt?

packages/rum/src/boot/recorderApi.ts
bcaudan: ok

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -158,6 +137,158 @@ describe('makeRecorderApi', () => {
...
+      beforeEach(() => {
+        session.setReplayPlan()
+      })
+
+      it('stops recording if startSessionReplayRecording was called', () => {
+        recorderApi.start()
+        session.setNotTracked()
+        lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
+        expect(stopRecordingSpy).toHaveBeenCalled()
+      })
```
bcaudan, line 220: should not we ensure that start is not called for the not tracked session

bcaudan (approved):
Nice!

PR #961, opened 2021-07-26
Enable foreground tracking for everyone

ThibautGeriz:
## Motivation

Enable foreground tracking for everyone

## Changes

Remove experimental flag

---

### DIFF
--- a/packages/rum-core/src/domain/foregroundContexts.ts
+++ b/packages/rum-core/src/domain/foregroundContexts.ts
@@ -1,6 +1,4 @@
 import {
-  Configuration,
-  noop,
   addEventListener,
   DOM_EVENT,
   RelativeTime,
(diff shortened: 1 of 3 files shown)


BenoitZugmeyer (approved):

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -60,10 +51,6 @@ function addNewForegroundPeriod() {
   }
   const currentForegroundPeriod = foregroundPeriods[foregroundPeriods.length - 1]
   if (currentForegroundPeriod !== undefined && currentForegroundPeriod.end === undefined) {
-    addMonitoringMessage('Previous foreground periods not closed. Continuing current one', {
```
bcaudan, line 63: What about other messages?

ThibautGeriz: AFAIK, They are not triggered at the moment, I would like to know if we change something and those limit case start to trigger.

bcaudan: By looking at the logs I shared with you earlier, we can see some occurrences for all of them.

ThibautGeriz:
Closed for https://github.com/DataDog/browser-sdk/pull/1053

PR #976, opened 2021-08-04
⚗ [RUMF-971] experiment to detect when the computer goes to sleep 

BenoitZugmeyer:
## Motivation

This is part of an investigation where we observed abnormally high resource durations and web vital timings

## Changes

Add an experimental way to detect when the computer goes to sleep. Use it on resources and internal monitoring logs to help investigation.

## Testing

Unit, manual

---

### DIFF
--- a/packages/rum-core/src/domain/trackSleep.ts
+++ b/packages/rum-core/src/domain/trackSleep.ts
@@ -1,0 +1,39 @@
+import { monitor, ONE_MINUTE, ONE_SECOND, TimeStamp, timeStampNow } from '@datadog/browser-core'
+
+export const SLEEP_CHECK_DELAY = ONE_SECOND
+export const SLEEP_THRESHOLD = ONE_MINUTE
+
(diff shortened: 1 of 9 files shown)


packages/rum-core/src/domain/trackSleep.ts
```diff
@@ -0,0 +1,39 @@
+import { monitor, ONE_MINUTE, ONE_SECOND, TimeStamp, timeStampNow } from '@datadog/browser-core'
+
+export const SLEEP_CHECK_DELAY = ONE_SECOND
+export const SLEEP_THRESHOLD = ONE_MINUTE
+
+let sleepPeriods: Array<{ start: TimeStamp; end: TimeStamp }> | undefined
+let lastWokeDate: TimeStamp | undefined
+
+export function trackSleep() {
+  lastWokeDate = timeStampNow()
```
bcaudan, line 10: nit: 
```suggestion
  lastWoke = timeStampNow()
```

to be consistent with other naming (now, start, end)

bcaudan (approved):
LGTM

PR #979, opened 2021-08-04
🔊 [RUMF-971] add some context on suspicious LCP and FCP monitoring

BenoitZugmeyer:
## Motivation

Investigate suspicious LCP and FCP timings

## Changes

* Add a count of FCP to see if multiple FCP are sent on the same page
* Add a list of sizes for LCP to see if LCP are correctly sent with an ever-increasing size

## Testing

---


packages/rum-core/src/domain/rumEventsCollection/view/trackInitialViewTimings.ts
```diff
@@ -129,12 +134,14 @@ export function trackLargestContentfulPaint(
         entry.startTime < firstInteractionTimestamp &&
         entry.startTime < firstHidden.timeStamp
       ) {
+        lcpSizes.push(entry.size)
```
bcaudan, line 137: Could it be interesting to have entry size + date?

bcaudan (approved):

PR #980, opened 2021-08-04
Configurable sessionUUID generator

cchantep:
## Motivation

Allow to generate a custom session ID, so it can be consistent with logs/traces from other services (to correlate).

## Changes

- Add an optional new setting `generateSessionUuid: () => string` in `Configuration
- If the setting is specified, use it when starting session to generate the ID.

## Testing

Unit test for `rumSession` updated

---

### DIFF
--- a/packages/core/src/domain/configuration.ts
+++ b/packages/core/src/domain/configuration.ts
@@ -80,6 +80,7 @@ export type Configuration = typeof DEFAULT_CONFIGURATION &
 
     service?: string
     beforeSend?: BeforeSendCallback
+    generateSessionUuid?: () => string,
 
     actionNameAttribute?: string
 
--- a/packages/core/src/domain/sessionManagement.ts
+++ b/packages/core/src/domain/sessionManagement.ts
@@ -22,13 +22,19 @@ export interface SessionState {
   [key: string]: string | undefined
 }
 
+type InitialSessionState<TrackingType> = {
+  id?: string
+  trackingType: TrackingType
+  isTracked: boolean 
+}
+
 /**
  * Limit access to cookie to avoid performance issues
  */
 export function startSessionManagement<TrackingType extends string>(
   options: CookieOptions,
   productKey: string,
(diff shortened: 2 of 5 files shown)


bcaudan:
Hello @cchantep,

RUM sessions are automatically correlated with datadog logs and traces.
If you want to correlate RUM sessions with IDs from other providers, have you considered passing them through the [global context](https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm#global-context)?

cchantep:
As far as I can see, RUM sessions are correlated with browser log, not with logs of other services.
In our case, an API service provide user session ID (and all its logs are collected with such session ID attribute), but as this session ID cannot be shared when init'ing the RUM, there is correlation.

bcaudan:
> In our case, an API service provide user session ID (and all its logs are collected with such session ID attribute), but as this session ID cannot be shared when init'ing the RUM, there is correlation.

not sure what you meant with "there is correlation" but you could retrieve the user session id from the API and pass it to the global context while initializing RUM.

How did you plan to use the `generateSessionUuid` option?

cchantep:
Adding to the RUM global context doesn't override/set the session ID, it's added in the RUM custom attributes, but not as "Session Id", so navigating by clicking on this session ID cannot correlated with other service.

Capture d’écran 2021-08-09 à 14 05 29

As indicated in the description, with the changes there we can:

1. Init a user session on another service (server-side session), get its ID.
2. Init RUM on client side, with the very same session ID using `generateSessionUuid: () => apiSessionId` in the configuration.

bcaudan:
> Init a user session on another service (server-side session), get its ID.

Since RUM session is used to bill the product, it is unlikely that we delegate its control to users.

I would suggest you to reach out to [datadog support](https://www.datadoghq.com/support/) to see how the UI could be improve to help you correlate RUM session with other service session.

cchantep:
I don't see how the session ID could be use for billing, or that's quite fragile, as it can be easily forge on the client side.

So this PR still need review, may be with an escalade.

bcaudan:
As suggested before, it will be preferable to go through support for this one.

PR #987, opened 2021-08-09
✨[REPLAY-379] add replay stats on view 

BenoitZugmeyer:
## Motivation

Allows replay statistics aggregation over sessions.

## Changes

This PR adds a `_dd.replay` field on views, populated by `trackViews` via lifecycle events sent by the `segmentCollection`.

Those stats are also used to simplify the `session.has_replay` field on views: we are guaranteed that if a view has at least one segment, it will have a replay. We don't need to maintain a `isRecording` state based on start and stop recording events.

## Testing

Unit, manual

---


packages/rum-core/src/domain/lifeCycle.ts
```diff
@@ -21,8 +21,7 @@ export enum LifeCycleEventType {
   BEFORE_UNLOAD,
   RAW_RUM_EVENT_COLLECTED,
   RUM_EVENT_COLLECTED,
-  RECORD_STARTED,
-  RECORD_STOPPED,
+  REPLAY_STATS_UPDATED,
```
amortemousque, line 24: IMO, it doesn't feel like a life cycle event. Former once felt better in that sense. Could we do the same without a life cycle event?

packages/rum-core/src/domain/rumEventsCollection/view/trackReplayStats.ts
```diff
@@ -0,0 +1,20 @@
...
+import { LifeCycle, LifeCycleEventType } from '../../lifeCycle'
+
+export function trackReplayStats(lifeCycle: LifeCycle, viewId: string, callback: (stats: ViewReplayStats) => void) {
+  const replayStats: ViewReplayStats = {
+    records_count: 0,
+    segments_count: 0,
+    segments_total_raw_size: 0,
+  }
+
+  const { unsubscribe: stop } = lifeCycle.subscribe(LifeCycleEventType.REPLAY_STATS_UPDATED, (update) => {
```
amortemousque, line 11: My mental model when reasoning about `trackers` is: 
`tracker` -> notify through `life cycle event or observable` -> `collection`. 
I know there is currently no real rules about it but don't you think it could be interesting to think of it?

packages/rum-core/src/domain/lifeCycle.ts
bcaudan: It feels a bit weird to me too.

BenoitZugmeyer: Would three events like `REPLAY_SEGMENT_CREATED` `REPLAY_RECORD_ADDED` and `REPLAY_BYTES_WRITTEN` do?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -217,6 +204,10 @@ function newView(
     loadingType
   )
 
+  const { stop: stopReplayStatsTracking } = trackReplayStats(lifeCycle, id, (newReplayStats) => {
```
bcaudan, line 207: It would be nice to be able to add replay metadata to views in view collection, something like:

```ts
function processViewUpdate(
  view: ViewEvent,
  foregroundContexts: ForegroundContexts,
  replayStats: any,
): RawRumEventCollectedData<RawRumViewEvent> {

const viewEvent = {
  ...
  ...replayStats.get()
}
...
}
```

packages/rum-core/src/domain/lifeCycle.ts
bcaudan: I would be in favor of not using Lifecycle events for this use case.
Maybe a stateful stats object used as a dependency by both view and record logic.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
BenoitZugmeyer: Okay, this was actually my first implementation. With this approach, we'd need to make sure the updated view is the one being recorded, which is not necessarily the case. My solution was to have a `getReplayStats(viewId)` function, to be sure to get the corresponding stats, but then we need to keep the stats indexed by view id just in case we need it again it the future, so it leaks memory.

Could we talk about it?

amortemousque: Maybe instead of a stat object, you could use a replay observable. Would it solve the issue?

PR #990, opened 2021-08-09
Limit log monitoring + add more details

ThibautGeriz:
## Motivation

Too much logging in the foregroundContext manager

## Changes

- do not log error if two events are coming real close: https://bugs.chromium.org/p/chromium/issues/detail?id=1237904
- add now to extimate the user impact better

### DIFF
--- a/packages/rum-core/src/domain/foregroundContexts.ts
+++ b/packages/rum-core/src/domain/foregroundContexts.ts
@@ -14,7 +14,9 @@ import { InForegroundPeriod } from '../rawRumEvent.types'
 
 // Arbitrary value to cap number of element (mostly for backend)
 export const MAX_NUMBER_OF_FOCUSED_TIME = 500
-const MAX_LAG_TIME = 10 as RelativeTime
+// ignore duplicate focus & blur events if coming in the right after the previous one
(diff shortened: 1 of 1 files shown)


packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -14,6 +14,7 @@ import { InForegroundPeriod } from '../rawRumEvent.types'
 
 // Arbitrary value to cap number of element (mostly for backend)
 export const MAX_NUMBER_OF_FOCUSED_TIME = 500
+const MAX_LAG_TIME = 10 as RelativeTime
```
bcaudan, line 17: Can you add a comment to explain your limiting strategy?

ThibautGeriz: This is because of this [chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1237904). It currently dispatches two events separated by ~1ms. I've set it at 10ms to be large & because if our foreground periods are off by 10ms we do not really care.
I also added the now in the log to actually get an insight by how much we're off. I am not really fixing the base issue, just ignoring this chrome bug & adding monitoring info.

PR #994, opened 2021-08-11
✨ [REPLAY-379] add replay stats on view (getter edition)

BenoitZugmeyer:
## Motivation

Allows replay statistics aggregation over sessions.

This PR supersed https://github.com/DataDog/browser-sdk/pull/987 with a different approach: uses a "getter" instead of lifecycle events.

## Changes

This PR adds a `_dd.replay` field on views, populated by `trackViews` via lifecycle events sent by the `segmentCollection`.

Those stats are also used to simplify the `session.has_replay` field on views: we are guaranteed that if a view has at least one segment, it will have a replay. We don't need to maintain a `isRecording` state based on start and stop recording events.

## Testing

Unit, manual
---


packages/rum/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -135,24 +137,34 @@ export function doStartSegmentCollection(
       return
     }
 
+    const viewStats = getOrCreateViewStats(context.view.id)
```
amortemousque, line 140: Might be just me, but it feels like with stats calculation `createNewSegment` does more that it should. Maybe it's just the naming or we can split the function?

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -101,6 +101,7 @@ export interface RawRumViewEvent {
   }
   _dd: {
     document_version: number
+    replay?: ViewReplayStats
```
bcaudan, line 104: it seems that we have several namings representing the same thing:

- replay
- viewStats
- viewReplayStats

could we use the same naming everywhere?

packages/rum/src/domain/segmentCollection/segmentCollection.ts
bcaudan: agree, could we move the calculation to the view stats file?

packages/rum/src/domain/segmentCollection/segment.spec.ts
```diff
@@ -143,4 +144,42 @@ describe('Segment', () => {
...
+    })
+
+    it('when adding a record', () => {
+      const segment = new Segment(worker, CONTEXT, 'init', FULL_SNAPSHOT_RECORD, noop, noop)
+      segment.addRecord(RECORD)
+      worker.processAllMessages()
+      expect(getViewStats('b')).toEqual({
+        segments_count: 1,
+        records_count: 2,
+        segments_total_raw_size: 73,
```
bcaudan, line 181: it could be nice to clarify somehow where 47, 202 and 73 come from

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -109,4 +115,24 @@ describe('viewCollection', () => {
...
     })
   })
+
+  it('should include replay information if available', () => {
+    const { lifeCycle, rawRumEvents } = setupBuilder.build()
+    getViewStatsSpy.and.callFake((viewId) => {
+      if (viewId === VIEW.id) {
+        return { segments_count: 4, records_count: 10, segments_total_raw_size: 1000 }
+      }
+    })
```
bcaudan, line 125: is the check on view id needed?

packages/rum/src/domain/segmentCollection/segment.ts
```diff
@@ -26,12 +29,19 @@ export class Segment {
     this.recordsCount = 1
     this.hasFullSnapshot = initialRecord.type === RecordType.FullSnapshot
 
+    this.viewStats = getOrCreateViewStats(context.view.id)
+    this.viewStats.segments_count += 1
+    this.viewStats.records_count += 1
```
bcaudan, line 34: Could we move the computation in view stats file?

BenoitZugmeyer: What are you thinking about? Instead of having a `getOrCreateViewStats`, three functions like `incrementSegmentsCount(viewId)`, `incrementRecordsCount(viewId)` and `incrementRawSize(viewId)` ?

packages/rum-core/src/rawRumEvent.types.ts
bcaudan: what about `replayStats` everywhere?

BenoitZugmeyer: Ok!

packages/rum/src/domain/segmentCollection/segment.ts
amortemousque: Feels weird to have function like addSegment(viewId) which does not really add a segment. Wouldn't it make more sense to mimic the RUM logic and have a `replayLifeCycle` with `segmentCreated`, `recordCreated` and given that object the `viewStats` are computed?
Plus, it's easily extendable for future similar use cases.

packages/rum/src/domain/segmentCollection/segment.ts
```diff
@@ -12,31 +14,39 @@ export class Segment {
...
+
+    let previousRawSize = 0
     const listener: DeflateWorkerListener = monitor(({ data }) => {
       if ('error' in data) {
         return
       }
 
       if (data.id === this.id) {
+        this.viewStats.segments_total_raw_size += data.rawSize - previousRawSize
```
webNeat, line 43: I am trying to understand why do we need this computation. is there a way to have the additional size directly in `data` instead of having to compute it each time? This would get rid of `previousRawSize` and reduce complexity I think.

packages/rum/src/domain/segmentCollection/segmentCollection.spec.ts
```diff
@@ -214,6 +216,47 @@ describe('startSegmentCollection', () => {
...
+      lifeCycle.notify(LifeCycleEventType.BEFORE_UNLOAD)
+      worker.processAllMessages()
+      expect(getViewStats('b')).toEqual({
+        segments_count: 1,
+        records_count: 1,
+        segments_total_raw_size: 193,
+      })
+    })
+
+    it('when creating a new segment', () => {
```
webNeat: I like to think of tests as being independant, but here this test depends on the one before it to give the expected results. is there a way to remove this implicit dependency ?

webNeat: Sorry, I was wrong, there is no dependency after all.
I didn't see the second call to `addRecord` here and thought it was counting the record from the previous test to give `records_count: 2`

packages/rum-core/src/domain/rumEventsCollection/view/viewCollection.spec.ts
```diff
@@ -109,4 +115,21 @@ describe('viewCollection', () => {
       },
     })
   })
+
+  it('should include replay information if available', () => {
+    const { lifeCycle, rawRumEvents } = setupBuilder.build()
+    getReplayStatsSpy.and.callFake(() => ({ segments_count: 4, records_count: 10, segments_total_raw_size: 1000 }))
```
bcaudan, line 121: nit
```suggestion
    getReplayStatsSpy.and.returnValue({ segments_count: 4, records_count: 10, segments_total_raw_size: 1000 })
```

bcaudan (approved):
LGTM

webNeat (approved):

PR #969, opened 2021-08-02
✨ [RUMF-974] use user-agent to detect synthetics sessions

bcaudan:
## Motivation

Synthetics browser tests now add a specific value in the user agent to ease the identification, cf [doc](https://docs.datadoghq.com/synthetics/guide/identify_synthetics_bots/?tab=browsertests)

## Changes

- Check user agent instead of global variable
- Check only once

## Testing

unit test
Need to check on staging

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #997, opened 2021-08-12
🐛[RUMF-990] restore global check to detect synthetics sessions

bcaudan:
## Motivation

User agent used by synthetics can be overwritten by the customer. So user agent check is not enough.

## Changes

Restore check on global. 
Previous PR:
#969 

## Testing

unit

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -29,6 +29,10 @@ import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { ParentContexts } from './parentContexts'
 import { RumSession, RumSessionPlan } from './rumSession'
 
+export interface BrowserWindow extends Window {
+  _DATADOG_SYNTHETICS_BROWSER?: unknown
+}
+
 enum SessionType {
   SYNTHETICS = 'synthetics',
   USER = 'user',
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer (approved):

packages/rum-core/src/domain/assembly.ts
```diff
@@ -88,7 +90,8 @@ export function startRumAssembly(
           date: timeStampNow(),
           service: configuration.service,
           session: {
-            type: sessionType,
+            // must be computed on each event because synthetics instrumentation can be done after sdk execution
```
webNeat, line 93: Does that mean that if the first events happen before the synthetics instrumentation, they will be sent with the wrong session type?

bcaudan: yes, it was the previous behavior

webNeat: We currently have a RFC that tries to relate RUM events to Synthetics tests, I think some solutions from that RFC could be used for this use case too. In particular, I like the idea of using a `Set-Cookie` header to pass data from Synthetics to the SDK. So we could set the session type based on the presence of a cookie.
Do you think it worth exploring that solution? may be as a follow-up to this PR?

bcaudan: We already discussed this topic during our daily yesterday, the solution describe in the RFC could indeed make the synthetics sessions detection more reliable. However, since this solution is not yet available, the goal of this PR is to fix the issue introduced with the previous change.

PR #1001, opened 2021-08-13
Add monitoring on untrusted events

ThibautGeriz:
## Motivation

Add check as part of investigating error in SDK

https://docs.google.com/document/d/17FsROHTNE8hh_Ra6blCgNwyZn8m2PqRefS_WIWjfBSk/edit#

## Testing

---


BenoitZugmeyer (commented):
What about discrading untrusted events directly instead of monitoring them?

ThibautGeriz:
> What about discarding untrusted events directly instead of monitoring them?

I would like to do it in two steps, discarding then would be the second one. I will eventually discard them but first let's see if it happens

BenoitZugmeyer (approved):

PR #1004, opened 2021-08-19
✨ [RUMF-998] introduce the initialPrivacyLevel configuration option 

BenoitZugmeyer:
## Motivation

Allow the user to set a default privacy level for session replay

## Changes

This PR is changing a bit the design of the Privacy by Default implementation #914  . 

Mainly, it removes the confusing distinct types `NodePrivacyLevel` and `NodePrivacyLevelInternal`, to allow simpler inheritance of the privacy level.

It also make  the privacy level a require argument for main recording functions, to allow passing an initial privacy level.

Finally, it implements the new configuration option publicly.

## Testing

---


packages/core/src/domain/configuration.ts
```diff
@@ -143,6 +152,13 @@ export function buildConfiguration(initConfiguration: InitConfiguration, buildEn
     configuration.actionNameAttribute = initConfiguration.actionNameAttribute
   }
 
+  if (
+    'initialPrivacyLevel' in initConfiguration &&
+    objectValues(InitialPrivacyLevel).indexOf(initConfiguration.initialPrivacyLevel) >= 0
+  ) {
+    configuration.initialPrivacyLevel = initConfiguration.initialPrivacyLevel!
```
ThibautGeriz, line 159: This code is super cryptic to me, I would refactor or at least add a comment.
From my understanding, since the user is not using typescript we need to check at runtime that the `initialPrivacyLevel` is in the config & it's a valid value. We cannot use [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) because of IE11 ?
Then, we add with a `!` I do not get it. It's supposed to be a string

ThibautGeriz (commented):
I don't think I can either approve or refuse this PR but it _looks_ good.

packages/core/src/domain/configuration.ts
webNeat: The `!` is there to tell Typescript that we are sure that `initConfiguration.initialPrivacyLevel` is not `undefined`. Otherwise it will complain about assigning `undefined` to a field of type `InitialPrivacyLevel`.

packages/rum/src/constants.ts
```diff
@@ -1,22 +1,15 @@
 export const NodePrivacyLevel = {
```
webNeat, line 1: Is it possible to reuse the `InitialPrivacyLevel` constant here and extend it like?
```ts
export const NodePrivacyLevel = {
  ... InitialPrivacyLevel,
  IGNORE: 'ignore',
  HIDDEN: 'hidden',
}
```

packages/rum/src/domain/record/serialize.ts
```diff
@@ -30,18 +24,29 @@ import {
 } from './serializationUtils'
 import { forEach } from './utils'
 
+// Those values are the only one that can be used when inheriting privacy levels from parent to
+// children during serialization, since HIDDEN and IGNORE shouldn't serialize their children. This
+// ensures that no children are serialized when they shouldn't.
+type ParentNodePrivacyLevel =
```
webNeat, line 30: Looks like this is the same as the `InitialPrivacyLevel` type

packages/core/src/domain/configuration.ts
```diff
@@ -1,9 +1,16 @@
...
-import { includes, ONE_KILO_BYTE, ONE_SECOND } from '../tools/utils'
+import { includes, objectValues, ONE_KILO_BYTE, ONE_SECOND } from '../tools/utils'
 import { computeTransportConfiguration } from './transportConfiguration'
 
+export const InitialPrivacyLevel = {
+  ALLOW: 'allow',
+  MASK: 'mask',
+  MASK_FORMS_ONLY: 'mask-forms-only',
+} as const
+export type InitialPrivacyLevel = typeof InitialPrivacyLevel[keyof typeof InitialPrivacyLevel]
```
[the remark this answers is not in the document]
BenoitZugmeyer: We chose to avoid exposing enums publicly especially to support CDN setups. We want to let the customer use plain values (ex: `'allow'`) so they don't need to import them (ex: in CDN setups importing stuff is not ideal). Also, when using TS definitions + a CDN setup like we do on web-ui, we can't import the enum value with`import type { SomeEnum } from "browser-rum"`. Because we cannot use this specific enum, typecheck fails. (Note: we can't do `import { SomeEnum } from  "browser-rum"` because our package have side effects that would be triggered by importing it)

packages/core/src/domain/configuration.ts
BenoitZugmeyer: That's right, we can't use `includes` because of IE11. `indexOf >= 0` is super idiomatic in JS, so I thought that was fine. I'll see what I can do to improve the situation.

packages/rum/src/constants.ts
BenoitZugmeyer: It is!

packages/rum/src/domain/record/serialize.ts
BenoitZugmeyer: Yes, but it is somewhat coincidental. The `InitialPrivacyLevel` is the privacy levels that we want to expose publicly, and `ParentNodePrivacyLevel` are the privacy levels that can be inherited during serialization, so they may not always be the same in the future. To avoid confusion, I'd keep those two separated.

webNeat: Makes sense.

amortemousque:
For the overall understanding of the recorder, it might be interesting to split the observer.ts file like so for instance:
/initObservers.ts
/observers
  /mutationObserver.ts
  /moveObserver.ts
  /scrollObserver.ts
  ...
Same logic could be applied for serialize.ts:
/serializers
 /serializeDocumentNode.ts
 /serializeElementNode.ts
/serializeAttribute.ts
...
Of course, it should be done in another PR :) 
What do you think?

packages/rum/src/domain/record/observer.ts
```diff
@@ -182,7 +194,7 @@ export function initInputObserver(cb: InputCallback): ListenerHandler {
 
     let inputState: InputState
     if (type === 'radio' || type === 'checkbox') {
-      if (nodePrivacyLevel === NodePrivacyLevel.MASK) {
+      if (shouldMaskNode(target, nodePrivacyLevel)) {
```
amortemousque, line 197: I don't really inderstand why we need `shouldMaskNode()`, if we already have `getNodePrivacyLevel()`
There is probably a confusion between the **Mask** term in `shouldMaskNode()` and the `NodePrivacyLevel.MASK`

jagracey:
> > For the overall understanding of the recorder, it might be interesting to split the observer.ts file like so for instance:
> > ```
> > /initObservers.ts
> > /observers
> >   /mutationObserver.ts
> >   /moveObserver.ts
> >   /scrollObserver.ts
> >   ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Same logic could be applied for serialize.ts:
> > ```
> > /serializers
> >   /serializeDocumentNode.ts
> >   /serializeElementNode.ts
> >   /serializeAttribute.ts
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ...
> > Of course, it should be done in another PR :)
> > What do you think?
> 
> I agree, we can split things up in another PR.

I'd agree with the first case, `observer`, particularly `mutationObserver`, however the serialisers (document, element, attribute) to me represent code that should be kept together.

PR #1007, opened 2021-08-20
 ✨ [RUMF-996] set synthetics ids on RUM events 

BenoitZugmeyer:
## Motivation

Allow linking RUM to Synthetics tests

## Changes

Add Synthetics test id and result id to all RUM events based on global variables. In the future, it will also use cookies, but since we don't know yet how they will be defined, let's do only global variables first.

## Testing

Manual, Unit

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.spec.ts
+++ b/packages/rum-core/src/domain/assembly.spec.ts
@@ -51,6 +51,7 @@ describe('rum assembly', () => {
 
   afterEach(() => {
     setupBuilder.cleanup()
+    cleanupSyntheticsGlobals()
   })
 
   describe('beforeSend', () => {
@@ -516,8 +517,7 @@ describe('rum assembly', () => {
     })
 
     it('should detect synthetics sessions from global', () => {
-      ;(window as BrowserWindow)._DATADOG_SYNTHETICS_PUBLIC_ID = 'foo'
-      ;(window as BrowserWindow)._DATADOG_SYNTHETICS_RESULT_ID = 'bar'
+      setSyntheticsGlobals('foo', 'bar')
 
       const { lifeCycle } = setupBuilder.build()
       notifyRawRumEvent(lifeCycle, {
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -516,7 +516,8 @@ describe('rum assembly', () => {
     })
 
     it('should detect synthetics sessions from global', () => {
-      ;(window as BrowserWindow)._DATADOG_SYNTHETICS_BROWSER = true
+      ;(window as BrowserWindow)._DATADOG_SYNTHETICS_PUBLIC_ID = 'foo'
+      ;(window as BrowserWindow)._DATADOG_SYNTHETICS_RESULT_ID = 'bar'
```
webNeat, line 520: I would create a utility function `setSyntheticsGlobals('foo', 'bar')` or similar to avoid repetition and make the test more expressive.
I would do the same and delete them with a `deleteSyntheticsGlobals()` utility function.

packages/rum-core/src/domain/assembly.ts
```diff
@@ -162,8 +164,19 @@ function needToAssembleWithAction(
...
-  return navigator.userAgent.indexOf('DatadogSynthetics') === -1 &&
-    (window as BrowserWindow)._DATADOG_SYNTHETICS_BROWSER === undefined
+  return navigator.userAgent.indexOf('DatadogSynthetics') === -1 && !getSyntheticsContext()
     ? SessionType.USER
     : SessionType.SYNTHETICS
 }
+
+function getSyntheticsContext() {
+  const testId = (window as BrowserWindow)._DATADOG_SYNTHETICS_PUBLIC_ID
+  const resultId = (window as BrowserWindow)._DATADOG_SYNTHETICS_RESULT_ID
```
webNeat, line 174: Should we handle the case when some of these variables is not a string? I know it shouldn't normally happen but if it does, what would be the impact? generating the events with wrong format, how would the backend handle it?

packages/rum-core/src/rumEvent.types.ts
```diff
@@ -14,7 +14,7 @@ export type RumActionEvent = CommonProperties & {
   /**
    * RUM event type
    */
-  readonly type: string
+  readonly type: 'action'
```
webNeat, line 17: This is not an issue, but I am just curious how did this type change from `string` to `'action'`? did we change the something in the tool that generates the types?

BenoitZugmeyer: Yes! we updated our `json-schema-to-typescript` fork with the  latest changes from upstream some time ago: https://github.com/bcaudan/json-schema-to-typescript/pull/1

webNeat (approved):
Looking good!

PR #1009, opened 2021-08-24
🐛 prevent recording to start when session renewed before onload

BenoitZugmeyer:
## Motivation

When the session is renewed before onload, there may be a case where the recording is incorrectly started.
This shouldn't happen that often, but it bothered me.

## Changes

Stop the recording during session renew if it is `starting`.

## Testing

CI

---


webNeat (approved):
Looks good to me.

amortemousque (approved):

ThibautGeriz (approved):

PR #1011, opened 2021-08-24
🐛 [RUMF-997]  dont take a FullSnapshot on view creation during session renew

BenoitZugmeyer:
## Motivation

When renewing the session from replay to lite (for example), a full snapshot was taken on the lite session, because the recording was stopped a tiny bit too late (after the new view creation)

## Changes

This PR introduces a new lifecycle event `SESSION_EXPIRED` to have a chance to do some cleanup before the session is renewed. It uses it to stop recording.

## Testing

Manual, e2e (no unit test because we don't have any test at this level: it involves the whole RUM and recording stack)

---

### DIFF
--- a/packages/rum-core/src/domain/lifeCycle.ts
+++ b/packages/rum-core/src/domain/lifeCycle.ts
@@ -17,6 +17,19 @@ export enum LifeCycleEventType {
   VIEW_ENDED,
   REQUEST_STARTED,
   REQUEST_COMPLETED,
+
+  // The SESSION_EXPIRED lifecycle event has been introduced to represent when a session has expired
+  // and trigger cleanup tasks related to this, prior to renewing the session. Its implementation is
+  // slightly naive: it is not triggered as soon as the session is expired, but rather just before
+  // notifying that the session is renewed. Thus, the session id is already set to the newly renewed
+  // session.
+  //
+  // This implementation is "good enough" for our use-cases. Improving this is not trivial,
+  // primarily because multiple instances of the SDK may be managing the same session cookie at
+  // the same time, for example when using Logs and RUM on the same page, or opening multiple tabs
+  // on the same domain.
+  SESSION_EXPIRED,
+
   SESSION_RENEWED,
   BEFORE_UNLOAD,
   RAW_RUM_EVENT_COLLECTED,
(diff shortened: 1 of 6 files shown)


webNeat (approved):

packages/rum/src/boot/recorderApi.ts
```diff
@@ -62,12 +62,14 @@ export function makeRecorderApi(startRecordingImpl: StartRecording): RecorderApi
       session: RumSession,
       parentContexts: ParentContexts
     ) => {
-      lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
+      lifeCycle.subscribe(LifeCycleEventType.SESSION_EXPIRED, () => {
         if (state.status === RecorderStatus.Started) {
           stopStrategy()
           state = { status: RecorderStatus.IntentToStart }
```
ThibautGeriz, line 68: If a session expired, we might not need to restart it if it's now `LITE` no?

packages/rum-core/src/domain/lifeCycle.ts
```diff
@@ -17,6 +17,19 @@ export enum LifeCycleEventType {
...
+  // The SESSION_EXPIRED lifecycle event has been introduced to represent when a session has expired
+  // and trigger cleanup tasks related to this, prior to renewing the session. Its implementation is
+  // slightly naive: it is not triggered as soon as the session is expired, but rather just before
+  // notifying that the session is renewed. Thus, the session id is already set to the newly renewed
+  // session.
+  //
+  // This implementation is "good enough" for our use-cases. Improving this is not trivial,
+  // primarily because multiple instances of the SDK may be managing the same session cookie at
+  // the same time, for example when using Logs and RUM on the same page, or opening multiple tabs
+  // on the same domain.
```
amortemousque, line 30: Good comment!

packages/rum/src/boot/recorderApi.ts
amortemousque: I think what you're referring to is a few lines below in the `startStrategy`
https://github.com/DataDog/browser-sdk/blob/b38ae1d939b5d90d48595390843976be1c5d4871/packages/rum/src/boot/recorderApi.ts#L78-L82

amortemousque (approved):

PR #1016, opened 2021-08-27
✨ [RUMF-993] New proxy strategy

amortemousque:
## Motivation

Some customer proxy configurations rely on the intake v1 path to root the requests. For instance they may use a wildcard to identify RUM like so `/v1/input/`. By enabling the intake version 2, the path would become `/api/v2/...` and these customers would break their proxy configuration.

In order to guarantee backward compatibility for future intake versions, we need to change our approach. 

## Changes

- Add a new **proxyUrl** option parameter
- Implement the new strategy

## Testing

---

### DIFF
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -98,11 +98,6 @@ export function createEndpointBuilder(
 
     let parameters = `ddsource=${source || 'browser'}&ddtags=${encodeURIComponent(tags)}`
 
-    if (proxyHost) {
-      const datadogHost = buildHost(endpointType)
-      parameters += `&ddhost=${datadogHost}`
-    }
-
     if (shouldUseIntakeV2(endpointType)) {
       parameters +=
         `&dd-api-key=${clientToken}&` +
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -46,6 +46,7 @@ export function createEndpointBuilder(
     clientToken,
     env,
     proxyHost,
+    proxyUrl,
     service,
     version,
     intakeApiVersion,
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -99,6 +89,30 @@ export function createEndpointBuilder(
...
+      `${env ? `,env:${env}` : ''}` +
+      `${service ? `,service:${service}` : ''}` +
+      `${version ? `,version:${version}` : ''}`
+
+    let parameters = `ddsource=${source || 'browser'}&ddtags=${encodeURIComponent(tags)}`
+
+    if (proxyHost) {
+      const datadogHost = buildHost(endpointType)
+      parameters += `&ddhost=${datadogHost}`
+    }
```
BenoitZugmeyer, line 104: It seems like you are adding the `ddhost` parameter twice (once here and ance in `build()`). Maybe add a test to check the full URL instead of `toContain` / `toMatch`?

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -46,41 +46,31 @@ export function createEndpointBuilder(
...
-    if (proxyHost) {
-      const datadogHost = buildHost(endpointType)
-      parameters += `&ddhost=${datadogHost}`
+    const host = buildHost(endpointType)
+    const path = buildPath(endpointType)
+    const queryParameters = buildQueryParameters(endpointType, source)
+    const endpoint = `https://${host}${path}?${queryParameters}`
+
+    if (proxyUrl) {
+      return `${proxyUrl}?ddforward=${encodeURIComponent(endpoint)}`
```
BenoitZugmeyer, line 63: You could move the `endpoint` variable declaration in the `if`

BenoitZugmeyer (approved):

webNeat (approved):

PR #1018, opened 2021-08-30
🐛 [RUMF-1005] Fix dd-request-id endpoint query param

amortemousque:
## Motivation

Fix intake v2 `dd-request-id` to be unique per request

## Changes

Add dd-request-id  to HttpRequest 

## Testing

Locally, Unit
---

### DIFF
--- a/packages/core/src/transport/httpRequest.spec.ts
+++ b/packages/core/src/transport/httpRequest.spec.ts
@@ -86,6 +86,14 @@ describe('httpRequest parameters', () => {
     expect(sendBeaconSpy.calls.argsFor(0)[0]).toContain(`&batch_time=`)
   })
 
+  it('should not add batch_time', () => {
+    const request = new HttpRequest('https://my.website', BATCH_BYTES_LIMIT, false)
+
+    request.send('{"foo":"bar1"}', 10)
+
+    expect(sendBeaconSpy.calls.argsFor(0)[0]).not.toContain(`&batch_time=`)
+  })
+
   it('should have dd-request-id', () => {
     const request = new HttpRequest('https://my.website', BATCH_BYTES_LIMIT)
 
--- a/packages/core/src/transport/httpRequest.spec.ts
+++ b/packages/core/src/transport/httpRequest.spec.ts
@@ -1,7 +1,5 @@
 /* eslint-disable @typescript-eslint/unbound-method */
 import sinon from 'sinon'
-import { BuildEnv, BuildMode } from '../boot/init'
-import { createEndpointBuilder } from '../domain/configuration/endpointBuilder'
 import { HttpRequest } from './httpRequest'
 
 describe('httpRequest', () => {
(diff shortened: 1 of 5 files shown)


packages/core/src/transport/httpRequest.ts
```diff
@@ -67,6 +77,14 @@ function addBatchTime(url: string) {
   return `${url}${url.indexOf('?') === -1 ? '?' : '&'}batch_time=${new Date().getTime()}`
 }
 
+function addRequestId(url: string) {
+  return `${url}${url.indexOf('?') === -1 ? '?' : '&'}dd-request-id=${generateUUID()}`
+}
+
+function isEndpointIntakeV2(url: string) {
+  return includes(url, '/api/v2')
```
BenoitZugmeyer, line 85: NIT: Since you are using `includes` here, maybe you could use it instead of `indexOf` above? (in addBatchTime too)

packages/core/src/transport/httpRequest.ts
```diff
@@ -14,7 +15,16 @@ export class HttpRequest {
   constructor(private endpointUrl: string, private bytesLimit: number, private withBatchTime: boolean = false) {}
 
   send(data: string | FormData, size: number) {
-    const url = this.withBatchTime ? addBatchTime(this.endpointUrl) : this.endpointUrl
+    let url = this.endpointUrl
+
+    if (isEndpointIntakeV2(url)) {
```
BenoitZugmeyer, line 20: Checking if we are using the intake v2 at this level is a bit unexpected.

What would be the impact of always adding the request id? Does the "intake v1" ignore it or not?

amortemousque: The intake ignores it, but it pass it along to rum backend. We have the issue with (also with intake v2) of the `dd-request-id` being in the JSON object section of the frontend. Once it's fixed, that should not be an issue.

BenoitZugmeyer: So, maybe we could wait until the issue is fixed? `intake v2` is still behind a flag anyway.

bcaudan: what about handling that with a `withRequestId` constructor param depending on intakeApiVersion config + FF?

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -63,3 +65,49 @@ describe('httpRequest', () => {
...
+    const request = new HttpRequest('https://my.website', BATCH_BYTES_LIMIT, true)
+
+    request.send('{"foo":"bar1"}', 10)
+
+    expect(sendBeaconSpy.calls.argsFor(0)[0]).toContain(`batch_time=`)
+  })
+
+  it('should add dd-request-id query attribute when the intake v2 enabled', () => {
+    const endpointBuilder = createEndpointBuilder({ clientToken, intakeApiVersion: 2 }, buildEnv, true)
+    const request = new HttpRequest(endpointBuilder.build('rum'), BATCH_BYTES_LIMIT)
```
bcaudan, line 98: what about avoid to make the test dependent of the endpoint builder logic?

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -63,3 +63,35 @@ describe('httpRequest', () => {
...
+    navigator.sendBeacon = sendBeaconSpy
+  })
+
+  afterEach(() => {
+    server.restore()
+  })
+
+  it('should add batch_time', () => {
+    const request = new HttpRequest('https://my.website', BATCH_BYTES_LIMIT, true)
+
```
bcaudan, line 83: what about having a test ensuring that we don't have batch time when `withBatchTime=false`?

bcaudan (approved):

PR #1020, opened 2021-08-30
👷 Replace istanbul-instrumenter-loader with coverage-istanbul-loader

bpinto:
## Motivation

To keep up to date with latest versions of all libraries which include bug fixes and improvements.

## Changes

This change removes a no longer supported library by replacing it with another.

## Testing

I ran both `yarn test:unit` and `yarn test:e2e`.

---


BenoitZugmeyer:
Thanks for reporting this. It looks like `coverage-istanbul-loader` is [not really maintained either](https://github.com/JS-DevTools/coverage-istanbul-loader/issues/8)... but still it may be better than the old one? At least it will fix the yarn warning 
```
warning " > istanbul-instrumenter-loader@3.0.1" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".
```
since it doesn't specify a webpack version. We'll take a look.

bpinto:
I agree it's not a perfect replacement but I could not find anything newer. I might look into opening a PR that bumps webpack to v5 later.

But as you said, it does help with some existing warnings.

BenoitZugmeyer:
This PR is superseded by #1023 (branches need to be on our own repository so the CI can be executed)

PR #1021, opened 2021-08-31
🐛[REPLAY-371] Truncate long "data:" URIs

jagracey:
## Motivation
A customer is embedding a large "data:" URI resource. This impacts performance without sufficient corresponding benefit to this customer. As a general quick fix, we shall truncate this attribute value. We do not wish to generalise recorder limits at this time, so have specifically targeted only attributes starting with "data:" exceeding 100,000 length (somewhat arbitrarily chosen, but corresponds to 5*X < 1MB).

## Testing
New tests added.

---

### DIFF
--- a/packages/rum/src/domain/record/privacy.spec.ts
+++ b/packages/rum/src/domain/record/privacy.spec.ts
@@ -397,12 +397,14 @@ describe('shouldMaskNode', () => {
 })
 
 describe('serializeAttribute ', () => {
-  it('truncates "data:" URIs after 32,000 string length', () => {
+  it('truncates "data:" URIs after 100,000 string length', () => {
     const node = document.createElement('p')
 
-    const maxAttributeValue = `data:${'a'.repeat(32000 - 5)}`
-    const exceededAttributeValue = `data:${'a'.repeat(32001 - 5)}`
-    const ignoredAttributeValue = `foos:${'a'.repeat(32001 - 5)}`
+    const longString = new Array(100000 - 5).fill('a').join('');
+    const maxAttributeValue = `data:${(longString)}`
+    const exceededAttributeValue = `data:${(longString)}1`
+    const ignoredAttributeValue = `foos:${longString}`
+
     node.setAttribute('test-okay', maxAttributeValue)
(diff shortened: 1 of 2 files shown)


packages/rum/src/domain/record/privacy.spec.ts
```diff
@@ -389,3 +395,26 @@ describe('shouldMaskNode', () => {
     })
   })
 })
+
+describe('serializeAttribute ', () => {
```
BenoitZugmeyer, line 399: Could you take some time to move `serializeAttribute` function and tests in `serialize.ts`? i think it would be better suited than `privacy.ts`, especially with this improvement where it is not really related to privacy.

packages/rum/src/domain/record/privacy.ts
```diff
@@ -187,10 +187,16 @@ export function serializeAttribute(
     }
   }
 
-  // Rebuild absolute URLs from relative (without using <base> tag)
   if (!attributeValue || typeof attributeValue !== 'string') {
     return attributeValue
   }
+
+  // Minimum Fix for customer.
+  if (attributeValue.length > 100_000 && attributeValue.slice(0, 5) === 'data:') {
```
BenoitZugmeyer, line 195: `100_000` should be a constant like `MAX_DATA_ATTRIBUTE_LENGTH`, and be used in the tests too instead of `100000` to be more explicit

jagracey: Good catch.

packages/rum/src/domain/record/privacy.spec.ts
jagracey: Yeah this should be cleaned up. Will tackle in a separate cleanup PR next.

BenoitZugmeyer (approved):

packages/rum/src/domain/record/privacy.spec.ts
jagracey: Addressed in PR https://github.com/DataDog/browser-sdk/pull/1033

Issue #1022, opened 2021-08-31
Datadog logs sends many request when using it together with the RUM SDK.


radualex:
I have the following issue:

**Trigger**

An uncaught exception is being thrown and datadog logs SDK sends request to log the incident.

**Expected outcome**

One request is sent per incident and the incident is logged only once in Datadog UI.

**Actual outcome**

Datadog logs SDK sends many requests per incident (between 1k-2.5k) and the incident is logged many times in Datadog UI.

**Additional Information**

When disabling the Datadog RUM SDK, then the Datadog logs SDK behaves as expected (logs the incident once). However, I want to run them both, so this is not an option at the moment.

I am using version `3.1.3` for both `@datadog/browser-logs` and `@datadog/browser-rum` packages.

Here's a screenshot to illustrate the issue: (1 error thrown, many requests sent => the error is logged many times in Datadog UI)

[Many requests being sent for one uncaught exception example][1]

This is the code I am using to initialise both logs and RUM SDKs:

        import { datadogLogs } from '@datadog/browser-logs';
        import { datadogRum } from '@datadog/browser-rum';
    
        if (process.env.NODE_ENV === 'production' && process.env.DATADOG_CLIENT_TOKEN) {
                const environment = getEnvironment();
                const config = {
                    site: 'datadoghq.eu',
                    clientToken: process.env.DATADOG_CLIENT_TOKEN,
                    service: typeof DATADOG_SERVICE !== 'undefined' ? DATADOG_SERVICE : undefined,
                    env: environment ? `${environment}` : undefined,
                    proxyHost: process.env.PROXY_HOST
                };
            
                datadogLogs.init(config);
            
                
                    if (process.env.DATADOG_APPLICATION_ID) {
                        datadogRum.init({
                            ...config,
                            trackInteractions: true,
                            applicationId: process.env.DATADOG_APPLICATION_ID
                        });
            
                        datadogRum.setUser({
                            name: service.getName(),
                            email: service.getEmail()
                        });
                    }
        }

  [1]: https://i.stack.imgur.com/8xX8L.png

Is this a know issue?

abereghici:
I encountered the same issue. `Reached max number of errors by minute: 3000`

bcaudan:
Hi,

Thanks for the report, we will look into it shortly.

bcaudan:
We have been able to identify the issue, it happens:

- for npm setup only
- with both logs and RUM installed
- since v2.18.0
- only for `console.error`

[...]

PR #1027, opened 2021-09-02
🐛 [RUMF-1012] fix console.error loop on npm setup

bcaudan:
## Motivation

On npm setup, with both logs and RUM enabled, `console.error` calls generate an unexpected number of logs.
Regression after error collection refactoring in https://github.com/DataDog/browser-sdk/pull/930.

fixes https://github.com/DataDog/browser-sdk/issues/1022

## Changes

Ensure that `console.error` can only be overwrite once.

## Testing

unit test

---

### DIFF
--- a/packages/core/src/domain/error/trackConsoleError.ts
+++ b/packages/core/src/domain/error/trackConsoleError.ts
@@ -12,29 +12,56 @@ import { find, jsonStringify } from '../../tools/utils'
 import { callMonitored } from '../internalMonitoring'
 import { computeStackTrace } from '../tracekit'
 
-let originalConsoleError: (...params: unknown[]) => void
-
 /* eslint-disable no-console */
 export function trackConsoleError(errorObservable: Observable<RawError>) {
-  originalConsoleError = console.error
-
-  console.error = (...params: unknown[]) => {
-    const handlingStack = createHandlingStack()
-    callMonitored(() => {
-      originalConsoleError.apply(console, params)
-      errorObservable.notify({
-        ...buildErrorFromParams(params, handlingStack),
-        source: ErrorSource.CONSOLE,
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/error/trackConsoleError.ts
```diff
@@ -12,29 +12,56 @@ import { find, jsonStringify } from '../../tools/utils'
...
+        }
+        afterCallCallbacks.forEach((callback) => callback(rawError))
       })
-    })
+    }
+    consoleErrorProxySingleton = {
+      afterCall(callback: AfterCallCallback) {
+        afterCallCallbacks.push(callback)
+      },
+    }
```
BenoitZugmeyer, line 55: It could be simpler to use an Observable here

packages/core/src/domain/error/trackConsoleError.ts
```diff
@@ -12,29 +12,56 @@ import { find, jsonStringify } from '../../tools/utils'
...
-      errorObservable.notify({
-        ...buildErrorFromParams(params, handlingStack),
-        source: ErrorSource.CONSOLE,
-        startClocks: clocksNow(),
-        handling: ErrorHandling.HANDLED,
+  startConsoleErrorProxy().afterCall((error) => errorObservable.notify(error))
+
+  return {
+    stop: resetConsoleErrorProxy,
+  }
```
BenoitZugmeyer, line 21: The `stop` here is a bit confusing since it stops console tracking globally, not only where `trackConsoleError` was used. Since it is only used in tests, i think it would be better to use `resetConsoleErrorProxy` directly there.

BenoitZugmeyer (approved):

webNeat (approved):

PR #1032, opened 2021-09-06
Introducing new max for foreground periods

ThibautGeriz:
## Motivation

We have quite a log of `Reached maximum of foreground time` errors but the views do not have 500 items.

## Changes

Let's split the max in two: 
- a max of periods stored to limit memory used
- a max of periods by view to limit bandwidth

## Testing

It's unit tested

### DIFF
--- a/packages/rum-core/src/domain/foregroundContexts.ts
+++ b/packages/rum-core/src/domain/foregroundContexts.ts
@@ -148,11 +148,9 @@ function getInForegroundPeriods(eventStartTime: RelativeTime, duration: Duration
   const eventEndTime = (eventStartTime + duration) as RelativeTime
   const filteredForegroundPeriods: InForegroundPeriod[] = []
 
-  for (let i = foregroundPeriods.length - 1; i >= 0; i--) {
+  const earliestIndex = Math.max(0, foregroundPeriods.length - MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW)
+  for (let i = foregroundPeriods.length - 1; i >= earliestIndex; i--) {
     const foregroundPeriod = foregroundPeriods[i]
-    if (foregroundPeriods.length - i > MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW) {
-      break
-    }
     if (foregroundPeriod.end !== undefined && eventStartTime > foregroundPeriod.end) {
       // event starts after the end of the current focus period
       // since the array is sorted, we can stop looking for foreground periods
--- a/packages/rum-core/src/domain/foregroundContexts.ts
+++ b/packages/rum-core/src/domain/foregroundContexts.ts
@@ -13,16 +13,16 @@ import {
 import { InForegroundPeriod } from '../rawRumEvent.types'
 
 // Arbitrary value to cap number of element mostly for backend & to save bandwidth
-export const MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW = 500
+export const MAX_NUMBER_OF_SELECTABLE_FOREGROUND_PERIODS = 500
 // Arbitrary value to cap number of element mostly for memory consumption in the browser
-export const MAX_NUMBER_OF_FOCUSED_TIME = 2500
+export const MAX_NUMBER_OF_STORED_FOREGROUND_PERIODS = 2500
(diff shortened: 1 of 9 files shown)


packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -12,8 +12,10 @@ import {
 } from '@datadog/browser-core'
 import { InForegroundPeriod } from '../rawRumEvent.types'
 
-// Arbitrary value to cap number of element (mostly for backend)
-export const MAX_NUMBER_OF_FOCUSED_TIME = 500
+// Arbitrary value to cap number of element mostly for backend & to save bandwidth
+export const MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW = 500
+// Arbitrary value to cap number of element mostly for memory consumption in the browser
+export const MAX_NUMBER_OF_FOCUSED_TIME = 2500
```
bcaudan, line 18: Do we need to keep and increase `MAX_NUMBER_OF_FOCUSED_TIME`?

BenoitZugmeyer: The idea is to have  two  limits  serving two different purposes;

* `MAX_NUMBER_OF_FOCUSED_TIME` is here to limit the memory usage. It can be fairly large as objects are small.
* `MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW` is here to limit  the bandwidth usage. We keep the same limit as before (500) as it should be enough.

packages/rum-core/src/domain/foregroundContexts.ts
```diff
@@ -148,6 +150,9 @@ function getInForegroundPeriods(eventStartTime: RelativeTime, duration: Duration
 
   for (let i = foregroundPeriods.length - 1; i >= 0; i--) {
     const foregroundPeriod = foregroundPeriods[i]
+    if (foregroundPeriods.length - i > MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW) {
+      break
+    }
```
BenoitZugmeyer, line 155: There is two stop conditions here. They may be merged together, with something like:

```suggestion
  const earliestIndex  = Math.max(0, foregroundPeriods.length - MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW)
  for (let i = foregroundPeriods.length - 1; i >= earliestIndex; i--) {
    const foregroundPeriod = foregroundPeriods[i]
```

packages/rum-core/src/domain/foregroundContexts.ts
ThibautGeriz: Agreed. Most of the views have less than 500 focused periods therefore the array is full because of SPA changes. We can drastically reduce the number of errors while keeping the same limit for the bandwidth.

bcaudan: IIUC:
- `MAX_NUMBER_OF_FOCUSED_TIME` is the max number of foreground periods stored 
- `MAX_NUMBER_OF_FOCUSED_TIME_PER_VIEW` is the max number of foreground periods than can be returned by `getInForegroundPeriods`.

if that so, we could probably clarify the namings:
- using "view" in the constant is "leaking" information about the current consumer which we don't need to know here
- I am confused by "focused time" vs "foreground period", is there a difference?
- we could maybe be more expressive than `get` for `getInForeground` and `getInForegroundPeriods` and use the same verb for `getInForegroundPeriods` and the related constant.

Given that, I would have in mind something like:

```ts
const MAX_NUMBER_OF_SELECTABLE_FOREGROUND_PERIODS = 500
const MAX_NUMBER_OF_STORED_FOREGROUND_PERIODS = 2500

isInForegroundAt: (startTime: RelativeTime) => boolean | undefined
selectInForegroundPeriodsFor: (startTime: RelativeTime, duration: Duration) => InForegroundPeriod[] | undefined
```

wdyt?

ThibautGeriz: Agreed, the naming you're proposing is better than the current one

BenoitZugmeyer (approved):

bcaudan (approved):
LGTM

PR #1034, opened 2021-09-07
Export serialise dom function

ThibautGeriz:
## Motivation

Export serialise dom function to be used in other datadog repo


BenoitZugmeyer:
You could also `export { NodePrivacyLevel } from './constants'` to pass the right constant when using `serializeNodeWithId`

index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
BenoitZugmeyer, line 1: Oops

BenoitZugmeyer (approved):

PR #1043, opened 2021-09-13
🎨  Update performance project structure

amortemousque:
## Motivation

Reorganise the performance project in order to be add tests more easily

## Changes

Update folder structure 

## Testing

Locally

---


BenoitZugmeyer (commented):
I think it could be nice to have Scenario objects like:

```ts
interface Scenario {
  description: string,
  run(page: Page, takeMeasurements: () => Promise<void>): Promise<void>
}

// and then

export const twitterScenario: Scenario = {
  description: `...`,
  async run(page, takeMeasurements) {
    ...
  }
}
```

so we don't have to explicitly declare all `run` arguments in each scenario, we only import `Scenario` and every types are automatically applied.

BenoitZugmeyer (approved):

PR #1055, opened 2021-09-15
♻️ [RUMF-1015] add callback capabilities to observable

bcaudan:
## Motivation

preliminary refactoring

## Changes

- add `Observable` callback as constructor param:

```ts
const observable = new Observable(() => {
  // on first subscribe logic
  ...
  
  return () => {
    // on last unsubscribe logic
    ...  
  }
})
```

- use it to not duplicate `Observable` logic in  `domMutationObservable`

## Testing

Unit

---

### DIFF
--- a/packages/rum-core/src/browser/domMutationObservable.ts
+++ b/packages/rum-core/src/browser/domMutationObservable.ts
@@ -2,32 +2,20 @@ import { monitor, Observable } from '@datadog/browser-core'
 
 export function createDOMMutationObservable() {
   const MutationObserver = getMutationObserverConstructor()
-  const observable = new Observable<void>(() => {
-    startDOMObservation()
-    return stopDOMObservation
-  })
-  const observer = MutationObserver ? new MutationObserver(monitor(() => observable.notify())) : undefined
 
-  function startDOMObservation() {
-    if (!observer) {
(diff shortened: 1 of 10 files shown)


packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -1,17 +1,12 @@
...
   const MutationObserver = getMutationObserverConstructor()
-  const observer = MutationObserver ? new MutationObserver(monitor(notify)) : undefined
-
-  function notify() {
-    callbacks.forEach((callback) => callback())
-  }
+  const observable = new Observable<void>(() => {
+    startDOMObservation()
+    return stopDOMObservation
+  })
```
BenoitZugmeyer, line 8: I think this could be simpler written as:

```suggestion
  const observable = new Observable<void>(() => {
    if (!MutationObserver) return
    
    const observer = new MutationObserver(monitor(() => observable.notify()))
 
    observer.observe(document, {
      attributes: true,
      characterData: true,
      childList: true,
      subtree: true,
    })

    return () => observer.disconnect()
  })
```

It avoids a maybe undefined observer.

BenoitZugmeyer (approved):

PR #1057, opened 2021-09-16
♻️ [RUMF-1015] extract ContextHistory

bcaudan:
## Motivation

Preliminary refactoring to use same context history logic as parent contexts

## Changes

- Introduce `ContextHistory` class
- Use it for action and view context history in parent contexts

## Testing

Unit

---


packages/rum-core/src/domain/contextHistory.ts
```diff
@@ -0,0 +1,66 @@
...
+
+interface PreviousContext<T> {
+  startTime: RelativeTime
+  endTime: RelativeTime
+  context: T
+}
+
+export const CLEAR_OLD_CONTEXTS_INTERVAL = ONE_MINUTE
+
+export class ContextHistory<Raw extends { startClocks: ClocksState }, Built> {
```
BenoitZugmeyer, line 11: It seems like only `startClocks.relative` is used, maybe change Raw to extend `{ startTime: RelativeTime }` ?

packages/rum-core/src/domain/contextHistory.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+        startClocks: relativeToClocks(20 as RelativeTime),
+        value: 'bar',
+      }
+
+      expect(contextHistory.find(relativeToClocks(15 as RelativeTime))).toBeUndefined()
+    })
+  })
+
+  describe('closeCurrent', () => {
+    it('should not clean current state to allow to use it for next current', () => {
```
BenoitZugmeyer, line 72: I don't really get the use case for it. In this PR we always override `.current` right after using `.closeCurrent()` (except in this test case of course). I think it might be the source of bugs if `closeCurrent` is used multiple times.

packages/rum-core/src/domain/contextHistory.ts
bcaudan: Indeed, internally we are only using relative time, but for parent contexts usage we'd like to use it with objects containing clocks (ViewCreatedEvent, AutoActionCreatedEvent).
Adding a startTime to these objects or creating new ones with it does not seem so great to me, converting everything to clock is a bit painful as well.
We could not force the `Raw` object structure and provide the current context start time outside of the current context object, it would allow us to work only with relative time.

something like:
```ts
const foo = { value, startClocks }
contextHistory.current = foo
contextHistory.currentStartTime = foo.startClocks.relative
```

wdyt?

packages/rum-core/src/domain/contextHistory.spec.ts
bcaudan: It is indeed a need for the following PR.

I could remove this behavior for now and see if it can be better handled in the PR that will need it.

packages/rum-core/src/domain/contextHistory.ts
amortemousque: What is the benefit of a class instead of a function?

amortemousque: What is the benefit of a class instead of a function?

bcaudan: > What is the benefit of a class instead of a function?

when I need to encapsulate some logic for a behavior that I want to use independently at several places, I tend to use classes by default. 
We could achieve a similar result with functions.

amortemousque: Even though it's just code style, I would move to a functional implementation for consistency reason then.

bcaudan: there are several other examples of classes in the code base.
it is a quite common concept, any issue with it?

BenoitZugmeyer: > Indeed, internally we are only using relative time, but for parent contexts usage we'd like to use it with objects containing clocks (ViewCreatedEvent, AutoActionCreatedEvent).
> Adding a startTime to these objects or creating new ones with it does not seem so great to me, converting everything to clock is a bit painful as well.
> We could not force the `Raw` object structure and provide the current context start time outside of the current context object, it would allow us to work only with relative time.
> 
> something like:
> 
> ```ts
> const foo = { value, startClocks }
> contextHistory.current = foo
> contextHistory.currentStartTime = foo.startClocks.relative
> ```
> 
> wdyt?

It looks good, and maybe we could go one step further and remove the `buildContext` function altogether. For the view context case, we could have:

```ts
contextHistory.current = {
      session: {
        id: viewCreatedEvent.sessionId,
      },
      view: {
        id: viewCreatedEvent.id,
        name: viewCreatedEvent.name,
        referrer: viewCreatedEvent.referrer,
        url: viewCreatedEvent.location.href,
      },
    }
context.currentStartTime = viewCreatedEvent.startClocks.relative
```

Maybe an `updateCurrent(current, startTime)` method would be nicer.

packages/rum-core/src/domain/contextHistory.spec.ts
BenoitZugmeyer: Sounds good to me

packages/rum-core/src/domain/contextHistory.ts
bcaudan: > Maybe an updateCurrent(current, startTime) method would be nicer.

👍 I started an implementation with something like that

> maybe we could go one step further and remove the `buildContext` function altogether

ok so the client is responsible to keep the info needed to generate the context.
it should remove the issue with https://github.com/DataDog/browser-sdk/pull/1057#discussion_r710937948 then 😉

bcaudan: In fact, [this test](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/parentContexts.spec.ts#L121) is preventing us to remove the `buildContext`  for now, but we should be able to remove it in the last PR.

BenoitZugmeyer (approved):

PR #1058, opened 2021-09-16
✨ Add rate limit on actions

amortemousque:
## Motivation

In normal conditions the max actions/min is around 200 but we have seen in some cases where we have >200k actions/min which impact the intake.

Add a rate limit mechanism for actions similar to what we have for error.

## Changes

Create an event rate limiter to handle error and action limit

## Testing

Unit

---


packages/core/src/tools/createEventRateLimiter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type EventRateLimiter = ReturnType<typeof createEventRateLimiter>
+
+export function createEventRateLimiter(
+  eventType: string,
+  limit: number,
+  onLimitReached: (limitError: RawError) => void
+) {
+  let actionCount = 0
+  let allowNextAction = false
```
bcaudan, line 11:
```suggestion
  let eventCount = 0
  let allowNextEvent = false
```
no?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -66,9 +67,14 @@ export function startRumAssembly(
...
-  const errorFilter = createErrorFilter(configuration, (error) => {
+  const reportError = (error: RawError) => {
     lifeCycle.notify(LifeCycleEventType.RAW_ERROR_COLLECTED, { error })
-  })
+  }
+
+  const eventRateLimiters = [
+    createEventRateLimiter(RumEventType.ERROR, configuration.maxErrorsPerMinute, reportError),
+    createEventRateLimiter(RumEventType.ACTION, configuration.maxActionsPerMinute, reportError),
+  ]
```
bcaudan, line 77:
```suggestion
  const eventRateLimiters = {
    [RumEventType.ERROR]: createEventRateLimiter(RumEventType.ERROR, configuration.maxErrorsPerMinute, reportError),
    [RumEventType.ACTION]: createEventRateLimiter(RumEventType.ACTION, configuration.maxActionsPerMinute, reportError),
  }
```
it would be more suited to how it is used

packages/rum-core/src/domain/assembly.ts
```diff
@@ -151,9 +157,12 @@ function shouldSend(
...
   }
-  if (event.type === RumEventType.ERROR) {
-    return !errorFilter.isLimitReached()
+
+  const rateLimitReached = eventRateLimiters.find((l) => event.type === l.eventType)?.isLimitReached()
+  if (rateLimitReached) {
+    return false
   }
+
   return true
```
bcaudan, line 166:
```ts
if (boolean) {
  return false
}
return true
```
=> 
```ts
return !boolean
```
🙏

packages/core/src/tools/createEventRateLimiter.ts
amortemousque: yes

packages/rum-core/src/domain/assembly.ts
amortemousque: Don't really like the `RumEventType.ERROR` repetition but it's nicer for `createEventRateLimiter()` and `shouldSend()`, so let's do it ;)

packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -652,6 +652,74 @@ describe('rum assembly', () => {
...
+      notifyRumActionEvent(lifeCycle, 'bar')
+      clock.tick(ONE_MINUTE)
+      notifyRumActionEvent(lifeCycle, 'baz')
+
+      expect(serverRumEvents.length).toBe(2)
+      expect((serverRumEvents[0] as RumActionEvent).action.target?.name).toBe('foo')
+      expect((serverRumEvents[1] as RumActionEvent).action.target?.name).toBe('baz')
+    })
+
+    function notifyRumActionEvent(lifeCycle: LifeCycle, name = 'oh great') {
```
bcaudan, line 715:
```suggestion
    function notifyRumActionEvent(lifeCycle: LifeCycle, name: string) {
```
default value does not seem to be used

bcaudan (approved):

PR #1059, opened 2021-09-16
✅ add CSP headers on e2e tests

BenoitZugmeyer:
## Motivation

Make sure tests are working with restrictive CSP

## Changes

Add CSP headers on e2e tests

## Testing

E2E

---


test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -47,24 +47,28 @@ export function createMockServerApp(endpoints: Endpoints, setup: string) {
   })
 
   app.get('/', (_req, res) => {
+    res.header(
+      'Content-Security-Policy',
+      `connect-src ${intakeUrl}; script-src 'self' 'unsafe-inline'; worker-src blob:;`
```
BenoitZugmeyer, line 52: * There is only one intake host, so no need for a star in connect-src
* We host the SDK bundles on the same domain as the app, so we can use `'self'` for `script-src`

acorretti (approved):

bcaudan (approved):
Great!

PR #1060, opened 2021-09-16
♻️ [RUMF-1015] allow to observe location changes

bcaudan:
## Motivation

preliminary refactoring to observe location changes from other components

## Changes

- convert `trackLocationChanges` to `locationChangesObservable`
- move `areDifferentLocation` logic to `trackViews`
- introduce `changeLocation` helper to simulate location change during unit tests 

## Testing

unit

---


amortemousque (approved):

packages/rum-core/src/browser/locationChangeObservable.ts
```diff
@@ -1,23 +1,34 @@
...
+    const { stop: stopHistoryTracking } = trackHistory(onLocationChange)
+    const { stop: stopHashTracking } = trackHash(onLocationChange)
+    return () => {
       stopHistoryTracking()
       stopHashTracking()
-    },
+    }
+  })
+
+  function onLocationChange() {
```
amortemousque, line 19: Nit: What do you think of moving this function inside the observable to slowly move toward an implementation where `notify()` would not be exposed?

bcaudan: FMU, it does not change anything for now. If we choose to go this way at some point, it would only be one easy change more to do, not a big deal IMO.

amortemousque: Fair enough!

BenoitZugmeyer (approved):

PR #1063, opened 2021-09-21
♻️ [RUMF-1015] use the url corresponding to the start of the event

bcaudan:
## Motivation

With the introduction of manual view tracking (#924), the creation of a view can be decoupled from URL changes.
In the case of a URL change occurring just before the end of a view, the view URL could wrongly be updated with the new location. 

**Previous behavior:**

- view events: URL updated during the lifecycle of the view
- other events: URL corresponding to the end of the event

**New behavior:**
- use the URL corresponding to the start of the event

## Changes

- Introduce `UrlContexts` to store URL states history
- Use `UrlContexts` from `assembly` to retrieve URL and referrer
- Simplify `ContextHistory` since we no more need to dynamically generate the current context

## Testing

Unit, manually

---

### DIFF
--- a/packages/rum-core/src/domain/contextHistory.ts
+++ b/packages/rum-core/src/domain/contextHistory.ts
@@ -8,22 +8,22 @@ interface PreviousContext<T> {
 
 export const CLEAR_OLD_CONTEXTS_INTERVAL = ONE_MINUTE
 
-export class ContextHistory<Raw, Built> {
-  private current: Raw | undefined
+export class ContextHistory<Context> {
+  private current: Context | undefined
   private currentStart: RelativeTime | undefined
-  private previousContexts: Array<PreviousContext<Built>> = []
+  private previousContexts: Array<PreviousContext<Context>> = []
   private clearOldContextsInterval: number
 
-  constructor(private buildContext: (r: Raw) => Built, private expireDelay: number) {
+  constructor(private expireDelay: number) {
     this.clearOldContextsInterval = setInterval(() => this.clearOldContexts(), CLEAR_OLD_CONTEXTS_INTERVAL)
   }
 
   find(startTime?: RelativeTime) {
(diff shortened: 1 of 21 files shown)


packages/rum-core/src/domain/contextHistory.ts
```diff
@@ -8,22 +8,22 @@ interface PreviousContext<T> {
...
 
-  constructor(private buildContext: (r: Raw) => Built, private expireDelay: number) {
+  constructor(private expireDelay: number) {
     this.clearOldContextsInterval = setInterval(() => this.clearOldContexts(), CLEAR_OLD_CONTEXTS_INTERVAL)
   }
 
   find(startTime?: RelativeTime) {
     if (startTime === undefined) {
-      return this.current ? this.buildContext(this.current) : undefined
+      return this.current ? this.current : undefined
```
BenoitZugmeyer, line 23: NIT: Could be:
```suggestion
      return this.current
```

packages/rum-core/src/domain/urlContexts.ts
```diff
@@ -0,0 +1,76 @@
...
+      urlContextHistory.closeCurrent(changeTime)
+      urlContextHistory.setCurrent(
+        buildUrlContext({
+          url: newLocation.href,
+          referrer: current.view.referrer,
+        }),
+        changeTime
+      )
+    }
+  })
```
amortemousque, line 58: FMU, if a location change then all these handlers will be called because tackView() also listen to it when views are tracked automatically. Is it the intended behaviour?

bcaudan: Do you see an issue with it?

bcaudan: as seen together, I've added an extra test

amortemousque (approved):

PR #1069, opened 2021-09-22
⚡ [RUMF-1022] remove `__sn` property in nodes

BenoitZugmeyer:
## Motivation

Consume less memory. When running the performance script:
* on Twitter scenario, we go from 1.67 MB to 890 kB median memory consumption
* on Wikipedia we go from 2.48 MB to  515 kB

(No noticeable change on CPU usage)

## Changes

* Use a weakmap instead of an internal property in nodes
* Store only the node ids instead of the whole serialized node structure.

## Testing

CI

---


amortemousque (approved):

packages/rum/src/domain/record/serializationUtils.ts
```diff
@@ -1,14 +1,13 @@
...
 import { CENSORED_STRING_MARK, NodePrivacyLevel } from '../../constants'
 import { shouldMaskNode } from './privacy'
-import { SerializedNodeWithId } from './types'
 
-export interface NodeWithSerializedNode extends Node {
-  __sn: SerializedNodeWithId
-}
+export type NodeWithSerializedNode = Node & { s: 'Node with serialized node' }
+
+const serializedNodeIds = new WeakMap<Node, number>()
```
bcaudan, line 7: what is the reasoning of using a `WeakMap` rather than continue storing the id directly on the node?

BenoitZugmeyer: We avoid setting a visible property in our customers app. Also it solves issues when the SDK is instanciated twice. And in my mind it's simply more elegent.

Do you see something wrong with this?

bcaudan: no, I was just wondering :sli

bcaudan (approved):

packages/rum/src/domain/record/serializationUtils.ts
amortemousque: Yes, I also think it's a great use of it :)

ThibautGeriz (approved):
Not sure my approval is worth much but it looks good

PR #1075, opened 2021-09-27
🔊 Add monitoring on session type change

bcaudan:
## Motivation

cf #1066, #1072

## Changes

Add log on unexpected session type change

## Testing

locally

---

### DIFF
--- a/packages/core/src/domain/sessionManagement.ts
+++ b/packages/core/src/domain/sessionManagement.ts
@@ -33,42 +33,37 @@ export function startSessionManagement<TrackingType extends string>(
   const sessionCookie = cacheCookieAccess(SESSION_COOKIE_NAME, options)
   tryOldCookiesMigration(sessionCookie)
   const renewObservable = new Observable<void>()
-  const sessionState = retrieveActiveSession(sessionCookie)
-  let initialSessionState = { ...sessionState }
+  let inMemorySession = retrieveActiveSession(sessionCookie)
 
   const { throttled: expandOrRenewSession } = utils.throttle(
     monitor(() => {
-      const session = retrieveActiveSession(sessionCookie)
-      const retrievedSession = { ...session }
-      const { trackingType, isTracked } = computeSessionState(session[productKey])
-      session[productKey] = trackingType
-      if (isTracked && !session.id) {
-        session.id = utils.generateUUID()
-        session.created = String(Date.now())
+      const cookieSession = retrieveActiveSession(sessionCookie)
+      const retrievedSession = { ...cookieSession }
+      const { trackingType, isTracked } = computeSessionState(cookieSession[productKey])
+      cookieSession[productKey] = trackingType
+      if (isTracked && !cookieSession.id) {
+        cookieSession.id = utils.generateUUID()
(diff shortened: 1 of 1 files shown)


amortemousque (approved):

packages/core/src/domain/sessionManagement.ts
```diff
@@ -48,10 +50,26 @@ export function startSessionManagement<TrackingType extends string>(
       persistSession(session, sessionCookie)
 
       // If the session id has changed, notify that the session has been renewed
-      if (isTracked && currentSessionId !== session.id) {
-        currentSessionId = session.id
+      if (isTracked && initialSessionState.id !== session.id) {
+        initialSessionState = { ...session }
```
BenoitZugmeyer, line 54: Since the `initialSessionState` is reaffected, maybe it should be called something else? I think `currentSessionState` (or `currentSession`, since the `State` suffix is not used for `session` and `retrievedSession` variables) is more accurate.

packages/core/src/domain/sessionManagement.ts
```diff
@@ -33,11 +33,13 @@ export function startSessionManagement<TrackingType extends string>(
   const sessionCookie = cacheCookieAccess(SESSION_COOKIE_NAME, options)
   tryOldCookiesMigration(sessionCookie)
   const renewObservable = new Observable<void>()
-  let currentSessionId = retrieveActiveSession(sessionCookie).id
+  const sessionState = retrieveActiveSession(sessionCookie)
+  let initialSessionState = { ...sessionState }
```
BenoitZugmeyer, line 37: `sessionState` is not used anywhere else, and `initialSessionState` is not mutated, you could use a single variable for this.

packages/core/src/domain/sessionManagement.ts
bcaudan: hum, IMO `current` is a bit confusing since the `retrieved` state could be qualified of `current`
maybe `stored` or `inMemory` could be more accurate
wdyt?

bcaudan: submitted with `inMemorySession` and `cookieSession`

BenoitZugmeyer (approved):

packages/core/src/domain/sessionManagement.ts
```diff
@@ -33,25 +33,38 @@ export function startSessionManagement<TrackingType extends string>(
...
+      const retrievedSession = { ...cookieSession }
+      const { trackingType, isTracked } = computeSessionState(cookieSession[productKey])
+      cookieSession[productKey] = trackingType
+      if (isTracked && !cookieSession.id) {
+        cookieSession.id = utils.generateUUID()
+        cookieSession.created = String(Date.now())
       }
       // save changes and expand session duration
-      persistSession(session, sessionCookie)
+      persistSession(cookieSession, sessionCookie)
```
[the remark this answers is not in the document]
bcaudan: I had a thought on this too. 
IMO it is good enough for now, let's see if we learn something from the monitoring and we could reassess naming during cleanup or rework.

PR #1082, opened 2021-09-28
[RUMF-1034] allow passing `undefined` options to RUM and Logs init

BenoitZugmeyer:
## Motivation

TS 4.4 introduced a new option called `exactOptionalPropertyTypes`, that disallows passing `undefined` to optional properties. Example:

```ts
interface Options {
  foo?: number
}

let options: Options = {
  foo: undefined // this will raise an error with `exactOptionalPropertyTypes: true`
}
```

To ease SDK integration in projects that are using this TS option, we want to explicitly allow passing `undefined` as a value for optional properties. 

## Changes

* add a small compatibility test to check for this configuration
* explicitly allow `undefined` for all `init` configuration optional properties

## Testing

CI

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -49,39 +49,39 @@ export const DEFAULT_CONFIGURATION = {
 
 export interface InitConfiguration {
   clientToken: string
-  applicationId?: string
-  actionNameAttribute?: string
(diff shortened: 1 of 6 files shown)


amortemousque (approved):

bcaudan (commented):
Would it worth doing it for other public APIs?
like `setUser({id: undefined})`

bcaudan (approved):

PR #1085, opened 2021-09-29
Base tag support: discussion of SDK side approach

jagracey:
## Motivation

Support `<Base href={url}/>` URL resolution for relative URLs. Currently the SDK resolves relative URLs against location.href without consideration for the base tag.

We can either handle support on the record or the replay side.
- This PR discusses the record side solution.
- [This other PR discusses the replay side solution.](https://github.com/DataDog/browser-sdk/pull/1086)

## Testing

---

### DIFF
--- a/packages/rum/src/domain/record/serializationUtils.ts
+++ b/packages/rum/src/domain/record/serializationUtils.ts
@@ -56,12 +56,27 @@ export function makeSrcsetUrlsAbsolute(attributeValue: string, baseUrl: string) 
   )
 }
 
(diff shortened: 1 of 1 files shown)


acorretti:
Am I mistaken or this isn't the approach we discussed?
We talked about recording all URLs _as they are_ and let the browser resolve them to the right location by adjusting a `<base>` tag on the sandbox iframe.

This PR is just turning all URLs absolute, yes according to the `<base>` tag, but still converting them.

We also talked about changing the approach behind a FF.

PR #1086, opened 2021-09-29
Base tag support: discussion of replay side approach

jagracey:
## Motivation

Support `<Base href={url}/>` URL resolution for relative URLs. Currently the SDK resolves relative URLs against location.href without consideration for the base tag.

We can either handle support on the record or the replay side.
- This PR discusses the replay side solution.
- [This other PR discusses the record side solution.](https://github.com/DataDog/browser-sdk/pull/1085)

Support `<Base href={url}/>` URL resolution for relative URLs. Currently the SDK resolves relative URLs against location.href without consideration for the base tag.

**Breakdown:**
Commit # 1 shows a feature flag friendly rollout for internal testing- however the FF implementation via prop drill-down, as is, may be more error prone, so we should figure out a cleaner FF methodology going forward.

Commit # 2/3  shows the final solution without using feature flags. 

## Testing

---

### DIFF
--- a/packages/rum/src/domain/record/privacy.ts
+++ b/packages/rum/src/domain/record/privacy.ts
@@ -20,8 +20,6 @@ import {
 
 export const MAX_ATTRIBUTE_VALUE_CHAR_LENGTH = 100_000
 
-import { makeStylesheetUrlsAbsolute } from './serializationUtils'
-
 import { shouldIgnoreElement } from './serialize'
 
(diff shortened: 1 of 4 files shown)


jagracey:
Commit # 1 shows a feature flag friendly rollout for internal testing- however the FF implementation via prop drill-down, as is, may be more error prone, so we should figure out a cleaner FF methodology going forward.

PR #1088, opened 2021-09-29
♻️  CI scripts refactoring

amortemousque:
## Motivation

During the new staging ci process development utils have been introduced. 
This PR updates all the scripts which can have a use for utils.

## Changes

- Update scripts

## Testing

Locally, Gitlab

---

### DIFF
--- a/scripts/bs-wrapper.js
+++ b/scripts/bs-wrapper.js
@@ -14,28 +14,22 @@
 // after killing it. There might be a better way of prematurely aborting the test command if we need
 // to in the future.
 
-const { spawn } = require('child_process')
 const request = require('request')
+const { spawnCommand, printLog } = require('./utils')
 
 const AVAILABILITY_CHECK_DELAY = 30_000
 // eslint-disable-next-line max-len
 const RUNNING_BUILDS_API = `https://${process.env.BS_USERNAME}:${process.env.BS_ACCESS_KEY}@api.browserstack.com/automate/builds.json?status=running`
 
-main()
-  .then((exitCode) => process.exit(exitCode))
-  .catch((error) => {
-    console.log(error)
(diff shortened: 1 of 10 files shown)


scripts/bs-wrapper.js
```diff
@@ -52,13 +46,15 @@ function hasRunningBuild() {
...
 }
 
 function timeout(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms))
 }
+
+main().catch((error) => {
+  printError('\nStacktrace:\n', error)
+  process.exit(1)
+})
```
bcaudan, line 60: We are doing that a lot, what about extracting it also?
maybe something like 
```ts
printFormattedError(error)
process.exit(1)
```
or
```ts
catch(logAndExit)
```

amortemousque: Yes, sure. I'll do the second proposition

scripts/replace-build-env.js
```diff
@@ -9,21 +10,20 @@ const buildEnv = require('./build-env')
...
-  console.log(
+  printLog(
     'Changed files:',
     results.filter((entry) => entry.hasChanged).map((entry) => entry.file)
   )
   process.exit(0)
 } catch (error) {
-  console.error('Error occurred:', error)
+  printError('\nStacktrace:\n', error)
   process.exit(1)
```
bcaudan, line 28:
```suggestion
  logAndExit(error)
```

bcaudan (approved):

scripts/replace-build-env.js
amortemousque: I've changed the design of `replace-build-env.js` to be like the other scripts

BenoitZugmeyer (approved):

PR #1089, opened 2021-09-29
⚗[REPLAY-341] Add VisualViewport tracking (Pinch Zoom)

jagracey:
## Motivation

Support pinch zoom and pinch scroll by tracking the "visual viewport".

As mentioned in the internal RFC ["Supporting Pinch Zoom Tracking"](https://docs.google.com/document/d/1GOUNsiQizPzANRUvLQW0R-WHuGsOe9p4b-oO3XALU2w/edit?usp=sharing), unfortunately, the dimensions we currently record are not consistent across devices and browsers as some are referenced. On certain devices (primarily mobile), when users pinch zoom, the replay will replay incorrectly- displaying the layout viewport but not the visual viewport.

Unfortunately, browsers have not standardised various dimension properties. Mobile devices typically report dimensions in reference to the visual viewport, while desktop uses the layout viewport. For example, mobile Chrome will change innerWidth when a pinch zoom takes place, while Chrome desktop (mac) will not

With the new Viewport API, we now calculate and normalise dimension properties to the layout viewport.

**Note:** If the VisualViewport API is not supported by a browser, it isn't reasonably possible to detect or normalise which viewport is being measured. Therefore these exported functions will fallback to assuming that the layout viewport is being measured by the browser

## Changes
1. Adds VisualViewport record as incremental record.
2. Changes innerWidth/innerHeight, scrollX/scrollY, and mouse positions to a normalised value that is referenced to the layout viewport (not pinch zoom's visual viewport) so that we do not factor in the visual viewport (pinch zoom) twice for mobile devices.

## Testing
Adds new unit tests comparing initial native properties to synthetic alternative.
Note that to support pinch zoom and scroll testing, we will need to utilise our E2E testing framework  to use Appium's gesture commands, or utilise Chrome Devtool Protocol to send emulated user interactions.

---


packages/rum/src/domain/record/observer.ts
```diff
@@ -46,6 +48,11 @@ export function initObservers(o: ObserverParam): ListenerHandler {
   const styleSheetObserver = initStyleSheetObserver(o.styleSheetRuleCb)
   const focusHandler = initFocusObserver(o.focusCb)
 
+  const visualViewportResizeHandler = isExperimentalFeatureEnabled('visualviewport')
+    ? initVisualViewportResizeObserver(o.visualViewportResizeCb)
+    : // eslint-disable-next-line @typescript-eslint/no-empty-function
+      () => {}
```
ThibautGeriz, line 54: **suggestion:** use noop
```ts
import { noop } from '@datadog/browser-core'
```

packages/rum/src/domain/record/record.ts
```diff
@@ -57,6 +58,24 @@ export function record(options: RecordOptions): RecordAPI {
     })
   }
 
+  if (isExperimentalFeatureEnabled('visualviewport')) {
+    if (visualViewport) {
```
ThibautGeriz, line 62: **suggestion:**  group both conditions

packages/rum/src/domain/record/record.ts
```diff
@@ -57,6 +58,24 @@ export function record(options: RecordOptions): RecordAPI {
...
+      const viewport = visualViewport
```
ThibautGeriz, line 63: **question:** Why renaming here?
**sugestion:** remove rename or in the first name the variable `visualViewport` as `viewport`

packages/rum/src/domain/record/record.ts
```diff
@@ -57,6 +58,24 @@ export function record(options: RecordOptions): RecordAPI {
...
+        data: {
+          scale: viewport.scale,
+          offsetLeft: viewport.offsetLeft,
+          offsetTop: viewport.offsetTop,
+          pageLeft: viewport.pageLeft,
+          pageTop: viewport.pageTop,
+          height: viewport.height,
+          width: viewport.width,
+        },
+        type: RecordType.VisualViewport,
```
ThibautGeriz, line 74: **question:** is there a web-ui PR sister PR to handle this new type of record?

If not, it should not break the player but this records won't have any handler and therefore will trigger an error in error tracking.
That said, to have the new types in web-ui, you'll need to publish this on NPM

packages/rum/src/domain/record/viewports.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { isSafari } from 'packages/core/test/specHelper'
```
ThibautGeriz, line 1: I am not sure to understand everything in `viewport.ts` but there is a lot of complexity and only two unit tests. Are you sure to have covered every case?

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,79 @@
...
+      innerHeight: window.innerHeight,
+      scrollX: window.scrollX,
+      scrollY: window.scrollY,
+    }
+    // scrollX/Y is subpixel precise, but has less support so needs a fallback
+    if (window.scrollX === undefined || window.scrollY === undefined) {
+      const scrollXFallback =
+        window.pageXOffset !== undefined
+          ? window.pageXOffset
+          : (document.documentElement || document.body.parentNode || document.body).scrollLeft
```
ThibautGeriz, line 32: **suggestion:** Can you extract variables? It's not super easy to read?
(same for the second one)

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,79 @@
...
+    normalised.visualViewportY = null
+  } else if (isVisualViewportFactoredIn()) {
+    normalised.layoutViewportX + visualViewport.offsetLeft
+    normalised.layoutViewportY + visualViewport.offsetTop
+  } else {
+    normalised.visualViewportX! - visualViewport.offsetLeft
+    normalised.visualViewportY! - visualViewport.offsetTop
+  }
+  return normalised
+}
```
ThibautGeriz, line 79: **suggestion:** you often create an object in that file and then you mutate its props.
For me, it's anti pattern, I would try to keep it immutable by computing all its value first. Honesty your code is readable so it might make it less readable, in that case please ignore.

packages/rum/src/domain/record/observer.ts
jagracey: amazing. perfect

packages/rum/src/domain/record/record.ts
jagracey: Web-UI has a corresponding listener here, though not ready for review:
```
 recordsEmitter.addEventListener(
      RecordType.VisualViewportRecord,
      visualViewportCallback
 );
```

packages/rum/src/domain/record/viewports.spec.ts
jagracey: It's true we don't have tests covering the _new_ pinch zoom + pinch scroll functionality. Unfortunately our unit tests don't expose hooks to the automation library, and our E2E tests need some large changes to synthetically trigger pinch zoom and page zoom while still being able to unit test inside the codebase.

So for this PR, under the safety of a feature flag we are testing to make sure we don't break existing functionality- specifically innerWidth and scroll.

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,79 @@
...
+    // Unable to normalise
+    normalised.visualViewportX = null
+    normalised.visualViewportY = null
+  } else if (isVisualViewportFactoredIn()) {
+    normalised.layoutViewportX + visualViewport.offsetLeft
+    normalised.layoutViewportY + visualViewport.offsetTop
+  } else {
+    normalised.visualViewportX! - visualViewport.offsetLeft
+    normalised.visualViewportY! - visualViewport.offsetTop
+  }
```
BenoitZugmeyer, line 77: Those statements do nothing

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,79 @@
+/**
```
BenoitZugmeyer, line 1: This whole file is not imported/used anywhere. Is something missing?

packages/rum/src/domain/record/record.ts
jagracey: TBH, I'm less certain here.

I believe our code minifier (Terser) won't minify global variables. Instead if we use an intermediate variable (line 63) I believe we can save 112 bytes (and move line 63 to line 60) without reducing readability- though it isn't conventional.

packages/rum/src/domain/record/viewports.ts
jagracey: I'll refactor this. I see a bug now.

packages/rum/src/domain/record/record.ts
jagracey: @BenoitZugmeyer agreed

packages/rum/src/domain/record/viewports.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(layoutDimensions).toEqual({
+      scrollX: 0,
+      scrollY: 0,
+      innerWidth: window.innerWidth,
+      innerHeight: window.innerHeight,
+    })
+
+    if (!isSafari) {
+      // Mobile Safari 12.0 (iOS 12.1) not responding to native scroll
```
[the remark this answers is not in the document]
jagracey: Exactly. Was very strange. I actually went to bizarre measures and wrapped in requestAnimationFrame callbacks (against spec) incase it was an async rendering bug, but this did not resolve this issue (see BrowserStack logs).

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,79 @@
...
+export const getLayoutViewportDimensions = () => {
+  if (!window.visualViewport || !isVisualViewportFactoredIn()) {
+    const dimensions = {
+      innerWidth: window.innerWidth,
+      innerHeight: window.innerHeight,
+      scrollX: window.scrollX,
+      scrollY: window.scrollY,
+    }
+    // scrollX/Y is subpixel precise, but has less support so needs a fallback
+    if (window.scrollX === undefined || window.scrollY === undefined) {
```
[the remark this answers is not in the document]
jagracey: I too was surprised. We received a scrollY is undefined bug here with Browserstack during an earlier commit. I can check again later though.

packages/rum/src/domain/record/record.ts
```diff
@@ -57,6 +58,24 @@ export function record(options: RecordOptions): RecordAPI {
...
+      emit({
+        data: {
+          scale: viewport.scale,
+          offsetLeft: viewport.offsetLeft,
+          offsetTop: viewport.offsetTop,
+          pageLeft: viewport.pageLeft,
+          pageTop: viewport.pageTop,
+          height: viewport.height,
+          width: viewport.width,
+        },
```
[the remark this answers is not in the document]
jagracey: Great point.

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Mobile Chrome will change innerWidth when a pinch zoom takes place, while Chrome Desktop (mac) will not.
+ *
+ * With the new Viewport API, we now calculate and normalize dimension properties to the layout viewport.
+ * If the VisualViewport API is not supported by a browser, it isn't reasonably possible to detect or normalize
+ * which viewport is being measured. Therefore these exported functions will fallback to assuming that the layout
+ * viewport is being measured by the browser
+ */
+
+// Scrollbar widths vary across properties on different devices and browsers
+const TOLERANCE = 10
```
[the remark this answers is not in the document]
jagracey: Just bumped to 25px

packages/rum/src/domain/record/record.ts
jagracey: Refactored to use this inside `viewports.ts`, saving perhaps 280 bytes.

acorretti (approved):

packages/rum/src/domain/record/observer.ts
```diff
@@ -141,12 +171,20 @@ function initScrollObserver(cb: ScrollCallback, defaultPrivacyLevel: DefaultPriv
...
+            x: getScrollX(),
+            y: getScrollY(),
+          })
+        } else {
+          const scrollEl = (document.scrollingElement || document.documentElement)!
+          cb({
+            id,
+            x: scrollEl.scrollLeft,
+            y: scrollEl.scrollTop,
+          })
```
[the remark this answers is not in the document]
jagracey: Thanks for noticing above scrollX/Y() was not feature gated in this instance. Just added a FF to these methods.

PR #1090, opened 2021-09-29
♻️ Feature flag refactor: no more "prop drilldown", now just import

jagracey:
## Motivation

Related to (internal) [RUMF-1004](https://datadoghq.atlassian.net/browse/RUMF-1004).

Currently prop drill-down is required to pass the configuration object around to access `configuration.isEnabled()`. This is a lot of work, and requires care to not introduce further bugs. Instead we can leverage a singleton to import `isEnabledExperimentalFeatures`.

## Changes

Moving from `configuration.isEnabled` prop drill-down  to instead `import {isEnabledExperimentalFeatures} from '@datadog/browser-core'`;

## Testing

---

### DIFF
--- a/eslint-local-rules/disallowSideEffects.js
+++ b/eslint-local-rules/disallowSideEffects.js
@@ -35,11 +35,7 @@ const pathsWithSideEffect = new Set([
 ])
 
 // Those packages either are known to have no side effects when evaluated, or are allow listed safely
-const packagesWithoutSideEffect = new Set([
-  '@datadog/browser-core',
-  '@datadog/browser-rum-core',
-  'packages/core/src/domain/configuration/experimentalFeatures',
-])
+const packagesWithoutSideEffect = new Set(['@datadog/browser-core', '@datadog/browser-rum-core'])
 
 /**
  * Iterate over the given node and its children, and report any node that may have a side effect
--- a/eslint-local-rules/disallowSideEffects.js
+++ b/eslint-local-rules/disallowSideEffects.js
@@ -38,7 +38,7 @@ const pathsWithSideEffect = new Set([
 const packagesWithoutSideEffect = new Set([
   '@datadog/browser-core',
   '@datadog/browser-rum-core',
-  'packages/core/src/domain/configuration/experimentalFeatures'
+  'packages/core/src/domain/configuration/experimentalFeatures',
 ])
 
 /**
--- a/eslint-local-rules/disallowSideEffects.js
+++ b/eslint-local-rules/disallowSideEffects.js
@@ -34,8 +34,12 @@ const pathsWithSideEffect = new Set([
   `${packagesRoot}/rum-slim/src/index.ts`,
 ])
 
-// Those packages are known to have no side effects when evaluated
-const packagesWithoutSideEffect = new Set(['@datadog/browser-core', '@datadog/browser-rum-core'])
+// Those packages either are known to have no side effects when evaluated, or are allow listed safely
+const packagesWithoutSideEffect = new Set([
+  '@datadog/browser-core',
+  '@datadog/browser-rum-core',
+  'packages/core/src/domain/configuration/experimentalFeatures'
+])
 
 /**
  * Iterate over the given node and its children, and report any node that may have a side effect
--- a/eslint-local-rules/disallowSideEffects.js
+++ b/eslint-local-rules/disallowSideEffects.js
@@ -34,7 +34,7 @@ const pathsWithSideEffect = new Set([
   `${packagesRoot}/rum-slim/src/index.ts`,
 ])
 
-// Those packages either are known to have no side effects when evaluated
+// Those packages are known to have no side effects when evaluated
 const packagesWithoutSideEffect = new Set(['@datadog/browser-core', '@datadog/browser-rum-core'])
 
 /**
--- a/eslint-local-rules/disallowSideEffects.js
+++ b/eslint-local-rules/disallowSideEffects.js
@@ -34,7 +34,7 @@ const pathsWithSideEffect = new Set([
   `${packagesRoot}/rum-slim/src/index.ts`,
 ])
 
-// Those packages either are known to have no side effects when evaluated, or are allow listed safely
+// Those packages either are known to have no side effects when evaluated
 const packagesWithoutSideEffect = new Set(['@datadog/browser-core', '@datadog/browser-rum-core'])
 
 /**
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -49,7 +49,7 @@ export interface BuildEnv {
 }
 
 export function commonInit(initConfiguration: InitConfiguration, buildEnv: BuildEnv) {
-  updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures || [])
+  updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures)
   const configuration = buildConfiguration(initConfiguration, buildEnv)
   const internalMonitoring = startInternalMonitoring(configuration)
 
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -3,7 +3,7 @@ import { buildConfiguration, InitConfiguration } from '../domain/configuration'
 import { setDebugMode, startInternalMonitoring } from '../domain/internalMonitoring'
 import { catchUserErrors } from '../tools/catchUserErrors'
 import { display } from '../tools/display'
-import { updateEnabledExperimentalFeatures } from '../domain/configuration/experimentalFeatures'
+import { updateExperimentalFeatures } from '../domain/configuration/experimentalFeatures'
 
 export function makePublicApi<T>(stub: T): T & { onReady(callback: () => void): void } {
   const publicApi = {
@@ -49,11 +49,7 @@ export interface BuildEnv {
 }
 
 export function commonInit(initConfiguration: InitConfiguration, buildEnv: BuildEnv) {
-  const enableExperimentalFeatures = Array.isArray(initConfiguration.enableExperimentalFeatures)
-    ? initConfiguration.enableExperimentalFeatures
-    : []
-  updateEnabledExperimentalFeatures(enableExperimentalFeatures)
-
+  updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures || [])
   const configuration = buildConfiguration(initConfiguration, buildEnv)
   const internalMonitoring = startInternalMonitoring(configuration)
 
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -3,6 +3,7 @@ import { buildConfiguration, InitConfiguration } from '../domain/configuration'
 import { setDebugMode, startInternalMonitoring } from '../domain/internalMonitoring'
 import { catchUserErrors } from '../tools/catchUserErrors'
 import { display } from '../tools/display'
+import { setEnabledExperimentalFeatures } from '../domain/configuration/experimentalFeatures'
 
 export function makePublicApi<T>(stub: T): T & { onReady(callback: () => void): void } {
   const publicApi = {
@@ -50,6 +51,12 @@ export interface BuildEnv {
 export function commonInit(initConfiguration: InitConfiguration, buildEnv: BuildEnv) {
   const configuration = buildConfiguration(initConfiguration, buildEnv)
   const internalMonitoring = startInternalMonitoring(configuration)
+
+  const enableExperimentalFeatures = Array.isArray(initConfiguration.enableExperimentalFeatures)
+  ? initConfiguration.enableExperimentalFeatures
+  : []
+
+  setEnabledExperimentalFeatures(enableExperimentalFeatures)
 
   return {
     configuration,
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -49,14 +49,13 @@ export interface BuildEnv {
 }
 
 export function commonInit(initConfiguration: InitConfiguration, buildEnv: BuildEnv) {
-  const configuration = buildConfiguration(initConfiguration, buildEnv)
-  const internalMonitoring = startInternalMonitoring(configuration)
-
   const enableExperimentalFeatures = Array.isArray(initConfiguration.enableExperimentalFeatures)
     ? initConfiguration.enableExperimentalFeatures
     : []
-
   setEnabledExperimentalFeatures(enableExperimentalFeatures)
+
+  const configuration = buildConfiguration(initConfiguration, buildEnv)
+  const internalMonitoring = startInternalMonitoring(configuration)
 
   return {
     configuration,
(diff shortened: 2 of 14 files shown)


acorretti (approved):

eslint-local-rules/disallowSideEffects.js
```diff
@@ -34,7 +34,7 @@ const pathsWithSideEffect = new Set([
   `${packagesRoot}/rum-slim/src/index.ts`,
 ])
 
-// Those packages are known to have no side effects when evaluated
+// Those packages either are known to have no side effects when evaluated
```
BenoitZugmeyer, line 37: Could you revert this change?

packages/core/src/boot/init.ts
```diff
@@ -51,6 +52,12 @@ export function commonInit(initConfiguration: InitConfiguration, buildEnv: Build
   const configuration = buildConfiguration(initConfiguration, buildEnv)
   const internalMonitoring = startInternalMonitoring(configuration)
 
+  const enableExperimentalFeatures = Array.isArray(initConfiguration.enableExperimentalFeatures)
+    ? initConfiguration.enableExperimentalFeatures
+    : []
+
+  setEnabledExperimentalFeatures(enableExperimentalFeatures)
+
```
BenoitZugmeyer, line 60: I think it would be better to move this at the beginning of `commonInit`, so we can use feature flags while building the configuration.

packages/core/src/boot/init.ts
```diff
@@ -48,6 +49,11 @@ export interface BuildEnv {
 }
 
 export function commonInit(initConfiguration: InitConfiguration, buildEnv: BuildEnv) {
+  const enableExperimentalFeatures = Array.isArray(initConfiguration.enableExperimentalFeatures)
+    ? initConfiguration.enableExperimentalFeatures
+    : []
+  setEnabledExperimentalFeatures(enableExperimentalFeatures)
+
   const configuration = buildConfiguration(initConfiguration, buildEnv)
   const internalMonitoring = startInternalMonitoring(configuration)
```
bcaudan, line 58: it could be nice to extract experimental features init to its own function to keep the same level of abstraction inside commonInit

bcaudan (commented):
note on PR form:
- it would deserve a better title to clearly understand what this refactor is about
- it could be interesting to clearly explicit the design change in the changes section (`configuration.isEnabled` -> `isEnabledExperimentalFeatures`), it would be easier for reviewer to have a general idea of the change
- it seems that there was an issue when you have updated your branch with main since some changes already in main appear in your diff

packages/core/src/domain/configuration/experimentalFeatures.ts
```diff
@@ -0,0 +1,9 @@
+let enabledExperimentalFeatures: Set<string>
+
+export function setEnabledExperimentalFeatures(enabledFeatures: string[]): void {
+  enabledExperimentalFeatures = new Set(enabledFeatures)
+}
+
+export function isEnabledExperimentalFeature(featureName: string): boolean {
+  return enabledExperimentalFeatures.has(featureName)
+}
```
bcaudan, line 9: any thoughts on usage during tests?

amortemousque: I have some thoughts about the naming. What do you think about:

- enableExperimentalFeatures(enabledFeatures)
- isExperimentalFeatureEnabled(featureName)

jagracey: This convention of {specifier + group_name} is less correct english, but it isn't uncommon on codebases.
I don't mind one way or the other. Do we have a convention of the reverse?

jagracey: Yes, a reset function makes sense. Just added `resetExperimentalFeatures`.

packages/core/src/boot/init.ts
```diff
@@ -48,6 +49,7 @@ export interface BuildEnv {
 }
 
 export function commonInit(initConfiguration: InitConfiguration, buildEnv: BuildEnv) {
+  updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures || [])
```
bcaudan, line 52: do we need to default to `[]` here?

packages/core/src/domain/configuration/experimentalFeatures.ts
```diff
@@ -0,0 +1,26 @@
+let enabledExperimentalFeatures: Set<string>
+
+export function updateExperimentalFeatures(enabledFeatures: string[]): void {
```
bcaudan, line 3: since you have added extra code to handle some cases, what about adding some tests to document them and  avoid regression?

packages/core/src/boot/init.ts
jagracey: FMU,
Since the FF data is arriving externally from a JS snippet, we should have minimum validation to ensure we have an array since TS won't help runtime issues. To handle the case of being possibly undefined, we can either have this inside the TS file, or here. To me it makes sense to trust TS checking inside the func, so we should default to array outside the func.

bcaudan: I though you were handling the case where no flags were provided here.

> To me it makes sense to trust TS checking inside the func

you are already doing some sanitization inside the function `Array.isArray` and `.filter((flag) => typeof flag === 'string')` which make me think you don't really trust the `enabledFeatures: string[]` typing.

what about doing all the sanitization at the same place?

packages/core/src/domain/configuration/experimentalFeatures.ts
jagracey: just added some tests.

packages/core/src/boot/init.ts
jagracey: 1. enableExperimentalFeatures is typed within initConfig as 
`enableExperimentalFeatures?: string[] | undefined`
so we handle the undefined case outside the `updateExperimentalFeatures` with `initConfiguration.enableExperimentalFeatures || []`
2. Inside updateExperimentalFeatures, as an API, we IMO, shouldn't allow the undefined case in the context of typescript checking because we should be passing string[] only.
3. We can't depend on the runtime following the TS rules given customer control, so we handle them now inside `updateExperimentalFeatures` (changed from inside `commonInit` as a previous commit in this PR ).

Do you have an alternative in mind?

bcaudan: > 1. enableExperimentalFeatures is typed within initConfig as
enableExperimentalFeatures?: string[] | undefined
so we handle the undefined case outside the updateExperimentalFeatures with initConfiguration.enableExperimentalFeatures || []
> 2. Inside updateExperimentalFeatures, as an API, we IMO, shouldn't allow the undefined case in the context of typescript checking because we should be passing string[] only.

why though?
do we have other use cases for `updateExperimentalFeatures` where we should not allow to pass `undefined`?

IMO, we could go with `updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures)` or even `updateExperimentalFeatures(initConfiguration)` since we only use it from the `initConfiguration`.

it would allow us to handle the undefined case in the unit tests as well.

packages/core/src/domain/configuration/experimentalFeatures.ts
```diff
@@ -0,0 +1,26 @@
...
+  // Safely handle external data
+  if (!Array.isArray(enabledFeatures)) {
+    return
+  }
+
+  if (!enabledExperimentalFeatures) {
+    enabledExperimentalFeatures = new Set(enabledFeatures)
+  }
```
bcaudan, line 11: no unit test for `!enabledExperimentalFeatures`, what was your intent?

packages/core/src/domain/configuration/experimentalFeatures.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  enabledFeatures
+    .filter((flag) => typeof flag === 'string')
+    .forEach((flag: string) => {
+      enabledExperimentalFeatures.add(flag)
+    })
+}
+
+export function isExperimentalFeatureEnabled(featureName: string): boolean {
+  return !!enabledExperimentalFeatures && enabledExperimentalFeatures.has(featureName)
```
bcaudan, line 21: no unit test for `!!enabledExperimentalFeatures`, what was your intent?

packages/core/src/domain/configuration/experimentalFeatures.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (!Array.isArray(enabledFeatures)) {
+    return
+  }
+
+  if (!enabledExperimentalFeatures) {
+    enabledExperimentalFeatures = new Set(enabledFeatures)
+  }
+
+  enabledFeatures
+    .filter((flag) => typeof flag === 'string')
```
bcaudan, line 14: no unit test for string filtering, do we need it?

packages/core/src/boot/init.ts
amortemousque: Would it be a valid solution to have a `sanitise` for the entire `initConfiguration`? This way we ensure to have a valid typescript object before any other logic.

jagracey: Cool, switched to `updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures)`

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  it('initial state is empty', () => {
+    resetExperimentalFeatures()
+    expect(isExperimentalFeatureEnabled(undefined)).toBeFalse()
+    expect(isExperimentalFeatureEnabled('foo')).toBeFalse()
+    expect(isExperimentalFeatureEnabled('bar')).toBeFalse()
+    expect(isExperimentalFeatureEnabled('')).toBeFalse()
+  })
+
+  it('updateExperimentalFeatures', () => {
+    updateExperimentalFeatures(['foo'])
```
bcaudan, line 17: it would be preferable to run a `resetExperimentalFeatures` in an `afterEach` block in order to not leak features between tests

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  updateExperimentalFeatures,
+  isExperimentalFeatureEnabled,
+  resetExperimentalFeatures,
+} from './experimentalFeatures'
+
+describe('experimentalFeatures', () => {
+  it('initial state is empty', () => {
+    resetExperimentalFeatures()
+    expect(isExperimentalFeatureEnabled(undefined as any)).toBeFalse()
```
bcaudan, line 10: do we want to use this API with `undefined`?

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    expect(isExperimentalFeatureEnabled('foo')).toBeFalse()
+    expect(isExperimentalFeatureEnabled('bar')).toBeFalse()
+    expect(isExperimentalFeatureEnabled('')).toBeFalse()
```
bcaudan, line 13: do we want to use this API with `''`?

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  })
+
+  it('updateExperimentalFeatures', () => {
```
bcaudan, line 16:
```suggestion
  it('should define enabled experimental features', () => {
```

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    updateExperimentalFeatures(['foo'])
+    expect(isExperimentalFeatureEnabled('foo')).toBeTrue()
+    expect(isExperimentalFeatureEnabled('bar')).toBeFalse()
+  })
+
+  it('resetExperimentalFeatures clears state', () => {
+    resetExperimentalFeatures()
+    expect(isExperimentalFeatureEnabled('foo')).toBeFalse()
+    expect(isExperimentalFeatureEnabled('bar')).toBeFalse()
+  })
```
bcaudan, line 26: IMO since it is only meant to be used in unit tests, no need to add tests for it

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  it("functions won't throw", () => {
```
bcaudan, line 28: what is your intent?

packages/core/src/domain/configuration/experimentalFeatures.ts
jagracey: `!!enabledExperimentalFeatures` is tested when `isExperimentalFeatureEnabled` is called after a `resetExperimentalFeatures`.

packages/core/src/domain/configuration/experimentalFeatures.ts
jagracey: You cannot add flags to a set if the set does not exist. This logic is tested via lines 37,39,40.

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
jagracey: Just checking as a valid common string. I haven't seen a spec around which string types we care about (uppercase, "Turkish dotless i", diacritics, emoji). 

Do we have a team preference on the extent of string validation?

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
jagracey: Testing via unusual calling patterns given sometimes errors can be thrown with unexpected duplicate calls.
Do we have a team preference here on test scope?

jagracey: Updated, moved the string filtering test here.

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    resetExperimentalFeatures()
+    resetExperimentalFeatures()
+    updateExperimentalFeatures(undefined)
+    updateExperimentalFeatures(undefined)
+    expect(isExperimentalFeatureEnabled('foo')).toBeFalse()
+
+    updateExperimentalFeatures(['foo'])
+    updateExperimentalFeatures([11 as any, [] as any, {} as any, null as any, undefined as any])
+    expect(isExperimentalFeatureEnabled('foo')).toBeTrue()
+  })
```
bcaudan, line 33:
```suggestion
  it('should allow to be shared between products', () => {
    updateExperimentalFeatures(['foo'])
    updateExperimentalFeatures(['bar'])
    
    expect(isExperimentalFeatureEnabled('foo')).toBeTrue()
    expect(isExperimentalFeatureEnabled('bar')).toBeTrue()
  })
  
  it('should support some edge cases', () => {
    updateExperimentalFeatures(['foo'])
    updateExperimentalFeatures(undefined)
    updateExperimentalFeatures([])
    updateExperimentalFeatures([11 as any])
    
    expect(isExperimentalFeatureEnabled('foo')).toBeTrue()
  })
```

bcaudan (approved):

PR #1095, opened 2021-09-30
🔊 [RUMF-1036] Add negative loading time internal monitoring

amortemousque:
## Motivation

Some views and actions have a negative loading time.
Because it happens more often on actions, this PR only adds internal monitoring to them.
The first step is to identify if this behaviour comes from `waitIdlePageActivity`.

## Changes

Add internal monitoring on actions with negative loading time

## Testing

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
@@ -7,7 +7,6 @@ import {
   generateUUID,
   ClocksState,
   clocksNow,
-  TimeStamp,
   Configuration,
   ONE_SECOND,
   Observable,
--- a/packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
@@ -11,6 +11,7 @@ import {
   Configuration,
   ONE_SECOND,
   Observable,
+  addMonitoringMessage,
 } from '@datadog/browser-core'
 import { ActionType } from '../../../rawRumEvent.types'
 import { LifeCycle, LifeCycleEventType } from '../../lifeCycle'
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
```diff
@@ -138,14 +139,27 @@ class PendingAutoAction {
...
   complete(endTime: TimeStamp) {
+    const actionDuration = elapsed(this.startClocks.timeStamp, endTime)
+    if (actionDuration < 0) {
+      addMonitoringMessage('auto action with negative loading time', {
+        debug: {
+          actionDuration,
+          type: this.type,
+          name: this.name,
+          startClocks: this.startClocks,
+          endTime,
```
BenoitZugmeyer, line 150: I'm not sure how this information will help you to understand what's going on. The `startClocks.timeStamp` will always be greater than `endTime`, which is the issue I know, but what could we deduce from that?

Maybe storing the "clock state" in `endTime` instead of a timestamp alone could help, we could maybe observe that comparing timestamps produces a negative time, but comparing relative times produces a positive time.

amortemousque: Noted, thanks for the advice. I'll add the end clock state.

BenoitZugmeyer (approved):

packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
```diff
@@ -137,15 +137,28 @@ class PendingAutoAction {
...
+    const actionDuration = elapsed(this.startClocks.timeStamp, endClocks.timeStamp)
+    if (actionDuration < 0) {
+      addMonitoringMessage('auto action with negative loading time', {
+        debug: {
+          actionDuration,
+          type: this.type,
+          name: this.name,
+          startClocks: this.startClocks,
+          endClocks,
+        },
```
bcaudan, line 150: what about adding currentDrift too?
to see if it can be related somehow

PR #1098, opened 2021-10-01
♻️ Use new observable pattern for page activity

amortemousque:
## Motivation

By looking at the trackPageActivity, I've seen that it could benefit from the new observable pattern.

## Changes

- trackPageActivities() => createPageActivityObservable()
- waitIdlePageActivity() => createIdlePageActivityObservable()

## Testing

Unit, Locally

---

### DIFF
--- a/packages/rum-core/src/domain/pageActivityObservable.spec.ts
+++ b/packages/rum-core/src/domain/pageActivityObservable.spec.ts
@@ -1,5 +1,5 @@
-import { noop, Observable, ONE_SECOND, TimeStamp, timeStampNow } from '@datadog/browser-core'
-import { Clock, mockClock } from '../../../core/test/specHelper'
+import { Observable, ONE_SECOND, TimeStamp, timeStampNow } from '@datadog/browser-core'
+import { Clock, mockClock } from '@datadog/browser-core/test/specHelper'
 import { RumPerformanceNavigationTiming, RumPerformanceResourceTiming } from '../browser/performanceCollection'
 import { LifeCycle, LifeCycleEventType } from './lifeCycle'
 import { RequestCompleteEvent } from './requestCollection'
@@ -7,10 +7,10 @@ import {
   PAGE_ACTIVITY_END_DELAY,
   PAGE_ACTIVITY_VALIDATION_DELAY,
   PageActivityEvent,
-  trackPageActivities,
+  createPageActivityObservable,
   waitPageActivitiesCompletion,
-  CompletionCallbackParameters,
-} from './trackPageActivities'
+  PageActivityCompletionEvent,
+} from './pageActivityObservable'
 
 // Used to wait some time after the creation of an action
 const BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY = PAGE_ACTIVITY_VALIDATION_DELAY * 0.8
@@ -34,25 +34,27 @@ function eventsCollector<T>() {
   }
 }
 
-describe('trackPagePageActivities', () => {
+describe('pageActivityObservable', () => {
   const { events, pushEvent } = eventsCollector<PageActivityEvent>()
 
   let lifeCycle: LifeCycle
   let domMutationObservable: Observable<void>
+  let pageActivityObservable: Observable<PageActivityEvent>
 
   beforeEach(() => {
     lifeCycle = new LifeCycle()
     domMutationObservable = new Observable()
+    pageActivityObservable = createPageActivityObservable(lifeCycle, domMutationObservable)
   })
 
   it('emits an activity event on dom mutation', () => {
-    trackPageActivities(lifeCycle, domMutationObservable).observable.subscribe(pushEvent)
+    pageActivityObservable.subscribe(pushEvent)
     domMutationObservable.notify()
     expect(events).toEqual([{ isBusy: false }])
   })
 
   it('emits an activity event on resource collected', () => {
-    trackPageActivities(lifeCycle, domMutationObservable).observable.subscribe(pushEvent)
+    pageActivityObservable.subscribe(pushEvent)
     const performanceTiming = {
       entryType: 'resource',
     }
@@ -61,7 +63,7 @@ describe('trackPagePageActivities', () => {
   })
 
   it('does not emit an activity event when a navigation occurs', () => {
-    trackPageActivities(lifeCycle, domMutationObservable).observable.subscribe(pushEvent)
+    pageActivityObservable.subscribe(pushEvent)
     const performanceTiming = {
       entryType: 'navigation',
     }
@@ -73,13 +75,13 @@ describe('trackPagePageActivities', () => {
   })
 
   it('stops emitting activities after calling stop()', () => {
-    const { stop, observable } = trackPageActivities(lifeCycle, domMutationObservable)
-    observable.subscribe(pushEvent)
+    const observable = createPageActivityObservable(lifeCycle, domMutationObservable)
+    const subscription = observable.subscribe(pushEvent)
 
     domMutationObservable.notify()
     expect(events).toEqual([{ isBusy: false }])
 
-    stop()
+    subscription.unsubscribe()
 
     domMutationObservable.notify()
     domMutationObservable.notify()
@@ -93,14 +95,16 @@ describe('trackPagePageActivities', () => {
     }
     let lifeCycle: LifeCycle
     let domMutationObservable: Observable<void>
+    let pageActivityObservable: Observable<PageActivityEvent>
 
     beforeEach(() => {
       lifeCycle = new LifeCycle()
       domMutationObservable = new Observable()
+      pageActivityObservable = createPageActivityObservable(lifeCycle, domMutationObservable)
     })
 
     it('emits an activity event when a request starts', () => {
-      trackPageActivities(lifeCycle, domMutationObservable).observable.subscribe(pushEvent)
+      pageActivityObservable.subscribe(pushEvent)
       lifeCycle.notify(LifeCycleEventType.REQUEST_STARTED, {
         requestIndex: 10,
       })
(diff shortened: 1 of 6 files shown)


packages/rum-core/src/domain/pageActivityObservable.ts
```diff
@@ -62,65 +50,55 @@ export function waitIdlePageActivity(
...
+
+export function createPageActivityObservable(
   lifeCycle: LifeCycle,
   domMutationObservable: Observable<void>
-): { observable: Observable<PageActivityEvent>; stop: () => void } {
-  const observable = new Observable<PageActivityEvent>()
+): Observable<PageActivityEvent> {
   const subscriptions: Subscription[] = []
   let firstRequestIndex: undefined | number
   let pendingRequestsCount = 0
```
BenoitZugmeyer, line 60: It would be best to move those declarations in the Observable callback

packages/rum-core/src/domain/pageActivityObservable.spec.ts
```diff
@@ -73,13 +75,13 @@ describe('trackPagePageActivities', () => {
   })
 
   it('stops emitting activities after calling stop()', () => {
-    const { stop, observable } = trackPageActivities(lifeCycle, domMutationObservable)
-    observable.subscribe(pushEvent)
+    const observable = createPageActivityObservable(lifeCycle, domMutationObservable)
+    const subscription = observable.subscribe(pushEvent)
```
BenoitZugmeyer, line 79: Any reason to not use `pageActivityObservable` here?

BenoitZugmeyer (approved):

packages/rum-core/src/domain/pageActivityObservable.ts
```diff
@@ -10,32 +10,20 @@ export interface PageActivityEvent {
   isBusy: boolean
 }
 
-export type CompletionCallbackParameters = { hadActivity: true; endClocks: ClocksState } | { hadActivity: false }
+export type PageActivityCompletionEvent = { hadActivity: true; endClocks: ClocksState } | { hadActivity: false }
 
-export function waitIdlePageActivity(
+export function createIdlePageActivityObservable(
```
bcaudan, line 15: should not the file be `idlePageActivityObservable.ts` then?

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts
```diff
@@ -108,15 +108,15 @@ function trackActivityLoadingTime(
...
 ) {
   const startTime = timeStampNow()
-  const { stop: stopWaitIdlePageActivity } = waitIdlePageActivity(lifeCycle, domMutationObservable, (params) => {
+  const subscription = createIdlePageActivityObservable(lifeCycle, domMutationObservable).subscribe((params) => {
     if (params.hadActivity) {
       callback(elapsed(startTime, params.endClocks.timeStamp))
     } else {
       callback(undefined)
     }
   })
```
bcaudan, line 117: it seems a bit weird to me to use an observable when we expect a single notification.

packages/rum-core/src/domain/pageActivityObservable.ts
```diff
@@ -10,32 +10,20 @@ export interface PageActivityEvent {
   isBusy: boolean
 }
 
-export type CompletionCallbackParameters = { hadActivity: true; endClocks: ClocksState } | { hadActivity: false }
+export type PageActivityCompletionEvent = { hadActivity: true; endClocks: ClocksState } | { hadActivity: false }
```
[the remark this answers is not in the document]
amortemousque: Yes, I agree 100%. I'll change it.

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts
amortemousque: Yes, valid argument. I'll change it back.

packages/rum-core/src/domain/pageActivityObservable.ts
amortemousque: true

packages/rum-core/src/domain/waitIdlePageActivity.ts
```diff
@@ -154,16 +59,84 @@ export function waitPageActivitiesCompletion(
...
     stop()
-    completionCallback(params)
+    idlePageActivityCallback(event)
   }
-
   return { stop }
 }
+
+// Automatic action collection lifecycle overview:
+//                      (Start new trackPageActivities)
```
BenoitZugmeyer, line 76: `trackPageActivities` does not exist anymore. Also, it could make sense to move this graph higher in the file.

packages/rum-core/src/domain/waitIdlePageActivity.ts
```diff
@@ -10,119 +10,24 @@ export interface PageActivityEvent {
...
-  )
-
-  const stop = () => {
-    stopWaitPageActivitiesCompletion()
-    stopPageActivitiesTracking()
-  }
-
+  const pageActivityObservable = createPageActivityObservable(lifeCycle, domMutationObservable)
+  const { stop } = doWaitIdlePageActivity(pageActivityObservable, idlePageActivityCallback, maxDuration)
   return { stop }
```
bcaudan, line 23: what about returning `doWaitIdlePageActivity` directly?

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts
bcaudan: Sure, but IMO it:
- is using a more complex concept for no reason, we could use Promises even for synchronous API 
- can be misleading since I would expect that `IdlePageActivityObservable` notify us for every new page idle which is not the case

BenoitZugmeyer: > Sure, but IMO it:
> 
> * is using a more complex concept for no reason, we could use Promises even for synchronous API

Few reasons:

* We can't use Promise since it is not supported by IE11
* Promise cannot be aborted/stopped
* While I would agree that Observables like Rx are more complex than promises, our observable implementation is less complex than Promises.

> 
> * can be misleading since I would expect that `IdlePageActivityObservable` notify us for every new page idle which is not the case

Then it may be a naming issue? It could be renamed as `FirstIdlePageActivityObservable` or `NextIdlePageActivityObservable`.

bcaudan: I was not proposing to use Promises here, it was just another example of using a more complex concept when it is not needed.

amortemousque: I'm not sure to understand the complexity argument since the observable pattern is widely used in our code base. It means that developers already are use to it. However I do understand the misleading aspect of  `IdlePageActivityObservable`, it sure can be solved by changing the name but I'm fine with the callback.

packages/rum-core/src/domain/waitIdlePageActivity.ts
```diff
@@ -10,119 +10,24 @@ export interface PageActivityEvent {
   isBusy: boolean
 }
 
-export type CompletionCallbackParameters = { hadActivity: true; endClocks: ClocksState } | { hadActivity: false }
+export type IdlePageActivityEvent = { hadActivity: true; endClocks: ClocksState } | { hadActivity: false }
```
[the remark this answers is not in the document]
amortemousque: Interesting, it's shorter, I like it :)

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts
bcaudan: About complexity, it is not that observable is complex to use it is just that a callback would do the job.
Why involving a higher concept when a simpler one is enough?

amortemousque: If it's for readability and consistency it's challengeable, but like I said I fine with the callback.

BenoitZugmeyer: I'm not sure it's worth more debate here, but: I had the exact same thought when Promises appeared in the JS ecosystem. We previously used callbacks (continuation passing style) or events everywhere, I didn't see the point of Promises at all and it took me a while to understand it and embrace it.

I'm not saying that it's exactly the same case for Observables. It hasn't reach the same consensus in JS ecosystem as Promises, and anyway the topic is highly subjective. But in my point of view, Observables provide the same advantages over callbacks than Promises. It helps to model and compose asynchronous scenarios, and could be used to unify our "data produced by a function over time that can be stopped" concept.

That said, I won't push for it. I'm happy to see more usages and to use it where it makes sense, but I'm not here to "evangelize". I think it's fine if we use both callbacks and observables.

bcaudan: My driver in this debate is I'd like to prevent a [golden hammer](https://sourcemaking.com/antipatterns/golden-hammer) situation.
From my point of view, observables are suited to notify state changes to many dependents with a light coupling.
We are already using observables in cases where we don't need light coupling or don't have several dependents, I don't think we should actively change it. 
However, I'd prefer to not refactor existing code to use this pattern when the situation does not require it.

BenoitZugmeyer: > I'd prefer to not refactor existing code to use this pattern when the situation does not require it.

Isn't that the point of this PR?

packages/rum-core/src/domain/waitIdlePageActivity.ts
amortemousque: Agreed

bcaudan (approved):

PR #1106, opened 2021-10-04
⚗[REPLAY-465] Base tag support (feature flagged): Remove URL transformation from relative to absolute

jagracey:
## Motivation

As per internal RFC ["Supporting Replay of Relative URLs When <base/> Tag Used"](https://docs.google.com/document/d/193t_-nHSN4YLGRKkf51-l--TWrn58QKnsmP5uA_1GI8/edit?usp=sharing), we want to support the[ <base/> tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base). In short, this tag will cause the document to resolve relative URLs with the provided URL instead.

## Changes
We don't need to transform URLs from relative to absolute on the SDK recorder side. Instead we will handle this on the replay side.

## Testing
Removes tests for code that transforms URLs. Manual testing for now behind a feature flag.

---


acorretti (approved):

bcaudan (approved):

packages/rum/src/domain/record/serializationUtils.ts
```diff
@@ -58,6 +64,9 @@ export function makeSrcsetUrlsAbsolute(attributeValue: string, baseUrl: string)
 
 export function makeUrlAbsolute(url: string, baseUrl: string): string {
   try {
+    if (isExperimentalFeatureEnabled('base-tag')) {
+      return url.trim()
```
BenoitZugmeyer, line 68: Do you need the `.trim()` here? I'd suggest removing it.

jagracey: I was pondering that myself. Why did we trim in the first place?

PR #1112, opened 2021-10-05
👷 Remove gitlab jobs on release branches

amortemousque:
## Motivation

During our release process, the release branch and the tag run nearly the same ci jobs, making the checks twice. 
This PR removes the release branch jobs to save resources and time.

## Changes

Update gitlab-ci.yml

## Testing

Gitlab

---

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,6 +25,36 @@ stages:
   - deploy
   - notify
 
+########################################################################################################################
+# Configuration helpers
+########################################################################################################################
+
+.base-configuration:
+  tags: ['runner:main', 'size:large']
+  image: $CI_IMAGE
+
+.feature-branch-configuration:
+  extends: .base-configuration
+  except:
+    refs:
+      - main
+      - tags
+      - /^staging-[0-9]+$/
+      - /^release\/v[0-9.]+$/
+      - schedules
+
+.test-configuration:
+  stage: test
+  extends: .base-configuration
+  except:
+    refs:
+      - /^release\/v[0-9.]+$/
+      - schedules
+
+########################################################################################################################
+# CI image
+########################################################################################################################
+
 ci-image:
   stage: ci-image
   when: manual
@@ -43,44 +73,24 @@ ci-image:
 ########################################################################################################################
 
 format:
-  stage: test
-  except:
-    refs:
-      - schedules
-  tags: ['runner:main', 'size:large']
-  image: $CI_IMAGE
+  extends: .test-configuration
   script:
     - yarn
     - yarn format
 
 woke:
-  stage: test
-  except:
-    refs:
-      - schedules
-  tags: ['runner:main', 'size:large']
-  image: $CI_IMAGE
+  extends: .test-configuration
   script:
     - yarn woke
 
 typecheck:
-  stage: test
-  except:
-    refs:
-      - schedules
-  tags: ['runner:main', 'size:large']
-  image: $CI_IMAGE
+  extends: .test-configuration
   script:
     - yarn
     - yarn typecheck
 
 build-and-lint:
-  stage: test
-  except:
-    refs:
-      - schedules
-  tags: ['runner:main', 'size:large']
-  image: $CI_IMAGE
+  extends: .test-configuration
   script:
     - yarn
     - yarn build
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -43,44 +75,24 @@ ci-image:
 ########################################################################################################################
 
 format:
-  stage: test
-  except:
-    refs:
-      - schedules
-  tags: ['runner:main', 'size:large']
-  image: $CI_IMAGE
+  extends: .test-configuration
```
BenoitZugmeyer, line 78: FMU this removes the `format` job from release branches. I think it's worth keeping it, because it applies on the CHANGELOG too.

amortemousque:
By removing the release branch jobs, browserstack tests are no longer running during the release process.
Two options:
- Adding them to the tag pipeline
- Considering reasonable that it has been checked at PR stage

.gitlab-ci.yml
```diff
@@ -25,12 +25,44 @@ stages:
...
+  image: $CI_IMAGE
+
+.feature-branch-configuration:
+  extends: .base-configuration
+  except:
+    refs:
+      - main
+      - tags
+      - /^staging-[0-9]+$/
+      - /^release\/v[0-9.]+$/
```
bcaudan, line 43:
```suggestion
      - /^release\//
```
let's be a bit more permissive on the release branch name
By looking at my history, I have some `release/X.Y.Z` branches and I don't think we need to enforce the `release/vX.Y.Z` format

.gitlab-ci.yml
```diff
@@ -25,12 +25,44 @@ stages:
...
+########################################################################################################################
+
 ci-image:
   stage: ci-image
   when: manual
   except:
     refs:
       - tags
+      - /^staging-[0-9]+$/
+      - /^release\/v[0-9.]+$/
```
BenoitZugmeyer, line 65: What about using `.feature-branch-configuration` here? I don't think we'll need to build the CI image outside of a PR.

amortemousque: Perfect, I'll use it then

.gitlab-ci.yml
BenoitZugmeyer: Ah right, the tags will trigger another pipeline. It looks fine then!

.gitlab-ci.yml
```diff
@@ -25,12 +25,44 @@ stages:
...
+      - schedules
+
+.test-configuration:
+  stage: test
+  extends: .base-configuration
+  except:
+    refs:
+      - /^release\/v[0-9.]+$/
+      - schedules
```
bcaudan, line 52: I found a bit hard to grasp currently since `test-configuration` and `feature-branch-configuration` are not comparable.
what about extracting configuration that only help us to select branches?
we could have configuration for feature branch, tag, main, ... and for each job specify on which kind of branch we want to run it.

something like:
```
format:
  extends: 
    - .main
    - .staging
    - .feature-branch
```
wdyt?

amortemousque: I understand your point, I first tried this strategy but I ended up having the same amount of code while I wanted to shorten our configuration. 
Also you're example is sweet because the items in the extends section are of the same nature, but this example bellow is way less elegant and understandable:
```
format:
  extends: 
    - .base-configuration
    - .main
    - .staging
    - .potential-future-config
```

.gitlab-ci.yml
amortemousque: Right

.gitlab-ci.yml
bcaudan: fair enough, we could avoid composition among branches configuration and still extract branches/tag logic:

- `test-configuration` -> `test-allowed-branches` and remove `stage: test` and `.base-configuration`
- `feature-branch-configuration` -> `feature-branch`
- `tag` for  only refs tags
- `staging` for current staging

then we can have:
```
format:
  extends:
    - .base-configuration
    - .test-allowed-branches

notify-staging-failure:
  extends: 
    - .prepare_notification
    - .staging
```

amortemousque: I'm not 100% convince of the benefit but why not. I will change `.staging` to `.staging-branch` though.

PR #1118, opened 2021-10-06
⚗[REPLAY-341] Add VisualViewport tracking (Pinch Zoom)

jagracey:
## Motivation

Support pinch zoom and pinch scroll by tracking the "visual viewport".

As mentioned in the internal RFC ["Supporting Pinch Zoom Tracking"](https://docs.google.com/document/d/1GOUNsiQizPzANRUvLQW0R-WHuGsOe9p4b-oO3XALU2w/edit?usp=sharing), unfortunately, the dimensions we currently record are not consistent across devices and browsers as some are referenced. On certain devices (primarily mobile), when users pinch zoom, the replay will replay incorrectly- displaying the layout viewport but not the visual viewport.

Unfortunately, browsers have not standardised various dimension properties. Mobile devices typically report dimensions in reference to the visual viewport, while desktop uses the layout viewport. For example, mobile Chrome will change innerWidth when a pinch zoom takes place, while Chrome desktop (mac) will not

With the new Viewport API, we now calculate and normalise dimension properties to the layout viewport.

**Note:** If the VisualViewport API is not supported by a browser, it isn't reasonably possible to detect or normalise which viewport is being measured. Therefore these exported functions will fallback to assuming that the layout viewport is being measured by the browser

## Changes
1. Adds VisualViewport record as incremental record.
2. Changes innerWidth/innerHeight, scrollX/scrollY, and mouse positions to a normalised value that is referenced to the layout viewport (not pinch zoom's visual viewport) so that we do not factor in the visual viewport (pinch zoom) twice for mobile devices.

## Testing
To support pinch zoom and pinch scroll unit tests, we use utilise our E2E testing framework to use Appium's gesture commands.

Unfortunately there are many pitfalls to be aware of that required work arounds:
- Scrolling behaviour may disappear the address bar changing screen height
- Address bar height can persist across page changes (record flushing)
- FullScreen mode isn't consistent with regard to pinch zoom
- Pinch/touch gestures must not exceed the bounds of the page, so are commonly repeated together.
- Pinch scroll is inverted on iOS vs Android
- Click() method will trigger a scroll into view, so testing the coordinate system of layout vs visual viewports for clicks doesn't appear possible
- Chrome Linux doesn't support the meta viewport tag to limit max scale
- Chrome Linux `innerWidth` vs `visual.width*scale` appears to differ by 15 pixels likely due to scroll bars.

---


packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,119 @@
...
+  const visual = window.visualViewport
+  if (isExperimentalFeatureEnabled('visualviewport') && visual) {
+    return visual.pageLeft
+  }
+  if (window.scrollX !== undefined) {
+    return window.scrollX
+  }
+  return window.pageXOffset !== undefined
+    ? window.pageXOffset
+    : (document?.documentElement || (document?.body?.parentNode as Element) || document?.body).scrollLeft || 0
```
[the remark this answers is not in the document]
jagracey: Good point. Using the existing scroll fallback approach probably isn't needed for our use-cases anymore.

packages/rum/src/domain/record/viewports.ts
```diff
@@ -0,0 +1,119 @@
...
+// Scrollbar widths vary across properties on different devices and browsers
+const TOLERANCE = 25
+
+const isVisualViewportFactoredIn = () => {
+  const visual = window.visualViewport
+  return (
+    Math.abs(visual.pageTop - visual.offsetTop - window.scrollY) > TOLERANCE ||
+    Math.abs(visual.pageLeft - visual.offsetLeft - window.scrollX) > TOLERANCE
+  )
+}
```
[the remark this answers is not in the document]
jagracey: Added a comment in code now:
>  Use the Visual Viewport API's properties to measure scrollX/Y in reference to the layout viewport in order to determine if window.scrollX/Y is measuring the layout or visual viewport. This finding corresponds to which viewport mouseEvent.clientX/Y and window.innerWidth/Height measures.

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -21,6 +29,38 @@ const INTEGER_RE = /^\d+$/
...
+
+const isGestureUnsupported = () => {
+  const { capabilities } = browser
+  return !!(
+    capabilities.browserName === 'firefox' ||
+    capabilities.browserName === 'Safari' ||
+    capabilities.browserName === 'msedge' ||
+    capabilities.platformName === 'windows' ||
+    capabilities.platformName === 'linux'
+  )
```
bcaudan, line 50: `!!()` seems unneeded

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -21,6 +29,38 @@ const INTEGER_RE = /^\d+$/
...
 
+const VIEWPORT_META_TAGS = `
+<meta name="apple-mobile-web-app-capable" content="yes">
+<meta name="mobile-web-app-capable" content="yes">
+<meta name="viewport"
+  content="width=device-width, initial-scale=1.0, maximum-scale=2.75, minimum-scale=1.0, user-scalable=yes"
+>
+`
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
```
bcaudan, line 40: https://webdriver.io/docs/api/browser/pause

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -21,6 +29,38 @@ const INTEGER_RE = /^\d+$/
...
+    capabilities.browserName === 'firefox' ||
+    capabilities.browserName === 'Safari' ||
+    capabilities.browserName === 'msedge' ||
+    capabilities.platformName === 'windows' ||
+    capabilities.platformName === 'linux'
+  )
+}
+
+// Flakiness: Working with viewport sizes has variations per device of a few pixels
+function expectToBeNearby(numA: number, numB: number) {
```
bcaudan, line 54: what about moving the new helpers at the bottom of the file with the others?

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
+
+        const lastViewportResizeRecord = ViewportResizeRecords.slice(-1)[0].data as ViewportResizeData
+
+        expectToBeNearby(Math.round(lastViewportResizeRecord.width), innerWidth)
+        expectToBeNearby(Math.round(lastViewportResizeRecord.height), innerHeight)
+        // Test the test: ensure the pinch zoom worked
+        expect(initialVisualViewport.scale < nextVisualViewport.scale).toBeTruthy()
+      })
+
+    createTest('scrollX/Y reports using layout viewport dimensions')
```
bcaudan, line 742: this test fails on my machine (I had to merge main to get the chrome driver update)

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
+        expectToBeNearby(nextVisualViewport.pageTop, preVisualViewport.pageTop + SCROLL_DOWN_PX - heightChange)
+        expectToBeNearby(nextVisualViewport.offsetTop, preVisualViewport.offsetTop + SCROLL_DOWN_PX - heightChange)
+
+        expectToBeNearby(lastScrollRecord.x, nextVisualViewport.pageLeft)
+        expectToBeNearby(lastScrollRecord.y, nextVisualViewport.pageTop)
+      })
+  })
+
+  describe('visual viewport properties - ', () => {
+    createTest('pinch scroll event tracked reports visual viewport page offset')
```
bcaudan, line 818: this test fails on my machine

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
+         */
+        if (isGestureUnsupported()) {
+          return // No Fallback test
+        }
+        await resetViewport()
+
+        const { innerWidth, innerHeight } = (await browserExecute(() => ({
+          innerWidth: window.innerWidth,
+          innerHeight: window.innerHeight,
+        }))) as { innerWidth: number; innerHeight: number }
```
bcaudan, line 717: https://webdriver.io/docs/api/browser/getWindowSize ?

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
+        })
+        await sleep(210)
+
+        await flushEvents()
+        const segment = getLastSegment(events)
+        const ViewportResizeRecords = findAllIncrementalSnapshots(segment, IncrementalSource.ViewportResize)
+
+        const lastViewportResizeRecord = ViewportResizeRecords.slice(-1)[0].data as ViewportResizeData
+
+        expectToBeNearby(Math.round(lastViewportResizeRecord.width), innerWidth)
```
bcaudan, line 736: what about rounding inside your helper?
it would remove some noise from the scenario

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
+        await browserExecute(() => {
+          window.dispatchEvent(new Event('resize'))
+        })
+        await sleep(210)
+
+        await flushEvents()
+        const segment = getLastSegment(events)
+        const ViewportResizeRecords = findAllIncrementalSnapshots(segment, IncrementalSource.ViewportResize)
+
+        const lastViewportResizeRecord = ViewportResizeRecords.slice(-1)[0].data as ViewportResizeData
```
bcaudan, line 734: no clue why:

- 150
- two pinches
- need to sleep
- 210
- window resize

it could worth some explanation

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
   return events.sessionReplay[0].segment.data
 }
 
-function getLastSegment(events: EventRegistry) {
+export function getLastSegment(events: EventRegistry) {
   return events.sessionReplay[events.sessionReplay.length - 1].segment.data
 }
 
-function initRumAndStartRecording(initConfiguration: RumInitConfiguration) {
+export function initRumAndStartRecording(initConfiguration: RumInitConfiguration) {
```
bcaudan, line 925: any reason to export them?

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
   window.DD_RUM!.init(initConfiguration)
   window.DD_RUM!.startSessionReplayRecording()
 }
+
+export async function pinchZoom(xChange = 50, durationMS = 400) {
```
bcaudan, line 930: why xChange=50?
it seems that most of the time you are using it with 150

you never use the durationMS parameter so it could be another constant inside the function

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,17 +694,305 @@ describe('recorder', () => {
...
+  const actions = [
+    {
+      type: 'pointer',
+      id: 'finger1',
+      parameters: { pointerType: 'touch' },
+      actions: [
+        { type: 'pointerMove', duration: 0, x: xBase, y: yBase },
+        { type: 'pointerDown', button: 0 },
+        { type: 'pause', duration: 150 },
+        { type: 'pointerMove', duration: durationMS, origin: 'pointer', x: -xChange, y: 0 },
```
bcaudan, line 943: it could be interesting to add some explanation for all the magic numbers, are they random or are they important?

packages/rum/test/utils.ts
```diff
@@ -152,6 +159,11 @@ export function findFullSnapshot(segment: Segment): FullSnapshotRecord | null {
   return segment.records.find((record) => record.type === RecordType.FullSnapshot) as FullSnapshotRecord
 }
 
+// Returns all the VisualViewportRecords in a Segment, if any.
+export function findAllVisualViewport(segment: Segment): VisualViewportRecord[] {
```
bcaudan, line 163:
```suggestion
// Returns all the VisualViewportRecords in a Segment, if any.
export function findAllVisualViewports(segment: Segment): VisualViewportRecord[] {
```

test/e2e/lib/helpers/browser.ts
```diff
@@ -74,3 +74,35 @@ export async function sendXhr(url: string, headers: string[][] = []): Promise<st
...
+    }
+  }) as Promise<VisualViewportData>
+}
+
+export function getWindowScroll() {
+  return browserExecute(() => ({
+    scrollX: window.scrollX,
+    scrollY: window.scrollY,
+  })) as Promise<{ scrollX: number; scrollY: number }>
+}
```
bcaudan, line 108: what about keeping them in recorder scenario if we don't use them elsewhere?

test/e2e/scenario/recorder.scenario.ts
jagracey: The underlying JSONWP spec wasn't clear what is being returned here, I suspect it is outerWidth/height- not sure if scrollbar width is included for all device browsers. 

While the code is less clean, it's perhaps more accurate to directly test `innerWidth`'s relationship with `visualViewport.scale * visualViewport.width` to be sure they are the same (which is done throughout this file)

test/e2e/scenario/recorder.scenario.ts
jagracey: Great questions:
- first off, not sure why sleeping helps, but it reliably helps with flakiness. 210ms was picked and used consistently because it was the highest amount that the SDK throttles that also feels low enough (100-200ms)- though FMU throttling should not be responsible for flakiness, but I have no other leads.
- Pinch zooms are not setting to 1.5x zoom, but rather moving fingers 150px in each direction- which can exceed the bounds of the device's screen causing an error- so two pinches are used to zoom in "a bunch" (an unknown/undefined amount). Having less pinch zoom (1 call) has less clear differences from `innerWidth/height`, while having 3 would be excessive IMO.
- The dispatched resize event triggers the SDK to take a record which is what we are testing.

The purpose of this test is to check that our new browser dimensions are measuring the layout viewport, and so they do not change when pinch zoom is applied.  I'll add a comment in the code.

test/e2e/scenario/recorder.scenario.ts
bcaudan: did you test it?

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,6 +670,236 @@ describe('recorder', () => {
...
+        // Scrollbar edge-case handling not considered right now, further investigation needed
+        let scrollbarWidth = 0
+        if (browser.capabilities.browserName === 'chrome' && browser.capabilities.platformName === 'mac os x') {
+          scrollbarWidth = await getScrollbarWidth()
+        }
+        expectToBeNearby(lastViewportResizeRecord.width, innerWidth - scrollbarWidth)
+        expectToBeNearby(lastViewportResizeRecord.height, innerHeight - scrollbarWidth)
+
+        // Test the test: ensure the pinch zoom worked
+        expect(initialVisualViewport.scale < nextVisualViewport.scale).toBeTruthy()
```
bcaudan, line 725: it would be way easier to read your test by abstracting some operations.
FMU, we could reduce your test to something like:
```suggestion
        const layoutViewport = await getLayoutViewport()
        await performSignificantZoom()
        const resizeRecord = await sendResizeRecord(events)
        const scrollbarCorrection = await getScrollbarCorrection()

        expectToBeNearby(resizeRecord.width, layoutViewport.width - scrollbarCorrection)
        expectToBeNearby(resizeRecord.height, layoutViewport.height - scrollbarCorrection)
```
wdyt?

test/e2e/scenario/recorder.scenario.ts
bcaudan: > first off, not sure why sleeping helps, but it reliably helps with flakiness. 210ms was picked and used consistently because it was the highest amount that the SDK throttles that also feels low enough (100-200ms)- though FMU throttling should not be responsible for flakiness, but I have no other leads.

From past experiences, my take is that relying on timeout is rarely a persistent solution. Could we find why we need to wait and see if we can rely on something else than time?

> Pinch zooms are not setting to 1.5x zoom, but rather moving fingers 150px in each direction- which can exceed the bounds of the device's screen causing an error- so two pinches are used to zoom in "a bunch" (an unknown/undefined amount). Having less pinch zoom (1 call) has less clear differences from innerWidth/height, while having 3 would be excessive IMO.

cf https://github.com/DataDog/browser-sdk/pull/1118#discussion_r733431423, it could be nice to abstract this operation and add this kind of explanation with it can also help

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,6 +670,236 @@ describe('recorder', () => {
...
+      .run(async ({ events }) => {
+        /**
+         * The purpose of this test is to check that browser dimensions are measuring the
+         * layout viewport (not visual viewport), and so they do not change when pinch zoom is applied.
+         *
+         * InnerWidth/Height on some devices/browsers are changed by pinch zoom.
+         * Read vendor discussions around the challenges of standardization:
+         * - https://bugs.chromium.org/p/chromium/issues/detail?id=489206
+         */
+        if (isGestureUnsupported()) {
```
bcaudan, line 689: is this comment related to the test or to the check on gesture?

test/e2e/scenario/recorder.scenario.ts
```diff
@@ -654,6 +670,236 @@ describe('recorder', () => {
...
+          return // No Fallback test
+        }
+        await resetViewport()
```
bcaudan, line 692: what is the goal of that?
I've removed it and all tests seems to still be successful 🤔

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,371 @@
...
+        const lastViewportResizeData = (
+          await getLastRecord(events, (segment) =>
+            findAllIncrementalSnapshots(segment, IncrementalSource.ViewportResize)
+          )
+        ).data as ViewportResizeData
+
+        const scrollbarWidth = await getScrollbarCorrection()
+
+        expectToBeNearby(lastViewportResizeData.width, innerWidth - scrollbarWidth)
+        expectToBeNearby(lastViewportResizeData.height, innerHeight - scrollbarWidth)
```
bcaudan, line 49:
```suggestion
        expectToBeNearby(lastViewportResizeData.height, innerHeight - scrollbarCorrection)
```
a bit weird to subtract a width from a height, no?

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,371 @@
...
+      height: visual.height,
+      offsetLeft: visual.offsetLeft,
+      offsetTop: visual.offsetTop,
+      pageLeft: visual.pageLeft,
+      pageTop: visual.pageTop,
+    }
+  }) as Promise<VisualViewportData>
+}
+
+function getWindowScroll() {
```
bcaudan, line 298:
```suggestion
function getLayoutScroll() {
```
to be consistent with `resetLayoutScroll` naming

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,371 @@
...
+
+        await buildScrollablePage()
+        await performSignificantZoom()
+        await resetLayoutScroll()
+
+        const initialVisualViewport = await getVisualViewport()
+        const { scrollX: initialScrollX, scrollY: initialScrollY } = await getWindowScroll()
+
+        // Add Visual Viewport Scroll
+        await pinchScrollVerticallyDown(VISUAL_SCROLL_DOWN_PX)
```
bcaudan, line 73: why not having a `visualScrollTo` helper to be symetric with `layoutScrollTo`?

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,371 @@
...
+          document.dispatchEvent(new Event('scroll'))
+        })
+
+        const lastScrollData = (
+          await getLastRecord(events, (segment) => findAllIncrementalSnapshots(segment, IncrementalSource.Scroll))
+        ).data as ScrollData
+
+        // Height changes because URL address bar changes due to scrolling
+        const navBarHeightChange = nextVisualViewport.height - initialVisualViewport.height
+        expect(navBarHeightChange).toBeLessThanOrEqual(30)
```
bcaudan, line 91: where does 30 come from?

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,371 @@
...
+        // Both Layout + Visual scroll should affect "page" dimensions
+        expectToBeNearby(
+          nextVisualViewport.pageTop,
+          LAYOUT_SCROLL_AMOUNT + VISUAL_SCROLL_DOWN_PX - navBarHeightChange + initialVisualViewport.pageTop
+        )
+        expectToBeNearby(nextVisualViewport.pageLeft, LAYOUT_SCROLL_AMOUNT + initialVisualViewport.pageLeft)
+      })
+  })
+
+  describe('visual viewport properties', () => {
```
bcaudan, line 116: IIRC, from last talk we said that we'll remove this test.
anything change?

test/e2e/scenario/recorder/viewports.scenario.ts
jagracey: The height change is usually be 20px, empirically. So add a bit of a buffer.

bcaudan: what about extracting a constant for it?

test/e2e/scenario/recorder/viewports.scenario.ts
jagracey: Unfortunately browsers are not consistent on the `window.scrollX/Y` specification. This could be either layout scroll, or layout+visual scroll. For the devices we test on, they happen to be consistent- but I'd recommend we stick with `getWindowScroll` to make it clear that it is the native behaviour- which is loosely defined.

test/e2e/scenario/recorder/viewports.scenario.ts
jagracey: Yeah. Correction is probably better than something like `scrollbarThickness`.

test/e2e/scenario/recorder/viewports.scenario.ts
bcaudan: so `resetWindowScroll` then?

jagracey: Funny enough `resetLayoutScroll` does in fact reset the layout scroll though has a somewhat poorly defined affect on visual scroll. But yes, this would follow.

test/e2e/scenario/recorder/viewports.scenario.ts
jagracey: Unfortunately `layoutScrollTo` uses the native JS method to scroll to an absolute position whereas `pinchScrollVerticallyDown` uses the browser automation API to relatively pinch.
Unfortunately the nature of pinchScroll means that the direction is poorly defined (iOS vs Android) and that pinch scrolling too much will "overflow" into the layout scroll.

Renaming `pinchScrollVerticallyDown` to `visualScrollVerticallyDown` might be more consistent, but it wouldn't be perfectly accurate either.

test/e2e/scenario/recorder/viewports.scenario.ts
jagracey: Resolved IRL

bcaudan (approved):
LGTM

packages/rum/src/domain/record/record.ts
```diff
@@ -37,26 +39,21 @@ export function record(options: RecordOptions): RecordAPI {
...
       type: RecordType.FullSnapshot,
     })
   }
 
+  if (isExperimentalFeatureEnabled('visualviewport') && window.visualViewport) {
+    emit({
+      data: getVisualViewport(),
+      type: RecordType.VisualViewport,
+    })
+  }
```
BenoitZugmeyer, line 55: Shouldn't this be done at every full snapshot like Focus and Meta?

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,351 @@
...
+      .withBody(html`${VIEWPORT_META_TAGS}`)
+      .run(async ({ events }) => {
+        await buildScrollablePage()
+
+        const { innerWidth, innerHeight } = await getWindowInnerDimensions()
+        await performSignificantZoom()
+
+        await browserExecute(() => {
+          window.dispatchEvent(new Event('resize'))
+        })
```
BenoitZugmeyer, line 39: Why is this needed?

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,351 @@
...
+
+        // Add Layout Viewport Scroll
+        await layoutScrollTo(LAYOUT_SCROLL_AMOUNT, LAYOUT_SCROLL_AMOUNT)
+
+        const nextVisualViewport = await getVisualViewport()
+        const { scrollX: nextScrollX, scrollY: nextScrollY } = await getWindowScroll()
+
+        await browserExecute(() => {
+          document.dispatchEvent(new Event('scroll'))
+        })
```
BenoitZugmeyer, line 84: Why is this needed?

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -0,0 +1,351 @@
...
+  )
+}
+
+// Flakiness: Working with viewport sizes has variations per device of a few pixels
+function expectToBeNearby(numA: number, numB: number) {
+  const roundedA = Math.round(numA)
+  const roundedB = Math.round(numB)
+  const test = Math.abs(roundedA - roundedB) <= 5
+  if (test) {
+    expect(test).toBeTruthy()
```
BenoitZugmeyer, line 158: This seems useless, maybe remove it?

BenoitZugmeyer: NIT: I would write the whole `expectToBeNearby` assertion as:

```ts
expect(Math.abs(numA - numB)).toBeLessThan(5)
```
(I don't think we need to round numbers)

test/e2e/scenario/recorder/viewports.scenario.ts
BenoitZugmeyer: For the record: this is to generate a `ViewportResize` event without having to explicitly resize the browser (which may be more complex/flaky). LGTM

PR #1122, opened 2021-10-07
🐛  Discard views and actions with negative loading time

amortemousque:
## Motivation

It happens to have negative loading time on view and actions. It’s really rare around 0.000175% of click actions.
To keep meaningful data, let's discard them.

## Changes

- Discard actions with negative loading time
- Discard view loading_time value if negative

## Testing

Unit

---


packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
```diff
@@ -136,18 +131,16 @@ class PendingAutoAction {
...
+  complete(event: IdlePageEvent) {
+    if (!event.hadActivity) {
+      this.discard()
+      return
+    }
+
+    const duration = elapsed(this.startClocks.timeStamp, event.endClocks.timeStamp)
+    if (duration < 0) {
+      this.discard()
+      return
```
bcaudan, line 143: since `complete` and `discard` are two separate behaviors, what about keeping them separated?
we could only pass duration to `complete` if you want to avoid to compute it twice

packages/rum-core/src/domain/rumEventsCollection/action/trackActions.spec.ts
```diff
@@ -121,6 +126,16 @@ describe('trackActions', () => {
     expect(discardSpy).toHaveBeenCalled()
   })
 
+  it('discards a pending action with a negative duration', () => {
+    const { domMutationObservable, clock } = setupBuilder.build()
+    mockValidatedClickAction(domMutationObservable, clock, button, -BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
```
bcaudan, line 131:
```suggestion
    mockValidatedClickAction(domMutationObservable, clock, button, -1)
```
or is there any reason to use this constant here?

PR #1123, opened 2021-10-07
👷 Auto deploy on org 2

amortemousque:
## Motivation

With the possibility of testing PRs on staging without merging them to main, we could choose to automatically deploy the canary bundle for every commit on main.
Everything merged on main should be production ready and able to be released publicly at any time. For now, the actual public release would still be manual.

## Changes

Update gitlab-ci

## Testing

---


.gitlab-ci.yml
```diff
@@ -286,7 +285,7 @@ notify-prod-canary-success:
   stage: notify:canary
   extends:
     - .prepare_notification
-    - .tags
```
BenoitZugmeyer, line 289: You probably need the same in `notify-prod-canary-failure`?

.gitlab-ci.yml
```diff
@@ -225,8 +225,7 @@ deploy-prod-canary:
   stage: deploy:canary
   extends:
     - .base-configuration
-    - .tags
-  when: manual
+    - .main
   allow_failure: false
```
bcaudan, line 229: allow failure should not be needed anymore

.gitlab-ci.yml
```diff
@@ -225,8 +225,7 @@ deploy-prod-canary:
   stage: deploy:canary
   extends:
     - .base-configuration
-    - .tags
-  when: manual
+    - .main
```
bcaudan, line 228: could we avoid to deploy release PR merge?
it will be unnecessary noise

.gitlab-ci.yml
```diff
@@ -225,8 +225,7 @@ deploy-prod-canary:
   stage: deploy:canary
   extends:
     - .base-configuration
-    - .tags
-  when: manual
```
bcaudan, line 229: we should not `BUILD_MODE=release yarn build:bundle` for canary anymore otherwise we won't be able to differentiate the commits

.gitlab-ci.yml
amortemousque: Sure, the only solution I found is to check the commit title:
https://github.com/DataDog/browser-sdk/blob/d1a707e2a22dcd2a290fa8e87e064890d5321f18/.gitlab-ci.yml#L72

.gitlab-ci.yml
amortemousque: sure

.gitlab-ci.yml
amortemousque: I created a `canary` build mode, which is the same as `staging`. I found it clearer instead of reusing staging. Wdyt?

.gitlab-ci.yml
```diff
@@ -68,6 +68,8 @@ stages:
   except:
     refs:
       - schedules
+    variables:
+      - $CI_COMMIT_TITLE =~ /^v[0-9.]+$/
```
bcaudan, line 72: By looking at commit on main, we should probably take into account that the commit title contains the related PR, ex:
```
v3.6.11 (#1126)
```

scripts/build-env.js
```diff
@@ -6,6 +6,7 @@ switch (process.env.BUILD_MODE) {
   case 'release':
     sdkVersion = lernaJson.version
     break
+  case 'canary':
```
bcaudan, line 9: what about reflecting it in [BuildMode](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/boot/init.ts#L40) too?

.gitlab-ci.yml
```diff
@@ -68,6 +68,8 @@ stages:
   except:
     refs:
       - schedules
+    variables:
+      - $CI_COMMIT_TITLE =~ /^v[0-9.]+ .+$/
```
bcaudan, line 72:
```suggestion
      - $CI_COMMIT_TITLE =~ /^v[0-9.]+/
```
nit

PR #1124, opened 2021-10-07
[RUMF-1046] make fetchProxy stoppable

BenoitZugmeyer:
## Motivation

This PR uses the previously introduced `instrumentMethod` function 

## Changes

This PR changes the fetchProxy to return an Observable. `fetchProxy` is used during network collection and resource collection, two independent parts. Using an `Observable` allows to keep lazily start it and make sure it stops when subscribers are unsubscribed. Because there is two kind of context to dispatch (start and complete), a new `context.state: "start" | "complete"` property has been introduced, allowing to distinguish between the two.

## Testing

---


packages/core/src/browser/fetchObservable.ts
```diff
@@ -0,0 +1,127 @@
...
+      context.response = response
+      context.responseText = text
+      context.responseType = response.type
+      context.status = response.status
+      context.isAborted = false
+
+      observable.notify(context)
+    }
+  }
+  responsePromise.then(monitor(reportFetch), monitor(reportFetch))
```
amortemousque, line 126: Nit: Would 2 different function to handle success and error avoid the if statement above?

BenoitZugmeyer:
> Since we are changing it. What do you think of moving the requestCollection to `/tracing` since its only responsibility is to trace request?

It is also used to generate REQUEST_COMPLETED / REQUEST_STARTED events to:
* generate RUM Resource events
* wait for page idle

packages/core/src/browser/fetchObservable.ts
bcaudan: IMO, not a big deal, I'd say let's stay focus on the SDK stoppable topic

bcaudan (approved):

amortemousque (approved):

PR #1133, opened 2021-10-12
[RUMF-1046] make xhrProxy stoppable

BenoitZugmeyer:
## Motivation

Similarly to #1124 , the PR changes the XHR proxy to make it stoppable.

## Changes

* Use Observable to track XHR calls
* Use instrumentMethod to be able to stop the instrumentation

## Testing

Manual, CI

---


packages/core/src/browser/xhrObservable.ts
```diff
@@ -0,0 +1,158 @@
...
+function createXhrObservable() {
+  const observable = new Observable<XhrContext>(() => {
+    const { stop: stopInstrumentingStart } = instrumentMethod(
+      XMLHttpRequest.prototype,
+      'open',
+      (original) =>
+        function (method, url) {
+          callMonitored(openXhr, this, [method, url])
+          return original.apply(this, arguments as any)
+        }
```
BenoitZugmeyer, line 60: This pattern seems to be quite used, so we may have a helper function to factorize it, ex:

```ts

const { stop: stopInstrumentingStart } = instrumentMethodAndCallOriginal(
      XMLHttpRequest.prototype,
      'open',
      openXhr
)
```

The only thing is that it doesn't allow to modify the arguments, which is needed for tracking fetch., so we need to keep the current `instrumentMethod` implementation

packages/core/test/specHelper.ts
```diff
@@ -189,7 +189,7 @@ class StubEventEmitter {
   }
 }
 
-class StubXhr extends StubEventEmitter {
+export class StubXhr extends StubEventEmitter {
```
bcaudan, line 192: needed?

PR #1134, opened 2021-10-14
♻️ 🔥 adjust tracekit codebase a bit

BenoitZugmeyer:
## Motivation

Tracekit is using a few global state that aren't needed because we don't `report` errors through it. Before using `instrumentMethod` on it, I wanted to cleanup a bit this unused logic.

Also, I wanted to move tests around so test cases where in their related module spec file. One small downside is that `computeStackTrace` spec file is now quite large, but I still think it's improving discoverability.

## Changes

* remove `report()` and associated code
* move `handler.spec` and `parser.spec` in more apprpriate module spec files.

## Testing

CI

---


packages/core/src/domain/tracekit/index.ts
```diff
@@ -1,3 +1,3 @@
 export { Handler, StackTrace, BrowserError } from './types'
 export { computeStackTrace } from './computeStackTrace'
-export { report, subscribe, unsubscribe } from './report'
+export { subscribe, unsubscribe } from './report'
```
bcaudan, line 3: `report` may not be the right name for the file anymore then.
`tracekit` instead?

packages/core/test/specHelper.ts
```diff
@@ -312,3 +312,18 @@ export function setUserAgent(ua: string) {
...
+ * information.
+ */
+export function disableJasmineUncaughtErrorHandler() {
+  const originalErrorHandler = window.onerror
+  window.onerror = null
+  return {
+    reset: () => {
+      window.onerror = originalErrorHandler
+    },
+  }
```
bcaudan, line 328: what about using `instrumentMethod`?

PR #1140, opened 2021-10-15
⚗ Investigate single spa integration

amortemousque:
## Motivation

We’d like to know if our strategy for supporting sub apps could be compatible with this framework. 

## Changes

Add setService to RUM public API

## Testing

Locally

---


bcaudan:
@amortemousque can we close that?

PR #1148, opened 2021-10-29
✨ [RUMF-1067] forward rum event to bridge

amortemousque:
## Motivation

This PR forward RUM events when the bridge is present.

## Changes

Forward RUM event if bridge present

## Testing

Local, Unit

---


packages/rum-core/src/boot/startRum.ts
```diff
@@ -107,7 +115,7 @@ export function startRumEventCollection(
     stop: () => {
       // prevent batch from previous tests to keep running and send unwanted requests
       // could be replaced by stopping all the component when they will all have a stop method
-      batch.stop()
+      stopBatch()
```
BenoitZugmeyer, line 118: `stopBatch` is undefined if the event bridge is used

packages/rum-core/src/boot/startRum.ts
```diff
@@ -96,7 +97,14 @@ export function startRumEventCollection(
   const parentContexts = startParentContexts(lifeCycle, session)
   const urlContexts = startUrlContexts(lifeCycle, locationChangeObservable, location)
   const foregroundContexts = startForegroundContexts()
-  const batch = startRumBatch(configuration, lifeCycle)
+
+  let stopBatch: () => void
+
+  if (isEventBridgePresent()) {
+    startRumEventBridge(lifeCycle)
```
BenoitZugmeyer, line 104: NIT: maybe expose a stop function to unsubscribe from the lifecycle

packages/core/src/transport/eventBridge.ts
```diff
@@ -8,10 +8,22 @@ export interface DatadogEventBridge {
...
+  }
+}
+
 export function isEventBridgePresent(): boolean {
-  return !!getEventBridge()
+  return !!getDatadogEventBridge()
 }
 
-function getEventBridge() {
+function getDatadogEventBridge() {
```
BenoitZugmeyer, line 27: NIT: maybe `getEventBridgeGlobal` would be more explicit to avoid naming clash with `getEventBridge`

packages/rum-core/src/boot/startRum.ts
amortemousque: It could simplify the code, but it's a bit weird to me since we never unsubscribe from lifecycle. Should we start doing it to make RUM stoppable?

BenoitZugmeyer: This should be discussed further, so you may ignore my comment for this PR, no problem :)

amortemousque: okay :)

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -33,3 +33,28 @@ describe('isEventBridgePresent', () => {
...
+
+  it('should serialize the event for the datadog bridge', () => {
+    const sendSpy = spyOn(eventBridgeStub, 'send')
+    const bridge = getEventBridge()
+    const event = { foo: 'bar' }
+    const eventType = 'view'
+
+    bridge.send(eventType, event)
+
+    expect(sendSpy).toHaveBeenCalledOnceWith(JSON.stringify({ eventType, event }))
```
bcaudan, line 58: what about expecting an actual string value instead of rewriting the implementation in the test?

packages/core/test/specHelper.ts
```diff
@@ -315,7 +315,9 @@ export function restoreUserAgent() {
 }
 
 export function initDatadogEventBridgeStub() {
```
bcaudan, line 317:
```suggestion
export function initEventBridgeStub() {
```

packages/core/test/specHelper.ts
```diff
@@ -315,7 +315,9 @@ export function restoreUserAgent() {
...
-  ;(window as BrowserWindow).DatadogEventBridge = { send: () => undefined }
+  const bridgeStub = { send: () => undefined }
```
bcaudan, line 318:
```suggestion
  const eventBridgeStub = { send: () => undefined }
```

packages/core/test/specHelper.ts
```diff
@@ -315,7 +315,9 @@ export function restoreUserAgent() {
...
+  ;(window as BrowserWindow).DatadogEventBridge = bridgeStub
+  return bridgeStub
 }
 
 export function deleteDatadogEventBridgeStub() {
```
bcaudan, line 323:
```suggestion
export function deleteEventBridgeStub() {
```

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -33,3 +33,28 @@ describe('isEventBridgePresent', () => {
     })
   })
 })
+
+describe('getEventBridge', () => {
+  let eventBridgeStub: DatadogEventBridge
```
bcaudan, line 38:
```suggestion
  let eventBridgeStub: EventBridge
```

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -33,3 +33,28 @@ describe('isEventBridgePresent', () => {
...
+    updateExperimentalFeatures(['event-bridge'])
+    eventBridgeStub = initDatadogEventBridgeStub()
+  })
+
+  afterEach(() => {
+    resetExperimentalFeatures()
+    deleteDatadogEventBridgeStub()
+  })
+
+  it('should serialize the event for the datadog bridge', () => {
```
bcaudan, line 50: instead of: 
`getEventBridge should serialize the event for the datadog bridge`
what about:
`event bridge should serialize sent events`

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -33,3 +33,28 @@ describe('isEventBridgePresent', () => {
...
+    const sendSpy = spyOn(eventBridgeStub, 'send')
+    const bridge = getEventBridge()
```
bcaudan, line 52:
```suggestion
    const eventBridge = getEventBridge()
```

packages/rum-core/src/boot/startRum.ts
```diff
@@ -96,7 +97,14 @@ export function startRumEventCollection(
...
+  } else {
```
bcaudan, line 105: could we have a unit test for it?

test/e2e/lib/framework/createTest.ts
```diff
@@ -28,6 +28,7 @@ interface TestContext {
   baseUrl: string
   crossOriginUrl: string
   events: EventRegistry
+  bridgeEvents: EventRegistry
```
bcaudan, line 31: what about renaming `events` to `serverEvents`?

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -141,6 +155,19 @@ export function html(parts: readonly string[], ...vars: string[]) {
   return parts.reduce((full, part, index) => full + vars[index - 1] + part)
 }
 
+function setupBridge() {
```
bcaudan, line 158:
```suggestion
function setupEventBridge() {
```

test/e2e/lib/helpers/sdk.ts
```diff
@@ -1,7 +1,8 @@
-import { getTestServers, waitForServersIdle } from '../framework'
-import { browserExecuteAsync, deleteAllCookies } from './browser'
+import { RumEvent } from '@datadog/browser-rum'
+import { EventRegistry, getTestServers, waitForServersIdle } from '../framework'
+import { browserExecute, browserExecuteAsync, deleteAllCookies } from './browser'
 
-export async function flushEvents() {
+export async function flushEvents(bridgeEvents?: EventRegistry) {
```
bcaudan, line 5: Do we need all the logic about servers for bridge events?
if not, we could introduce a dedicated flush method instead of mixing the concerns

test/e2e/lib/helpers/sdk.ts
```diff
@@ -11,6 +12,18 @@ export async function flushEvents() {
...
+  // Flush bridge events
+  if (bridgeEvents) {
+    const events = (await browserExecute(() => {
+      const eventBridge = (window as any).DatadogEventBridge
+      if (!eventBridge) {
+        return []
+      }
+      return eventBridge.events as RumEvent[]
+    })) as Array<{ eventType: string; event: RumEvent }>
+    events.forEach(({ eventType, event }) => bridgeEvents.push(eventType === 'log' ? 'logs' : 'rum', event))
```
bcaudan, line 24: do we need to differentiate monitoring logs from logger logs?

test/e2e/scenario/rum/bridge.scenario.ts
```diff
@@ -0,0 +1,63 @@
+import { browserExecute, flushBrowserLogs } from '../../lib/helpers/browser'
```
bcaudan, line 1: the file name should be `eventBridge.scenario.ts`

test/e2e/scenario/rum/bridge.scenario.ts
```diff
@@ -0,0 +1,63 @@
...
+import { createTest, html } from '../../lib/framework'
+import { flushEvents } from '../../lib/helpers/sdk'
+
+describe('bridge present', () => {
+  createTest('forward action to the bridge')
```
bcaudan, line 6:
```suggestion
describe('rum event bridge', () => {
  createTest('send action')
```

test/e2e/lib/helpers/sdk.ts
amortemousque: For now the bridge does not support log or monitoring so I'll remove the check

test/e2e/lib/helpers/sdk.ts
amortemousque: The flush events does not focus on server or bridge events. It basically wait 200ms and wait for idle to be sure that the SDK send all events before testing. Flush event also does a page redirection to handle session replay event. I can't get bridge event after the redirection which is why I put the code inside the flush event.

bcaudan: IMO, we probably don't need to wait for server idle since we don't send events to servers, but it should not hurt.

test/e2e/lib/helpers/sdk.ts
```diff
@@ -11,6 +12,18 @@ export async function flushEvents() {
...
+  // Flush bridge events
+  if (bridgeEvents) {
+    const events = (await browserExecute(() => {
+      const eventBridge = (window as any).DatadogEventBridge
+      if (!eventBridge) {
+        return []
+      }
+      return eventBridge.events as RumEvent[]
+    })) as Array<{ eventType: string; event: RumEvent }>
+    events.forEach(({ event }) => bridgeEvents.push('rum', event))
```
bcaudan, line 24: what about extracting a helper for that?

packages/core/src/transport/eventBridge.ts
```diff
@@ -8,10 +8,22 @@ export interface DatadogEventBridge {
   send(msg: string): void
 }
 
+export type BridgeEventType = 'log' | 'view' | 'error' | 'action' | 'resource' | 'long_task'
```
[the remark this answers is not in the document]
amortemousque: I'm wondering if it should be generic. To me it represents the contract between us and mobile. If a new type if added to RumEvent.type do we want to be warn by typescript that it does match BridgeEventType or not ?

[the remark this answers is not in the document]
amortemousque: I made it generic

test/e2e/lib/framework/serverApps/intake.ts
```diff
@@ -2,17 +2,22 @@ import { createInflate } from 'zlib'
...
-import { SegmentFile, SerssionReplayCall } from '../../types/serverEvents'
+import cors from 'cors'
+import { SegmentFile, SessionReplayCall } from '../../types/serverEvents'
 import { EventRegistry } from '../eventsRegistry'
 
-export function createIntakeServerApp(events: EventRegistry) {
+export function createIntakeServerApp(serverEvents: EventRegistry, bridgeEvents: EventRegistry) {
   const app = express()
 
+  app.use(cors())
```
amortemousque, line 12: Weirdly, using `XMLHttpRequest` in the e2e bridge makes CORS errors but not sendBeacon. However to have an easy implementation and to be compatible with IE, I stick with XMLHttpRequest and added `cors()` to the e2e intake server.
In any case this configuration is closer to real life scenario.

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -11,26 +11,26 @@ export interface SetupOptions {
...
+export type PageFactory = (options: PageSetupOptions, intakeUrl: string) => string
 
 const isBrowserStack =
   browser.config.services &&
   browser.config.services.some((service) => (Array.isArray(service) ? service[0] : service) === 'browserstack')
 
-export const DEFAULT_SETUPS = isBrowserStack
-  ? [{ name: 'bundle', factory: bundleSetup }]
+export const DEFAULT_PAGE_SETUPS = isBrowserStack
+  ? [{ name: 'bundle', pageFactory: bundleSetup }]
```
bcaudan, line 21: we had `setup`, now we have `page`, `setup` and `pageSetup`.
What are the differences?

test/e2e/lib/framework/serverApps/intake.ts
bcaudan: wrong header with xhr maybe?
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests

test/e2e/lib/framework/pageSetups.ts
amortemousque: I renamed `setup` to `pageSetup` to match the file name, which I found clearer.
I renamed `factory` by `pageFactory` mainly because at some point in the code, it's passed as parameter and we loose the context. I could also change the code to pass the pageSetup and keep the context.
Wdyt?

test/e2e/lib/framework/serverApps/intake.ts
amortemousque: Yes, probably. However, I used the same implementation as the SDK request fallback which doesn't pass headers. Wdyt, should I change the e2e xhr bridge implementation or are we good to go with `cors()`?

test/e2e/lib/framework/pageSetups.ts
amortemousque: Topic for another time. I roll back.

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -47,10 +47,16 @@ export function computeTransportConfiguration(
...
+
+    const e2eEndpointBuilders = {
+      logsEndpointBuilder: e2eEndpointBuilder('<<< E2E LOGS ENDPOINT >>>'),
+      rumEndpointBuilder: e2eEndpointBuilder('<<< E2E RUM ENDPOINT >>>'),
+      sessionReplayEndpointBuilder: e2eEndpointBuilder('<<< E2E SESSION REPLAY ENDPOINT >>>'),
+      internalMonitoringEndpointBuilder: e2eEndpointBuilder('<<< E2E INTERNAL MONITORING ENDPOINT >>>'),
+    }
+
+    intakeEndpoints = objectValues(e2eEndpointBuilders).map((builder) => builder.build())
+    objectEntries(e2eEndpointBuilders).forEach(([key, builder]) => (configuration[key] = builder))
```
bcaudan, line 59: industrializing the override of the configuration variable that we have just created does not seem right to me.
Could we have something more natural like:
```
// compute endpointBuilders and intakeEndpoints depending on build mode
// compute replica configuration and add intakes
// create configuration with correct values
```

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -47,10 +47,16 @@ export function computeTransportConfiguration(
...
-    configuration.sessionReplayEndpointBuilder = e2eEndpointBuilder('<<< E2E SESSION REPLAY ENDPOINT >>>')
+
+    const e2eEndpointBuilders = {
+      logsEndpointBuilder: e2eEndpointBuilder('<<< E2E LOGS ENDPOINT >>>'),
+      rumEndpointBuilder: e2eEndpointBuilder('<<< E2E RUM ENDPOINT >>>'),
+      sessionReplayEndpointBuilder: e2eEndpointBuilder('<<< E2E SESSION REPLAY ENDPOINT >>>'),
+      internalMonitoringEndpointBuilder: e2eEndpointBuilder('<<< E2E INTERNAL MONITORING ENDPOINT >>>'),
+    }
+
+    intakeEndpoints = objectValues(e2eEndpointBuilders).map((builder) => builder.build())
```
bcaudan, line 58: what about faking `buildIntakeUrl` too?
it would avoid to have something specific to do for e2e

test/e2e/lib/framework/serverApps/intake.ts
bcaudan: IMO `cors()` is fine

packages/core/src/domain/configuration/transportConfiguration.ts
amortemousque: True, it could be better. I'm gonna try it

packages/core/src/domain/configuration/transportConfiguration.ts
amortemousque: sure

packages/core/src/tools/utils.ts
```diff
@@ -244,6 +244,10 @@ export function mapValues<A, B>(object: { [key: string]: A }, fn: (arg: A) => B)
   return newObject
 }
 
+export function notEmpty<T>(value: T | null | undefined): value is T {
```
bcaudan, line 247: Empty seems more suited to a collection type, from you implementation [nullish](https://developer.mozilla.org/en-US/docs/Glossary/Nullish) could be a more suited term

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -24,62 +24,80 @@ export function computeTransportConfiguration(
...
 ): TransportConfiguration {
-  const endpointBuilders = {
-    logsEndpointBuilder: createEndpointBuilder(initConfiguration, buildEnv, 'logs'),
-    rumEndpointBuilder: createEndpointBuilder(initConfiguration, buildEnv, 'rum'),
-    sessionReplayEndpointBuilder: createEndpointBuilder(initConfiguration, buildEnv, 'sessionReplay'),
-  }
-  const intakeEndpoints: string[] = objectValues(endpointBuilders).map((builder) => builder.buildIntakeUrl())
+  const endpointBuilders = computeEndpointBuilders(initConfiguration, buildEnv)
+  const intakeEndpoints = objectValues(endpointBuilders)
+    .filter(notEmpty)
```
bcaudan, line 29: why do we need to check for null here?
could we even not do this check by not providing undefined value?

amortemousque: It's a tradeoff I made, to have a clearer `computeEndpointBuilders` but let's do it differently then.

packages/core/src/tools/utils.ts
```diff
@@ -228,8 +228,8 @@ export function objectHasValue<T extends { [key: string]: unknown }>(object: T,
   return Object.keys(object).some((key) => object[key] === value)
 }
 
-export function objectEntries(object: { [key: string]: unknown }): Array<[string, unknown]> {
-  return Object.keys(object).map((key) => [key, object[key]])
+export function objectEntries<T = unknown>(object: T): Array<[keyof T, T[keyof T]]> {
+  return (Object.keys(object) as Array<keyof T>).map((key) => [key, object[key]])
```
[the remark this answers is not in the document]
amortemousque: Sure, even though having a typed objectEntries seems better to me.

bcaudan (approved):

PR #542, opened 2020-09-23
[RUM] 🐛 call previously installed onunhandledrejection handler

michaelmcchen:
## Motivation

https://github.com/DataDog/browser-sdk/issues/541 confirmed that the existing behavior of ignoring the previously installed `onunhandledrejection` handler is a bug; this PR fixes it

## Testing

I tested this out against my project (that first raised the issue) as well as added an unit test to `tracekit.spec.ts`.

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/master/CONTRIBUTING.md) documentation.

PR #1151, opened 2021-11-04
[RUMF-1046] make Tracekit stoppable

### LINKED ISSUE #541: SDK override existing installed `onunhandledrejection` handler
Hi, 

I have a project where I've installed both Datadog RUM and Sentry (https://sentry.io) and when Sentry is initialized _before_ RUM, RUM will override the `onunhandledrejection` handler (see the [code](https://github.com/DataDog/browser-sdk/blob/8f86472688dbb3cf117ad422c11cd5ba7fa5aab8/packages/core/src/tracekit.ts#L312-L331)) 

The fix for this project was to initialize Datadog before Sentry, so Sentry calls both its handler & the previously installed handler (RUM). I'm wondering if the RUM team will be receptive of a PR changing the installation behavior to respect previously installed handlers. Thanks!

BenoitZugmeyer:
## Motivation

In the same spirit as #1133 and #1124 , this PR allows to stop/unpatch the tracekit instrumentation.

By using the `instrumentMethod` helper, it also fixes #541 (superseds #542)

## Changes

* Simplify tracekit tests a bit to make them synchronous and factorize the subscribe/unsubscribe in before/afterEach hooks
* Make Tracekit stoppable by using `instrumentMethod`
* Add a non-regression test for #541 collocated to the similar test for `onerror`

## Testing

Unit, manual, staging

---

### DIFF
--- a/packages/core/src/domain/tracekit/tracekit.spec.ts
+++ b/packages/core/src/domain/tracekit/tracekit.spec.ts
@@ -1,21 +1,22 @@
 import { disableJasmineUncaughtErrorHandler } from '../../../test/specHelper'
-import { subscribe, unsubscribe, traceKitWindowOnError } from './tracekit'
-import { Handler } from './types'
+import { startUnhandledErrorCollection } from './tracekit'
+import { Callback } from './types'
 
-describe('traceKitWindowOnError', () => {
+describe('startUnhandledErrorCollection', () => {
   const testLineNo = 1337
 
-  let subscriptionHandler: jasmine.Spy<Handler>
+  let callbackSpy: jasmine.Spy<Callback>
+  let stopCollectingUnhandledError: () => void
   let resetJasmineUncaughtErrorHandler: () => void
 
   beforeEach(() => {
     ;({ reset: resetJasmineUncaughtErrorHandler } = disableJasmineUncaughtErrorHandler())
-    subscriptionHandler = jasmine.createSpy()
-    subscribe(subscriptionHandler)
+    callbackSpy = jasmine.createSpy()
+    ;({ stop: stopCollectingUnhandledError } = startUnhandledErrorCollection(callbackSpy))
   })
 
   afterEach(() => {
-    unsubscribe(subscriptionHandler)
+    stopCollectingUnhandledError()
     resetJasmineUncaughtErrorHandler()
   })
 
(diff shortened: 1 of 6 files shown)


packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -1,225 +1,104 @@
...
-    } catch (inner) {
-      exception = inner
-    }
-  })
-  if (exception) {
-    throw exception
+  return {
+    stop: () => {
+      stopInstrumentingOnError()
+      stopInstrumentingOnUnhandledRejection
```
bcaudan, line 48:
```suggestion
      stopInstrumentingOnUnhandledRejection()
```

packages/core/src/domain/tracekit/tracekit.spec.ts
```diff
@@ -1,165 +1,110 @@
...
-        expect(stack.name).toEqual('ReferenceError')
-        expect(stack.message).toEqual('foo is undefined')
-        unsubscribe(subscriptionHandler!)
-        done()
-      }
-      subscribe(subscriptionHandler)
-      traceKitWindowOnError('ReferenceError: foo is undefined', 'http://example.com', testLineNo)
+    it('should separate name, message for default error types (e.g. ReferenceError)', () => {
+      window.onerror!('ReferenceError: foo is undefined', 'http://example.com', testLineNo)
+      const [stack, ,] = callbackSpy.calls.mostRecent().args
```
bcaudan, line 37:
```suggestion
      const [stack] = callbackSpy.calls.mostRecent().args
```
or do we need it?
(other occurrences with or without)

packages/core/src/domain/tracekit/tracekit.ts
BenoitZugmeyer: How did I miss that!

packages/core/src/domain/tracekit/tracekit.spec.ts
BenoitZugmeyer: Unneeded, I'll remove those.

bcaudan (approved):
Nice!

packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -1,225 +1,104 @@
...
-      handlers.splice(i, 1)
-    }
-  }
-
-  if (handlers.length === 0) {
-    uninstallGlobalHandler()
-    uninstallGlobalUnhandledRejectionHandler()
-  }
-}
+export function startUnhandledErrorCollection(callback: UnhandledErrorCallback) {
```
amortemousque, line 41: Like it!

amortemousque (approved):

PR #1155, opened 2021-11-08
✨ [RUMF-1070] forward logs event to bridge

amortemousque:
## Motivation

This PR forward Logs events when the bridge is present.

## Changes

Forward Logs event if bridge present

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/logs/src/transport/startLoggerBatch.ts
```diff
@@ -0,0 +1,27 @@
...
+    return new Batch(
+      new HttpRequest(endpointBuilder, configuration.batchBytesLimit),
+      configuration.maxBatchSize,
+      configuration.batchBytesLimit,
+      configuration.maxMessageSize,
+      configuration.flushTimeout
+    )
+  }
+
+  return (message: Context) => {
```
BenoitZugmeyer, line 21: NIT: what about keeping an object with an `add` method, and do the same for returned value of `startLoggerEventBridge`? It could be a bit more explicit.

BenoitZugmeyer (approved):

packages/logs/src/boot/startLogs.ts
```diff
@@ -55,7 +58,7 @@ export function doStartLogs(
   )
 
   const assemble = buildAssemble(session, configuration, reportError)
-  const batch = startLoggerBatch(configuration)
+  const transport = isEventBridgePresent() ? startLoggerEventBridge() : startLoggerBatch(configuration)
```
bcaudan, line 61: not super fan of the `transport` naming.
what about something a bit like
```suggestion
let onLogEventCollected 
if (isEventBridgePresent()) {
  const bridge = getEventBridge()
  onLogCollected = (message: Context) => bridge.send('log', message) 
} else {
  const batch = startLoggerBatch(configuration)
  onLogCollected = (message: Context) => batch.add(message) 
}
```

amortemousque: To m, It feels a bit convoluted to use and event handler for this case. We loose the intent when reading 
https://github.com/DataDog/browser-sdk/blob/19b1552111f4afe7e5e73fcadc1aa140d8d3b1a0/packages/logs/src/boot/startLogs.ts#L90-L95

packages/logs/src/transport/startLoggerBatch.ts
amortemousque: So you prefer` transport.add(message)` instead of `transport(message)`. I suggest you look at the conversation with Bastien: https://github.com/DataDog/browser-sdk/pull/1155#discussion_r745408409

packages/logs/src/boot/startLogs.ts
bcaudan: >  It feels a bit convoluted to use and event handler 

it is still the same callback that transport but we keep the concept of batch.add and bridge.send that is lost in startLogs with your current version.

> We loose the intent

which intent?

amortemousque: send the message to whatever process it. Also, worth mentioning that the transport "concept" already exist. `TransportConfiguration`, `transport/` (folder).

BenoitZugmeyer: I feel like `transport` is not an action on its own. Maybe if we had `transport.send()` or `transport.add()` it would be nicer. Or a function named `send` instead of `transport`?

bcaudan (approved):

BenoitZugmeyer (approved):

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -150,6 +150,25 @@ describe('logs', () => {
...
+      updateExperimentalFeatures(['event-bridge'])
+      const sendSpy = spyOn(initEventBridgeStub(), 'send')
+
+      let configuration = { ...baseConfiguration, ...{ sampleRate: 0 } } as LogsInitConfiguration
+      let sendLog = originalStartLogs(configuration, new Logger(noop))
+      sendLog(DEFAULT_MESSAGE, {})
+
+      expect(sendSpy).not.toHaveBeenCalled()
+
+      configuration = { ...baseConfiguration, ...{ sampleRate: 100 } } as LogsInitConfiguration
```
amortemousque, line 164: `baseConfiguration` already has sampleRate to 100 but I found it more expressive to assign it here.

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -150,6 +150,25 @@ describe('logs', () => {
     })
   })
 
+  describe('sampling', () => {
+    it('should be sampled when event bridge is present', () => {
```
bcaudan, line 154: what do you mean by `sampling should be sampled`?

packages/logs/src/boot/startLogs.ts
```diff
@@ -37,9 +37,8 @@ export function startLogs(initConfiguration: LogsInitConfiguration, errorLogger:
     trackNetworkError(configuration, errorObservable)
   }
 
-  const session = canUseEventBridge()
-    ? startStubLoggerSession()
-    : startLoggerSession(configuration, areCookiesAuthorized(configuration.cookieOptions))
+  const canUseCookies = areCookiesAuthorized(configuration.cookieOptions) && !canUseEventBridge()
```
bcaudan, line 40: what is the relation between cookies and bridge?

amortemousque: Created a startLoggerSessionStub

packages/logs/src/domain/loggerSession.ts
```diff
@@ -12,14 +12,7 @@ export enum LoggerTrackingType {
   TRACKED = '1',
 }
 
-export function startLoggerSession(configuration: Configuration, areCookieAuthorized: boolean): LoggerSession {
-  if (!areCookieAuthorized) {
```
BenoitZugmeyer, line 16: We probably want to keep this

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -127,6 +137,36 @@ describe('logs', () => {
...
+      const sendLog = startLogs()
+      sendLog(DEFAULT_MESSAGE, {})
+
+      expect(server.requests.length).toEqual(0)
+      expect(sendSpy).toHaveBeenCalled()
+    })
+  })
+
+  describe('sampling', () => {
+    it('should be applied when event bridge is present', () => {
```
BenoitZugmeyer, line 154: I thought we agree to not apply the sampling on logs for now?

bcaudan: we agreed to recheck it with product and finally went with sampling at each init like we do in other env without session/cookie.
Any concern?

bcaudan (approved):

packages/logs/src/boot/startLogs.spec.ts
BenoitZugmeyer: Alright, I find this a bit weird product wise but other than that it's fine

packages/logs/src/domain/loggerSession.ts
amortemousque: The check is done in [startLogs](https://github.com/DataDog/browser-sdk/blob/c0356a2b8fb852dba1d40689fbd09217011bf650/packages/logs/src/boot/startLogs.ts#L40)

BenoitZugmeyer: Ah, I missed it. Nice!

BenoitZugmeyer (approved):

PR #1157, opened 2021-11-09
✨ [RUMF-1074] bridge host checking

amortemousque:
## Motivation

The mobile SDKs (Android and iOS) want to track events that happen inside a webview using the browser SDK. To do so the browser SDK will send events to a bridge injected by the mobile SDK. 

Because WebView can be used to display websites not owned by the customer, we need to guarantee that if these websites also have the browser SDK, they won’t send their events to the mobile SDK. 

## Changes

Add webView host checking

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/core/src/transport/eventBridge.spec.ts
+++ b/packages/core/src/transport/eventBridge.spec.ts
@@ -3,7 +3,7 @@ import { deleteEventBridgeStub, initEventBridgeStub } from '../../test/specHelpe
 import { getEventBridge, isEventBridgePresent } from './eventBridge'
 
 describe('isEventBridgePresent', () => {
-  const notAllowedWebViewHosts = ['foo.bar']
+  const allowedWebViewHosts = ['foo.bar']
 
   afterEach(() => {
     resetExperimentalFeatures()
@@ -25,7 +25,7 @@ describe('isEventBridgePresent', () => {
     })
 
     it('should not detect when the bridge is present and the webView host is not allowed', () => {
-      initEventBridgeStub(notAllowedWebViewHosts)
+      initEventBridgeStub(allowedWebViewHosts)
       expect(isEventBridgePresent()).toBeFalse()
     })
   })
@@ -37,7 +37,7 @@ describe('isEventBridgePresent', () => {
     })
 
     it('should not detect when the bridge is present and the webView host is not allowed', () => {
-      initEventBridgeStub(notAllowedWebViewHosts)
+      initEventBridgeStub(allowedWebViewHosts)
       expect(isEventBridgePresent()).toBeFalse()
     })
 
--- a/packages/core/src/transport/eventBridge.spec.ts
+++ b/packages/core/src/transport/eventBridge.spec.ts
@@ -62,7 +62,7 @@ describe('getEventBridge', () => {
   })
 
   it('event bridge should serialize sent events', () => {
-    const eventBridge = getEventBridge()
+    const eventBridge = getEventBridge()!
 
     eventBridge.send('view', { foo: 'bar' })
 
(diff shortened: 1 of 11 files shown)


packages/core/src/transport/eventBridge.ts
```diff
@@ -1,25 +1,35 @@
...
+      eventBridgeGlobal.send(JSON.stringify({ eventType, event }))
     },
   }
 }
 
 export function isEventBridgePresent(): boolean {
-  return !!getEventBridgeGlobal()
+  return getEventBridge()
+    .getAllowedWebViewHosts()
+    .some((o) => includes(o, window.location.hostname))
```
bcaudan, line 32:
```suggestion
    .some((host) => includes(host, window.location.hostname))
```

bcaudan: why `includes` and not `===`? 
should we add a test for it?

amortemousque: It was to handle port but know that I see it, we might have some issue with the include. I've double checked with Marius and an equality will do.

amortemousque: For the test it's done within the `isEventBridgePresent`.

packages/core/src/transport/eventBridge.ts
```diff
@@ -5,21 +5,31 @@ export interface BrowserWindow extends Window {
...
+      eventBridgeGlobal.send(JSON.stringify({ eventType, event }))
     },
   }
 }
 
 export function isEventBridgePresent(): boolean {
-  return !!getEventBridgeGlobal()
+  return getEventBridge()
+    .getAllowedWebViewHosts()
+    .some((host) => host === window.location.hostname)
```
BenoitZugmeyer, line 32:
```suggestion
  return includes(window.location.hostname, getEventBridge().getAllowedWebViewHosts())
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -5,21 +5,31 @@ export interface BrowserWindow extends Window {
...
   send(msg: string): void
 }
 
 export function getEventBridge<T, E>() {
-  const datadogEventBridge = getEventBridgeGlobal()
+  const eventBridgeGlobal = getEventBridgeGlobal()
+
+  if (!eventBridgeGlobal) {
+    return { getAllowedWebViewHosts: () => [], send: () => undefined }
+  }
```
BenoitZugmeyer, line 17: Do we need to return a fake event bridge here? Why not returning `undefined`?

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -3,6 +3,8 @@ import { deleteEventBridgeStub, initEventBridgeStub } from '../../test/specHelpe
 import { getEventBridge, isEventBridgePresent } from './eventBridge'
 
 describe('isEventBridgePresent', () => {
+  const notAllowedWebViewHosts = ['foo.bar']
```
BenoitZugmeyer, line 6: Shouldn't it be
```suggestion
  const allowedWebViewHosts = ['foo.bar']
```
?

packages/core/src/transport/eventBridge.ts
amortemousque: Sure it can work.

packages/core/src/transport/eventBridge.ts
amortemousque: nice!

packages/core/src/transport/eventBridge.ts
```diff
@@ -1,25 +1,34 @@
...
-      datadogEventBridge?.send(JSON.stringify({ eventType, event }))
+      eventBridgeGlobal.send(JSON.stringify({ eventType, event }))
     },
   }
 }
 
 export function isEventBridgePresent(): boolean {
-  return !!getEventBridgeGlobal()
+  const bridge = getEventBridge()
+  return !!bridge && includes(bridge.getAllowedWebViewHosts(), window.location.hostname)
```
bcaudan, line 31: we don't check only if the bridge is present now.
what about renaming the function to something like `canUseEventBridge`?

bcaudan (approved):

packages/core/src/transport/eventBridge.ts
amortemousque: yes sure

PR #1158, opened 2021-11-10
🔊 [RUMF-1080] try to detect segment sent without start

bcaudan:
## Motivation

Start field missing during the parsing of some segments

## Changes

Add a monitoring log when segment meta start is not a number

## Testing

- [ ] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/types.ts
+++ b/packages/rum/src/types.ts
@@ -12,7 +12,6 @@ export interface SegmentMeta extends SegmentContext {
   has_full_snapshot: boolean
   records_count: number
   creation_reason: CreationReason
-  [key: string]: any
 }
 
 export interface SegmentContext {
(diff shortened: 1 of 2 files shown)


packages/rum/src/types.ts
```diff
@@ -12,6 +12,7 @@ export interface SegmentMeta extends SegmentContext {
   has_full_snapshot: boolean
   records_count: number
   creation_reason: CreationReason
+  [key: string]: any
```
BenoitZugmeyer, line 15: I'd prefer a `as Context` cast when using it in `addMonitoringMessage`, so we keep a clean type for the rest of the application

BenoitZugmeyer (approved):

PR #1162, opened 2021-11-12
[RUMF-1068] Forward browser_sdk_version to mobile

amortemousque:
## Motivation

The mobile SDKs (Android and iOS) want to track events that happen inside a webView using the browser SDK. The proposed way to do that is to inject a JS bridge interface that the browser SDK will call instead of sending the events to the intake.

To easily monitor browser sdk event sent though mobile, we need to keep the browser SDK version tag.
Because it is not easy for the mobile to keep browser tag, this first version send it as `_dd` attribute.

## Changes

Add `browser_sdk_version` in `_dd`

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/assembly.ts
```diff
@@ -91,6 +91,7 @@ export function startRumAssembly(
             session: {
               plan: session.hasReplayPlan() ? RumSessionPlan.REPLAY : RumSessionPlan.LITE,
             },
+            browser_sdk_version: configuration.sdkVersion,
```
amortemousque, line 94: I decided to set the value even if the bridge isn't present. It will allow us to check the same version attribute regardless of the webview context.

bcaudan: > It will allow us to check the same version attribute regardless of the webview context.

only from this version though, so we probably won't use it, no?

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -188,6 +188,7 @@ export interface RumContext {
     session: {
       plan: RumSessionPlan
     }
+    browser_sdk_version: string
```
bcaudan, line 191: should not we update rum-event-format too?

amortemousque: Indeed.

packages/rum-core/src/domain/assembly.ts
amortemousque: Yes, we will probably still use both the tag and the attribute to have the full picture. IYO, does it adds value to put it only on the webview context?

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -107,6 +108,7 @@ export function buildConfiguration(initConfiguration: InitConfiguration, buildEn
       initConfiguration.beforeSend && catchUserErrors(initConfiguration.beforeSend, 'beforeSend threw an error:'),
     cookieOptions: buildCookieOptions(initConfiguration),
     service: initConfiguration.service,
+    ...buildEnv,
```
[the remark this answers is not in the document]
amortemousque: yes, sure

BenoitZugmeyer (approved):
OK for me

packages/rum-core/src/domain/assembly.ts
bcaudan: it avoids to bloat events with unnecessary attributes

amortemousque: okay

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -107,6 +108,7 @@ export function buildConfiguration(initConfiguration: InitConfiguration, buildEn
       initConfiguration.beforeSend && catchUserErrors(initConfiguration.beforeSend, 'beforeSend threw an error:'),
     cookieOptions: buildCookieOptions(initConfiguration),
     service: initConfiguration.service,
+    sdkVersion: buildEnv.sdkVersion,
```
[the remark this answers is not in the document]
amortemousque: FMU, we prefer arguments instead of static import, like in "buildConfiguration". 
Instead of adding `buildEnv` all the way down, my approach is to merge it to the configuration. So we send it to `buildConfiguration()` and we get the configuration with everything we need for the rest of the code.
Wdyt?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -89,6 +90,7 @@ export function startRumAssembly(
             session: {
               plan: session.hasReplayPlan() ? RumSessionPlan.REPLAY : RumSessionPlan.LITE,
             },
+            browser_sdk_version: canUseEventBridge() ? configuration.sdkVersion : undefined,
```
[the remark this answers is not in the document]
amortemousque: sure

packages/rum-core/test/specHelper.ts
```diff
@@ -188,6 +188,7 @@ function validateRumEventFormat(rawRumEvent: RawRumEvent) {
       session: {
         plan: RumSessionPlan.REPLAY,
       },
+      browser_sdk_version: 'dev',
```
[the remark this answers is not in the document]
amortemousque: not required anymore

packages/core/src/domain/configuration/configuration.ts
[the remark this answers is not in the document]
bcaudan: FMU, buildEnv is a constant, we import constant in a lot of places.

bcaudan (approved):

PR #1166, opened 2021-11-15
♻️ [RUMF-1083] extract session store from session management

bcaudan:
## Motivation

First part of session management rework

Next:

- Replace session cookie cache
- Introduce session context history

## Changes

Extract session store from session management

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/core/src/domain/session/oldCookiesMigration.ts
+++ b/packages/core/src/domain/session/oldCookiesMigration.ts
@@ -1,4 +1,4 @@
-import { CookieCache, getCookie } from '../browser/cookie'
+import { CookieCache, getCookie } from '../../browser/cookie'
 import { persistSession, SessionState } from './sessionManagement'
 
 export const OLD_SESSION_COOKIE_NAME = '_dd'
--- a/packages/core/src/domain/session/oldCookiesMigration.ts
+++ b/packages/core/src/domain/session/oldCookiesMigration.ts
@@ -1,5 +1,5 @@
 import { getCookie, CookieOptions, cacheCookieAccess } from '../../browser/cookie'
-import { persistSession, SessionState, SESSION_COOKIE_NAME } from './sessionManagement'
+import { persistSession, SessionState, SESSION_COOKIE_NAME } from './sessionStore'
 
 export const OLD_SESSION_COOKIE_NAME = '_dd'
 export const OLD_RUM_COOKIE_NAME = '_dd_r'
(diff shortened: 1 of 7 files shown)


packages/core/src/domain/session/oldCookiesMigration.ts
```diff
@@ -13,7 +13,8 @@ export const LOGS_SESSION_KEY = 'logs'
  * This migration should remain in the codebase as long as older versions are available/live
  * to allow older sdk versions to be upgraded to newer versions without compatibility issues.
  */
-export function tryOldCookiesMigration(sessionCookie: CookieCache) {
+export function tryOldCookiesMigration(options: CookieOptions) {
+  const sessionCookie = cacheCookieAccess(SESSION_COOKIE_NAME, options)
```
BenoitZugmeyer, line 17: Since `sessionCookie` will be read only once, does it need to be cached?

bcaudan: No but we still need the logic of [persistSession](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/domain/sessionManagement.ts#L124) that take a cookieCache as input.
Since the plan is to remove cookie cache in the next PR, I was planning to deal with that refactoring then.

BenoitZugmeyer: Ok!

BenoitZugmeyer (approved):

amortemousque:
Not directly related to this PR, but it seems there is no need to export 
https://github.com/DataDog/browser-sdk/blob/eb2173c0fd1cef3aaf56a26b007225ecc89c48a3/packages/core/src/domain/session/sessionStore.ts#L79

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -1,39 +1,34 @@
...
   expire?: string
+
   [key: string]: string | undefined
 }
 
-/**
- * Limit access to cookie to avoid performance issues
- */
-export function startSessionManagement<TrackingType extends string>(
+export const SESSION_COOKIE_NAME = '_dd_s'
```
amortemousque, line 21: Can we group all global const together with
```
const SESSION_ENTRY_REGEXP = /^([a-z]+)=([a-z0-9-]+)$/
```

PR #1170, opened 2021-11-16
✨ [RUMF-1084] ignore init if a RUM instance is or will be injected by synthetics

BenoitZugmeyer:
## Motivation

Allows the Synthetics worker to prevent any RUM SDK initialisation if it plans to inject a RUM instance itself.

## Changes

* Move synthetics context in its own module
* Ignore `init()` call based on values (global variables/cookies) provided by the Synthetics worker

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -103,7 +104,7 @@ export function makeRumPublicApi<C extends RumInitConfiguration>(startRumImpl: S
       return
     }
 
-    if (!isValidInitConfiguration(initConfiguration)) {
+    if (willSyntheticsInjectRum() || !isValidInitConfiguration(initConfiguration)) {
```
amortemousque, line 107: By adding the check here, we still check `canHandleSession`. So we might have init cookie error message even if the SDK won't be started.

packages/rum-core/src/tools/syntheticsContext.ts
```diff
@@ -0,0 +1,30 @@
+import { getCookie } from '@datadog/browser-core'
```
amortemousque, line 1: `tools/syntheticsContext.ts`, should it be consider as a tool?

BenoitZugmeyer: Would you rather put it in `domain`? I thought it wasn't directly related to how RUM works, but happy to move it elsewhere.

packages/rum-core/src/boot/rumPublicApi.ts
BenoitZugmeyer: Good point, I'll move it

packages/rum-core/src/tools/syntheticsContext.ts
amortemousque: Why not to have tools, but I would also think of moving other components like `ContextHistory` which also looks like tools.

bcaudan: IMO
- syntheticsContext -> domain
- contextHistory -> tools

packages/rum-core/src/tools/syntheticsContext.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    expect(getSyntheticsContext()).toBeUndefined()
+  })
+})
+
+describe('willSyntheticsInjectRum', () => {
+  afterEach(() => {
+    cleanupSyntheticsWorkerValues()
+  })
+
+  it('returns false if nothing is defined"', () => {
```
bcaudan, line 51:
```suggestion
  it('returns false if nothing is defined', () => {
```

packages/rum-core/src/tools/syntheticsContext.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    mockSyntheticsWorkerValues({}, 'globals')
+
+    expect(willSyntheticsInjectRum()).toBeFalse()
+  })
+
+  it('returns false if the INJECTS_RUM global variable is false"', () => {
```
bcaudan, line 57:
```suggestion
  it('returns false if the INJECTS_RUM global variable is false', () => {
```

packages/rum-core/src/tools/syntheticsContext.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    mockSyntheticsWorkerValues({ injectsRum: false }, 'globals')
+
+    expect(willSyntheticsInjectRum()).toBeFalse()
+  })
+
+  it('returns true if the INJECTS_RUM global variable is truthy"', () => {
```
bcaudan, line 63:
```suggestion
  it('returns true if the INJECTS_RUM global variable is truthy', () => {
```

packages/rum-core/src/tools/syntheticsContext.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    mockSyntheticsWorkerValues({ injectsRum: true }, 'globals')
+
+    expect(willSyntheticsInjectRum()).toBeTrue()
+  })
+
+  it('returns true if the INJECTS_RUM cookie is truthy"', () => {
```
bcaudan, line 69:
```suggestion
  it('returns true if the INJECTS_RUM cookie is truthy', () => {
```

bcaudan (approved):

amortemousque (approved):

PR #1177, opened 2021-11-22
✨ [RUMF-1071] forward internal monitoring to bridge

amortemousque:
## Motivation

This PR forward internal monitoring events when the bridge is present.

## Changes

Forward internal monitoring event if bridge present

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/transport/httpRequest.ts
```diff
@@ -1,5 +1,5 @@
 import { EndpointBuilder } from '../domain/configuration/endpointBuilder'
-import { monitor, addErrorToMonitoringBatch, addMonitoringMessage } from '../domain/internalMonitoring'
+import { monitor, addErrorToMonitoring, addMonitoringMessage } from '../domain/internalMonitoring'
```
BenoitZugmeyer, line 2: It would be more natural to rename `addErrorToMonitoring` to `addMonitoringError`, to follow the `addMonitoringMessage` name.

packages/core/src/domain/internalMonitoring.ts
```diff
@@ -26,24 +26,29 @@ export interface MonitoringMessage extends Context {
...
 const monitoringConfiguration: {
-  batch?: Batch
   debugMode?: boolean
   maxMessagesPerPage: number
   sentMessageCount: number
 } = { maxMessagesPerPage: 0, sentMessageCount: 0 }
 
 export let externalContextProvider: () => Context
 
+let onInternalMonitoringEventCollected: ((message: MonitoringMessage) => void) | undefined
```
BenoitZugmeyer, line 36: Since this function takes a `MonitoringMessage` as argument, I think it should be called `onInternalMonitoringMessageCollected`

packages/core/src/domain/internalMonitoring.ts
```diff
@@ -32,7 +32,7 @@ const monitoringConfiguration: {
   sentMessageCount: number
 } = { maxMessagesPerPage: 0, sentMessageCount: 0 }
 
-let externalContextProvider: () => Context
+export let externalContextProvider: () => Context
```
BenoitZugmeyer, line 34: Mutable export is fun but a bit confusing :) I think you could either:
* Consolidate the Message with the Context in this module instead of in the batch, so `externalContextProvider` stays local to this module
* Pass the `externalContextProvider` to the batch by using an indirection, ex: `startBatch(() => externalContextProvider?.())`

packages/core/src/transport/httpRequest.ts
amortemousque: Yes, could be nice

packages/core/src/domain/internalMonitoring.ts
amortemousque: Right, and in fact I was also missing the external context when sending to the event bridge. 
It's now fixed. 
Does it look good to you?

packages/core/src/domain/internalMonitoring.ts
amortemousque: Hum, why not but we could say the same for log [`onLogEventCollected`](https://github.com/DataDog/browser-sdk/blob/c0356a2b8fb852dba1d40689fbd09217011bf650/packages/logs/src/boot/startLogs.ts#L63).
Should I update both?

test/e2e/scenario/rum/eventBridge.scenario.ts
```diff
@@ -60,4 +60,23 @@ describe('bridge present', () => {
       expect(serverEvents.rumViews.length).toEqual(0)
       expect(bridgeEvents.rumViews.length).toBeGreaterThan(0)
     })
+
+  createTest('forward internal monitoring to the bridge')
```
bcaudan, line 64: that is not RUM specific, should it be in internal monitoring scenario instead?

packages/core/src/domain/internalMonitoring.ts
BenoitZugmeyer: Looks good!

packages/core/src/domain/internalMonitoring/index.ts
```diff
@@ -0,0 +1,14 @@
...
+  monitor,
+  callMonitored,
+  addMonitoringMessage,
+  addMonitoringError,
+  startFakeInternalMonitoring,
+  resetInternalMonitoring,
+  setDebugMode,
+  startInternalMonitoring,
+} from './internalMonitoring'
+export { startMonitoringBatch } from './startMonitoringBatch'
```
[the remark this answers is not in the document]
BenoitZugmeyer: I would be in favor of this index file. It allows to separate the implementation into different sub-modules but keeping `imports` from other modules short and focused.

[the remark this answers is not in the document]
bcaudan: I am not against it but:
- should we start do that everywhere?
- do we need to explicitly list all the exported members or could we go with `*`?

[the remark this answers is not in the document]
BenoitZugmeyer: > should we start do that everywhere?

* If the folder is grouping modules that are mostly used together and expose a coherent API together accessed by other parts of the SDK. This can be seen as a privacy convention at the modules level.

* If the folder is grouping similar but mostly indepentent modules (for example, `src/domain`), I don't think it's worth it.

> do we need to explicitly list all the exported members or could we go with *?

The advantage of explicit exports is to chose what should be public and what should be kept private at the module level (ex: for spec files). Again, this is a convention, but it feels natural to me. 

IMO `startMonitoringBatch` should not be exported here, because it is only used by sibling modules.

packages/core/src/domain/internalMonitoring/internalMonitoring.ts
```diff
@@ -26,85 +26,61 @@ export interface MonitoringMessage extends Context {
...
-  batch?: Batch
   debugMode?: boolean
   maxMessagesPerPage: number
   sentMessageCount: number
 } = { maxMessagesPerPage: 0, sentMessageCount: 0 }
 
-let externalContextProvider: () => Context
+let onInternalMonitoringEventCollected: ((message: MonitoringMessage) => void) | undefined
 
 export function startInternalMonitoring(configuration: Configuration): InternalMonitoring {
```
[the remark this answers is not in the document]
amortemousque: I thought the same but there is also the instrumentation functions in this file.  I general, I like having the exported function name as the file name. The question comes when there is more exports. 
If I do it, I'd like we stick to this logic every time. Wdyt?

test/e2e/scenario/rum/eventBridge.scenario.ts
amortemousque: Indeed

packages/core/src/domain/internalMonitoring/index.ts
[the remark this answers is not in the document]
amortemousque: Yes could be nice to have a consistent approach here. I also, think that being able to import everything from everywhere is not great. Even if that's just a convention, I'd say let's discuss about it ;)

test/e2e/scenario/rum/eventBridge.scenario.ts
bcaudan: both approaches are fine to me

packages/core/src/domain/internalMonitoring.ts
BenoitZugmeyer: I think that you could, it's a small enough change to be included in this PR. But feel free to keep the current naming, this is more like nitpicking.

packages/core/src/domain/internalMonitoring/internalMonitoring.ts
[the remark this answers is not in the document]
bcaudan: indeed, nevermind then

bcaudan (approved):

BenoitZugmeyer (approved):

PR #1178, opened 2021-11-22
👷 extension improvements

bcaudan:
## Motivation

hackathon

## Changes

- move extension content from popup to devtools
- add tab system
- retrieve rum and logs init configuration

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/background/domain/endSession.ts
```diff
@@ -3,7 +3,6 @@ import { evaluateCodeInActiveTab } from '../utils'
 
 listenAction('endSession', () => {
   evaluateCodeInActiveTab(() => {
-    console.log('plop')
```
BenoitZugmeyer, line 6: 😆

developer-extension/src/background/store.ts
```diff
@@ -7,16 +7,32 @@ export const store: Store = {
...
+    store.local[tabId] = {
+      rumConfig: {},
+      logsConfig: {},
+    }
+  }
+  const localStore = store.local[tabId]
+  if (wouldModifyStore(newStore, localStore)) {
+    Object.assign(localStore, newStore)
+    sendAction('newStore', store)
+    chrome.storage.local.set({ store })
```
BenoitZugmeyer, line 32: I'm not that sure we should persist "local store", because tabId are likely to be changed/reused between launches. I think we should keep it in memory.

developer-extension/src/contentscript/domEventToActionMessage.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Receive an HTML event from the page
+ * and forward it as an action message to extension logic
+ */
+
+document.querySelector('html').addEventListener('extension', (event: any) => {
```
BenoitZugmeyer, line 6:
```suggestion
document.documentElement.addEventListener('extension', (event: any) => {
```

developer-extension/src/background/utils.ts
```diff
@@ -1,19 +1,19 @@
...
+function evaluateCodeInline(tabId: number, code: (arg?: string) => void, arg?: string) {
   chrome.tabs.executeScript(tabId, {
     code: `{
       const script = document.createElement('script')
       script.setAttribute("type", "module")
-      script.textContent = ${JSON.stringify(`(${String(code)})()`)}
+      script.textContent = ${JSON.stringify(`(${String(code)})('${arg}')`)}
       document.body.appendChild(script)
       script.remove()
     }`,
```
BenoitZugmeyer, line 19: To avoid the contentscript and custom event passing, we could imagine using the `executeScript` return value to pass the result of the evaluated function. Something like that:

```ts
function evaluateCodeInline<R>(tabId: number, code: (arg?: string) => R, arg?: string): Promise<R> {
  return chrome.tabs.executeScript(tabId, {
    code: `{
      const script = document.createElement('script')
      script.setAttribute("type", "module")
      script.textContent = ${JSON.stringify(`window.DD_EXTENSION_RESULT = (${String(code)})(${JSON.stringify(arg)})`)}
      document.body.appendChild(script)
      script.remove();
      const result = window.DD_EXTENSION_RESULT;
      delete window.DD_EXTENSION_RESULT;
      result
    }`,
  })
}
```

developer-extension/src/background/utils.ts
```diff
@@ -1,19 +1,19 @@
...
 }
 
-function evaluateCodeInline(tabId: number, code: () => void) {
+function evaluateCodeInline(tabId: number, code: (arg?: string) => void, arg?: string) {
   chrome.tabs.executeScript(tabId, {
     code: `{
       const script = document.createElement('script')
       script.setAttribute("type", "module")
-      script.textContent = ${JSON.stringify(`(${String(code)})()`)}
+      script.textContent = ${JSON.stringify(`(${String(code)})('${arg}')`)}
```
BenoitZugmeyer, line 16:
```suggestion
      script.textContent = ${JSON.stringify(`(${String(code)})(${JSON.stringify(arg)})`)}
```

developer-extension/src/panel/app.tsx
```diff
@@ -1,18 +1,11 @@
...
 const theme = {
   global: {
     fontSize: 14,
-    styles: {
-      base: css`
-        body {
-          width: 300px;
-        }
-      `,
-    },
```
BenoitZugmeyer, line 15: Could you keep a global style like that:
```ts
    styles: {
      base: css`
        body {
          min-height: 100vh;
        }
      `,
```
so the `body` background color is applied to the whole panel? This is mostly visible for dark themes:
Screenshot 2021-11-22 at 14 43 07

developer-extension/src/background/store.ts
bcaudan: For now, it is not an issue since even if it was stored in memory, the content stored in the tab is refreshed regularly.
Any reason to make the change now or could it wait for a future usage?

BenoitZugmeyer: Ok, makes sense, it can be done later :)

developer-extension/src/panel/app.tsx
bcaudan: This issue is actually fixed by newer version of bumbag but newer version of bumbag has type issues 😢 
I used the css trick.

BenoitZugmeyer: Ok, we'll upgrade bumbag someday.

developer-extension/src/background/domain/getConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+    sendActionAsDomEvent('configReceived', {
+      type,
+      config:
+        type === 'rum'
+          ? (window as BrowserWindow).DD_RUM?.getInitConfiguration()
+          : (window as BrowserWindow).DD_LOGS?.getInitConfiguration(),
+    })
+
+    function sendActionAsDomEvent(action: string, payload: any) {
+      document.querySelector('html').dispatchEvent(
```
amortemousque, line 25:
```suggestion
      document.documentElement.dispatchEvent(
```

developer-extension/src/common/types.ts
```diff
@@ -15,4 +17,12 @@ export interface Store {
   useRumSlim: boolean
   logEventsFromRequests: boolean
   blockIntakeRequests: boolean
+  local: {
+    [tabId: number]: LocalStore
```
amortemousque, line 21: Is the localStore cleaned up at some point?

developer-extension/manifest.json
```diff
@@ -9,8 +9,14 @@
...
+  "content_scripts": [
+    {
+      "js": ["contentscript.js"],
+      "matches": ["<all_urls>"]
+    }
+  ],
+  "devtools_page": "devtools.html",
   "browser_action": {
-    "default_title": "Browser SDK extension",
-    "default_popup": "popup.html"
```
amortemousque, line 14: Nit: When clicking on the extension icon nothing happens. It could be nice to have a tooltip telling the user where to look at or to open the devtool to the right tab.

developer-extension/src/common/types.ts
bcaudan: not that I know of.
FMU, tab ids are likely to be reused so it should not grow indefinitely.

any concerns?

developer-extension/manifest.json
bcaudan: > open the devtool to the right tab

that could be nice but not sure that is doable

> have a tooltip telling the user where to look at

if we need to explain that, it could be done on the extension readme at some point.
I'd be in favor of removing the `browser_action` in the mean time

developer-extension/src/background/utils.ts
bcaudan: By testing your approach, it appears that window context is not shared between the page (where the added script tag is evaluated) and where the execute script code is evaluated. 
So the `DD_EXTENSION_RESULT` approach does not seem to work.
Did you have different results?

BenoitZugmeyer: Ah indeed. Maybe the event dispatch/listening could be done here then.

bcaudan: probably, IMO this change would shift the complexity but not drastically reduce it like the first suggestion.
I'd be in favor of keeping the current implementation and revisit later.
wdyt?

amortemousque (approved):

cswatt (approved):

developer-extension/src/background/utils.ts
BenoitZugmeyer: Sure, that's fine :)

BenoitZugmeyer (approved):

PR #1179, opened 2021-11-22
✨ [RUMF-1047] implement a Synthetics bundle

BenoitZugmeyer:
## Motivation

We need a bundle to be injected in Synthetics tests, that skips the [previously introduced](https://github.com/DataDog/browser-sdk/pull/1170) check to ignore `init()` calls during Synthetics tests, and expose a different global variable.

## Changes

* Document build modes. Initially I was planning to use a new build mode, so I wrote some doc, but then I chose not to use a build mode (so we can build synthetics bundles with various build modes (staging / release)). I chose to keep the documentation in the PR.

* Create and deploy the new synthetics bundle
* Skip init check for Synthetics bundle
* Rename global variable for Synthetics bundle

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/boot/rum.entry.ts
+++ b/packages/rum/src/boot/rum.entry.ts
@@ -1,5 +1,5 @@
 import { defineGlobal, getGlobalObject } from '@datadog/browser-core'
-import { makeRumPublicApi, RumPublicApi, startRum } from '@datadog/browser-rum-core'
+import { buildEnv, makeRumPublicApi, RumPublicApi, startRum } from '@datadog/browser-rum-core'
 
 import { startRecording } from './startRecording'
 import { makeRecorderApi } from './recorderApi'
@@ -9,5 +9,10 @@ export const datadogRum = makeRumPublicApi(startRum, recorderApi)
 
 interface BrowserWindow extends Window {
   DD_RUM?: RumPublicApi
+  DD_SYNTHETICS_INJECTED_RUM?: RumPublicApi
 }
-defineGlobal(getGlobalObject<BrowserWindow>(), 'DD_RUM', datadogRum)
+defineGlobal(
+  getGlobalObject<BrowserWindow>(),
+  buildEnv.syntheticsBundle ? 'DD_SYNTHETICS_INJECTED_RUM' : 'DD_RUM',
+  datadogRum
+)
--- a/packages/rum/webpack.config.js
+++ b/packages/rum/webpack.config.js
@@ -14,5 +14,8 @@ module.exports = (_env, argv) => [
     mode: argv.mode,
     entry,
     filename: 'datadog-rum-synthetics.js',
+    extraBuildEnv: {
+      SYNTHETICS_BUNDLE: 'true',
+    },
   }),
 ]
--- a/packages/rum/webpack.config.js
+++ b/packages/rum/webpack.config.js
@@ -2,9 +2,17 @@ const path = require('path')
 
 const webpackBase = require('../../webpack.base')
 
-module.exports = (_env, argv) =>
+const entry = path.resolve(__dirname, 'src/boot/rum.entry.ts')
+
+module.exports = (_env, argv) => [
   webpackBase({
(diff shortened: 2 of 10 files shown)


scripts/build-env.js
```diff
@@ -1,8 +1,44 @@
 const execSync = require('child_process').execSync
 const lernaJson = require('../lerna.json')
 
+/**
+ * Allows to produce slightly custom builds of bundles and packages to be used in various
+ * environment. On top of allowing some internal use cases, it also drives the SDK version being
+ * used when requesting the intake (see below).
```
bcaudan, line 7:
```suggestion
 * environment, it allows:
 * - some internal uses cases
 * - different strategies for sdk_version sent to the intake (see below)
```

I had a hard time understanding what you referred to with SDK version

scripts/build-env.js
```diff
@@ -17,6 +53,10 @@ switch (process.env.BUILD_MODE) {
 }
 
 module.exports = {
-  BUILD_MODE: process.env.BUILD_MODE,
+  BUILD_MODE: buildMode,
   SDK_VERSION: sdkVersion,
+
+  // This will be injected by webpack when building the RUM synthetics bundle. When set to 'true',
+  // the bundle will skip the check to ignore init() calls during synthetics tests.
+  SYNTHETICS_BUNDLE: '',
```
bcaudan, line 61: > I chose not to use a build mode (so we can build synthetics bundles with various build modes (staging / release))

do we plan to use various mode for synthetics?

packages/rum/src/boot/rum.entry.ts
```diff
@@ -9,5 +9,10 @@ export const datadogRum = makeRumPublicApi(startRum, recorderApi)
 
 interface BrowserWindow extends Window {
   DD_RUM?: RumPublicApi
+  DD_SYNTHETICS_INJECTED_RUM?: RumPublicApi
 }
-defineGlobal(getGlobalObject<BrowserWindow>(), 'DD_RUM', datadogRum)
+defineGlobal(
+  getGlobalObject<BrowserWindow>(),
+  buildEnv.syntheticsBundle ? 'DD_SYNTHETICS_INJECTED_RUM' : 'DD_RUM',
```
bcaudan, line 16: did not we discuss about having a different name for session cookie as well?

packages/rum/webpack.config.js
```diff
@@ -2,9 +2,20 @@ const path = require('path')
...
+    entry,
     filename: 'datadog-rum.js',
-  })
+  }),
+  webpackBase({
+    mode: argv.mode,
+    entry,
+    filename: 'datadog-rum-synthetics.js',
+    extraBuildEnv: {
+      SYNTHETICS_BUNDLE: 'true',
```
amortemousque, line 18: Instead of a boolean, can't we pass the BUNDLE_NAME to have something like 
```
webpackBase(...{
   BUNDLE_NAME: 'DD_RUM'
}),
webpackBase(..., {
   BUNDLE_NAME: 'DD_SYNTHETICS_INJECTED_RUM'
})
```

scripts/build-env.js
BenoitZugmeyer: No, why?

packages/rum/src/boot/rum.entry.ts
BenoitZugmeyer: With the new approach where Synthetics takes precedence, it is not required anymore, since the customer instance will be ignored completely (it won't create cookies). Do you see a reason why we would need a different cookie?

scripts/build-env.js
```diff
@@ -1,8 +1,44 @@
 const execSync = require('child_process').execSync
 const lernaJson = require('../lerna.json')
 
+/**
```
amortemousque, line 4: Nice! This doc was clearly missing ;)

BenoitZugmeyer:
Superseded by #1188

PR #1187, opened 2021-11-30
♻️ [RUMF-1083] introduce session context history

bcaudan:
## Motivation

Third part of session management rework, cf #1166, #1180

## Changes

- move `ContextHistory` to core
- expose `sessionStore.expireObservable`
- store session context history in session management
- use session history in RUM and Logs

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -192,9 +193,6 @@ export interface RumContext {
 }
 
 export interface ViewContext extends Context {
-  session: {
-    id: string | undefined
-  }
```
BenoitZugmeyer, line 197: ❤️

BenoitZugmeyer (commented):
This looks great!

Maybe we could change the session API a bit to have something like:

```ts
interface SessionHistory {
  findSession(date: RelativeTime): { isTracked(): boolean, getId(): boolean, ...}
}
```

So when accessing multiple attributes from the same session we don't have to look for it through the history every time:

```ts
const session = sessionHistory.findSession(startTime) 
if (session.isTracked()) {
  ...
  session: { id: session.getId()! }
}
```

bcaudan:
> Maybe we could change the session API a bit to have something like:
> 
> ```ts
> interface SessionHistory {
>   findSession(date: RelativeTime): { isTracked(): boolean, getId(): boolean, ...}
> }
> ```

@BenoitZugmeyer that looks interesting, I'd be in favor of exploring it in a following PR though

packages/core/src/domain/session/sessionManagement.ts
```diff
@@ -24,14 +33,34 @@ export function startSessionManagement<TrackingType extends string>(
   const sessionStore = startSessionStore(options, productKey, computeSessionState)
   stopCallbacks.push(() => sessionStore.stop())
 
+  const sessionContextHistory = new ContextHistory<SessionContext<TrackingType>>(SESSION_CONTEXT_TIMEOUT_DELAY)
```
amortemousque, line 36: Since we are using ContextHistory here now, wouldn't it make sens to rename `sessionManagment` to `sessionContexts`. We would be consistent with `urlContexts`, `parentContexts` and we could have the API looking like `sessionContexts.findSession(startTime)`

bcaudan: I would say no.
`urlContext` and `parentContexts` only maintain the contexts, `sessionManagement` has more responsibilities.
About `findSession(startTime)`, it seems to be the same idea as https://github.com/DataDog/browser-sdk/pull/1187#issuecomment-983723926, no?

amortemousque: Let's discuss the naming when doing [ #1187 (comment)](https://github.com/DataDog/browser-sdk/pull/1187#issuecomment-983723926)

amortemousque (approved):

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -32,58 +33,71 @@ export function startSessionStore<TrackingType extends string>(
...
+      cookieSession.created = String(Date.now())
+    }
+    // save changes and expand session duration
+    persistSession(cookieSession, options)
+    return isTracked
+  }
+
+  function hasSessionCache() {
+    return sessionCache.id !== undefined
+  }
```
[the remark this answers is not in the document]
bcaudan: done in 604c333

BenoitZugmeyer (approved):
LGTM!

PR #1188, opened 2021-11-30
:✨⚗ [RUMF-1047] implement a RUM synthetics entry point

BenoitZugmeyer:
## Motivation

Allows the Synthetics worker to inject RUM instances into test runs.

This PR superseds #1179 

## Changes

Provide a new entry point to be consumed by the Synthetics worker. As per RFC, this entry point won't define any global API, and won't ignore the `init()` call if the Synthetics worker is set to automatically inject RUM instance (as this *is* the instance that will be injected).

## Testing

- [x] Local: tested with the e2e test app
- [ ] Staging: can't be tested until a new release is published
- [x] Unit
- [ ] End to end

---


bcaudan (approved):

.eslintrc.js
```diff
@@ -252,7 +252,7 @@ module.exports = {
       },
     },
     {
-      files: ['packages/{rum,logs,rum-slim}/src/index.ts', 'packages/rum/src/internal.ts'],
+      files: ['packages/{rum,logs,rum-slim}/src/index.ts', 'packages/rum/src/{internal,internalSynthetics}.ts'],
```
amortemousque, line 255: I'm wondering about `internalSynthetics.ts`. Shouldn't it be named like `rum.synthetics.entry.ts` since it provide the same ability as `rum.entry.ts` but for synthetics?

BenoitZugmeyer: Good point. While it does about the same job as `rum.entry.ts`, it is a "package entry point" similar to `index.ts`, that's why I think it should stay at the same level as `index.ts`. But maybe we could improve things a bit and have a "entries" folder, like so:

```
 packages/rum/entries/
    main.ts // combination of rum.entry.ts and index.ts
    internal.ts
    internalSynthetics.ts
```
WDYT? cc @bcaudan

amortemousque: Yes, I think it could be nice.

amortemousque (approved):

bcaudan (approved):

PR #1191, opened 2021-12-01
🔊 [RUMF-1083] add monitoring message for session inconsistencies

bcaudan:
## Motivation

Allow to monitor session inconsistencies improvements

## Changes

- consider session outdated if type change
- add monitoring message fo session inconsistencies

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/session/sessionStore.ts
+++ b/packages/core/src/domain/session/sessionStore.ts
@@ -81,6 +81,7 @@ export function startSessionStore<TrackingType extends string>(
   function isSessionInCacheOutdated(cookieSession: SessionState) {
     if (sessionCache.id !== cookieSession.id) {
       if (cookieSession.id) {
+        // cookie id undefined could be due to cookie expiration
         addSessionInconsistenciesMessage(cookieSession, 'different id')
       }
       return true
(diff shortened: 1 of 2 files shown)


amortemousque (approved):

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -79,7 +79,28 @@ export function startSessionStore<TrackingType extends string>(
   }
 
   function isSessionInCacheOutdated(cookieSession: SessionState) {
-    return sessionCache.id !== cookieSession.id || sessionCache[productKey] !== cookieSession[productKey]
+    if (sessionCache.id !== cookieSession.id) {
+      if (cookieSession.id) {
```
BenoitZugmeyer, line 83: Could you explain this condition? I'm not sure to see why we shouldn't display a message when `cookieSession.id` is undefined.

bcaudan: so I see 3 cases:

1. cached id x / stored id y => unexpected change of id
2. cached id undefined / stored id defined => can be new session created in another tab
3. cached id defined / stored id undefined => can be the expiration of the cookie

I should indeed add an extra condition to avoid to report messages for case 2

BenoitZugmeyer: Oh ok. I think it's fine for case 2, since `undefined === undefined`, it won't enter the first `if`. Maybe add a small comment ? `// if cookie session id is undefined, could be the expiration of the cookie`

bcaudan: about case 2, it would enter the first if since one id is defined and not the other.
However, it does mean that a session is created in another tab, this behavior would be from state: no session in cache / stored id undefined.
So it seems all good.
I've added a comment

BenoitZugmeyer (approved):

PR #1194, opened 2021-12-02
⚗✨ [RUMF-1085] implement the `synthetics.injected` field

BenoitZugmeyer:
## Motivation

Help customers distinguish events coming from automatically injected RUM instances.

## Changes

Add a `@synthetics.injected` field

## Testing

- [ ] Local
- [x] Staging: the field is present
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/syntheticsContext.spec.ts
```diff
@@ -21,9 +22,16 @@ describe('getSyntheticsContext', () => {
...
       test_id: 'foo',
       result_id: 'bar',
+      injected: false,
     })
   })
 
+  it('sets the `injected` field to true if the Synthetics test is configured to automatically inject RUM', () => {
+    mockSyntheticsWorkerValues({ publicId: 'foo', resultId: 'bar', injectsRum: true }, 'globals')
+
+    expect(getSyntheticsContext()?.injected).toBeTrue()
```
bcaudan, line 32:
```suggestion
    expect(getSyntheticsContext()!.injected).toBeTrue()
```
no?

bcaudan (approved):

packages/rum-core/src/domain/syntheticsContext.spec.ts
BenoitZugmeyer: Both work! But yes, using `!` would raise slightly less confusing error messages if it fails.

PR #1197, opened 2021-12-06
♻️ [RUMF-1083] rework session management API

bcaudan:
## Motivation

Avoid multiple session history search when accessing multiple session properties
Increase consistency between common / logs / rum session management 

## Changes

- Rename `(rum|logger)Session` to `(rum|logs)SessionManager`
- Replace `isTracked()`, `getId()`, `has(Replay|Lite)Plan()` by `find(Active|Tracked)Session(): Session`
- Stop exposing `isTracked` in Rum and Logs Session, just don't return not tracked session with `findTrackedSession`

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/assembly.ts
```diff
@@ -79,15 +79,16 @@ export function startRumAssembly(
     ({ startTime, rawRumEvent, domainContext, savedCommonContext, customerContext }) => {
       const viewContext = parentContexts.findView(startTime)
       const urlContext = urlContexts.findUrl(startTime)
-      if (session.isTracked(startTime) && viewContext && urlContext) {
+      const session = sessionManager.findSession(startTime)
+      if (session && viewContext && urlContext) {
```
amortemousque, line 83: Is there any cases where we can have a session and no viewContext or no urlContext?

packages/rum-core/src/domain/internalContext.ts
```diff
@@ -10,19 +10,20 @@ import { UrlContexts } from './urlContexts'
...
   parentContexts: ParentContexts,
   urlContexts: UrlContexts
 ) {
   return {
     get: (startTime?: number): InternalContext | undefined => {
       const viewContext = parentContexts.findView(startTime as RelativeTime)
       const urlContext = urlContexts.findUrl(startTime as RelativeTime)
-      if (session.isTracked(startTime as RelativeTime) && viewContext && urlContext) {
+      const session = sessionManager.findSession(startTime as RelativeTime)
+      if (session && viewContext && urlContext) {
```
amortemousque, line 22: Same as https://github.com/DataDog/browser-sdk/pull/1197/files#r764020341

packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -24,46 +27,49 @@ export enum RumTrackingType {
...
+    },
   }
 }
 
 /**
  * Start a tracked replay session stub
  * It needs to be a replay plan in order to get long tasks
  */
-export function startRumSessionStub(): RumSession {
+export function startRumSessionStub(): RumSessionManager {
```
amortemousque, line 62: What about
```suggestion
export function startRumSessionManagerStub(): RumSessionManager {
```

packages/rum-core/src/domain/assembly.ts
bcaudan: I'd say no

amortemousque: To me this check is misleading. I would be happy if we could improve it. 
For intance:
```js
if (session) {
  const viewContext = parentContexts.findView(startTime)
  const urlContext = urlContexts.findUrl(startTime)
  ...
}
```
Wdyt?

bcaudan: It should not be the case but it is not a guarantee that we have by design. 
Your proposal can't be added as is, we should either:

1.  still have to check if viewContext and urlContext are defined
2. add `!` to inform ts about that it should be defined at this point

For 1, I'd say it does not look that great and we would have to do it at several places, I'd prefer to stay with the current implem
For 2, in case I'm wrong about the initial question, I'd prefer to not introduce potential unrelated behavior changes with all pending changes on session. We could mitigate that but I don't think it worth the effort.

Another approach could be to have a design that is better expressing this state, I think we could revisit that when it will be a bit more bothering.

amortemousque: Indeed, it does not have to be mixed with the current PR. 
I'm a bit concerned though. Aren't we trading the meaning of the code to please the linter?

To me, the code should reflect the behaviour we intend from the SDK. 
If we fear an unwanted behaviour, we could add an internal monitoring like for any other investigation. 

Of course, another design could better reflect the state, but if we are always exprecting  `findView` and `findUrl` to return a value, why not doing:

findUrl: (startTime?: RelativeTime) => UrlContext **~~| undefined~~**
findView: (startTime?: RelativeTime) => ViewContext **~~| undefined~~**

We can talk about it outside the PR if you want.

packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -1,14 +1,16 @@
...
-  hasReplayPlan: (startTime?: RelativeTime) => boolean
-  hasLitePlan: (startTime?: RelativeTime) => boolean
+  findSession: (startTime?: RelativeTime) => RumSession | undefined
+}
+
+export type RumSession = {
+  id: string
+  hasReplayPlan: boolean
+  hasLitePlan: boolean
 }
```
BenoitZugmeyer, line 14: I think it would have been better with your proposal in the RFC:
* RumSession has a `isTracked` field
* The `findSession` functions always return an object

Here, `findSession` only returns tracked sessions, which could be confusing. I think we should a single session concept accross the lib, which could be either tracked or not.

bcaudan: From my understanding, across the lib there are 3 different session states:

1. tracked, the session has an id and is updated along the user navigation
2. not tracked, the session does not have an id but it is updated along the user navigation
3. inactive, no session in cookie and waiting for sdk to start or an user interaction after expiration

In the current implementation, `isTracked()` or `getId() !== undefined` is used to identify state 1 VS state 2 and 3, since outside session management we don't care about the difference between not tracked and inactive.

It is why I choose to expose only tracked session outside of session management, it allow us to:

- not expose track / no tracked / inactive concepts outside of session management (as long as we don't need to)
- guarantee that we always have an id (for RUM)

With your proposal:

- the findSession functions always return an object
- have a single session concept accross the lib

I think we should then:

- maintain a state for inactive session, in order to always return something for find session
- expose the difference between not tracked and inactive, in order to have a single concept across the lib

IMO 

- `(rum|logs)Manager.findSession` returning tracked session only is simpler for callers
- introducing extra logic / state is not required for current needs

wdyt?

BenoitZugmeyer: FMU, when starting the session manager, we get or create an active session (by [calling `sessionStore.expandOrRenewSession()`](https://github.com/DataDog/browser-sdk/blob/b8d1a67ee92da9d30b3b75b96fd8b157d9c58d2e/packages/core/src/domain/session/sessionManager.ts#L45)). So there can't be any "inactive" session coming from the session manager. This initial session state is stored in the `sessionContextHistory` starting from the `origin` time, so calling `findSession` will always return a session in practice, right?

bcaudan: seen IRL, `sessionContextHistory.findSession` can return undefined when the session is expired.

added some clarification on c4e7625

BenoitZugmeyer (approved):

amortemousque (approved):

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -27,6 +27,12 @@ export const SESSION_TIME_OUT_DELAY = 4 * utils.ONE_HOUR
 const SESSION_ENTRY_REGEXP = /^([a-z]+)=([a-z0-9-]+)$/
 const SESSION_ENTRY_SEPARATOR = '&'
 
+/**
+ * Different session concepts:
+ * - tracked, the session has an id and is updated along the user navigation
```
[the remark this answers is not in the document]
bcaudan: nice catch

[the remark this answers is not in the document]
bcaudan: oh, but in fact from a session store point of view, it is the case since logs without session id don't use the store.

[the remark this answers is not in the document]
BenoitZugmeyer: Ah nice!

PR #1198, opened 2021-12-07
:bug: remove readonly from all LogsEvent properties

### LINKED ISSUE #1195: [Logs] allowed to use beforeSend to change the log level?
In the following I want to set the log level to `warn` in case of a network error.
This seems to work correctly, but since I am using TS, I get a type error.

```ts
datadogLogs.init({
    ...,
    beforeSend: (log) => {
      if (log.error?.origin === "network") {
        log.status = "warn"; // error TS2540: Cannot assign to 'status' because it is a constant or a read-only property.
      }
    },
    ...
});
```

Is the `status` readonly because you don't allow the `status` to be changed?

kotarella1110:
## Motivation

In some cases, I want to update the readonly `LogsEvent` properties by using `beforeSend`.

resolve #1195

## Changes

remove `readonly` from `LogsEvent` all properties.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


kotarella1110:
> Actually, you could even remove all readonly from logsEvent properties to avoid similar issues to be raised.

@bcaudan I see. will do 👍🏻

kotarella1110:
@bcaudan I think we should avoid updating the following properties, so it is better to leave the `readonly` property as it is.

- `application_id`
- `service`
- `session_id`
- `view.id`

Please let us know what you think.

bcaudan:
> I think we should avoid updating the following properties

it is already possible in javascript, so I would say no reason to have the limitation only in typescript.

kotarella1110:
> it is already possible in javascript, so I would say no reason to have the limitation only in typescript.

Thank for your reply! I will fix.

kotarella1110:
@bcaudan Removed readonly from all properties. Please review 🙏🏻

bcaudan (approved):

BenoitZugmeyer (approved):

BenoitZugmeyer:
Thank you for your contribution! I'll make a new release shortly.

Issue #1193, opened 2021-12-01
Getting "Converting circular structure to JSON" in React App


rafaelguinho:
I'm getting "Converting circular structure to JSON
    --> starting at object with constructor 'XMLHttpRequest'
    |     property '_datadog_xhr' -> object with constructor 'Object'
    --- property 'xhr' closes the circle"

This error ocur after inicialize browser agent and breaks the entire aplication.

rafaelguinho:
I have tried to use NPM module, and CDN sync and async but still the same issue.

metasean:
I'm running into the exact same error in a Nuxt v.2/Vue v.2 app.

bcaudan:
Hello,

I can reproduce by

```js
// init SDK

const xhr = new XMLHttpRequest()
xhr.open('get', 'http://localhost:8080')
xhr.send()

JSON.stringify(xhr)
```

is there any stringification done on xhr on your apps or on the framework you are using? 
or is it something else?

rafaelguinho:
Hello @bcaudan, not in my code,  I'm using React 17.0.1 and axios 0.21.1 for http requests. I did not find any exemple using React on the documentation.

metasean:
Hey @bcaudan - We don't have any in our code and I'm not seeing any in either the @nuxt or @vue module directories.  Like Rafael, we're also using axios, but we're using a slightly more up-to-date version 0.21.4.

metasean:
I've got a bit more info.  I was doing dependency updates when I ran into this, and the Datadog update was a pretty big update (moving up from ^1.26.3).  The only change that I saw in the [changelog](https://github.com/DataDog/browser-sdk/blob/main/CHANGELOG.md) _that looked relevant_ was the removal of `datacenter` from the init.

I've just funneled my way down to the breaking version.  For our app @datadog/browser-rum@v.2.13.0 _works_ while version v.2.14.0 introduces this circular error.  

On the one hand, since v.2.14.0 was released over 6 months ago, I'll be genuinely surprised if other developers haven't already run into this.  On the other hand, in the [v.2.14.0 changelog entry](https://github.com/DataDog/browser-sdk/blob/main/CHANGELOG.md#v2140) there's an XHR bugfix entry (RUMF-876) which _might_ be related.  Unfortunately, I'm not seeing any stringify usage in there, so this could just be a red herring. 🤷‍♂️

bcaudan:
[...]

PR #1204, opened 2021-12-08
🐛 remove self-referential xhr loop

metasean:
The _datadog_xhr object already extends the xhr object, so putting the
xhr object on the _datadog_xhr object causes a self-referential loop,
which can cause problems for anyone using that any of that data, such
as in an error situation.

## Motivation

Removes a self-referential loop. (https://github.com/DataDog/browser-sdk/issues/1193)

## Changes

Literally just removes a self-referential loop that causes the following if the xhr or _datadog_xhr object are referenced.
self-referential-loop-error

## Testing

Added a test to verify the self-reference doesn't exist.

- [ ] Local - not sure how you want these tested, other than using the modified version in our app no longer causes the error
- [ ] Staging - not sure how you want this tested
- [x] Unit - success
- [ ] End to end - COULDN'T RUN - I've got a current version of Chrome, meanwhile the ChromeDriver will only work with an older version of Chrome:
```
[0-4] 2021-12-08T22:24:57.827Z ERROR webdriver: Request failed with status 500 due to session not created: session not created: This version of ChromeDriver only supports Chrome version 94
Current browser version is 96.0.4664.55 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
```

---

### DIFF
--- a/packages/core/src/browser/xhrObservable.spec.ts
+++ b/packages/core/src/browser/xhrObservable.spec.ts
@@ -90,6 +90,7 @@ describe('xhr proxy', () => {
         expect(request.isAborted).toBe(false)
         expect(request.startTime).toEqual(jasmine.any(Number))
         expect(request.duration).toEqual(jasmine.any(Number))
+        expect(request.xhr).not.toBeDefined()
         done()
       },
     })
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer:
Thank you for your contribution. Sadly, we need the `xhr` value [here](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/requestCollection.ts#L68) and [here](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/requestCollection.ts#L88). We'll remove the `_datadog_xhr` property altogether in the upcoming v4 release, so your issue will be fixed as well.

PR #1205, opened 2021-12-09
:sparkles: allows specification of forward errors to logs

### LINKED ISSUE #1202: [Logs] configure which errors to forward
The logs forwarded in case of `forwardErrorsToLogs: true` is as follows:

- `console.error` logs
- exceptions
- network errors

Currently, I use `beforeSend` to avoid forwarding the `console.error` logs as follows:

```ts
datadogLogs.init({
    ...,
    forwardErrorsToLogs: true,
    beforeSend: (log) => {
      if (log.error?.origin === "console") {
        return false;
      }
    },
    ...
});
```

I think it would be more useful to achieve this without using `beforeSend`.
I request `forwardErrorsToLogs` to accept both `booleans` and `error.origins`, and the ability to forward only specific error logs.

```ts
datadogLogs.init({
    ...,
    forwardErrorsToLogs: ["source", "network"], // (property) LogsInitConfiguration.forwardErrorsToLogs?: boolean | ("console" | "source" | "network")[] | undefined
    ...
});
```

If you guys like this feature, I will create a PR for it.

kotarella1110:
## Motivation

It would be useful to have an option to forward only specific errors to logs.

resolve #1202

## Changes

I added an `Array` type to the `LogsInitConfiguration.forwardErrorsToLogs` option so that we can specify which logs we want to forward.

```diff
- LogsInitConfiguration.forwardErrorsToLogs?: boolean | undefined
+ LogsInitConfiguration.forwardErrorsToLogs?: boolean | Array<'network' | 'source' | 'console'> | undefined
```

## Testing

memo:I tried to add unit tests, but Jasmine doesn't support mocking named export modules, so I gave up on adding tests.

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan:
cf https://github.com/DataDog/browser-sdk/issues/1202#issuecomment-992359646

PR #1217, opened 2021-12-14
♻️ [RUMF-1097] revamp configuration - logs

BenoitZugmeyer:
## Motivation

Followup of #1216 . This PR focuses on the logs part.

## Changes

* Move init configuration typings into a new `logs/src/domain/configuration` module
* Introduce a new `LogsConfiguration` to be used internally in the Logs SDK 
* Validate and build the configuration in `init()`, by using the previously introduced `validateAndBuildConfiguration` function from `core`
* Add `forwardErrorToLogs` to `LogsConfiguration`, to avoid using `InitConfiguration` outside of `init()`

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/boot/logsPublicApi.spec.ts
+++ b/packages/logs/src/boot/logsPublicApi.spec.ts
@@ -7,10 +7,10 @@ import {
   resetExperimentalFeatures,
 } from '@datadog/browser-core'
 import { Clock, deleteEventBridgeStub, initEventBridgeStub, mockClock } from '../../../core/test/specHelper'
+import { HybridInitConfiguration, LogsInitConfiguration } from '../domain/configuration'
 
 import { HandlerType, LogsMessage, StatusType } from '../domain/logger'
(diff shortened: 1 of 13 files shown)


packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -76,41 +79,28 @@ describe('logs entry', () => {
       setDebug(false)
     })
 
-    it('init should log an error if sampleRate is invalid', () => {
-      LOGS.init({ clientToken: 'yes', sampleRate: 'foo' as any })
-      expect(displaySpy).toHaveBeenCalledTimes(1)
-
-      LOGS.init({ clientToken: 'yes', sampleRate: 200 })
-      expect(displaySpy).toHaveBeenCalledTimes(2)
```
bcaudan, line 84: we have a test covering the 'foo' case in core but it does not seem that we still have one about '200'.
what about adding it to keep this case?

bcaudan (approved):

amortemousque (approved):

PR #1219, opened 2021-12-15
✅ [RUMF-1090] update minimal version to 3.8.2

BenoitZugmeyer:
## Motivation

Allows to use newer TS features in consumed typings (ex: `import type`)

## Changes

Update minimal TS version to 3.8.2 (the first stable release of TS 3.8)

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/test/app/package.json
+++ b/test/app/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "ts-loader": "6.2.1",
-    "typescript": "3.0.1",
+    "typescript": "3.8.2",
     "webpack": "5.28.0"
   }
 }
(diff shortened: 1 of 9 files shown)


test/app/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "ts-loader": "6.2.1",
-    "typescript": "3.0.1",
```
BenoitZugmeyer, line 22: Stable 3.8.0 and 3.8.1 don't exist: they were used as dev and rc versions. Let's use the first 3.8 stable version, which is 3.8.2

test/app/package.json
```diff
@@ -19,7 +19,7 @@
...
+    "typescript": "3.8.2",
```
BenoitZugmeyer, line 22: Stable 3.8.0 and 3.8.1 don't exist: they were used as dev and rc versions. Let's use the first 3.8 stable version, which is 3.8.2

bcaudan: We should update RUM and Logs Readme where we mention:

> Types are compatible with TypeScript >= 3.0

bcaudan (approved):

sarina-dd (approved):

PR #1221, opened 2021-12-15
♻️ [RUMF-1097] revamp configuration - rum

BenoitZugmeyer:
## Motivation

Followup of #1216 . This PR focuses on the logs part.

## Changes

* Move init configuration typings into a new `rum/src/domain/configuration` module
* Introduce a new `RumConfiguration` to be used internally in the RUM SDK 
* Validate and build the configuration in `init()`, by using the previously introduced `validateAndBuildConfiguration` function from `core`
* Add `applicationId` to `RumConfiguration`, to avoid using `InitConfiguration` outside of `init()`
* Cleanup previous `buildConfiguration` function.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/configuration.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export type RumConfiguration = Configuration &
+  typeof DEFAULT_RUM_CONFIGURATION & {
+    applicationId: string
+
+    actionNameAttribute: string | undefined
+  }
+
+export function validateAndBuildRumConfiguration(
```
amortemousque, line 39: I tried a different implementation [approach](https://github.com/DataDog/browser-sdk/blob/aymeric/revamp-configuration-rum-refacto/packages/rum-core/src/domain/configuration.ts):
My pros:
- Shorter
- Isolate init conf check at the top of the function
- No configuration mutation
- Less objective but avoid the RumConfiguration type union which I found a bit complicated 
Cons 
- Init config undefined check is done twice
- The default config might be a bit more hard to read

Let me know what you think. If we choose to do it. It can be done as a follow up to not block the v4.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -112,68 +86,50 @@ export function validateAndBuildConfiguration(
...
-
-  if ('allowedTracingOrigins' in initConfiguration) {
-    configuration.allowedTracingOrigins = initConfiguration.allowedTracingOrigins!
-  }
-
-  if ('sampleRate' in initConfiguration) {
-    configuration.sampleRate = initConfiguration.sampleRate!
+    silentMultipleInit: !!initConfiguration.silentMultipleInit,
+
+    /**
```
amortemousque, line 107: Nit: I'd be in favour of adding comments on the type above. It will make it available for the completion

amortemousque (approved):

packages/core/src/domain/configuration/configuration.ts
BenoitZugmeyer: But those comments are here to justify the values used, not to provide documentation on each key.. I think we should keep it close to the value.

amortemousque: Fair enough. Lgtm then

bcaudan (approved):
Nice!

PR #1222, opened 2021-12-16
♻️ [RUMF-1092] use a WeakMap to store XHR context

BenoitZugmeyer:
## Motivation

Don't expose internal XHR context to users. Also, it will avoid any conflict between double instrumentation when two SDKs are present.

## Changes

Use a WeakMap instead of a property to store the XHR context

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/xhrObservable.ts
+++ b/packages/core/src/browser/xhrObservable.ts
@@ -4,18 +4,6 @@ import { Observable } from '../tools/observable'
 import { Duration, elapsed, relativeNow, RelativeTime, ClocksState, clocksNow, timeStampNow } from '../tools/timeUtils'
 import { normalizeUrl } from '../tools/urlPolyfill'
 
-interface BrowserXHR<T> extends XMLHttpRequest {
-  /**
-   * @deprecated this property is shared by both logs and rum and can be used by different code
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/xhrObservable.ts
```diff
@@ -40,6 +28,7 @@ export interface XhrCompleteContext extends Omit<XhrStartContext, 'state'> {
 export type XhrContext = XhrOpenContext | XhrStartContext | XhrCompleteContext
 
 let xhrObservable: Observable<XhrContext> | undefined
+const xhrContexts: WeakMap<XMLHttpRequest, XhrContext> = new WeakMap()
```
amortemousque, line 31: Nit: 
```suggestion
const xhrContexts = new WeakMap<XMLHttpRequest, XhrContext>()
```

amortemousque (approved):

packages/core/src/browser/xhrObservable.ts
BenoitZugmeyer: Simpler indeed!

bcaudan (approved):

PR #1223, opened 2021-12-16
🐛 Fix proxy url intake detection

amortemousque:
## Motivation

With the current proxyUrl implementation, if a user set the same url for his proxy and his backend, we con't collect backend requests as resources. 

Example:
```js
init({proxyUrl: 'https://www.proxy.com'...})
fetch('https://www.proxy.com/foo') // no resource collected
```

## Changes

Update buildIntakeUrl logic 

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/transportConfiguration.spec.ts
+++ b/packages/core/src/domain/configuration/transportConfiguration.spec.ts
@@ -218,7 +218,7 @@ describe('transportConfiguration', () => {
       ).toBe(true)
     })
 
-    it('should detect proxy intake request', () => {
+    it('should detect proxy intake request with proxyHost', () => {
       let configuration = computeTransportConfiguration({ clientToken, proxyHost: 'www.proxy.com' }, buildEnv)
       expect(configuration.isIntakeUrl(`https://www.proxy.com${v1IntakePath}?xxx`)).toBe(true)
       configuration = computeTransportConfiguration(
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/configuration/transportConfiguration.spec.ts
```diff
@@ -230,11 +230,27 @@ describe('transportConfiguration', () => {
...
 
-    it('should not detect request done on the same host as the proxy', () => {
+    it('should not detect request done on the same host as the proxy with proxyHost', () => {
       const configuration = computeTransportConfiguration({ clientToken, proxyHost: 'www.proxy.com' }, buildEnv)
       expect(configuration.isIntakeUrl('https://www.proxy.com/foo')).toBe(false)
     })
 
+    it('should detect proxy intake request with proxyUrl', () => {
+      let configuration = computeTransportConfiguration({ clientToken, proxyUrl: 'https://www.proxy.com' }, buildEnv)
+      expect(configuration.isIntakeUrl(`https://www.proxy.com?ddforward=xxx`)).toBe(true)
```
BenoitZugmeyer, line 240: If the `proxyUrl` is set to `https://www.proxy.com` (like the example), `isIntakeUrl` will actually be called with an URL like `https://www.proxy.com/?ddforward=xxx`: an ending slash will be inserted by the browser.

So with your implementation, `isIntakeUrl` will return `false`, and intake requests may produce a RUM event.

See [this jsbin](https://jsbin.com/neburisika/edit?js,console,output) to illustrate the issue.

To fix this issue, you may want to use [`normalizeUrl`](https://github.com/DataDog/browser-sdk/blob/e9fc8ff542d4df5be2cd27ea6f107586a4330381/packages/core/src/tools/urlPolyfill.ts#L3) to sanitize the `proxyUrl` before using it in `endpointBuilder`

amortemousque: Nice catch thx!

amortemousque: Do you know why the browser transform:
https://www.proxy.com?ddforward=xxx to https://www.proxy.com/?ddforward=xxx
and not
https://www.proxy.com/foo?ddforward=xxx to https://www.proxy.com/foo/?ddforward=xxx
?

amortemousque: Done

BenoitZugmeyer: I think the 'path' of an URL is "absolute" and should start with a `/`. After that, it may or may not end with a `/` (it could symbolize a directory or a file)

BenoitZugmeyer (approved):

PR #1226, opened 2021-12-17
🐛 improve privacy computation on a single node

BenoitZugmeyer:
## Motivation

When specifying multiple privacy mode on a single node, the result is not intuitive:
* First, if a privacy attribute is present, it will always take precedence over the class name.
* Then, if two privacy class names are specified, the precedence order is weird (ex: ALLOW takes precedence over MASK or HIDDEN).

## Changes

This commit make sure class names and attributes are treated equally and that the most restrictive privacy level takes precedence over the other.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


bcaudan (approved):
LGTM

PR #1227, opened 2021-12-20
🔥 [RUMF-1093] deprecated proxyhost option

amortemousque:
## Motivation

Remove the proxyHost option and related implementation

## Changes

Update endpoint builder

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -62,11 +62,6 @@ export interface InitConfiguration {
   silentMultipleInit?: boolean | undefined
   trackInteractions?: boolean | undefined
   trackViewsManually?: boolean | undefined
-
-  /**
-   * @deprecated Favor proxyUrl option
-   */
-  proxyHost?: string | undefined
   proxyUrl?: string | undefined
   beforeSend?: BeforeSendCallback | undefined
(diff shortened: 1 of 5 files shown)


BenoitZugmeyer (commented):
Because the logic to build endpoint URLs is getting simpler, I was wondering if we couldn't reduce the complexity by:

* removing internal functions
* returning URLs directly instead of 'build' functions

See https://gist.github.com/BenoitZugmeyer/3c4120d9c0a64afeafdf54cb0e2b44ef 
WDYT?

amortemousque:
> Because the logic to build endpoint URLs is getting simpler, I was wondering if we couldn't reduce the complexity by:
> 
> * removing internal functions
> * returning URLs directly instead of 'build' functions
> 
> See https://gist.github.com/BenoitZugmeyer/3c4120d9c0a64afeafdf54cb0e2b44ef WDYT?

The only issue I see, I that dd-request-id and batch_time have to be build for each request. How would you manage it?

BenoitZugmeyer:
Ah right. Well we could still simplify it, but keep the `build()` function?

amortemousque:
> 

Yes nice :) I'm doing it

BenoitZugmeyer (approved):
Nice!

bcaudan (approved):

PR #1228, opened 2021-12-20
🐛 [RUMF-1122] fix view updates while session is expired

bcaudan:
## Motivation

Bump in session time spent matching the deployment of #1187. 
Suspected root cause: regression allowing views to send updates while the session is expired

## Changes

For view updates, check if the session is currently tracked instead of if the session was tracked at the beginning of the view.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -81,7 +81,7 @@ export function startRumAssembly(
     ({ startTime, rawRumEvent, domainContext, savedCommonContext, customerContext }) => {
       const viewContext = parentContexts.findView(startTime)
       const urlContext = urlContexts.findUrl(startTime)
-      const session = sessionManager.findTrackedSession(startTime)
+      const session = sessionManager.findTrackedSession(rawRumEvent.type !== RumEventType.VIEW ? startTime : undefined)
       if (session && viewContext && urlContext) {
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/assembly.ts
```diff
@@ -81,7 +81,7 @@ export function startRumAssembly(
     ({ startTime, rawRumEvent, domainContext, savedCommonContext, customerContext }) => {
       const viewContext = parentContexts.findView(startTime)
       const urlContext = urlContexts.findUrl(startTime)
-      const session = sessionManager.findTrackedSession(startTime)
+      const session = sessionManager.findTrackedSession(rawRumEvent.type !== RumEventType.VIEW ? startTime : undefined)
```
BenoitZugmeyer, line 84: When looking at the `assembly` code alone, this seems risky. It indicates that a View could potentially be associated to multiple sessions when updates are sent before and after a session renew. What about adding a comment to explain:

* why are we using the current session for view events (why are we making this exception)
* why this is fine and view events are only generated for the current session (contrary to other events)

bcaudan: added some notes + creating a task for the todo

BenoitZugmeyer (approved):

PR #1230, opened 2021-12-21
⚗[RUMF-1079] add lock mechanism for session cookie writes

bcaudan:
## Motivation

Mitigate session inconsistencies due to concurrent access to cookies by different tabs (chromium browsers).

## Changes

- introduce a sessionCookieStore for session cookie methods 
- introduce an API to write into session cookie protected by lock
- lock mechanism enabled by flag `cookie-lock`

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -1,27 +1,76 @@
...
+      postpone(operations)
+      return
+    }
+  }
   if (processedSession) {
-    persistSession(processedSession, options)
+    persistSession(processedSession, operations.options)
+  }
+  if (isExperimentalFeatureEnabled('cookie-lock')) {
+    if (!processedSession || !utils.isEmptyObject(processedSession)) {
```
BenoitZugmeyer, line 70: **❓ question:** ‏ I wonder why there is a different behavior when writing the `processedSession` if it is empty or not. FMU, when the `processedSession` is not an empty object, we "persist" it with the `lock` attribute present, and then we remove the `lock` attribute, for integrity check. Don't we need the same integrity check when removing the session / writing an empty object?

packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -1,27 +1,76 @@
...
+        postpone(operations)
+        return
+      }
+      delete currentSession.lock
+      setSession(currentSession, operations.options)
+      processedSession = currentSession
+    }
   }
-  after?.(processedSession || initialSession)
+  operations.after?.(processedSession || currentSession)
```
BenoitZugmeyer, line 82: **💬 suggestion:** ‏Add a comment to justify calling `after` even if `process` returns undefined. FMU this is to handle the case when the session has been modified in another tab?

packages/core/src/domain/session/sessionCookieStore.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+    cookie.getSpy.and.callFake(() => {
+      const currentOnLockCheck = onLockChecks.shift()
+      if (!currentOnLockCheck) {
+        return currentSession
+      }
+      cookie.getSpy.and.callThrough()
+      cookie.setSpy.and.callThrough()
+      const { currentState, retryState } = currentOnLockCheck()
+      persistSession(retryState, COOKIE_OPTIONS)
+      return `${SESSION_COOKIE_NAME}=${toSessionString(currentState)}`
```
BenoitZugmeyer: **💬 suggestion:** ‏This is really hard to get. What about not relying on the original cookie stub spies and keep using the `currentSession` instead?

```ts
      const currentOnLockCheck = onLockChecks.shift()
      if (currentOnLockCheck) {
        const { currentState, retryState } = currentOnLockCheck()
        currentSession = `${SESSION_COOKIE_NAME}=${toSessionString(currentState)}`
        return `${SESSION_COOKIE_NAME}=${toSessionString(retryState)}`
      } else {
        return currentSession
      }
```

packages/core/src/domain/session/sessionCookieStore.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+  const COOKIE_OPTIONS = {}
+  let initialSession: SessionState
+  let otherSession: SessionState
+  let processSpy: jasmine.Spy<jasmine.Func>
+  let afterSpy: jasmine.Spy<jasmine.Func>
+  let cookie: ReturnType<typeof stubCookie>
+
+  type OnLockCheck = () => { currentState: SessionState; retryState: SessionState }
+
+  function lockScenario({
```
BenoitZugmeyer, line 25: **🥜 nitpick:** ‏since this function is only used in the last `it`, what about moving it closer to the end of the file?

packages/core/src/domain/session/sessionCookieStore.spec.ts
```diff
@@ -195,5 +206,24 @@ describe('session cookie store', () => {
         })
       })
     })
+
+    it('should abort after a max number of retry', (done) => {
+      const maxNumberOfRetriesShouldBeReached = MAX_NUMBER_OF_LOCK_RETRIES * LOCK_RETRY_DELAY * 2
```
BenoitZugmeyer, line 207: **💬 suggestion:** ‏maybe mock the clock to avoid `setTimeout` usages that may be flaky

packages/core/src/domain/session/sessionCookieStore.spec.ts
```diff
@@ -225,5 +225,31 @@ describe('session cookie store', () => {
...
+      persistSession(initialSession, COOKIE_OPTIONS)
+
+      withCookieLockAccess({
+        options: COOKIE_OPTIONS,
+        process: (session) => ({ ...session, first: 'added' }),
+        after: afterSpy,
+      })
+      withCookieLockAccess({
+        options: COOKIE_OPTIONS,
+        process: (session) => ({ ...session, second: 'added' }),
```
BenoitZugmeyer, line 241: **💬 suggestion:** ‏maybe write the cookie based on the previous value, ex:

```
withCookieLockAccess({
   ...
 process: (session) => ({ ...session, value: 'foo' }),
 ...
})
withCookieLockAccess({
 ...
 process: (session) => ({ ...session, value: session.value + 'bar' }),
 ...
// then check if value is `foobar`
```

So we know that both actions have been called in the good order.

bcaudan: indeed 👍

packages/core/src/domain/session/sessionCookieStore.spec.ts
bcaudan: indeed 👍

packages/core/src/domain/session/sessionCookieStore.spec.ts
bcaudan: I've made a simplification based on your proposition but not sure where the `currentSession` came from on the proposed code, so let me know if it was what you had in mind.

packages/core/src/domain/session/sessionCookieStore.ts
bcaudan: > FMU this is to handle the case when the session has been modified in another tab?

indeed, with `watchSession`

added a comment

packages/core/src/domain/session/sessionCookieStore.ts
bcaudan: Indeed, the issue with writing an empty session and keeping the lock is that the session is not empty anymore since we need to keep track of the lock in the session.
We could handle that by tweaking the clear session logic or by requiring withCookieLockAccess clients to handle the lock but it would increase the complexity.
This tradeoff seems simpler and I'd expect less concurrency issues around session expiration.

If that sounds ok to you, I could add some comments to clarify that though.

packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -1,27 +1,76 @@
...
+}
+
+function postpone(operations: Operations) {
```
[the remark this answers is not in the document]
bcaudan: IMO `postpone` conveys the information that it will not be executed again immediately.
maybe `retryLater`?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -40,28 +40,58 @@ export function startSessionStore<TrackingType extends string>(
...
-    if (hasSessionInCache()) {
-      persistSession(cookieSession, options)
+    withCookieLockAccess({
+      process: (cookieSession) => (hasSessionInCache() ? synchronizeSession(cookieSession) : undefined),
+    })
+  }
+
+  function watchSession() {
+    withCookieLockAccess({
+      process: (cookieSession) => (!isActiveSession(cookieSession) ? {} : undefined),
```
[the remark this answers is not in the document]
bcaudan: indeed, added some documentation to clarify

packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -19,9 +19,20 @@ type Operations = {
   after?: (cookieSession: SessionState) => void
 }
 
+const bufferedOperations: Operations[] = []
+let onGoingOperations: Operations | undefined
```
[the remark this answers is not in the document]
bcaudan: Since the ongoing operations can be postponed, we need to know if the current operations are the ongoing ones or not, in order to either attempt the operations or buffered them:
```ts
if (operations !== ongoingOperations) {
  bufferedOperations.push(operations)
  return
}
```

packages/core/test/specHelper.ts
```diff
@@ -326,3 +326,27 @@ export function disableJasmineUncaughtErrorHandler() {
...
+    get: () => doc.getCookie(),
+    set: (cookie) => doc.setCookie(cookie),
+  })
+  return {
+    stop: () => {
+      delete (document as any).cookie
+    },
+    setSpy,
+    getSpy,
+  }
```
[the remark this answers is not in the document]
bcaudan: oh nice 👍

packages/core/src/domain/session/sessionCookieStore.ts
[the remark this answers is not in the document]
BenoitZugmeyer: I see. Having two functions would do the trick then, something like (the `hasOngoingOperations` flag would be true even when waiting for a retry):

```ts
const bufferedOperations: Operations[] = []
let hasOngoingOperations = false

export function withCookieLockAccess(operations: Operations) {
  if (hasOngoingOperations) {
    bufferedOperations.push(operations)
    return
  }
  hasOngoingOperations = true
  runOperations(operations, 0)
}

function runOperations(operations: Operations, numberOfRetries: number) {
   // ... rest of the current `withCookieLockAccess` logic
}

function next() { // could be renamed `onOperationsEnd`
  hasOngoingOperations = false
  const nextOperations = bufferedOperations.shift()
  if (nextOperations) {
    withCookieLockAccess(nextOperations)
  }
}
```

packages/core/src/domain/session/sessionCookieStore.ts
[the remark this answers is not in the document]
BenoitZugmeyer: `retryLater` LGTM!

packages/core/src/domain/session/sessionCookieStore.ts
BenoitZugmeyer: We could simply make sure the lock is present in the `processedSession` object by doing something like this in `withCookieLockAccess`: 

```ts
processedSession.lock = lock
```

packages/core/src/domain/session/sessionCookieStore.spec.ts
BenoitZugmeyer: The `currentSession` is [the local variable you introduced to store the current session cookie](https://github.com/DataDog/browser-sdk/blob/8eb1617f3c9db8176a8abbafdec6b518f45d1548/packages/core/src/domain/session/sessionCookieStore.spec.ts#L34) . Your change does exactly the same but with a bit of indirection, it looks good to me :)

packages/core/src/domain/session/sessionCookieStore.ts
bcaudan: This approach is the one that I had in mind behind the "tweaking the clear session logic".
FMU, by implementing it as is, when wanting to clear the session, `persistSession` will put `lock=x&expire=${now + expirationDelay}` in the cookie.
So this session could be considered as active, which even if does not seem to be an issue now is not ideal.
I'd then be in favor of handling this case differently in order to not cause unexpected issue.
Then, to release the lock for this case, we would need to handle it differently to also set the cookie expiration delay at 0.

Let's discuss it synchronously if you want.

packages/core/src/domain/session/sessionCookieStore.ts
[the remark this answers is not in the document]
bcaudan: It should work, yes.
Does it worth to do the change though? IMO this one move the complexity to another part of the design rather than reducing it like the first suggestion.

BenoitZugmeyer (approved):

amortemousque (approved):

PR #1231, opened 2021-12-21
👷 [RUMF-1106] enable some ESLint import rules

BenoitZugmeyer:
## Motivation

Enforce `import` path consistency

## Changes

* Enable the `import/no-useless-path-segments` rule to make sure import paths are the simplest possible. 

* Enable the `import/no-unresolved` rule and  extend the `plugin:import/typescript` ESLint configuration to make sure ESLint is able to resolve all files (currently there is some imports that work with TS but not with ESLint)

* Enable the` import/no-extraneous-dependencies` rule to make sure we don’t import from the same package (ex: importing from @datadog/browser-core in the core package). 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/developer-extension/package.json
+++ b/developer-extension/package.json
@@ -12,6 +12,7 @@
     "@types/react-dom": "16",
     "copy-webpack-plugin": "8.1.0",
     "html-webpack-plugin": "5.3.1",
+    "webpack": "5.28.0",
     "webpack-webextension-plugin": "0.4.1"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "replace-in-file": "6.2.0",
     "sinon": "9.2.4",
     "string-replace-loader": "3.0.1",
+    "terser-webpack-plugin": "5.1.1",
     "ts-loader": "8.0.18",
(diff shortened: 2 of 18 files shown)


developer-extension/package.json
```diff
@@ -12,6 +12,7 @@
     "@types/react-dom": "16",
     "copy-webpack-plugin": "8.1.0",
     "html-webpack-plugin": "5.3.1",
+    "webpack": "5.28.0",
```
BenoitZugmeyer, line 15: Added, because we `require` it in `developer-extension/webpack.config.js`

package.json
```diff
@@ -79,6 +79,7 @@
     "replace-in-file": "6.2.0",
     "sinon": "9.2.4",
     "string-replace-loader": "3.0.1",
+    "terser-webpack-plugin": "5.1.1",
```
BenoitZugmeyer, line 82: Added because we `require` it in `webpack.base.js`

bcaudan (approved):
Nice!

amortemousque (approved):
Nice !

Issue #822, opened 2021-04-29
Improve beforeSend typing


nmbro:
Currently:

https://github.com/DataDog/browser-sdk/blob/270cff1aa97aa1202a88e63fc20d429289b33ee4/packages/core/src/domain/configuration.ts#L54

`beforesend(event)` is of type any, is it possible to get a better type here instead of any?

bcaudan:
Hello @nmbro,

It is actually but from Logs and RUM configuration interfaces:

https://github.com/DataDog/browser-sdk/blob/270cff1aa97aa1202a88e63fc20d429289b33ee4/packages/logs/src/boot/logs.ts#L25

https://github.com/DataDog/browser-sdk/blob/270cff1aa97aa1202a88e63fc20d429289b33ee4/packages/rum-core/src/boot/rumPublicApi.ts#L25

How are you using it?
There is probably something for us to improve here.

nmbro:
It did autocomplete it in my IDE.

I was using it for the RUM part

bcaudan:
Hum, not sure why your IDE tried to pick the one from core 😕 
We should probably get rid of it anyway.

Have you managed to make it work with the right types?

nmbro:
It probably didn't, I was just browsing the source code, so I was probably just oblivious to the actual correct implementations / interfaces

bcaudan:
Things should have been improved in v4, feel free to reopen otherwise.

PR #1232, opened 2021-12-21
🏷️ [RUMF-1098] move init options into their related interfaces

BenoitZugmeyer:
## Motivation

Improve init configuration typings 

## Changes

* 🚚 Move some `InitConfiguration` properties to their own package.
* 🎨 Group/reorder properties

## Testing

- [ ] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -16,7 +16,7 @@ export type DefaultPrivacyLevel = typeof DefaultPrivacyLevel[keyof typeof Defaul
 export interface InitConfiguration {
   // global options
   clientToken: string
-  beforeSend?: BeforeSendCallback | undefined
+  beforeSend?: GenericBeforeSendCallback | undefined
   sampleRate?: number | undefined
   silentMultipleInit?: boolean | undefined
 
@@ -41,7 +41,9 @@ export interface InitConfiguration {
   replica?: ReplicaUserConfiguration | undefined
 }
 
-export type BeforeSendCallback = (event: any, context?: any) => unknown
+// This type is only used to build the core configuration. Logs and RUM SDKs are using a proper type
+// for this option.
+type GenericBeforeSendCallback = (event: any, context?: any) => unknown
 
 interface ReplicaUserConfiguration {
   applicationId?: string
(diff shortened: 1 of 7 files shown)


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -14,30 +14,29 @@ export const DefaultPrivacyLevel = {
 export type DefaultPrivacyLevel = typeof DefaultPrivacyLevel[keyof typeof DefaultPrivacyLevel]
 
 export interface InitConfiguration {
+  // global options
   clientToken: string
-  applicationId?: string | undefined
-  actionNameAttribute?: string | undefined
-  internalMonitoringApiKey?: string | undefined
-  allowedTracingOrigins?: Array<string | RegExp> | undefined
+  beforeSend?: BeforeSendCallback | undefined
```
bcaudan, line 19: do we need to keep these common `beforeSend` and `BeforeSendCallback`?
cf #822

packages/rum-core/src/domain/configuration.ts
```diff
@@ -12,9 +12,23 @@ import { RumEventDomainContext } from '../domainContext.types'
 import { RumEvent } from '../rumEvent.types'
 
 export interface RumInitConfiguration extends InitConfiguration {
+  // global options
   applicationId: string
   beforeSend?: ((event: RumEvent, context: RumEventDomainContext) => void | boolean) | undefined
+
+  // tracing options
+  allowedTracingOrigins?: Array<string | RegExp> | undefined
```
bcaudan, line 20: Should we use `ReadonlyArray` now?

packages/core/src/domain/configuration/configuration.ts
BenoitZugmeyer: Mmh I can move it. The downside is that we'll need to have the following code in both RUM and Logs configuration build:
```ts
    beforeSend:
```
but I guess that's fine

bcaudan: indeed.
Alternate proposal: keep the code here but make more clear here that it is a more generic type with real implementation in logs / rum.
it should avoid the confusion of the related ticket.
wdyt?

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -41,7 +41,9 @@ export interface InitConfiguration {
   replica?: ReplicaUserConfiguration | undefined
 }
 
-export type BeforeSendCallback = (event: any, context?: any) => unknown
+// This type is only used to build the core configuration. Logs and RUM SDKs are using a proper type
+// for this option.
+type GenericBeforeSendCallback = (event: any, context?: any) => unknown
```
bcaudan, line 46: 👍

bcaudan (approved):

amortemousque (approved):

PR #1233, opened 2021-12-22
✨ [RUMF-1119] Implement dual ship for other orgs

amortemousque:
## Motivation

Other Datadog data sources are dual shipped to US1 (org 2). In order to allow correlation of this data with RUM data, we would like to dual ship browser data from other data centers (except govcloud) to US1.

The PR enable the replica configuration for other orgs

## Changes

- Add the new datacenter tag
- Remove the browser SDK staging safe guard

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -234,7 +234,7 @@ deploy-staging:
     - .staging
   script:
     - yarn
-    - BUILD_MODE=staging yarn build:bundle
+    - BUILD_MODE=canary yarn build:bundle
     - ./scripts/deploy.sh staging staging
 
 deploy-prod-canary:
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -26,6 +26,13 @@ const INTAKE_TRACKS = {
 
 export type EndpointType = keyof typeof ENDPOINTS[IntakeType]
 
+const INTAKE_DATACENTERS: { [site: string]: string } = {
+  'datadoghq.eu': 'eu1.prod.dog',
+  'datadoghq.com': 'us1.prod.dog',
+  'us3.datadoghq.com': 'us3.prod.dog',
+  'us5.datadoghq.com': 'us5.prod.dog',
+}
+
 export const INTAKE_SITE_US = 'datadoghq.com'
 const INTAKE_SITE_US3 = 'us3.datadoghq.com'
 const INTAKE_SITE_GOV = 'ddog-gov.com'
@@ -41,6 +48,7 @@ export function createEndpointBuilder(
   initConfiguration: InitConfiguration,
   buildEnv: BuildEnv,
   endpointType: EndpointType,
+  sourceSite?: string,
   source?: string
 ) {
   const sdkVersion = buildEnv.sdkVersion
@@ -96,11 +104,13 @@ export function createEndpointBuilder(
   }
 
   function buildQueryParameters(endpointType: EndpointType, source?: string) {
+    const datacenter = INTAKE_DATACENTERS[sourceSite ?? site]
     const tags =
       `sdk_version:${sdkVersion}` +
       `${env ? `,env:${env}` : ''}` +
       `${service ? `,service:${service}` : ''}` +
-      `${version ? `,version:${version}` : ''}`
+      `${version ? `,version:${version}` : ''}` +
+      `${datacenter ? `,datacenter:${datacenter}` : ''}`
 
     let parameters = `ddsource=${source || 'browser'}&ddtags=${encodeURIComponent(tags)}`
 
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -26,17 +26,12 @@ const INTAKE_TRACKS = {
 
 export type EndpointType = keyof typeof ENDPOINTS[IntakeType]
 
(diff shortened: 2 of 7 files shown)


packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -77,7 +77,7 @@ function computeReplicaConfiguration(
   buildEnv: BuildEnv,
   intakeEndpoints: string[]
 ): ReplicaConfiguration | undefined {
-  if (buildEnv.buildMode !== BuildMode.STAGING || initConfiguration.replica === undefined) {
+  if (!initConfiguration.replica) {
```
amortemousque, line 80: After checking on prod I saw that replica is set to "null" therefore I replaced  `initConfiguration.replica === undefined` by `!initConfiguration.replica`.

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -26,6 +26,13 @@ const INTAKE_TRACKS = {
 
 export type EndpointType = keyof typeof ENDPOINTS[IntakeType]
 
+const INTAKE_DATACENTERS: { [site: string]: string } = {
+  'datadoghq.eu': 'eu1.prod.dog',
+  'datadoghq.com': 'us1.prod.dog',
+  'us3.datadoghq.com': 'us3.prod.dog',
+  'us5.datadoghq.com': 'us5.prod.dog',
```
BenoitZugmeyer, line 33: I would be in favor of a `replica.datacenter` Init configuration, so we don't need to maintain a list here. Usage example:

```ts
DD_RUM.init({
  ....
  site: 'datadoghq.eu',
  replica: {
    ....
    datacenter: 'eu1.prod.dog'
  }
})
```

amortemousque: Good point!

packages/core/src/domain/configuration/transportConfiguration.ts
bcaudan, line 80: what about enabling replica only for canary mode?
it should not be used outside of DD applications, right?

bcaudan, line 80: outdated comments:
- https://github.com/DataDog/browser-sdk/blob/f3bab0a29aa3eef0679f023b26c3e3e3d4e2248d/packages/core/src/domain/configuration/transportConfiguration.ts#L12
- https://github.com/DataDog/browser-sdk/blob/f3bab0a29aa3eef0679f023b26c3e3e3d4e2248d/packages/core/src/domain/configuration/configuration.ts#L49

packages/core/src/domain/configuration/transportConfiguration.spec.ts
```diff
@@ -141,11 +141,46 @@ describe('transportConfiguration', () => {
...
+        expect(decodeURIComponent(configuration.replica!.rumEndpointBuilder.build())).toMatch(
+          `&ddtags=(.*),datacenter:${datacenter}`
+        )
+        expect(decodeURIComponent(configuration.replica!.logsEndpointBuilder.build())).toMatch(
+          `&ddtags=(.*),datacenter:${datacenter}`
+        )
+      })
+    })
+
+    it('should not be set as tag for other datacenters in the logs and rum endpoints', () => {
```
bcaudan, line 173: should not we set it for us1.prod.dog in order to have all rum events in us1 tagged with a datacenter?
it seems to be the case for other data sources

.gitlab-ci.yml
```diff
@@ -234,7 +234,7 @@ deploy-staging:
     - .staging
   script:
     - yarn
-    - BUILD_MODE=staging yarn build:bundle
+    - BUILD_MODE=canary yarn build:bundle
     - ./scripts/deploy.sh staging staging
```
bcaudan, line 238: Should we use the `canary` suffix on staging env then?
or is it still useful to differentiate canary and staging files?

amortemousque: Yes it is useful as they are not build at the same time / frequency.
- canary on commit on main
- staging on commit on staging

packages/core/src/domain/configuration/transportConfiguration.ts
amortemousque: Right, 
I was wondering, if the goal is to prevent users from using it:
Not having the option officially documented is it not enough?
What about `enableExperimentalFeatures` then?

bcaudan: I'd say that `enableExperimentalFeatures` is more for temporary experiment rather than long lasting code.
If we only need it on DD app, using a specific build mode has the advantage of not shipping related code to customers.

.gitlab-ci.yml
bcaudan: ok, fair enough

packages/core/src/domain/configuration/transportConfiguration.ts
amortemousque: I checked the consul config for eu, us3 and us5 and we use the stable v3 version on them. 
So, in order to allow the replica on these envs either:
- We remove the buildMode check to keep the stable version
- We add `buildEnv.buildMode !== BuildMode.CANARY` check and set the canary version also on these envs

My personal feeling, is that I prefer to keep the canary version only for US1 to scope our impact in case something went wrong but it's not a strong opinion. Wdyt?

PR #1235, opened 2021-12-23
👷 [RUMF-1107] implement the disallowProtectedDirectoryImport rule 

BenoitZugmeyer:
## Motivation

Make sure we don't import from protected modules

## Changes

Implement an ESLint rule for it, then fix issues that it yields.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     'plugin:import/typescript',
     'prettier',
     'prettier/@typescript-eslint',
+    'plugin:import/typescript',
   ],
   parser: '@typescript-eslint/parser',
   parserOptions: {
@@ -187,6 +188,7 @@ module.exports = {
       rules: {
         '@typescript-eslint/no-unsafe-call': 'off',
         '@typescript-eslint/no-unsafe-return': 'off',
+        '@typescript-eslint/restrict-plus-operands': 'off',
         '@typescript-eslint/restrict-template-expressions': 'off',
       },
     },
@@ -230,6 +232,13 @@ module.exports = {
       rules: {
         // E2E codebase is importing @datadog/browser-* packages referenced by tsconfig.
         'import/no-extraneous-dependencies': 'off',
+      },
+    },
+    {
+      files: ['**/*'],
+      excludedFiles: ['packages/*/test/**', '**/*.spec.ts'],
+      rules: {
+        'local-rules/disallow-protected-directory-import': 'error',
       },
     },
   ],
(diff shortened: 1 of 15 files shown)


.eslintrc.js
```diff
@@ -232,5 +234,12 @@ module.exports = {
         'import/no-extraneous-dependencies': 'off',
       },
     },
+    {
+      files: ['**/*'],
+      excludedFiles: ['packages/*/test/**', '**/*.spec.ts'],
```
bcaudan, line 239: What is the reasoning for not enforcing the protection in spec files?

amortemousque (approved):

.eslintrc.js
```diff
@@ -187,6 +188,7 @@ module.exports = {
       rules: {
         '@typescript-eslint/no-unsafe-call': 'off',
         '@typescript-eslint/no-unsafe-return': 'off',
+        '@typescript-eslint/restrict-plus-operands': 'off',
```
amortemousque, line 191: Why do we need to disable this rule?

BenoitZugmeyer: With this rule enabled, I get the following error:
```
[...]/browser-sdk/eslint-local-rules/disallowProtectedDirectoryImport.js
  85:64  error  Operands of '+' operation must either be both strings or both numbers  @typescript-eslint/restrict-plus-operands
```
This is because TS tries to infer types in JS files, but often it fails to do so and just uses `any` in most places. This rule forbids to use the `+` operator with `any` types. Same thing as what I tried to document [above](https://github.com/DataDog/browser-sdk/pull/1235/files/43449ab679af405fd24d124d36f11ffbec647ca2#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R186),

amortemousque: Got it ;)

packages/rum/test/utils.ts
```diff
@@ -1,4 +1,13 @@
-import { DeflateWorker, DeflateWorkerAction, DeflateWorkerListener } from '../src/domain/segmentCollection'
+import {
+  DeflateWorker,
+  DeflateWorkerAction,
+  DeflateWorkerListener,
+  // To avoid importing too much code in E2E tests, import from deflateWorker directly. This solves
+  // the "Timeout" typing conflict between DOM and @types/node for now. We might find a better
+  // solution in the future.
+  // (cf https://github.com/Microsoft/TypeScript/issues/30128)
+  // eslint-disable-next-line local-rules/disallow-protected-directory-import
```
bcaudan, line 9: Did you consider prefixing `setTimeout` with `window` to disambiguate the conflict?

BenoitZugmeyer: I didn't went with it because this is what we wanted to avoid by doing #708. Another solution would be to use something like:

```ts
type Timeout = ReturnType<typeof setTimeout>
```

and use it instead of `number`. WDYT?

bcaudan: That seems nicer to me 👍

bcaudan (approved):

PR #1242, opened 2022-01-04
⚡️ [RUMF-1030] Decrease BoundedBuffer limitation to 500

amortemousque:
## Motivation

The BoundedBuffer is used to buffer api calls before the SDK initialisation. It stores views, actions, and errors and has a limit of 10000 api calls. In case of big user defined contexts or if the init() is never called, it can have a significant impact.
This PR decrease the limit from 10000 to 500 buffered api calls to lower our memory impact.

## Changes

Decrease bounded buffer limitation to 500

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/boundedBuffer.spec.ts
+++ b/packages/core/src/tools/boundedBuffer.spec.ts
@@ -26,4 +26,17 @@ describe('BoundedBuffer', () => {
     buffered.drain()
     expect(spy.calls.count()).toEqual(5)
   })
+
+  it('store at most 500 callbacks by default', () => {
+    const spy = jasmine.createSpy<() => void>()
+    const buffered = new BoundedBuffer()
+    const limit = 500
+
+    for (let i = 0; i < limit + 1; i += 1) {
+      buffered.add(spy)
+    }
+
+    buffered.drain()
+    expect(spy.calls.count()).toEqual(limit)
+  })
 })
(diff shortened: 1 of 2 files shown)


bcaudan (approved):

packages/core/src/tools/boundedBuffer.spec.ts
```diff
@@ -26,4 +26,17 @@ describe('BoundedBuffer', () => {
     buffered.drain()
     expect(spy.calls.count()).toEqual(5)
   })
+
+  it('store at most 500 callbacks by default', () => {
```
BenoitZugmeyer, line 30: **💭 thought:** ‏this test seems to duplicate the previous one with a different limit. I think we slould keep only one of them

BenoitZugmeyer (approved):

packages/core/src/tools/boundedBuffer.spec.ts
amortemousque: I had the same feeling, but I wanted to test the default value as it's important for performance matter.
Don't you think we could remove the parameter since there is not place in the code where we use it?

BenoitZugmeyer: I added the parameter because the limit was reallly high and testing it with the limit could have make tests somewhat slower. But now that the limit is smaller, I think it's fine to remove the parameter and only test the default limit.

PR #1247, opened 2022-01-05
[REPLAY-564] ♻️ Remove FF - enable visual viewport recording

jagracey:
## Motivation

It's been a while, time to enable visual viewport recording by default. The is a follow up from implementation https://github.com/DataDog/browser-sdk/pull/1089.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/observer.ts
+++ b/packages/rum/src/domain/record/observer.ts
@@ -7,7 +7,6 @@ import {
   addEventListener,
   includes,
   DefaultPrivacyLevel,
-  isExperimentalFeatureEnabled,
   noop,
 } from '@datadog/browser-core'
 import { NodePrivacyLevel } from '../../constants'
@@ -60,10 +59,7 @@ export function initObservers(o: ObserverParam): ListenerHandler {
   const mediaInteractionHandler = initMediaInteractionObserver(o.mediaInteractionCb, o.defaultPrivacyLevel)
   const styleSheetObserver = initStyleSheetObserver(o.styleSheetRuleCb)
   const focusHandler = initFocusObserver(o.focusCb)
-
-  const visualViewportResizeHandler = isExperimentalFeatureEnabled('visualviewport')
-    ? initVisualViewportResizeObserver(o.visualViewportResizeCb)
-    : noop
+  const visualViewportResizeHandler = initVisualViewportResizeObserver(o.visualViewportResizeCb)
 
   return () => {
     mutationHandler()
@@ -99,7 +95,7 @@ function initMoveObserver(cb: MousemoveCallBack): ListenerHandler {
           x: clientX,
           y: clientY,
         }
-        if (isExperimentalFeatureEnabled('visualviewport') && window.visualViewport) {
+        if (window.visualViewport) {
           const { visualViewportX, visualViewportY } = convertMouseEventToLayoutCoordinates(clientX, clientY)
           position.x = visualViewportX
           position.y = visualViewportY
@@ -146,7 +142,7 @@ function initMouseInteractionObserver(
       x: clientX,
       y: clientY,
     }
-    if (isExperimentalFeatureEnabled('visualviewport') && window.visualViewport) {
+    if (window.visualViewport) {
       const { visualViewportX, visualViewportY } = convertMouseEventToLayoutCoordinates(clientX, clientY)
       position.x = visualViewportX
       position.y = visualViewportY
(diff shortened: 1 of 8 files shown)


packages/rum/src/domain/record/viewports.spec.ts
```diff
@@ -14,11 +14,9 @@ describe('layout viewport', () => {
     window.scrollTo(0, 0)
   })
 
-  describe('get window width and height', () => {
+  describe('get window height', () => {
     it('normalized scroll matches native behaviour', () => {
-      const initialInnerWidth = getWindowWidth()
       const initialInnerHeight = getWindowHeight()
-      expect(initialInnerWidth).toBe(window.innerWidth)
       expect(initialInnerHeight).toBe(window.innerHeight)
```
bcaudan, line 20: **❓ question:** ‏no need to test width anymore?

packages/rum/src/domain/record/observer.ts
```diff
@@ -172,7 +168,7 @@ function initScrollObserver(cb: ScrollCallback, defaultPrivacyLevel: DefaultPriv
       }
       const id = getSerializedNodeId(target)
       if (target === document) {
-        if (isExperimentalFeatureEnabled('visualviewport')) {
+        if (window.visualViewport) {
```
bcaudan, line 171: **❓ question:** ‏since this check was not there before, was it a bug or do we not need this check?

jagracey: I added an additional check here since to be safe as I'm not convinced there are zero behavioural affects.

I'm not 100% sure of the various cross browser + cross version affects of:
- `(document.scrollingElement || document.documentElement).scrollLeft` 
vs
- `window.scrollX ?? window.pageXOffset`

**In terms of:**
- inclusion of scrollbar thickness
- measuring visual viewport vs layout
- changes across browser versions

Not a big deal, but I think this should be addressed in a follow up to isolate cause and affect of any replay time rendering differences

packages/rum/test/utils.ts
```diff
@@ -469,3 +469,12 @@ export function createMutationPayloadValidatorFromSegment(segment: Segment) {
     validate: (expected: ExpectedMutationsPayload) => mutationPayloadValidator.validate(mutations[0].data, expected),
   }
 }
+
+/**
+ * Simplify asserting record lengths across multiple devices when not all record types are supported
+ */
+export const addRecordsPerFullSnapshot = (otherRecordsCount: number) => {
```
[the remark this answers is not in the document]
jagracey: Great ideas here. I really like `recordsPerFullSnapshotPlus` for the simplicity.

packages/rum/src/domain/record/viewports.spec.ts
jagracey: Originally thinking it's low value being redundant and it's quite the headache to wrangle the various edge case behaviours that using the Visual Viewport API solves. innerWidth/Height should include the thickness of the sidebar while `visualViewport.width` excludes it- however conformance isn't correct with our device browsers.

But on second glance we should put in some work to describe the issues so it's been re-added with explicit handling of the scrollbar thickness by checking if there is a match with or without scrollbars included - IFF the document is scrollable.

packages/rum/src/domain/record/viewports.spec.ts
```diff
@@ -1,30 +1,57 @@
...
 }
 
 describe('layout viewport', () => {
-  beforeEach(() => {
+  const addVerticalScrollBar = () => {
     document.body.style.setProperty('margin-bottom', '2000px')
-  })
+  }
+  const addHorizontalScrollBar = () => {
+    document.body.style.setProperty('margin-bottom', '2000px')
```
BenoitZugmeyer, line 12: **🔨 warning:** ‏this function is the same as `addVerticalScrollBar`, shouldn't be `margin-left`?


**💬 suggestion:** ‏My screen is 2.5k width, could you use a bigger value?

packages/rum/src/domain/record/observer.ts
jagracey: Got consensus, will change.

jagracey: Got consensus, will change.

bcaudan (approved):
LGTM

BenoitZugmeyer (approved):

PR #1249, opened 2022-01-05
🎨 [RUMF-1118] Enable the @typescript/consistent-type-imports rule

amortemousque:
## Motivation

With TS 3.8.0, let’s enable the consistent-type-imports ESLint rule. Basically, it forces to use import type to import a type. This is mostly for consistency for now, but import type statements could help to make sure some part of the codebase are treeshakable or side-effect free. This rule is auto-fixable, so it won’t change our habits too much.

## Changes

- Replace @typescript-eslint/consistent-type-imports by import/no-duplicates
- Enable @typescript-eslint/consistent-type-imports

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan (approved):
LGTM

urseberry (commented):
What input do you need from the Documentation team? I don't see any user-facing text, but I might have missed something.

Issue #606, opened 2020-11-05
Request buffer system is incompatible with requests that are larger than available memory


eligrey:
I haven't tested to confirm this, but it appears that you buffer all `fetch` and `XMLHttpRequest` requests. This will likely cause memory issues for sites that use streaming requests for resources that are larger than available memory.

I suggest that you switch to a streaming chunked solution. You should refactor all of your `FetchCompleteContext` (and `XhrCompleteContext `?) code to either have a maximum bufferable size (at which point, the Datadog SDK should stop processing & revoke the buffer) or use `ReadableStream` & `TransformStream` with polyfills as needed.

- https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/browser/fetchProxy.ts#L96

BenoitZugmeyer:
Better late than never, this issue has been fixed :) Thank you for your feedback! Let us know if you still see anything weird.

PR #1263, opened 2022-01-11
⚡️[RUMF-1111] compute response text in trackNetworkError 

BenoitZugmeyer:
## Motivation

This is a first PR to address https://github.com/DataDog/browser-sdk/issues/606 . Because we only use the response text when a network error is reported by the Logs SDK, this PR moves the response text retrieval to `trackNetworkError`. This way, the response text won't be collected in the majority of cases.

A future PR will focus on fetching only a limited amount of bytes from the response, by adjusting the `computeResponseData` implementation.

## Changes

Move response text collection into `trackNetworkError`.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/core/test/specHelper.ts
```diff
@@ -151,15 +138,91 @@ export function stubFetch(): FetchStubManager {
...
 
-export interface ResponseStub extends Partial<Response> {
+export interface ResponseStubOptions {
+  status?: number
+  method?: string
+  type?: ResponseType
   responseText?: string
   responseTextError?: Error
 }
+function todo(): never {
```
bcaudan, line 148: **🥜 nitpick:** since it is not really something that we have to do‏
```suggestion
function notYetImplemented(): never {
```

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -46,6 +50,38 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
   }
 }
 
+// TODO: ideally, computeResponseData should always call the `callback` with a string instead of
+// `any`. But to keep retrocompatibility, in the case of XHR with a `responseType` different than
+// "text", the response data should be whatever `xhr.response` is. This is a bit confusing as Logs
+// event 'stack' is expected to be a string. This should be changed in a future major version as it
+// could be a breaking change.
```
bcaudan, line 57: 👍

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -46,6 +50,38 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
...
+      callback(xhr.response)
+    }
+  } else if (response) {
+    response
+      .clone()
+      .text()
+      .then(
+        (text) => callback(text),
+        (error) => callback(`Unable to retrieve response: ${error as string}`)
+      )
```
bcaudan, line 76: **❓ question:** ‏I am not sure if it was working fine with previous async/await version but should not we wrap the callback with a `monitor()`?

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -46,6 +50,38 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
...
+  } else if (error) {
+    callback(truncateResponseText(toStackTraceString(computeStackTrace(error)), configuration))
+  } else {
+    // This case should not happen in practice, but better safe than sorry.
+    callback()
```
bcaudan, line 81: **❓ question:** ‏Could it be related to the 'Failed to load' case? or is it something else?
https://github.com/DataDog/browser-sdk/blob/23b7b2779c44ce1a157d3f1cdcf3f4b5401bc631/packages/logs/src/domain/trackNetworkError.ts#L38

packages/logs/src/domain/trackNetworkError.ts
BenoitZugmeyer: Indeed! I'll wrap it in a `monitor()`.

packages/logs/src/domain/trackNetworkError.ts
BenoitZugmeyer: It can't happen because `XhrCompleteContext | FetchCompleteContext` has always either response, error or xhr defined, but those types alone are not enough to guarantee it. Maybe we should improve typings and have something like:

```ts
interface FetchErrorContext {
  error: Error
...
}

interface FetchCompleteContext {
  response: Response
...
}
```

and then using `XhrCompleteContext | FetchErrorContext | FetchCompleteContext`, so TS can ensure than at least one of the three is define?

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -46,6 +51,38 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
   }
 }
 
+// TODO: ideally, computeResponseData should always call the `callback` with a string instead of
+// `any`. But to keep retrocompatibility, in the case of XHR with a `responseType` different than
```
amortemousque, line 55:
```suggestion
// `any`. But to keep backward compatibility, in the case of XHR with a `responseType` different than
```

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -46,6 +51,38 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
...
+    }
+  } else if (response) {
+    response
+      .clone()
+      .text()
+      .then(
+        monitor((text) => callback(text)),
+        monitor((error) => callback(`Unable to retrieve response: ${error as string}`))
+      )
+  } else if (error) {
```
amortemousque, line 78: I got a bit confused at first on identifying what `response` and `error` cases are compared to `xhr`. As far as I understand `response` and `error` cases are only on fetch requests. 
Maybe better distinguishing when handling `xhr` vs `fetch` could help.

packages/logs/src/domain/trackNetworkError.ts
bcaudan: hum ok, so 'Failed to load' would only be used when responseData is `''`, we should not have undefined value.
Then, yes, it could be nice to avoid this empty callback if it is not needed or add a monitoring log if we have some doubts.

BenoitZugmeyer: Ok to remove the empty callback. The test case still make sense, because as you observed an empty response still needs the "Failed to load" fallback, right?

packages/logs/src/domain/trackNetworkError.ts
BenoitZugmeyer: I split the function in three with more explicit namings

packages/logs/src/domain/trackNetworkError.ts
bcaudan: Yes, but `responseText: undefined` should then be `responseText: ''` and we could clarify the test description.
it could worth a quick chat if I am missing something.

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -46,6 +59,37 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
   }
 }
 
+// TODO: ideally, computeXhrResponseData should always return a string instead of `any`. But to keep
+// backward compatibility, in the case of XHR with a `responseType` different than "text", the
+// response data should be whatever `xhr.response` is. This is a bit confusing as Logs event 'stack'
+// is expected to be a string. This should be changed in a future major version as it could be a
+// breaking change.
+export function computeXhrResponseData(xhr: XMLHttpRequest, configuration: LogsConfiguration): any {
```
bcaudan, line 67: **💬 suggestion:** ‏even if `xhr.response` type is `any`, what about handling it as `unknown` on our side? or reflect what we expect with something like `string | object`? 

and having the same type for `onResponseDataAvailable(responseData: any) {`

packages/logs/src/domain/trackNetworkError.ts
```diff
@@ -24,6 +27,16 @@ export function trackNetworkError(configuration: LogsConfiguration, errorObserva
 
   function handleCompleteRequest(type: RequestType, request: XhrCompleteContext | FetchCompleteContext) {
     if (!configuration.isIntakeUrl(request.url) && (isRejected(request) || isServerError(request))) {
+      if ('xhr' in request) {
+        onResponseDataAvailable(computeXhrResponseData(request.xhr, configuration))
+      } else if (request.response) {
+        computeFetchResponseText(request.response, configuration, onResponseDataAvailable)
+      } else if (request.error) {
+        onResponseDataAvailable(computeFetchErrorText(request.error, configuration))
+      }
```
bcaudan, line 36: **🥜 nitpick:** ‏it could be nice to have a more symmetric API to ease the understanding of the flow, something like:

```suggestion
      if ('xhr' in request) {
        computeXhrResponseData(request.xhr, configuration, onResponseDataAvailable)
      } else if (request.response) {
        computeFetchResponseText(request.response, configuration, onResponseDataAvailable)
      } else if (request.error) {
       computeFetchErrorText(request.error, configuration, onResponseDataAvailable)
      }
```

or by handling promise in `onResponseDataAvailable`:

```suggestion
      onResponseDataAvailable(
        'xhr' in request 
        ? computeXhrResponseData(request.xhr, configuration) 
        : request.response 
          ? computeFetchResponseText(request.response, configuration) 
          : computeFetchErrorText(request.error, configuration)
      )
```

packages/logs/src/domain/trackNetworkError.ts
BenoitZugmeyer: I went with `unknown` because in reality it can be a variety of things: https://developer.mozilla.org/fr/docs/Web/API/XMLHttpRequest/response

bcaudan (approved):

PR #1277, opened 2022-01-19
🔊 [RUMF-1079] audit cookie operations at start of the session

bcaudan:
## Motivation

Try to understand what leads to missing value in session cookie 
Will be reverted after investigation

## Changes

- Audit read/write to session cookie
- Expose audit logs in a global variable to share the between logs/rum
- Add audit logs to session inconsistencies messages

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/session/sessionCookieStore.ts
+++ b/packages/core/src/domain/session/sessionCookieStore.ts
@@ -182,9 +182,9 @@ export class Audit {
 
   private addEntry(phase: string, operation: string, session: SessionState) {
     if (this.entries.length < MAX_AUDIT_ENTRIES) {
-      const currentDate = new Date()
+      const now = new Date()
       // hh:mm:ss.xxx in the browser timezone
-      const formattedTime = `${currentDate.toTimeString().substr(0, 8)}.${currentDate.getMilliseconds()}`
+      const formattedTime = `${now.toTimeString().substr(0, 8)}.${`00${now.getMilliseconds()}`.slice(-3)}`
       this.entries.push(`${formattedTime} - ${this.productKey} ${phase} ${operation} ${toSessionString(session)}`)
     }
   }
--- a/packages/core/src/domain/session/sessionCookieStore.ts
+++ b/packages/core/src/domain/session/sessionCookieStore.ts
@@ -3,6 +3,7 @@ import { getCookie, setCookie } from '../../browser/cookie'
 import * as utils from '../../tools/utils'
 import { isExperimentalFeatureEnabled } from '../configuration'
 import { monitor, addMonitoringMessage } from '../internalMonitoring'
+import { timeStampNow } from '../../tools/timeUtils'
 import type { SessionState } from './sessionStore'
 import { SESSION_EXPIRATION_DELAY } from './sessionStore'
 
@@ -17,6 +18,8 @@ export const MAX_NUMBER_OF_LOCK_RETRIES = 100
 
 type Operations = {
   options: CookieOptions
+  audit?: Audit
+  phase?: string
   process: (cookieSession: SessionState) => SessionState | undefined
   after?: (cookieSession: SessionState) => void
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -155,3 +162,28 @@ function isExpiredState(session: SessionState) {
 function clearSession(options: CookieOptions) {
   setCookie(SESSION_COOKIE_NAME, '', 0, options)
 }
+
+const MAX_AUDIT_ENTRIES = 50
```
bcaudan, line 166: Observed cases seems to happen at the beginning of the session, 50 should be enough to debug with impacting memory

packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -155,3 +162,28 @@ function isExpiredState(session: SessionState) {
...
+
+  addRead(phase: string, session: SessionState) {
+    this.addEntry(phase, 'read', session)
+  }
+
+  private addEntry(phase: string, operation: string, session: SessionState) {
+    if (this.entries.length < MAX_AUDIT_ENTRIES) {
+      const currentDate = new Date(timeStampNow())
+      this.entries.push(
+        `${currentDate.toLocaleTimeString()}.${currentDate.getMilliseconds()} - ${
```
BenoitZugmeyer, line 183: **💬 suggestion:** ‏format the string as ISO or at least non-locale, to avoid looking at translated or unexpected time formats

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -40,13 +47,20 @@ export function startSessionStore<TrackingType extends string>(
   const renewObservable = new Observable<void>()
   const expireObservable = new Observable<void>()
 
+  ;(window as BrowserWindow).DD_SDK_AUDIT = (window as BrowserWindow).DD_SDK_AUDIT || []
```
BenoitZugmeyer, line 50: **💬 suggestion:** ‏I don't think it's a good idea to introduce a new global variable on customer websites. What about using a feature flag?

packages/core/src/domain/session/sessionCookieStore.ts
```diff
@@ -155,3 +162,28 @@ function isExpiredState(session: SessionState) {
...
+    this.addEntry(phase, 'write', session)
+  }
+
+  addRead(phase: string, session: SessionState) {
+    this.addEntry(phase, 'read', session)
+  }
+
+  private addEntry(phase: string, operation: string, session: SessionState) {
+    if (this.entries.length < MAX_AUDIT_ENTRIES) {
+      const currentDate = new Date(timeStampNow())
```
BenoitZugmeyer, line 181: **❓ question:** ‏why do we need `timeStampNow()` here?

```suggestion
      const currentDate = new Date()
```

packages/core/src/domain/session/sessionCookieStore.ts
BenoitZugmeyer: **💬 suggestion:** ‏maybe the formatting can be done only if we find an inconsistency (optimization).

packages/core/src/domain/session/sessionStore.ts
bcaudan: 👍
I'll put it behind `cookie-lock`

packages/core/src/domain/session/sessionCookieStore.ts
bcaudan: 👍
leftover from previous iteration

packages/core/src/domain/session/sessionCookieStore.ts
bcaudan: - used `toTimeString`
- about optim, it'll only be done when FF is enabled so I'd say it is good enough

amortemousque (approved):

BenoitZugmeyer (approved):

PR #1309, opened 2022-01-27
✨ [RUMF-1174] forward event to bridge with rum type

amortemousque:
## Motivation

The mobile SDKs tracks events that happen inside a webview using the browser SDK. Mobile SDK inject a JS bridge to receive event from the browser SDK.
Now, RUM event won't be deserialized on the Mobile side so we don't have to differentiate rum event types in the bridge. We can flag them as "rum".

## Changes

- update rum event bridge
- improve bridge unit tests

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/boot/startLogs.spec.ts
+++ b/packages/logs/src/boot/startLogs.spec.ts
@@ -168,10 +168,9 @@ describe('logs', () => {
       sendLog(DEFAULT_MESSAGE, {})
 
       expect(server.requests.length).toEqual(0)
-      expect(sendSpy).toHaveBeenCalledOnceWith(
-        '{"eventType":"log",' +
-          '"event":{"service":"service","session_id":"session-id","status":"info","message":"message"}}'
-      )
+      const [msg] = sendSpy.calls.mostRecent().args
+      expect(msg).toContain('"eventType":"log"')
+      expect(msg).toContain('"message":"message"')
     })
   })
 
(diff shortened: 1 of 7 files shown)


packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -168,7 +168,10 @@ describe('logs', () => {
       sendLog(DEFAULT_MESSAGE, {})
 
       expect(server.requests.length).toEqual(0)
-      expect(sendSpy).toHaveBeenCalled()
+      expect(sendSpy).toHaveBeenCalledOnceWith(
+        '{"eventType":"log",' +
+          '"event":{"service":"service","session_id":"session-id","status":"info","message":"message"}}'
+      )
```
bcaudan, line 174: **💬 suggestion:** ‏to make those tests less brittle in case the structure or the content of the message changes, it could be interesting to only check what we are interested in, I'd say the `eventType` and something specific to the log/event sent by the test. Something like:

```js
expect(msg).toContain('"eventType":"log"')
expect(msg).toContain('"message":"message"')
```

bcaudan (approved):

packages/logs/src/boot/startLogs.spec.ts
BenoitZugmeyer: I would go one step further by parsing the message and using jasmine assertions on the resulting object

amortemousque: Is there any advantages of parsing the message? 
To me, it seems okay to test a string if the expected message output is a string.

amortemousque: I've done the change

BenoitZugmeyer (approved):

PR #1310, opened 2022-01-28
[RUMF-1174] collect reports and csp violation

amortemousque:
## Motivation

Naive implementation to start collecting reports and csp violation.
Feature under feature flag

## Changes

- implement trackReports
- implement trackCspViolation

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/logs/src/boot/startLogs.ts
```diff
@@ -88,6 +91,29 @@ export function doStartLogs(
...
+    trackReports(['deprecation', 'intervention'], (message: string, report: Report) => {
+      const loggerMethod = report.type === 'deprecation' ? 'warn' : 'error'
+      errorLogger[loggerMethod](message, { report: JSON.parse(JSON.stringify(report)) })
+    })
+    trackCspViolation((message: string, event: SecurityPolicyViolationEvent) => {
+      errorLogger.error(message, {
+        report: {
+          blockedURI: event.blockedURI,
+          columnNumber: event.columnNumber,
+          documentURI: event.blockedURI,
```
bcaudan:
```suggestion
          documentURI: event.documentURI,
```

packages/logs/src/boot/startLogs.ts
```diff
@@ -88,6 +91,29 @@ export function doStartLogs(
   }
   errorObservable.subscribe(reportError)
 
+  if (isExperimentalFeatureEnabled('forward-reports')) {
+    trackReports(['deprecation', 'intervention'], (message: string, report: Report) => {
+      const loggerMethod = report.type === 'deprecation' ? 'warn' : 'error'
+      errorLogger[loggerMethod](message, { report: JSON.parse(JSON.stringify(report)) })
```
bcaudan, line 97: **❓ question:** ‏what is the reasoning for doing the stringify + parse?

packages/logs/src/boot/startLogs.ts
```diff
@@ -88,6 +91,29 @@ export function doStartLogs(
...
+    })
+    trackCspViolation((message: string, event: SecurityPolicyViolationEvent) => {
+      errorLogger.error(message, {
+        report: {
+          blockedURI: event.blockedURI,
+          columnNumber: event.columnNumber,
```
bcaudan, line 103: **❓ question:** what is the reasoning for mapping each attributes individually? could we do `report: event`?‏

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -161,6 +163,10 @@ describe('logs', () => {
     })
 
     it('should send bridge event when bridge is present', () => {
+      if (!isChromium()) {
+        pending('no ReportingObserver support')
+      }
```
bcaudan, line 168: **❓ question:** ‏how is this test related to `ReportingObserver`?

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -170,6 +176,37 @@ describe('logs', () => {
...
+
+      reportingObserverStub.raiseReport()
+
+      expect(logErrorSpy).toHaveBeenCalled()
+
+      resetExperimentalFeatures()
+      reportingObserverStub.reset()
+    })
+
+    it('should not log reports if ff "forward-reports" disabled', () => {
```
bcaudan, line 198: **❓ question:** ‏should not this test be skipped on non chromium browsers?

packages/logs/src/domain/trackReports.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { isChromium } from '@datadog/browser-core'
+import { FAKE_REPORT, stubReportingObserver } from '../test/stubReportingObserver'
+import { trackReports } from './trackReports'
+
+describe('track reports', () => {
```
bcaudan, line 5: **❓ question:** ‏could we have a test for csp as well?

amortemousque: I sure, I wanted first to be sure I was on track.

packages/logs/src/boot/startLogs.ts
amortemousque: I use it to have a raw json object. With the web API `Report` object, our `jsonStringify` function used in the log batch return an empty object. I could have mapped the attributes one by one but for the naive implem I though it was easier. Do you have another idea?

packages/logs/src/boot/startLogs.ts
amortemousque: I tried. With our `jsonStringify` function or the native `JSON.stringify` I only get `{"isTrusted":true}`

packages/logs/src/boot/startLogs.spec.ts
amortemousque: sure

packages/logs/src/boot/startLogs.spec.ts
amortemousque: Bad copy past..

packages/logs/src/boot/startLogs.ts
bcaudan: Interesting, no idea yet about the actual implementation however it could be interesting to reproduce this behavior with the stub.

**💬 suggestion:** ‏On a design perspective, it could be nice to also perform the report transformation inside the tracking function to avoid each trackReports user to do it (same with trackCsp).

packages/logs/src/domain/trackReports.spec.ts
amortemousque: Done

packages/logs/src/boot/startLogs.ts
amortemousque: For the design suggestion, sure. 
Less sure for the `stringify` though. It's not a behaviour of the reports. It is a due to our `jsonStringify` implementation which removes the `toJSON` of the Report prototype.

bcaudan: If this issue comes from our `jsonStringify` implementation, should not we fix it?
`JSON.parse(JSON.stringify(...))` seems really hacky to me.

amortemousque: Reports provide a toJSON method. So I now use it.

packages/logs/src/test/stubCspEventListener.ts
```diff
@@ -0,0 +1,31 @@
+export function stubCspEventListener() {
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const originalAddEventListener = document.addEventListener
```
BenoitZugmeyer, line 3: **💬 suggestion:** ‏you could use `spyOn(document, 'addEventListener').and.callFake(...)`

packages/logs/src/domain/trackReports.ts
```diff
@@ -0,0 +1,105 @@
...
+      effectiveDirective: event.effectiveDirective,
+      lineNumber: event.lineNumber,
+      originalPolicy: event.originalPolicy,
+      referrer: event.referrer,
+      sourceFile: event.sourceFile,
+      statusCode: event.statusCode,
+      violatedDirective: event.violatedDirective,
+    })
+  }
+  return addEventListener(document, DOM_EVENT.SECURITY_POLICY_VIOLATION, handleCspEvent).stop
```
BenoitZugmeyer, line 104: **🥜 nitpick:** ‏by convention, we usually return an object with a `stop` method instead of a "stop function" directly.

packages/logs/src/domain/trackReports.ts
```diff
@@ -0,0 +1,105 @@
...
+export interface ReportingObserverOption {
+  types: ReportType[]
+  buffered: boolean
+}
+
+interface ReportingObserverCallback {
+  (reports: Report[], observer: ReportingObserver): void
+}
+
+declare const ReportingObserver: {
```
BenoitZugmeyer, line 22: **💬 suggestion:** ‏maybe `declare class ReportingObserver {` would be more idiomatic?

```ts
declare class ReportingObserver extends PerformanceObserver {
  static readonly supportedEntryTypes: readonly string[]
  constructor(callback: ReportingObserverCallback, option: ReportingObserverOption)
  disconnect(): void
  observe(): void
  takeRecords(): Report[]
}
```

packages/logs/src/boot/startLogs.ts
```diff
@@ -88,6 +91,16 @@ export function doStartLogs(
   }
   errorObservable.subscribe(reportError)
 
+  if (isExperimentalFeatureEnabled('forward-reports')) {
+    trackReports(['deprecation', 'intervention'], (message: string, report: Report) => {
+      const loggerMethod = report.type === 'deprecation' ? 'warn' : 'error'
+      errorLogger[loggerMethod](message, { report })
```
BenoitZugmeyer, line 97: **💬 suggestion:** ‏
```suggestion
      const status = report.type === 'deprecation' ? StatusType.warn : StatusType.error
      errorLogger.log(message, { report }, status)
```

PR #1314, opened 2022-02-02
🐛 chromium detection did not work in headless mode

bcaudan:
## Motivation

`isChromium` check does not work in unit test because `window.chrome === false` in headless mode 🤦 

## Changes

Check user agent for HeadlessChrome

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/browserDetection.ts
+++ b/packages/core/src/tools/browserDetection.ts
@@ -3,5 +3,5 @@ export function isIE() {
 }
 
 export function isChromium() {
-  return !!(window as any).chrome
+  return !!(window as any).chrome || /HeadlessChrome/.test(window.navigator.userAgent)
 }


amortemousque (approved):

amortemousque (approved):

packages/core/src/tools/browserDetection.ts
```diff
@@ -3,5 +3,5 @@ export function isIE() {
 }
 
 export function isChromium() {
-  return !!(window as any).chrome
```
BenoitZugmeyer, line 6: **❓ question:** ‏if `window.chrome === false` in headless, what about `'chrome' in window`?

bcaudan: my bad it is actually `undefined`

BenoitZugmeyer (approved):

PR #1315, opened 2022-02-04
🐛 [RUMF-1153] fix initial view loading time computation

bcaudan:
## Motivation

Loading time was computed from the start of the activity tracking. In the case of initial view, the loading time should be computed from the start of the view which can be quite different from the start of the tracking, especially for async setup.

## Changes

Compute the activity loading time from the start of the associated parent

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
@@ -138,6 +138,44 @@ describe('rum track view metrics', () => {
       expect(getViewUpdateCount()).toEqual(2)
       expect(getViewUpdate(1).loadingTime).toEqual(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
     })
+
+    it('should use computed loading time from time origin for initial view', () => {
+      // introduce a gap between time origin and tracking start
+      // ensure that `load event > activity delay` and `load event < activity delay + clock gap`
+      // to make the test fail if the clock gap is not correctly taken into account
+      const CLOCK_GAP =
+        FAKE_NAVIGATION_ENTRY_WITH_LOADEVENT_AFTER_ACTIVITY_TIMING.loadEventEnd -
+        BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY +
+        1
+
+      setupBuilder.beforeBuild((buildContext) => {
+        // need to restart the view test to introducing a clock gap before the tracking start
+        viewTest.stop()
+        buildContext.clock.tick(CLOCK_GAP)
+        viewTest = setupViewTest(buildContext)
+        return viewTest
+      })
+
+      const { lifeCycle, domMutationObservable, clock } = setupBuilder.build()
+      const { getViewUpdate, getViewUpdateCount } = viewTest
+
+      expect(getViewUpdateCount()).toEqual(1)
+
+      clock.tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+
+      lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, [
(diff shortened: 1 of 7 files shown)


packages/rum-core/src/domain/waitIdlePage.ts
```diff
@@ -80,7 +81,7 @@ export function doWaitIdlePage(
     const lastChangeTime = timeStampNow()
     if (!isBusy) {
       idleTimeoutId = setTimeout(
-        monitor(() => complete({ hadActivity: true, duration: elapsed(startTime, lastChangeTime) })),
+        monitor(() => complete({ hadActivity: true, duration: elapsed(startClocks.timeStamp, lastChangeTime) })),
```
BenoitZugmeyer, line 84: **💬 suggestion:** ‏Instead of passing the start clocks around, I would change the `IdlePageEvent` so instead of a `duration` it has an `end` timestamp, and use this `end` to compute the final duration wherever `waitIdlePage` is used.

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
```diff
@@ -138,6 +138,44 @@ describe('rum track view metrics', () => {
...
+        FAKE_NAVIGATION_ENTRY_WITH_LOADEVENT_AFTER_ACTIVITY_TIMING.loadEventEnd -
+        BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY +
+        1
+
+      setupBuilder.beforeBuild((buildContext) => {
+        // need to restart the view test to introducing a clock gap before the tracking start
+        viewTest.stop()
+        buildContext.clock.tick(CLOCK_GAP)
+        viewTest = setupViewTest(buildContext)
+        return viewTest
```
BenoitZugmeyer, line 156: **💭 thought:** ‏This feels a bit confusing. What about either:
* always advance the clock before calling `setupViewTest`, as it could reflect a more realistic situation (the SDK will never be initialized at "navigation start"/relative time 0)
* exposing the clock as a `TestSetupBuilder` property, so we can do:
```ts
setupBuilder.clock.tick(CLOCK_GAP)
const { ... } = setupBuilder.build()
```
* using the `mockClock` function directly instead of using `setupBuilder.withFakeClock()`

packages/rum-core/src/domain/waitIdlePage.ts
amortemousque: +1 good suggestion

bcaudan: hum, duration is also used internally to handle cases with max duration:
https://github.com/DataDog/browser-sdk/blob/2ee259897465789c05f404e8a57fa98a4183c73d/packages/rum-core/src/domain/waitIdlePage.ts#L71-L76
wdyt?

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
bcaudan: I went with the second one in [f499364](https://github.com/DataDog/browser-sdk/pull/1315/commits/f49936432d4e81fffc273035661014653be4ebac)

packages/rum-core/src/domain/waitIdlePage.ts
BenoitZugmeyer: What would be the issue with
```ts
monitor(() => complete({ hadActivity: true, end: timeStampNow() })), 
```
?

amortemousque: I don't really see the relation with maxDuration.

bcaudan: nevermind, I get confused, it should be ok

BenoitZugmeyer (approved):

PR #1316, opened 2022-02-04
[RUMF-1176] collect other console logs new

amortemousque:
## Motivation

Collect all console logs.
Under feature flag
## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/tools/observable.ts
```diff
@@ -27,3 +27,14 @@ export class Observable<T> {
     this.observers.forEach((observer) => observer(data))
   }
 }
+
+export function mergeObservables<T>(...observables: Array<Observable<T>>) {
+  const globalObservable = new Observable<T>(() => {
+    const subscriptions: Subscription[] = observables.map((observable) =>
+      observable.subscribe((log) => globalObservable.notify(log))
```
BenoitZugmeyer, line 34:
```suggestion
      observable.subscribe((data) => globalObservable.notify(data))
```

packages/logs/src/boot/startLogs.ts
```diff
@@ -63,30 +64,37 @@ export function doStartLogs(
...
+
+  function reportConsoleLog(log: ConsoleLog) {
+    const messageContext: Partial<LogsEvent> = {}
+    if (log.status === StatusType.error) {
+      messageContext.error = {
+        origin: log.source,
+        stack: log.stack,
+      }
+    }
+    logger.log(log.message, messageContext, log.status)
```
BenoitZugmeyer, line 93: **💬 suggestion:** ‏the message context is not really a LogsEvent... I think it could be simpler as:
```suggestion
    const messageContext
    if (log.status === StatusType.error) {
      messageContext = {
        error: {
          origin: log.source,
          stack: log.stack,
        }
      }
    }
    logger.log(log.message, messageContext, log.status)
```

packages/logs/src/boot/startLogs.ts
```diff
@@ -18,14 +18,14 @@ import type { LogsSessionManager } from '../domain/logsSessionManager'
...
-export function startLogs(configuration: LogsConfiguration, errorLogger: Logger) {
+export function startLogs(configuration: LogsConfiguration, logger: Logger) {
   const internalMonitoring = startInternalMonitoring(configuration)
 
   const errorObservable = new Observable<RawError>()
-
+  const consoleObservable = initConsoleObservable(['log', 'debug', 'info', 'warn', 'error'])
   if (configuration.forwardErrorsToLogs) {
-    trackConsoleError(errorObservable)
     trackRuntimeError(errorObservable)
```
bcaudan, line 29: **❗ warning:** ‏careful about not loosing current behavior

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,96 @@
...
+const LogStatusForApi = {
+  log: StatusType.info,
+  debug: StatusType.debug,
+  info: StatusType.info,
+  warn: StatusType.warn,
+  error: StatusType.error,
+}
+
+export interface ConsoleLog {
+  startClocks: ClocksState
```
bcaudan, line 29: **💭 thought:** ‏it seems weird to me to have a `startClocks` since console calls:

- are punctual events, no difference between start and end
- can't be collected before the SDK init

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -24,11 +24,25 @@ export interface ProvidedError {
...
+      error: {
+        startClocks: consoleError.startClocks,
+        message: consoleError.message,
+        stack: consoleError.stack,
+        source: consoleError.source,
+        handling: ErrorHandling.HANDLED,
+        handlingStack: consoleError.handlingStack,
+      },
+    })
+  )
```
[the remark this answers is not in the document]
amortemousque: I agree, it would be clearer

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,96 @@
...
+  message: string
+  status: StatusType
+  source: 'console'
```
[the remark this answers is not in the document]
amortemousque: The source is also used in Logs to map error.origin. Would you also hardcode it in Logs?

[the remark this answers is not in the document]
amortemousque: Nevermind, I changed it

packages/logs/src/boot/startLogs.ts
BenoitZugmeyer: Makes sense, ok!

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,96 @@
...
+  const observable = new Observable<ConsoleLog>(() => {
+    const originalConsoleApi = console[api]
+
+    console[api] = (...params: unknown[]) => {
+      const handlingStack = createHandlingStack()
+      callMonitored(() => {
+        originalConsoleApi.apply(console, params)
+        observable.notify(buildConsoleLog(params, api, handlingStack))
+      })
+    }
```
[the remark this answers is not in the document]
BenoitZugmeyer: > Won't be easy to use instrumentMethod because createHandlingStack is also used in [here](https://github.com/DataDog/browser-sdk/blob/462592a85a86cb2bfebd6b60808703add65b54a5/packages/rum-core/src/boot/rumPublicApi.ts#L193). It as to be at the same call stack depth everywhere it's called.

Yup! definitly not easy.

> 
> I moved the original console api call outside of callMonitored

Great!

packages/core/src/browser/consoleObservable.ts
```diff
@@ -2,41 +2,29 @@ import { callMonitored } from '../domain/internalMonitoring'
...
-type ApiNameType = 'log' | 'debug' | 'info' | 'warn' | 'error'
-
-const LogStatusForApi = {
-  log: StatusType.info,
-  debug: StatusType.debug,
-  info: StatusType.info,
-  warn: StatusType.warn,
-  error: StatusType.error,
-}
+type ConsoleApiNameType = typeof ConsoleApiName[keyof typeof ConsoleApiName]
```
BenoitZugmeyer, line 15: **💬 suggestion:** ‏It's fine to use an enum here (this is not public facing). It would be a bit simpler.

packages/logs/src/boot/startLogs.ts
amortemousque: Fixed and I added the feature flag

BenoitZugmeyer (approved):
Appart from typechecking issue, LGTM!

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,84 @@
...
+export const CONSOLE_APIS = Object.keys(ConsoleApiName) as ConsoleApiName[]
+
+export interface ConsoleLog {
+  message: string
+  apiName: ConsoleApiName
+  stack?: string
+  handlingStack?: string
+}
+
+const consoleObservables: { [k in ConsoleApiName]?: Observable<ConsoleLog> } = {}
```
bcaudan, line 26: **💬 suggestion:** ‏Can you add a test for the need to have a singleton?

packages/core/src/tools/observable.ts
```diff
@@ -27,3 +27,14 @@ export class Observable<T> {
...
+export function mergeObservables<T>(...observables: Array<Observable<T>>) {
+  const globalObservable = new Observable<T>(() => {
+    const subscriptions: Subscription[] = observables.map((observable) =>
+      observable.subscribe((data) => globalObservable.notify(data))
+    )
+    return () => subscriptions.forEach((subscription) => subscription.unsubscribe())
+  })
+
+  return globalObservable
+}
```
bcaudan, line 40: **💬 suggestion:** ‏what about adding some tests for it?

packages/core/src/tools/utils.ts
```diff
@@ -564,3 +564,9 @@ export function combine(...sources: any[]): unknown {
 }
 
 export type TimeoutId = ReturnType<typeof setTimeout>
+
+export function removeDuplicates<T>(array: T[]) {
+  const deduplicated: T[] = []
+  new Set(array).forEach((item) => deduplicated.push(item))
+  return deduplicated
```
bcaudan, line 571: **💬 suggestion:** ‏
```suggestion
  return Array.from(new Set(array))
```

packages/logs/src/domain/configuration.ts
```diff
@@ -28,11 +39,30 @@ export function validateAndBuildLogsConfiguration(
...
   }
 
+  let forwardConsoleLogs: StatusType[] = []
+  if (isExperimentalFeatureEnabled('forward-logs')) {
+    if (
+      initConfiguration.forwardConsoleLogs !== undefined &&
+      initConfiguration.forwardConsoleLogs !== 'all' &&
+      !Array.isArray(initConfiguration.forwardConsoleLogs)
+    ) {
+      display.error('Forward Console Logs should be an array')
```
bcaudan, line 49: **💬 suggestion:** the value it is not necessarily an array ‏
```suggestion
      display.error('Forward Console Logs should be "all" or an array')
```

packages/logs/src/domain/configuration.ts
```diff
@@ -28,11 +39,30 @@ export function validateAndBuildLogsConfiguration(
     return
   }
 
+  let forwardConsoleLogs: StatusType[] = []
+  if (isExperimentalFeatureEnabled('forward-logs')) {
+    if (
+      initConfiguration.forwardConsoleLogs !== undefined &&
+      initConfiguration.forwardConsoleLogs !== 'all' &&
+      !Array.isArray(initConfiguration.forwardConsoleLogs)
```
bcaudan, line 47: **💬 suggestion:** ‏it could be nice to ensure that only allowed api are instrumented.

packages/logs/src/boot/startLogs.ts
```diff
@@ -63,30 +67,47 @@ export function doStartLogs(
...
+      date: error.startClocks.timeStamp,
+      error: {
+        kind: error.type,
+        origin: error.source,
+        stack: error.stack,
+      },
+    }
+    if (error.resource) {
+      messageContext.http = {
+        method: error.resource.method as any, // TODO: Set method to upper case (it's a breaking change),
```
bcaudan, line 80: **🥜 nitpick:** ‏not sure to get the intent of adding the todo, it does not seem related to the current PR and we already have a task for it

packages/logs/src/boot/startLogs.ts
```diff
@@ -63,30 +67,47 @@ export function doStartLogs(
...
-          : undefined
-      )
-    )
+      }
+    }
+    logger.log(log.message, messageContext, LogStatusForApi[log.apiName])
   }
+
   errorObservable.subscribe(reportError)
+  consoleObservable.subscribe(reportConsoleLog)
```
bcaudan, line 110: **💭 thought:** ‏it may just be me but I was a bit bugged by the fact that console errors don't go through `reportError` since they are ... errors 🤔.
Maybe tweaking the namings to`rawErrorObservable` and `reportRawError` could help.
wdyt?

packages/core/src/tools/utils.ts
amortemousque: Nice ! I though we could not use `Array.from` because of IE.

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,84 @@
...
+  error: 'error',
+} as const
+
+export type ConsoleApiName = typeof ConsoleApiName[keyof typeof ConsoleApiName]
+
+export const CONSOLE_APIS = Object.keys(ConsoleApiName) as ConsoleApiName[]
+
+export interface ConsoleLog {
+  message: string
+  apiName: ConsoleApiName
```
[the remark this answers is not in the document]
amortemousque: Sure

packages/logs/src/domain/configuration.ts
```diff
@@ -28,11 +39,30 @@ export function validateAndBuildLogsConfiguration(
...
+    }
+
+    forwardConsoleLogs =
+      initConfiguration.forwardConsoleLogs === 'all' ? CONSOLE_APIS : initConfiguration.forwardConsoleLogs || []
+  }
+
+  if (initConfiguration.forwardErrorsToLogs) {
+    forwardConsoleLogs.push(ConsoleApiName.error)
+  }
+
```
BenoitZugmeyer, line 60: **💬 suggestion:** you could remove the need for a `removeDuplicates` function by either:

* testing if the value isn't there yet:
```suggestion
  if (initConfiguration.forwardErrorsToLogs && !includes(forwardConsoleLogs, ConsoleApiName.error)) {
    forwardConsoleLogs.push(ConsoleApiName.error)
  }
```

* using a `Set` for `forwardConsoleLogs` value (also use a `Set` as parameter to `initConsoleObservable`)

BenoitZugmeyer: Actually, it seems like `new Set(arg)` is not supported by IE11 (according to https://kangax.github.io/compat-table/es6/#test-Set ) so maybe use the first solution

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const ConsoleApiName = {
+  log: 'log',
+  debug: 'debug',
+  info: 'info',
+  warn: 'warn',
+  error: 'error',
+} as const
+
+export type ConsoleApiName = typeof ConsoleApiName[keyof typeof ConsoleApiName]
```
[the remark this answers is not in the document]
amortemousque: `ConsoleApiName` is used in LogsInitConfiguration which is a user facing type. Isn't it an issue?

packages/logs/src/domain/configuration.ts
amortemousque: The first solution does not cover if a user send `forwardConsoleLogs: ['error', 'error']`. Or maybe I'm missing something?

packages/logs/src/boot/startLogs.ts
amortemousque: The todo is here to explain why there is a cast. I can replace it by an other comment like:
`// Cast resource method because case mismatch cf issue RUMF-1152`
or let the cast unexplained?

packages/logs/src/domain/configuration.ts
BenoitZugmeyer: Ah right.

packages/core/src/browser/consoleObservable.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Ah that's right, nevermind

packages/logs/src/boot/startLogs.ts
bcaudan: oh, I missed the introduction of the cast compared to previous version.
your new comment seems clearer to me 👍

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const CONSOLE_APIS = Object.keys(ConsoleApiName) as ConsoleApiName[]
```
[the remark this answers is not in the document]
amortemousque: Yes

packages/logs/src/boot/startLogs.ts
amortemousque: I see. I'll update the names. I keep this comment in mind for log assembly improvement task.

packages/core/src/browser/consoleObservable.ts
```diff
@@ -0,0 +1,84 @@
...
+function buildConsoleLog(params: unknown[], apiName: ConsoleApiName, handlingStack: string): ConsoleLog {
+  const log: ConsoleLog = {
+    message: [`console ${apiName}:`, ...params].map((param) => formatConsoleParameters(param)).join(' '),
+    apiName,
+  }
+
+  if (apiName === ConsoleApiName.error) {
+    const firstErrorParam = find(params, (param: unknown): param is Error => param instanceof Error)
+    log.stack = firstErrorParam ? toStackTraceString(computeStackTrace(firstErrorParam)) : undefined
+    log.handlingStack = handlingStack
```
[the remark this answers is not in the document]
amortemousque: yes, done

packages/core/src/browser/consoleObservable.ts
amortemousque: Isn't the "should allow multiple caller" test enough?

bcaudan: 🤔 removing the singleton does not fail the test, so something is not right

packages/core/src/tools/observable.ts
amortemousque: Sure

packages/core/src/browser/consoleObservable.ts
amortemousque: Its because the observable references of `consoleObservablesByApi` live inside the merged observable.
I've updated the test to ensure the console api as been instrumented only once.
https://github.com/DataDog/browser-sdk/blob/18b2dbe7a138f0dbf50abc4ee3115801c6ac514e/packages/core/src/domain/console/consoleObservable.spec.ts#L65

packages/logs/src/boot/startLogs.ts
```diff
@@ -62,31 +66,48 @@ export function doStartLogs(
...
+    logger.error(error.message, messageContext)
+  }
+
+  const LogStatusForApi = {
+    [ConsoleApiName.log]: StatusType.info,
+    [ConsoleApiName.debug]: StatusType.debug,
+    [ConsoleApiName.info]: StatusType.info,
+    [ConsoleApiName.warn]: StatusType.warn,
+    [ConsoleApiName.error]: StatusType.error,
+  }
```
bcaudan, line 94: **🥜 nitpick:** ‏we could move this configuration constant at the top of the file, it would reduce a bit the length of this method

packages/logs/src/domain/configuration.ts
```diff
@@ -28,11 +39,35 @@ export function validateAndBuildLogsConfiguration(
...
+  if (isExperimentalFeatureEnabled('forward-logs') && initConfiguration.forwardConsoleLogs !== undefined) {
+    if (
+      (initConfiguration.forwardConsoleLogs !== 'all' && !Array.isArray(initConfiguration.forwardConsoleLogs)) ||
+      (Array.isArray(initConfiguration.forwardConsoleLogs) &&
+        initConfiguration.forwardConsoleLogs.some((api) => !includes(allowedConsoleApis, api)))
+    ) {
+      display.error(
+        `Forward Console Logs should be "all" or an array with allowed values ${allowedConsoleApis
+          .map((api) => `"${api}"`)
+          .join(', ')}`
```
bcaudan, line 53: **🥜 nitpick:** to avoid an extra traversal of the array‏
```suggestion
        `Forward Console Logs should be "all" or an array with allowed values "${allowedConsoleApis.join('", "')}"`
```

packages/logs/src/domain/configuration.spec.ts
```diff
@@ -22,4 +23,72 @@ describe('validateAndBuildLogsConfiguration', () => {
...
+          .forwardConsoleLogs
+      ).toEqual(['error'])
+    })
+
+    it('contains all apis when forwardConsoleLogs is set to "all"', () => {
+      expect(
+        validateAndBuildLogsConfiguration({ ...DEFAULT_INIT_CONFIGURATION, forwardConsoleLogs: 'all' })!
+          .forwardConsoleLogs
+      ).toEqual(Object.keys(ConsoleApiName) as ConsoleApiName[])
+    })
```
bcaudan, line 76: **💬 suggestion:** it could be nice to have a test where both options are enabled, to ensure that `error` is not tracked twice (`removeDuplicates`)‏

bcaudan (approved):

packages/logs/src/domain/configuration.spec.ts
amortemousque: Sure.

packages/logs/src/domain/configuration.ts
```diff
@@ -40,23 +39,24 @@ export function validateAndBuildLogsConfiguration(
...
   }
 
+  const allowedConsoleApis = Object.keys(ConsoleApiName) as ConsoleApiName[]
   let forwardConsoleLogs: StatusType[] = []
   if (isExperimentalFeatureEnabled('forward-logs') && initConfiguration.forwardConsoleLogs !== undefined) {
     if (
       (initConfiguration.forwardConsoleLogs !== 'all' && !Array.isArray(initConfiguration.forwardConsoleLogs)) ||
       (Array.isArray(initConfiguration.forwardConsoleLogs) &&
-        initConfiguration.forwardConsoleLogs.some((api) => !includes(CONSOLE_APIS, api)))
+        initConfiguration.forwardConsoleLogs.some((api) => !includes(allowedConsoleApis, api)))
```
BenoitZugmeyer, line 48: **💬 suggestion:** ‏you could use `objectHasValue` to check this, see [example here](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/configuration.ts#L84)

packages/logs/src/domain/configuration.ts
```diff
@@ -28,11 +39,33 @@ export function validateAndBuildLogsConfiguration(
...
   }
 
+  const allowedConsoleApis = Object.keys(ConsoleApiName) as ConsoleApiName[]
+  let forwardConsoleLogs: StatusType[] = []
+  if (isExperimentalFeatureEnabled('forward-logs') && initConfiguration.forwardConsoleLogs !== undefined) {
+    if (
+      (initConfiguration.forwardConsoleLogs !== 'all' && !Array.isArray(initConfiguration.forwardConsoleLogs)) ||
+      (Array.isArray(initConfiguration.forwardConsoleLogs) &&
+        initConfiguration.forwardConsoleLogs.some((api) => !includes(allowedConsoleApis, api)))
+    ) {
```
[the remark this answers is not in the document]
amortemousque: yep

PR #1317, opened 2022-02-04
[RUMF-1141] implement a POC frustration signal tracking

BenoitZugmeyer:
## Motivation

Explore how we could collect signals.

## Changes

Track frustration signals behind the `frustration-signals` feature flag. The `frustration-signals-debug` flag can be enabled to print debug values in the console, and will be used along a playground page to help anyone willing to give it a try understand what's going on. 

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/action/trackFrustrationSignals.ts
```diff
@@ -0,0 +1,243 @@
...
+    activeElement.selectionStart !== activeElement.selectionEnd
+  ) {
+    return true
+  }
+
+  const selection = window.getSelection()!
+  return !!selection && !selection.isCollapsed
+}
+
+function collectSignals(clicks: Click[]): FrustrationSignal[] {
```
amortemousque, line 150: I know that it's a POC, but to me, it's a bit hard to understand what about rage, dead or error. 
Is it possible to separate the logic by signal type.

BenoitZugmeyer: Well I could separate the logic a bit but it will still be intricate, because there is some kind of precendence between signals:
* we have to wait to make sure a click is not part of a "rage click" sequence
* if it's not, the "error click" takes precedence
* then the "dead click" is used as a fallback on certain condition

amortemousque:
From my tests in the playground here are my observations:

Suggestions:
- When selecting a text, If I by mistake I click 4 or 5 times instead of 3, it could be nice not to consider it as a rage click.
- To add value to dead clicks, it could be nice to get the closest element the user tried to click. Might be complicated though.
- When rage clicking on the error button, it could be nice to report a rage click and an error click. Or to find a way to say "I'm rage clicking because of an error"

Issues:
- When resizing the textarea, I get a dead click. 
- When I change the pointer position in the textarea, I get a dead click.

packages/rum-core/src/domain/rumEventsCollection/action/trackFrustrationSignals.ts
bcaudan: Depending on how we want to integrate that with automatic click actions, it could be even more intricate.
It could be nice if we can make that easier to follow.

PR #1332, opened 2022-02-11
✨ [RUMF-1175] collect reports and csp violation

amortemousque:
## Motivation

Collect reports and csp violation.
Feature under feature flag

Worth reviewing commit by commit 

## Changes

- add a reportObserbvable 
- collect error reports in RUM
- collect all reports in Logs

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,152 @@
...
+
+interface InterventionReportBody {
+  id: string
+  message: string
+  lineNumber: number
+  columnNumber: number
+  sourceFile: string
+}
+
+export const ReportType = {
```
amortemousque, line 48: Can't be an enum because, it will be exposed with a new init config option

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.ts
```diff
@@ -0,0 +1,21 @@
...
+import { clocksNow, ErrorHandling, ErrorSource, initReportObservable, ReportType } from '@datadog/browser-core'
+
+export function trackReportError(errorObservable: Observable<RawError>) {
+  const subscription = initReportObservable([ReportType.csp_violation, ReportType.intervention]).subscribe(
+    (reportError) =>
+      errorObservable.notify({
+        startClocks: clocksNow(),
+        message: reportError.message,
+        stack: reportError.stack,
+        source: ErrorSource.REPORT,
```
amortemousque, line 11: The rum events format will need to be updated

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export type ReportType = typeof ReportType[keyof typeof ReportType]
+
+export interface Report {
+  type: ReportType
+  message: string
+  stack?: string
+}
+
+export type BrowserReportType = 'intervention' | 'deprecation'
```
bcaudan, line 63: **💭 thought:** ‏It is a bit hard to follow what is Browser types and what is our types. 
It could be interesting to extract the browser types to a dedicated file with the actual browser names somewhere.

amortemousque: I've put browser types into report/browser.type.ts

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,119 @@
...
+import type {
+  Report,
+  BrowserWindow,
+  ReportType,
+  ReportingObserverCallback,
+  ReportingObserverOption,
+  ReportingObserver as ReportingObserverInterface,
+} from './browser.types'
+
+declare const ReportingObserver: {
```
amortemousque, line 13: I tried to move the ReportingObserver type declaration in the `browser.types` without success. I run into several typescript issues. This is the best solution I could found.

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,119 @@
...
+  new (callback: ReportingObserverCallback, option: ReportingObserverOption): ReportingObserverInterface
```
bcaudan, line 14: **💬 suggestion:** could you:

- move the constructor type in the existing ReportingObserver interface
- ensure to use ReportingObservable variable from the BrowserWindow type
- remove the declare const

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export const CustomReportType = {
```
bcaudan, line 17: **❓ question:** ‏what do you want to express with "custom"?

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,119 @@
...
+  intervention: 'intervention',
+  deprecation: 'deprecation',
+  csp_violation: 'csp_violation',
```
bcaudan, line 20: **💭 thought:** ‏we don't have a rule for that but this case seems odd compared to what we currently do.
I would have expected either:

- `CSP_VIOLATION: 'csp_violation'`
- `cspViolation: 'csp_violation'`

packages/core/src/domain/report/reportObservable.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    reportingObserverStub = stubReportingObserver()
+    cspEventListenerStub = stubCspEventListener()
+    notifyReport = jasmine.createSpy('notifyReport')
+  })
+
+  afterEach(() => {
+    reportingObserverStub.reset()
+    consoleSubscription.unsubscribe()
+  })
+  ;[CustomReportType.deprecation, CustomReportType.deprecation].forEach((type) => {
```
BenoitZugmeyer, line 26: **🔨 warning:** ‏looks like it defines the same test twice

bcaudan, line 26: **🥜 nitpick:** ‏
```suggestion
  })
  
  ;[CustomReportType.deprecation, CustomReportType.deprecation].forEach((type) => {
```

bcaudan, line 26: **🥜 nitpick:** ‏
```suggestion
  })
  
  ;[CustomReportType.deprecation, CustomReportType.deprecation].forEach((type) => {
```

packages/core/src/domain/report/reportObservable.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+describe(`report observable`, () => {
+  let reportingObserverStub: { reset(): void; raiseReport(type: string): void }
+  let cspEventListenerStub: { dispatchEvent(): void }
+  let consoleSubscription: Subscription
+  let notifyReport: jasmine.Spy
+
+  beforeEach(() => {
+    if (!isChromium()) {
+      pending('no ReportingObserver support')
+    }
```
BenoitZugmeyer, line 15: **❓ question:** Since we are stubing browser APIs for those tests, do we need to worry about whether the API is supported or not? I think it could be run in all browsers, no?

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,119 @@
...
+  sourceFile: string | null,
+  lineNumber: number | null,
+  columnNumber: number | null
+): string | undefined {
+  if (!sourceFile) {
+    return undefined
+  }
+
+  return `${type}: ${message}
+at <anonymous> @ ${[sourceFile, lineNumber, columnNumber].filter((e) => e).join(':')}`
```
BenoitZugmeyer, line 118: **💬 suggestion:** ‏maybe use [`toStackTraceString()`](https://github.com/DataDog/browser-sdk/blob/5f884b7b23de943d4b70b3ca192b4dfd455db8ca/packages/core/src/tools/error.ts#L63) here

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    clock = mockClock()
+  })
+
+  afterEach(() => {
+    subscription.unsubscribe()
+    clock.cleanup()
+    reportingObserverStub.reset()
+  })
+
+  it('should error report', () => {
```
bcaudan, line 29: **❓ question:** ‏should it not be enable only behind the FF?

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.ts
bcaudan: Should not it be done before this PR?

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -202,6 +208,24 @@ describe('logs', () => {
...
     })
+
+    it('should send reports', () => {
+      if (!isChromium()) {
+        pending('no ReportingObserver support')
+      }
+      const logger = new Logger(noop)
+      const logErrorSpy = spyOn(logger, 'log')
+      const reportingObserverStub = stubReportingObserver()
+      updateExperimentalFeatures(['forward-reports'])
```
bcaudan, line 219: **💬 suggestion:** ‏what about adding a test without the FF to ensure that it also behave correctly?

packages/core/test/specHelper.ts
```diff
@@ -417,3 +424,80 @@ export function stubCookie() {
     },
   }
 }
+
+export function stubReportingObserver() {
```
BenoitZugmeyer, line 428: **💭 thought:** Maybe we'll need to split this file one day, it's beginning to be a bit large

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -202,6 +208,24 @@ describe('logs', () => {
...
+      originalStartLogs({ ...baseConfiguration, forwardReports: ['intervention'] }, logger)
```
bcaudan, line 220: **💬 suggestion:**  it could be interesting to add a test for the expected default behavior as well‏

packages/core/src/domain/report/reportObservable.ts
BenoitZugmeyer, line 17: **❓ question:** ‏I wonder, why is it called `CustomReportType` and not simply `ReportType`?

amortemousque: I just don't want to conflict with native types that lives in `browser.types.ts`.

amortemousque: Same comment as https://github.com/DataDog/browser-sdk/pull/1332#discussion_r816747785

packages/core/test/specHelper.ts
amortemousque: I agree

packages/core/src/domain/report/reportObservable.spec.ts
amortemousque: Yes, why not :)

packages/core/src/domain/report/reportObservable.ts
amortemousque: I see, let go with the` cspViolation: 'csp_violation'`

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.spec.ts
amortemousque: Of course

packages/logs/src/boot/startLogs.spec.ts
amortemousque: By the expected default behaviour, you mean without the FF?

packages/logs/src/boot/startLogs.spec.ts
amortemousque: Sure

packages/core/src/domain/report/reportObservable.spec.ts
amortemousque: This is the prettier issue reported here: https://github.com/DataDog/browser-sdk/blob/ae205749e656fb477c870625b075a6c341d09250/packages/core/src/domain/console/consoleObservable.spec.ts#L7-L8

packages/core/src/domain/report/reportObservable.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { isChromium } from '../../tools/browserDetection'
+import type { Subscription } from '../../tools/observable'
+import { stubReportingObserver, stubCspEventListener } from '../../../test/specHelper'
+import { initReportObservable, CustomReportType } from './reportObservable'
+
+describe(`report observable`, () => {
```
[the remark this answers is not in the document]
amortemousque: I prepare the PR to enable the TS rule :)

packages/core/src/domain/report/browser.types.ts
```diff
@@ -0,0 +1,43 @@
+export interface BrowserWindow {
+  ReportingObserver?: ReportingObserver
```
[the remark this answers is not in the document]
amortemousque: Yes sure

packages/core/src/domain/report/reportObservable.ts
amortemousque: If I declare the ReportingObserver in browser.types like so:
```typescript
export interface BrowserWindow {
  ReportingObserver: ReportingObserver
}

declare const ReportingObserver: {
  prototype: PerformanceEntry
  new (callback: ReportingObserverCallback, option: ReportingObserverOption): ReportingObserver
}

interface ReportingObserver {
  disconnect(): void
  observe(): void
  takeRecords(): Report[]
}
```

TS complains when I try use ReportingObservable variable from the BrowserWindow type:
```typescript
const test = new (window as BrowserWindow).ReportingObserver()
```
TS error message:
`This expression is not constructable. Type 'ReportingObserver' has no construct signatures.`

amortemousque: Same if I declare a class like so:
```typescript
declare class ReportingObserver { 
  constructor(callback: ReportingObserverCallback, option: ReportingObserverOption) 
  disconnect(): void 
  observe(): void 
  takeRecords(): Report[] 
}
```

packages/core/src/domain/report/reportObservable.ts
amortemousque: I can also use an alias for the native types that I import. Wdyt?

BenoitZugmeyer: To avoid conflict, we usually prefix those names with `Rum`: `RumReportType`. WDYT?

packages/core/src/domain/report/reportObservable.ts
bcaudan: what I had in mind: https://github.com/DataDog/browser-sdk/commit/4c522570562a01b4271facaf84873338da6084d0

packages/core/src/domain/report/reportObservable.ts
amortemousque: I thought that we use `RUM` and `Logs` prefixes when the code is specific to one package like:
`Configuration`, `RumConfiguration`, `LogsConfiguration`

I think `RumReportType` will be misleading. Wdyt?

bcaudan: Agreed, we use `Raw` prefix at several places, could it fit here?

BenoitZugmeyer: Ah right, it is used in Logs too. `Raw` seems fine then. In my mind, `Custom` indicates that it is somehow customer related.

packages/core/src/domain/report/reportObservable.spec.ts
bcaudan: nevermind then

packages/core/src/domain/report/reportObservable.ts
amortemousque: Way nicer indeed!

packages/logs/src/boot/startLogs.spec.ts
amortemousque: Okay, done

packages/core/src/domain/report/reportObservable.ts
amortemousque: Okay, I'm good with Raw ;)

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  afterEach(() => {
+    subscription.unsubscribe()
+    clock.cleanup()
+    reportingObserverStub.reset()
+    resetExperimentalFeatures()
+  })
+
+  it('should error report when ff forward-reports enabled', () => {
```
bcaudan, line 37:
```suggestion
  it('should report when ff forward-reports enabled', () => {
```

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const subscription = initReportObservable([RawReportType.cspViolation, RawReportType.intervention]).subscribe(
+    (reportError) =>
+      errorObservable.notify({
+        startClocks: clocksNow(),
+        message: reportError.message,
+        stack: reportError.stack,
+        type: reportError.id,
+        source: ErrorSource.REPORT,
+        handling: ErrorHandling.HANDLED,
```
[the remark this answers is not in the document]
amortemousque: Indeed logical

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,118 @@
...
+    stack: buildStack(body.id, body.message, body.sourceFile, body.lineNumber, body.columnNumber),
+  }
+}
+
+function buildRawReportFromCspViolation(event: SecurityPolicyViolationEvent): RawReport {
+  const type = RawReportType.cspViolation
+  const message = `'${event.blockedURI}' blocked by '${event.effectiveDirective}' directive`
+  return {
+    type,
+    id: type,
```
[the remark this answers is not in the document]
amortemousque: Yes it could fit as a subtype.

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -0,0 +1,118 @@
...
+    type,
+    subtype: body.id,
+    message: `${type}: ${body.message}`,
+    stack: buildStack(body.id, body.message, body.sourceFile, body.lineNumber, body.columnNumber),
+  }
+}
+
+function buildRawReportFromCspViolation(event: SecurityPolicyViolationEvent): RawReport {
+  const type = RawReportType.cspViolation
+  const message = `'${event.blockedURI}' blocked by '${event.effectiveDirective}' directive`
```
amortemousque, line 87: After some thought, I think it is valuable to add the `originalPolicy` to the crafted message. Could be done like so:
`${event.blockedURI}' blocked by '${event.effectiveDirective}' directive of the policy "${event.originalPolicy}"`.

amortemousque: @bcaudan @BenoitZugmeyer

bcaudan: Some directives could list a large number of domains 🤔
I'd say that it could be added in the stack to keep a minimal message.
We should maybe also truncate the policy to avoid sending too much data.

amortemousque: I added it to the stack. Let me know what you think

BenoitZugmeyer: LGTM!

packages/rum-core/src/domain/rumEventsCollection/error/trackReportError.ts
amortemousque: Done

bcaudan (approved):

PR #1333, opened 2022-02-11
[RUMF-1103] Remove event bridge FF

amortemousque:
## Motivation

Now that the event bridge implementation is stable, remove the feature flag

## Changes

Remove event-bridge FF

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -268,7 +268,12 @@ export function makeRumPublicApi(
   }
 
   function overrideInitConfigurationForBridge<C extends InitConfiguration>(initConfiguration: C): C {
-    return { ...initConfiguration, applicationId: 'empty', clientToken: 'empty', sampleRate: 100 }
+    return {
+      ...initConfiguration,
+      applicationId: '00000000-aaaa-0000-aaaa-000000000000',
```
amortemousque, line 273: Need a valid application id for the e2e format test. The issue wasn't there with the FF because the `initRum` => `canUseEventBridge` => `isExperimentalFeatureEnabled('event-bridge')` always returned false. It wasn't really problematic though because it was only to override some init config options.

BenoitZugmeyer: **🥜 nitpick:** ‏Maybe change the e2e test instead?

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: I still prefer our event to match the RUM event format, as it serve as a doc to communicate with others
Wdyt?

BenoitZugmeyer: Ok, makes sense!

PR #1335, opened 2022-02-11
[RUMF-1169] cleanup compute stack trace

BenoitZugmeyer:
## Motivation

Reduce bundles size by removing old browser support in tracekit imported code. Those browsers are not supported by our SDK in the first place (the code won't execute at all), so it won't have any impact.

## Changes

Remove various strategies to compute the call stack. I made sure that removed code cannot be used in supported browsers:

* old Opera (presto-based) browsers don't support WeakMaps, which is [a requirement](https://github.com/DataDog/browser-sdk/blob/ef19508a61c2be66b07561986bd6233fe76c0ba0/packages/core/src/browser/xhrObservable.ts#L31) for our SDK

* the IE11 errors may not have a `stack` (they need to be thrown to have a stack), but the fallback strategy to produce a stack by walking the caller stack wasn't working because our SDK is evaluated in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). Also, this strategy would probably return something similar to the "handling stack" instead of the actual error stack.

Before:
```
du -b packages/*/bundle/*.js
35349	packages/logs/bundle/datadog-logs.js
66922	packages/rum-slim/bundle/datadog-rum-slim.js
116266	packages/rum/bundle/datadog-rum.js
```

After
```
du -b packages/*/bundle/*.js
32539	packages/logs/bundle/datadog-logs.js
64113	packages/rum-slim/bundle/datadog-rum-slim.js
113457	packages/rum/bundle/datadog-rum.js
```

8.6% size reduction for logs bundle 🎉 

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -1,556 +1,128 @@
...
-      }
-    } else {
-      continue
-    }
-
-    if (!element.func && element.line) {
-      element.func = UNKNOWN_FUNCTION
-    }
-    stack.push(element)
+export function computeStackTraceOfCaller(): StackTrace {
```
bcaudan, line 39: **❓ question:** ‏it seems to only be used by the test, do we need to keep it?

packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -1,556 +1,128 @@
-import type { StackFrame, StackTrace } from './types'
+import type { StackTrace, StackFrame } from './types'
 
 const UNKNOWN_FUNCTION = '?'
 
 /**
  * computeStackTrace: cross-browser stack traces in JavaScript
- *
```
bcaudan, line 7: **💬 suggestion:** ‏what about removing this comment too?

bcaudan (approved):

amortemousque (approved):

PR #1337, opened 2022-02-14
⚡️ Process buffered performance entries in an idle callback

amortemousque:
## Motivation

Process buffered resources in an idle callback could prevent the SDK from blocking the main thread on page load too much.

## Changes

Update performance collection

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -154,6 +161,17 @@ export function startPerformanceCollection(lifeCycle: LifeCycle, configuration:
...
 
+function waitForRequestIdleOrUnload(lifeCycle: LifeCycle, callback: () => void) {
+  let hasBeenCalled = false
+  const callOnce = () => {
+    if (hasBeenCalled) return
+    hasBeenCalled = true
+    callback()
+  }
+  requestIdleCallback(callOnce)
+  lifeCycle.subscribe(LifeCycleEventType.BEFORE_UNLOAD, callOnce)
```
bcaudan, line 172: **❓ question:** ‏what is the reasoning for calling it on unload?

amortemousque: I don't know the detail behind the idle period detection but, FMU we can have cases where they are no idle periods before the unload. This is why the requestIdleCallback API provides a timeout option. 

Btw the documentation states: "A timeout option is strongly recommended for required work, as otherwise it's possible multiple seconds will elapse before the callback is fired." but in our case BEFORE_UNLOAD seems more appropriate.

bcaudan: IMO, these kind of behaviors should be explained in the code and tested. 
I am wondering if this processing is not likely to be aborted by the unload anyway.

On the other hand, if we are worried about the requestIdleCallback not being called due to too much activity, let's just use a setTimeout, it should at least split this processing from the init.

wdyt?

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -154,6 +161,17 @@ export function startPerformanceCollection(lifeCycle: LifeCycle, configuration:
...
 }
 
+function waitForRequestIdleOrUnload(lifeCycle: LifeCycle, callback: () => void) {
+  let hasBeenCalled = false
+  const callOnce = () => {
+    if (hasBeenCalled) return
+    hasBeenCalled = true
+    callback()
+  }
+  requestIdleCallback(callOnce)
```
BenoitZugmeyer, line 171: **💭 thought:** ‏maybe we should add some logic to make sure they are sent before a new view is created?

amortemousque: I suppose your fear is that the resources are being associated to the wrong view? 
The only case I can think of, would be to have a location change before the setTimeout callback is called.
I guess that could happen. 
Considering our last discussion, to me it feels a bit of the same concern that the BEFORE_UNLOAD, meaning: 
it's very unlikely to happen but can technically happen.

Here is the list of our options:
- ignore these cases
- consider all cases:  view update, before_unload, visibility change
- only consider view update
- drop the optimisation

My personal opinion would be in favour of either ignoring these cases or drop the optimisation
Wdyt?

packages/rum-core/src/browser/performanceCollection.ts
amortemousque: Other concern about the async call:
https://github.com/DataDog/browser-sdk/pull/1337#discussion_r808080557

packages/core/src/tools/utils.ts
```diff
@@ -564,3 +564,23 @@ export function combine(...sources: any[]): unknown {
 }
 
 export type TimeoutId = ReturnType<typeof setTimeout>
+
+export function requestIdleCallback(callback: () => void, opts?: { timeout?: number }) {
```
amortemousque, line 568: I think we could still let this function here as it could be valuable at some point in other packages but let me know.

BenoitZugmeyer (approved):

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -109,8 +109,12 @@ export function startPerformanceCollection(lifeCycle: LifeCycle, configuration:
   })
 
   if (supportPerformanceObject()) {
-    handleRumPerformanceEntries(lifeCycle, configuration, performance.getEntries())
+    const performanceEntries = performance.getEntries()
+    // Because the performance entry list can be quite large
+    // delay the computation to prevent the SDK from blocking the main thread on init
+    setTimeout(() => handleRumPerformanceEntries(lifeCycle, configuration, performanceEntries))
```
bcaudan, line 115: **❗ warning:** ‏we should monitor this function

amortemousque: erf, of course. Done

bcaudan (approved):

PR #1342, opened 2022-02-15
✨ [RUMF-1177] add index_in_view to segment meta 

BenoitZugmeyer:
## Motivation

Helps debugging replay issues

## Changes

Add a `segment.meta.index_in_view` field to the segment structure. This is also sent to the intake as a `index_in_view` FormData field.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/rum/src/domain/segmentCollection/segment.ts
```diff
@@ -9,27 +9,30 @@ let nextId = 0
 export class Segment {
   public isFlushed = false
   public flushReason?: string
+  public readonly meta: SegmentMeta
```
amortemousque, line 12: Does `meta` means `metadata`?

BenoitZugmeyer: Yes

packages/rum/src/domain/replayStats.ts
```diff
@@ -3,6 +3,10 @@ import type { ReplayStats } from '@datadog/browser-rum-core'
 export const MAX_STATS_HISTORY = 10
 let statsPerView: Map<string, ReplayStats> | undefined
 
+export function getSegmentCount(viewId: string) {
+  return getOrCreateReplayStats(viewId).segments_count
```
bcaudan, line 7:
```suggestion
export function getSegmentsCount(viewId: string) {
  return getOrCreateReplayStats(viewId).segments_count
```

amortemousque (approved):

packages/rum/src/domain/segmentCollection/segment.ts
amortemousque: Then I would find it less confusing using the naming `metadata`

packages/rum/src/types.ts
```diff
@@ -12,6 +12,7 @@ export interface SegmentMeta extends SegmentContext {
   has_full_snapshot: boolean
   records_count: number
   creation_reason: CreationReason
+  sequence_number: number
```
bcaudan, line 15: **🥜 nitpick:** `‏sequence_number` seems a bit blurry to me, it could be more precise with a naming indicating that it is related to the view, something like `view_index` or `position_in_view`, wdyt?

bcaudan (approved):

packages/rum/src/domain/segmentCollection/segment.ts
BenoitZugmeyer: Sure

packages/rum/src/types.ts
BenoitZugmeyer: This is something we agreed on with @acorretti . What do you think about those alternate namings Alfonso? Any preference?

acorretti: Well they're not really _views_, but I'm okay with something more specific like `segment_index`

BenoitZugmeyer: After discussion with @acorretti , we agreed on `index_in_view`. I'll make the change

bcaudan (approved):

Issue #1183, opened 2021-11-25
tslib updates?


liorosh:
Is there a special reason not to update `tslib`?
at the moment it is set to very old versions, and npm will not update to any version greater than 1.x.x
This causes a lot of duplications if a consumer already uses newer versions of `tslib` (e.g 2.x.x) and forces the consumer to bring both versions.

bcaudan:
Hello @liorosh,

We have this on our radar, we'll probably try to remove the dependency. 
We'll ping here when we'll make progress on this topic.

BenoitZugmeyer:
As of v4.6.0, tslib is no longer a Browser SDK dependency.

PR #1347, opened 2022-02-17
⚡️ [RUMF-1043] remove TSLib dependency

BenoitZugmeyer:
## Motivation

We observed that sometimes, our Browser SDK NPM packages is used with a bundler configuration which does not always remove unused parts of TSLib (ex: when using Webpack with a [`resolve.mainFields`](https://webpack.js.org/configuration/resolve/#resolvemainfields) set to `"main"`). In this case, TSLib adds a large amount of unused code to the user bundle. This is [one of the actions identified to reduce the SDK size](https://github.com/DataDog/browser-sdk/issues/317#issuecomment-1022076725).

Our TSLib version is also a bit old, and our users are probably using a newer, incompatible version, which could produce duplicate code in the bundle (see https://github.com/DataDog/browser-sdk/issues/1183).

## Changes

This PR removes the TSLib dependency and replace its usage with inlined TS helpers via the [`importHelpers` TS config](https://www.typescriptlang.org/tsconfig#importHelpers).

By doing so, the bundle grows a bit because helpers are duplicated in various modules. To mitigate this, this PR also forbid two syntaxes (object and array spread) that are known to produce such helpers (respectively [`assign`](https://github.com/microsoft/tslib/blob/c827964226e85118e2fd35b1cc68d4a5ad867f39/tslib.es6.js#L32-L41) and [spreadArrays](https://github.com/microsoft/tslib/blob/c827964226e85118e2fd35b1cc68d4a5ad867f39/tslib.es6.js#L161-L167)).

Note: some TS helpers are still used:

* [`decorate`](https://github.com/microsoft/tslib/blob/main/tslib.es6.js#L55-L60), [used by the Logger class](https://github.com/DataDog/browser-sdk/blob/4999eea80f19c42132dd771298c7a9667a7bdbba/packages/logs/src/domain/logger.ts#L49). Because it is used a single time, no need to change / factorize it.

* [commonjs module declaration helpers](https://github.com/microsoft/tslib/blob/c827964226e85118e2fd35b1cc68d4a5ad867f39/tslib.es6.js#L214-L226), which are used when using tsc to produce commonjs modules. Those can't really be replaced with alternative code without important drawbacks. In total, those two functions are duplicated 4 times in core and 3 times in rum. This may be improved by shiping pre-built NPM packages.

## Testing

To verify that our build does not produce more code than before, I run `yarn build` on both the PR branch and on `main` (in two different directories), and compare the result with something like:

```
diff -x '*.map' -u -r ../browser-sdk2/packages/rum/esm packages/rum/esm
```

The difference should be minimal.

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/core/src/browser/xhrObservable.ts
```diff
@@ -109,7 +110,7 @@ function sendXhr(this: XMLHttpRequest, observable: Observable<XhrContext>) {
     completeContext.state = 'complete'
     completeContext.duration = elapsed(startContext.startClocks.timeStamp, timeStampNow())
     completeContext.status = this.status
-    observable.notify({ ...completeContext })
+    observable.notify(assign({}, completeContext))
```
bcaudan, line 113: **💭 thought:** ‏what about extracting a `shallowClone` method for those cases?

amortemousque:
**💭 thought:** ‏I get the gain of this PR but of course I'm not a big fan to forbid object and array spread.
Do we agree that if we expose pre build packages for NPM users or remove the support of IE11 we will be able to use them again?

bcaudan (approved):

PR #1350, opened 2022-02-18
Update README.md

vacas:
## Motivation

I'm unsure if this piece of documentation was accidental or if I may be missing something, but it seemed like this was swapped by mistake.

## Changes

Swapped example `to collect 100% of your sessions using only the Browser RUM option` with ` to collect 100% of your sessions using only the Session Replay RUM option without recording a replay`.

### DIFF
--- a/packages/rum/README.md
+++ b/packages/rum/README.md
@@ -268,7 +268,7 @@ For example, to collect 100% of your sessions using only the Browser RUM option:
 datadogRum.init({
     ....
     sampleRate: 100,
-    replaySampleRate: 0
+    replaySampleRate: 100 // Note: if this is not included it will default to 100%
 });
 ```
 
@@ -278,7 +278,7 @@ For example, to collect 100% of your sessions using only the Session Replay RUM 
 datadogRum.init({
     ....
     sampleRate: 100,
-    replaySampleRate: 100 // Note: if this is not included it will default to 100%
+    replaySampleRate: 0
 });
 ```
 


BenoitZugmeyer:
Hello @vacas, thanks for your feedback.

This is not a mistake: `replaySampleRate: 100` indicates that all (100%) sessions will have full "Session Replay RUM" data, and `replaySampleRate: 0` indicates that none (0%) will have "Session Replay RUM" data.

Is there something specific that lead you to understand the contrary?

vacas:
Hi @BenoitZugmeyer! Sorry for the late reply! I guess what was confusing to me was that in the text above the last example is said `For example, to collect 100% of your sessions using only the Session Replay RUM option without recording a replay:` and then the example indicated `sessionReplay: 100` ([here](https://github.com/DataDog/browser-sdk/pull/1350/files#diff-a9c53af1f3e9e5b103b5bdef39b52eac991688984575a03139b6fcfdf143ecc5L281)), which lead me to believe this was a mistake, and found it a bit confusing. Either way, I see you have a PR open (https://github.com/DataDog/browser-sdk/pull/1362) and pretty much fixed the copy above [here](https://github.com/DataDog/browser-sdk/pull/1362/files#diff-a9c53af1f3e9e5b103b5bdef39b52eac991688984575a03139b6fcfdf143ecc5R278-R286), so thank you for that! I think that should take care of the confusion.

BenoitZugmeyer:
Will be fixed by #1362 then. Thanks for the reply!

PR #1354, opened 2022-02-22
👷[RUMF-1170] Tweak build env

bcaudan:
## Motivation

Remove e2e test specific code from release bundles.

## Changes

- Inject `BUILD_MODE` and `SDK_VERSION` through global variable replaced by their value at build time
- Do not replace the build env variable in unit tests in order to test behaviors related to those variables

## Testing

- [x] Local: e2e test code remove from webpack build, variables replaced for cjs/esm
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/transportConfiguration.ts
+++ b/packages/core/src/domain/configuration/transportConfiguration.ts
@@ -1,5 +1,3 @@
-import type { BuildEnv } from '../../boot/init'
-import { BuildMode } from '../../boot/init'
 import { objectValues } from '../../tools/utils'
 import type { InitConfiguration } from './configuration'
 import type { EndpointBuilder } from './endpointBuilder'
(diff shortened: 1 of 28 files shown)


packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -40,8 +35,9 @@ export function computeTransportConfiguration(
   }
 }
 
-function computeEndpointBuilders(initConfiguration: InitConfiguration, buildEnv: BuildEnv, tags: string[]) {
-  if (buildEnv.buildMode === BuildMode.E2E_TEST) {
+function computeEndpointBuilders(initConfiguration: InitConfiguration, tags: string[]) {
+  // @ts-ignore replaced at build time
```
BenoitZugmeyer, line 39: **💬 suggestion:** to avoid those ts-ignore, maybe add something like this in each module where you are using those kind of constants:

```ts
// replaced at build time
declare const __BUILD_ENV__BUILD_MODE__: string
```

BenoitZugmeyer (commented):
This looks great

webpack.base.js
```diff
@@ -20,10 +20,14 @@ module.exports = ({ entry, mode, filename, types }) => ({
         test: /\.ts$/,
         loader: 'string-replace-loader',
         options: {
-          multiple: [
```
BenoitZugmeyer, line 23: **💬 suggestion:** ‏you could probably replace the `string-replace-loader` with the [DefinePlugin](https://webpack.js.org/plugins/define-plugin/), which is part of Webpack, so that would be one dependency less

webpack.base.js
```diff
@@ -64,4 +54,15 @@ module.exports = ({ entry, mode, filename, types }) => ({
       }),
     ],
   },
+
+  plugins:
+    // do not replace in unit test in order to test behaviors related to build env variables
+    mode !== 'development'
```
BenoitZugmeyer, line 60: **🔨 warning:** ‏This breaks `yarn dev`, since it is also using a development mode

bcaudan: Nice catch! 👀

BenoitZugmeyer (approved):

PR #1368, opened 2022-03-01
 📦 [RUMF-1168] update typescript 

BenoitZugmeyer:
## Motivation

Have an up-to-date TS toolchain

## Changes

Update typescript, ts-loader and ts-node dependencies to their latest version

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/segmentCollection/startDeflateWorker.ts
+++ b/packages/rum/src/domain/segmentCollection/startDeflateWorker.ts
@@ -94,7 +94,7 @@ function onInitialized(worker: DeflateWorker) {
   }
 }
 
-function onError(error: ErrorEvent | Error | string) {
+function onError(error: unknown) {
   if (state.status === DeflateWorkerStatus.Loading) {
     display.error('Session Replay recording failed to start: an error occurred while creating the Worker:', error)
     if (error instanceof Event || (error instanceof Error && includes(error.message, 'Content Security Policy'))) {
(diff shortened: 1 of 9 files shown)


packages/rum/src/domain/segmentCollection/startDeflateWorker.ts
```diff
@@ -94,7 +94,7 @@ function onInitialized(worker: DeflateWorker) {
   }
 }
 
-function onError(error: ErrorEvent | Error | string) {
+function onError(error: unknown) {
```
amortemousque, line 97: Why do we have to set the type to `unknown`?

test/e2e/lib/framework/httpServers.ts
```diff
@@ -86,7 +86,7 @@ async function instantiateServer(): Promise<http.Server> {
     try {
       return await instantiateServerOnPort(port)
     } catch (error) {
-      if (error.code === 'EADDRINUSE') {
+      if (error instanceof Error && (error as NodeJS.ErrnoException).code === 'EADDRINUSE') {
```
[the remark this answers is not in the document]
BenoitZugmeyer: Yes, could work, as long as `error` is an instance of `Error`, which should always be the case here! Let's go with your suggestion as it is simpler.

packages/rum/src/domain/segmentCollection/startDeflateWorker.ts
amortemousque: All right! good

amortemousque (approved):

bcaudan (approved):

PR #1371, opened 2022-03-02
♻️ [RUMF-1181] preliminary refactorings for telemetry events collection

bcaudan:
## Motivation

Various refactoring mentioned in https://github.com/DataDog/browser-sdk/pull/1351 to ease the future collection of telemetry events

## Changes

- Move RUM batch creation earlier: in order to start collecting telemetry events as soon as possible
- Mutualize monitoring batch and logger batch: easily reuse a simple batch with replica structure
- Move RUM and Logs internal monitoring at same level: increase consistency between RUM and logs
- Move internal monitoring transport to RUM and logs: telemetry transport strategy will vary depending on the package, so let the caller handle it

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/boot/startRum.ts
```diff
@@ -102,24 +107,13 @@ export function startRumEventCollection(
   const urlContexts = startUrlContexts(lifeCycle, locationChangeObservable, location)
   const foregroundContexts = startForegroundContexts()
 
-  let stopBatch: () => void
-
-  if (canUseEventBridge()) {
-    startRumEventBridge(lifeCycle)
-  } else {
-    ;({ stop: stopBatch } = startRumBatch(configuration, lifeCycle))
```
BenoitZugmeyer, line 110: **💬 suggestion:** ‏since we don't use the `stop` method anymore, you could remove it from `startRumBatch`

packages/rum-core/src/boot/startRum.ts
```diff
@@ -95,6 +101,22 @@ export function startRum(
   }
 }
 
+function startRumInternalMonitoring(configuration: RumConfiguration) {
```
BenoitZugmeyer, line 104: **❓ question:** ‏This piece of code is now duplicated in both Logs and RUM. The goal is to be able to modify the RUM version to integrate it with LifeCycle in the follow-up PR, right?

bcaudan: exactly, it the exact same thing for now but it will diverge to handle telemetry events

BenoitZugmeyer (approved):

PR #1379, opened 2022-03-07
📦 [RUMF-1162] update developer-extension

BenoitZugmeyer:
## Motivation

Keep dependencies up to date and maintanable

## Changes

* Replace [bumbag](https://bumbag.style/) with [mantine](https://mantine.dev/).
  *  Reasons:
     * Bumbag still ships with a [fixed and outdated version of lodash](https://github.com/jxom/bumbag-ui/blob/44ffaa1c90b799edaa7c152c3d144833dcebb3bf/packages/bumbag/package.json#L34). This version is known to have [a vulnerability](https://github.com/advisories/GHSA-35jh-r3h4-6jhm), but the bumbag team doesn't [seem to be motivated to change that](https://github.com/jxom/bumbag-ui/issues/105#issue-823987605). Overriding the dependency version with yarn works, but is combursome.
     * Bumbag typings [are incompatible with `strictNullChecks`](https://github.com/jxom/bumbag-ui/issues/97), so the whole extension codebase lacks stricter types.
     * Bumbag did release a major version, and by upgrading it, we would need to make a pass on all usages anyway
  *  Why mantine:
     * Very similar to bumbag (uses same dependencies (emotion), provides the same components), so upgrade is easy
     * Lots more traction (6.5k stars on github, vs 971 for bumbag)
* Adjust the extension codebase to comply with stryctNullChecks
* Update react and types

The webpack plugins used by the extension has not been updated. This will be done in a future task where we update webpack.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,8 +23,6 @@ updates:
       # update jasmine: RUMF-1161
       - dependency-name: '*jasmine*'
       - dependency-name: '*sinon*'
-      # update extension: RUMF-1162
-      - dependency-name: '*react*'
       # update ajv: RUMF-1164
       - dependency-name: 'ajv'
         update-types: ['version-update:semver-major']
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,7 +25,6 @@ updates:
       - dependency-name: '*sinon*'
       # update extension: RUMF-1162
       - dependency-name: '*react*'
-      - dependency-name: 'bumbag'
       # update ajv: RUMF-1164
       - dependency-name: 'ajv'
         update-types: ['version-update:semver-major']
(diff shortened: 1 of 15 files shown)


.github/dependabot.yml
```diff
@@ -23,9 +23,18 @@ updates:
       # update jasmine: RUMF-1161
       - dependency-name: '*jasmine*'
       - dependency-name: '*sinon*'
+<<<<<<< HEAD
```
amortemousque, line 26: Seems like unresolved conflicts

BenoitZugmeyer: Oops. Fixed

amortemousque (approved):

PR #1388, opened 2022-03-08
👷 Add event list panel in the developer extension

amortemousque:
## Motivation

To help visualize sent events when developing I added an event list panel in the developer extension
The Log events option needs to be enabled

## Changes

Update developer extension

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/background/utils.ts
```diff
@@ -1,4 +1,4 @@
-export function evaluateCodeInActiveTab(code: (arg?: string) => void, arg?: string) {
+export function evaluateCodeInActiveTab(code: (arg?: string) => void, arg?: any) {
```
BenoitZugmeyer, line 1: **💬 suggestion:** ‏I don't see why you changed the type here, could you revert?

(Same for the function below)

developer-extension/src/background/utils.ts
```diff
@@ -19,3 +19,10 @@ function evaluateCodeInline(tabId: number, code: (arg?: string) => void, arg?: s
     }`,
   })
 }
+
+export function generateUUID(placeholder?: string): string {
```
BenoitZugmeyer, line 23: **💭 thought:** ‏I know this is a copy-paste from somewhere else, but this function could be a bit more readable

BenoitZugmeyer (approved):

developer-extension/src/background/domain/logEventsFromRequests.ts
```diff
@@ -21,25 +25,38 @@ chrome.webRequest.onBeforeRequest.addListener(
...
+    storeEvents(newEvents, info.tabId)
   },
   {
     urls: intakeUrlPatterns,
   },
   ['requestBody']
 )
 
+function storeEvents(newEvents: StoredEvent[], tabId: number) {
+  const previousEvents = Object.prototype.hasOwnProperty.call(store.local, tabId) ? store.local[tabId].events : []
```
BenoitZugmeyer, line 50: **💭 thought:** ‏Using the store for storing events isn't great. The whole `store.local` concept is especially not great, because everything will persist forever. We should think of a better design in the future.

amortemousque: Yes clearly

developer-extension/src/background/utils.ts
amortemousque: I removed it and use our existing function in core/src/tools/utils

PR #1405, opened 2022-03-14
👷 [RUMF-1156] update wdio to latest version 

BenoitZugmeyer:
## Motivation

Keep packages up to date

Note: there is an issue observed with this new version, see https://github.com/webdriverio/webdriverio/issues/8126 . Since it happens only when the spec fails in the first place, it shouldn't block us from upgrading

## Changes

* Update @wdio/* and webdriver packages
* Work around stricter type checks
* Improve browserstack capabilities configuration in both e2e and unit tests

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


test/e2e/lib/framework/pageSetups.ts
```diff
@@ -15,6 +15,7 @@ export interface SetupOptions {
 export type SetupFactory = (options: SetupOptions, servers: Servers) => string
 
 const isBrowserStack =
+  'services' in browser.config &&
   browser.config.services &&
```
amortemousque, line 19: Is it required to have both checks?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -104,9 +104,9 @@ describe('logs', () => {
 
       expect(servers.base.app.getLargeResponseWroteSize()).toBeLessThan(
         // When reading the request, chunks length are probably not aligning perfectly with the
-        // response length limit, so it sends few more bytes than necessary. Add a 50% margin of
-        // error to verify that it's still close to the expected limit.
-        DEFAULT_REQUEST_ERROR_RESPONSE_LENGTH_LIMIT * 1.5
+        // response length limit, so it sends few more bytes than necessary. Add a margin of error
+        // to verify that it's still close to the expected limit.
+        DEFAULT_REQUEST_ERROR_RESPONSE_LENGTH_LIMIT * 2
```
amortemousque, line 109: Why this change?

test/e2e/lib/framework/pageSetups.ts
BenoitZugmeyer: Yes, because `browser.config` is something like `{ services: something } | { other stuff }`, so we need the `'services' in browser.config` otherwise TS yield as us because "services" is not in `{ other stuff }`.

test/e2e/scenario/logs.scenario.ts
BenoitZugmeyer: Cf commit message: https://github.com/DataDog/browser-sdk/pull/1405/commits/12f6dd2b745898629185aef0bc83cd4096e50572

test/e2e/lib/helpers/browser.ts
```diff
@@ -15,6 +17,59 @@ export async function browserExecuteAsync<A extends any[]>(fn: (...params: A) =>
...
+  if ('bstack:options' in capabilities && capabilities['bstack:options']) {
+    platformName = (capabilities['bstack:options'] as any).os
+  }
+  if (platformName === undefined && 'platformName' in capabilities && typeof capabilities.platformName === 'string') {
+    platformName = capabilities.platformName
+  }
+  if (platformName === undefined) {
+    // The test is run locally, use the local os name
+    platformName = os.type()
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: Done

test/e2e/scenario/logs.scenario.ts
amortemousque: The comment does not explain why we move from 1.5 to 2. Or am I messing somethingN

BenoitZugmeyer: Maybe I wasn't clear enough? We need a bigger margin of error because apparently with the new wdio packages browsers are reading a bit more data at once (or maybe something in-between the browser and our node server, like the browserstack local tunnel). Let's chat about it on Monday if you want :)

amortemousque: Okay! It's clear now ;)

amortemousque (approved):

PR #1425, opened 2022-03-18
♻️ 🐛 [RUMF-1178] improve logs assembly - Part 1

amortemousque:
## Motivation

First step of the logs assembly refactoring
[RUMF-576] Fix the logs printed twice in the console when logs 'forwardErrorsToLogs' is enabled

## Changes

- Extract log assembly from startLogs
- Extract the send/print logic from the logger with a Sender delegate (needed for the fix)

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

[RUMF-576]: https://datadoghq.atlassian.net/browse/RUMF-576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/logs/src/domain/sender.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+  const contextManager = createContextManager()
+  contextManager.set(loggerContext)
+
+  function isAuthorized(status: StatusType, handlerType: HandlerType) {
+    const sanitizedHandlerType = Array.isArray(conf.handlerType) ? conf.handlerType : [conf.handlerType]
+    return STATUS_PRIORITIES[status] >= STATUS_PRIORITIES[conf.level] && includes(sanitizedHandlerType, handlerType)
+  }
+
+  function sendHttpRequest(message: string, messageContext: object | undefined, status: StatusType) {
```
BenoitZugmeyer, line 33: **💬 suggestion:** ‏Naming: maybe we could have `sendToHttp` / `sendToConsole`?

packages/logs/src/domain/sender.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const STATUS_PRIORITIES: { [key in StatusType]: number } = {
+  [StatusType.debug]: 0,
+  [StatusType.info]: 1,
+  [StatusType.warn]: 2,
+  [StatusType.error]: 3,
+}
+
+export function buildSender(
+  sendLog: (message: LogsMessage) => void,
```
BenoitZugmeyer, line 16: **💬 suggestion:** Naming: ‏`sendLog` is also used as a returned property, with a different meaning. Maybe we can find a more explicit name? Like `sendToHttpImpl`?

packages/logs/src/boot/startLogs.ts
```diff
@@ -164,7 +155,7 @@ export function doStartLogs(
         },
       }
     }
-    logger.log(log.message, messageContext, LogStatusForApi[log.api])
+    sender.sendHttpRequest(log.message, messageContext, LogStatusForApi[log.api])
```
BenoitZugmeyer, line 158: **👏 praise:** ‏I like that :)

packages/logs/src/domain/sender.ts
```diff
@@ -0,0 +1,64 @@
...
+export type Sender = ReturnType<typeof buildSender>
+
+export const STATUS_PRIORITIES: { [key in StatusType]: number } = {
+  [StatusType.debug]: 0,
+  [StatusType.info]: 1,
+  [StatusType.warn]: 2,
+  [StatusType.error]: 3,
+}
+
+export function buildSender(
```
BenoitZugmeyer, line 15: **💬 suggestion:** What about naming it `createSender`? I feel like `build` is more about formating immutable values, where `create` is to define a live value with some mutable state. WDYT?

packages/logs/src/domain/sender.ts
amortemousque: Yes better

packages/logs/src/domain/sender.ts
amortemousque: Sure

packages/logs/src/domain/sender.ts
amortemousque: Let's do that ;)

BenoitZugmeyer (approved):

PR #1432, opened 2022-03-21
♻️ [RUMF-1207] collect concurrent actions groundwork - move action history closer to action collection

BenoitZugmeyer:
## Motivation

Currently, actions are collected in `actionCollection`, and an history is built in `parentContext`. To communicate between the two `LifeCycle` events are used, and the currently active action is stored in those two places.

With concurrent actions, this implementation will be a bit more complex because `LifeCycle` events will need to be extended to track which actions are created/discarded, and a list of current actions have to be maintained in both places. This makes the whole process unnecessarily complex and error-prone.

To make things a bit simpler, let's make the action collection in charge of managing the action history, and expose an `ActionContexts` class the same way we do for URLs and sessions.

## Changes

See commits.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/parentContexts.ts
```diff
@@ -24,15 +24,6 @@ export function startParentContexts(lifeCycle: LifeCycle): ParentContexts {
     viewContextHistory.setCurrent(buildViewContext(view), view.startClocks.relative)
   })
 
-  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (view) => {
```
BenoitZugmeyer, line 27: This change is unrelated to the PR, and I meant to include it in a follow-up PR, but I merged it on staging before realizing it was there. Sorry about that! I think it's small enough to include it here, you'll let me know.

packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
```diff
@@ -41,117 +50,135 @@ export interface AutoActionCreatedEvent {
...
-  private id: string
-  private eventCountsSubscription: { eventCounts: EventCounts; stop(): void }
-
-  constructor(private lifeCycle: LifeCycle, private type: AutoActionType, private name: string, private event: Event) {
-    this.id = generateUUID()
-    this.startClocks = clocksNow()
-    this.eventCountsSubscription = trackEventCounts(lifeCycle)
-    this.lifeCycle.notify(LifeCycleEventType.AUTO_ACTION_CREATED, { id: this.id, startClocks: this.startClocks })
-  }
+function createAction(
```
amortemousque, line 122: Why not `newAction` like in view (newView)?

packages/rum-core/src/domain/rumEventsCollection/action/trackActions.ts
```diff
@@ -41,117 +50,135 @@ export interface AutoActionCreatedEvent {
 
 // Maximum duration for automatic actions
 export const AUTO_ACTION_MAX_DURATION = 10 * ONE_SECOND
+export const ACTION_CONTEXT_TIME_OUT_DELAY = 5 * ONE_MINUTE // arbitrary
+
+interface ActionController {
```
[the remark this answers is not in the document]
BenoitZugmeyer: I would be fine with it, but this type has a `discard()` method, which is not really context-y (like, it wouldn't be compatible with actual [Context typings](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/context.ts)), don't you think?

packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
```diff
@@ -20,8 +22,11 @@ export function startActionCollection(
     lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action, foregroundContexts))
   )
 
+  let actionContexts: ActionContexts
   if (configuration.trackInteractions) {
-    trackActions(lifeCycle, domMutationObservable, configuration)
+    actionContexts = trackActions(lifeCycle, domMutationObservable, configuration).actionContexts
+  } else {
+    actionContexts = { findActionId: noop as () => undefined }
```
amortemousque, line 29: Could be a bit shorter like so:
  ```typescript
let actionContexts: ActionContexts = { findActionId: noop as () => undefined }
if (configuration.trackInteractions) {
     actionContexts = trackActions(lifeCycle, domMutationObservable, configuration).actionContexts
}
```

amortemousque (approved):

PR #1459, opened 2022-03-28
Update telemetry data model

amortemousque:
## Motivation

Update telemetry date model according to the event format modifications

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/internalMonitoring/internalMonitoring.ts
```diff
@@ -71,15 +71,17 @@ export function startInternalMonitoring(configuration: Configuration): InternalM
...
         version: __BUILD_ENV__SDK_VERSION__,
+        source: 'browser' as const,
         _dd: {
-          event_type: 'internal_telemetry' as const,
+          format_version: 2 as const,
         },
       },
       telemetryContextProvider !== undefined ? telemetryContextProvider() : {},
-      message
+      { telemetry: message as any } // https://github.com/microsoft/TypeScript/issues/48457
```
BenoitZugmeyer, line 84: **💬 suggestion:** ‏move the `telemetry` property in the object passed as first argument?

BenoitZugmeyer (approved):

packages/core/src/domain/internalMonitoring/internalMonitoring.ts
amortemousque: Yes indeed

PR #1470, opened 2022-04-01
♻️ Simplify developer extension sdks config panel

amortemousque:
## Motivation

Use chrome.devtools.inspectedWindow to get the SDKs configuration. chrome.devtools.inspectedWindow is accessible in the devtools context, so we can simply create a custom hook to handle the data. No need to go through the background and to persist the data in the storage.

## Changes

- Create a useSdkConfig custom hook
- Update the configuration panel 

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/panel/useSdkConfig.ts
```diff
@@ -0,0 +1,42 @@
+import { useEffect, useState } from 'react'
+
+export function useSdkConfig() {
+  const [rumConfig, setRumConfig] = useState<object>()
+  const [logsConfig, setLogsConfig] = useState<object>()
+
+  useEffect(() => {
+    retry(getInitConfiguration('rum')).then(setRumConfig).catch(console.error)
+    retry(getInitConfiguration('logs')).then(setLogsConfig).catch(console.error)
+  }, [])
```
BenoitZugmeyer, line 10: **💬 suggestion:** ‏
Maybe keep a 2s refresh like in the previous version, so it'll refresh when we navigate to another page, and won't need to `retry` anymore

```suggestion
  useEffect(() => {
    function getInitConfigurations() {
      getInitConfiguration('rum').then(setRumConfig).catch(console.error)
      getInitConfiguration('logs').then(setLogsConfig).catch(console.error)
    }
    const id = setInterval(getInitConfigurations, 2000)
    getInitConfigurations()
    return () => clearInterval(id)
  }, [])
```

BenoitZugmeyer (approved):

developer-extension/src/panel/useSdkConfig.ts
amortemousque: yes

BenoitZugmeyer (approved):

PR #1471, opened 2022-04-01
👷 Improve developer extension event-inspector

amortemousque:
## Motivation

Improve developer extension event-inspector. 
image

## Changes

- Replace local storage by useEvents custom hooks which uses chrome.devtools.network.onRequestFinished under the hood
- Add a filter capability 
  - Search syntax: `type:view application.id:40d8ca4b`
- Log event by default (remove logEventsFromRequests)
- Move the actionTab as an actionBar to always see it

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/common/types.ts
```diff
@@ -12,13 +12,13 @@ export interface BackgroundActions {
 
 export interface PopupActions {
   newStore: Store
+  newEvents: StoredEvent[]
```
BenoitZugmeyer, line 15: I don't think you need this?

developer-extension/src/common/types.ts
```diff
@@ -12,13 +12,13 @@ export interface BackgroundActions {
...
 }
 
 export interface Store {
   devServerStatus: 'unavailable' | 'checking' | 'available'
   useDevBundles: boolean
   useRumSlim: boolean
-  logEventsFromRequests: boolean
   blockIntakeRequests: boolean
   local: {
```
BenoitZugmeyer, line 23: **💬 suggestion:** ‏Since the local store is not used anymore after your two PRs, could you remove it?

developer-extension/src/panel/hooks/useEvents.ts
```diff
@@ -0,0 +1,89 @@
+import { useEffect, useState } from 'react'
+import { generateUUID } from '../../../../packages/core/src/tools/utils'
+import type { StoredEvent } from '../../common/types'
```
BenoitZugmeyer, line 3: **💬 suggestion:** ‏remove the `StoredEvent` from `common`, since it is not used in the background anymore

BenoitZugmeyer (approved):
Looks great!

developer-extension/src/common/types.ts
amortemousque: Yes, this is the idea ;)

developer-extension/src/panel/hooks/useEvents.ts
amortemousque: Sure

PR #1479, opened 2022-04-04
✨ ⚗  [RUMF-1224] remove console APIs prefix

amortemousque:
## Motivation

Remove the console APIs prefix except for console error
Add the origin to the root of the log

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/console/consoleObservable.ts
```diff
@@ -57,21 +57,23 @@ function createConsoleObservable(api: ConsoleApiName) {
...
+  const prefix = api === ConsoleApiName.error ? `console ${api}: ` : ''
+  let message = params.map((param) => formatConsoleParameters(param)).join(' ')
+  let stack
 
   if (api === ConsoleApiName.error) {
     const firstErrorParam = find(params, (param: unknown): param is Error => param instanceof Error)
-    log.stack = firstErrorParam ? toStackTraceString(computeStackTrace(firstErrorParam)) : undefined
-    log.handlingStack = handlingStack
+    stack = firstErrorParam ? toStackTraceString(computeStackTrace(firstErrorParam)) : undefined
+    message = `${prefix}${message}`
```
BenoitZugmeyer, line 68: **💬 suggestion:** ‏simply add the prefix here, since it's the same condition
```suggestion
    message = `console error: ${message}`
```

packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -227,6 +234,14 @@ describe('logs entry', () => {
       })
     })
 
+    it('logs a message with "logger" origin', () => {
+      updateExperimentalFeatures(['forward-logs'])
+      LOGS.logger.log('message')
+
+      expect(getLoggedMessage(0).message.origin).toEqual(ErrorSource.LOGGER)
+      resetExperimentalFeatures()
```
BenoitZugmeyer, line 242: **🥜 nitpick:** ‏you should reset global state in a `afterEach` callback because if you don't and the test fails, the state won't be reset.

packages/logs/src/domain/logger.spec.ts
```diff
@@ -26,6 +26,19 @@ describe('Logger', () => {
       expect(getLoggedMessage(0).status).toEqual(StatusType.info)
     })
 
+    it("'logger.log' should set 'logger' origin", () => {
+      updateExperimentalFeatures(['forward-logs'])
```
BenoitZugmeyer, line 30: **💬 suggestion:** add `with 'forward-logs' enabled, ...` to the test description

packages/logs/src/domain/logger.spec.ts
```diff
@@ -26,6 +26,19 @@ describe('Logger', () => {
...
+      logger.log('message')
```
BenoitZugmeyer, line 31: **💬 suggestion:** ‏add a test to make sure `origin` can be overriden by the customer, like:

```ts
logger.log('message', { origin: 'foo' })
expect(getLoggedMessage(0).origin).toEqual('foo')
```

packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -227,6 +234,14 @@ describe('logs entry', () => {
       })
     })
 
+    it('logs a message with "logger" origin', () => {
```
BenoitZugmeyer, line 237: **💬 suggestion:** ‏add `when 'forward-logs' is enabled, ...`

packages/core/src/domain/console/consoleObservable.ts
amortemousque: My bad, indeed :)

packages/logs/src/boot/logsPublicApi.spec.ts
amortemousque: Sure

packages/logs/src/domain/logger.spec.ts
amortemousque: Done :)

BenoitZugmeyer (approved):

bcaudan (approved):

PR #1497, opened 2022-04-12
♻️  [RUMF-1178] Improve log assembly part-3

amortemousque:
## Motivation

Improve the Logs assembly.

Since the Logs assembly was intertwined with the rest of the logic, I had a hard time to limit the modifications so
the PR is quite big.  However, the flow is now very similar to RUM, so you should get use to it relatively fast.
Let me know if it is indigestible, I'll try to break it down.

## Changes

- Create a `CommonContext` to encapsulate the current `date`, `view` and `loggerGlobalContext`
- Create a Log `LifeCycle` with `RAW_LOG_COLLECTED` and `LOG_COLLECTED`
- Assemble the different contexts in `startLogsAssembly`
- Add logger  getLevel, getHandler and getContext
- Pass down the main logger to startLogsAssembly to be assembled for all collection

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,8 +29,5 @@ updates:
       - dependency-name: 'ansi-regex'
       # update webpack: RUMF-1166
       - dependency-name: '*webpack*'
-      # update node-fetch: RUMF-1167
-      - dependency-name: 'node-fetch'
-        update-types: ['version-update:semver-major']
       - dependency-name: '*'
         update-types: ['version-update:semver-minor', 'version-update:semver-patch']
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "karma-webpack": "5.0.0",
     "lerna": "4.0.0",
     "minimatch": "5.0.1",
-    "node-fetch": "2",
+    "node-fetch": "3.2.3",
     "npm-run-all": "4.1.5",
     "prettier": "2.6.2",
     "sinon": "9.2.4",
--- a/scripts/bs-wrapper.js
+++ b/scripts/bs-wrapper.js
@@ -17,8 +17,9 @@
 const { spawnCommand, printLog, logAndExit, executeCommand, fetch } = require('./utils')
 
 const AVAILABILITY_CHECK_DELAY = 30_000
-// eslint-disable-next-line max-len
-const RUNNING_BUILDS_API = `https://${process.env.BS_USERNAME}:${process.env.BS_ACCESS_KEY}@api.browserstack.com/automate/builds.json?status=running`
+const BS_USERNAME = process.env.BS_USERNAME
+const BS_ACCESS_KEY = process.env.BS_ACCESS_KEY
+const BS_BUILD_URL = 'https://api.browserstack.com/automate/builds.json?status=running'
 
 async function main() {
   if (await executeCommand('git tag --points-at HEAD')) {
@@ -39,7 +40,11 @@ async function waitForAvailability() {
 }
 
 async function hasRunningBuild() {
-  return (await fetch(RUNNING_BUILDS_API)) !== '[]'
+  return (
+    (await fetch(BS_BUILD_URL, {
+      headers: { Authorization: `Basic ${Buffer.from(`${BS_USERNAME}:${BS_ACCESS_KEY}`).toString('base64')}` },
+    })) !== '[]'
+  )
 }
 
 function runTests() {
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -2,7 +2,9 @@ const util = require('util')
 const fs = require('fs/promises')
 const execute = util.promisify(require('child_process').exec)
 const spawn = require('child_process').spawn
-const fetch = require('node-fetch')
+// node-fetch v3.x only support ESM syntax.
+// Todo: Remove node-fetch when node v18 LTS is released with fetch out of the box
+const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args))
 
 const CI_FILE = '.gitlab-ci.yml'
 
@@ -92,8 +94,8 @@ function printLog(...params) {
   console.log(greenColor, ...params, resetColor)
 }
 
-async function fetchWrapper(url) {
-  const response = await fetch(url)
+async function fetchWrapper(url, options) {
+  const response = await fetch(url, options)
   if (!response.ok) {
     throw new Error(`HTTP Error Response: ${response.status} ${response.statusText}`)
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3746,6 +3746,11 @@ dashdash@^1.12.0:
   dependencies:
     assert-plus "^1.0.0"
 
+data-uri-to-buffer@^4.0.0:
+  version "4.0.0"
+  resolved "https://registry.yarnpkg.com/data-uri-to-buffer/-/data-uri-to-buffer-4.0.0.tgz#b5db46aea50f6176428ac05b73be39a57701a64b"
+  integrity sha512-Vr3mLBA8qWmcuschSLAOogKgQ/Jwxulv3RNE4FXnYWRGujzrRWQI4m12fQqRkwX06C0KanhLr4hK+GydchZsaA==
+
 date-format@0.0.2:
   version "0.0.2"
   resolved "https://registry.yarnpkg.com/date-format/-/date-format-0.0.2.tgz#fafd448f72115ef1e2b739155ae92f2be6c28dd1"
@@ -4842,6 +4847,14 @@ fd-slicer@~1.1.0:
   dependencies:
     pend "~1.2.0"
 
+fetch-blob@^3.1.2, fetch-blob@^3.1.4:
+  version "3.1.5"
+  resolved "https://registry.yarnpkg.com/fetch-blob/-/fetch-blob-3.1.5.tgz#0077bf5f3fcdbd9d75a0b5362f77dbb743489863"
+  integrity sha512-N64ZpKqoLejlrwkIAnb9iLSA3Vx/kjgzpcDhygcqJ2KKjky8nCgUQ+dzXtbrLaWZGZNmNfQTsiQ0weZ1svglHg==
+  dependencies:
+    node-domexception "^1.0.0"
+    web-streams-polyfill "^3.0.3"
+
 figures@^3.0.0:
   version "3.1.0"
   resolved "https://registry.yarnpkg.com/figures/-/figures-3.1.0.tgz#4b198dd07d8d71530642864af2d45dd9e459c4ec"
@@ -4943,6 +4956,13 @@ form-data@~2.3.2:
     asynckit "^0.4.0"
     combined-stream "^1.0.6"
     mime-types "^2.1.12"
+
+formdata-polyfill@^4.0.10:
+  version "4.0.10"
+  resolved "https://registry.yarnpkg.com/formdata-polyfill/-/formdata-polyfill-4.0.10.tgz#24807c31c9d402e002ab3d8c720144ceb8848423"
+  integrity sha512-buewHzMvYL29jdeQTVILecSaZKnt/RJWjoZCF5OW60Z67/GmSLBkOFM7qh1PI3zFNtJbaZL5eQu1vLfazOwj4g==
+  dependencies:
+    fetch-blob "^3.1.2"
 
 forwarded@0.2.0:
   version "0.2.0"
@@ -7195,12 +7215,26 @@ no-case@^3.0.4:
     lower-case "^2.0.2"
     tslib "^2.0.3"
 
-node-fetch@2, node-fetch@2.6.7, node-fetch@^2.6.1:
+node-domexception@^1.0.0:
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/node-domexception/-/node-domexception-1.0.0.tgz#6888db46a1f71c0b76b3f7555016b63fe64766e5"
+  integrity sha512-/jKZoMpw0F8GRwl4/eLROPA3cfcXtLApP0QzLmUT/HuPCZWyB7IY9ZrMeKw2O/nFIqPQB3PVM9aYm0F312AXDQ==
+
+node-fetch@2.6.7, node-fetch@^2.6.1:
   version "2.6.7"
   resolved "https://registry.yarnpkg.com/node-fetch/-/node-fetch-2.6.7.tgz#24de9fba827e3b4ae44dc8b20256a379160052ad"
   integrity sha512-ZjMPFEfVx5j+y2yF35Kzx5sF7kDzxuDj6ziH4FFbOp87zKDZNx8yExJIb05OGF4Nlt9IHFIMBkRl41VdvcNdbQ==
   dependencies:
     whatwg-url "^5.0.0"
+
+node-fetch@3.2.3:
+  version "3.2.3"
+  resolved "https://registry.yarnpkg.com/node-fetch/-/node-fetch-3.2.3.tgz#a03c9cc2044d21d1a021566bd52f080f333719a6"
+  integrity sha512-AXP18u4pidSZ1xYXRDPY/8jdv3RAozIt/WLNR/MBGZAz+xjtlr90RvCnsvHQRiXyWliZF/CpytExp32UU67/SA==
+  dependencies:
+    data-uri-to-buffer "^4.0.0"
+    fetch-blob "^3.1.4"
+    formdata-polyfill "^4.0.10"
 
 node-forge@1.3.0:
   version "1.3.0"
@@ -9774,6 +9808,11 @@ wcwidth@>=1.0.1, wcwidth@^1.0.0, wcwidth@^1.0.1:
   integrity sha1-8LDc+RW8X/FSivrbLA4XtTLaL+g=
   dependencies:
     defaults "^1.0.3"
+
+web-streams-polyfill@^3.0.3:
+  version "3.2.1"
+  resolved "https://registry.yarnpkg.com/web-streams-polyfill/-/web-streams-polyfill-3.2.1.tgz#71c2718c52b45fd49dbeee88634b3a60ceab42a6"
+  integrity sha512-e0MO3wdXWKrLbL0DgGnUV7WHVuw9OUvL4hjgnPkIeEvESk74gAITi5G606JtZPp39cd8HA9VQzCIvA49LpPN5Q==
 
 webdriver@7.18.0:
   version "7.18.0"


packages/logs/src/domain/logger.ts
```diff
@@ -64,22 +60,42 @@ export class Logger {
...
+) {
+  const contextManager = createContextManager()
+  // Todo: merge the logger name into the context in logger collection in next major version
+  contextManager.set(assign({}, context, { logger: { name } }))
+
+  return {
+    handlerType,
+    level,
+    name,
+    contextManager,
```
bcaudan, line 99: **💬 suggestion:** ‏I think we should handle separately options used to build the current message (name, context) from the one used to "handle" this message (handler, level).

We could partially build the message inside the logger like we did before:
```ts
combine({ message, status }, this.contextManager.get(), messageContext)
```

and let that only override more general contexts (global, rum, service/env/version) during assembly.

wdyt?

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
```diff
@@ -0,0 +1,50 @@
...
+export function startLoggerCollection(lifeCycle: LifeCycle) {
+  function addLog(logsMessage: LogsMessage, loggerOptions: LoggerOptions, savedCommonContext?: CommonContext) {
+    const messageContext = logsMessage.context
+
+    if (isAuthorized(logsMessage.status, HandlerType.console, loggerOptions)) {
+      display.log(
+        `${logsMessage.status}: ${logsMessage.message}`,
+        combine(loggerOptions.contextManager.get(), messageContext)
+      )
+    }
```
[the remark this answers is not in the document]
amortemousque: I suppose it could always be printed for this type of error :)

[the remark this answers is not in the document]
bcaudan: 🤔 I had mind that we had some kind of global parameters for handler type but in fact no, it is only logger specific :facep

packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -222,67 +212,24 @@ describe('logs entry', () => {
 
     it('logs a message', () => {
       LOGS.logger.log('message')
-
-      expect(getLoggedMessage(0)).toEqual({
-        context: {
```
amortemousque, line 227: I removed the common context check since it's only combined assemble stage

packages/logs/src/domain/logsCollection/console/consoleCollection.spec.ts
```diff
@@ -1,120 +1,100 @@
...
     console.error('foo', 'bar')
 
-    expect(sendLogSpy.calls.mostRecent().args[0].error).toEqual({
+    expect(rawLogs[0].rawLog.error).toEqual({
       origin: ErrorSource.CONSOLE,
       stack: undefined,
     })
   })
-
-  it('should not print the log twice when console handler is enabled', () => {
```
amortemousque, line 103: This test now requires that `consoleCollection` and `loggerCollection` start so I moved to startLogs.spec

packages/logs/src/domain/logger.spec.ts
```diff
@@ -30,126 +30,83 @@ describe('Logger', () => {
       expect(getLoggedMessage(0).status).toEqual(StatusType.info)
     })
 
-    it("'logger.log' should set 'logger' origin when ff forward-logs enabled", () => {
```
amortemousque, line 33: Moved to loggerCollection.spec

packages/logs/src/domain/logsCollection/console/consoleCollection.ts
```diff
@@ -19,24 +19,25 @@ const LogStatusForApi = {
...
-          origin: ErrorSource.CONSOLE, // Todo: Remove in the next major release
-          stack: log.stack,
-        },
-      }
-    }
-    if (isExperimentalFeatureEnabled('forward-logs')) {
-      messageContext.origin = ErrorSource.CONSOLE
-    }
-    sender.sendToHttp(log.message, messageContext, LogStatusForApi[log.api])
+    lifeCycle.notify(LifeCycleEventType.RAW_LOG_COLLECTED, {
```
bcaudan, line 27: **🥜 nitpick:** ‏we could inline this function to avoid to name it

packages/logs/src/domain/logsCollection/console/consoleCollection.ts
```diff
@@ -19,24 +19,25 @@ const LogStatusForApi = {
   [ConsoleApiName.warn]: StatusType.warn,
   [ConsoleApiName.error]: StatusType.error,
 }
-export function startConsoleCollection(configuration: LogsConfiguration, sender: Sender) {
+export function startConsoleCollection(configuration: LogsConfiguration, lifeCycle: LifeCycle) {
   const consoleObservable = initConsoleObservable(configuration.forwardConsoleLogs)
   const consoleSubscription = consoleObservable.subscribe(reportConsoleLog)
```
bcaudan, line 24: **🥜 nitpick:** ‏`consoleObservable` could be inlined

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    if (isAuthorized(logsMessage.status, HandlerType.console, logger)) {
+      display.log(`${logsMessage.status}: ${logsMessage.message}`, combine(logger.getContext(), messageContext))
+    }
+
+    lifeCycle.notify(LifeCycleEventType.RAW_LOG_COLLECTED, {
+      rawLog: {
+        message: logsMessage.message,
+        status: logsMessage.status,
+        origin: isExperimentalFeatureEnabled('forward-logs') ? ErrorSource.LOGGER : undefined,
```
bcaudan, line 27: **💬 suggestion:** ‏related to previous PRs, should we rename `ErrorSource` since it does not seem to be related to error anymore?

packages/logs/src/domain/logsCollection/report/reportCollection.ts
```diff
@@ -18,18 +18,16 @@ const LogStatusForReport = {
...
 
   function logReport(report: RawReport) {
     let message = report.message
-    const messageContext: Partial<LogsEvent> = {
-      origin: ErrorSource.REPORT,
-    }
-    const logStatus = LogStatusForReport[report.type]
-    if (logStatus === StatusType.error) {
-      messageContext.error = {
+    const status = LogStatusForReport[report.type]
```
bcaudan, line 27: **🥜 nitpick:** we could inline this function to avoid to name it‏

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -0,0 +1,34 @@
+import type { Context, ErrorSource, TimeStamp } from '@datadog/browser-core'
+import type { StatusType } from './domain/logger'
+
+export interface RawLogsEvent {
```
bcaudan, line 4: **💬 suggestion:** ‏could it be interesting to document the different raw types, console, report, logger, ...?

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
```diff
@@ -0,0 +1,46 @@
...
+  [StatusType.error]: 3,
+}
+
+export function startLoggerCollection(lifeCycle: LifeCycle) {
+  function handleLog(logsMessage: LogsMessage, logger: Logger, savedCommonContext?: CommonContext) {
+    const messageContext = logsMessage.context
+
+    if (isAuthorized(logsMessage.status, HandlerType.console, logger)) {
+      display.log(`${logsMessage.status}: ${logsMessage.message}`, combine(logger.getContext(), messageContext))
+    }
```
BenoitZugmeyer, line 21: **💭 thought:** ‏at first glance I expected that the `display.log` call should be done in another module using `lifeCycle.subscribe(LOG_COLLECTED, ...)`. But:

* it seems that LOG_COLLECTED (and the whole LifeCycle, and the assembly module) only expects to handle logs that should be sent to datadog. It's a bit unexpected but seems reasonable.

* I think that we should keep the `display.log` call close to the actual `Logger.*` usage, i.e. not moving it in a LifeCycle `subscribe` callback, but maybe moving it back to the `Logger` class. With #1501 we'll have a stack trace displayed below "error" and "warn" logs, which is great! But having `display.log` deep into the stack trace adds a bit of noise and impacts the developper experience. For example, with your PR, the stack trace would look like this:
Screenshot 2022-04-22 at 17 56 38
Which is not dramatic, but we might want to minimize this stack trace :)

BenoitZugmeyer:
I'm actually curious about your decision to remove the `LoggerOptions` and extend the `Logger` class instead :) I thought it was a bit nicer to have a "dumb" state used accross the SDK instead of a `Logger` instance that:
* constrains us to the current design, as it is also publicly exposed
* should not be used internally by convention (except to access their state)

packages/logs/src/domain/assembly.ts
```diff
@@ -0,0 +1,89 @@
...
+      configuration.eventRateLimiterThreshold,
+      reportAgentError
+    ),
+    [StatusType.debug]: createEventRateLimiter(
+      StatusType.debug,
+      configuration.eventRateLimiterThreshold,
+      reportAgentError
+    ),
+    ['custom']: createEventRateLimiter('custom', configuration.eventRateLimiterThreshold, reportAgentError),
+  }
```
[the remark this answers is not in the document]
amortemousque: 👍

packages/logs/src/domain/lifeCycle.ts
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      unsubscribe: () => {
+        this.callbacks[eventType] = this.callbacks[eventType]!.filter((other) => callback !== other)
+      },
+    }
+  }
+}
+
+export interface RawLogCollectedData<E extends RawLogsEvent = RawLogsEvent> {
+  rawLog: E
```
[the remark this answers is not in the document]
amortemousque: Indeed

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -104,19 +115,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
...
-          view: {
-            referrer: document.referrer,
-            url: window.location.href,
-          },
-        },
-        globalContextManager.get()
-      )
-    )
+  function createLogger(options: LoggerOptions) {
+    return new Logger(options, (...params) => addLogStrategy(...params))
```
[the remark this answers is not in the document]
amortemousque: Sure

packages/logs/src/domain/assembly.ts
```diff
@@ -0,0 +1,89 @@
...
+        logger.getContext(),
+        messageContext
+      )
+
+      if (
+        configuration.beforeSend?.(log) === false ||
+        (log.error?.origin !== ErrorSource.AGENT &&
+          (logRateLimiters[log.status] ?? logRateLimiters['custom']).isLimitReached()) ||
+        // Todo: [RUMF-1230] Move this check to the logger collection in the next major release
+        !isAuthorized(rawLog.status, HandlerType.http, logger)
```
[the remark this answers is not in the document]
amortemousque: true

packages/logs/src/domain/assembly.ts
```diff
@@ -0,0 +1,89 @@
...
+      const commonContext = savedCommonContext || getCommonContext()
+      const log = combine(
+        { service: configuration.service, session_id: session.id },
+        { date: commonContext.date, view: commonContext.view },
+        commonContext.context,
+        getRUMInternalContext(startTime),
+        rawLog,
+        logger.getContext(),
+        messageContext
+      )
```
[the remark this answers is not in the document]
amortemousque: It would mean that at assembly level, we can not guarantee the format for `rawLogsEvent` coming from `Loggers` since it can be overridden by the user. This can be problematic for logic relying on certain field like `date`, `origin` or `status`. Wdyt?

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
amortemousque: Indeed. I'm wondering if the issue is only on naming. Now that it serve `log origin` and `rum error source`, I feel like it starts to be different concepts. We already have divergence:
```
  AGENT: 'agent',
  CONSOLE: 'console',
  CUSTOM: 'custom', // specific RUM
  LOGGER: 'logger', // specific Logs
  NETWORK: 'network', // specific Logs
  SOURCE: 'source',
  REPORT: 'report',
```
IMO, there is two options:

- keep a single list in coe packahe with a better name like: `apiSource` 

- have two different list:

`logOrigins` in Logs package
`errorSource` in RUM core package

Wdyt?

packages/logs/src/rawLogsEvent.types.ts
amortemousque: Indeed it could be nice to have.

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
amortemousque: Yes right. Since this PR is quite big and it takes times to validate, I'm in favor of working on this improvement while doing https://github.com/DataDog/browser-sdk/pull/1501

packages/logs/src/domain/assembly.ts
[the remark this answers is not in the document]
bcaudan: FMU, those fields can still be overwritten by the customer. I would find it less surprising to use customer value as soon as possible rather than using our value for an internal behavior (like associate RUM context from the date) and then overwrite our value by the customer one which would make thing quite hard to debug.

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
bcaudan: lgtm, we should probably postpone that to a next PR though

amortemousque: What option exactly look good to you? :)

bcaudan: 😅 two different lists

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -91,13 +91,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
...
 
     createLogger: monitor((name: string, conf: LoggerConfiguration = {}) => {
-      customLoggers[name] = new Logger(
-        (...params) => handleLogStrategy(...params),
-        name,
-        conf.handler,
-        conf.level,
-        conf.context
-      )
+      customLoggers[name] = new Logger(handleLogStrategy, name, conf.handler, conf.level, conf.context)
```
bcaudan, line 94: **❗ warning:** FMU using `handleLogStrategy` directly here should break loggers created before the init 🤔 
do we have a test for this one?‏

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -1,29 +1,67 @@
...
-  date?: TimeStamp
+export type RawLogsEvent =
+  | RawConsoleLogsEvent
+  | RawNetworkLogsEvent
+  | RawLoggerLogsEvent
+  | RawAgentLogsEvent
+  | RawReportLogsEvent
+  | RawRuntimeLogsEvent
+
+export type RawLogsEventOrigin = RawLogsEvent['origin']
```
bcaudan, line 12: **❓ question:** ‏is it used somewhere?

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -1,29 +1,67 @@
...
-    [k: string]: unknown
-  }
-  http?: {
+  error?: Error
+}
+
+interface CommonRawErrorLogsEvent extends CommonRawLogsEvent {
+  status: typeof StatusType.error
+  error: Error
+}
```
bcaudan, line 30: **🥜 nitpick:** ‏personal preference, I would inline this interface to avoid to navigating across the hierarchy to understand what we actually have in an interface

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -1,29 +1,67 @@
...
+
+export interface RawConsoleLogsEvent extends CommonRawLogsEvent {
+  origin: typeof ErrorSource.CONSOLE
+}
+
+export interface RawLoggerLogsEvent extends CommonRawLogsEvent {
+  origin: typeof ErrorSource.LOGGER
+}
+
```
bcaudan, line 39: **💭 thought:** ‏I would have love to have `logger?: {name: string}` here

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -1,29 +1,67 @@
...
-    [k: string]: unknown
-  }
-  error?: {
-    kind?: string
-    origin: ErrorSource
-    stack?: string
-    [k: string]: unknown
-  }
-  http?: {
+  error?: Error
```
bcaudan, line 24: **💭 thought:** ‏it could be nice to have a design allowing to have `date: TimeStamp` here

packages/logs/src/domain/assembly.ts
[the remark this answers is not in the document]
amortemousque: Fair enough. Since it's a non negligible change I'm in favor of doing it when tackling the [RUMF-1230](https://datadoghq.atlassian.net/browse/RUMF-1230)
Wdyt?

packages/logs/src/domain/logsCollection/logger/loggerCollection.ts
amortemousque: Yes I'll do it in a following PR

packages/logs/src/rawLogsEvent.types.ts
amortemousque: Yes me too. It directly related to https://github.com/DataDog/browser-sdk/pull/1497#discussion_r853177027
and It will be done with [RUMF-1230](https://datadoghq.atlassian.net/browse/RUMF-1230)

packages/logs/src/boot/logsPublicApi.ts
amortemousque: Indeed. I fixed it and added a test.

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -37,9 +31,11 @@ export interface RawLoggerLogsEvent extends CommonRawLogsEvent {
   origin: typeof ErrorSource.LOGGER
 }
 
-export interface RawNetworkLogsEvent extends CommonRawErrorLogsEvent {
+export interface RawNetworkLogsEvent extends CommonRawLogsEvent {
   date: TimeStamp
```
bcaudan, line 35: **🥜 nitpick:** ‏date should not be needed here anymore

bcaudan (approved):

PR #1499, opened 2022-04-13
✨ [RUMF-1237] The event bridge allowed hosts should also match subdomains

amortemousque:
## Motivation

When  a bridge is injected by mobile with `AllowedWebViewHosts`
The browser SDK should send data to the bridge if on domains and subdomains that match AllowedWebViewHosts

## Changes

Update canUseEventBridge

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/eventBridge.spec.ts
+++ b/packages/core/src/transport/eventBridge.spec.ts
@@ -9,16 +9,25 @@ describe('canUseEventBridge', () => {
   })
 
   it('should detect when the bridge is present and the webView host is allowed', () => {
-    initEventBridgeStub()
-    expect(canUseEventBridge()).toBeTrue()
-  })
-
-  it('should not detect when the bridge is absent', () => {
-    expect(canUseEventBridge()).toBeFalse()
+    initEventBridgeStub(allowedWebViewHosts)
+    expect(canUseEventBridge('foo.bar')).toBeTrue()
+    expect(canUseEventBridge('baz.foo.bar')).toBeTrue()
+    expect(canUseEventBridge('www.foo.bar')).toBeTrue()
   })
 
   it('should not detect when the bridge is present and the webView host is not allowed', () => {
     initEventBridgeStub(allowedWebViewHosts)
+    expect(canUseEventBridge('foo.com')).toBeFalse()
+    expect(canUseEventBridge('foo.bar.baz')).toBeFalse()
+    expect(canUseEventBridge('bazfoo.bar')).toBeFalse()
+  })
+
+  it('should not detect when the bridge on the parent domain if only the subdomain is allowed', () => {
+    initEventBridgeStub(['baz.foo.bar'])
+    expect(canUseEventBridge('foo.bar')).toBeFalse()
+  })
+
+  it('should not detect when the bridge is absent', () => {
     expect(canUseEventBridge()).toBeFalse()
   })
 })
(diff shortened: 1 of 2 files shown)


packages/core/src/transport/eventBridge.ts
```diff
@@ -26,10 +26,16 @@ export function getEventBridge<T, E>() {
...
-export function canUseEventBridge(): boolean {
+export function canUseEventBridge(hostname = window.location.hostname): boolean {
   const bridge = getEventBridge()
-
-  return !!bridge && includes(bridge.getAllowedWebViewHosts(), window.location.hostname)
+  return (
+    !!bridge &&
+    bridge.getAllowedWebViewHosts().some((host) => {
+      const escapedHost = host.replace(/\./g, '\\$&')
+      const isDomainOrSubDomain = new RegExp(`(^|.+\\.)${escapedHost}$`)
```
bcaudan, line 35:
```suggestion
      const isDomainOrSubDomain = new RegExp(`^(.+\\.)*${escapedHost}$`)
```
personal preference, it seems more explicit to me that we can have 0 to n subdomains

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -9,16 +9,25 @@ describe('canUseEventBridge', () => {
...
-    initEventBridgeStub()
-    expect(canUseEventBridge()).toBeTrue()
-  })
-
-  it('should not detect when the bridge is absent', () => {
-    expect(canUseEventBridge()).toBeFalse()
+    initEventBridgeStub(allowedWebViewHosts)
+    expect(canUseEventBridge('foo.bar')).toBeTrue()
+    expect(canUseEventBridge('baz.foo.bar')).toBeTrue()
+    expect(canUseEventBridge('www.foo.bar')).toBeTrue()
```
bcaudan, line 15:
```suggestion
    expect(canUseEventBridge('www.foo.bar')).toBeTrue()
    expect(canUseEventBridge('www.qux.foo.bar')).toBeTrue()
```
Testing multiple subdomains

packages/core/src/transport/eventBridge.ts
amortemousque: Why not

packages/core/src/transport/eventBridge.spec.ts
amortemousque: Yes, done

packages/core/src/transport/eventBridge.ts
```diff
@@ -26,10 +26,16 @@ export function getEventBridge<T, E>() {
...
+export function canUseEventBridge(hostname = getGlobalObject<Window>().location?.hostname): boolean {
   const bridge = getEventBridge()
-
-  return !!bridge && includes(bridge.getAllowedWebViewHosts(), window.location.hostname)
+  return (
+    !!bridge &&
+    bridge.getAllowedWebViewHosts().some((host) => {
+      const escapedHost = host.replace(/\./g, '\\.')
+      const isDomainOrSubDomain = new RegExp(`^(.+\\.)*${escapedHost}$`)
+      return !!isDomainOrSubDomain.exec(hostname)
```
bcaudan, line 36: **🥜 nitpick:** ‏
```suggestion
      return isDomainOrSubDomain.test(hostname)
```

bcaudan (approved):

PR #1521, opened 2022-05-09
🔊  [RUMF-1038] Add failed sendBeacon telemetry

amortemousque:
## Motivation

When  RUM / Logs / Replay send requests at the same time with sendBeacon then collision may happen.
To mitigate these collisions, we plan to lower batch sizes. 

To see the impact of lower batch size, this PR collects logs send beacon failure that will allow us: 

- To see how often collision happens
- To identify the amount of failed request correlated to the flush reason
- To see if collisions happens more often on slow connection 'slow-2g', '2g', '3g', or '4g'.

## Implementation strategy

Because collisions are most likely to happen at the `unload` if we were to send the logs straight away the request would be cancelled by the browser. To prevent it, I use the `localStorage` to store the logs and defer the sending. I choose  to use a `setTimeout` to avoid blocking the main thread at init 

## Changes

Add internal log in httpRequest

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/transport/httpRequest.ts
```diff
@@ -21,6 +22,8 @@ export class HttpRequest {
         if (isQueued) {
           return
         }
+
+        addFailedSendBeacon(this.endpointBuilder.endpointType, size, reason)
```
bcaudan, line 26: **❓ question:**  ‏how do we plan to differentiate collisions from other possible failure sources?

packages/core/src/transport/failedSendBeacon.ts
```diff
@@ -0,0 +1,32 @@
+import { isExperimentalFeatureEnabled } from '../domain/configuration'
+import { addMonitoringMessage, monitor } from '../domain/internalMonitoring'
+
+const FAILED_SEND_BEACON_FLUSH_INTERVAL = 2000
+
+export function startFlushFailedSendBeacons() {
+  if (!isExperimentalFeatureEnabled('lower-batch-size')) return
+
+  setInterval(monitor(flushFailedSendBeacon), FAILED_SEND_BEACON_FLUSH_INTERVAL)
```
bcaudan, line 9: **💭 thought:** ‏it could be interesting to add some explanations about your overall strategy, cases that you expect to handle or not to handle. Hard to guess why we need local storage or why we need a set interval.

packages/core/src/transport/batch.ts
```diff
@@ -149,15 +149,15 @@ export class Batch {
...
+          this.flush('visibility_hidden')
         }
       })
       /**
        * Safari does not support yet to send a request during:
        * - a visibility change during doc unload (cf: https://bugs.webkit.org/show_bug.cgi?id=194897)
        * - a page hide transition (cf: https://bugs.webkit.org/show_bug.cgi?id=188329)
        */
-      addEventListener(window, DOM_EVENT.BEFORE_UNLOAD, () => this.flush())
+      addEventListener(window, DOM_EVENT.BEFORE_UNLOAD, () => this.flush('before_unload'))
```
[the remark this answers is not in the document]
amortemousque: It's because I mainly want to know when we are loosing data:
- before_unload/visibility_hidden => data loss
- other cases => the fallback should works.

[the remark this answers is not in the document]
bcaudan: The PR description mention collision but IIUC we are collecting all failed requests, with every possible reason on replay but with only a subset of reason on logs/rum.
Here, you are mentioning data loss but we could have collisions without data loss, no?
I am a bit confused on what we are trying to achieve and how we plan to analyze the results out of it 😕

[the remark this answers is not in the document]
amortemousque: I updated the description

packages/core/src/transport/failedSendBeacon.ts
amortemousque: I've updated the PR description. Let me know what you think?

bcaudan: FMU, the updated PR description provide more information about our goal with this PR, however it does not provide more reasoning about your strategy in this file, cf

> Hard to guess why we need local storage or why we need a set interval.

amortemousque: I see. I added more precision in the `Implementation strategy` section of the description

bcaudan: ok, btw It could be even more useful in the file directly for future reference.

packages/core/src/transport/failedSendBeacon.ts
```diff
@@ -0,0 +1,38 @@
...
+// replaced at build time
+declare const __BUILD_ENV__SDK_VERSION__: string
+
+const FAILED_SEND_BEACON_FLUSH_INTERVAL = 2000
+export const LOCAL_STORAGE_KEY = 'datadog-browser-sdk-failed-send-beacon'
+
+export function startFlushFailedSendBeacons() {
+  if (!isExperimentalFeatureEnabled('lower-batch-size')) return
+
+  setInterval(monitor(flushFailedSendBeacon), FAILED_SEND_BEACON_FLUSH_INTERVAL)
```
bcaudan, line 13: **💭 thought:** ‏it is a shame to always write/read the local storage while we only need it for unload/init. Moreover, a flush could also happen during the unload phase, no?
What about only writing in local storage at unload and only reading from it at init?

amortemousque: Done

packages/core/src/transport/failedSendBeacon.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// replaced at build time
+declare const __BUILD_ENV__SDK_VERSION__: string
+
+export const LOCAL_STORAGE_KEY = 'datadog-browser-sdk-failed-send-beacon'
+
+export function startFlushFailedSendBeacons() {
+  if (!isExperimentalFeatureEnabled('lower-batch-size')) return
+
+  setTimeout(monitor(flushFailedSendBeacon))
```
bcaudan, line 13: **💬 suggestion:** ‏what about adding a comment to explain this `setTimeout`?

packages/core/src/transport/failedSendBeacon.ts
```diff
@@ -0,0 +1,43 @@
...
+    JSON.stringify({
+      reason,
+      endpointType,
+      version: __BUILD_ENV__SDK_VERSION__,
+      connection: navigator.connection ? (navigator.connection as any).effectiveType : undefined,
+      onLine: navigator.onLine,
+      size,
+    })
+  )
+}
```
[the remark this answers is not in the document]
amortemousque: I made so that we only buffer logs in local storage on before_unload and visibility_hidden. I used both event because before_unload isn't reliable in all case (like mobile). Let me know what you think

packages/core/src/transport/failedSendBeacon.ts
amortemousque: sure

bcaudan (approved):

PR #1526, opened 2022-05-11
✨[RUMF-1253] add `tracingSampleRate` option

bcaudan:
## Motivation

APM will soon start to bill ingested traces, so add an option to allow customers to configure the percentage of requests they want to trace to control their costs.

## Changes

- add `tracingSampleRate` init option
- remove useless `x-datadog-sampled` tracing header
- set `x-datadog-sampling-priority` header value according to the draw
- remove tracing infos from not sampled requests

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -113,11 +115,11 @@ function getCrypto() {
...
 }
 
-function makeTracingHeaders(traceId: TraceIdentifier, spanId: TraceIdentifier): TracingHeaders {
+function makeTracingHeaders(traceId: TraceIdentifier, spanId: TraceIdentifier, traceSampled: boolean): TracingHeaders {
   return {
     'x-datadog-origin': 'rum',
     'x-datadog-parent-id': spanId.toDecimalString(),
-    'x-datadog-sampling-priority': '1',
+    'x-datadog-sampling-priority': traceSampled ? '1' : '0',
     'x-datadog-trace-id': traceId.toDecimalString(),
```
BenoitZugmeyer, line 123: **❓ question:** ‏FMU, if `x-datadog-sampling-priority` is `0`, the agent won't create a trace. So do we need the `x-datadog-parent-id` and `x-datadog-trace-id` in this case?

kayayarai (approved):

packages/rum-core/src/domain/tracing/tracer.ts
bcaudan: > FMU, if x-datadog-sampling-priority is 0, the agent won't create a trace.

FMU, the backend could still create the trace if it does not have enough but we will not have a root span in this case. We could not have the tracing headers but the idea was to have an implementation more close to the next step.
Don't hesitate to have a second look at Xavier's RFC.

BenoitZugmeyer: Ok I've taken another look and it makes sense. Maybe add a comment to give a bit of context?

bcaudan: sure

BenoitZugmeyer (approved):

PR #1553, opened 2022-05-19
⚡️ [RUMF-1038] Lower batch size limit

amortemousque:
## Motivation

Given  RUM / Logs / Replay
When  they send requests at the same time with sendBeacon
Then  collisions may happen

This PR experiment lower batch size limits to be under the sendBeacon limit when RUM / logs / replay are enabled and leave some space for third party apps.

## Changes

- Batch limit from: `60KB (replay) + 16KB (RUM) + 16KB (Logs) ` to `22KB (replay) + 10KB (RUM) + 10KB (Logs)`
- Remove SEND_BEACON_BYTES_LIMIT and only use SEGMENT_BYTES_LIMIT

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -105,7 +105,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
        * beacon payload max queue size implementation is 64kb
        * ensure that we leave room for logs, rum and potential other users
        */
-      batchBytesLimit: 16 * ONE_KILO_BYTE,
+      batchBytesLimit: 10 * ONE_KILO_BYTE,
 
       eventRateLimiterThreshold: 3000,
       maxInternalMonitoringMessagesPerPage: 15,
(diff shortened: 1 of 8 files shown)


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -105,7 +105,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
        * beacon payload max queue size implementation is 64kb
        * ensure that we leave room for logs, rum and potential other users
        */
-      batchBytesLimit: 16 * ONE_KILO_BYTE,
```
BenoitZugmeyer, line 108: **❓ question:** ‏Shouldn't this be behind the `lower-batch-size` FF?

amortemousque: Damn yes

BenoitZugmeyer (approved):

PR #1566, opened 2022-05-24
[RUMF-1280] collect click position

amortemousque:
## Motivation

To build clickmaps, the browser SDK needs to collect the action click position.

## Changes

- Created a `getSelectorFromElement` to build the CSS path selector. It uses strategies the same way we have in `getActionNameFromElement`
- Every `newClick` of `trackClickActions`:
  - adds `action.position`: Contains x and y click coordinates
  - adds `action.target.selector`: CSS selector path of the target element
  - adds `action.target.width` and `action.target.height`: Dimension of the target element used to compute relative coordinate in the backend

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
```diff
@@ -0,0 +1,52 @@
...
+  while (element && element.nodeName !== 'HTML' && strategy !== strategies.getIDSelector) {
+    for (strategy of objectValues(strategies)) {
+      const elementSelector = strategy(element)
+      if (elementSelector) {
+        targetElementSelector.unshift(elementSelector)
+        break
+      }
+    }
+    element = element.parentElement
+  }
```
bcaudan, line 49: **💭 thought:** It could be interesting to compare the implementation with a simpler approach using the strategies directly instead of iterating over an array of strategies. 
IIUC, we don't plan to add a bunch of other strategies so it could just be more explicit.

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
```diff
@@ -0,0 +1,52 @@
...
+      const siblings = Array.from(element.parentElement!.children).filter((child) => child !== element)
+      const classUniqueAmongSiblings = siblings.every(
+        (sibling) =>
+          sibling.tagName !== element.tagName ||
+          orderedClassList.some((className) => !sibling.classList.contains(className))
+      )
+
+      if (classUniqueAmongSiblings)
+        return `${element.tagName}${orderedClassList.map((className) => `.${className}`).join('')}`
+    },
```
bcaudan, line 25: **💭 thought:** ‏I had a quick look but it probably worth iterating on this part:

- tests seems a bit light
- not sure why sorting is needed (perf concern)
- what if `sibling.classList = [...orderedClassList, otherClass]`

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
amortemousque: Yes sure

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
amortemousque: Sorting is needed to always have the same css path string regardless of the class name order.

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
```diff
@@ -0,0 +1,47 @@
...
+    targetElementSelector.unshift(getClassSelector(element) || getPositionSelector(element))
+
+    element = element.parentElement
+  }
+
+  return targetElementSelector.join('>')
+}
+
+function getIDSelector(element: Element): string | undefined {
+  const isUnique = element.id && element.ownerDocument.body.querySelectorAll(`#${element.id}`).length === 1
```
BenoitZugmeyer, line 23: **🔨 warning:** ‏You should escape id and class names when using them as query selectors, see [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape)

BenoitZugmeyer, line 23: **🔨 warning:** ‏You should escape id and class names when using them as query selectors, see [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape)

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  if (isUnique) return `#${element.id}`
+}
+
+function getClassSelector(element: Element): string | undefined {
+  const orderedClassList = arrayFrom(element.classList).sort()
+  const siblings = arrayFrom(element.parentElement!.children).filter((child) => child !== element)
```
BenoitZugmeyer, line 30: **💬 suggestion:** ‏Don't use `arrayFrom` so much, as it has performances implications:
* for `orderedClassList`, use `arrayFrom`/`sort` only if you are actually using it (in the `classUniqueAmongSiblings` case)
* for `siblings`, use a for loop over children, you don't need an array for it

BenoitZugmeyer, line 30: **💬 suggestion:** ‏Don't use `arrayFrom` so much, as it has performances implications:
* for `orderedClassList`, use `arrayFrom`/`sort` only if you are actually using it (in the `classUniqueAmongSiblings` case)
* for `siblings`, use a for loop over children, you don't need an array for it

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (classUniqueAmongSiblings)
+    return `${element.tagName}${orderedClassList.map((className) => `.${className}`).join('')}`
+}
+
+function getPositionSelector(element: Element): string | undefined {
+  const isUniqueChild = !element.previousElementSibling && !element.nextElementSibling
+  if (isUniqueChild) return `${element.tagName}`
+
+  const index = arrayFrom(element.parentElement!.children).indexOf(element)
+  return `${element.tagName}:nth-of-type(${index + 1})`
```
BenoitZugmeyer, line 46: **🔨 warning:** ‏to compute `nth-of-type`, you should only take elements with the same tag name into account


**💬 suggestion:** ‏use a for-loop instead of creating an array for it

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
```diff
@@ -0,0 +1,47 @@
...
+  while (element && element.nodeName !== 'HTML') {
+    const idSelector = getIDSelector(element)
+    if (idSelector) {
+      targetElementSelector.unshift(idSelector)
+      break
+    }
+
+    targetElementSelector.unshift(getClassSelector(element) || getPositionSelector(element))
+
+    element = element.parentElement
```
BenoitZugmeyer, line 16: **❓ question:** ‏Maybe you could stop when the `targetElementSelector` is unique on the page? We can talk about it.

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -195,6 +203,20 @@ function newClick(
...
+
+  if (isExperimentalFeatureEnabled('clickmap')) {
+    target = {
+      selector: getSelectorFromElement(base.event.target),
+      width: base.event.target.offsetWidth,
+      height: base.event.target.offsetHeight,
+    }
+    position = {
+      x: base.event.offsetX,
+      y: base.event.offsetY,
```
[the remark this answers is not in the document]
amortemousque: How did you do your test? It tested on SVG element and it works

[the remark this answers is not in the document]
BenoitZugmeyer: Does it report the right value? Like, related to the SVG element?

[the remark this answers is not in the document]
amortemousque: Interesting, the offset values are relative to the <svg> element even when clicking in a nested element

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { getSelectorFromElement } from './getSelectorFromElement'
+
+describe('getSelectorFromElement', () => {
+  const iframes: HTMLIFrameElement[] = []
+
+  function element(s: TemplateStringsArray) {
+    // Simply using a DOMParser does not fit here, because script tags created this way are
+    // considered as normal markup, so they are not ignored when getting the textual content of the
+    // target via innerText
```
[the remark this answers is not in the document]
amortemousque: Yes, I created a test/createIsolatedDom.ts

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -89,7 +97,7 @@ export function trackClickActions(
     }
   }
 
-  function processClick(event: MouseEvent & { target: Element }) {
+  function processClick(event: MouseEvent & { target: HTMLElement }) {
```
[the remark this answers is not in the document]
amortemousque: Yep right. Fixed

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -195,6 +203,20 @@ function newClick(
...
 ) {
   const id = generateUUID()
+  let target: ClickAction['target']
+  let position: ClickAction['position']
+
+  if (isExperimentalFeatureEnabled('clickmap')) {
+    target = {
+      selector: getSelectorFromElement(base.event.target),
+      width: base.event.target.offsetWidth,
+      height: base.event.target.offsetHeight,
```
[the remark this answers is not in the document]
amortemousque: Yes right

packages/rum-core/src/rumEvent.types.ts
```diff
@@ -51,6 +63,20 @@ export type RumActionEvent = CommonProperties & {
       readonly type: ('rage_click' | 'dead_click' | 'error_click')[]
       [k: string]: unknown
     }
+    /**
+     * Action position properties
+     */
+    readonly position?: {
```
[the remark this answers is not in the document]
amortemousque: I can update the rum-event-format comment to be more specific. 
Wdyt?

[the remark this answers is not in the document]
BenoitZugmeyer: It would be good, yes 🙂

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorFromElement.ts
amortemousque: yes lets talk

BenoitZugmeyer (approved):

PR #1567, opened 2022-05-24
[RUMF-1267] resolve remaining cyclic dependencies related to telemetry

BenoitZugmeyer:
## Motivation

Remove cyclic dependencies. Followup of https://github.com/DataDog/browser-sdk/pull/1559 , focuses on solving dependency cycle related to telemetry

## Changes

* make sure `display` does not depend on monitor functions
* split monitor functions from the `telemetry` module
* enable the `import/no-cycle` eslint rule

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -9,6 +9,7 @@ import { computeStackTrace } from '../tracekit'
 import { Observable } from '../../tools/observable'
 import { timeStampNow } from '../../tools/timeUtils'
 import { ConsoleApiName } from '../console/consoleObservable'
+import { startMonitorErrorCollection } from '../monitor'
```
bcaudan, line 12: **💭 thought:** ‏about `monitor` location, I am still a bit bothered about `tools/utils` requiring something from `domain`. Not sure about moving specific utilities to `monitor` or `moving` monitor to `tools`.
wdyt?

BenoitZugmeyer: I'm fine with moving `monitor` to `tools`. I'll do that!

packages/core/src/tools/monitor/index.ts
```diff
@@ -0,0 +1 @@
+export * from './monitor'
```
amortemousque, line 1: Do we need a folder for monitor?

packages/core/src/tools/display.ts
```diff
@@ -7,7 +7,15 @@
  * but we should be safe from infinite loop nonetheless.
  */
 
-import { ConsoleApiName } from '../domain/console/consoleObservable'
+export const ConsoleApiName = {
```
amortemousque, line 10: I'd find more logic to have it in `domain/console/consoleApiNames.ts`
Wdyt?

amortemousque: Hum, okay then we can keep it here. I'm not sure having it in `browser/consoleApiNames` will make things clearer. 

Note for later: If we really don't want to have domain dependencies out of `/domain` it could be nice to add an eslint rule.

amortemousque (approved):

PR #1569, opened 2022-05-25
🔊 [RUMF-1038] Add failed sendbeacon replay reason

amortemousque:
## Motivation

For the failed sendbeacon investigation the flush reason of replay were missing

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


BenoitZugmeyer (commented):
**💬 suggestion:** ‏There is no need to send the reason to the worker. You could simply call  `send(data, segment.metadata, rawSegmentBytesCount, segment.reason)`  [here](https://github.com/DataDog/browser-sdk/blob/5c1bf1ea7c1d0a9da033ebf0f196274eaa6d70a8/packages/rum/src/domain/segmentCollection/segmentCollection.ts#L142)

Also I would call it `flushReason` to be more explicit

BenoitZugmeyer (approved):

PR #1577, opened 2022-05-30
🐛 [RUMF-1267] remove last circular dependencies

BenoitZugmeyer:
## Motivation

Following https://github.com/DataDog/browser-sdk/pull/1559 and https://github.com/DataDog/browser-sdk/pull/1567, this PR removes some more circular dependencies, as well as taking the opportunity to clean recorder types a bit.

Those two last depency loops did not have been caught because of [an issue with `import/no-cycle` rule 
](https://github.com/import-js/eslint-plugin-import/issues/2461). They have been caught by importing the RUM SDK into a small rollup project.

## Changes

* Remove dependency loop related to eventBridge
* Remove dependency loop related to recorder types
* Reorganize and rename some recorder types

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum/src/types/record.ts
```diff
@@ -0,0 +1,236 @@
+import type { TimeStamp } from '@datadog/browser-core'
```
amortemousque, line 1: I like the split of the types but now that I'm looking at it, it feels a bit weird to have `domain/record/types.ts` and `/types/record.ts`. 
Either we collocate the types to the module or use a `types` folder. My personal preference is to collocate the type with the module but it's not a strong opinion. 
Wdyt

amortemousque (approved):

PR #1584, opened 2022-06-08
✨ [RUMF-1288] Collect viewport size

amortemousque:
## Motivation

When looking at clickmaps users need to be able to filter events by viewport size.
I also took to chance to clean a bit RUM demain folder by moving context files in a dedicated folder.

## Changes

- Move context files in a dedicated folder
- Collocate context types with their modules
- Add displayContext

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/assembly.ts
```diff
@@ -121,6 +122,7 @@ export function startRumAssembly(
           },
           synthetics: syntheticsContext,
           ci_test: ciTestContext,
+          display: getDisplayContext(),
```
amortemousque, line 125: Getting the display context at the assembly stage is the more straightforward way to implement it.
However, for pre-init calls, it means we won't get the display info of the time the event happens, but at the init time. I think it's good enough though

packages/rum-core/src/domain/contexts/displayContext.ts
```diff
@@ -0,0 +1,12 @@
+import { isExperimentalFeatureEnabled } from '@datadog/browser-core'
+
+export function getDisplayContext() {
+  if (!isExperimentalFeatureEnabled('clickmap')) return
+
+  return {
+    viewport: {
+      width: window.innerWidth,
+      height: window.innerHeight,
```
BenoitZugmeyer, line 9: **💬 suggestion:** ‏Apparently getting the window dimension might force the browser to recompute the layout (cf list [here](https://gist.github.com/paulirish/5d52fb081b3570c81e3a)). Maybe listen for `resize` event and cache its value?

amortemousque: Indeed, good.  Do you think we should do the same [here](https://github.com/DataDog/browser-sdk/blob/336ebe7c6cd3dc9c87499cbf9cac6bb5e8fe6da8/packages/rum/src/domain/record/observer.ts#L223-L234)

BenoitZugmeyer: that's already what we do :) (but also we throttle the listener)

amortemousque: Nice!

BenoitZugmeyer (approved):

abramjstamper:
I've checked the documentation [here](https://docs.datadoghq.com/real_user_monitoring/). What configuration of `@datadog/browser-rum` is required to enable getting the viewport size sent to DataDog? I was going to implement a crudely made custom metric and send the data myself, but would rather use a natively-available Data Dog feature. Thanks

PR #1588, opened 2022-06-10
♻️ Simplify RUM assembly

amortemousque:
## Motivation

RUM assembly has a quite convoluted logic to handle the various contexts. More precisely:
- All context extends `Context` making us beleave they can contain any attribute (`[x: string]: ContextValue`) which is false.
- It's hard to follow the overridden logic of context with the [combine](https://github.com/DataDog/browser-sdk/blob/336ebe7c6cd3dc9c87499cbf9cac6bb5e8fe6da8/packages/rum-core/src/domain/assembly.ts#L134-L138)
- It's hard to understand the [service and version logic](https://github.com/DataDog/browser-sdk/blob/336ebe7c6cd3dc9c87499cbf9cac6bb5e8fe6da8/packages/rum-core/src/domain/assembly.ts#L127-L132)

## Changes

- Only RumContext now extends Context
- Remove urlContext, viewContext, { action: { id: actionId } } from the combine and use simple assignment
- Remove convoluted viewContext service and version logic and use simple assignment

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -125,17 +125,17 @@ export function startRumAssembly(
             type: syntheticsContext ? SessionType.SYNTHETICS : ciTestContext ? SessionType.CI_TEST : SessionType.USER,
           },
           view: {
-            id: viewContext.view.id,
-            name: viewContext.view.name,
-            url: urlContext.view.url,
-            referrer: urlContext.view.referrer,
+            id: viewContext.id,
+            name: viewContext.name,
+            url: urlContext.url,
+            referrer: urlContext.referrer,
           },
           action: needToAssembleWithAction(rawRumEvent) && actionId ? { id: actionId } : undefined,
           synthetics: syntheticsContext,
           ci_test: ciTestContext,
         }
 
-        const serverRumEvent = combine(rumContext, rawRumEvent) as RumEvent & Context
+        const serverRumEvent = combine(rumContext as RumContext & Context, rawRumEvent) as RumEvent & Context
         serverRumEvent.context = combine(commonContext.context, customerContext)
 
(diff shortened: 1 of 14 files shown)


packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -181,7 +181,7 @@ export type RawRumEvent =
   | RawRumLongTaskEvent
   | RawRumActionEvent
 
-export interface RumContext {
+export interface RumContext extends Context {
```
BenoitZugmeyer, line 184: **💬 suggestion:** ‏ we need to extend `Context` here just to work around the `Index signature for type 'string' is missing in type ...` error, but `RumContext` is never actually used with an index, except when using `combine`.

Maybe we could remove the `extends` here, and cast the object as `RumContext & Context` where `RumContext` is used with `combine`. WDYT?

I know this is a bit confusing, and there is a [long standing TS issue](
https://github.com/microsoft/TypeScript/issues/15300) related to that.

packages/rum-core/src/domain/assembly.ts
```diff
@@ -113,30 +115,27 @@ export function startRumAssembly(
...
           source: 'browser',
           session: {
             id: session.id,
             type: syntheticsContext ? SessionType.SYNTHETICS : ciTestContext ? SessionType.CI_TEST : SessionType.USER,
           },
+          view: {
+            id: viewContext.view.id,
+            name: viewContext.view.name,
+            url: urlContext.view.url,
+            referrer: urlContext.view.referrer,
```
BenoitZugmeyer, line 131: **🥜 nitpick:** ‏maybe remove the `view` extra-level from those contexts

PR #1593, opened 2022-06-15
✨ Feature flag sampling

amortemousque:
## Motivation

While experimenting the lower batch size, I was wondering "**how to test the impact of a change in a short period of time**"
The answer is that  I created a `sampleExperimentalFeature` which enable a feature flag only for X% of users. This way I can gather data with and without the feature enabled as the same time.

## Changes

- Create `sampleExperimentalFeature`
- Sample 'lower-batch-size' ff by 50%

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.spec.ts
+++ b/packages/core/src/domain/configuration/configuration.spec.ts
@@ -136,8 +136,8 @@ describe('validateAndBuildConfiguration', () => {
     })
 
     it('should be set to 10KB when lower-batch-size is enabled', () => {
-      updateExperimentalFeatures(['lower-batch-size'])
       setSampledExperimentalFeatures({ 'lower-batch-size': 100 })
+      updateExperimentalFeatures(['lower-batch-size'])
 
       const configuration = validateAndBuildConfiguration({ clientToken })!
       expect(configuration.batchBytesLimit).toEqual(10 * ONE_KILO_BYTE)
--- a/packages/core/src/domain/configuration/configuration.spec.ts
+++ b/packages/core/src/domain/configuration/configuration.spec.ts
@@ -6,7 +6,6 @@ import { validateAndBuildConfiguration } from './configuration'
 import {
   isExperimentalFeatureEnabled,
   resetExperimentalFeatures,
-  setSampledExperimentalFeatures,
   updateExperimentalFeatures,
 } from './experimentalFeatures'
 
@@ -136,7 +135,6 @@ describe('validateAndBuildConfiguration', () => {
     })
 
     it('should be set to 10KB when lower-batch-size is enabled', () => {
-      setSampledExperimentalFeatures({ 'lower-batch-size': 100 })
       updateExperimentalFeatures(['lower-batch-size'])
 
       const configuration = validateAndBuildConfiguration({ clientToken })!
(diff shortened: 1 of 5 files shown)


packages/core/src/domain/configuration/experimentalFeatures.ts
```diff
@@ -25,9 +28,29 @@ export function updateExperimentalFeatures(enabledFeatures: string[] | undefined
 }
 
 export function isExperimentalFeatureEnabled(featureName: string): boolean {
-  return !!enabledExperimentalFeatures && enabledExperimentalFeatures.has(featureName)
+  return (
+    !!enabledExperimentalFeatures &&
+    enabledExperimentalFeatures.has(featureName) &&
+    isExperimentalFeatureSampled(featureName)
```
BenoitZugmeyer, line 34: **💬 suggestion:** ‏Instead of checking if the FF is sampled at every use, you could just:
* set the sample rate before calling `updateExperimentalFeature`
* in `updateExperimentalFeature`, make sure to add only sampled FF to `enableExperimentalFeatures`

amortemousque: By doing so the `sampleExperimentalFeature` as to be called before  the `updateExperimentalFeatures` while currently it can be called at any time. Wdyt?

BenoitZugmeyer: Yes, but I don't think we'll ever want to change the sample rate once it's set. I think it's fine.

packages/core/src/domain/configuration/configuration.spec.ts
```diff
@@ -136,6 +137,8 @@ describe('validateAndBuildConfiguration', () => {
 
     it('should be set to 10KB when lower-batch-size is enabled', () => {
       updateExperimentalFeatures(['lower-batch-size'])
+      setSampledExperimentalFeatures({ 'lower-batch-size': 100 })
```
BenoitZugmeyer, line 140: **🔨 warning:** ‏this should be set before update

BenoitZugmeyer (approved):

packages/core/src/domain/configuration/experimentalFeatures.spec.ts
```diff
@@ -37,3 +38,14 @@ describe('experimentalFeatures', () => {
     expect(isExperimentalFeatureEnabled('foo')).toBeTrue()
   })
 })
+
+describe('setSampledExperimentalFeatures', () => {
+  it('should sample experimental features', () => {
+    setSampledExperimentalFeatures({ foo: 100, bar: 0 })
+    updateExperimentalFeatures(['foo', 'bar', 'baz'])
```
bcaudan, line 45: **💭 thought:** ‏what about allowing us to configure the sampling rate from the init configuration?
```ts
enableExperimentalFeatures: [{foo: 100}, {bar: 0}, 'baz']
```
it would:

- be closer to what other DD feature flags can do
- allow us to change the rate without updating SDK code

cc @BenoitZugmeyer

amortemousque: Sounds nice. Lets do it

amortemousque: Done! Let me know what you think

BenoitZugmeyer: Yes it looks nice! A follow-up could be to list enabled FF in telemetry events, so we can compare A/B data.

BenoitZugmeyer (approved):

PR #1594, opened 2022-06-15
🔊  Clear batch before send to allow telemetry in httpRequest.send

amortemousque:
## Motivation

During my investigation on lower batch size, since the new telemetry is enable, some of the telemetry events were missing.
It appears that telemetry events added inside httpRequest.send were remove by the batch. To prevent it, lets clear the batch before we use `httpRequest.send`

## Changes

clear the batch before we use `httpRequest.send`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/transport/batch.spec.ts
```diff
@@ -150,4 +150,21 @@ describe('batch', () => {
...
   })
+
+  it('should be able to use telemetry in the httpRequest.send', () => {
+    const fakeRequest = {
+      send(data: string) {
+        addTelemetryDebugFake()
+        transport.send(data, BATCH_BYTES_LIMIT)
+      },
+    } as unknown as HttpRequest
+    const batch = new Batch(fakeRequest, 10, 100, 100, 100)
```
bcaudan, line 161: **💬 suggestion:** ‏unless we care about having those exact values for the batch, we could use the global constants

```suggestion
    const batch = new Batch(fakeRequest, BATCH_MESSAGES_LIMIT, BATCH_BYTES_LIMIT, MESSAGE_BYTES_LIMIT, FLUSH_TIMEOUT)
```

bcaudan (approved):

PR #1601, opened 2022-06-20
✨[RUMF-1225] Enable service and version update on startView

amortemousque:
## Motivation

In the context of sub-apps, web and mobile developers want to only see RUM events happening in their area of ownership. 
For example, a developer working on the Log Management product at Datadog wants to see RUM events that happened in the Log Management only. To do so we make the service attribute first-class-citizen in RUM and allow customers to update the current `service` and `version` with the `startView` API.

## Changes

Remove the `sub-app` feature flag

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/boot/rumPublicApi.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.ts
@@ -156,8 +156,8 @@ export function makeRumPublicApi(
 
   const startView: {
     (name?: string): void
-    // (options: ViewOptions): void // uncomment when removing the feature flag
-  } = monitor((options?: string) => {
+    (options: ViewOptions): void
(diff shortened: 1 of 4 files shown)


BenoitZugmeyer (approved):

ashishmahi:
is there way to reset the service names to whatever it was initialized with ? for example, I've shell app A with all variables set to "A" initially now app B/C/D.. would change service name in their render method and before they get unmounted they can just reset the variables with a method like .resetView without need of knowing what my parent variables were.

bcaudan:
Hi @ashishmahi, please open a new issue to start a new discussion.

PR #1629, opened 2022-07-06
🐛 [RUMF-1327] rum synthetics: fix logs session conflict

BenoitZugmeyer:
## Motivation

We identified an issue when the RUM SDK is injected by the Synthetics Worker and the Web application is using the Logs SDK: cookie options might be different, and the session cookie accesses have unexpected behaviors.

## Changes

To avoid that, this PR makes sure that the Logs SDK does not create a cookie session when the RUM SDK will be injected by the Synthetics Worker

In this context, the Logs SDK will rely on the session created by the inject RUM SDK

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/domain/assembly.ts
+++ b/packages/logs/src/domain/assembly.ts
@@ -1,5 +1,5 @@
-import type { Context, EventRateLimiter, RawError, RelativeTime } from '@datadog/browser-core'
-import { ErrorSource, combine, createEventRateLimiter, getRelativeTime } from '@datadog/browser-core'
+import type { Context, EventRateLimiter, RawError, RelativeTime} from '@datadog/browser-core';
+import { willSyntheticsInjectRum , ErrorSource, combine, createEventRateLimiter, getRelativeTime } from '@datadog/browser-core'
 import type { CommonContext } from '../rawLogsEvent.types'
 import type { LogsConfiguration } from './configuration'
 import type { LifeCycle } from './lifeCycle'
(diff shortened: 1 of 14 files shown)


bcaudan (approved):
Great 👍

packages/rum-core/test/specHelper.ts
```diff
@@ -277,46 +271,6 @@ export const noopRecorderApi: RecorderApi = {
   getReplayStats: () => undefined,
 }
 
-// Duration to create a cookie lasting at least until the end of the test
-const COOKIE_DURATION = ONE_MINUTE
-
-export function mockSyntheticsWorkerValues(
```
liywjl, line 283: **👏 praise:** nice job lifting this, I feel this file is a little long

packages/logs/src/domain/assembly.ts
```diff
@@ -63,10 +72,40 @@ export function startLogsAssembly(
...
+    return context
+  }
+
+  return getFromGlobal(browserWindow.DD_RUM)
+
+  function getFromGlobal(rum?: Rum): Context | undefined {
+    if (rum && rum.getInternalContext) {
+      return rum.getInternalContext(startTime)
+    }
+  }
```
liywjl, line 106: **💭 thought:**
Do we interchangeably use `Global` and `Context`? If not, maybe might be worth renaming the function name to explicitly mention `Context`?

Also we can use TS `?` to remove the need for `if` checks
```suggestion
  const getFromGlobal = (rum?: Rum): Context | undefined => rum?.getInternalContext(startTime)
```
(Note: if using `const` link in the example, we would need to lift where this function is defined)

liywjl (approved):
👏

packages/logs/src/domain/assembly.ts
BenoitZugmeyer: * I agree that I could have been more explicit. I pushed a commit to address that

* I tend to avoid the using `?.` operator because I observed that it produces more code than necessary when transpiled. If we compare the two approaches, transpiled + minimized:

```
// current implementation
function n(n){if(n&&n.getInternalContext)return n.getInternalContext()}
// your proposal
var t=function(n){var t;return null===(t=null==n?void 0:n.getInternalContext)||void 0===t?void 0:t.call(n)};
```

It's not a strong rule yet, but still something to keep in mind.

PR #1631, opened 2022-07-07
🐛 frustration signals: track window open

BenoitZugmeyer:
## Motivation

When the user clicks and the application is using `window.open` to open an url in a new tab/window, a dead click is reported as we are not aware of it.

## Changes

Track `window.open` calls and consider them as "page activity", in the same spirit as DOM mutations.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


bcaudan (approved):
Nice!

PR #1633, opened 2022-07-08
[RUMF-1328] better data transfer with retry and throttling

amortemousque:
## Motivation

This is a draft PR of what the data transfer might look like with the retry and throttling strategies.
So it:
- assume httRequest.send already switch from sendbeacon to fetch keepalive and return an observable
- assume a function`isFailedRequest(response)` exists and check is the http status code tell the intake is unavailable
- does not handle buffer bytesLimit

The goal is to agree on the target design to then implementing it incrementally like so:
- Implement the switch from sendbeacon to fetch keepalive
- Implement the request throttling
- Implement the retry strategy

## Changes

- Add a httpRetry module
- Add sendWithRetry in httpRequest
- Use sendWithRetry in batch

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -1,6 +1,8 @@
-import type { EndpointBuilder } from '../domain/configuration'
+import type { Configuration, EndpointBuilder } from '../domain/configuration'
 import { addTelemetryError } from '../domain/telemetry'
+import type { Observable } from '../tools/observable'
 import { addFailedSendBeacon } from './failedSendBeacon'
+import { httpRetry } from './httpRetry'
 
 /**
  * Use POST request without content type to:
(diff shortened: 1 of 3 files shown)


packages/core/src/transport/httpRetry.ts
```diff
@@ -0,0 +1,133 @@
+import type { TimeoutId } from '../tools/utils'
+import { ONE_SECOND } from '../tools/utils'
+import type { Observable } from '../tools/observable'
+import { monitor } from '../tools/monitor'
+
+const MIN_RETRY_INTERVAL = ONE_SECOND
+const MAX_RETRY_INTERVAL = 8 * ONE_SECOND
+
+export function httpRetry(
+  send: (data: string | FormData, bytesCount: number) => Observable<Response>,
```
amortemousque, line 10: In this PR I assume `send` already uses fetch keepalive and return an `Observable`

packages/core/src/transport/httpRequest.ts
```diff
@@ -11,11 +13,15 @@ import { addFailedSendBeacon } from './failedSendBeacon'
  * to be parsed correctly without content type header
  */
 export class HttpRequest {
-  constructor(private endpointBuilder: EndpointBuilder, private bytesLimit: number) {}
+  sendWithRetry: (data: string | FormData, bytesCount: number) => Observable<Response>
+
+  constructor(private endpointBuilder: EndpointBuilder, private configuration: Configuration) {
+    ;({ retried: this.sendWithRetry } = httpRetry(this.send, configuration))
```
bcaudan, line 19: **❗ warning:** ‏that does not seem right:

- httpRetry depends on a HttpRequest method 
- httpRequest depends on httpRetry

packages/core/src/transport/httpRetry.ts
```diff
@@ -0,0 +1,133 @@
...
+  configuration: {
+    bufferBytesLimit: number
+    ongoingBytesLimit: number
+    ongoingRequestsLimit: number
+  }
+) {
+  const buffer: Array<{ data: string | FormData; bytesCount: number }> = []
```
bcaudan, line 17: **💭 thought:** ‏Extracting a type for `{ data: string | FormData; bytesCount: number }` and also use it in send functions would be great

bcaudan:
closed in favor of #1700

PR #1640, opened 2022-07-13
Setup Code Scanning on browser-sdk

ganeshkumarsv:
Static code analysis on browser-sdk using Github's Code Scanning.

This PR has Github workflow configs to trigger the Code Scanning on every PR and on push to master branch. As of now only default security based CodeQL queries are configured. We can gradually add queries for Code Quality later.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,49 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
BenoitZugmeyer, line 20: **💭 thought:** ‏since we are now using GitHub action, we might want to setup dependabot to check for actions updates ([example](https://github.com/BenoitZugmeyer/eslint-plugin-html/blob/main/.github/dependabot.yml#L8-L12))

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

.github/workflows/codeql-analysis.yml
bcaudan: 👍, let's do that in a following PR

Issue #1582, opened 2022-06-06
Can't append user information with `setUser`


annerajb:
Right now [setUser](https://github.com/DataDog/browser-sdk/blob/cb1918e8a94a9fdabef9fcc3396a2bd0ddf8ad62/packages/rum-core/src/boot/rumPublicApi.ts#L203)

Does not have an easy way to append data to a user.

What is the recommended way?

BenoitZugmeyer:
There is not, sorry. We'll consider it! Does something like `mergeUser()` would sound good to you?

annerajb:
It does

liywjl:
Hello @annerajb,

We have have as of v4.17.0 added new methods to update the Session User Object. 

You can find these APIs documented [here](https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm#user-session).

The APIs most relevant to updating specific properties are [`setUserProperty`](https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm#addoverride-user-session-property) and [`removeUserProperty`](https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm#remove-user-session-property).

Don't hesitate to open another ticket if you have a specific issue related to the SDK or contact [datadog support](https://www.datadoghq.com/support/) directly.

PR #1641, opened 2022-07-15
✨ [RUMF-1315] Extend user object methods

liywjl:
## Motivation

Wanted to extend methods used to interact with user object

https://github.com/DataDog/browser-sdk/issues/1582

## Challenges

- Deprecate methods
Expanding user object methods required us to deprecate some methods in favour of new ones. 

- Update shared source of truth
The user object shares similar code to `rumGlobalContext` and `loggerGlobalContext`. Therefore the same method depreciation needed to be applied.

## Updated APIs

| user  |                              | rumGlobalContext   |                                 | loggerGlobalContext   |   |
|---|---|---|---|---|---|
|current              |deprecated      |current   | deprecated                                |current   |deprecated   |
| setUser             |                | setGlobalContext  |  setRumGlobalContext             | setGlobalContext  | setLoggerGlobalContext  |
| getUser             |                | getGlobalContext  | getRumGlobalContext              | getGlobalContext  | getLoggerGlobalContext   |
| setUserProperty     |                | setGlobalContextProperty  | addRumGlobalContext      | setGlobalContextProperty  | addLoggerGlobalContext  |
| removeUserProperty  |                | removeGlobalContextProperty | removeRumGlobalContext | removeLoggerGlobalContext  | removeGlobalContextProperty   |
| clearUser           | removeUser     | clearGlobalContext  |                                | clearGlobalContext  |   |

## Changes

- Extend `contextManager` APIs
- Use `contextManager` for user object APIs
- Expose new APIs to mutate user object
- Expand `globalContextManager` API to use new `contextManager` APIs
- Deprecate old `contextManager` APIs
- Remove defensive code used to prevent exposing context reference (`deepClone` is now done by `contextManager`)

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -203,14 +203,24 @@ export function makeRumPublicApi(
...
-      user = {}
+      userContextManager.set({})
+    }),
+
+    addUserAttribute: monitor((attribute, value) => {
+      userContextManager.add(attribute, value)
+    }),
+
+    removeUserAttribute: monitor((attribute) => {
+      userContextManager.remove(attribute)
```
amortemousque, line 223: **💬 suggestion:** ‏I think it's worth having tests to cover these new API

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,128 @@ describe('rum public api', () => {
...
+
+    beforeEach(() => {
+      addActionSpy = jasmine.createSpy()
+      rumPublicApi = makeRumPublicApi(
+        () => ({
+          ...noopStartRum(),
+          addAction: addActionSpy,
+        }),
+        noopRecorderApi
+      )
```
BenoitZugmeyer, line 471: **💬 suggestion:** ‏`addActionSpy` isn't used, you should simplify this:

```suggestion
      rumPublicApi = makeRumPublicApi(
        noopStartRum,
        noopRecorderApi
      )
```

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,128 @@ describe('rum public api', () => {
...
+      setupBuilder = setup()
```
BenoitZugmeyer, line 472: **💬 suggestion:** ‏`setupBuilder` isn't used, you should remove it

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,128 @@ describe('rum public api', () => {
...
+    afterEach(() => {
+      setupBuilder.cleanup()
+    })
+
+    it('should remove attribute', () => {
+      rumPublicApi.setUser(user)
+      rumPublicApi.removeUserAttribute('foo')
+      const userClone = rumPublicApi.getUser()
+      const newUser = { ...user }
+      delete newUser.foo
```
BenoitZugmeyer, line 575: **🥜 nitpick:** ‏you shouldn't mutate the variable declared outside of `beforeEach` or `it`, because it might be used by other test cases. It is not the case there, as there is a single test case in the `describe`, but still a bad practice

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,40 +217,47 @@ export function makeRumPublicApi(
...
+    setUserProperty: monitor(userContextManager.setContextProperty),
+
+    removeUserProperty: monitor(userContextManager.removeContextProperty),
+
+    clearUser: monitor(userContextManager.clearContext),
+
     removeUser: monitor(() => {
-      user = {}
+      userContextManager.clearContext()
     }),
```
BenoitZugmeyer, line 237: **💬 suggestion:** ‏`removeUser: monitor(userContextManager.clearContext)` for consistency?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,40 +217,47 @@ export function makeRumPublicApi(
...
 
+    getUser: monitor(userContextManager.getContext),
+
+    setUserProperty: monitor(userContextManager.setContextProperty),
+
+    removeUserProperty: monitor(userContextManager.removeContextProperty),
+
+    clearUser: monitor(userContextManager.clearContext),
+
     removeUser: monitor(() => {
```
BenoitZugmeyer, line 235: **💬 suggestion:** ‏add a `@deprecated` comment above the `removeUser` function

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,40 +217,47 @@ export function makeRumPublicApi(
...
-      user = {}
+      userContextManager.clearContext()
     }),
 
+    addUserAttribute: monitor(userContextManager.setContextProperty),
+
+    removeUserAttribute: monitor(userContextManager.removeContextProperty),
+
```
BenoitZugmeyer, line 242: **🔨 warning:** ‏those should be removed

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,40 +217,47 @@ export function makeRumPublicApi(
...
+    if ('name' in newUser) {
+      newUser.name = String(newUser.name)
     }
-    if ('email' in result) {
-      result.email = String(result.email)
+    if ('email' in newUser) {
+      newUser.email = String(newUser.email)
     }
-    return result
+    return newUser
```
BenoitZugmeyer, line 260: **🔨 warning:** ‏you should not mutate the variable coming from outside. Also, it could be nice to sanitize values set with `setUserProperty` as well?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -129,7 +134,7 @@ export function makeRumPublicApi(
     const startRumResults = startRumImpl(
       configuration,
       () => ({
-        user,
+        user: userContextManager.get(),
         context: globalContextManager.get(),
```
bcaudan, line 138: **❓ question:** ‏ ‏Should we not use those `get` methods or not deprecate them?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,17 +217,22 @@ export function makeRumPublicApi(
...
 
     setUser: monitor((newUser: User) => {
-      const sanitizedUser = sanitizeUser(newUser)
-      if (sanitizedUser) {
-        user = sanitizedUser
-      } else {
+      if (typeof newUser !== 'object' || !newUser) {
         display.error('Unsupported user:', newUser)
+      } else {
+        userContextManager.setContext(sanitizeUser(newUser as Context))
```
bcaudan, line 223: **💬 suggestion:** ‏I's a bit a shame to perform two deep clones here. What about doing a shallow clone just for the sanitization (in order to not overwriting customer `id`, `name` or `email`) and let the `setContext` perform the deep clone?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,17 +217,22 @@ export function makeRumPublicApi(
...
       }
     }),
 
-    removeUser: monitor(() => {
-      user = {}
-    }),
+    getUser: monitor(userContextManager.getContext),
+
+    setUserProperty: monitor((key, property) => userContextManager.setContextProperty(key, sanitizeUser(property))),
```
bcaudan, line 229: **❗ warning:** ‏sanitizing a user property is probably not the same as sanitizing a user.

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,85 @@ describe('rum public api', () => {
...
+      const userClone = rumPublicApi.getUser()
+      const userClone2 = rumPublicApi.getUser()
+
+      expect(userClone).not.toBe(user)
+      expect(userClone).not.toBe(userClone2)
+      expect(userClone).toEqual(user)
+    })
+  })
+
+  describe('setUserProperty', () => {
```
bcaudan, line 482: **💬 suggestion:** ‏you should add a test case about sanitization

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,85 @@ describe('rum public api', () => {
...
+    it('should override attribute', () => {
+      rumPublicApi.setUser(user)
+      rumPublicApi.setUserProperty('foo', addressAttribute)
+      const userClone = rumPublicApi.getUser()
+
+      expect(userClone).toEqual({ ...user, foo: addressAttribute })
+    })
+  })
+
+  describe('removeUserAttribute', () => {
```
bcaudan, line 518:
```suggestion
  describe('removeUserProperty', () => {
```

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,85 @@ describe('rum public api', () => {
...
+
+    it('should remove property', () => {
+      const user: Context = { id: 'foo', name: 'bar', email: 'qux', foo: { bar: 'qux' } }
+
+      rumPublicApi.setUser(user)
+      rumPublicApi.removeUserProperty('foo')
+      const userClone = rumPublicApi.getUser()
+      const newUser = { ...user }
+      delete newUser.foo
+      expect(userClone).toEqual(newUser)
```
bcaudan, line 533: **💬 suggestion:** ‏You are replicating the implementation in your test
 ```suggestion
      expect(userClone.foo).toBeUndefined()
```

File: packages/rum-core/src/boot/rumPublicApi.spec.ts [hunk 15]
```diff
@@ -455,6 +455,85 @@ describe('rum public api', () => {
...
+    beforeEach(() => {
+      rumPublicApi = makeRumPublicApi(noopStartRum, noopRecorderApi)
+    })
+
+    it('should add attribute', () => {
+      rumPublicApi.setUser(user)
+      rumPublicApi.setUserProperty('address', addressAttribute)
+      const userClone = rumPublicApi.getUser()
+
+      expect(userClone).toEqual({ ...user, address: addressAttribute })
```
bcaudan, line 496: **💬 suggestion:** ‏do we need to check the whole object?
```suggestion
      expect(userClone.address).toEqual(addressAttribute)
```

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,85 @@ describe('rum public api', () => {
...
+    })
+
+    it('should not contain original reference to object', () => {
+      const userDetails: { [key: string]: any } = { name: 'john' }
+      rumPublicApi.setUser(user)
+      rumPublicApi.setUserProperty('userDetails', userDetails)
+      userDetails.DOB = '11/11/1999'
+      const userClone = rumPublicApi.getUser()
+
+      expect(userClone.userDetails).toEqual({ name: 'john' })
```
bcaudan, line 506: **💬 suggestion:** ‏should we not check the reference in this test?
```suggestion
    it('should not contain original reference to object', () => {
      const userDetails: { [key: string]: any } = { name: 'john' }
      rumPublicApi.setUser(user)
      rumPublicApi.setUserProperty('userDetails', userDetails)
      userDetails.DOB = '11/11/1999'
      const userClone = rumPublicApi.getUser()

      expect(userClone.userDetails).not.toBe(userDetails)
```

packages/rum-core/src/boot/rumPublicApi.ts
liywjl: good point, I have removed it as ti doesn't make sense

bcaudan: Sanitizing user properties seemed to be a good idea to me, what would prevent customers to do:

```ts
setUserProperty('id', 2)
```

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -455,6 +455,94 @@ describe('rum public api', () => {
...
+
+    it('should sanitize properties', () => {
+      rumPublicApi.setUserProperty('id', 123)
+      rumPublicApi.setUserProperty('name', ['Adam', 'Smith'])
+      rumPublicApi.setUserProperty('email', { foo: 'bar' })
+      const userClone = rumPublicApi.getUser()
+
+      expect(userClone.id).toEqual('123')
+      expect(userClone.name).toEqual('Adam,Smith')
+      expect(userClone.email).toEqual('[object Object]')
```
liywjl, line 525: This matches tests defined in `setUser`

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -129,8 +134,8 @@ export function makeRumPublicApi(
     const startRumResults = startRumImpl(
       configuration,
       () => ({
-        user,
-        context: globalContextManager.get(),
+        user: userContextManager.getContext(),
+        context: globalContextManager.getContext(),
```
bcaudan, line 138: **💬 suggestion:** ‏on second look, if a customer relied on the fact that we did not clone the global context to pass some modifications by reference, it may be a breaking change, no? 
If so, it could be safer to keep using `globalContextManager.get()` until the next major.
wdyt?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,40 +217,45 @@ export function makeRumPublicApi(
...
 
-  function sanitizeUser(newUser: unknown) {
-    if (typeof newUser !== 'object' || !newUser) {
-      return
-    }
-    const result = deepClone(newUser as Context)
-    if ('id' in result) {
-      result.id = String(result.id)
+  function sanitizeUser(newUser: Context) {
+    const shallowClondeUser = assign(newUser, {})
```
bcaudan, line 248: **💬 suggestion:** ‏typo, if you don't already use one, having some spell checker in your IDE could be handy 😉
```suggestion
    const shallowClonedUser = assign(newUser, {})
```

packages/rum-core/src/boot/rumPublicApi.ts
liywjl: I believe in this case it is fine; this 2nd param in `startRumImpl` refers to `getCommonContext` in [`startRum`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/boot/startRum.ts#L28). 

That function is then passed to [`startRumAssembly`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/assembly.ts#L63) which uses it to define the variable [`commonContext`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/assembly.ts#L102). 

Finally that variable is used to set [`serverRumEvent.context`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/assembly.ts#L138), which is in turn dispatched by [`lifeCycle.notify(LifeCycleEventType.RUM_EVENT_COLLECTED, serverRumEvent)`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/assembly.ts#L152)

To my (hesitant) understanding I believe that is variable is not mutated or exposed publicly therefore making such a change ok. Thoughts?

bcaudan: It could have been exposed through `beforeSend` in `event.context` but since there is a deep clone performed by `limitModification` it is not, so all good :swea

bcaudan (approved):

packages/rum-core/src/boot/rumPublicApi.ts
BenoitZugmeyer: **💭 thought:** ‏Since it is internal usage only, and we know that it's not mutateted, maybe we can still use `globalContextManager.get()`? Using `getContext` doesn't seem useful here?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -201,40 +217,45 @@ export function makeRumPublicApi(
...
       }
     }),
 
-    removeUser: monitor(() => {
-      user = {}
+    getUser: monitor(userContextManager.getContext),
+
+    setUserProperty: monitor((key, property) => {
+      const sanitizedProperty = sanitizeUser({ [key]: property })[key]
+      userContextManager.setContextProperty(key, sanitizedProperty)
```
BenoitZugmeyer, line 231: **🥜 nitpick:** ‏ ‏This is a bit weird. We could have a `sanitizeUserProperty(key, value)` function that cast the value if key is `id`, `name` or `email` instead, and use it in `sanitizeUser`. But it's not dramatic and we can live with it :)

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.ts
liywjl: @BenoitZugmeyer knowing we are looking  at deprecating `get` in contextManager, I think it makes sense to change this, even though this is only an internal API

BenoitZugmeyer: Nothing forces us to deprecate it (we might want to rename it to make it clearer though)

liywjl: I am not sure I follow? is this regarding the name `getContext`?

To recap:
- `contextManager.get()` is being phased out because it exposes the context. 
- `contextManager.getContext()` is now the preferred way of getting context, and returns a clone of the context (thus resolving the issue of the `get` method)

packages/rum-core/src/boot/rumPublicApi.ts
liywjl: I agree this isn't the nicest, but I really wanted to avoid having 2x the same sanitize function for different input types. Having once source of truth is also nicer if we decide to expand `sanitizeUser`

Issue #1592, opened 2022-06-15
tracekit computeStackTrace misses some stack traces and fails to upload stack with error


hipitihop:
We have found that the [tracekit.computeStackTrace](
https://github.com/DataDog/browser-sdk/blob/1ea6cc8867925b5984cf5755b7a26d8f6a27bbb5/packages/core/src/domain/tracekit/computeStackTrace.ts#L34) regular expression misses some stack traces. In some cases, it matches no stack lines, so only uploads the exception message, instead of the trace. In other cases, it misses some stack lines.

1. Example from a local development environment where the entire stack fails to be matched:

  ```
  Error: RTE Simulation
      at day8$curves$events$handler_COLON_simulate_exception (events.cljs:1060:12)
      at eval (std_interceptors.cljc:132:29)
      at re_frame$std_interceptors$fx_handler__GT_interceptor_$_fx_handler_before (std_interceptors.cljc:128:19)
      at Object.re_frame$interceptor$invoke_interceptor_fn [as invoke_interceptor_fn] (interceptor.cljc:70:6)
      at Object.re_frame$interceptor$invoke_interceptors [as invoke_interceptors] (interceptor.cljc:108:24)
      at Object.re_frame$interceptor$execute [as execute] (interceptor.cljc:199:8)
      at Object.re_frame$events$handle [as handle] (events.cljc:65:14)
      at Object.eval [as re_frame$router$IEventQueue$_process_1st_event_in_queue$arity$1] (router.cljc:179:10)
      at Object.eval [as re_frame$router$IEventQueue$_run_queue$arity$1] (router.cljc:198:44)
      at eval (router.cljc:146:64)
  ```
  > NOTE: in this example, this is not advanced compiled JS from a development environment, so we are not expecting actual source mapping to work, nonetheless, the stack trace should still be logged along with this exception.

2. Example from release environment, where the stack is uploaded, however, the first two lines of the stack don't match the regex and are excluded.
  ```
  Error: RTE Simulation
      at https://test.curve8.day8.com.au/client/compiled_prod/curves.js:8489:191
      at https://test.curve8.day8.com.au/client/compiled_prod/curves.js:5329:312
      at ajb (https://test.curve8.day8.com.au/client/compiled_prod/curves.js:5302:157)
      at cjb (https://test.curve8.day8.com.au/client/compiled_prod/curves.js:5302:487)
      at kjb (https://test.curve8.day8.com.au/client/compiled_prod/curves.js:5315:435)
[...]

Issue #957, opened 2021-07-23
Stack traces are not formated correctly for chrome extensions


dickfickling:
In computeStackTrace.ts, the regex for matching a Chrome stack line marks the open parenthesis as required, when it shouldn't be. This causes my stack traces to be mangled on upload to RUM. This is making RUM unusable for me as the uploaded stack traces don't correspond to the actual code.

I created a PR (#954) which updates a test that was "working", but actually silently parsing certain lines as winjs stack lines, to add a line for a chrome extension stack trace example. By making the open parenthesis in the regex optional, that specific test starts to pass again, but it causes other tests to fail.

See #954 for more details.

liywjl:
Hello @dickfickling, 

Thank you for reporting this issue. We have resolved it as of v4.16.0.

It appears that our regex struggled to handle cases where anonymous functions are swallowed by the stack trace. I was not able to reproduce the stack trace myself but using the examples provided in your [PR](https://github.com/DataDog/browser-sdk/pull/954) I added some new [unit test](https://github.com/DataDog/browser-sdk/pull/1647/files#diff-ff9c61c635cc727f4558a106898fe5f24808fca1b1a5f748305d36929a7e0f45R516-R522) to cover this new case.

I hope this solves your issue. Don't hesitate to open another ticket if you have a specific issue related to the SDK or contact [datadog support](https://www.datadoghq.com/support/) directly.

PR #1647, opened 2022-07-19
[RUMF-1310] handle extra stacktrace parsing cases

liywjl:
## Motivation

Some users have reported cases in which we fail to parse the error stack in Chrome:

https://github.com/DataDog/browser-sdk/issues/957
https://github.com/DataDog/browser-sdk/issues/1592

## Changes

After some digging, it appears there are 2 Chrome cases we do not handle:
- `closureScripe` logs file URL differently (only displaying the filename and not the path)
- `anonymous function` in some cases Chrome seems to ignore the text that indicates the stack was called by an anonymous function

## Solution
- `closureScripe`:  match URLs that match a filename without prefix
Add `\w+\.` to URL regex

- `anonymous function`: ignore regex trying to identify function
Remove ` (.*?)` from regex, and create new parse function

## Testing

Add unit tests to handle cases

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -61,6 +63,26 @@ function parseChromeLine(line: string): StackFrame | undefined {
   }
 }
 
+const CHROME_ANONYMOUS_FUNCTION_RE =
+  // eslint-disable-next-line max-len
+  /^\s*at ?((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\w+\.[cljs|cljc]|\/).*?)(?::(\d+))?(?::(\d+))??\s*$/i
```
amortemousque, line 68: **❓ question:** ‏Couldn't we make a single regex for both cases with the function name and the url parentheses optional?

liywjl: I found making the `(` optional broke a bunch of the other edge cases that we tested for. I played around with some other ideas to no avail. 

Although not ideal, this was at least a working solution that didn't break any of the other cases. Happy to explore other ideas you might have?

packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -31,7 +32,8 @@ export function computeStackTrace(ex: unknown): StackTrace {
 
 const CHROME_LINE_RE =
   // eslint-disable-next-line max-len
-  /^\s*at (.*?) ?\(((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\/).*?)(?::(\d+))?(?::(\d+))?\)?\s*$/i
+  /^\s*at (.*?) ?\(((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\w+\.[cljs|cljc]|\/).*?)(?::(\d+))?(?::(\d+))?\)?\s*$/i
```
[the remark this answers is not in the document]
liywjl: Great idea! I changed that section to `\w+\.` which seems to have worked and not broken previous cases

packages/core/src/domain/tracekit/computeStackTrace.ts
bcaudan: **💬 suggestion:** ‏what about extracting what we can have between paranthesis and generate a new regexp containing both options?

```js
const filePart =  "?:file|https?|blob|chrome-extension|native|eval|web..."
const regex = new Regex(`...(\(${filePart}\)|${filePart})...`)
```

BenoitZugmeyer: What @bcaudan suggests sounds great. I would do something like this:

```js
const url = '((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\\w+\\.|\\/).*?)'
const position = '((?::(\\d+))?(?::(\\d+))?'
const chromeLine = new Regex(`^\s*at(?: (.*?))? (?:\(${url}\)|${url})${position}\s*$`, 'i')
```

packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -61,6 +65,26 @@ function parseChromeLine(line: string): StackFrame | undefined {
...
+  if (!parts) {
+    return
+  }
+
+  return {
+    args: [],
+    column: parts[3] ? +parts[3] : undefined,
+    func: UNKNOWN_FUNCTION,
+    line: parts[2] ? +parts[2] : undefined,
+    url: parts[1] ?? undefined,
```
BenoitZugmeyer, line 84: **💬 suggestion:** ‏the URL part isn't optional
```suggestion
    url: parts[1],
```

packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -28,11 +29,14 @@ export function computeStackTrace(ex: unknown): StackTrace {
...
   }
 }
-
+const url = '((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\\w+\\.|\\/).*?)'
+const position = '(?::(\\d+))'
 const CHROME_LINE_RE =
-  // eslint-disable-next-line max-len
-  /^\s*at (.*?) ?\(((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\/).*?)(?::(\d+))?(?::(\d+))?\)?\s*$/i
-const CHROME_EVAL_RE = /\((\S*)(?::(\d+))(?::(\d+))\)/
+  // eslint-disable-next-line local-rules/disallow-side-effects
```
BenoitZugmeyer, line 35: **🔨 warning:** ‏You should not disable this rule, as it guarantees that our code will stay mostly tree-shakable. In this case, we have two possibilities:
* allow `new RegExp` by changing the rule [here](https://github.com/DataDog/browser-sdk/blob/dfcd3dc6a861be9a25d3b1cc533262096a687742/eslint-local-rules/disallowSideEffects.js#L197)
* build the RE in a function, like:
```ts
let chromeLineRe: RegExp | undefined
function getChromeLineRe() {
  if (!chromeLineRe) chromeLineRe = new RegExp(...)
  return chromeLineRe
}
```

We know that `new RegExp` is side-effect free, so the first solution could be good. One small drawback is that Webpack+Terser fails to treeshake regex constructed with variables like you do. This can be observed like that:
```
$ cat src/index.js
const a = new RegExp("foo");

const bar = "bar";
const b = new RegExp(`foo${bar}`);

$ npx --package webpack-cli webpack --mode production
assets by status 21 bytes [cached] 1 asset
./src/index.js 84 bytes [built] [code generated]
webpack 5.72.0 compiled successfully in 99 ms

$ cat dist/main.js
new RegExp("foobar");
```
`a` has been treeshaken in the resulting file, but `b` not entirely.

As this is a minor issue, and Terser might be improved someday, I suggest changing the rule nonetheless. (Note that [rollup does not have this issue](https://rollupjs.org/repl/?version=2.77.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmNvbnN0JTIwYSUyMCUzRCUyMG5ldyUyMFJlZ0V4cCglNUMlMjJmb28lNUMlMjIpJTNCJTVDbiU1Q25jb25zdCUyMGJhciUyMCUzRCUyMCU1QyUyMmJhciU1QyUyMiUzQiU1Q25jb25zdCUyMGIlMjAlM0QlMjBuZXclMjBSZWdFeHAoJTYwZm9vJTI0JTdCYmFyJTdEJTYwKSUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=) 🎉  )

packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -28,11 +29,11 @@ export function computeStackTrace(ex: unknown): StackTrace {
     stack,
   }
 }
+const url = '((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\\w+\\.|\\/).*?)'
+const position = '(?::(\\d+))'
```
bcaudan, line 33: **💬 suggestion:**

- make clearer url and position of what
- avoid duplicating position structure‏

```suggestion
const fileUrl = '((?:file|https?|blob|chrome-extension|native|eval|webpack|<anonymous>|\\w+\\.|\\/).*?)'
const filePosition = '(?::(\\d+))?(?::(\\d+))?'
```

bcaudan (approved):

packages/core/src/domain/tracekit/computeStackTrace.ts
liywjl: `filePosition` contains `?` in-between each position. In the use case of `CHROME_EVAL_RE` this char doesn't exist, therefore not making it composable. Happy to however use the prefix `file` to the variable names

BenoitZugmeyer (approved):

packages/core/src/domain/tracekit/computeStackTrace.ts
bcaudan: oh ok, missed this one, nice then

PR #1650, opened 2022-07-20
👷 [RUMF-1336] guard against prototype pollution

bcaudan:
## Motivation

To please the code scanner

## Changes

In `limitModification`, ensure that modified field is owned by the object and not only exist on the object.
No risk on the current state since we are controlling which fields can be modified and that those fields are owned by the objects.
This change should not change current behavior.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/limitModification.ts
+++ b/packages/core/src/tools/limitModification.ts
@@ -54,5 +54,5 @@ function set(object: unknown, path: string, value: unknown) {
 }
 
 function isValidObjectContaining(object: unknown, field: string): object is { [key: string]: unknown } {
-  return typeof object === 'object' && object !== null && field in object
+  return typeof object === 'object' && object !== null && Object.prototype.hasOwnProperty.call(object, field)
 }


packages/core/src/tools/limitModification.ts
```diff
@@ -54,5 +54,5 @@ function set(object: unknown, path: string, value: unknown) {
 }
 
 function isValidObjectContaining(object: unknown, field: string): object is { [key: string]: unknown } {
-  return typeof object === 'object' && object !== null && field in object
+  return typeof object === 'object' && object !== null && Object.prototype.hasOwnProperty.call(object, field)
```
bcaudan, line 57: This change actually don't fix the issue due to https://github.com/github/codeql/issues/9869.
I'd be in favor of going with this fix and dismiss the alert as false positive.

liywjl (approved):

PR #1652, opened 2022-07-20
👷 [REPLAY-922] Generate Session Replay types from JSON-schema

acorretti:
## Motivation
Browser Session Replay types are manually-defined. Now that [rum-events-format](https://github.com/DataDog/rum-events-format) got a JSON-schema definition for Session Replay data, let's generate the Typescript definitions the same way we do for RUM events.

Additionally, the `source` property is added to the segment metadata. On Browser SDK, this is a hardcoded `'browser'` value.

## Changes
* Adds segment metadata property `source` with constant value `'browser'`.
* Removes manually-defined Session Replay definitions.
* Generates `packages/rum/src/types/session-replay.ts` out of the `schemas/session-replay-browser-schema.json` rum-events-format schema.
* Adds some manually-defined utility types & const at `packages/rum/src/types/session-replay-utils.ts`.
* Adjusts all import references to use _Browser*_ generated types, aliasing them to their un-prefixed name.

## Testing

This PR aims to be a mere drop-in replacement of the manual types by the automatically-generated ones. As such, building & successfully running the tests is expected.

It's been deployed and tested in staging (`sdk_version:4.15.0_6c3be9dfac758d714432dae86fb7b2d773e4ccad`).

- [X] Local
- [X] Staging
- [X] Unit
- [X] End to end

---


packages/rum/src/domain/segmentCollection/segment.spec.ts
```diff
@@ -14,7 +14,7 @@ const ENCODED_SEGMENT_HEADER_BYTES_COUNT = 12 // {"records":[
 const ENCODED_RECORD_BYTES_COUNT = 25
 const ENCODED_FULL_SNAPSHOT_RECORD_BYTES_COUNT = 35
 const ENCODED_SEPARATOR_BYTES_COUNT = 1 // ,
-const ENCODED_META_BYTES_COUNT = 173 // this should stay accurate as long as less than 10 records are added
+const ENCODED_META_BYTES_COUNT = 192 // this should stay accurate as long as less than 10 records are added
```
liywjl, line 17: **❓ question:** I am curious to know why this changed?

liywjl (approved):
LGTM

packages/rum/src/domain/segmentCollection/segment.spec.ts
acorretti: Because we've added the `source: 'browser'` property to the segment metadata

bcaudan:
@acorretti could you update `rum-events-format` reference and regenerate the types? it seems that we are missing the latest commits

packages/rum/src/types/session-replay-utils.ts
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable unicorn/filename-case */
```
bcaudan, line 1: **❓ question:** ‏why not following the file name convention?

scripts/generate-schema-types.js
```diff
@@ -16,9 +16,14 @@ async function main() {
     path.join(__dirname, '../packages/core/src/domain/telemetry/telemetryEvent.types.ts'),
     'telemetry-events-schema.json'
   )
+  await generateTypesFromSchema(
+    path.join(__dirname, '../packages/rum/src/types/session-replay.ts'),
```
bcaudan, line 20: **💬 suggestion:** ‏what about following file name pattern?
```suggestion
    path.join(__dirname, '../packages/rum/src/types/sessionReplay.types.ts'),
```

packages/rum/test/htmlAst.ts
```diff
@@ -32,8 +32,8 @@ export const generateLeanSerializedDoc = (htmlContent: string, privacyTag: strin
     serializeNodeWithId(newDoc, {
       document: newDoc,
       parentNodePrivacyLevel: NodePrivacyLevel.ALLOW,
-    })!
-  ) as SerializedNodeWithId
+    })! as unknown as Record<string, unknown>
+  ) as unknown as SerializedNodeWithId
```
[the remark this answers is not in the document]
BenoitZugmeyer: I agree, those types are far from perfect, but I don't think it should be changed in this PR since they were already there. Let's change that in a follow-up task

scripts/generate-schema-types.js
acorretti: I agree. I think I didn't give it too much thought, can't recall where I got the need for kebab case from.

packages/rum/src/types/session-replay-utils.ts
acorretti: Done!

packages/rum/src/types/session-replay.ts
```diff
@@ -0,0 +1,640 @@
+/* eslint-disable */
+/**
+ * DO NOT MODIFY IT BY HAND. Run `yarn rum-events-format:sync` instead.
+ */
+
+/**
+ * Schema of all properties of browser Session Replay
+ */
+export type BrowserSessionReplay = BrowserSegment
```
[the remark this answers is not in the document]
acorretti: Sadly and for the time being, yes.
The schema entrypoint defines this type: https://github.com/DataDog/rum-events-format/blob/master/schemas/session-replay-browser-schema.json

I've tried and we can get rid of the type, just point at the `$ref` to the `segment-schema.json`, however it requires a change to _rum-event-format_ schemas and we'll want this change on the mobile side as well for the sake of consistency. I need to check with mobile recorders to see they don't have a problem doing that.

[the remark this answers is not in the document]
bcaudan: oh ok.
Not related to this PR then but it could worth to check if there is a need for this entrypoint and remove it if it is useless.

packages/rum/test/htmlAst.ts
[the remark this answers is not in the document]
bcaudan: Sounds good to me

packages/rum/src/domain/record/utils.ts
```diff
@@ -1,5 +1,8 @@
 import { assign, timeStampNow } from '@datadog/browser-core'
-import type { IncrementalData, IncrementalSnapshotRecord } from '../../types'
+import type {
+  BrowserIncrementalData as IncrementalData,
+  BrowserIncrementalSnapshotRecord as IncrementalSnapshotRecord,
```
BenoitZugmeyer, line 4: **💬 suggestion:** ‏Rename those types in `src/types/index.ts`, before exporting them, so we have a consistent names accross the codebase. The drawback is that you couldn't use `export * from`. WDYT?

packages/rum/src/types/sessionReplayUtils.ts
```diff
@@ -0,0 +1,84 @@
+import type * as SessionReplay from './sessionReplay'
```
BenoitZugmeyer, line 1: **🥜 nitpick:** ‏this file could be renamed `sessionReplayConstants` instead of `Utils`

packages/rum/src/domain/record/utils.ts
acorretti: Well, it adds another process on top of re-generating the types.
From the player (and the product) perspective, there's nothing wrong in calling those types `Browser*`.

BenoitZugmeyer: > Well, it adds another process on top of re-generating the types.

I don't think so? We would change `src/types/index.ts` only when we need to, no need to be exhaustive here.

>  From the player (and the product) perspective, there's nothing wrong in calling those types Browser*.

I understand, no issue with that.

I would be OK with using `Browser*` types in the Browser SDK instead of using `import as` everytime we want to use them. It is not enforced so we need to think about it all the time, and it breaks auto-import.

acorretti: I was talking about the `index.ts` sure.

It's fine to get rid of the aliases, just that I didn't want to make all those changes in the scope of this PR.

acorretti: As agreed with @BenoitZugmeyer, I'm going to get rid of the aliases right in this PR.
Will send a follow-up commit.

BenoitZugmeyer (approved):

PR #1653, opened 2022-07-21
🐛 improve `jsonStringify` implementation

BenoitZugmeyer:
## Motivation

The previous implementation only removed toJSON on the input value and its prototype. We usually pass an Object as input, so it was focused on Object.

But websites using outdated libraries that define toJSON to Array objects (ex: prototype.js before 1.7) are still an issue with this strategy.

## Changes

This new implementation assumes that internally, the SDK won't use other objects than Array and Object (primitive types like string and number are unaffected), and those objects won't be defined with an own toJSON method. So, it focuses on removing toJSON to Array and Object prototypes, which completely solves the issue.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/utils.ts
+++ b/packages/core/src/tools/utils.ts
@@ -146,61 +146,51 @@ export function round(num: number, decimals: 0 | 1 | 2 | 3 | 4) {
 // eslint-disable-next-line @typescript-eslint/no-empty-function
 export function noop() {}
 
-interface ObjectWithToJSON {
+interface ObjectWithToJsonMethod {
   toJSON: (() => object) | undefined
 }
+function removeToJsonMethod(object: ObjectWithToJsonMethod) {
+  const objectToJson = object.toJSON
+  if (objectToJson) {
+    delete object.toJSON
+    return () => {
+      object.toJSON = objectToJson
+    }
+  }
 
-type OriginalToJSON = [boolean, undefined | (() => object)]
+  return noop
+}
 
 /**
  * Custom implementation of JSON.stringify that ignores value.toJSON.
  * We need to do that because some sites badly override toJSON on certain objects.
- * Note this still supposes that JSON.stringify is correct...
+ * Note this still assumes that JSON.stringify is correct...
  */
 export function jsonStringify(
   value: unknown,
   replacer?: Array<string | number>,
   space?: string | number
 ): string | undefined {
-  if (value === null || value === undefined) {
-    return JSON.stringify(value)
-  }
-  let originalToJSON: OriginalToJSON = [false, undefined]
-  if (hasToJSON(value)) {
-    // We need to add a flag and not rely on the truthiness of value.toJSON
-    // because it can be set but undefined and that's actually significant.
-    originalToJSON = [true, value.toJSON]
-    delete value.toJSON
-  }
+  // Some sites are badly overriding toJSON on certain objects. We expect that the Browser SDK won't
+  // use other objects than Array and Object to represent a value to stringify, and those objects
+  // won't have an own "toJSON" method defined.
+  //
+  // Thus, we only need to make sure that Array and Object prototypes don't define their own toJSON
+  // methods. User-provided values might have `toJSON` methods, but it is safe to keep them, since
+  // it won't impact the SDK operations or data model representation.
+  //
+  // Note: Object should be handled before Array because Array is a subclass of Object.
+  const restoreObjectPrototypeToJson = removeToJsonMethod(Object.prototype as ObjectWithToJsonMethod)
+  const restoreArrayPrototypeToJson = removeToJsonMethod(Array.prototype as unknown as ObjectWithToJsonMethod)
 
-  let originalProtoToJSON: OriginalToJSON = [false, undefined]
-  let prototype
-  if (typeof value === 'object') {
-    prototype = Object.getPrototypeOf(value) as object
-    if (hasToJSON(prototype)) {
-      originalProtoToJSON = [true, prototype.toJSON]
-      delete prototype.toJSON
-    }
-  }
-
-  let result: string
   try {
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/utils.ts
```diff
@@ -146,61 +146,51 @@ export function round(num: number, decimals: 0 | 1 | 2 | 3 | 4) {
...
-      delete prototype.toJSON
-    }
-  }
+  // Some sites are badly overriding toJSON on certain objects. We expect that the Browser SDK won't
+  // use other objects than Array and Object to represent a value to stringify, and those objects
+  // won't have an own "toJSON" method defined.
+  //
+  // Thus, we only need to make sure that Array and Object prototypes don't define their own toJSON
+  // methods. User-provided values might have `toJSON` methods, but it is safe to keep them, since
+  // it won't impact the SDK operations or data model representation.
```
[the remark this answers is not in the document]
BenoitZugmeyer: > If a customer is giving us an object owning a toJSON method throwing an error, the current implementation works fine but with this changes it will throw

The current behavior will not change in this case. It didn't work with the current implementation because it is removing `value.toJSON` only on the *root* value, which is never a user-provided value. For example, if a user does something like this with the current version, it won't work with both current and new implementations:

```ts
DD_RUM.addRumGlobalContext('toto', { toJSON() { throw new Error('toto') } } )
```

> IIRC, the current jsonStringify implementation is inherited from the synthetics team, it could worth checking with them if they recall why this case was introduced

Ok, I'll check that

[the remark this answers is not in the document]
BenoitZugmeyer: Ah my mistake, `jsonStringify` is also used to serialize errors and console arguments, which are user-provided. I'll remove the `toJSON` method on the value then.

liywjl (approved):
LGTM 👍

packages/core/src/tools/utils.ts
```diff
@@ -146,61 +146,55 @@ export function round(num: number, decimals: 0 | 1 | 2 | 3 | 4) {
...
-    if (originalToJSON[0]) {
-      ;(value as ObjectWithToJSON).toJSON = originalToJSON[1]
-    }
-    if (originalProtoToJSON[0]) {
-      ;(prototype as ObjectWithToJSON).toJSON = originalProtoToJSON[1]
-    }
+    restoreObjectPrototypeToJson()
+    restoreArrayPrototypeToJson()
+    restoreValueToJson()
+    restoreValuePrototypeToJson()
```
bcaudan, line 181: **💬 suggestion:** a bit surprising to not restore in the same order‏
```suggestion
    restoreValuePrototypeToJson()
    restoreValueToJson()
```

packages/core/src/tools/utils.ts
```diff
@@ -146,61 +146,55 @@ export function round(num: number, decimals: 0 | 1 | 2 | 3 | 4) {
...
+ *
+ * Note: this still assumes that JSON.stringify is correct.
  */
 export function jsonStringify(
   value: unknown,
   replacer?: Array<string | number>,
   space?: string | number
 ): string | undefined {
-  if (value === null || value === undefined) {
+  if (!value || typeof value !== 'object') {
```
bcaudan, line 162: **💬 suggestion:** ‏`!value` seems a bit too broad to have a clear intent, should it be:
```suggestion
  if (typeof value !== 'object' || value === null) {
```

packages/core/src/tools/utils.ts
```diff
@@ -146,61 +146,55 @@ export function round(num: number, decimals: 0 | 1 | 2 | 3 | 4) {
...
+    restoreValueToJson()
+    restoreValuePrototypeToJson()
   }
-  return result
 }
 
-function hasToJSON(value: unknown): value is ObjectWithToJSON {
-  return typeof value === 'object' && value !== null && Object.prototype.hasOwnProperty.call(value, 'toJSON')
+interface ObjectWithToJsonMethod {
+  toJSON: (() => object) | undefined
```
bcaudan, line 186: **🥜 nitpick:** while we are here, no need to know its structure‏
```suggestion
  toJSON: unknown
```

bcaudan (approved):

packages/core/src/tools/utils.ts
```diff
@@ -146,61 +146,55 @@ export function round(num: number, decimals: 0 | 1 | 2 | 3 | 4) {
...
-      originalProtoToJSON = [true, prototype.toJSON]
-      delete prototype.toJSON
-    }
-  }
+  // Note: The order matter here. We need to detach toJSON methods on parent classes before their
+  // subclasses.
+  const restoreObjectPrototypeToJson = detachToJsonMethod(Object.prototype)
+  const restoreArrayPrototypeToJson = detachToJsonMethod(Array.prototype)
+  const restoreValuePrototypeToJson = detachToJsonMethod(Object.getPrototypeOf(value))
+  const restoreValueToJson = detachToJsonMethod(value)
```
ArmorDarks, line 171: @BenoitZugmeyer could you please clarify while all objects are stripped of toJSON method before adding it to logs? Many classes because of it are rendered simply as `{}` which isn't very convenient and now needs extra logic to invoke `toJSON` on those objects _before_ passing it to the DataDog library.

Any chance it will be made configurable?

Thank you

bcaudan: Hi @ArmorDarks, can you please open an issue to start a new discussion?

BenoitZugmeyer: We found out that some custom `toJSON` implementations were throwing exceptions and broke everything. That's why we are removing those. But we are in the process of re-designing how we sanitize/serialize user data, and we are talking of respecting the `toJSON` methods (behind a try/catch)

PR #1667, opened 2022-08-02
[RUMF-1346] heatmaps: move action event attributes

BenoitZugmeyer:
## Motivation

For now, we move some fields introduced for heatmaps in `_dd` so we can adjust them at will. We'll make them public once heatmaps are released and there is a need for it.

Related: https://github.com/DataDog/rum-events-format/pull/80

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/action/actionCollection.ts
```diff
@@ -76,6 +74,10 @@ function processAction(
             count: action.counts.resourceCount,
           },
         },
+        _dd: {
+          target: action.target,
+          position: action.position,
+        },
```
bcaudan, line 80: **❓ question:** ‏From schema, it seems to be `_dd.action.target` and `_dd.action.position` but all your updates are on `_dd.target` and `_dd.position`. Is it a mistake or am I missing something?

BenoitZugmeyer: It's a mistake, thanks :)

bcaudan (approved):

PR #1668, opened 2022-08-02
[RUMF-1357] Add a peer dependency between rum and logs packages

### LINKED ISSUE #1666: Establish peer dependency between rum and logs packages?
I'm curious if there is appetite for (or a specific reason for not) establishing a peer dependency between the rum and logs packages. It'd be nice to tell NPM/Yarn that if `@datadog/rum` is at 4.16.0, `@datadog/logs` also must be at 4.16.0 (if present), and vice versa. Luckily, [`peerDependencies`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependencies) does the trick for us.

Considering the RUM docs ([here](https://docs.datadoghq.com/real_user_monitoring/browser/#choose-the-right-installation-method) and [here](https://docs.datadoghq.com/logs/log_collection/javascript/#choose-the-right-installation-method)) state that it is recommended to match versions, this seems like a beneficial addition! I also believe (though I'm having a hard time fully parsing the [Dependabot](https://github.com/dependabot/dependabot-core) source to be sure) it will force Dependabot to update both `@datadog/rum` and `@datadog/logs` in tandem if both are included dependencies of a project.

I would love to hear thoughts on this option!

agrobbin:
## Motivation

I suggested this, so made the PR too! This closes #1666.

## Changes

The RUM docs ([here](https://docs.datadoghq.com/real_user_monitoring/browser/#choose-the-right-installation-method) and [here](https://docs.datadoghq.com/logs/log_collection/javascript/#choose-the-right-installation-method)) state that it is recommended to match versions, so this seems like a beneficial addition!

I also believe (though I'm having a hard time fully parsing the [Dependabot](https://github.com/dependabot/dependabot-core) source to be sure) it will force Dependabot to update both @datadog/rum and @datadog/logs in tandem if both are included dependencies of a project.

## Testing

My primary concern from a "testing" perspective is whether Lerna will know to update these when a new version is released. I'm not familiar w/ Lerna, or the configuration for it, but hopefully this is a starting point!


BenoitZugmeyer:
Thank you for your PR! Could you make those peer dependencies "optional" with [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta)?

As you guessed, lerna will be an issue for that. They [intentionally removed support for it](https://github.com/lerna/lerna/commit/bdbfc62966e5351abfeac77830f9d47b6d69f1b1), and [closed a feature request asking for an option to bump them on publish](https://github.com/lerna/lerna/issues/1575). I think we'll need to tweak [our `version` script](https://github.com/DataDog/browser-sdk/blob/d4c60b83bf09e890966383cca1b6c01336d9bb56/scripts/cli#L59-L63) to bump it manually.

You can try it yourself with something like `npx lerna version --no-push --no-git-tag-version -y major`

agrobbin:
@BenoitZugmeyer good call, I just added `peerDependenciesMeta`. I'm not sure what the idea might be re: the `cli` script. It'd be ideal to programmatically manipulate the contents of `package.json`, but I'm not terribly familiar with that landscape. A quick Google suggests `npe`, `json`, and a few others, but I'm definitely open to some suggestions here!

Additionally, after reading some of the comments in https://github.com/lerna/lerna/issues/1575, I do want to confirm that this is the correct usage of `peerDependencies`. I still think it makes sense, but this isn't my project. 😆

BenoitZugmeyer:
> I just added `peerDependenciesMeta`. 

Thank you!

> I'm not sure what the idea might be re: the `cli` script. It'd be ideal to programmatically manipulate the contents of `package.json`, but I'm not terribly familiar with that landscape. A quick Google suggests `npe`, `json`, and a few others, but I'm definitely open to some suggestions here!

Mmh I was thinking about a simple "find and replace" script (ex: with `sed`), or if we want to do something fancier, a 3 lines nodejs script like:
```js
const pkg = JSON.parse(fs.readFileSync('package.json'))
pkg.peerDependencies["@browser-logs"] = pkg.version
fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2))
```

But don't worry about that, we can take over your PR to make the final adjustments 🙂 

> Additionally, after reading some of the comments in [lerna/lerna#1575](https://github.com/lerna/lerna/issues/1575), I do want to confirm that this is the correct usage of `peerDependencies`. I still think it makes sense, but this isn't my project. 😆

I agree that it still makes sense for us. I think arguments are a bit outdated, for example “`peerDependencies` themselves only warn, not auto-install” is not true anymore: it either don't warn (optional) or auto-install (required). It made sense when peer dependencies were kind of required, but we want our packages to be optional.

agrobbin:
@BenoitZugmeyer ah got it, that makes sense! I've enabled maintainer editing access on my branch, which hopefully is what you need to finalize the `bin/cli` script changes. Let me know if there is anything else I need to do.

agrobbin:
@BenoitZugmeyer love it!

scripts/update-peer-dependency-versions.js
```diff
@@ -0,0 +1,25 @@
...
+
+// This script updates the peer dependency versions between rum and logs packages to match the new
+// version during a release.
+//
+// Lerna [intentionally removed support for it][1], and [closed a feature request asking for an
+// option to bump them on publish][2], but we still want to make sure this version is updated as
+// well to warn users who use different versions of RUM and Logs packages.
+//
+// [1]: https://github.com/lerna/lerna/commit/bdbfc62966e5351abfeac77830f9d47b6d69f1b1
+// [2]: https://github.com/lerna/lerna/issues/1575
```
liywjl, line 12: should we use JSDoc?

liywjl (approved):
Nice work!

bcaudan (approved):

BenoitZugmeyer:
Thank you for your contribution @agrobbin , this will be released shortly.

agrobbin:
Awesome, thanks @BenoitZugmeyer!

agrobbin:
Well, looks like I was wrong about Dependabot bumping these together. 😢

PR #1669, opened 2022-08-02
✨ [RUMF-1306] Send the tracing sample rate in _dd.rule_psr for resources

liywjl:
## Motivation

We want to be able to display the sample rate on RUM BE

## Changes

Pull `sampleRate` out of `configuration` and set it to the `rumContext` under `_dd.rule_psr` if present

## Updated Schema

https://github.com/DataDog/rum-events-format/pull/83

## Testing

- [x] Local
- [x] Staging 
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/assembly.ts
```diff
@@ -109,6 +109,7 @@ export function startRumAssembly(
             session: {
               plan: session.hasPremiumPlan ? RumSessionPlan.PREMIUM : RumSessionPlan.LITE,
             },
+            rule_psr: 'sampleRate' in configuration ? configuration.sampleRate : undefined,
```
BenoitZugmeyer, line 112: **🔨 warning:**  it should be `configuration.tracingSampleRate`, not `configuration.sampleRate`


**💬 suggestion:** ‏what about putting this into [`computeRequestTracingInfo`](https://github.com/DataDog/browser-sdk/blob/d4c60b83bf09e890966383cca1b6c01336d9bb56/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts#L130)?


**💬 suggestion:** ‏`tracingSampleRate` should always be defined on `RumConfiguration` values, so no need to check for its presence

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -130,6 +134,7 @@ function computeRequestTracingInfo(request: RequestCompleteEvent) {
     _dd: {
       span_id: request.spanId!.toDecimalString(),
       trace_id: request.traceId!.toDecimalString(),
+      rule_psr: configuration && 'tracingSampleRate' in configuration ? configuration.tracingSampleRate : undefined,
```
BenoitZugmeyer, line 137: **❓ question:** ‏ ‏`configuration` and `configuration.tracingSampleRate` should always be defined. Did you find some case were it's not, or is it just for extra prevention?


**💬 suggestion:** ‏
```suggestion
      rule_psr: configuration.tracingSampleRate,
```

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.spec.ts
```diff
@@ -8,14 +8,19 @@ import { RumEventType } from '../../../rawRumEvent.types'
...
 
 describe('resourceCollection', () => {
   let setupBuilder: TestSetupBuilder
 
   beforeEach(() => {
     setupBuilder = setup().beforeBuild(({ lifeCycle }) => {
-      startResourceCollection(lifeCycle)
+      startResourceCollection(
+        lifeCycle,
+        validateAndBuildRumConfiguration({ clientToken: 'xxx', applicationId: 'xxx' }) as RumConfiguration
```
BenoitZugmeyer, line 22: **💬 suggestion:** ‏using the `!` operator would be more explicit on what we want to do here
```suggestion
        validateAndBuildRumConfiguration({ clientToken: 'xxx', applicationId: 'xxx' })!
```

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.spec.ts
```diff
@@ -186,6 +191,31 @@ describe('resourceCollection', () => {
...
+
+    it('should pull tracingSampleRate from config if present', () => {
+      setupBuilder = setup().beforeBuild(({ lifeCycle }) => {
+        startResourceCollection(
+          lifeCycle,
+          validateAndBuildRumConfiguration({
+            clientToken: 'xxx',
+            applicationId: 'xxx',
+            tracingSampleRate: 60,
+          }) as RumConfiguration
```
BenoitZugmeyer, line 203: **💬 suggestion:** ‏same here
```suggestion
          })!
```

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -38,6 +38,7 @@ export interface RawRumResourceEvent {
   _dd?: {
     trace_id?: string
     span_id?: string // not available for initial document tracing
+    rule_psr?: number
```
bcaudan, line 41: **❗ warning:** ‏we should also add that to the rum-event-format

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
liywjl: This was a really interesting question. Yes `tracingSampleRate` is always defined in configuration. However `_dd` is not always defined in Resource events and thus added to some confusion on my end.

BenoitZugmeyer: Ah, gotcha. Yes, a lot of attributes are optional in RUM events, because we want to keep strong backward compatibility: some older SDK versions might not send a newly introduced attribute, so the backend/UI should handle both cases.

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -130,12 +137,25 @@ function computeRequestTracingInfo(request: RequestCompleteEvent) {
...
+        trace_id: entry.traceId,
+        rule_psr: configuration.tracingSampleRate / 100, // rule_psr needs to be between 0-1
+      },
+    }
+  }
+  return {
+    _dd: {
+      rule_psr: configuration.tracingSampleRate / 100, // rule_psr needs to be between 0-1
+    },
+  }
```
BenoitZugmeyer, line 158: **💬 suggestion:** ‏I think this would be fine:
```suggestion
  return {
    _dd: {
      trace_id: entry.traceId,
      rule_psr: configuration.tracingSampleRate / 100, // rule_psr needs to be between 0-1
    },
  }
```

If `_dd.trace_id` is undefined, it will be skipped during JSON serialization.

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -130,12 +137,25 @@ function computeRequestTracingInfo(request: RequestCompleteEvent) {
     _dd: {
       span_id: request.spanId!.toDecimalString(),
       trace_id: request.traceId!.toDecimalString(),
+      rule_psr: configuration.tracingSampleRate / 100, // rule_psr needs to be between 0-1
```
BenoitZugmeyer, line 140: **💬 suggestion:** ‏add a function like `function getRulePsr(configuration)` to keep the computation only once

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
BenoitZugmeyer: **❓ question:** ‏in `computeRequestTracingInfo`, we define `rule_psr` only if the request have been traced. Shouldn't we do the same here? Or should it be changed in `computeRequestTracingInfo`?

liywjl: Good point! I think yes, furthermore I think the condition of this check should be lifted to `processRequest` and `processResourceEntry`

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -38,6 +38,7 @@ export interface RawRumResourceEvent {
   _dd?: {
     trace_id?: string
     span_id?: string // not available for initial document tracing
+    rule_psr: number
```
bcaudan, line 41: **💬 suggestion:** ‏to be consistent with schema
```suggestion
    rule_psr?: number
```

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -121,21 +131,23 @@ function computePerformanceEntryMetrics(timing: RumPerformanceResourceTiming) {
   }
 }
 
-function computeRequestTracingInfo(request: RequestCompleteEvent) {
-  const hasBeenTraced = request.traceSampled && request.traceId && request.spanId
-  if (!hasBeenTraced) {
-    return undefined
-  }
+function computeRequestTracingInfo(request: RequestCompleteEvent, configuration: RumConfiguration) {
   return {
```
bcaudan, line 135: **❓ question:** ‏What is the reasoning for moving this logic to the callers?

liywjl: **💭 thought:** We do some conditional logic to the variables `startClocks` and `correspondingTimingOverrides` in the lines just above the variable `tracingInfo`. IMO it feels nicer to have conditional logic in that one function. 

Happy to revert it if you feel this is unnecessary

bcaudan: IMO, this function was encapsulating all the logic around tracing info so it is a bit a shame to split it and duplicate some it. 

To improve the current state, I'd rather limit process functions responsibilities to only run various computations and assemble the results, leaving all the conditional logic to sub parts.
IMO, no need to go there now.

wdyt?

liywjl: Makes sense, I agree having logic coupled and contained to relevant functions makes it a lot easier to read through.

I will revert it and lower the newly added conditional logic down to `computeEntryTracingInfo`

BenoitZugmeyer (approved):

bcaudan (approved):

PR #1670, opened 2022-08-02
⚡[RUMF-1344] Serialize scroll positions only for full snapshots

bcaudan:
## Motivation

Accessing scroll positions of DOM elements during serialization is costly so let's avoid to do it when we can.

## Changes

First step, when serializing:

- DOM mutations => don't serialize scroll positions
- full snapshots => access scroll attributes to serialize scroll positions

Follow up in next PR:
- access scroll attributes only on initial full snapshot

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/rum/src/domain/record/serialize.spec.ts
```diff
@@ -114,22 +115,41 @@ describe('serializeNodeWithId', () => {
...
+      },
+    ].forEach(({ description, serializationContext, shouldSerializeScroll }) => {
+      it(description, () => {
+        const element = document.createElement('div')
+        Object.assign(element.style, { width: '100px', height: '100px', overflow: 'scroll' })
+        const inner = document.createElement('div')
+        Object.assign(inner.style, { width: '200px', height: '200px' })
+        element.appendChild(inner)
+        sandbox.appendChild(element)
+        element.scrollBy(10, 20)
```
liywjl, line 137: **🥜: nitpick:** would be nice to have this run in a `beforeEach` block

liywjl (approved):
Nice work!

packages/rum/src/domain/record/serialize.spec.ts
bcaudan: What benefit do you see in extracting that to a `beforeEach` statement?

liywjl: On second thought, I find these lines quite useful to understand the context of each test. Feel free to ignore

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.spec.ts
```diff
@@ -351,10 +353,17 @@ describe('rum track view metrics', () => {
...
-        isLayoutShiftSupported ? ['layout-shift'] : []
-      )
+      Object.defineProperty(PerformanceObserver, 'supportedEntryTypes', {
+        get: () => (isLayoutShiftSupported ? ['layout-shift'] : []),
+      })
+    })
+
+    afterEach(() => {
+      if (originalSupportedEntryTypes) {
+        Object.defineProperty(PerformanceObserver, 'supportedEntryTypes', originalSupportedEntryTypes)
```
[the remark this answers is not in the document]
bcaudan: By bumping safari version, `supportedEntryTypes` is now available ... but not in the same way than the other browsers 😅 
So, since there is no `get` method on his property descriptor, I had to find an alternate strategy.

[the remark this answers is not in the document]
BenoitZugmeyer: I see, thanks!

BenoitZugmeyer (approved):

PR #1677, opened 2022-08-04
Improve legend in Changelog to provide context on more icons

fhightower:
## Motivation

There are some icons in the changelog (e.g. the 🔒 in [v4.16.0](https://github.com/DataDog/browser-sdk/blob/main/CHANGELOG.md#v4160)) which are not mentioned in the legend. I think it would be helpful to help readers know what each icon means.

## Changes

There are two approaches:

1. Add all of the icons which are used in the changelog to the legend
2. Keep commonly used icons in the legend and add a link to the [gitmoji guide](https://gitmoji.dev/) where readers can look up less commonly used icons if they are interested

I've opted for the later approach as it is more extensible and won't require regular updates, but please let me know if you'd rather go with the first approach (or a different approach entirely) and I'm happy to make that change.

Thanks!


BenoitZugmeyer:
Thank you! We only intended to list user-facing changes in the changelog, so not all gitmojis make sense. But this rule is not always followed and some non-user-facing changes are showing up, bringing undocumented gitmojis. I'll bring this topic to the team, to see how we can improve the situation :)

PR #1680, opened 2022-08-05
⚡ [RUMF-1344] Access scroll attributes only on initial full snapshot

bcaudan:
## Motivation

Accessing scroll positions of DOM elements during serialization is costly so let's avoid to do it when we can.

## Changes

Following #1670, remove more scroll attributes accesses by:

- keeping in memory scroll positions during initial full snapshot and scroll events
- retrieve from memory scroll positions during subsequent full snapshot

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/elementsScrollPositions.ts
+++ b/packages/rum/src/domain/record/elementsScrollPositions.ts
@@ -1,0 +1,23 @@
+import { addTelemetryDebug } from '@datadog/browser-core'
+
+export type ElementsScrollPositions = ReturnType<typeof createElementsScrollPositions>
+export type ScrollPositions = { scrollLeft: number; scrollTop: number }
+
+export function createElementsScrollPositions() {
+  const scrollPositionsByElement = new WeakMap<Element, ScrollPositions>()
+  const documentScrollingElement = document.scrollingElement!
+  if (!documentScrollingElement) {
+    addTelemetryDebug('document without scrollingElement')
+  }
+  return {
+    set(element: Element | Document, scrollPositions: ScrollPositions) {
+      scrollPositionsByElement.set(
+        element === document ? documentScrollingElement : (element as Element),
+        scrollPositions
+      )
+    },
+    get(element: Element) {
+      return scrollPositionsByElement.get(element)
+    },
+  }
+}
--- a/packages/rum/src/domain/record/elementsScrollPositions.ts
+++ b/packages/rum/src/domain/record/elementsScrollPositions.ts
@@ -19,5 +19,8 @@ export function createElementsScrollPositions() {
     get(element: Element) {
       return scrollPositionsByElement.get(element)
     },
+    has(element: Element) {
+      return scrollPositionsByElement.has(element)
+    },
   }
 }
--- a/packages/rum/src/domain/record/record.ts
+++ b/packages/rum/src/domain/record/record.ts
@@ -13,7 +13,7 @@ import type {
   ViewportResizeData,
 } from '../../types'
 import { RecordType, IncrementalSource } from '../../types'
-import { serializeDocument, SerializationContext } from './serialize'
+import { serializeDocument, SerializationContextStatus } from './serialize'
 import { initObservers } from './observers'
 
 import { MutationController } from './mutationObserver'
@@ -45,7 +45,7 @@ export function record(options: RecordOptions): RecordAPI {
 
   const takeFullSnapshot = (
     timestamp = timeStampNow(),
-    serializationContext = SerializationContext.INITIAL_FULL_SNAPSHOT
+    serializationContext = { status: SerializationContextStatus.INITIAL_FULL_SNAPSHOT, elementsScrollPositions }
   ) => {
     mutationController.flush() // process any pending mutation before taking a full snapshot
     const { width, height } = getViewportDimension()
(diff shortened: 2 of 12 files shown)


packages/rum/src/domain/record/elementsScrollPositions.ts
```diff
@@ -0,0 +1,27 @@
+export type ElementsScrollPositions = ReturnType<typeof createElementsScrollPositions>
+export type ScrollPositions = { scrollLeft: number; scrollTop: number }
+
+export function createElementsScrollPositions() {
+  const scrollPositionsByElement = new WeakMap<Element, ScrollPositions>()
+  /**
+   * The element carrying the document scroll positions is not necessarily the same on all browsers.
+   * HTML and BODY elements has been observed but other cases might exist.
+   * It should be the first serialized elements though.
+   */
```
BenoitZugmeyer, line 10: **💭 thought:** ‏re: “It should be the first serialized elements though.”: FMU, `set` is not necessarily called in serialization order though.


**💬 suggestion:** ‏use [`document.scrollingElement`](https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement) when available

packages/rum/src/domain/record/serialize.ts
```diff
@@ -34,25 +35,33 @@ type ParentNodePrivacyLevel =
   | typeof NodePrivacyLevel.MASK
   | typeof NodePrivacyLevel.MASK_USER_INPUT
 
-export type SerializationContext = 'full-snapshot' | 'mutation'
+export enum SerializationContext {
```
BenoitZugmeyer, line 38: **🥜 nitpick:** ‏you could use a const enum here

packages/rum/src/domain/record/serialize.ts
```diff
@@ -34,25 +35,33 @@ type ParentNodePrivacyLevel =
...
+  MUTATION,
+}
 
 export interface SerializeOptions {
   document: Document
   serializedNodeIds?: Set<number>
   ignoreWhiteSpace?: boolean
   parentNodePrivacyLevel: ParentNodePrivacyLevel
   serializationContext: SerializationContext
+  elementsScrollPositions?: ElementsScrollPositions
```
BenoitZugmeyer, line 50: **💭 thought:** ‏it would be better to have this always defined. I know it is not used during `serializationContext === mutation`, but it is not obvious that those two options are linked together.


**💬 suggestion:** ‏If you want to avoid making this mandatory when it is not used, what about using a more explicit `SerializationContext` like this:

```ts
enum SerializationContextStatus {
  MUTATION,
  INITIAL_FULL_SNAPSHOT,
  SUBSEQUENT_FULL_SNAPSHOT,
}
type SerializationContext =
  {
    status: SerializationContextStatus.MUTATION
  } |
  {
    status: SerializationContextStatus.INITIAL_FULL_SNAPSHOT,
    elementsScrollPositions: ElementsScrollPosition,
  } |
  {
    elementsScrollPositions: ElementsScrollPosition,
  }
```

packages/rum/src/domain/record/serialize.ts
```diff
@@ -401,13 +416,26 @@ function getAttributesForPrivacyLevel(
...
-    }
-    if (element.scrollTop) {
-      safeAttrs.rr_scrollTop = Math.round(element.scrollTop)
-    }
+  let scrollPositions
+  switch (serializationContext) {
+    case SerializationContext.INITIAL_FULL_SNAPSHOT:
+      scrollPositions = {
+        scrollTop: element.scrollTop && Math.round(element.scrollTop),
+        scrollLeft: element.scrollLeft && Math.round(element.scrollLeft),
```
BenoitZugmeyer, line 424: **💬 suggestion:** ‏is testing for scroll position truthiness useful here?
```suggestion
        scrollTop: Math.round(element.scrollTop),
        scrollLeft: Math.round(element.scrollLeft),
```

packages/rum/src/domain/record/serialize.ts
```diff
@@ -401,13 +416,26 @@ function getAttributesForPrivacyLevel(
...
    * Serialize the scroll state for each element only for full snapshot
    */
-  if (serializationContext === 'full-snapshot') {
-    if (element.scrollLeft) {
-      safeAttrs.rr_scrollLeft = Math.round(element.scrollLeft)
-    }
-    if (element.scrollTop) {
-      safeAttrs.rr_scrollTop = Math.round(element.scrollTop)
-    }
+  let scrollPositions
```
BenoitZugmeyer, line 419: **🥜 nitpick:** ‏I think using two independent variables here would be easier on the bundle size (no repeated property access, no `?.` operator):

```suggestion
  let scrollTop: number | undefined
  let scrollLeft: number | undefined
```

packages/rum/src/domain/record/record.ts
```diff
@@ -28,7 +29,7 @@ export interface RecordOptions {
 
 export interface RecordAPI {
   stop: () => void
-  takeFullSnapshot: (timestamp?: TimeStamp) => void
+  takeFullSnapshot: (timestamp?: TimeStamp, serializationContext?: SerializationContext) => void
```
BenoitZugmeyer, line 32: **💬 suggestion:** ‏FMU, the `serializationContext` argument should never be the default value when invoked outside of this file. So I would make it mandatory, as it would be less error-prone (= we wouldn't inadvertently miss it).

BenoitZugmeyer (approved):
Great!

liywjl (approved):

PR #1682, opened 2022-08-08
⚗️ [RUMF-1333] Experiment with fetch keepalive

bcaudan:
## Motivation

In order to improve our data transfer, experiment with fetch keepalive in order to later be able to access to intake response.

## Changes

Behind `fetch-keepalive` flag, use fetch keepalive except on visibility hidden and before unload

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/httpRequest.spec.ts
+++ b/packages/core/src/transport/httpRequest.spec.ts
@@ -25,7 +25,7 @@ describe('httpRequest', () => {
     interceptor.restore()
   })
 
-  describe('send (without fetch-keepalive FF)', () => {
+  describe('send (without fetch_keepalive FF)', () => {
     it('should use xhr when sendBeacon is not defined', () => {
       interceptor.withSendBeacon(false)
 
@@ -84,9 +84,9 @@ describe('httpRequest', () => {
     })
   })
 
-  describe('send (with fetch-keepalive FF)', () => {
+  describe('send (with fetch_keepalive FF)', () => {
     beforeEach(() => {
-      updateExperimentalFeatures(['fetch-keepalive'])
+      updateExperimentalFeatures(['fetch_keepalive'])
     })
 
     afterEach(() => {
--- a/packages/core/src/transport/httpRequest.spec.ts
+++ b/packages/core/src/transport/httpRequest.spec.ts
@@ -142,11 +142,11 @@ describe('httpRequest', () => {
     })
   })
 
-  describe('sendBeacon', () => {
+  describe('sendOnExit', () => {
     it('should use xhr when sendBeacon is not defined', () => {
       interceptor.withSendBeacon(false)
 
-      request.sendBeacon('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
+      request.sendOnExit('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
 
       expect(requests.length).toEqual(1)
       expect(requests[0].type).toBe('xhr')
@@ -159,14 +159,14 @@ describe('httpRequest', () => {
         pending('no sendBeacon support')
       }
 
-      request.sendBeacon('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
+      request.sendOnExit('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
 
       expect(requests.length).toEqual(1)
       expect(requests[0].type).toBe('sendBeacon')
     })
 
     it('should use xhr over sendBeacon when the bytes count is too high', () => {
-      request.sendBeacon('{"foo":"bar1"}\n{"foo":"bar2"}', BATCH_BYTES_LIMIT)
+      request.sendOnExit('{"foo":"bar1"}\n{"foo":"bar2"}', BATCH_BYTES_LIMIT)
 
       expect(requests.length).toEqual(1)
       expect(requests[0].type).toBe('xhr')
@@ -178,7 +178,7 @@ describe('httpRequest', () => {
       }
       interceptor.withSendBeacon(() => false)
 
-      request.sendBeacon('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
+      request.sendOnExit('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
 
       expect(requests.length).toEqual(1)
       expect(requests[0].type).toBe('xhr')
@@ -194,7 +194,7 @@ describe('httpRequest', () => {
         throw new TypeError()
       })
 
-      request.sendBeacon('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
+      request.sendOnExit('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
 
       expect(sendBeaconCalled).toBe(true)
       expect(requests.length).toEqual(1)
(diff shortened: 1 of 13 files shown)


liywjl (approved):

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -1,101 +1,105 @@
...
     request.send('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
 
-    expect(server.requests.length).toEqual(1)
-    expect(server.requests[0].url).toContain(ENDPOINT_URL)
-    expect(server.requests[0].requestBody).toEqual('{"foo":"bar1"}\n{"foo":"bar2"}')
+    expect(requests.length).toEqual(1)
+    expect(requests[0].type).toBe('xhr')
+    expect(requests[0].url).toContain(ENDPOINT_URL)
+    expect(requests[0].body).toEqual('{"foo":"bar1"}\n{"foo":"bar2"}')
+    navigator.sendBeacon = originalSendBeacon
```
BenoitZugmeyer: **💬 suggestion:** restore the state in an `afterEach` callback

packages/rum/src/boot/startRecording.spec.ts
```diff
@@ -61,7 +57,13 @@ describe('startRecording', () => {
...
         })
         .beforeBuild(({ lifeCycle, configuration, viewContexts, sessionManager }) => {
-          const recording = startRecording(lifeCycle, configuration, sessionManager, viewContexts, worker!)
+          const httpRequest = createHttpRequest(configuration.sessionReplayEndpointBuilder, SEGMENT_BYTES_LIMIT)
+
+          const requestSendSpy = spyOn(httpRequest, 'send')
+          ;({ waitAsyncCalls: waitRequestSendCalls, expectNoExtraAsyncCall: expectNoExtraRequestSendCalls } =
+            collectAsyncCalls(requestSendSpy))
+
+          const recording = startRecording(lifeCycle, configuration, sessionManager, viewContexts, worker!, httpRequest)
```
BenoitZugmeyer, line 66: **💬 suggestion:** ‏use `interceptRequests` instead of introducing a sixth parameter here

packages/core/src/transport/httpRequest.ts
```diff
@@ -13,26 +15,55 @@ import { addTelemetryError } from '../domain/telemetry'
...
+  return {
+    send: (data: string | FormData, bytesCount: number) => {
+      // TODO check tests without flag
+      if (!isExperimentalFeatureEnabled('fetch-keepalive')) {
+        sendBeacon(data, bytesCount)
+      } else {
+        sendFetch(data, bytesCount)
+      }
     },
+    sendBeacon,
```
BenoitZugmeyer, line 66: **💬 suggestion:** ‏This method is a bit confusing because it might not even use `sendBeacon` at all.  Maybe rename this method (and related methods in `Batch`) to something more explicit like `sendUrgently` (since it is used when leaving the page)

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -19,62 +19,185 @@ describe('httpRequest', () => {
...
 
-    expect(requests.length).toEqual(1)
-    expect(requests[0].type).toBe('xhr')
+      request.send('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
+
+      expect(requests.length).toEqual(1)
+      expect(requests[0].type).toBe('xhr')
+      expect(requests[0].url).toContain(ENDPOINT_URL)
+      expect(requests[0].body).toEqual('{"foo":"bar1"}\n{"foo":"bar2"}')
+      window.Request = originalRequest
```
BenoitZugmeyer, line 101: **💬 suggestion:** ‏restore the state in an `afterEach` callback

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -19,62 +19,185 @@ describe('httpRequest', () => {
...
+      const originalSendBeacon = navigator.sendBeacon && navigator.sendBeacon.bind(navigator)
+      ;(navigator.sendBeacon as any) = false
+
+      request.sendBeacon('{"foo":"bar1"}\n{"foo":"bar2"}', 10)
+
+      expect(requests.length).toEqual(1)
+      expect(requests[0].type).toBe('xhr')
+      expect(requests[0].url).toContain(ENDPOINT_URL)
+      expect(requests[0].body).toEqual('{"foo":"bar1"}\n{"foo":"bar2"}')
+      navigator.sendBeacon = originalSendBeacon
```
BenoitZugmeyer, line 154: **💬 suggestion:** ‏restore the state in a `afterEach` callback

packages/core/src/transport/httpRequest.ts
```diff
@@ -13,26 +15,55 @@ import { addTelemetryError } from '../domain/telemetry'
...
+    request.open('POST', url, true)
+    request.send(data)
+  }
 
-      const request = new XMLHttpRequest()
-      request.open('POST', url, true)
-      request.send(data)
+  return {
+    send: (data: string | FormData, bytesCount: number) => {
+      // TODO check tests without flag
```
BenoitZugmeyer, line 59: **❓ question:** ‏not sure what this comment means?

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -19,62 +19,185 @@ describe('httpRequest', () => {
...
+    it('should fallback to xhr when fetch keepalive is not queued', (done) => {
+      if (!('fetch' in window) || !('keepalive' in new window.Request(''))) {
+        pending('no fetch keepalive support')
+      }
+
+      let notQueuedFetch: Promise<never>
+      spyOn(window, 'fetch').and.callFake(() => {
+        notQueuedFetch = Promise.reject()
+        return notQueuedFetch
+      })
```
BenoitZugmeyer, line 131: **💬 suggestion:** ‏`jasmine.getEnv().allowRespy(true)` bothers me a bit. In the past, I got yield at by jasmine because some spies were not correctly cleaned up. By allowing "respy" globally, I think it opens the door to more mistakes and overly complex patterns.

```ts
// in beforeEach
const { requests, mockNextRequest } = interceptRequests()

// in test case
mockNextRequest('fetch', () => Promise.reject())
```

packages/core/src/transport/httpRequest.ts
bcaudan: just a leftover :swe

packages/rum/src/boot/startRecording.spec.ts
bcaudan: 🤔 it would not play well with the `collectAsyncCalls` logic though and `interceptRequests` only deal with sync case for now.
Unless I am missing something, I'd be in favor of keeping those separated for now and see how it will evolve with the retry logic before trying to merge those helpers.

packages/core/src/transport/httpRequest.ts
bcaudan: yep, not super fan of the naming as well.

> since it is used when leaving the page

`sendOnExit`?

BenoitZugmeyer: Sounds good to me!

packages/rum/src/boot/startRecording.spec.ts
BenoitZugmeyer: Alright 👌

BenoitZugmeyer (approved):

BenoitZugmeyer:
Hum I think using `snake_case` for the feature flag would make sense now. See https://github.com/DataDog/consul-config/pull/130587#discussion_r908262285

PR #1684, opened 2022-08-09
✨⚗ [RUMF-1355] add selector with stable attributes

BenoitZugmeyer:
## Motivation

Use common "stable" attributes to improve CSS selector generation for heatmaps.

## Changes

* refactor how we compute the click action base. This has two purposes:
  * avoid computing selector twice when cloning the Click object
  * helps passing the `actionNameAttribute` option to the function used to compute selectors
* tweak types to allow using multiple selectors
* change selector computation to make it easier to customize
* add selector with stable attributes

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/action/getSelectorsFromElement.ts
```diff
@@ -0,0 +1,136 @@
...
+    const uniqueSelector = findSelector(element, globallyUniqueSelectorStrategies, isSelectorUniqueGlobally)
+    if (uniqueSelector) {
+      targetElementSelector.unshift(uniqueSelector)
+      break
+    }
+
+    targetElementSelector.unshift(
+      findSelector(element, uniqueAmongChildrenSelectorStrategies, isSelectorUniqueAmongChildren) ||
+        getPositionSelector(element)
+    )
```
bcaudan, line 59: **🥜 nitpick:** ‏having a similar flow could ease the understanding

```suggestion
    const uniqueSelector = findSelector(element, globallyUniqueSelectorStrategies, isSelectorUniqueGlobally)
    if (uniqueSelector) {
      targetElementSelector.unshift(uniqueSelector)
      break
    }

    const uniqueSelectorAmongChildren = findSelector(element, uniqueAmongChildrenSelectorStrategies, isSelectorUniqueAmongChildren)
    if (uniqueSelectorAmongChildren) {
      targetElementSelector.unshift(uniqueSelectorAmongChildren)
    } else {
      targetElementSelector.unshift(getPositionSelector(element))
    }
```

bcaudan (approved):
Great!

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorsFromElement.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function isSelectorUniqueAmongChildren(element: Element, selector: string): boolean {
+  for (let i = 0; i < element.parentElement!.children.length; i++) {
+    const sibling = element.parentElement!.children[i]
+    if (sibling !== element && elementMatches(sibling, selector)) {
+      return false
+    }
+  }
+  return true
+}
```
liywjl, line 136: **👏 praise:** great naming

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -172,6 +167,36 @@ export function trackClickActions(
...
+
+  return {
+    type: 'click',
+    target,
+    position,
+    name: getActionNameFromElement(event.target, actionNameAttribute),
+    event,
+    startClocks: clocksNow(),
+  }
+}
```
liywjl, line 198: 👏

liywjl (approved):
Awesome work!

packages/rum-core/src/domain/rumEventsCollection/action/getSelectorsFromElement.ts
```diff
@@ -0,0 +1,136 @@
...
+  'data-qa-id',
+  'data-testing',
+  // Fullstory decorator attributes:
+  'data-component',
+  'data-element',
+  'data-source-file',
+]
+
+export function getSelectorsFromElement(element: Element, actionNameAttribute: string | undefined) {
+  const attributeSelectors = STABLE_ATTRIBUTES.map((attribute) => getAttributeSelector.bind(null, attribute))
```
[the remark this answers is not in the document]
BenoitZugmeyer: `bind` does not call the function. `bind` returns a new function with some arguments defined (cf [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind)). Basically, 

```
STABLE_ATTRIBUTES.map((attribute) => getAttributeSelector.bind(null, attribute))
```
is the same as
```
STABLE_ATTRIBUTES.map((attribute) => (element: Element) => getAttributeSelector(attribute, element))
```
Would it be clearer?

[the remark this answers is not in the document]
liywjl: Ah so we are using it as a form of currying? Thanks for the clarification

I personally wouldn't use `bind` unless we care about `this`; but that is more personal preference.

PR #1685, opened 2022-08-10
🐛 fix compatibility check

BenoitZugmeyer:
## Motivation

* We previously removed node_modules without reinstalling, so the parent TS version was used instead of the test/app one. By installing test/app dependencies, we make sure the correct version of TS is installed. Note: compatibility errors were caught by our ssr compatibility check

* A sed command worked on Mac but not on Linux, so we didn't test for `exactOptionalPropertyTypes` compatibility on CI

## Changes

* Install test/app dependencies before checking compatibility
* Replace sed by perl

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/cli
+++ b/scripts/cli
@@ -72,6 +72,7 @@ cmd_check_typescript_compatibility () {
   yarn build
   cd test/app
   rm -rf node_modules
+  yarn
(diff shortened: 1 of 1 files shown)


bcaudan (approved):
LGTM

PR #1687, opened 2022-08-11
🐛[RUMF-1344] handle document scroll when document has no scrolling element

bcaudan:
## Motivation

Following #1680, some telemetry events show that some documents seems to lack scrolling element:

- specific debug log: 'unable to find document scrolling element'
- error on setting null as a weak map key

I have not been able to reproduce the behavior while going on available customer pages.

## Changes

- when document scrolling element is not available, try to find it after a document scroll
- catch weak map set error to see if we face other cases

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/elementsScrollPositions.spec.ts
+++ b/packages/rum/src/domain/record/elementsScrollPositions.spec.ts
@@ -1,0 +1,45 @@
+import { createElementsScrollPositions } from './elementsScrollPositions'
+
+describe('elementsScrollPositions', () => {
+  beforeEach(() => {
+    document.body.style.setProperty('height', '5000px')
+    document.body.style.setProperty('width', '5000px')
+    window.scroll(10, 20)
+  })
+
+  afterEach(() => {
+    window.scroll(0, 0)
+    document.body.style.removeProperty('height')
+    document.body.style.removeProperty('width')
+  })
+
+  it('should attach document scroll positions to document scrolling element', () => {
+    const elementsScrollPositions = createElementsScrollPositions()
+
+    elementsScrollPositions.set(document, { scrollLeft: 10, scrollTop: 20 })
+
+    expect(elementsScrollPositions.get(document.scrollingElement!)).toEqual({ scrollLeft: 10, scrollTop: 20 })
+  })
+
+  it('should attach document scroll positions to unavailable document scrolling element', () => {
+    const documentScrollingElement = document.scrollingElement!
+    Object.defineProperty(document, 'scrollingElement', { value: null, configurable: true })
+    const elementsScrollPositions = createElementsScrollPositions()
+    elementsScrollPositions.set(document, { scrollLeft: 10, scrollTop: 20 })
+
+    expect(elementsScrollPositions.get(documentScrollingElement)).toEqual({ scrollLeft: 10, scrollTop: 20 })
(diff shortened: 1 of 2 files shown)


liywjl (approved):
Nice one!

liywjl (approved):

packages/rum/src/domain/record/elementsScrollPositions.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+
+  it('should attach document scroll positions to unavailable document scrolling element', () => {
+    const documentScrollingElement = document.scrollingElement!
+    Object.defineProperty(document, 'scrollingElement', { value: null, configurable: true })
+    const elementsScrollPositions = createElementsScrollPositions()
+    elementsScrollPositions.set(document, { scrollLeft: 10, scrollTop: 20 })
+
+    expect(elementsScrollPositions.get(documentScrollingElement)).toEqual({ scrollLeft: 10, scrollTop: 20 })
+    Object.defineProperty(document, 'scrollingElement', { value: documentScrollingElement, configurable: true })
```
BenoitZugmeyer, line 31: **💬 suggestion:** ‏Restore original scrollingElement in an `afterEach` callback to make sure it is correctly restored

packages/rum/src/domain/record/elementsScrollPositions.ts
```diff
@@ -1,20 +1,29 @@
...
 export type ElementsScrollPositions = ReturnType<typeof createElementsScrollPositions>
 export type ScrollPositions = { scrollLeft: number; scrollTop: number }
 
 export function createElementsScrollPositions() {
   const scrollPositionsByElement = new WeakMap<Element, ScrollPositions>()
-  const documentScrollingElement = document.scrollingElement!
-  if (!documentScrollingElement) {
-    addTelemetryDebug('document without scrollingElement')
-  }
+  let documentScrollingElement = document.scrollingElement
```
BenoitZugmeyer, line 9: From MDN documentation, “In standards mode, [`document.scrollingElement`] is the root element of the document, [document.documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement).”. 

Thus, in standard mode, we can fallback to `document.documentElement` if we don't have `scrollingElement`.


**❓ question:** ‏ how far should we go to support non-standard modes? Is it even supported by the player?

packages/rum/src/domain/record/elementsScrollPositions.ts
```diff
@@ -24,3 +33,21 @@ export function createElementsScrollPositions() {
...
 }
+
+function findDocumentScrollingElement(childNodes: NodeList, scrollPositions: ScrollPositions) {
+  let scrollingElement: Element | null = null
+  forEach(childNodes, (node) => {
+    if (!scrollingElement && node.nodeType === node.ELEMENT_NODE) {
+      const element = node as Element
+      if (
+        Math.round(element.scrollTop) === scrollPositions.scrollTop &&
+        Math.round(element.scrollLeft) === scrollPositions.scrollLeft
```
BenoitZugmeyer, line 44: **🔨 warning:** ‏If `scrollTop` and `scrollLeft` are 0, it would match any non-scrollable element.

PR #1688, opened 2022-08-11
🐛 [RUMF-1344] try to improve document scrolling element handling

bcaudan:
## Motivation

Following #1680, some telemetry events show that some documents seems to lack scrolling element:

- specific debug log: 'unable to find document scrolling element'
- error on setting null as a weak map key

I have not been able to reproduce the behavior while going on available customer pages.

## Changes

- access document.scrollingElement only on document scroll
- try to fallback on document.documentElement
- improve logging

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/elementsScrollPositions.ts
+++ b/packages/rum/src/domain/record/elementsScrollPositions.ts
@@ -36,6 +36,7 @@ function tryToFindScrollingElement(scrollPositions: ScrollPositions) {
   if (document.scrollingElement) {
     return document.scrollingElement
   }
+  addTelemetryDebug('null document scrolling element')
   if (scrollPositions.scrollLeft === 0 && scrollPositions.scrollTop === 0) {
     addTelemetryDebug('Unable to find scrolling element for scroll (0,0)')
     return null
--- a/packages/rum/src/domain/record/elementsScrollPositions.ts
+++ b/packages/rum/src/domain/record/elementsScrollPositions.ts
@@ -5,15 +5,18 @@ export type ScrollPositions = { scrollLeft: number; scrollTop: number }
 
 export function createElementsScrollPositions() {
   const scrollPositionsByElement = new WeakMap<Element, ScrollPositions>()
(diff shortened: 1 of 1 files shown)


packages/rum/src/domain/record/elementsScrollPositions.ts
```diff
@@ -24,3 +31,21 @@ export function createElementsScrollPositions() {
...
+  }
+  if (scrollPositions.scrollLeft === 0 && scrollPositions.scrollTop === 0) {
+    addTelemetryDebug('Unable to find scrolling element for scroll (0,0)')
+    return null
+  }
+  if (
+    Math.round(document.documentElement.scrollLeft) === scrollPositions.scrollLeft &&
+    Math.round(document.documentElement.scrollTop) === scrollPositions.scrollTop
+  ) {
+    return document.documentElement
```
BenoitZugmeyer, line 47: **💬 suggestion:** Previously, we used `document.scrollingElement` before any scroll occured. Now that we are using it only when a scroll occurs, ‏I'd be curious of how often we get in this case now. Would you mind adding a telemetry log here?

BenoitZugmeyer (approved):

packages/rum/src/domain/record/elementsScrollPositions.ts
bcaudan: is [0c65a8d](https://github.com/DataDog/browser-sdk/pull/1688/commits/0c65a8d5f03efe11f9e62a3994cdeb82f641c2e6) what you had in mind?

BenoitZugmeyer (approved):

liywjl (approved):
Unit tests might be worth adding

packages/rum/src/domain/record/elementsScrollPositions.ts
BenoitZugmeyer: Yes it does the job. The idea was to have a log at every codepath when `scrollingElement` was missing. Adding it here works for me.

PR #1694, opened 2022-08-16
🐛 [RUMF-1344] remove fallback for null scrollingElement

bcaudan:
## Motivation

Following #1688, current fallback does not improve much the situation.

## Changes

- Remove fallback when scrolling element is null and don't handle document scroll in this case.
- Remove try/catch around weakmap as it should not happen anymore

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/elementsScrollPositions.ts
+++ b/packages/rum/src/domain/record/elementsScrollPositions.ts
@@ -1,27 +1,19 @@
-import { addTelemetryDebug } from '@datadog/browser-core'
-
 export type ElementsScrollPositions = ReturnType<typeof createElementsScrollPositions>
 export type ScrollPositions = { scrollLeft: number; scrollTop: number }
 
 export function createElementsScrollPositions() {
   const scrollPositionsByElement = new WeakMap<Element, ScrollPositions>()
-  let documentScrollingElement: Element | null
   return {
     set(element: Element | Document, scrollPositions: ScrollPositions) {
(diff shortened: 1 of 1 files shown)


packages/rum/src/domain/record/elementsScrollPositions.ts
```diff
@@ -1,27 +1,19 @@
...
-        addTelemetryDebug(`invalid element: ${String(element)}`)
+      if (element === document && !document.scrollingElement) {
+        // cf https://drafts.csswg.org/cssom-view/#dom-document-scrollingelement,
+        // in some cases scrolling elements can not be defined, we don't support those for now
+        return
       }
+      scrollPositionsByElement.set(
+        element === document ? document.scrollingElement! : (element as Element),
+        scrollPositions
+      )
```
BenoitZugmeyer, line 16: **🥜 nitpick:** ‏I feel writing it like that would be a tiny bit clearer, but either is fine

```suggestion
      if (element === document) {
        if (!document.scrollingElement) {
          // cf https://drafts.csswg.org/cssom-view/#dom-document-scrollingelement,
          // in some cases scrolling elements can not be defined, we don't support those for now
          return
        }
        element = document.scrollingElement
      }
      scrollPositionsByElement.set((element as Element), scrollPositions)
```

BenoitZugmeyer (approved):

liywjl (approved):

Issue #1570, opened 2022-05-25
Cannot read properties of undefined (reading 'apply') 


michaelAtCoalesce:
**Cannot read properties of undefined (reading 'apply')** 

TypeError: Cannot read properties of undefined (reading 'apply')```

the code that fails is clearly in the datadog RUM 

../../src/tools/instrumentMethod.ts::stop:17 
../../src/browser/xhrProxy.ts:(XMLHttpRequest.onreadystatechange::119 

```

This caused a failure on load for our application

        **"@datadog/browser-rum": "^4.7.1",**

BenoitZugmeyer:
Hello @michaelAtCoalesce, thanks for the report.

The file you are indicating, `xhrProxy.ts`, does not exist since version `3.6.12`. Also it didn't use `instrumentMethod` back then. So I suspect something is wrong either with the stack trace or the version you are providing.

Could you provide additional details? A small example to reproduce the issue would be great!

Ryankey:
I've recently come into what I think is the same issue on "^4.14.0". In our case, we have Datadog RUM, Sentry (for error watching) and a third library molstar combining to produce this error on only one part of our application.

The order of events is...

1. Molstar makes an initial xhr request
2. Datadog RUM wraps the `xhr.onreadystatechange` (which is currently null, as many callbacks are) on [L13](https://github.com/DataDog/browser-sdk/blob/ed8a9bfdb21e4067c795f81804617b3ddf121408/packages/core/src/tools/instrumentMethod.ts#L13). This wrapping is good and will succeed on our first request because the wrapping function handles non-functions on [L55](https://github.com/DataDog/browser-sdk/blob/ed8a9bfdb21e4067c795f81804617b3ddf121408/packages/core/src/tools/instrumentMethod.ts#L55)
3. Sentry wraps `xhr.onreadystatechange` for error watching
4. `stop()` is called on the wrapper from 2, but because of Sentry's wrap from 3 we fail the condition on [L23](https://github.com/DataDog/browser-sdk/blob/ed8a9bfdb21e4067c795f81804617b3ddf121408/packages/core/src/tools/instrumentMethod.ts#L23), which causes us to fall back to L26, replacing our safe monitoring wrapper with the original null
5. Molstar makes another xhr request, and uses a pool of xhr objects, reusing the previous one
[...]

PR #1697, opened 2022-08-17
🐛 Fix instrumention of null function with 3rd party wrapper (#1570)

Ryankey:
## Motivation

Issue #1570 is causing noisy errors when wrapping a null/undefined method that also has third party wrappers.

## Changes

Adds a condition when executing a wrapped function to protect against the case when we "unwrap" the function ourselves and set it to null (its previous value).

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/instrumentMethod.spec.ts
+++ b/packages/core/src/tools/instrumentMethod.spec.ts
@@ -54,6 +54,17 @@ describe('instrumentMethod', () => {
     expect(instrumentationSpy).toHaveBeenCalled()
   })
 
+  it('handles undefined methods with third party wrappers after stop', () => {
+    const object: { method?: () => number } = {}
+    const instrumentationStub = () => 2
+    const { stop } = instrumentMethod(object, 'method', () => instrumentationStub)
+
+    thirdPartyInstrumentation(object)
+
+    stop()
+    expect(object.method).not.toThrow()
+  })
+
   describe('stop()', () => {
     it('restores the original behavior', () => {
       const object = { method: () => 1 }
@@ -102,9 +113,11 @@ describe('instrumentMethod', () => {
     })
   })
 
-  function thirdPartyInstrumentation(object: { method: () => number }) {
+  function thirdPartyInstrumentation(object: { method?: () => number }) {
     const originalMethod = object.method
-    object.method = () => originalMethod() + 2
+    if (typeof originalMethod === 'function') {
+      object.method = () => originalMethod() + 2
+    }
   }
 })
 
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/instrumentMethod.ts
```diff
@@ -12,9 +12,12 @@ export function instrumentMethod<OBJECT extends { [key: string]: any }, METHOD e
...
 
-  const instrumentationWrapper = function (this: OBJECT): ReturnType<OBJECT[METHOD]> {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-    return instrumentation.apply(this, arguments as unknown as Parameters<OBJECT[METHOD]>)
+  const instrumentationWrapper = function (this: OBJECT): ReturnType<OBJECT[METHOD]> | null {
+      if (typeof instrumentation === 'function') {
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+          return instrumentation.apply(this, arguments as unknown as Parameters<OBJECT[METHOD]>)
+      }
+      return null
```
BenoitZugmeyer, line 20: **💬 suggestion:** any reason to return `null` here? I think it would be fine to return `undefined`. It's a bit subjective, since we don't know what could be returned, but explicitly returning `null` could raise some questions.

Ryankey: `undefined` makes sense to me, I think I just got caught up on `null` from staring at errors. I also considered `void` with an empty `return` but I think `undefined` is probably more correct.

packages/core/src/tools/instrumentMethod.spec.ts
```diff
@@ -54,6 +54,17 @@ describe('instrumentMethod', () => {
     expect(instrumentationSpy).toHaveBeenCalled()
   })
 
+  it('handles undefined methods with third party wrappers after stop', () => {
```
bcaudan, line 57: **💬 suggestion:** ‏what about moving this test to the `stop()` > `when the method has been instrumented by a third party` suite?

Ryankey: Yeah, I was back and forth on which suite it belonged under. The condition is a result of stop being called, but the fix itself was more general (don't try to `apply` on a non-function), which is why I left it here.

bcaudan: OK, since this test specifically target `stop` for third party, colocating it with the other related tests would be more appropriate.
If you see more general test cases related to this fix, please add them 🙂

Ryankey: Thanks for the guidance, moved the test under `stop() -> third party`.

bcaudan (approved):

PR #1699, opened 2022-08-19
RUMF-1309: Implement nested CSS support

N-Boutaib:
## Motivation

Some customers use CSS-in-JS libraries to style their applications. One of those libraries [(Stitches)](https://stitches.dev/) insert CSS rules inside CSSMedialRule `@media`, which is not tracked nowadays. The replay of those sessions seems to be broken and non-styled. 

## Changes

Not all browsers support `CSSGroupingRule` (see [CSSGroupingRule](https://caniuse.com/?search=cssgroupingrule), [CSSConditionRule](https://caniuse.com/?search=cssconditionrule), [CSSMediaRule](https://caniuse.com/?search=cssmediarule) & [CSSSupportsRule](https://caniuse.com/?search=csssupportsrule)

- New observers are added to watch over `insertRule` and `deleteRule` of `CSSGroupingRule` prototype if supported by the browser. If not supported, we fall back to `CSSMediaRule` (which is supported by all browsers) and `CSSSupportsRule` (which is not supported only by early version of browsers) and report the rule that was inserted or deleted, and how to access it from the parent `CSSStyleSheet`.
- If we assume that the `sdk` is run on a browser that support `CSSMediaRule` and doesn't support `CSSSupportsRule`, this later will be ignored.

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/observers.spec.ts
+++ b/packages/rum/src/domain/record/observers.spec.ts
@@ -194,7 +194,7 @@ describe('initStyleSheetObserver', () => {
         expect(styleSheetRule.id).toBeDefined()
         expect(styleSheetRule.removes).toBeUndefined()
         expect(styleSheetRule.adds?.length).toEqual(1)
-        expect(styleSheetRule.adds?.pop()?.index).toEqual(undefined)
+        expect(styleSheetRule.adds?.[0]?.index).toEqual(undefined)
       })
 
       it('should capture CSSStyleRule insertion when index is provided', () => {
@@ -208,7 +208,7 @@ describe('initStyleSheetObserver', () => {
         expect(styleSheetRule.id).toBeDefined()
         expect(styleSheetRule.removes).toBeUndefined()
         expect(styleSheetRule.adds?.length).toEqual(1)
-        expect(styleSheetRule.adds?.pop()?.index).toEqual(index)
+        expect(styleSheetRule.adds?.[0]?.index).toEqual(index)
       })
     })
 
@@ -225,7 +225,7 @@ describe('initStyleSheetObserver', () => {
         expect(styleSheetRule.id).toBeDefined()
         expect(styleSheetRule.adds).toBeUndefined()
         expect(styleSheetRule.removes?.length).toEqual(1)
-        expect(styleSheetRule.removes?.pop()).toEqual({ index })
+        expect(styleSheetRule.removes?.[0]).toEqual({ index })
       })
     })
   })
@@ -245,7 +245,7 @@ describe('initStyleSheetObserver', () => {
         expect(styleSheetRule.id).toBeDefined()
         expect(styleSheetRule.removes).toBeUndefined()
         expect(styleSheetRule.adds?.length).toEqual(1)
-        expect(styleSheetRule.adds?.pop()?.index).toEqual([1, 0, 1])
+        expect(styleSheetRule.adds?.[0]?.index).toEqual([1, 0, 1])
       })
 
       it('should not create record when inserting into a detached CSSGroupingRule', () => {
@@ -280,7 +280,7 @@ describe('initStyleSheetObserver', () => {
         expect(styleSheetRule.id).toBeDefined()
         expect(styleSheetRule.adds).toBeUndefined()
         expect(styleSheetRule.removes?.length).toEqual(1)
-        expect(styleSheetRule.removes?.pop()?.index).toEqual([1, 0, 0])
+        expect(styleSheetRule.removes?.[0]?.index).toEqual([1, 0, 0])
       })
 
       it('should not create record when removing from a detached CSSGroupingRule', () => {
--- a/packages/rum/src/domain/record/observers.spec.ts
+++ b/packages/rum/src/domain/record/observers.spec.ts
@@ -6,7 +6,7 @@ import { createNewEvent, isFirefox } from '../../../../core/test/specHelper'
 import { NodePrivacyLevel, PRIVACY_ATTR_NAME, PRIVACY_ATTR_VALUE_MASK_USER_INPUT } from '../../constants'
 import { RecordType } from '../../types'
 import type { FrustrationCallback, InputCallback, StyleSheetRuleCallback } from './observers'
-import { initStyleSheetObserver, initFrustrationObserver, initInputObserver } from './observers'
+import { initCSSObservers, initFrustrationObserver, initInputObserver } from './observers'
 import { serializeDocument, SerializationContextStatus } from './serialize'
 import { createElementsScrollPositions } from './elementsScrollPositions'
 import type { GroupingCSSRule } from './utils'
@@ -156,9 +156,9 @@ describe('initFrustrationObserver', () => {
   })
 })
 
-describe('initStyleSheetObserver', () => {
-  let stopStyleSheetObserver: () => void
-  let styleSheetCallbackSpy: jasmine.Spy<StyleSheetRuleCallback>
+describe('initCSSObservers', () => {
+  let stopCSSObservers: () => void
+  let cssRulesCallbackSpy: jasmine.Spy<StyleSheetRuleCallback>
   let styleElement: HTMLStyleElement
   let styleSheet: CSSStyleSheet
   const styleRule = '.selector-1 { color: #fff }'
@@ -167,7 +167,7 @@ describe('initStyleSheetObserver', () => {
     if (isIE()) {
       pending('IE not supported')
     }
-    styleSheetCallbackSpy = jasmine.createSpy()
+    cssRulesCallbackSpy = jasmine.createSpy()
     styleElement = document.createElement('style')
     document.head.appendChild(styleElement)
     styleSheet = styleElement.sheet!
@@ -179,18 +179,17 @@ describe('initStyleSheetObserver', () => {
   })
 
   afterEach(() => {
-    stopStyleSheetObserver()
+    stopCSSObservers()
     styleElement.remove()
   })
 
   describe('observing high level css stylesheet', () => {
     describe('when inserting rules into stylesheet', () => {
       it('should capture CSSStyleRule insertion when no index is provided', () => {
-        // When
-        stopStyleSheetObserver = initStyleSheetObserver(styleSheetCallbackSpy)
+        stopCSSObservers = initCSSObservers(cssRulesCallbackSpy)
         styleSheet.insertRule(styleRule)
-        // Then
-        const styleSheetRule = styleSheetCallbackSpy.calls.first().args[0]
+
+        const styleSheetRule = cssRulesCallbackSpy.calls.first().args[0]
         expect(styleSheetRule.id).toBeDefined()
         expect(styleSheetRule.removes).toBeUndefined()
         expect(styleSheetRule.adds?.length).toEqual(1)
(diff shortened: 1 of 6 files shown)


packages/rum/src/domain/record/observers.ts
```diff
@@ -371,9 +371,41 @@ function initStyleSheetObserver(cb: StyleSheetRuleCallback): ListenerHandler {
     },
   })
 
+  const { stop: restoreInsertNestedRule } = instrumentMethodAndCallOriginal(CSSGroupingRule.prototype, 'insertRule', {
+    before(rule, index) {
+      if (hasSerializedNode(this.parentStyleSheet!.ownerNode!)) {
```
BenoitZugmeyer, line 376: **🔨 warning:** ‏don't assume that a CSSGroupingRule has a stylesheet, because it might not be inserted

Screenshot 2022-08-19 at 16 26 01
.

packages/rum/src/domain/record/observers.ts
```diff
@@ -371,9 +371,41 @@ function initStyleSheetObserver(cb: StyleSheetRuleCallback): ListenerHandler {
...
+        cb({
+          id: getSerializedNodeId(this.parentStyleSheet!.ownerNode),
```
BenoitZugmeyer, line 378: **💬 suggestion:** ‏factorize a bit with something like

```ts
function withStyleSheetId(styleSheet: StyleSheet | null, callback: (id: number) => {
  if (styleSheet && hasSerializeNode(styleSheet.ownerNode)) {
    callback(getSerializedNode(styleSheet.ownerNode)
  }
})

// usage
withStyleSheetId(this.parentStyleSheet, id => cb({ id, adds: [...] )})
```

packages/rum/src/domain/record/utils.ts
```diff
@@ -28,3 +28,19 @@ export function assembleIncrementalSnapshot<Data extends BrowserIncrementalData>
     timestamp: timeStampNow(),
   }
 }
+
+export function getPathToNestedCSSRule(rule: CSSRule): number[] {
+  const path: number[] = []
+  let currentRule = rule
+  while (currentRule.parentRule instanceof CSSGroupingRule) {
```
BenoitZugmeyer, line 35: **💬 suggestion:** Do you really need to check the parentRule type? I would go with a simpler condition:

```suggestion
  while (currentRule.parentRule) {
```

We would some cast to make sure TS doesn't complain, but I think it's worth it. Especially when implementing support for browser who don't implement CSSGroupingRule

N-Boutaib: I thought I removed this casting 😅

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -154,3 +154,142 @@ describe('initFrustrationObserver', () => {
...
+  const styleRule = '.selector-1 { color: #fff }'
+
+  beforeEach(() => {
+    if (isIE()) {
+      pending('IE not supported')
+    }
+    styleSheetCallbackSpy = jasmine.createSpy()
+    styleElement = document.createElement('style')
+    document.head.appendChild(styleElement)
+    styleSheet = <CSSStyleSheet>styleElement.sheet
```
BenoitZugmeyer, line 172: **💬 suggestion:** ‏use a non-null assertion, as it is simpler and shows the intent more clearly
```suggestion
    styleSheet = styleElement.sheet!
```

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -154,3 +154,142 @@ describe('initFrustrationObserver', () => {
...
+        const groupingRule = (styleSheet.cssRules[1] as CSSGroupingRule).cssRules[0] as CSSGroupingRule
+        // When
+        stopStyleSheetObserver = initStyleSheetObserver(styleSheetCallbackSpy)
+        groupingRule.insertRule(styleRule, 1)
+        // Then
+        const styleSheetRule = styleSheetCallbackSpy.calls.first().args[0]
+        expect(styleSheetRule.id).toBeDefined()
+        expect(styleSheetRule.removes).toBeUndefined()
+        expect(styleSheetRule.adds?.length).toEqual(1)
+        expect(styleSheetRule.adds?.pop()?.index).toEqual([1, 0, 1])
```
BenoitZugmeyer, line 247: **💬 suggestion:** ‏no need to mutate anything

```suggestion
        expect(styleSheetRule.adds?.[0]?.index).toEqual([1, 0, 1])
```

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -154,3 +154,142 @@ describe('initFrustrationObserver', () => {
...
+        styleSheet.insertRule('@media cond-2 { @media cond-1 { .nest-1 { color: #ccc } } }')
+        // Given
+        const parentRule = styleSheet.cssRules[0] as CSSGroupingRule
+        const groupingRule = parentRule.cssRules[0] as CSSGroupingRule
+        parentRule.deleteRule(0)
+        // When
+        stopStyleSheetObserver = initStyleSheetObserver(styleSheetCallbackSpy)
+        groupingRule.insertRule(styleRule, 0)
+        // Then
+        expect(styleSheetCallbackSpy.calls.all().length).toEqual(0)
```
BenoitZugmeyer, line 260: **💬 suggestion:** ‏
```suggestion
        expect(styleSheetCallbackSpy).not.toHaveBeenCalled()
```

packages/rum/src/domain/record/utils.ts
```diff
@@ -28,3 +29,64 @@ export function assembleIncrementalSnapshot<Data extends BrowserIncrementalData>
     timestamp: timeStampNow(),
   }
 }
+
+export function checkStyleSheetAndCallback(styleSheet: CSSStyleSheet | null, callback: (id: number) => void): void {
```
BenoitZugmeyer, line 33: **💬 suggestion:** ‏move this function in `initStyleSheetObserver` as it is rather specific to its implementation, and there is no tests associated

packages/rum/src/domain/record/utils.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+describe('getPathToNestedCSSRule', () => {
+  let styleSheet: CSSStyleSheet
+  let styleElement: HTMLStyleElement
+  beforeEach(() => {
+    if (isIE()) {
+      pending('IE not supported')
+    }
+    styleElement = document.createElement('style')
+    document.head.appendChild(styleElement)
+    styleSheet = <CSSStyleSheet>styleElement.sheet
```
BenoitZugmeyer, line 27: **💬 suggestion:** ‏
```suggestion
    styleSheet = styleElement.sheet!
```

mbeaudru (approved):

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -154,3 +155,150 @@ describe('initFrustrationObserver', () => {
...
+        const index = 0
+        // When
+        stopStyleSheetObserver = initStyleSheetObserver(styleSheetCallbackSpy)
+        styleSheet.deleteRule(index)
+        // Then
+        const styleSheetRule = styleSheetCallbackSpy.calls.first().args[0]
+        expect(styleSheetRule.id).toBeDefined()
+        expect(styleSheetRule.adds).toBeUndefined()
+        expect(styleSheetRule.removes?.length).toEqual(1)
+        expect(styleSheetRule.removes?.[0]).toEqual({ index })
```
bcaudan, line 228: **💭 thought:** ‏Not super fan of adding `given` / `when` / `then` comments, IMO you can separate the setup / action / expectation phases of the test with spacing alone but not a big deal.
However, I am a bit bothered by the instructions that are before the `given` comment, what is the intent there?

packages/rum/src/domain/record/observers.ts
```diff
@@ -348,33 +350,61 @@ export function initInputObserver(cb: InputCallback, defaultPrivacyLevel: Defaul
...
-    before(index) {
-      if (hasSerializedNode(this.ownerNode!)) {
-        cb({
-          id: getSerializedNodeId(this.ownerNode),
-          removes: [{ index }],
-        })
-      }
-    },
-  })
+export function initStyleSheetObserver(cb: StyleSheetRuleCallback): ListenerHandler {
```
bcaudan, line 353: **💭 thought:** ‏since we now observe other APIs than stylesheet, what about renaming this method to something more general, like `initCSSObservers`?

test/e2e/scenario/recorder/recorder.scenario.ts
```diff
@@ -648,6 +649,52 @@ describe('recorder', () => {
...
+              .baz {
+              }
+            }
+          </style>
+        `
+      )
+      .run(async ({ serverEvents }) => {
+        await browserExecute(() => {
+          const firstGroupingRule = document.styleSheets[0].cssRules[0] as CSSMediaRule
+          const secondGroupingRule = document.styleSheets[0].cssRules[1] as GroupingCSSRule
```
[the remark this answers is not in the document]
N-Boutaib: No specific reason tbh. To make it more general, I would cast the first one to `GroupingCSSRule` instead of `CSSMediaRule`

packages/rum/src/domain/record/utils.ts
```diff
@@ -28,3 +28,26 @@ export function assembleIncrementalSnapshot<Data extends BrowserIncrementalData>
     timestamp: timeStampNow(),
   }
 }
+
+export type GroupingCSSRule = CSSGroupingRule | CSSSupportsRule | CSSMediaRule
```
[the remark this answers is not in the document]
N-Boutaib: At a given point, we do cast a `CSSRule` to `GroupingCSSRule` in order to build the path to that rule. Casting directly to `CSSGroupingRule` would cause an issue in browser that don't support `CSSGroupingRule`

packages/rum/src/domain/record/observers.spec.ts
N-Boutaib: I test adding and removing CSS Rules at a given index. The steps before `// Given` is a setup of the style sheet.

test/e2e/scenario/recorder/recorder.scenario.ts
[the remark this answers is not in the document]
N-Boutaib: It would make sense to not only use `@media` rules, but also `@supports` rules

liywjl (approved):

packages/rum/src/domain/record/utils.ts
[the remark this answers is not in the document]
bcaudan: Casting is performed by ts compiler not by browsers, so we should not have this kind of issue.

packages/rum/src/domain/record/observers.ts
```diff
@@ -91,7 +93,7 @@ interface ObserverParam {
   visualViewportResizeCb: VisualViewportResizeCallback
   inputCb: InputCallback
   mediaInteractionCb: MediaInteractionCallback
-  styleSheetRuleCb: StyleSheetRuleCallback
+  cssRulesCb: StyleSheetRuleCallback
```
bcaudan, line 96: **💭 thought:** ‏hum, not great if we end up mixing `css` and `stylesheet` namings 🤔
wdyt?

packages/rum/src/domain/record/observers.spec.ts
bcaudan: I saw that you removed the comments but don't hesitate if you want to continue the discussion on that.

packages/rum/src/domain/record/utils.ts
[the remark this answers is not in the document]
N-Boutaib: My bad, the instruction that was failing is `instanceof`, I mixed two stuff. I'll use `CSSGroupingRule` instead

packages/rum/src/domain/record/observers.ts
bcaudan: But you just moved the problem:
https://github.com/DataDog/browser-sdk/blob/7c141506176099d07c003142a2f99d73f1976993/packages/rum/src/domain/record/observers.ts#L69

N-Boutaib: `StyleSheetRule` is the type of the record we create. I don't know if we can rename the whole typing of this record 🤔

bcaudan (approved):
LGTM

BenoitZugmeyer (approved):

PR #1700, opened 2022-08-19
⚗️[RUMF-1351] experiment request retry strategy

bcaudan:
## Motivation

We want to improve our data transfer to:

- limit our impact on end user bandwidth
- not loose data in case of short intake unavailability

## Changes

Introduce new request strategy allowing us to:
- limit the number of concurrent requests to the intake -> max 32 by product
- limit the total ongoing bytes of requests to the intake -> max 80kb by product
- retry requests if the intake is momentarily unavailable -> exponential backoff strategy 1s to 256s
- limit the size of queued requests due to throttling or failure -> max 3Mb by product

Behind `retry` flag, use this new strategy for logs/rum for non exit requests.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -1,6 +1,8 @@
 import type { EndpointBuilder } from '../domain/configuration'
 import { addTelemetryError } from '../domain/telemetry'
 import { monitor } from '../tools/monitor'
+import { isExperimentalFeatureEnabled } from '../domain/configuration'
+import { newRetryState, sendWithRetryStrategy } from './sendWithRetryStrategy'
 
 /**
  * Use POST request without content type to:
@@ -22,9 +24,17 @@ export interface Payload {
 }
 
 export function createHttpRequest(endpointBuilder: EndpointBuilder, bytesLimit: number) {
+  const retryState = newRetryState()
+  const sendStrategyForRetry = (payload: Payload, onResponse: (r: HttpResponse) => void) =>
+    fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload, onResponse)
+
   return {
     send: (payload: Payload) => {
-      fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload)
+      if (!isExperimentalFeatureEnabled('retry')) {
+        fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload)
+      } else {
+        sendWithRetryStrategy(payload, retryState, sendStrategyForRetry)
+      }
     },
     /**
      * Since fetch keepalive behaves like regular fetch on Firefox,
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -19,55 +19,36 @@ export interface Payload {
 }
 
 export function createHttpRequest(endpointBuilder: EndpointBuilder, bytesLimit: number) {
-  function sendBeaconStrategy({ data, bytesCount }: Payload) {
-    const url = endpointBuilder.build()
-    const canUseBeacon = !!navigator.sendBeacon && bytesCount < bytesLimit
-    if (canUseBeacon) {
-      try {
-        const isQueued = navigator.sendBeacon(url, data)
-
-        if (isQueued) {
-          return
-        }
-      } catch (e) {
-        reportBeaconError(e)
-      }
-    }
-
-    sendXHR(url, data)
-  }
-
-  function fetchKeepAliveStrategy({ data, bytesCount }: Payload) {
-    const url = endpointBuilder.build()
-    const canUseKeepAlive = window.Request && 'keepalive' in new Request('') && bytesCount < bytesLimit
-    if (canUseKeepAlive) {
-      fetch(url, { method: 'POST', body: data, keepalive: true }).catch(
-        monitor(() => {
-          // failed to queue the request
-          sendXHR(url, data)
-        })
-      )
-    } else {
-      sendXHR(url, data)
-    }
-  }
-
-  function sendXHR(url: string, data: Payload['data']) {
-    const request = new XMLHttpRequest()
-    request.open('POST', url, true)
-    request.send(data)
-  }
-
   return {
     send: (payload: Payload) => {
-      fetchKeepAliveStrategy(payload)
+      fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload)
     },
     /**
      * Since fetch keepalive behaves like regular fetch on Firefox,
      * keep using sendBeaconStrategy on exit
      */
-    sendOnExit: sendBeaconStrategy,
+    sendOnExit: (payload: Payload) => {
+      sendBeaconStrategy(endpointBuilder, bytesLimit, payload)
+    },
   }
+}
+
+function sendBeaconStrategy(endpointBuilder: EndpointBuilder, bytesLimit: number, { data, bytesCount }: Payload) {
+  const url = endpointBuilder.build()
+  const canUseBeacon = !!navigator.sendBeacon && bytesCount < bytesLimit
+  if (canUseBeacon) {
+    try {
+      const isQueued = navigator.sendBeacon(url, data)
+
+      if (isQueued) {
+        return
+      }
+    } catch (e) {
+      reportBeaconError(e)
+    }
+  }
+
+  sendXHR(url, data)
 }
 
 let hasReportedBeaconError = false
@@ -77,4 +58,25 @@ function reportBeaconError(e: unknown) {
     hasReportedBeaconError = true
     addTelemetryError(e)
   }
+}
+
+function fetchKeepAliveStrategy(endpointBuilder: EndpointBuilder, bytesLimit: number, { data, bytesCount }: Payload) {
+  const url = endpointBuilder.build()
+  const canUseKeepAlive = window.Request && 'keepalive' in new Request('') && bytesCount < bytesLimit
+  if (canUseKeepAlive) {
+    fetch(url, { method: 'POST', body: data, keepalive: true }).catch(
+      monitor(() => {
+        // failed to queue the request
+        sendXHR(url, data)
+      })
+    )
+  } else {
+    sendXHR(url, data)
+  }
+}
+
+function sendXHR(url: string, data: Payload['data']) {
+  const request = new XMLHttpRequest()
+  request.open('POST', url, true)
+  request.send(data)
 }
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -12,6 +12,9 @@ import { monitor } from '../tools/monitor'
  */
 
 export type HttpRequest = ReturnType<typeof createHttpRequest>
+export interface HttpResponse {
+  status: number
+}
 
 export interface Payload {
   data: string | FormData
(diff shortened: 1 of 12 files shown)


packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function wasRequestSuccessful(response: HttpResponse) {
+  return response.status < 500
+}
+
+export function newRetryState(): RetryState {
+  return {
+    isIntakeAvailable: true,
+    currentBackoffTime: ONE_SECOND,
```
BenoitZugmeyer, line 97: **🥜 nitpick:** ‏use a constant like `INITIAL_BACKOFF_TIME`

packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -0,0 +1,153 @@
...
+export function sendWithRetryStrategy(payload: Payload, state: RetryState, sendStrategy: SendStrategy) {
+  if (state.isIntakeAvailable && state.bandwidthMonitor.canHandle(payload)) {
+    send(payload, state, sendStrategy, {
+      onSuccess: () => sendNextPayload(state, sendStrategy),
+      onFailure: () => {
+        state.queuedPayloads.enqueue(payload)
+        scheduleRetry(state, sendStrategy)
+      },
+    })
+    sendNextPayload(state, sendStrategy)
```
BenoitZugmeyer, line 29: **💬 suggestion:** Try to send the queued payloads (if any) before sending the new payload.

Currently, if a payload bigger than `MAX_ONGOING_BYTES_COUNT` is added to the queue, it wont ever be sent (assuming the intake always succeeds), because `bandwidthMonitor.ongoingRequestCount` will always be greater than `1` (since we just started sending a request) and the payload size always bigger than `MAX_ONGOING_BYTES_COUNT`, so `canHandle` will always return false.

bcaudan: > it wont ever be sent (assuming the intake always succeeds), because bandwidthMonitor.ongoingRequestCount will always be greater than 1 (since we just started sending a request) and the payload size always bigger than MAX_ONGOING_BYTES_COUNT, so canHandle will always return false.


not exactly, we will wait for no more ongoing requests to send it, cf: https://github.com/DataDog/browser-sdk/blob/82c719f00de87b1c89e5ffeed18c13fbd2ca6eac/packages/core/src/transport/sendWithRetryStrategy.ts#L137-L143

BenoitZugmeyer: Hum I missed the `sendNextPayload` call in `onSuccess`. But I still think we should first try to send payloads from the queue before sending the new one, so the "first in first out" rule is respected. It could be useful in the Session Replay case where segments follow each other: if we miss an intermediary segment, the replay won't work, but if we miss the last one, it will only be truncated.

packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -0,0 +1,153 @@
...
+export const MAX_ONGOING_REQUESTS = 32
+export const MAX_QUEUE_BYTES_COUNT = 3 * ONE_MEGA_BYTE
+export const MAX_BACKOFF_TIME = 256 * ONE_SECOND
+
+export interface RetryState {
+  isIntakeAvailable: boolean
+  currentBackoffTime: number
+  bandwidthMonitor: ReturnType<typeof newBandwidthMonitor>
+  queuedPayloads: ReturnType<typeof newPayloadQueue>
+}
```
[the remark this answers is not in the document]
bcaudan: hum, if we want to go this way, I would probably differentiate status this way:
 `INTAKE_UP` | `INTAKE_ERROR_DETECTED` | `INTAKE_DOWN` 
 in order to not switch between `ACCEPTING_REQUESTS` and `QUEUING_REQUESTS` every time we reach a bandwidth limit.
We also need `queuedPayloads` and `bandwidthMonitor` to be shared among all `httpRequestInstance.send`.
So maybe something like:

```ts
interface RetryContext {
  queuedPayloads: ...,
  currentBackoffTime: ...,
  state: RetryState
}

type RetryState = {
  status: INTAKE_UP,
} | {
  status: INTAKE_ERROR_DETECTED,
} | {
  status: INTAKE_DOWN,
  currentBackoffTime: number,
}
```
wdyt?

packages/core/src/transport/sendWithRetryStrategy.ts
bcaudan: Currently, we are maximizing our bandwidth consumption:
- if a new request can be sent, it is sent
- when a request complete, we try to send all possible queued requests than can fit the available bandwidth.
So it indeed may not respect the order when the bandwidth is under constraint.

Enforcing the order indeed make sense for session replay, we missed that during previous reflexion.
I'll have a look.

packages/core/src/transport/httpRequest.ts
```diff
@@ -73,3 +72,36 @@ function reportBeaconError(e: unknown) {
     addTelemetryError(e)
   }
 }
+
+export function fetchKeepAliveStrategy(
```
liywjl, line 76: **❓ question:**if `fetch` `keepalive` just uses [`sendBeacon` under the hood](https://xgwang.me/posts/you-may-not-know-beacon/#fetch-keepalive), why do we not use `sendBeacon` as it has wider support?

bcaudan: cf [previous step of this topic](https://github.com/DataDog/browser-sdk/pull/1701)

> In order to improve our data transfer, send request with fetch keepalive and fallback in order to later be able to access to intake response.

packages/core/src/transport/sendWithRetryStrategy.ts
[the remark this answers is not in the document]
BenoitZugmeyer: I think that if we need the queue and currentBackoffTime at any time, having separate states is less interesting. Let's keep it your way 🙂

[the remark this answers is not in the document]
bcaudan: `currentBackoffTime` is only needed when intake is down and having named state can ease the understanding.
I'll experiment a bit and see what it could look like.

packages/core/src/transport/sendWithRetryStrategy.ts
BenoitZugmeyer: Maybe it could be configured per intake? We should make sure the queue limit is big enough to avoid reaching it in "normal usage". This limit is only there to prevent memleaks (ex: when a page is continuously sending data to a failing intake for a long time), right?

bcaudan: > We should make sure the queue limit is big enough to avoid reaching it in "normal usage". This limit is only there to prevent memleaks (ex: when a page is continuously sending data to a failing intake for a long time), right?

yes, the reasoning for removing oldest elements was "if they are very old, they will be ignored by the backend anyway" 

> Maybe it could be configured per intake?

given the original reasoning I don't think it worth the trouble

BenoitZugmeyer: How about adding some telemetry to see if it happens often, and raise the limit if that's the case?

packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -0,0 +1,154 @@
...
+      onSuccess()
+    } else {
+      state.isIntakeAvailable = false
+      onFailure()
+    }
+  })
+}
+
+function wasRequestSuccessful(response: HttpResponse) {
+  return response.status < 500
```
BenoitZugmeyer, line 92: **💬 suggestion:** ‏retry payloads when internet is down:
```suggestion
  return response.status === 0 || response.status < 500
```

packages/core/src/transport/sendWithRetryStrategy.ts
bcaudan: is it still related to choosing between discarding oldest or newest queued payloads?
or is it another point?

packages/core/src/transport/sendWithRetryStrategy.ts
bcaudan: Indeed, I've seen network status check listed as improvement and discarded the retry when request fail due to network down :facepal

packages/core/src/transport/sendWithRetryStrategy.ts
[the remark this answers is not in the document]
bcaudan: splitting into context and state added quite some complexity without a lot of benefit, I've promoted the boolean into a status, it could make things a bit clearer.

packages/core/src/transport/sendWithRetryStrategy.ts
BenoitZugmeyer: It is to evaluate how much it matters. If the limit is reached very occasionnaly, let's keep discarding oldest payloads. Else, we should consider a different approach.

bcaudan: I already made the change, it seems more consistent if we want to favor request order.
I'll add some monitoring, let me know if you think we should get back to discarding oldest payloads.

packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -79,22 +80,32 @@ function send(
...
   })
 }
 
+function retryQueuedPayloads(state: RetryState, sendStrategy: SendStrategy) {
+  const pendingPayloads = []
+  while (state.queuedPayloads.size() > 0) {
+    pendingPayloads.push(state.queuedPayloads.dequeue()!)
+  }
+  pendingPayloads.map((payload) => sendWithRetryStrategy(payload, state, sendStrategy))
+}
```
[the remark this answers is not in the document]
liywjl: `retryQueuedPayloads` is a lot simpler to understand IMO, and given the complexity already present in this file, I think it makes sense to keep it simple. I think it would be really interesting to exploring finite state machines here to manage the complexity. WDYT?

Also, I am curious to know what you think about maybe using [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)'`transferables` to avoid having to  run `state.queuedPayloads.dequeue`? There might be a performance gain to have here as data is transferred rather than cloned? Plus it would avoid having the double iteration issue you pointed out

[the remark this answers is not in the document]
bcaudan: > I think it would be really interesting to exploring finite state machines here to manage the complexity. WDYT?

IIRC, we have already discussed this topic, decided to investigate it but outside of the current project, right?

> Also, I am curious to know what you think about maybe using [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)'transferables to avoid having to run state.queuedPayloads.dequeue? There might be a performance gain to have here as data is transferred rather than cloned? Plus it would avoid having the double iteration issue you pointed out

I don't know this API, not sure that this point would need this capability.
It gave me another idea though: [e214c57](https://github.com/DataDog/browser-sdk/pull/1700/commits/e214c571ff37cc73518b553a027f50b7a51fa06f)
Let me know if you have any concerns.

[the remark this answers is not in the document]
liywjl: > IIRC, we have already discussed this topic, decided to investigate it but outside of the current project, right?

Yes, agreed

> It gave me another idea though: [e214c57](https://github.com/DataDog/browser-sdk/pull/1700/commits/e214c571ff37cc73518b553a027f50b7a51fa06f)

This is a much simpler solution! yes happy to go with that

liywjl (approved):

packages/core/src/transport/sendWithRetryStrategy.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      expect(state.bandwidthMonitor.ongoingRequestCount).toBe(1)
+      sendStub.respondWith(1, { status: 500 })
+
+      clock.tick(2 * INITIAL_BACKOFF_TIME)
+      expect(state.bandwidthMonitor.ongoingRequestCount).toBe(1)
+      sendStub.respondWith(2, { status: 500 })
+
+      clock.tick(4 * INITIAL_BACKOFF_TIME)
+      expect(state.bandwidthMonitor.ongoingRequestCount).toBe(1)
+      sendStub.respondWith(3, { status: 500 })
```
BenoitZugmeyer, line 202: **🥜 nitpick:** Responding to the request does not look useful if we don't assert anything after that

packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -0,0 +1,174 @@
...
+import { ONE_KIBI_BYTE, ONE_MEBI_BYTE, ONE_SECOND } from '../tools/utils'
+import type { Payload, HttpResponse } from './httpRequest'
+
+export const MAX_ONGOING_BYTES_COUNT = 80 * ONE_KIBI_BYTE
+export const MAX_ONGOING_REQUESTS = 32
+export const MAX_QUEUE_BYTES_COUNT = 3 * ONE_MEBI_BYTE
+export const MAX_BACKOFF_TIME = 256 * ONE_SECOND
+export const INITIAL_BACKOFF_TIME = ONE_SECOND
+
+enum TransportStatus {
```
BenoitZugmeyer, line 12: **💬 suggestion:** ‏we meant to enforce using `const enum` to help minification when possible, but the ESLint configuration is incorrect and fails to detect this. I fixed it in https://github.com/DataDog/browser-sdk/pull/1709

```suggestion
const enum TransportStatus {
```

BenoitZugmeyer (approved):

PR #1706, opened 2022-08-26
👷 e2e bs: improve stability with more recent browsers

bcaudan:
## Motivation

Failure to start browser and idle timeout seems quite frequent, especially with Safari.

## Changes

Bump versions for all browsers/OS

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


liywjl (approved):

BenoitZugmeyer (commented):
This will prevent us from finding incompatibility issues in older browsers. Does that mean we drop the support for those browsers? If not, how can we check that the SDK works for them?

test/browsers.conf.js
```diff
@@ -4,30 +4,30 @@ module.exports = [
...
-    version: '86',
+    version: '100.0',
     os: 'Windows',
-    osVersion: '10',
+    osVersion: '11',
   },
   {
     sessionName: 'Firefox',
     name: 'Firefox',
-    version: '78.0',
```
BenoitZugmeyer, line 14: **💬 suggestion:** ‏Use `91.13.0` for Firefox: it's the oldest version still supported (ESR) https://en.wikipedia.org/wiki/Firefox_version_history#Current_supported_official_releases

BenoitZugmeyer (approved):

PR #1709, opened 2022-08-30
👷 fix ESLint rule on non-const enum

BenoitZugmeyer:
## Motivation

The `no-restricted-syntax` related to enums introduced in #1364 is overridden by other restricted syntaxes on transpiled files introduced by #1347, so it's not actually applied to source files.

## Changes

Because the enum limitation is also about transpilation optimization, simply move it to the special casing.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


liywjl (approved):

bcaudan (approved):
Good catch

PR #1712, opened 2022-08-31
🐛 [RUMF-1333] fix keepalive support check

bcaudan:
## Motivation

Request can throw if the URL contains credentials

## Changes

- use endpoint url instead of current url
- wrap Request check in try/catch

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -34,7 +34,7 @@ export function createHttpRequest(endpointBuilder: EndpointBuilder, bytesLimit: 
 
   function fetchKeepAliveStrategy(data: string | FormData, bytesCount: number) {
     const url = endpointBuilder.build()
-    const canUseKeepAlive = window.Request && 'keepalive' in new Request('') && bytesCount < bytesLimit
+    const canUseKeepAlive = isKeepAliveSupported(url) && bytesCount < bytesLimit
     if (canUseKeepAlive) {
       fetch(url, { method: 'POST', body: data, keepalive: true }).catch(
         monitor(() => {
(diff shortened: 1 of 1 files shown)


packages/core/src/transport/httpRequest.ts
```diff
@@ -47,6 +47,18 @@ export function createHttpRequest(endpointBuilder: EndpointBuilder, bytesLimit:
     }
   }
 
+  /**
+   * Request can throw
+   * cf https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#errors
+   */
+  function isKeepAliveSupported(url: string) {
+    try {
+      return window.Request && 'keepalive' in new Request(url)
```
BenoitZugmeyer, line 56: **💬 suggestion:** What about using a hardcoded URL instead? `http://a`

packages/core/src/transport/httpRequest.ts
```diff
@@ -47,6 +47,18 @@ export function createHttpRequest(endpointBuilder: EndpointBuilder, bytesLimit:
     }
   }
 
+  /**
+   * Request can throw
+   * cf https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#errors
+   */
```
BenoitZugmeyer, line 53: **💬 suggestion:** ‏this comment is not a function documentation, move it inside the function

packages/core/src/transport/httpRequest.ts
bcaudan: why not

BenoitZugmeyer (approved):

packages/core/src/transport/httpRequest.ts
```diff
@@ -47,6 +47,15 @@ export function createHttpRequest(endpointBuilder: EndpointBuilder, bytesLimit:
...
 
+  function isKeepAliveSupported() {
+    // Request can throw, cf https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#errors
+    try {
+      return window.Request && 'keepalive' in new Request('http://a')
+    } catch {
+      return false
+    }
+  }
+
```
liywjl, line 58: **💬 suggestion:** using `Request('http://a')` means this block is no longer needed. Might be worth moving back to the original pattern and just adding the comment?

bcaudan: Since a failure caused by this check can be quite impacting, I'd be in favor of keeping this try/catch just for undocumented / future throw or patched API.

liywjl (approved):

PR #1717, opened 2022-09-02
✨ [RUMF-1371] Collect view time to first byte

bcaudan:
## Motivation

Allow to track the [time to first byte](https://developer.mozilla.org/en-US/docs/Glossary/time_to_first_byte) of the resource corresponding to the document loading at the view level.

## Changes

Add `view.first_byte` from response start [navigation timing](https://www.w3.org/TR/navigation-timing/#processing-model).

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackInitialViewTimings.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackInitialViewTimings.ts
@@ -16,6 +16,7 @@ export const TIMING_MAXIMUM_DELAY = 10 * ONE_MINUTE
 
 export interface Timings {
   firstContentfulPaint?: Duration
+  firstByte?: Duration
   domInteractive?: Duration
   domContentLoaded?: Duration
   domComplete?: Duration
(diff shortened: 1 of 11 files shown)


BenoitZugmeyer (approved):

packages/rum-core/src/domain/rumEventsCollection/view/trackInitialViewTimings.ts
```diff
@@ -67,6 +68,7 @@ export function trackNavigationTimings(lifeCycle: LifeCycle, callback: (timings:
           domContentLoaded: entry.domContentLoadedEventEnd,
           domInteractive: entry.domInteractive,
           loadEvent: entry.loadEventEnd,
+          firstByte: entry.responseStart,
```
liywjl, line 71: **💭 thought:** isn't [`TTFB = responseStart - navigationStart`](https://developer.mozilla.org/en-US/docs/Glossary/time_to_first_byte) and not simply `responseStart`?

BenoitZugmeyer: Right, but "`navigationStart`" is `0` 🙂  (or would be 0 if it was set in the PerformanceTiming entry)

liywjl: I suspected that might be the case 😅 was just double checking

liywjl (approved):

PR #1719, opened 2022-09-02
✨ [RUMF-1286] test for expected features before starting recording

BenoitZugmeyer:
## Motivation

Don't crash with telemetry errors when start recording on unsupported browsers

## Changes

Do some light feature test before returning the actual recording API

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum/src/boot/recorderApi.ts
```diff
@@ -155,3 +155,7 @@ export function makeRecorderApi(
     isRecording: () => state.status === RecorderStatus.Started,
   }
 }
+
+function isBrowserSupported() {
+  return typeof Array.from === 'function' && typeof CSSSupportsRule === 'function'
+}
```
N-Boutaib, line 161: **Nitpick**: Would you please add a comment to explain the reason behind identifying browser support from the support of `Array.from` and `CSSSupportsRule` ?

BenoitZugmeyer: Done

liywjl (approved):

PR #1725, opened 2022-09-08
🐛 [RUMF-1369] Exclude error message from stacktrace parsing

liywjl:
## Motivation

We accidentally parse error messages that looked similar to stackTrace:

Example:
```
HttpError: 400 Bad Request, see context.apiResponse for more details
    at s (spa.min.js:2:2418592)
```
would output
```
    [{
      args: [],
      column: undefined,
      func: '?',
      line: undefined,
      url: 'HttpError: 400 Bad Request, see context.apiResponse for more details',
    }, {
      args: [],
      column: 5,
      func: 's',
      line: 3,
      url: 'spa.min.js',
    }]
```

`'HttpError: 400 Bad Request, see context.apiResponse for more details'` is not a valid URL

## Changes

Get the `toString` of the error (casting error as a `String`) and remove it from the beginning of the stack trace if present

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
+++ b/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
@@ -502,7 +502,7 @@ Error: foo
   it('should not include error message into stacktrace ', () => {
     const stackFrames = computeStackTrace(new Error('bar@http://path/to/file.js:1:1'))
 
-    expect(stackFrames.stack[0].url).not.toBe('http://path/to/file.js')
+    expect(stackFrames.stack[0]?.url).not.toBe('http://path/to/file.js')
   })
 
   it('should parse Chrome anonymous function errors', () => {
--- a/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
+++ b/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
@@ -499,6 +499,38 @@ Error: foo
     })
   })
 
+  it('should ignore custom error message', () => {
+    const stack = `HttpError: 400 Bad Request, see context.apiResponse for more details
+    at s (spa.min.js:3:5)
+    at b (spa.min.js:6:8)
+    at spa.min.js:2:1`
+
+    const stackFrames = computeStackTrace({ stack } as Error)
+
+    expect(stackFrames.stack.length).toEqual(3)
+    expect(stackFrames.stack[0]).toEqual({
+      args: [],
+      column: 5,
+      func: 's',
+      line: 3,
+      url: 'spa.min.js',
+    })
+    expect(stackFrames.stack[1]).toEqual({
+      args: [],
+      column: 8,
+      func: 'b',
+      line: 6,
+      url: 'spa.min.js',
+    })
+    expect(stackFrames.stack[2]).toEqual({
+      args: [],
+      column: 1,
+      func: '?',
+      line: 2,
+      url: 'spa.min.js',
+    })
+  })
+
   it('should parse Chrome anonymous function errors', () => {
     const stack = `Error: RTE Simulation
     at https://datadoghq.com/somefile.js:8489:191
--- a/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
+++ b/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
@@ -499,7 +499,7 @@ Error: foo
     })
   })
 
-  it('should ignore custom error messages on chrome ', () => {
+  it('should ignore custom error messages ', () => {
     const stackFrames = computeStackTrace(new Error('bar@http://path/to/file.js:1:1'))
 
     expect(stackFrames.stack[0].url).not.toBe('http://path/to/file.js')
--- a/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
+++ b/packages/core/src/domain/tracekit/computeStackTrace.spec.ts
@@ -502,13 +502,7 @@ Error: foo
   it('should ignore custom error messages on chrome ', () => {
     const stackFrames = computeStackTrace(new Error('bar@http://path/to/file.js:1:1'))
 
-    expect(stackFrames.stack[0]).not.toEqual({
-      args: [],
-      column: 1,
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/tracekit/computeStackTrace.ts
```diff
@@ -104,9 +104,8 @@ const GECKO_EVAL_RE = /(\S+) line (\d+)(?: > eval line \d+)* > eval/i
 
 function parseGeckoLine(line: string): StackFrame | undefined {
   const parts = GECKO_LINE_RE.exec(line)
-  if (!parts) {
-    return
-  }
+  if (!parts) return
+  if (parts[3] === line && line !== '[native code]') return // overly greedy regex likely means this is a custom error message (exception native code)
```
BenoitZugmeyer, line 108: **💭 thought:** ‏ Ideally, the `parseGeckoLine` should not be used to parse React Native stacks :D anyway.


**💬 suggestion:** ‏Can I suggest an alternative fix? Let's be pragmatical here. The issue seems to be that:
* the error message is included in the `error.stack` property
* and the error message is considered as a stack "line" by our regexes.

Because the error message is often crafted by the application, this can happen unpredictibly. The message could literally be `    at namedFunc0 (http://example.com/js/script.js:10:2)` and we wouldn't be able to distinghuish it from a normal stack line. What we could do instead is, before trying to parse the `error.stack` value, test if it contains the error message and remove it from the stack if that's the case. Something like this:

```ts
let stackProperty = tryToGetString(ex, 'stack')
const exString = String(ex)
if (startsWith(stack, exString) {
  exProperty = stack.slice(exString.length)
}
stackProperty.split('\n').forEach(...)
```

WDYT?

liywjl: `parseGeckoLine` is currently catching a lot of cases unrelated to Gecko parsing errors (`'[native code]'` is just one of the examples). This is due to a greedy regex used (the main culprit is `.*?` just after `\[native)` in the `GECKO_LINE_RE`). 

Renaming or breaking down this part of the regex would be ideal to better understand what is getting matched.

liywjl: This is a great suggestion, I tried out something similar previously (but with different assumptions). 

The issue is that in some cases, the first line and the stack trace are the same, and this solution will inadvertently remove the string.

A lot of this argument is based on unit tests that failed. I suspect that despite it's flaws, `parseGeckoLine` does not regex match these edge cases.

BenoitZugmeyer: > The issue is that in some cases, the first line and the stack trace are the same, and this solution will inadvertently remove the string.

Which case? Do you have something in mind? I suspect that it would not happen very often, right? It would be really particular, hand crafted errors and stacks. And even if it happens, we would still send all the data (message + stack): even if the stack would miss its first line, it would be in the message anyway, so I think this is acceptable.

liywjl: 🤦‍♂️ `stack.slice` should have been `stackProperty.slice`. The solution you suggest actually solves this edge case without breaking anything else. Will update and add tests 

I still think we should maybe revisit `parseGeckoLine` at some point

packages/core/src/domain/tracekit/computeStackTrace.spec.ts
```diff
@@ -499,6 +499,53 @@ Error: foo
...
+  it('should ignore custom error messages on chrome ', () => {
+    const stackFrames = computeStackTrace(new Error('bar@http://path/to/file.js:1:1'))
+
+    expect(stackFrames.stack[0]).not.toEqual({
+      args: [],
+      column: 1,
+      func: 'Error: bar',
+      line: 1,
+      url: 'http://path/to/file.js',
+    })
```
BenoitZugmeyer, line 511: **💬 suggestion:** If anything change in the StackFrame format, this test will succeed without actually testing the expectation. Maybe only test for the URL?

```suggestion
    expect(stackFrames.stack[0].url).not.toBe('http://path/to/file.js')
```

packages/core/src/domain/tracekit/computeStackTrace.spec.ts
```diff
@@ -499,6 +499,53 @@ Error: foo
...
+  })
+
+  it('should ignore messages that come from the `toString` of the error on chrome ', () => {
```
BenoitZugmeyer, line 514: **💬 suggestion:** ‏this test case is about the same as the test case above. Maybe remove it?

packages/core/src/domain/tracekit/computeStackTrace.spec.ts
```diff
@@ -499,6 +499,44 @@ Error: foo
...
+      line: 6,
+      url: 'spa.min.js',
+    })
+    expect(stackFrames.stack[2]).toEqual({
+      args: [],
+      column: 1,
+      func: '?',
+      line: 2,
+      url: 'spa.min.js',
+    })
```
liywjl, line 537:
```suggestion
    const stackFrames = computeStackTrace(new Error('bar@http://path/to/file.js:1:1'))

    expect(stackFrames.stack[0].url).not.toBe('http://path/to/file.js')
```

BenoitZugmeyer (approved):

packages/core/src/domain/tracekit/computeStackTrace.spec.ts
```diff
@@ -499,6 +499,12 @@ Error: foo
     })
   })
 
+  it('should ignore custom error messages ', () => {
```
bcaudan, line 502: **💬 suggestion:** ‏Could you clarify test description?

bcaudan (approved):

packages/core/src/domain/tracekit/computeStackTrace.spec.ts
liywjl, line 502:
```suggestion
  it('should not include error message into stacktrace ', () => {
```

PR #1737, opened 2022-09-15
🐛 [RUMF-1393] don't mask attributes used to create CSS selectors

BenoitZugmeyer:
## Motivation

 Same as #1721, but for other attributes used by CSS selectors.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


liywjl (approved):

packages/rum/src/domain/record/serialize.spec.ts
```diff
@@ -655,33 +661,46 @@ describe('serializeAttribute ', () => {
...
+    it('does not mask the user-supplied programmatic action name attributes when it not a data attribute', () => {
+      const node = document.createElement('div')
+      node.setAttribute('testid', 'foo')
+      expect(
+        serializeAttribute(node, NodePrivacyLevel.MASK, 'testid', {
+          ...DEFAULT_CONFIGURATION,
+          actionNameAttribute: 'testid',
+        })
+      ).toBe('foo')
+    })
```
bcaudan, line 698: **💬 suggestion:** ‏we should probably not use testid anymore in these tests since it now overlaps with stable attributes

packages/rum/src/domain/record/serialize.spec.ts
```diff
@@ -655,33 +661,46 @@ describe('serializeAttribute ', () => {
...
+      node.setAttribute(DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE, 'foo')
+      expect(
+        serializeAttribute(
+          node,
+          NodePrivacyLevel.MASK,
+          DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE,
+          DEFAULT_CONFIGURATION
+        )
+      ).toBe('foo')
+    })
```
bcaudan, line 676: **💭 thought:** ‏I was wondering if we could remove the `DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE` export but it is still used here.
I am assuming it is still interesting to have a test for this case.
I don't mind keeping the export for it since at least we know that we are referencing the same thing.
However, with the same logic, we could maybe also reference `STABLE_ATTRIBUTES` instead of using `'data-test-id'` for the other test:
```ts
it('does not mask other attributes used to generate CSS selectors', () => {
      const node = document.createElement('div')
      node.setAttribute('data-test-id', 'foo')
      expect(serializeAttribute(node, NodePrivacyLevel.MASK, 'data-test-id', DEFAULT_CONFIGURATION)).toBe('foo')
    })
```
wdyt?

bcaudan (approved):

PR #1740, opened 2022-09-16
✨ [RUMF-1353] Add Error.cause property

liywjl:
## Motivation

We want to support the [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property when it is present

## Changes

- use `formatUnknownError` and `processError` to add `causes` to the RawError object
- use `flattenErrorCauses` to flattern nested error causes
- update tests and fixtures

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.spec.ts
```diff
@@ -61,6 +63,33 @@ describe('error collection', () => {
       })
     })
 
+    it('should extract causes from error', () => {
+      const { rawRumEvents } = setupBuilder.build()
+      const error1 = new Error('foo') as unknown as RawError
+      const error2 = new Error('bar') as unknown as RawError
+      const error3 = new Error('biz') as unknown as RawError
```
liywjl, line 70: it isnt ideal, but it avoids us having to use @ts-ignore

packages/core/src/tools/error.ts
```diff
@@ -110,3 +137,20 @@ export function createHandlingStack(): string {
 
   return formattedStack!
 }
+
+export function flattenErrorCauses(error: RawError, parentSource: ErrorSource): ErrorCause[] {
```
BenoitZugmeyer, line 141: **💬 suggestion:** ‏this function should not take a `RawError` as argument. It should take an `Error` or `unknown` argument and check whether it has a `cause` property.

packages/core/src/tools/error.ts
```diff
@@ -110,3 +137,20 @@ export function createHandlingStack(): string {
...
+  let count = 0
+  let currentError = error
+  const causes: Array<RawError | ErrorCause> = []
+  while (currentError?.cause && currentError.cause && count < 10) {
+    causes.push({
+      message: currentError.cause.message,
+      source: currentError.cause.source ?? parentSource,
+      type: currentError.cause?.type,
+      stack: currentError.cause?.stack,
```
BenoitZugmeyer, line 150: **💬 suggestion:** you should "normalize" the stack by parsing then formating it again. Maybe refactor a bit to use the same logic as `formatUnknownError`?

packages/core/src/tools/error.ts
```diff
@@ -110,3 +137,20 @@ export function createHandlingStack(): string {
 
   return formattedStack!
 }
+
+export function flattenErrorCauses(error: RawError, parentSource: ErrorSource): ErrorCause[] {
+  let count = 0
+  let currentError = error
+  const causes: Array<RawError | ErrorCause> = []
+  while (currentError?.cause && currentError.cause && count < 10) {
```
BenoitZugmeyer, line 145: **💬 suggestion:** ‏replace `count < 10` with `causes.length < 10`

packages/core/src/tools/error.ts
liywjl: the schema asks for `stack` to be of type `string`. But ideally here we should have `stack` as type `StackTrace`. Should I update the schema?

packages/core/src/tools/error.spec.ts
```diff
@@ -107,3 +157,18 @@ describe('createHandlingStack', () => {
   at userCallOne @ (.*)`)
   })
 })
+
+describe('flattenErrorCauses', () => {
+  it('should return empty array if  no cause found', () => {
```
bcaudan, line 162:
```suggestion
  it('should return empty array if no cause found', () => {
```

packages/core/src/tools/error.spec.ts
```diff
@@ -107,3 +157,18 @@ describe('createHandlingStack', () => {
...
+    const error = new Error('foo') as ErrorWithCause
+    const errorCauses = flattenErrorCauses({ errorObject: error, parentSource: ErrorSource.LOGGER })
+    expect(errorCauses.length).toEqual(0)
+  })
+
+  it('should only return the first 10 errors if nested chain is longer', () => {
+    const error = new Error('foo') as ErrorWithCause
+    error.cause = error
```
bcaudan, line 170: Nice one!

packages/core/src/tools/error.ts
```diff
@@ -110,3 +141,21 @@ export function createHandlingStack(): string {
 
   return formattedStack!
 }
+
+type FormatErrorCausesParams = { errorObject: ErrorWithCause; parentSource: ErrorSource }
+
+export function flattenErrorCauses({ errorObject, parentSource }: FormatErrorCausesParams): RawErrorCause[] {
+  let currentError = errorObject
+  const causes: RawErrorCause[] = []
+  while (currentError?.cause && currentError.cause && causes.length < 10) {
```
bcaudan, line 150: **💬 suggestion:** ‏no?
```suggestion
  while (currentError?.cause && causes.length < 10) {
```

packages/core/src/tools/error.ts
```diff
@@ -110,3 +141,21 @@ export function createHandlingStack(): string {
...
+    const stackTrace = currentError.cause instanceof Error ? computeStackTrace(currentError.cause) : undefined
+    causes.push({
+      message: currentError.cause.message,
+      source: parentSource,
+      type: stackTrace?.name,
+      stack: stackTrace && toStackTraceString(stackTrace),
+    })
+    currentError = currentError.cause
```
bcaudan, line 158: **💭 thought:** ‏we check if `cause` is an instance of `Error` but the type says it is an `Error`.
So we then use `cause.message` and pass the `cause` as the next error regardless of it is really an `Error` or not.
Should we handle things differently if `cause` is not an `Error`?
We should probably at least have a test for it.

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -47,7 +48,7 @@ export interface PerformanceResourceDetailsElement {
   start: ServerDuration
 }
 
-export interface RawRumErrorEvent {
```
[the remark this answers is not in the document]
liywjl: I imported `RawRumErrorEvent` as a type to use. TS was complaining it wasn't a type. In the end I didn't end up using it so have reverted

packages/core/src/tools/error.ts
liywjl: I have lifted the `stackTrace` logic further up the chain and have added some logic to handle if `cause` is not of type error

liywjl: I am making the assumption that we should stop recursion if the type isnt error. IDK if this is the right assumption to make however

bcaudan: > I am making the assumption that we should stop recursion if the type isnt error. IDK if this is the right assumption to make however

IMO that seems OK. We could always tweak the logic if a customer has another valid case.

packages/core/src/tools/error.spec.ts
```diff
@@ -50,26 +65,94 @@ describe('formatUnknownError', () => {
       stack: [],
     }
 
-    const formatted = formatUnknownError(stack, undefined, 'Uncaught')
+    const formatted = computeRawError({
+      stackTrace: stack,
+      error: undefined,
+      startClocks: clocksNow(),
+      nonErrorPrefix: 'Uncaught',
+      source: 'custom',
```
bcaudan, line 73: **🥜 nitpick:**
```suggestion
      source: ErrorSource.CUSTOM
```
and following occurrences

packages/core/src/tools/error.spec.ts
```diff
@@ -33,7 +42,13 @@ describe('formatUnknownError', () => {
...
     }
 
-    const formatted = formatUnknownError(stack, undefined, 'Uncaught')
+    const formatted = computeRawError({
+      stackTrace: stack,
+      error: undefined,
+      startClocks: clocksNow(),
+      nonErrorPrefix: 'Uncaught',
+      source: ErrorSource.CUSTOM,
+    })
```
bcaudan, line 51: **💬 suggestion:** ‏In order to avoid to clutter the tests with non relevant parameters, you could extract some default raw errors params.
It would make clearer for each case what are the important inputs of the test subject.
```suggestion
    const formatted = computeRawError({
      ...DEFAULT_RAW_ERROR_PARMS,
      stackTrace: stack,
      error: undefined,
    })
```

packages/core/src/tools/error.spec.ts
```diff
@@ -50,26 +65,94 @@ describe('formatUnknownError', () => {
       stack: [],
     }
 
-    const formatted = formatUnknownError(stack, undefined, 'Uncaught')
+    const formatted = computeRawError({
+      stackTrace: stack,
```
bcaudan, line 69: **🥜 nitpick:** ‏renaming those `stack` variables could be nice as well
```suggestion
      stackTrace,
```

packages/core/src/tools/error.spec.ts
```diff
@@ -50,26 +65,94 @@ describe('formatUnknownError', () => {
...
+      error,
+      startClocks: clocksNow(),
+      nonErrorPrefix: 'Uncaught',
+      source: 'custom',
+    })
 
     expect(formatted.message).toEqual('Uncaught {"foo":"bar"}')
   })
+
+  it('should return handling set', () => {
```
bcaudan, line 107:
```suggestion
  it('should set handling according to given parameter', () => {
```

packages/core/src/tools/error.spec.ts
```diff
@@ -50,26 +65,94 @@ describe('formatUnknownError', () => {
...
+        error,
+        startClocks: clocksNow(),
+        nonErrorPrefix: 'Uncaught',
+        source: 'custom',
+        handling: ErrorHandling.UNHANDLED,
+      }).handling
+    ).toEqual(ErrorHandling.UNHANDLED)
+  })
+
+  it('should format an object error with cause', () => {
```
bcaudan, line 133:
```suggestion
  it('should compute an object error with causes', () => {
```

packages/core/src/tools/error.spec.ts
```diff
@@ -107,3 +190,48 @@ describe('createHandlingStack', () => {
...
+    expect(errorCauses).toEqual(undefined)
+  })
+
+  it('should stop recursive loop if cause is not of type Error', () => {
+    const error = new Error('foo') as ErrorWithCause
+    const nestedError = new Error('bar')
+    const deepNestedError = { biz: 'buz', cause: new Error('boo') }
+    error.cause = nestedError
+    // @ts-ignore: want to test we can handle non error cases
+    nestedError.cause = deepNestedError
```
bcaudan, line 207: **💬 suggestion:** ‏what about having the non-error cause in the first error?
it would simplify the current test and I don't think that testing that it continues to work at a deeper level bring much more value.
wdyt?

packages/core/src/tools/error.ts
```diff
@@ -32,26 +52,48 @@ export const enum ErrorHandling {
...
+type RawErrorParams = {
+  stackTrace?: StackTrace
+  error: unknown
   handlingStack?: string
-) {
-  if (!stackTrace || (stackTrace.message === undefined && !(errorObject instanceof Error))) {
+  startClocks: ClocksState
+  nonErrorPrefix: string
+  source: ErrorSource
+  handling?: ErrorHandling
```
bcaudan, line 62: **💬 suggestion:** ‏from usages
```suggestion
  handling: ErrorHandling
```

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -92,6 +89,7 @@ function processError(
       handling_stack: error.handlingStack,
       type: error.type,
       handling: error.handling,
+      causes: error.causes ?? [],
```
bcaudan, line 92: **💬 suggestion:** ‏following previous comment
```suggestion
      causes: error.causes,
```

packages/core/src/tools/error.ts
```diff
@@ -110,3 +145,19 @@ export function createHandlingStack(): string {
 
   return formattedStack!
 }
+
+export function flattenErrorCauses(error: ErrorWithCause, parentSource: ErrorSource): RawErrorCause[] | undefined {
+  let currentError = error
+  const causes: RawErrorCause[] = []
+  while (currentError?.cause instanceof Error && causes.length < 10) {
+    const stackTrace = error instanceof Error ? computeStackTrace(error) : undefined
```
liywjl, line 153:
```suggestion
    const stackTrace = currentError instanceof Error ? computeStackTrace(currentError) : undefined
```

packages/core/src/tools/error.spec.ts
```diff
@@ -1,8 +1,23 @@
...
+} from './error'
+
+describe('computeRawError', () => {
   const NOT_COMPUTED_STACK_TRACE: StackTrace = { name: undefined, message: undefined, stack: [] } as any
+  const DEFAULT_RAW_ERROR_PARMS = {
+    originalError: undefined,
+    startClocks: clocksNow(),
+    nonErrorPrefix: 'Uncaught',
+    source: ErrorSource.CUSTOM,
+  }
```
bcaudan, line 20: **💬 suggestion:** ‏always set
```suggestion
  const DEFAULT_RAW_ERROR_PARMS = {
    startClocks: clocksNow(),
    nonErrorPrefix: 'Uncaught',
    source: ErrorSource.CUSTOM,
  }
```

packages/core/src/tools/error.spec.ts
```diff
@@ -107,3 +188,38 @@ describe('createHandlingStack', () => {
...
+
+  it('should stop recursive loop if cause is not of type Error', () => {
+    const error = new Error('foo') as ErrorWithCause
+    const nestedError = { biz: 'buz', cause: new Error('boo') }
+    // @ts-ignore: want to test we can handle non error cases
+    error.cause = nestedError
+
+    const errorCauses = flattenErrorCauses(error, ErrorSource.LOGGER)
+    expect(errorCauses?.length).toEqual(undefined)
+  })
```
bcaudan, line 207: **💬 suggestion:** ‏

- more precise test description
- avoid ts-ignore annotation
- inline unneeded variable

```suggestion
  it('should return undefined if cause is not of type Error', () => {
    const error = new Error('foo') as ErrorWithCause
    error.cause = { biz: 'buz', cause: new Error('boo') } as unknown as Error

    const errorCauses = flattenErrorCauses(error, ErrorSource.LOGGER)
    expect(errorCauses?.length).toEqual(undefined)
  })
```

packages/core/src/tools/error.spec.ts
```diff
@@ -107,3 +188,38 @@ describe('createHandlingStack', () => {
...
+
+  it('should use error to extract stack trace', () => {
+    const error = new Error('foo') as ErrorWithCause
+    const nestedError = new Error('bar')
+
+    error.cause = nestedError
```
bcaudan, line 213: **💬 suggestion:** ‏inline unneeded variable
```suggestion
    error.cause = new Error('bar')
```

packages/core/src/tools/error.ts
```diff
@@ -110,3 +152,23 @@ export function createHandlingStack(): string {
...
+): RawErrorCause[] | undefined {
+  let currentError = error
+  const causes: RawErrorCause[] = []
+  while (currentError?.cause && causes.length < 10) {
+    if (!(currentError.cause instanceof Error)) break
+    causes.push({
+      message: currentError.cause.message,
+      source: parentSource,
+      type: stackTrace?.name,
+      stack: stackTrace && toStackTraceString(stackTrace),
```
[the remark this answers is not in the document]
bcaudan: Could you add a test that ensure that stack properties used for causes are indeed related to the corresponding error?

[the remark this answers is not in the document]
liywjl: done, the test is called `should compute stack trace form nested error, and not use param stacktrace`. A bit verbose for my liking, so happy to hear suggestions

packages/core/src/tools/error.spec.ts
```diff
@@ -44,32 +63,118 @@ describe('formatUnknownError', () => {
...
+      stackTrace,
+      originalError: error,
+      handling: ErrorHandling.HANDLED,
+    })
+
+    expect(formatted?.type).toEqual('TypeError')
+    expect(formatted?.message).toEqual('some typeError message')
+    expect(formatted?.causes?.[0].type).toEqual('Error')
+    expect(formatted?.causes?.[0].message).toEqual('bar')
+  })
```
bcaudan, line 118: **💬 suggestion:** ‏instead of having a dedicated test for it, what about enriching the existing `should compute an object error with causes`?
We could override error stack to validate that each stack is at the right place
```ts
    const error = new Error('foo: bar') as ErrorWithCause
    error.stack = 'Error: foo: bar\n    at <anonymous>:1:15'
   
    const nestedError = new Error('biz: buz') as ErrorWithCause
    nestedError.stack = 'NestedError: biz: buz\n    at <anonymous>:1:15'

...

    expect(formatted.type).toBe('Error')
    expect(formatted.stack).toBe(...)
    expect(causes[0].type).toBe('NestedError')
    expect(causes[0].stack).toBe(...)
```

packages/core/src/tools/error.spec.ts
```diff
@@ -153,27 +128,42 @@ describe('computeRawError', () => {
...
+    expect(formatted?.type).toEqual('TypeError')
+    expect(formatted?.message).toEqual('some typeError message')
     expect(formatted.causes?.length).toBe(2)
     const causes = formatted.causes as RawErrorCause[]
     expect(causes[0].message).toContain(nestedError.message)
     expect(causes[0].source).toContain(ErrorSource.SOURCE)
+    expect(causes[0].type).toEqual('Error')
     expect(causes[1].message).toContain(deepNestedError.message)
     expect(causes[1].source).toContain(ErrorSource.SOURCE)
+    expect(causes[1].type).toEqual('Error')
```
bcaudan, line 166: **💬 suggestion:** ‏it seems that:
- the input errors stack are not used in the assertions  
- except the message we can't distinguish the two causes
could we improve that?

liywjl: 🤦‍♂️ just saw that in face we were not actually extracting the stack traces from causes correctly. I have fixed that and updated the tests

bcaudan: Could you ensure that the types are coming from the expected errors as well?

bcaudan (approved):

BenoitZugmeyer (approved):

PR #1751, opened 2022-09-30
🐛[RUMF-1276] handle performance entry without `toJSON`

bcaudan:
## Motivation

fix issue raised by telemetry

## Changes

remove useless toJSON call and checks

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
@@ -173,10 +173,9 @@ function computeEntryTracingInfo(entry: RumPerformanceResourceTiming, configurat
 }
 
 function toPerformanceEntryRepresentation(entry: RumPerformanceEntry): PerformanceEntryRepresentation {
-  if (supportPerformanceEntry() && entry instanceof PerformanceEntry) {
-    entry.toJSON()
-  }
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -173,10 +173,9 @@ function computeEntryTracingInfo(entry: RumPerformanceResourceTiming, configurat
 }
 
 function toPerformanceEntryRepresentation(entry: RumPerformanceEntry): PerformanceEntryRepresentation {
-  if (supportPerformanceEntry() && entry instanceof PerformanceEntry) {
-    entry.toJSON()
```
BenoitZugmeyer, line 177: **❓ question:** ‏I wonder if `toJSON` is actually useful here, since we didn't even use its returned value in the original code. Maybe we could just remove it?

bcaudan: fair point, probably not.
We should probably drop `PerformanceEntryRepresentation` and use directly `PerformanceEntry` then (in the next major).

liywjl (approved):

PR #1760, opened 2022-10-06
🔊 Collect configuration telemetry event

amortemousque:
## Motivation

Engineering and product would like to easily analyze how customers configure the SDKs in order to:

- Ease troubleshooting, ex: traffic spike/drop due to sampleRate change or to customer traffic
- Track feature usage, ex: what percentage of orgs use a proxy 

This PR collect customer configuration through telemetry events.

## Changes

Add `addTelemetryConfiguration`

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -3,7 +3,7 @@ import { getCurrentSite } from '../../browser/cookie'
 import { catchUserErrors } from '../../tools/catchUserErrors'
 import { display } from '../../tools/display'
 import { assign, isPercentage, ONE_KIBI_BYTE, ONE_SECOND } from '../../tools/utils'
-import type { RawConfigurationTelemetryEvent } from '../telemetry'
+import type { RawTelemetryConfiguration } from '../telemetry'
 import { updateExperimentalFeatures } from './experimentalFeatures'
 import { initSimulation } from './simulation'
 import type { TransportConfiguration } from './transportConfiguration'
@@ -153,9 +153,7 @@ function mustUseSecureCookie(initConfiguration: InitConfiguration) {
   return !!initConfiguration.useSecureSessionCookie || !!initConfiguration.useCrossSiteSessionCookie
 }
 
-export function serializeConfiguration(
-  configuration: InitConfiguration
-): Partial<RawConfigurationTelemetryEvent['configuration']> {
+export function serializeConfiguration(configuration: InitConfiguration): Partial<RawTelemetryConfiguration> {
   return {
     session_sample_rate: configuration.sampleRate,
     telemetry_sample_rate: configuration.telemetrySampleRate,
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -149,3 +149,16 @@ export function buildCookieOptions(initConfiguration: InitConfiguration) {
 function mustUseSecureCookie(initConfiguration: InitConfiguration) {
   return !!initConfiguration.useSecureSessionCookie || !!initConfiguration.useCrossSiteSessionCookie
 }
+
+export function serializeConfiguration(configuration: InitConfiguration) {
+  return {
+    session_sample_rate: configuration.sampleRate,
+    telemetry_sample_rate: configuration.telemetrySampleRate,
+    use_before_send: !!configuration.beforeSend,
+    use_cross_site_session_cookie: configuration.useCrossSiteSessionCookie,
+    use_secure_session_cookie: configuration.useSecureSessionCookie,
+    use_proxy: configuration.proxyUrl !== undefined ? !!configuration.proxyUrl : undefined,
+    silent_multiple_init: configuration.silentMultipleInit,
+    track_session_across_subdomains: configuration.trackSessionAcrossSubdomains,
+  }
+}
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -63,6 +63,7 @@ export interface Configuration extends TransportConfiguration {
   cookieOptions: CookieOptions
   sampleRate: number
   telemetrySampleRate: number
+  telemetryConfigurationSampleRate: number
   service: string | undefined
   silentMultipleInit: boolean
 
@@ -105,6 +106,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
       cookieOptions: buildCookieOptions(initConfiguration),
       sampleRate: initConfiguration.sampleRate ?? 100,
       telemetrySampleRate: initConfiguration.telemetrySampleRate ?? 20,
+      telemetryConfigurationSampleRate: 20,
       service: initConfiguration.service,
       silentMultipleInit: !!initConfiguration.silentMultipleInit,
 
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -42,6 +42,7 @@ export interface InitConfiguration {
   enableExperimentalFeatures?: string[] | undefined
   replica?: ReplicaUserConfiguration | undefined
   datacenter?: string
+  telemetryConfigurationSampleRate?: number
 
   // simulation options
   simulationStart?: string | undefined
@@ -95,6 +96,14 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     return
   }
 
+  if (
+    initConfiguration.telemetryConfigurationSampleRate !== undefined &&
+    !isPercentage(initConfiguration.telemetryConfigurationSampleRate)
+  ) {
+    display.error('Telemetry Configuration Sample Rate should be a number between 0 and 100')
+    return
+  }
+
   // Set the experimental feature flags as early as possible, so we can use them in most places
   updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures)
 
@@ -107,7 +116,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
       cookieOptions: buildCookieOptions(initConfiguration),
       sampleRate: initConfiguration.sampleRate ?? 100,
       telemetrySampleRate: initConfiguration.telemetrySampleRate ?? 20,
-      telemetryConfigurationSampleRate: 5,
+      telemetryConfigurationSampleRate: initConfiguration.telemetryConfigurationSampleRate ?? 5,
       service: initConfiguration.service,
       silentMultipleInit: !!initConfiguration.silentMultipleInit,
 
@@ -157,6 +166,7 @@ export function serializeConfiguration(configuration: InitConfiguration): Partia
   return {
     session_sample_rate: configuration.sampleRate,
     telemetry_sample_rate: configuration.telemetrySampleRate,
+    telemetry_configuration_sample_rate: configuration.telemetryConfigurationSampleRate,
     use_before_send: !!configuration.beforeSend,
     use_cross_site_session_cookie: configuration.useCrossSiteSessionCookie,
     use_secure_session_cookie: configuration.useSecureSessionCookie,
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -3,6 +3,7 @@ import { getCurrentSite } from '../../browser/cookie'
 import { catchUserErrors } from '../../tools/catchUserErrors'
 import { display } from '../../tools/display'
 import { assign, isPercentage, ONE_KIBI_BYTE, ONE_SECOND } from '../../tools/utils'
+import type { RawConfigurationTelemetryEvent } from '../telemetry'
 import { updateExperimentalFeatures } from './experimentalFeatures'
 import { initSimulation } from './simulation'
 import type { TransportConfiguration } from './transportConfiguration'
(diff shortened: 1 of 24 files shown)


packages/core/src/domain/telemetry/rawTelemetryEvent.types.ts
```diff
@@ -0,0 +1,34 @@
...
+    kind?: string
+    stack: string
+  }
+}
+
+export interface RawErrorTelemetryEvent extends Context {
+  message: string
+  status: StatusType.error
+  type: TelemetryType.log
+}
```
bcaudan, line 29: **❗ warning:** ‏that does not seem right 😕


**💭 thought:** ‏Is there a reason to declare Raw types structure? could we use directly the `telemetry` field from the format?
```ts
export type RawDebugTelemetryEvent = TelemetryDebugEvent['telemetry']
```

packages/rum-core/src/domain/configuration.ts
```diff
@@ -142,3 +143,25 @@ export function validateAndBuildRumConfiguration(
...
+  const baseSerializedConfiguration = serializeConfiguration(configuration)
+
+  return assign(
+    {
+      premium_sample_rate: configuration.premiumSampleRate,
+      replay_sample_rate: configuration.replaySampleRate,
+      session_replay_sample_rate: configuration.sessionReplaySampleRate,
+      action_name_attribute: configuration.actionNameAttribute,
+      use_allowed_tracing_origins:
+        configuration.allowedTracingOrigins !== undefined ? !!configuration.allowedTracingOrigins : undefined,
```
bcaudan, line 157: **💬 suggestion:** ‏IIUC, there are 3 cases: 

1. `allowedTracingOrigins: undefined`
2. `allowedTracingOrigins: []`
3. `allowedTracingOrigins: ['some origin', ..]`

The current implementation will send

1.  `use_allowed_tracing_origins: undefined`
2. `use_allowed_tracing_origins: true`
3. `use_allowed_tracing_origins: true`

IMO, it could be more interesting to send:

1.  `use_allowed_tracing_origins: false` (we could send undefined but do we care about this difference?)
2. `use_allowed_tracing_origins: false`
3. `use_allowed_tracing_origins: true`

wdyt?

(probably same reasoning for `excludedActivityUrls`)

amortemousque: Fair point. Seems reasonable

liywjl (approved):
LGTM

packages/core/src/domain/telemetry/rawTelemetryEvent.types.ts
amortemousque: The main reason is that Event types have `[k: string]: unknown` at every depth. Meaning that unknown attributes are allowed. Ex:
```typescript
  addTelemetry(
    assign(
      {
        type: TelemetryType.log as const,
        message,
        status: StatusType.debug as const,
        foo: 'bar', // not detected by the linter
      },
      context
    )
  )`
Also, we have to be backward compat in the format so field like `telemetry.type` is nullable where it should not.
At least if we where to remove `[k: string]: unknown` I would be more comfortable. Do you know why is it set in the first place?

Comment (amortemousque) [reply packages/core/src/domain/telemetry/rawTelemetryEvent.types.ts:29]: My example is not the best since debug and error telemetry can accept other field. However configuration telemetry should not.

File: packages/core/src/domain/telemetry/rawTelemetryEvent.types.ts [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+
+export interface RawConfigurationTelemetryEvent {
+  type: TelemetryType.configuration
+  configuration: any
```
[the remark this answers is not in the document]
amortemousque: Was waiting for my other PR but I it's good I updated  the rum-event-format manually

packages/core/src/domain/telemetry/rawTelemetryEvent.types.ts
amortemousque: I made the change

packages/rum-core/src/boot/startRum.ts
```diff
@@ -72,6 +74,7 @@ export function startRum(
     getCommonContext,
     reportError
   )
+  addTelemetryConfiguration(serializeRumConfiguration(initConfiguration))
```
bcaudan, line 77: **🥜 nitpick:** ‏what about colocating this code with other code related to telemetry?

packages/logs/src/boot/startLogs.ts
```diff
@@ -75,6 +82,7 @@ export function startLogs(configuration: LogsConfiguration, getCommonContext: ()
     startLogsBridge(lifeCycle)
   }
 
+  addTelemetryConfiguration(serializeLogsConfiguration(initConfiguration))
```
bcaudan, line 85: **🥜 nitpick:** ‏what about colocating this code with other code related to telemetry?

packages/core/src/domain/telemetry/telemetry.spec.ts
```diff
@@ -31,6 +42,39 @@ describe('telemetry', () => {
     expect(notifySpy).toHaveBeenCalledTimes(1)
   })
 
+  describe('addTelemetryConfiguration', () => {
+    afterEach(() => {
+      resetExperimentalFeatures()
+    })
+
+    it('should notify configuration when sampled and ff is enabled', () => {
```
bcaudan, line 50: **💬 suggestion:** ‏`notify configuration` seems a bit odd to me, what about something like `collects configuration` or `notify with configuration` instead?

packages/core/src/domain/telemetry/telemetry.spec.ts
```diff
@@ -31,6 +42,39 @@ describe('telemetry', () => {
     expect(notifySpy).toHaveBeenCalledTimes(1)
   })
 
+  describe('addTelemetryConfiguration', () => {
```
bcaudan, line 45: **💬 suggestion:** ‏it could be nice to add a test to ensure that `telemetryConfigurationEnabled` depends on `telemetryEnabled`

packages/rum/src/domain/segmentCollection/startDeflateWorker.spec.ts
```diff
@@ -142,7 +142,12 @@ describe('startDeflateWorker', () => {
         throw UNKNOWN_ERROR
       })
       expect(telemetryEvents).toEqual([
-        { status: 'error' as any, message: 'boom', error: { kind: 'Error', stack: jasmine.any(String) } },
+        {
+          type: 'log' as any,
+          status: 'error' as any,
```
bcaudan, line 147: **❓ question:** ‏Could we get rid of those `any`?

test/e2e/scenario/telemetry.scenario.ts
```diff
@@ -16,8 +16,9 @@ describe('telemetry', () => {
         window.DD_LOGS!.logger.log('hop', context as any)
       })
       await flushEvents()
-      expect(serverEvents.telemetry.length).toBe(1)
-      const event = serverEvents.telemetry[0] as TelemetryErrorEvent
+
+      // get the last telemetry event because telemetry configuration may have been collected first
+      const event = serverEvents.telemetry[serverEvents.telemetry.length - 1] as TelemetryErrorEvent
```
bcaudan, line 21: **💬 suggestion:** ‏what about having a dedicated utility to return only telemetry errors? something like:
```ts
event = serverEvents.telemetryErrors()[0]
```

packages/core/src/domain/telemetry/rawTelemetryEvent.types.ts
bcaudan: > At least if we where to remove [k: string]: unknown I would be more comfortable. Do you know why is it set in the first place?

I think it is because the types are generated to allow forward compatibility.
I am fine with using RawTypes but they have to match the telemetry types (cf previous inversion between Error and Debug) and we should not use `any`.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -149,3 +161,17 @@ export function buildCookieOptions(initConfiguration: InitConfiguration) {
 function mustUseSecureCookie(initConfiguration: InitConfiguration) {
   return !!initConfiguration.useSecureSessionCookie || !!initConfiguration.useCrossSiteSessionCookie
 }
+
+export function serializeConfiguration(configuration: InitConfiguration): Partial<RawTelemetryConfiguration> {
+  return {
+    session_sample_rate: configuration.sampleRate,
+    telemetry_sample_rate: configuration.telemetrySampleRate,
+    telemetry_configuration_sample_rate: configuration.telemetryConfigurationSampleRate,
```
amortemousque, line 169: Should I update the rum-events-format since telemetryConfigurationSampleRate is ment for tests?

test/e2e/scenario/telemetry.scenario.ts
```diff
@@ -37,11 +36,41 @@ describe('telemetry', () => {
...
-      const event = serverEvents.telemetry[0] as TelemetryErrorEvent
+      expect(serverEvents.telemetryErrors.length).toBe(1)
+      const event = serverEvents.telemetryErrors[0]
       expect(event.telemetry.message).toBe('bar')
       expect(event.telemetry.error!.kind).toBe('Error')
       expect(event.telemetry.status).toBe('error')
       serverEvents.empty()
     })
+
+  createTest('send init configuration for logs')
```
amortemousque, line 47: For `addTelemetryConfiguration` to work, `session`, `viewContexts`, `actionContexts` and `batch` need to be started which is a lot of dependancies.  
I tried to unit test it but it means testing the `startRum` which is hard since we have to be sure to remove all the instrumentation after the test.  
IMO e2e test is a good option, but it has the disadvantage of having to expose `telemetryConfigurationSampleRate`. 
Wdyt?

packages/rum-core/src/boot/startRum.ts
amortemousque: As we discussed, colocating the code is not easy due to the telemetry dependencies. I made an e2e test to ensure the feature works. https://github.com/DataDog/browser-sdk/pull/1760/files#r991413769

packages/core/src/domain/telemetry/telemetry.spec.ts
amortemousque: https://github.com/DataDog/browser-sdk/blob/1efc794192f0c3d71d3ece214e07869728f23ed3/packages/core/src/domain/telemetry/telemetry.spec.ts#L77

packages/core/src/domain/configuration/configuration.ts
bcaudan: Do you mean add this parameter in the configuration telemetry?
If yes, any reason to add it or not?

sandbox/index.html
```diff
@@ -6,21 +6,30 @@
...
-      DD_LOGS.init({
-        clientToken: 'xxx',
-        telemetrySampleRate: 100,
-        enableExperimentalFeatures: [],
-      })
+      // DD_LOGS.init({
+      //   clientToken: 'xxx',
+      //   telemetrySampleRate: 100,
+      //   enableExperimentalFeatures: [],
+      // })
```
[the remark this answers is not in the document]
amortemousque: Nop..

test/e2e/scenario/telemetry.scenario.ts
```diff
@@ -37,11 +36,41 @@ describe('telemetry', () => {
...
+      enableExperimentalFeatures: ['telemetry_configuration'],
+      forwardErrorsToLogs: true,
+    })
+    .run(async ({ serverEvents }) => {
+      await flushEvents()
+      expect(serverEvents.telemetryConfigurations.length).toBe(1)
+      const event = serverEvents.telemetryConfigurations[0]
+      expect(event.telemetry.type).toBe('configuration')
+      expect(event.telemetry.configuration.forward_errors_to_logs).toEqual(true)
+      serverEvents.empty()
```
bcaudan, line 59: **🥜 nitpick:** ‏Do we need those `serverEvents.empty()`?

test/e2e/scenario/telemetry.scenario.ts
```diff
@@ -37,11 +36,41 @@ describe('telemetry', () => {
...
+    .withSetup(bundleSetup)
+    .withLogs({
+      enableExperimentalFeatures: ['telemetry_configuration'],
+      forwardErrorsToLogs: true,
+    })
+    .run(async ({ serverEvents }) => {
+      await flushEvents()
+      expect(serverEvents.telemetryConfigurations.length).toBe(1)
+      const event = serverEvents.telemetryConfigurations[0]
+      expect(event.telemetry.type).toBe('configuration')
```
bcaudan, line 57: **🥜 nitpick:** ‏testing that the type is `configuration` is a bit redundant with the fact that we have already filtered the telemetry events on this value

test/e2e/scenario/telemetry.scenario.ts
bcaudan: LGTM

test/e2e/scenario/telemetry.scenario.ts
amortemousque: I updated the `tearDownTest` to only test that we should not have telemetry errors so I can remove `serverEvents.empty()`
https://github.com/DataDog/browser-sdk/blob/f22ccfcc69c9ac16466a778e49dc299252514160/test/e2e/lib/framework/createTest.ts#L193-L195

liywjl (approved):
LGTM

test/e2e/scenario/telemetry.scenario.ts
bcaudan: ok got it!

bcaudan (approved):
Nice!

PR #1771, opened 2022-10-14
[RUMF-1337] Instrument fetch response

liywjl:
## Motivation

Most browsers do not add resource entries to the performance object until after the payload has been downloaded.
This means that when we call `performance.getEntriesByName(request.url, 'resource')` after receiving a response from `Fetch`, we rarely find a corresponding entry.

To overcome this, we should create the resource when the response payload has been downloaded by instrumenting the methods used to trigger the download.

## Changes

- create a new method `reportFetchResolve` inside `afterSend` of the `fetchObservable.ts` file
- `reportFetchResolve` attaches a callback if the response payload downloading methods are used
- if the payload is not downloaded in a certain number of milliseconds (defined by `REPORT_FETCH_TIMER`), we trigger the resource entry

## Next steps
- unit testing
- e2e testing
- determining best value for `REPORT_FETCH_TIMER`

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/fetchObservable.ts
+++ b/packages/core/src/browser/fetchObservable.ts
@@ -89,6 +89,7 @@ function beforeSend(observable: Observable<FetchContext>, input: RequestInfo, in
 }
 
 type OverloadMethodKeys = 'json' | 'text' | 'formData' | 'blob' | 'arrayBuffer'
+export const responseMethodsToOverload: OverloadMethodKeys[] = ['json', 'text', 'formData', 'blob', 'arrayBuffer']
 export const REPORT_FETCH_TIMER = 3000
 
 function afterSend(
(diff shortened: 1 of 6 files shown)


packages/core/src/browser/fetchObservable.ts
```diff
@@ -113,5 +116,61 @@ function afterSend(
...
+    Promise.race([
+      new Promise((resolve) => {
+        stopInstruments = (['json', 'text', 'formData', 'blob', 'arrayBuffer'] as OverloadMethodKeys[]).map(
+          (method) => {
+            const { stop } = instrumentMethod(
+              response,
+              method,
+              (originalMethod: () => Promise<any>) =>
+                function () {
+                  const res = originalMethod.call(response)
```
amortemousque, line 133: **❓ question:** ‏ ‏Should not you send `this` and `args` like in [here](https://github.com/DataDog/browser-sdk/blob/1bf03b1dc17d8d3c1aa3d729c44bbd17361c6ab1/packages/core/src/tools/instrumentMethod.ts#L60) to avoid changing the scope and be certain of the parameters?

packages/core/src/browser/fetchObservable.ts
```diff
@@ -113,5 +116,61 @@ function afterSend(
       observable.notify(context)
     }
   }
-  responsePromise.then(monitor(reportFetch), monitor(reportFetch))
+
+  const reportFetchResolve = (response: Response) => {
+    let stopInstruments: Array<() => void> = []
+    let timeOutId: null | number = null
+
+    Promise.race([
```
amortemousque, line 124: **❓ question:** ‏Why is a `race` required?

packages/core/src/browser/fetchObservable.ts
```diff
@@ -113,5 +116,61 @@ function afterSend(
...
+      stopInstruments.forEach(stop)
+      timeOutId && clearTimeout(timeOutId)
+
+      timeOutId = null
+      stopInstruments = []
+    }
+  }
+
+  responsePromise.then(
+    monitor((response) => (response.status !== 200 ? reportFetch(response) : reportFetchResolve(response))),
```
amortemousque, line 173: **💬 suggestion:** Maybe you can do 
```suggestion
    monitor((response) => (!response.ok ? reportFetch(response) : reportFetchResolve(response))),
```

packages/core/src/browser/fetchObservable.ts
```diff
@@ -113,5 +116,61 @@ function afterSend(
...
+
+  const reportFetchResolve = (response: Response) => {
+    let stopInstruments: Array<() => void> = []
+    let timeOutId: null | number = null
+
+    Promise.race([
+      new Promise((resolve) => {
+        stopInstruments = (['json', 'text', 'formData', 'blob', 'arrayBuffer'] as OverloadMethodKeys[]).map(
+          (method) => {
+            const { stop } = instrumentMethod(
```
amortemousque, line 128: **💬 suggestion:** ‏Maybe a new helper function `instrumentMethods(object: OBJECT,
  methods: METHOD[],...` could help simplifying

packages/core/src/browser/fetchObservable.ts
liywjl: Good point on using `this`, I will update. These methods however do not have `args` so on need

packages/core/src/browser/fetchObservable.ts
liywjl: great idea!

liywjl:
closing in favour https://github.com/DataDog/browser-sdk/pull/1778

PR #1775, opened 2022-10-17
✨ accept functions in allowedTracingOrigins/excludedActivityUrls arguments list

idanci:
## Motivation

In our product we allow users to create subdomains, ex:
* https://janedoe.example.com
* https://custom-name.example.com

At the same time we have a bunch of reserved subdomains, ex:
* https://billing.example.com
* https://auth.example.com

Last one is not technically under our control, it's auth0 [custom domain](https://auth0.com/docs/customize/custom-domains). It will reject the request with `x-datadog-X` headers with a [CORS error](https://community.auth0.com/t/cors-error-xxx-is-not-allowed-by-access-control-allow-headers-in-preflight-response/77229).
We have `/example\.com/` in our allowedTracingOrigins, but there are some reserved subdomains that are not that easy remove from the list, because we still need the subdomains with suffixes, prefixes and derived words included ex:
* https://some-author.example.com
* https://some-authority.example.com

## Changes

This change allows functions in `allowedTracingOrigins` list. This makes it easier to apply some rules to the origin instead of giving complex regexps like `https://(?!auth)(.+\\.)?${domain}\\.(com|net)($|/)`

example usage:
```typescript
const allowedOrigin = (origin: string) => {
  if (origin === 'https://auth.example.com') return false;
  if (process.env.NODE_ENV === 'development') return true;

  return RegExp(`https://(?!auth)(.+\\.)?example\\.(com|net)($|/)`).test(origin)
};

datadogRum.init({
  site: 'datadoghq.eu',
  allowedTracingOrigins: [allowedOrigin, /lvh\.me/],
});
```

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


bcaudan:
Hi @idanci,

Thanks for your contribution!
I have added a couple of changes, we will review the change internally and should integrate it in the coming days.

amortemousque (approved):

packages/core/src/tools/utils.ts
```diff
@@ -660,8 +661,22 @@ export function removeDuplicates<T>(array: T[]) {
   return arrayFrom(set)
 }
 
-export function matchList(list: Array<string | RegExp>, value: string) {
-  return list.some((item) => item === value || (item instanceof RegExp && item.test(value)))
+export type MatchPredicate = (value: string) => boolean
```
BenoitZugmeyer, line 664: **💬 suggestion:** ‏Use the 3 supported types:

```suggestion
export type MatchPredicate = string | RegExp | ((value: string) => boolean)
```
So you can simplify list types with `MatchPredicate[]`

liywjl (approved):

BenoitZugmeyer (approved):

bcaudan:
Available in v4.22.0 🚀

PR #1778, opened 2022-10-18
🐛 [RUMF-1337] report fetch on payload callback

liywjl:
## Motivation

Most browsers do not add resource entries to the performance object until after the payload has been downloaded.
This means that when we call `performance.getEntriesByName(request.url, 'resource')` after receiving a response from `Fetch`, we rarely find a corresponding entry.

To overcome this, we listen to `reponse.body` ReadableStream and only trigger the resource creation on done.

## Changes

- lift `readLimitedAmountOfBytes` into its own function in core
- create a new method `reportFetchOnPayloadComplete` inside `afterSend` of the `fetchObservable.ts` file
- use `readLimitedAmountOfBytes` to listen to the stream
- attach `reportFetchOnPayloadComplete` on the `fetch` `then` resolve function

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #1782, opened 2022-10-20
🐛 [RUMF-1337] fix fetch duration and missing fetch timings

liywjl:
## Motivation

It appears that we have am incorrect `fetch` duration value and missing `fetch` timings. This was due to the fact that we were unable to associate a given `fetch` call to an associated browser [Performance Entry](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry); that would allow us to get more detailed timings about the `fetch`.

## Changes
- `fetchStub` make async (update unit tests to use async await)
- add `ok` field in ResponseStub
- Rename `matchRequestTiming.ts` to `matchResponseToPerformanceEntry.ts`
- Have 2 functions in `matchResponseToPerformanceEntry.ts` to
- Delete `matchRequestTiming.ts`
   `matchOnPerformanceObserverCallback` to use `PerformanceObserver` to get performance entry
   `matchOnPerformanceGetEntriesByName` to use `performance.getEntriesByName`
- We rename `duration` in fetch to `resolveDuration`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

## Previous implementations
- Instrumenting `response.body` https://github.com/DataDog/browser-sdk/pull/1778
- Instrumenting `response` methods https://github.com/DataDog/browser-sdk/pull/1771


packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -37,7 +43,28 @@ export function startResourceCollection(
   sessionManager: RumSessionManager
 ) {
   lifeCycle.subscribe(LifeCycleEventType.REQUEST_COMPLETED, (request: RequestCompleteEvent) => {
-    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processRequest(request, configuration, sessionManager))
+    if (request.type === RequestType.XHR || request.status === 0) {
```
liywjl, line 46: `request.status === 0` seems to resolve before we have time to set the `PerformanceObserver` so we need to capture this edge case

packages/rum-core/src/domain/rumEventsCollection/resource/matchResponseToPerformanceEntry.ts
```diff
@@ -0,0 +1,99 @@
...
+          // log that there is an issue
+          if (candidates.length > 2) resolve(undefined)
+          if (candidates.length === 2 && firstCanBeOptionRequest(candidates)) resolve(candidates[1])
+          if (candidates.length === 1) resolve(candidates[0])
+        }
+      })
+      observer.observe({ entryTypes: ['resource'] })
+    })
+      // @ts-ignore: if a browser supports fetch, it likely supports finally
+      .finally(reset)
```
liywjl: In the unlikely case that the `PerformanceObserver` callback does not fire, we do not register the `fetch`. I am unsure if there is a way around this as the fetch request may just be taking a long time, meaning setting a `Timeout` would accidentally cause us to not fetch the PerformanceEntry

packages/rum-core/src/domain/rumEventsCollection/resource/matchResponseToPerformanceEntry.ts
```diff
@@ -0,0 +1,110 @@
...
+  entries
+    .map((entry) => entry.toJSON() as RumPerformanceResourceTiming)
+    .filter(toValidEntry)
+    .filter((entry) =>
+      isBetween(
+        entry,
+        startClocks.relative,
+        endTime({
+          startTime: startClocks.relative,
+          duration: elapsed(startClocks.timeStamp, timeStampNow()),
```
liywjl, line 89: Note: we do not use the response duration value, and instead dynamically calculate based on execution

packages/rum-core/src/domain/rumEventsCollection/resource/matchResponseToPerformanceEntry.ts
```diff
@@ -0,0 +1,110 @@
...
+            resolve(undefined)
+          }
+          if (candidates.length === 2 && firstCanBeOptionRequest(candidates)) {
+            addTelemetryDebug('2 matching performance entries found')
+            resolve(candidates[1])
+          }
+          if (candidates.length === 1) resolve(candidates[0])
+        }
+      })
+      observer.observe({ entryTypes: ['resource'] })
```
liywjl, line 49: In the unlikely case that the `PerformanceObserver` callback does not fire, we do not register the fetch. 

I am unsure if there is a way around this as the fetch request may just be taking a long time, meaning setting a "timeout" would accidentally cause us to not fetch the PerformanceEntry

liywjl:
closing in favour of https://github.com/DataDog/browser-sdk/pull/1802

Issue #807, opened 2021-04-23
Handle serialization of cyclic object


dnepro:
Hey there!

I work with a very large Angular Application which creates a an Error like:
```
11:02:55.366 ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'collectionSize: 0'. Current value: 'collectionSize: 4'.
    Angular (7)
    View_FlightTimeListViewComponent_0 FlightTimeListViewComponent.html:21
    Angular (66)
    RxJS (6)
    Angular (9)
automaticErrorCollection.js:49
```
followed by a second error like:
```
11:02:55.377 ERROR CONTEXT 
{…}
​
elDef: Object { nodeIndex: 29, bindingIndex: 3, outputIndex: 0, … }
​
elView: Object { def: {…}, parent: {…}, state: 1036, … }
​
nodeDef: Object { nodeIndex: 30, bindingIndex: 3, outputIndex: 1, … }
​
nodeIndex: 30
​
view: Object { def: {…}, parent: {…}, state: 1036, … }
​
<prototype>: Object { elOrCompView: Getter, injector: Getter, component: Getter, … }
automaticErrorCollection.js:49
```

the first error shows up correctly in datadog, but the second results in:
`console error: ERROR CONTEXT <error: unable to serialize object>`

When i tried to serialize the error with "JSON.stringify", i receive this: 
```
11:21:16.362 Uncaught TypeError: cyclic object value
    <anonymous> debugger eval code:1
debugger eval code:1:6
```
It would be great if something like from this [MDN Article on TypeError: cyclic object value](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value) could be added directly to the browser logs or could be passed to datadog

Best Regards,
Roman

dnepro:
I just found the [createCircularReferenceChecker](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/tools/context.ts#L53) but it seems it is not used or has some other issues

bcaudan:
Hello @dnepro,

Indeed that is our current behavior when trying to serialize an object with a cycle: cf [related test](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/tools/utils.spec.ts#L262).
Removing the cyclic property from the serialization could be a nice improvement.
We'd be happy to review a PR for it if you want to give it a try.

yannickadam:
Hi!
[...]

PR #1783, opened 2022-10-20
🐛 [RUMF-1410] Allow serialization of objects with cyclic references

yannickadam:
## Motivation

Fix for https://github.com/DataDog/browser-sdk/issues/807

## Changes

Added a replacer within the jsonStringify function that removes cyclic references while serializing.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/utils.ts
+++ b/packages/core/src/tools/utils.ts
@@ -173,14 +173,14 @@ export function jsonStringify(value: unknown, space?: string | number): string |
   const getCyclicReplacer = <T>() => {
     // Using a weakmap instead of a weakset to support IE11
     const visited = new WeakMap<object, boolean>()
-    return (_k: string, v: T) => {
-      if (isValidObject(v)) {
-        if (visited.has(v)) {
+    return (_key: string, value: T) => {
+      if (isValidObject(value)) {
+        if (visited.has(value)) {
           return '<warning: cyclic reference not serialized>'
         }
-        visited.set(v, true)
+        visited.set(value, true)
       }
-      return v
+      return value
     }
   }
 
--- a/packages/core/src/tools/utils.ts
+++ b/packages/core/src/tools/utils.ts
@@ -171,14 +171,10 @@ export function jsonStringify(value: unknown, space?: string | number): string |
   // Line below can be removed when migrating to TS4.8+
   const isValidObject = (v: any): v is object => typeof v === 'object' && v !== null
   const getCyclicReplacer = <T>() => {
-    // Using a weakmap instead of a weakset to support IE11
-    const visited = new WeakMap<object, boolean>()
+    const circularReferenceChecker = createCircularReferenceChecker()
     return (_key: string, value: T) => {
-      if (isValidObject(value)) {
-        if (visited.has(value)) {
-          return '<warning: cyclic reference not serialized>'
-        }
-        visited.set(value, true)
+      if (isValidObject(value) && circularReferenceChecker.hasAlreadyBeenSeen(value)) {
+        return '<warning: cyclic reference not serialized>'
       }
       return value
     }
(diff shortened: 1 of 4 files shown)


liywjl (approved):

packages/core/src/tools/utils.ts
```diff
@@ -156,24 +156,36 @@ export function noop() {}
...
   const restoreValuePrototypeToJson = detachToJsonMethod(Object.getPrototypeOf(value))
   const restoreValueToJson = detachToJsonMethod(value)
 
+  // Line below can be removed when migrating to TS4.8+
+  const isValidObject = (v: any): v is object => typeof v === 'object' && v !== null
+  const getCyclicReplacer = <T>() => {
+    // Using a weakmap instead of a weakset to support IE11
+    const visited = new WeakMap<object, boolean>()
+    return (_k: string, v: T) => {
+      if (isValidObject(v)) {
```
bcaudan, line 177: **🥜 nitpick:** ‏using a meaningful word instead of a letter could be easier to follow, something like `candidate` or `value` instead of `v`.

bcaudan (approved):
Nice!

amortemousque:
Looks great!

amortemousque (approved):

liywjl (approved):

bcaudan (approved):

packages/core/src/tools/utils.ts
```diff
@@ -156,24 +156,32 @@ export function noop() {}
...
-  // Note: The order matter here. We need to detach toJSON methods on parent classes before their
+  // Note: The order matters here. We need to detach toJSON methods on parent classes before their
   // subclasses.
   const restoreObjectPrototypeToJson = detachToJsonMethod(Object.prototype)
   const restoreArrayPrototypeToJson = detachToJsonMethod(Array.prototype)
   const restoreValuePrototypeToJson = detachToJsonMethod(Object.getPrototypeOf(value))
   const restoreValueToJson = detachToJsonMethod(value)
 
+  // Line below can be removed when migrating to TS4.8+
+  const isValidObject = (v: any): v is object => typeof v === 'object' && v !== null
```
BenoitZugmeyer, line 172: **💬 suggestion:** ‏use `getType(destination) === 'object'` instead

packages/core/src/tools/utils.ts
```diff
@@ -156,24 +156,32 @@ export function noop() {}
...
+  const getCyclicReplacer = <T>() => {
```
BenoitZugmeyer, line 173: **🥜 nitpick:** ‏the type parameter `T` isn't useful here, and could be replaced by `unknown` in the argument list


**🥜 nitpick:** ‏FMU you create a closure just to keep the `circularReferenceChecker` variable private. IMO this is not worth it and the following would be a bit easier to read:

```ts
const circularReferenceChecker = createCircularReferenceChecker()
const cyclicReplacer = (_key: string, value: unknown) => {
  if (isValidObject(value) && circularReferenceChecker.hasAlreadyBeenSeen(value)) {
    return '<warning: cyclic reference not serialized>'
  }
  return value
}
```

packages/core/src/tools/utils.ts
```diff
@@ -490,24 +498,13 @@ interface CircularReferenceChecker {
...
-        if (!has) {
-          set.add(value)
-        }
-        return has
-      },
-    }
-  }
-  const array: any[] = []
+  // Using a weakmap instead of a weakset to support IE11
+  const map: WeakMap<any, boolean> = new WeakMap()
```
BenoitZugmeyer, line 502: **👏 praise:** ‏great, love it :)

BenoitZugmeyer (approved):

PR #1787, opened 2022-10-24
[REPLAY] Add support for shadow DOM

ThibautGeriz:
## Motivation

Add support for shadow DOM

##### External doc:
- https://javascript.info/shadow-dom
- https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM

_best reviewed by commits_

## Changes

- [x] extract functions in `packages/rum-core/src/browser/htmlDomUtils.ts`
- [x] update typings:  https://github.com/DataDog/rum-events-format/pull/108
- [x] support for serialisation
- [x] support events: more info here: https://javascript.info/shadow-dom-events#event-composed
- [x] fix privacy setting to access parent across shadow DOM
- [x] support `change` event that do not cross the shadow root 
- [x] add e2e tests for shadow DOM
- [x] hide the feature behind the `record_shadow_dom` experiment flag

## Testing

- [x] Local
  - [x] simple shadow DOM site
  - [x] TODO shadow DOM app
- [x] Staging
  - [ ] [~lit~](https://lit.dev/): won't work because of the lack of support of [`ShadowRoot.adoptedStyleSheets`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/adoptedStyleSheets)
  - [ ] ~ionic~: won't work because of the lack of support of [`ShadowRoot.adoptedStyleSheets`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/adoptedStyleSheets)
- [x] Unit
- [x] End to end

### Only support open shadow root

Working
```typescript
element.attachedShadow({ mode: 'open' })
```

Not Working
```typescript
element.attachedShadow({ mode: 'closed' })
```

### No support for dynamic shadow DOM

This part only apply when a node is added to the DOM after the FS. When a node is added we check if it has shadowNode. If we were to add the host to the DOM and then add the shadow root then it would not be recorded. We would require monkey patching to solve that and.

Working
```ts
const recordedDOM = document.querySelector('body');
const shadowHost = document.createElement('div');
shadowHost.attachShadow();
recordedDOM.appendChild(shadowHost); // recorded with shadow DOM
shadowHost.shadowRoot.appendChild('span');
```

Not Working
```ts
const recordedDOM = document.querySelector('body');
const shadowHost = document.createElement('div');
recordedDOM.appendChild(shadowHost); // recorded without shadow DOM

// asynchronously wait for a bit of time,
// causing the serialization to happen in between

shadowHost.attachShadow();
shadowHost.shadowRoot.appendChild('span');
```

### No support for [`ShadowRoot.adoptedStyleSheets`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/adoptedStyleSheets)

This is used by [Stencil](https://www.npmjs.com/package/@stencil/core) or [Lit](https://www.npmjs.com/package/lit)


packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -50,6 +55,16 @@ describe('initInputObserver', () => {
...
 
+  it('collects input values when an "input" event is composed', () => {
+    stopInputObserver = initInputObserver(inputCallbackSpy, DefaultPrivacyLevel.ALLOW)
+    dispatchComposedInputEvent('foo')
+
+    expect(inputCallbackSpy).toHaveBeenCalledOnceWith({
+      text: 'foo',
+      id: jasmine.any(Number) as unknown as number,
+    })
+  })
```
ThibautGeriz, line 66: I cannot test with a shadow DOM because it seems that untrusted event do not go through. I create fake composed event and it _works_

packages/rum/src/domain/record/serialize.ts
```diff
@@ -167,6 +169,7 @@ export function serializeElementNode(element: Element, options: SerializeOptions
       },
       childNodes: [],
       isSVG,
+      isShadowHost,
```
BenoitZugmeyer, line 179: **💬 suggestion:** ‏For hidden nodes, `isShadowHost` should not matter. I would remove this property and move the `const isShadowHost = ...` closer to where it is used

packages/rum/src/domain/record/node.typeGuards.ts
```diff
@@ -1,3 +1,7 @@
 export function isElement(node: Node): node is Element {
   return node.nodeType === node.ELEMENT_NODE
 }
+
+export function isShadowRoot(node: any): node is ShadowRoot {
```
BenoitZugmeyer, line 5: **💬 suggestion:** ‏`any` is a bit too generic here. For example, one could pass `null`, and the function would crash. What about accepting only `Node` and do some cast in the condition below?

packages/rum/src/domain/record/record.ts
```diff
@@ -43,6 +44,13 @@ export function record(options: RecordOptions): RecordAPI {
   const mutationController = new MutationController()
   const elementsScrollPositions = createElementsScrollPositions()
 
+  const mutationCb = (mutation: BrowserMutationPayload) =>
+    emit(assembleIncrementalSnapshot<BrowserMutationData>(IncrementalSource.Mutation, mutation))
+
+  const shadowDomCreatedCb = (shadowRoot: ShadowRoot) => {
+    startMutationObserver(mutationController, mutationCb, options.configuration, shadowDomCreatedCb, shadowRoot)
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: Same way we test for mutations? Add a shadowroot element, then add an element inside, and check for mutation records?

packages/rum/src/domain/record/serialize.spec.ts
```diff
@@ -40,10 +40,17 @@ const DEFAULT_SERIALIZATION_CONTEXT = {
   elementsScrollPositions: createElementsScrollPositions(),
 }
 
+const shadowDomCreatedCb = jasmine.createSpy()
+
+beforeEach(() => {
+  shadowDomCreatedCb.calls.reset()
+})
+
```
BenoitZugmeyer, line 48: **🥜 nitpick:** ‏we usually:
* declare the spy in the `describe` callback (with a `Spy` suffix)
* create it in the `beforeEach` callback

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -85,6 +95,13 @@ describe('initInputObserver', () => {
     input.value = newValue
     input.dispatchEvent(createNewEvent('input', { target: input }))
   }
+
+  function dispatchComposedInputEvent(newValue: string) {
+    input.value = newValue
+    const event = createNewEvent('input', { target: sandbox, composed: true })
+    spyOn(event, 'composedPath').and.returnValue([input, sandbox, document.body])
```
BenoitZugmeyer, line 102: **💬 suggestion:** ‏no need for a Spy here:

```js
event.composedPath = () => [...]
```

packages/rum/src/domain/record/observers.spec.ts
BenoitZugmeyer: Could you add a comment about that in the file, so we don't struggle again when we try to dispatch events in shadow DOM in the future?

packages/rum/src/domain/record/observers.ts
```diff
@@ -497,3 +498,10 @@ export function initFrustrationObserver(lifeCycle: LifeCycle, frustrationCb: Fru
     }
   }).unsubscribe
 }
+
+function getEventTarget(event: Event): Node {
+  if (event.composed === false) {
```
BenoitZugmeyer, line 503: **💬 suggestion:** ‏
```ts
if (!event.composed) { ... }
```
this way, it would still use `event.target` for emulated events without a `composed` property

packages/rum/src/domain/record/mutationObserver.ts
```diff
@@ -98,15 +109,17 @@ export class MutationController {
...
+  target: Node
 ) {
   // Discard any mutation with a 'target' node that:
   // * isn't injected in the current document or isn't known/serialized yet: those nodes are likely
   // part of a mutation occurring in a parent Node
   // * should be hidden or ignored
   const filteredMutations = mutations.filter(
     (mutation): mutation is WithSerializedTarget<RumMutationRecord> =>
-      document.contains(mutation.target) &&
+      target.contains(mutation.target) &&
```
BenoitZugmeyer, line 122: **💭 thought:** ‏I think that when `mutation.target` is a Shadow Root, we should use its host instead.

packages/rum/src/domain/record/observers.ts
```diff
@@ -96,10 +97,16 @@ interface ObserverParam {
   styleSheetCb: StyleSheetCallback
   focusCb: FocusCallback
   frustrationCb: FrustrationCallback
+  shadowDomCreatedCb: ShadowDomCallBack
```
BenoitZugmeyer, line 100: **🥜 nitpick:** ‏all other callbacks are used to emit a record. It feels a bit strange to have this one at the same level.

BenoitZugmeyer (commented):
You could move all your introduced typeguards and utilities like `getChildNodes` and `getParentNode` to `record/utils.ts`, making them colocated could improve their discoverability.

packages/rum/src/domain/record/mutationObserver.ts
```diff
@@ -50,19 +53,28 @@ export type RumMutationRecord =
 export function startMutationObserver(
   controller: MutationController,
   mutationCallback: MutationCallBack,
-  configuration: RumConfiguration
+  configuration: RumConfiguration,
+  shadowDomCreatedCb: ShadowDomCallBack,
```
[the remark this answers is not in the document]
ThibautGeriz: hum I took my inspiration from this piece of code
```typescript
interface ObserverParam {
  lifeCycle: LifeCycle
  mutationController: MutationController
  elementsScrollPositions: ElementsScrollPositions
  mutationCb: MutationCallBack
  mousemoveCb: MousemoveCallBack
  mouseInteractionCb: MouseInteractionCallBack
  scrollCb: ScrollCallback
  viewportResizeCb: ViewportResizeCallback
  inputCb: InputCallback
  mediaInteractionCb: MediaInteractionCallback
  styleSheetCb: StyleSheetCallback
  focusCb: FocusCallback
  frustrationCb: FrustrationCallback
  shadowDomCreatedCb: ShadowDomCallBack
}
```
Do you want to rename only here and leave with the the `cb` suffix in the observer?

packages/rum/src/domain/record/mutationObserver.ts
ThibautGeriz: It works with the `shadowRoot`, Why do you want to use the host instead?

packages/rum/src/domain/record/observers.ts
ThibautGeriz: I do not have a strong opinion on this, if you provide a suggestion, I will implement it

ThibautGeriz:
> You could move all your introduced typeguards and utilities like `getChildNodes` and `getParentNode` to `record/utils.ts`, making them colocated could improve their discoverability.

In the frontend we avoid have file named `utils.ts`. I can do it since you own that code and its convention but I like to be explicit, what do you think of:
- leaving type guard where it is
- move getChildNodes` and `getParentNode` to `html-getter.utils.ts` or something else but not simply utils

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -0,0 +1,232 @@
...
+  })
+  describe('incremental snapshot', () => {
+    createTest('record click')
+      .withRum({ enableExperimentalFeatures: ['recordShadowDom'] })
+      .withRumInit(initRumAndStartRecording)
+      .withSetup(bundleSetup)
+      .withBody(
+        html`
+          ${simpleShadowDom}
+          <my-open-web-component id="some" />
```
BenoitZugmeyer, line 124: **💬 suggestion:** it appears that ‏the id is not relevent for this test and those below. Maybe remove it?

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -0,0 +1,232 @@
...
+        `
+      )
+      .run(async ({ serverEvents }) => {
+        await browserExecute(() => document.documentElement.outerHTML)
```
BenoitZugmeyer, line 128: **💬 suggestion:** ‏This does not appear to be useful. Remove it here and in the other test cases too.

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -0,0 +1,232 @@
...
+        const input = await getNodeInsideShadowDom('input')
+        await input.addValue('t')
```
BenoitZugmeyer, line 183: **💬 suggestion:** ‏make a mutation without triggering other events, so the test case stays simpler/focused.

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -0,0 +1,232 @@
...
+        expect(mutationRecord.data.source).toBe(IncrementalSource.Mutation)
+        const mutationData = mutationRecord.data as BrowserMutationData
+        expect(mutationData.adds.length).toBe(1)
+        expect(mutationData.adds[0].parentId).toBe(pNode.id)
+        const textNode = mutationData.adds[0].node as TextNode
+        expect(textNode.type).toBe(NodeType.Text)
+        expect(textNode.textContent).toBe('totot')
+        expect(mutationData.removes.length).toBe(1)
+        expect(mutationData.removes[0].parentId).toBe(pNode.id)
+        expect(mutationData.removes[0].id).toBe(pNode.childNodes[0].id)
```
BenoitZugmeyer, line 201: **💬 suggestion:** ‏Use [`createMutationPayloadValidatorFromSegment`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum/test/utils.ts#L476) to test for records thoroughly. This should greatly simplify those assertions. See other usages for examples how to use it.

mquentin (approved):
LGTM session replay. Nice addition and pretty neat code.

I am not super fan of the [getNodeOrShadowHost(parent) as NodeWithSerializedNode](https://github.com/DataDog/browser-sdk/pull/1787/files#diff-3bd29f6c39d52d61a43a4ed27579824f4029e66a8a1abcf849a6a6941a975887R243) tho

packages/rum/src/domain/record/serialize.ts
```diff
@@ -232,8 +243,9 @@ function serializeCDataNode(): CDataNode {
 
 export function serializeChildNodes(node: Node, options: SerializeOptions): SerializedNodeWithId[] {
   const result: SerializedNodeWithId[] = []
+  const childNodes = isExperimentalFeatureEnabled('recordShadowDom') ? getChildNodes(node) : node.childNodes
```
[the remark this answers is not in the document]
ThibautGeriz: I had the same thought and went against it but mostly because I think helper should be pure and should not have side effect (aka FF). Happy to change it if you prefer

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -0,0 +1,232 @@
...
+    })
+    div1.appendChild(label);
+    div1.appendChild(input);
+    div1.setAttribute("id", "div1-" + this.getAttribute("id"));
+    this.shadowRoot.appendChild(div1);
+    this.shadowRoot.appendChild(div2);
+  }
+}
+window.customElements.define("my-open-web-component", MyOpenWebComponent);
+</script>
```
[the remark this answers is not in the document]
ThibautGeriz: > maybe this could be mitigated by factorizing common parts with functions.

This seems fragile

Happy to discuss this over zoom because It's a bit of a pain to maintain, when write the e2e tests I had to have that in a HTML file on the side to check if it was working as expected, having multiple of them, albeit simpler, might complexify the whole process

packages/rum/src/domain/record/record.ts
```diff
@@ -42,28 +48,53 @@ export function record(options: RecordOptions): RecordAPI {
...
+
   const mutationCb = (mutation: BrowserMutationPayload) => {
     emit(assembleIncrementalSnapshot<BrowserMutationData>(IncrementalSource.Mutation, mutation))
   }
   const inputCb: InputCallback = (s) => emit(assembleIncrementalSnapshot<InputData>(IncrementalSource.Input, s))
-
   const shadowDomCreatedCallback = (shadowRoot: ShadowRoot) => {
-    startMutationObserver(mutationController, mutationCb, options.configuration, shadowDomCreatedCallback, shadowRoot)
+    const shadowDomMutationController = new MutationController()
```
BenoitZugmeyer, line 69: **💬 suggestion:** drop mutation controller (seen IRL)

packages/rum/src/domain/record/mutationObserver.ts
```diff
@@ -243,6 +270,7 @@ function processChildListMutations(
         parentId: getSerializedNodeId(getNodeOrShadowHost(parent) as NodeWithSerializedNode),
         id: getSerializedNodeId(node),
       })
+      traverseRemovedShadowDom(node, shadowDomCallBacks.shadowDomRemovedCallback)
```
BenoitZugmeyer, line 273: Move this line above (seen IRL)

packages/rum/src/domain/record/record.ts
```diff
@@ -119,15 +166,23 @@ export function record(options: RecordOptions): RecordAPI {
...
+    },
     takeSubsequentFullSnapshot: (timestamp) =>
       takeFullSnapshot(timestamp, {
         status: SerializationContextStatus.SUBSEQUENT_FULL_SNAPSHOT,
         elementsScrollPositions,
       }),
-    flushMutations: () => mutationController.flush(),
+    flushMutations: () => {
+      shadowDomCallBacks.forEach(({ flush }) => flush())
+      documentMutationController.flush()
```
BenoitZugmeyer, line 184: **💬 suggestion:** ‏move those lines in a `flushMutations` function to factorize them

packages/rum/src/domain/record/mutationObserver.spec.ts
```diff
@@ -431,6 +447,93 @@ describe('startMutationCollection', () => {
         ],
       })
     })
+
+    describe('for shadow DOM', () => {
+      it('should call shadowDomCreatedCallback when host is added', () => {
+        updateExperimentalFeatures(['recordShadowDom'])
```
BenoitZugmeyer, line 453: **💬 suggestion:** ‏use `snake_case` for flag names

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -363,12 +363,37 @@ describe('record', () => {
       expect(mutationData.removes.length).toBe(1)
     })
 
-    function createShadow(children: Element[]) {
+    it('should clean the state when both the parent  shadow host is removed to avoid memory leak', () => {
```
BenoitZugmeyer, line 366: **💬 suggestion:** ‏I think you're missing a "and the" in the title

packages/rum/src/domain/record/record.ts
```diff
@@ -80,21 +78,23 @@ export function record(options: RecordOptions): RecordAPI {
...
     timestamp = timeStampNow(),
-    serializationContext = { status: SerializationContextStatus.INITIAL_FULL_SNAPSHOT, elementsScrollPositions }
+    serializationContext = { status: SerializationContextStatus.INITIAL_FULL_SNAPSHOT, elementsScrollPositions },
+    flushMutationsFromPreviousFs?: () => void
   ) => {
     shadowDomCallBacks.forEach(({ flush }) => flush())
-    documentMutationController.flush() // process any pending mutation before taking a full snapshot
+    if (flushMutationsFromPreviousFs) {
+      flushMutationsFromPreviousFs() // process any pending mutation before taking a full snapshot
+    }
```
BenoitZugmeyer, line 73: **💬 suggestion:** ‏Remove those lines from `takeFullSnapshot`.  Flush the mutations before calling `takeFullSnapshot()`, when needed. This should simplify things a bit

packages/rum/src/domain/record/record.ts
```diff
@@ -15,22 +16,29 @@ import type {
...
+interface ShadowDomCallBacks {
+  stop: () => void
+  flush: () => void
+}
 export interface RecordAPI {
   stop: () => void
   takeSubsequentFullSnapshot: (timestamp?: TimeStamp) => void
+  // the following is only used for testing purposes
   flushMutations: () => void
+  shadowDomCallBacks: Map<ShadowRoot, ShadowDomCallBacks>
```
bcaudan, line 41: **💬 suggestion:** flushMutations is still used in prod code‏
```suggestion
  flushMutations: () => void
   // the following is only used for testing purposes
  shadowDomCallBacks: Map<ShadowRoot, ShadowDomCallBacks>
```

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -196,6 +213,191 @@ describe('record', () => {
...
+      updateExperimentalFeatures(['record_shadow_dom'])
+      const div = document.createElement('div')
+      div.className = 'toto'
+      const shadowRoot = createShadow([div])
+      startRecording()
+      expect(getEmittedRecords().length).toBe(recordsPerFullSnapshot())
+      expect(recordApi.shadowDomCallBacks.size).toBe(1)
+      shadowRoot.host.remove()
+      recordApi.flushMutations()
+      expect(recordApi.shadowDomCallBacks.size).toBe(0)
```
bcaudan, line 357: **💬 suggestion:** ‏IMO having unit test for the shadow dom file to test the behavior of adding/removing shadow roots could be enough and would avoid to expose the map through recordingAPI.

wdyt?

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -0,0 +1,242 @@
...
+    this.shadowRoot.appendChild(div);
+  }
+}
+      window.customElements.define("my-div", CustomDiv);
+ </script>
+ `
+
+describe('recorder with shadow DOM', () => {
+  describe('full snapshot', () => {
+    createTest('can overwrite with mask-user-input')
```
bcaudan, line 84: **💭 thought:** ‏Looking at your test cases description, I'm not sure to get what is the goal of each and if it matches what I would expect to be tested with e2e, meaning:

- mutations inside shadow DOM are collected
- clicks inside shadow DOM are collected with correct target
- change inside shadow DOM are collected with correct target

Could we clarify that?

packages/rum/src/domain/record/record.spec.ts
ThibautGeriz: It was an important thing that I missed and would create memory leaks therefore I added a unit tests to avoid introducing a regression that being said I agree that we're testing an implementation details. I can delete it if you want

test/e2e/scenario/recorder/shadowDom.scenario.ts
ThibautGeriz: I am not sure what you expect it, can we jump in a call quickly tomorrow?

packages/rum/src/domain/record/record.ts
```diff
@@ -40,14 +48,24 @@ export function record(options: RecordOptions): RecordAPI {
...
+    emit(assembleIncrementalSnapshot<BrowserMutationData>(IncrementalSource.Mutation, mutation))
+  }
+  const inputCb: InputCallback = (s) => emit(assembleIncrementalSnapshot<InputData>(IncrementalSource.Input, s))
+
+  const {
+    shadowDomCallBacks,
+    flush: flushMutationsFromShadowRoots,
+    shadowDomCreatedCallback,
+    shadowDomRemovedCallback,
+  } = withShadowDomHelpers(options.configuration, { mutationCb, inputCb })
```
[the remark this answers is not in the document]
ThibautGeriz: I've kept `shadowDomCallBacks` in there until we agree on the other comment here: https://github.com/DataDog/browser-sdk/pull/1787#discussion_r1047290624
Since it's leaking a lot of implementation, instead of exporting the Map we could export
```ts
interface ShadowRootsController {
  addShadowRoot: (shadowRoot: ShadowRoot) => void
  removeShadowRoot: (shadowRoot: ShadowRoot) => void
  stop: () => void
  flush: () => void
+  activeShadowRootCountRef: {
     value: number
  }
}
```

packages/rum/src/domain/record/serialize.ts
```diff
@@ -61,24 +61,29 @@ export type SerializationContext =
...
 
+export type ShadowDomCallBack = (shadowRoot: ShadowRoot) => void
+
 export interface SerializeOptions {
   serializedNodeIds?: Set<number>
   ignoreWhiteSpace?: boolean
   parentNodePrivacyLevel: ParentNodePrivacyLevel
   serializationContext: SerializationContext
   configuration: RumConfiguration
+  shadowDomCreatedCallback: ShadowDomCallBack
```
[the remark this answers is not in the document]
BenoitZugmeyer: I agree, if it can fit in SerializationContext typewise, it could be nice

packages/rum-core/src/browser/htmlDomUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  return isElementNode(node) && node.shadowRoot !== null
+}
+
+export function isNodeShadowRoot(node: Node): node is ShadowRoot {
+  const shadowRoot = node as ShadowRoot
+  return !!shadowRoot.host && isElementNode(shadowRoot.host)
+}
+
+export function getChildNodes(node: Node) {
+  return isElementNode(node) && node.shadowRoot ? node.shadowRoot.childNodes : node.childNodes
```
BenoitZugmeyer, line 23: **🥜 nitpick:** ‏
```suggestion
  return isNodeShadowHost(node) ? node.shadowRoot.childNodes : node.childNodes
```

packages/rum/src/domain/record/mutationObserver.spec.ts
```diff
@@ -431,6 +452,94 @@ describe('startMutationCollection', () => {
         ],
       })
     })
+
+    describe('for shadow DOM', () => {
+      it('should call shadowDomCreatedCallback when host is added', () => {
```
BenoitZugmeyer, line 457: **💭 thought:** ‏
```suggestion
      it('should call addShadowRoot when host is added', () => {
```

packages/rum/src/domain/record/mutationObserver.spec.ts
```diff
@@ -431,6 +452,94 @@ describe('startMutationCollection', () => {
...
+              parent: expectInitialNode({ idAttribute: 'sandbox' }),
+              node: expectedHost,
+            },
+          ],
+        })
+        expect(addShadowRootSpy).toHaveBeenCalledOnceWith(shadowRoot)
+        expect(removeShadowRootSpy).not.toHaveBeenCalled()
+      })
+
+      it('should call shadowDomCreatedCallback when host is removed', () => {
```
BenoitZugmeyer, line 488: **💬 suggestion:** ‏
```suggestion
      it('should call removeShadowRoot when host is removed', () => {
```

packages/rum/src/domain/record/mutationObserver.spec.ts
```diff
@@ -431,6 +452,94 @@ describe('startMutationCollection', () => {
...
+              parent: expectInitialNode({ idAttribute: 'sandbox' }),
+              node: expectInitialNode({ idAttribute: 'host' }),
+            },
+          ],
+        })
+        expect(addShadowRootSpy).not.toHaveBeenCalled()
+        expect(removeShadowRootSpy).toHaveBeenCalledOnceWith(shadowRoot)
+      })
+
+      it('should call shadowDomCreatedCallback when parent of host is removed', () => {
```
BenoitZugmeyer, line 514: **💬 suggestion:** ‏
```suggestion
      it('should call removeShadowRoot when parent of host is removed', () => {
```

BenoitZugmeyer (approved):

packages/rum/src/domain/record/shadowDom.ts
```diff
@@ -0,0 +1,74 @@
...
+export type ShadowDomCallBack = (shadowRoot: ShadowRoot) => void
+
+export interface ShadowRootsController {
+  addShadowRoot: ShadowDomCallBack
+  removeShadowRoot: ShadowDomCallBack
+  stop: () => void
+  flush: () => void
+}
+
+export const initShadowRootsController = (
```
bcaudan, line 21: **💬 suggestion:** ‏what about renaming `shadowDom.ts` to `shadowRootsController.ts`?

packages/rum/src/domain/record/shadowDom.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  function addShadowRoot(shadowRoot: ShadowRoot) {
+    const { stop: stopMutationObserver, flush } = startMutationObserver(
+      mutationCb,
+      configuration,
+      shadowRootsController,
+      shadowRoot
+    )
+    // the change event no do bubble up across the shadow root, we have to listen on the shadow root
```
bcaudan, line 50: **🥜 nitpick:** ‏typo?
```suggestion
    // the change event do not bubble up across the shadow root, we have to listen on the shadow root
```

packages/rum/src/domain/record/shadowDom.ts
```diff
@@ -0,0 +1,74 @@
...
+    stop: () => {
+      shadowDomCallBacks.forEach(({ stop }) => stop())
+    },
+    flush: () => {
+      shadowDomCallBacks.forEach(({ flush }) => flush())
+    },
+    addShadowRoot,
+    removeShadowRoot,
+  }
+  return shadowRootsController
```
bcaudan, line 73: **💬 suggestion:** ‏What about inlining some parts?
```suggestion
  return {
    addShadowRoot: (shadowRoot: ShadowRoot) => {
      ...
    },
    removeShadowRoot: (shadowRoot: ShadowRoot) => {
      ...
    },
    stop: () => {
      shadowDomCallBacks.forEach(({ stop }) => stop())
    },
    flush: () => {
      shadowDomCallBacks.forEach(({ flush }) => flush())
    },
  }
```

packages/rum/src/domain/record/record.ts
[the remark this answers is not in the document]
bcaudan: `ShadowRootsController` looks good, we should still do something for `shadowDomCallBacks` though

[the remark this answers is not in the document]
ThibautGeriz: The `ShadowDomCallBacks` is an internal type only used by the `new Map<ShadowRoot, ShadowDomCallBacks>()`. What is the issue with it? What would you like to change?

packages/rum/src/domain/record/shadowDom.ts
ThibautGeriz: I find it a bit less readable but I am okay with it

packages/rum/src/domain/record/record.ts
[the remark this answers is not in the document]
bcaudan: cf top of the thread:

> - shadowDomCallBacks is quite confusing since it is used for various different concepts:
>   - shadowDomCallBacks = new Map<ShadowRoot, ShadowDomCallBacks>()
>   - interface ShadowDomCallBacks (declared both in record.ts and shadowDom.ts)
>   - shadowDomCallBacks: { shadowDomCreatedCallback, shadowDomRemovedCallback }
>   - bonus: ShadowDomCallBack = (shadowRoot: ShadowRoot) => void

the third one has been removed but we still have three different concepts with similar names.

IMO:

- `interface ShadowDomCallBacks` could be `ShadowRootController` since it exposes controls for a single shadow root
- `shadowDomCallBacks = new Map<ShadowRoot, ShadowDomCallBacks>()` would then be `controllerByShadowRoot = new Map<ShadowRoot, ShadowRootController>()`

For `ShadowDomCallBack = (shadowRoot: ShadowRoot) => void`, would it not more precise to name it `ShadowRootCallBack`?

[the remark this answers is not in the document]
ThibautGeriz: thank you for the explainations, The type is gone from the `record.ts` (unless I am missing something when grepping the code base)

> interface ShadowDomCallBacks could be ShadowRootController since it exposes controls for a single shadow root

_loosely held opinion:_ We would have have `ShadowRootController` & `ShadowRootsController`. Having a single `s` of difference make it less clear to me But I can rename it `ShadowRootCallBacks`. What do you think?


> For ShadowDomCallBack = (shadowRoot: ShadowRoot) => void, would it not more precise to name it ShadowRootCallBack?

Indeed it is.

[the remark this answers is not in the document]
bcaudan: > The type is gone from the record.ts (unless I am missing something when grepping the code base)

all good!

> We would have have ShadowRootController & ShadowRootsController. Having a single s of difference make it less clear to me But I can rename it ShadowRootCallBacks. What do you think?

IMO, the two components have similar responsibilities (flush, stop) so having a similar naming for it (Controller) could help the understanding.
For the single `s` of difference, I'd say it should not be too problematic as long are we are not mixing too much the two in the same context, cf https://github.com/DataDog/browser-sdk/pull/1787#discussion_r1055446995, inlining `shadowRootsController` could remove some occurrences.

bcaudan (approved):

PR #1799, opened 2022-10-27
✅ fix flaky test caused by an async scroll event

BenoitZugmeyer:
## Motivation

When the [viewports.spec tests](https://github.com/DataDog/browser-sdk/blob/62b88f934a5cd5c949435c6b5d2df51b79e864cf/packages/rum/src/domain/record/viewports.spec.ts#L8) are run, a "scroll" event is dispatched *after* they finish to run. This scroll event can impact test cases that are run directly after them, for example the [record.spec tests](https://github.com/DataDog/browser-sdk/blob/62b88f934a5cd5c949435c6b5d2df51b79e864cf/packages/rum/src/domain/record/record.spec.ts) which are doing many assertions on generated records.

This is reproductible with Jasmine seed 67707.

Note: I also observed a similar issue with the `visualViewport` resize event, but couldn't reproduce it anymore.

## Changes

Conditionnally wait for the scroll event to happen before ending the test.

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/viewports.spec.ts
+++ b/packages/rum/src/domain/record/viewports.spec.ts
@@ -1,4 +1,4 @@
-import { addEventListener, DOM_EVENT } from '@datadog/browser-core'
+import { addEventListener, DOM_EVENT, isIE } from '@datadog/browser-core'
 import { getScrollX, getScrollY } from './viewports'
 
 function isMobileSafari12() {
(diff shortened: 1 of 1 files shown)


liywjl (approved):

bcaudan (approved):
Nice catch!

Issue #1675, opened 2022-08-03
setUser info not sent with logs


hipitihop:
We use the RUM `setUser` and see correctly tagged RUM events including manual view changes, actions etc. However anything that goes via the `DD_LOGS.logger` fns like info, log, warn, do not include the user info. Is this an oversight that is being worked on ? or do we need to manually add the attributes e.g. `usr.id` and `user.name` to our logger payload ?

BenoitZugmeyer:
Hello @hipitihop,
To be consistend with RUM, we'll add a `setUser` API to Logs! In the meantime, you can use `DD_LOGS.addLoggerGlobalContext('usr', { ... })` to achieve the same result.

Bear in mind that the RUM and Logs SDKs are independent and you would have to set the user on both.

hipitihop:
> Hello @hipitihop,
> 
> To be consistend with RUM, we'll add a `setUser` API to Logs! In the meantime, you can use `DD_LOGS.addLoggerGlobalContext('usr', { ... })` to achieve the same result.
> 
> Bear in mind that the RUM and Logs SDKs are independent and you would have to set the user on both.

Hi @BenoitZugmeyer thanks. Does 14.7.x address this ? namely https://github.com/DataDog/browser-sdk/pull/1641 ?

BenoitZugmeyer:
No, implementing `setUser` on logs is still in our backlog.

yannickadam:
Hi!
This feature is available with SDK 14.24.0.
Feel free to open another issue if you experience any problem when using it.

PR #1801, opened 2022-10-27
✨ [RUMF-1409] Provide setUser and related functions for logs SDK

yannickadam:
## Motivation

Issue raised here: https://github.com/DataDog/browser-sdk/issues/1675
The RUM SDK provides convenient user-related functions. This change brings the same functions to the LOGS SDK.

## Changes

Added the following functions to the logs public API: 
setUser, 
setUserProperty, 
getUser,
removeUserProperty,
clearUser

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/user/user.ts
+++ b/packages/core/src/domain/user/user.ts
@@ -1,6 +1,6 @@
 import type { Context } from '../../tools/context'
 import { display } from '../../tools/display'
-import { deepClone, getType } from '../../tools/utils'
+import { assign, getType } from '../../tools/utils'
 import type { User } from './user.types'
 
 /**
@@ -9,7 +9,8 @@ import type { User } from './user.types'
  * https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#user-related-attributes
  */
 export function sanitizeUser(newUser: Context): Context {
-  const user = deepClone(newUser)
+  // We shallow clone only to prevent mutation of user data.
+  const user = assign({}, newUser)
   const keys = ['id', 'name', 'email']
   keys.forEach((key) => {
     if (key in user) {
--- a/packages/core/src/domain/user/user.ts
+++ b/packages/core/src/domain/user/user.ts
@@ -1,21 +1,18 @@
 import type { Context } from '../../tools/context'
-import { assign } from '../../tools/utils'
+import { deepClone } from '../../tools/utils'
 
 /**
  * Clone input data and ensure known user properties (id, name, email)
  * are strings, as defined here:
  * https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#user-related-attributes
  */
-export function sanitizeUser(newUser: Context) {
-  const shallowClonedUser = assign({}, newUser)
-  if ('id' in shallowClonedUser) {
-    shallowClonedUser.id = String(shallowClonedUser.id)
-  }
-  if ('name' in shallowClonedUser) {
-    shallowClonedUser.name = String(shallowClonedUser.name)
-  }
-  if ('email' in shallowClonedUser) {
-    shallowClonedUser.email = String(shallowClonedUser.email)
-  }
-  return shallowClonedUser
+export function sanitizeUser(newUser: Context): Context {
+  const user = deepClone(newUser)
+  const keys = ['id', 'name', 'email']
+  keys.forEach((key) => {
+    if (key in user) {
+      user[key] = String(user[key])
+    }
+  })
+  return user
 }
--- a/packages/core/src/domain/user/user.ts
+++ b/packages/core/src/domain/user/user.ts
@@ -1,5 +1,7 @@
 import type { Context } from '../../tools/context'
-import { deepClone } from '../../tools/utils'
+import { display } from '../../tools/display'
+import { deepClone, getType } from '../../tools/utils'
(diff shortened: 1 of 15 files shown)


liywjl (approved):
Nice one!

packages/core/src/domain/user/user.types.ts
```diff
@@ -0,0 +1,6 @@
+export interface User {
+  id?: string
+  email?: string
+  name?: string
+  [key: string]: unknown
+}
```
bcaudan, line 6: **❗ warning:** ‏as hinted by compatibility tests, this is a breaking change

packages/logs/src/domain/assembly.spec.ts
```diff
@@ -270,6 +277,52 @@ describe('startLogsAssembly', () => {
   })
 })
 
+describe('user management', () => {
```
bcaudan, line 280: **❗ warning:** ‏As you pointed out, we could have conflict between user context and global context, we should have a test to guarantee the behavior on that front

packages/logs/src/domain/assembly.ts
```diff
@@ -52,6 +54,10 @@ export function startLogsAssembly(
         messageContext
       )
 
+      if (!isEmptyObject(commonContext.user)) {
+        log.usr = shallowClone(commonContext.user)
```
bcaudan, line 58: **❓ question:** ‏what is your reasoning for using a shallow clone here?

packages/core/src/domain/user/user.ts
```diff
@@ -0,0 +1,18 @@
+import type { Context } from '../../tools/context'
+import { deepClone } from '../../tools/utils'
+
+/**
+ * Clone input data and ensure known user properties (id, name, email)
+ * are strings, as defined here:
+ * https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#user-related-attributes
+ */
+export function sanitizeUser(newUser: Context): Context {
+  const user = deepClone(newUser)
```
BenoitZugmeyer, line 10: **❓ question:** ‏any reason to deepClone here instead of shallow clone?

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -125,6 +129,25 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
     getInitConfiguration: monitor(() => getInitConfigurationStrategy()),
 
     getInternalContext: monitor((startTime?: number | undefined) => getInternalContextStrategy(startTime)),
+
+    setUser: monitor((newUser: User) => {
+      if (typeof newUser !== 'object' || !newUser) {
+        display.error('Unsupported user:', newUser)
+      } else {
```
amortemousque, line 136: **💬 suggestion:** ‏Maybe this check can also be mutualized in the `code/src/domain/user` module?

packages/core/src/domain/user/user.ts
yannickadam: It's possible to receive nested objects. It should not happen often, but it's a precaution.

liywjl: Example:
```
const obj = { foo: 'bar' }
const nested = { biz: 'buz' };
obj.nested = nested

console.log(obj) // { foo: 'bar', nested: { biz: 'buz' } }

nested.biz = 'break it'

console.log(obj) // { foo: 'bar', nested: { biz: 'break it' } }
```

liywjl (approved):
LGTM

packages/core/src/domain/user/user.ts
BenoitZugmeyer: Yes, since the Context Manager [already deep clones it when getting/setting the user](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/contextManager.ts#L27-L32), this seems unnecessary. See @bcaudan remark here: https://github.com/DataDog/browser-sdk/pull/1641#discussion_r932438621

buraizu:
Thanks, created DOCS-4334 to track as an editorial review

amortemousque (approved):

BenoitZugmeyer (approved):

packages/logs/README.md
```diff
@@ -572,6 +572,94 @@ window.DD_LOGS && DD_LOGS.getGlobalContext() // => {}
...
+window.DD_LOGS && DD_LOGS.removeUserProperty('type')
+
+window.DD_LOGS && DD_LOGS.getUser() // => {id: '1234', name: 'John Doe', email: 'john@doe.com'}
+
+window.DD_LOGS && DD_LOGS.clearUser()
+
+window.DD_LOGS && DD_LOGS.getUser() // => {}
+```
+
+**Note**: The `window.DD_LOGS` check is used to prevent issues if a loading failure occurs with the SDK.
```
urseberry, line 661:
```suggestion
**Note**: The `window.DD_LOGS` check prevents issues when a loading failure occurs with the SDK.
```

urseberry (approved):
I added a wording suggestion that changes a sentence to active voice. Feel free to ignore it if it changes the meaning of the note.

packages/logs/README.md
yannickadam: The wording was used in 8 spots - all of them have been updated.

PR #1805, opened 2022-10-31
♻️ [RUMF-1424] factorize page exit logic

BenoitZugmeyer:
## Motivation

Factorize page exit logic between Replay segment collection and RUM/Logs events.

This PR superseds https://github.com/DataDog/browser-sdk/pull/1788

## Changes

Introduce a new `PageExit` observable to have an event when the page is exiting.
Use this new object:
* in Batch to flush the pending data
* in track Views to emit a last View update and end the view (when the page is unloading)
* in SegmentCollection to flush any pending segment

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -75,7 +75,6 @@ export function trackClickActions(
     history.reset()
   })
 
-  lifeCycle.subscribe(LifeCycleEventType.BEFORE_UNLOAD, stopClickChain)
   lifeCycle.subscribe(LifeCycleEventType.VIEW_ENDED, stopClickChain)
```
BenoitZugmeyer, line 78: Note: the `BEFORE_UNLOAD` line have been removed intentionally here: because `VIEW_ENDED` is also triggered on `BEFORE_UNLOAD`, having both was unnecessary.

packages/core/src/browser/pageExitObservable.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  it('notifies multiple times', () => {
+    window.dispatchEvent(createNewEvent('beforeunload'))
+    window.dispatchEvent(createNewEvent('beforeunload'))
+    emulatePageVisibilityChange('hidden')
+
+    expect(onExitSpy).toHaveBeenCalledTimes(3)
+  })
+
+  it('does notify when the page becomes visible', () => {
```
bcaudan, line 40:
```suggestion
  it('does not notify when the page becomes visible', () => {
```

packages/rum/src/domain/segmentCollection/segmentCollection.spec.ts
```diff
@@ -124,27 +117,11 @@ describe('startSegmentCollection', () => {
...
-      setPageVisibility('hidden')
-      const { addRecord, eventEmitter, sendCurrentSegment, worker, sendSpy } = startSegmentCollection(CONTEXT)
-      addRecord(RECORD)
-      eventEmitter.dispatchEvent(createNewEvent(DOM_EVENT.VISIBILITY_CHANGE))
-
-      worker.processAllMessages()
-      expect(sendSpy).toHaveBeenCalledTimes(1)
+      const { lifeCycle, sendCurrentSegment } = startSegmentCollection(CONTEXT)
+      lifeCycle.notify(LifeCycleEventType.PAGE_EXITED, { isUnloading: false })
       expect(sendCurrentSegment().creation_reason).toBe('visibility_hidden')
```
bcaudan, line 122: **💬 suggestion:** ‏here, we are deducing that a page exit that is not unloading is a visibility hidden, maybe we could use an exit reason instead of a boolean

packages/rum/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -89,19 +88,11 @@ export function doStartSegmentCollection(
...
-
-  const { stop: unsubscribeVisibilityChange } = addEventListener(
-    emitter,
-    DOM_EVENT.VISIBILITY_CHANGE,
-    () => {
-      if (document.visibilityState === 'hidden') {
-        flushSegment('visibility_hidden')
-      }
-    },
-    { capture: true }
```
bcaudan, line 104: **❓ question:** ‏should we keep using capture?

packages/rum/src/domain/segmentCollection/segmentCollection.spec.ts
```diff
@@ -124,27 +117,11 @@ describe('startSegmentCollection', () => {
...
     it('flushes segment when the page become hidden', () => {
-      setPageVisibility('hidden')
-      const { addRecord, eventEmitter, sendCurrentSegment, worker, sendSpy } = startSegmentCollection(CONTEXT)
-      addRecord(RECORD)
-      eventEmitter.dispatchEvent(createNewEvent(DOM_EVENT.VISIBILITY_CHANGE))
-
-      worker.processAllMessages()
-      expect(sendSpy).toHaveBeenCalledTimes(1)
+      const { lifeCycle, sendCurrentSegment } = startSegmentCollection(CONTEXT)
+      lifeCycle.notify(LifeCycleEventType.PAGE_EXITED, { isUnloading: false })
```
bcaudan, line 121: **💬 suggestion:** update test name to something like `flushes segment on page exit`‏

packages/core/src/browser/pageExitObservable.ts
```diff
@@ -0,0 +1,35 @@
+import { Observable } from '../tools/observable'
+import { addEventListener, DOM_EVENT } from '../tools/utils'
+
+export interface PageExitEvent {
+  isUnloading: boolean
+}
+
+export function createPageExitObservable(): Observable<PageExitEvent> {
```
amortemousque, line 8: **👏 praise:** ‏Love it !

packages/rum-core/src/boot/startRum.ts
```diff
@@ -61,7 +62,11 @@ export function startRum(
     lifeCycle.notify(LifeCycleEventType.RAW_ERROR_COLLECTED, { error })
   }
   if (!canUseEventBridge()) {
-    startRumBatch(configuration, lifeCycle, telemetry.observable, reportError)
+    const pageExitObservable = createPageExitObservable()
+    pageExitObservable.subscribe((event) => {
+      lifeCycle.notify(LifeCycleEventType.PAGE_EXITED, event)
+    })
```
amortemousque, line 68: **❓ question:** ‏Do we have to create a lifeCycle event if we have the observable?

liywjl (approved):
LGTM

packages/rum/src/domain/segmentCollection/segmentCollection.ts
BenoitZugmeyer: Ah, good catch. I think that it should not have a big impact, but we should use `capture` most of the time to avoid being impacted by the app stopping the event propagation. In the long run I'd be in favor of using capture by default for all listeners.

packages/rum/src/domain/segmentCollection/segmentCollection.spec.ts
BenoitZugmeyer: In the previous PR you [suggested](https://github.com/DataDog/browser-sdk/pull/1788#discussion_r1004680648) to keep a test for `{ isUnloading: false }`. Since there is also a test for `{ isUnloading: true }`, we should have more specific names than `flushes segment on page exit`.


What do you think about:
* `flushes segment on page exit because it gets hidden`
* `flushes segment on page exit because it is unloading`

packages/rum-core/src/boot/startRum.ts
BenoitZugmeyer: I'm not quite against it, but we introduced the `LifeCycle` specifically to avoid passing a bunch of individual `Observable` all around. Using the `LifeCycle` allows to have a central event pub/sub in RUM and simplifies writing tests. Using the `Observable` would be a step backward from this design. I'd prefer to keep using the `LifeCycle` for now. @bcaudan WDYT?

amortemousque: I do get the convenience of the LifeCycle but then could we also use it in `startRumBatch`. It's not pretty straightforward to understand why we sometimes pass down the observable and not other times. Now that both RUM and Logs have a lifecycle we could add the PAGE_EXITED event to Logs and use it in the batch. Wdyt?

BenoitZugmeyer: FTR: we agreed that this topic requires further discussion. I'll merge this as-is and we'll continue the discussion around an RFC

packages/rum/src/domain/segmentCollection/segmentCollection.spec.ts
bcaudan: sounds good to me!

packages/rum/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -91,7 +91,7 @@ export function doStartSegmentCollection(
   const { unsubscribe: unsubscribePageExited } = lifeCycle.subscribe(
     LifeCycleEventType.PAGE_EXITED,
     (pageExitEvent) => {
-      flushSegment(pageExitEvent.isUnloading ? 'before_unload' : 'visibility_hidden')
+      flushSegment(pageExitEvent.reason === PageExitReason.UNLOADING ? 'before_unload' : 'visibility_hidden')
```
bcaudan, line 94: **🥜 nitpick:** it ‏could be nice to use page exit reason as a creation reason
```ts
flushSegment(pageExitEvent.reason)
```

bcaudan (approved):

PR #1809, opened 2022-11-03
:sparkles: add cookie handling override config (i.e. getCookie, setCookie) to support Electron

bryzaguy:
## Motivation

This will allow instrumenting Electron apps.

## Changes

Add getCookie and setCookie to initializer configuration 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/cookie.ts
+++ b/packages/core/src/browser/cookie.ts
@@ -3,13 +3,28 @@ import { findCommaSeparatedValue, generateUUID, ONE_SECOND } from '../tools/util
 
 export const COOKIE_ACCESS_DELAY = ONE_SECOND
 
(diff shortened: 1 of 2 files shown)


liywjl:
Hello @bryzaguy 

Thank you for opening this PR. We currently do not officially support Electron (yet), although we have explored the idea https://github.com/DataDog/browser-sdk/pull/1393  

I noticed that in your changes, you were seeking to expose arguments passed into `setCookie` to an external function. It would be interesting to investigate how to allow users to connect to their own storage via getters and setters. 

However, since cookie sessions can be related to sensitive information, such as billing, we need to be careful about what we send over.

I have created a ticket internally for us to export this.

PR #1810, opened 2022-11-07
⚗️ [RUMF-1337] add `fetch_duration` experiment to resolve incorrect fetch duration

liywjl:
## Motivation

It appears that `.then` is not a good measurement for when a `fetch` request has completed. This means that our own calculate for `fetch` duration have been under reporting the total time of the request.

## Why?

It appears that `fetch(xxx).then(xx)` is triggered when a  response is received. That does not include the response payload being downloaded, which should be included in the duration estimation.

## Solution

We therefore need to instrument the response and listen to it's stream to only calculate duration after the payload has been downloaded.

We should do this behind `fetch_duration` experiment flag.

## Other impacts

- This should increase the likelihood of finding [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) for `fetch` requests
- Users will no longer be able to use `Response.body.cancel()` or `Response.body.getReader().cancel()` and expect the stream no longer to stop (because we clone the `Response.body` stream)

## Changes

- Refactor and mutualise `readBytesFromStream`
- Add function `waitForResponseToComplete` in `requestCollection` to generate `REQUEST_COMPLETED` after stream is finished
- Replace `context.duration` with `context.resolveDuration` in `fetchObservable`
- add `computeResponseDurationInfo` in `resourceCollection` to get diff details between `resolveDuration` and new `duration` value
- Hide changes behind `fetch_duration` feature flag

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/browser/fetchObservable.ts
```diff
@@ -93,25 +94,58 @@ function afterSend(
...
+              return pump()
+            })
+          }
+        },
+      })
+    } else {
+      context.duration = elapsed(context.startClocks.timeStamp, timeStampNow())
+      reportFetch(context, response)
     }
   }
```
BenoitZugmeyer, line 141: 💬 **suggestion**: ‏extract this logic in a function that you would use like this:

```ts
waitForResponseToFinish(response).then(() => {
   context.duration = elapsed(context.startClocks.timeStamp, timeStampNow())
   reportFetch(context, response)
})
```


**❓ question:** ‏in a previous PR, I remember you used to factorize this code with the [logs network error collection](https://github.com/DataDog/browser-sdk/blob/0ab07cb70ac32cdd92677e80dd2942e7770e2cb9/packages/logs/src/domain/logsCollection/networkError/networkErrorCollection.ts#L142). Would it possible to do the same here?

packages/core/src/browser/fetchObservable.ts
```diff
@@ -93,25 +94,58 @@ function afterSend(
...
+    const responseClone = response.clone()
+    const reader = responseClone.body?.getReader()
+
+    if (reader && ReadableStream) {
+      new ReadableStream({
+        start(controller) {
+          return pump()
+
+          function pump(): Promise<undefined> {
+            return (reader as ReadableStreamDefaultReader<Uint8Array>).read().then(({ done }) => {
```
BenoitZugmeyer, line 125: 💬 **suggestion**: You should handle the case where `.read()` returns a rejected promise (for example, when the request has been cancelled via an `AbortController`). Also `monitor()` those callbacks.

test/e2e/scenario/logs.scenario.ts
```diff
@@ -104,7 +101,7 @@ describe('logs', () => {
         // When reading the request, chunks length are probably not aligning perfectly with the
         // response length limit, so it sends few more bytes than necessary. Add a margin of error
         // to verify that it's still close to the expected limit.
-        DEFAULT_REQUEST_ERROR_RESPONSE_LENGTH_LIMIT * 2
+        DEFAULT_REQUEST_ERROR_RESPONSE_LENGTH_LIMIT * 4
```
BenoitZugmeyer, line 104: **🔨 warning:** this assertion is now pointless:

* `getLargeResponseWroteSize` is [around 100K](https://github.com/DataDog/browser-sdk/blob/0ab07cb70ac32cdd92677e80dd2942e7770e2cb9/test/e2e/lib/framework/serverApps/mock.ts#L7)
* `DEFAULT_REQUEST_ERROR_RESPONSE_LENGTH_LIMIT * 4` is [around 128K](https://github.com/DataDog/browser-sdk/blob/0ab07cb70ac32cdd92677e80dd2942e7770e2cb9/packages/logs/src/domain/configuration.ts#L35)

So this test does not ensure that we "read only the first bytes of the response" when doing a fetch.

test/e2e/scenario/logs.scenario.ts
```diff
@@ -78,18 +78,15 @@ describe('logs', () => {
...
-          },
-          (error) => console.log(error)
-        )
+        const controller = new AbortController()
+        const signal = controller.signal
+
+        setTimeout(() => {
+          controller.abort()
+          setTimeout(() => done(undefined), 1000)
+        }, 2000)
```
BenoitZugmeyer, line 87: **💭 thought:** ‏those timeout usages seems strange to me, we should talk about it

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -196,6 +196,50 @@ describe('rum resources', () => {
       })
   })
 
+  createTest('track redirect fetch timings')
+    .withRum()
+    .run(async ({ serverEvents }) => {
+      await browserExecuteAsync((done) => {
+        fetch(`/redirect?duration=${200}`).then(
```
BenoitZugmeyer, line 203: **🥜 nitpick:** ‏
```suggestion
        fetch(`/redirect?duration=200`).then(
```

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -196,6 +196,50 @@ describe('rum resources', () => {
...
+      expectToHaveValidTimings(resourceEvent)
+      expect(resourceEvent.resource.redirect).not.toBeUndefined()
+      expect(resourceEvent.resource.redirect!.duration).toBeGreaterThan(0)
+    })
+
+  createTest('track dns lookup fetch timings')
+    .withRum()
+    .run(async ({ serverEvents }) => {
+      await browserExecuteAsync((done) => {
+        fetch(`/redirect?duration=${200}`).then(
```
BenoitZugmeyer, line 225: **🥜 nitpick:** ‏
```suggestion
        fetch(`/redirect?duration=200`).then(
```

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -60,6 +61,14 @@ function processRequest(
   const type = request.type === RequestType.XHR ? ResourceType.XHR : ResourceType.FETCH
 
   const matchingTiming = matchRequestTiming(request)
+  if (!matchingTiming) {
+    addTelemetryDebug('Missing PerformanceResourceTiming', {
+      duration: request.duration,
+      resolveDuration: request.resolveDuration,
+      status: request.status,
+      requestType: request.type === RequestType.XHR ? 'xhr' : 'fetch',
+    })
```
BenoitZugmeyer, line 70: **💬 suggestion:** ‏this might yield a *lot* of logs. Maybe log it only the first time we got a missing timing?

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -196,6 +196,50 @@ describe('rum resources', () => {
...
+      expect(resourceEvent).not.toBeUndefined()
+      expect(resourceEvent.resource.method).toEqual('GET')
+      expect(resourceEvent.resource.status_code).toEqual(200)
+      expect(resourceEvent._dd.resolveDuration).toBeGreaterThan(0)
+      expectToHaveValidTimings(resourceEvent)
+      expect(resourceEvent.resource.redirect).not.toBeUndefined()
+      expect(resourceEvent.resource.redirect!.duration).toBeGreaterThan(0)
+    })
+
+  createTest('track dns lookup fetch timings')
```
BenoitZugmeyer, line 221: **❓ question:** ‏ ‏this test is an exact copy of the one before. Typo?

packages/core/src/browser/fetchObservable.ts
liywjl: we already use `monitor` on `responsePromise.then(monitor(onFulfilled), monitor(onRejected))`, does it make sense to have it here too?

BenoitZugmeyer: Ah that's right the promise is handled somewhere else. Nevermind my comment then!

test/e2e/scenario/logs.scenario.ts
liywjl: This wasn't ideal I agree. I was trying to call `controller.abort()` and `done()` on `.then` but found that no entry would be present in `serverEvents`.

This workaround gets around the issue. But I agree it isnt ideal

packages/core/src/browser/fetchObservable.ts
```diff
@@ -93,25 +94,63 @@ function afterSend(
...
+
+  responsePromise.then(
+    monitor((response) =>
+      waitForResponseToFinish(response).then(
+        (resolveDuration: Duration) => reportFetch(response, resolveDuration),
+        () => reportFetch(response)
+      )
+    ),
+    monitor(reportFetch)
+  )
```
bcaudan, line 155: **💭 thought:** Since we are already reading responses for logs network error collection, there are probably some thoughts to have on that front.

- Do we really need to wait for response to finish for all fetch instrumentation usages? 
- Should we mutualise some things with our current [response reading code](https://github.com/DataDog/browser-sdk/blob/main/packages/logs/src/domain/logsCollection/networkError/networkErrorCollection.ts#L201)?

test/e2e/scenario/logs.scenario.ts
bcaudan, line 87: **💭 thought:** ‏Previous comment was useful to understand why we were canceling the response.
No idea what was the issue with your changes and what you are trying to do.
Could you clarify that?

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -41,6 +41,7 @@ export interface RawRumResourceEvent {
     span_id?: string // not available for initial document tracing
     rule_psr?: number
     discarded: boolean
+    resolveDuration?: ServerDuration
```
bcaudan, line 44: **❓ question:** ‏is it only for while we are monitoring the change or do we plan to keep this property?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -60,6 +63,17 @@ function processRequest(
   const type = request.type === RequestType.XHR ? ResourceType.XHR : ResourceType.FETCH
 
   const matchingTiming = matchRequestTiming(request)
+
+  if (!hasLoggedMissingResource && !matchingTiming) {
+    addTelemetryDebug('Missing PerformanceResourceTiming', {
```
bcaudan, line 68: **❓ question:** ‏matchTiming can be undefined for several reasons, what do you want to monitor exactly?‏

packages/core/src/browser/fetchObservable.ts
liywjl: > Should we mutualise some things with our current response reading code?
Yes agreed

> Do we really need to wait for response to finish for all fetch instrumentation usages?
I agree this isn't ideal, however if we wanted to finish earlier we cannot determine a precise duration. Happy to sync up about this issue

packages/core/src/browser/fetchObservable.ts
```diff
@@ -96,22 +96,18 @@ function afterSend(
   const reportFetch = (response: Response | Error) => {
     const context = startContext as unknown as FetchCompleteContext
     context.state = 'complete'
-    context.duration = elapsed(context.startClocks.timeStamp, timeStampNow())
```
liywjl, line 99: We know duration is incorrect here so it makes sense for us to remove it

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -52,6 +60,22 @@ export function startResourceCollection(
   })
 }
 
+function waitForResponseToFinish(request: RequestCompleteEvent, callback: (duration: Duration) => void) {
+  const duration = request.duration || elapsed(request.startClocks.timeStamp, timeStampNow())
```
liywjl, line 64: for `XHR` request.duration should be defined, else we should create one as a fallback

test/e2e/scenario/logs.scenario.ts
```diff
@@ -74,22 +74,15 @@ describe('logs', () => {
...
-        fetch('/throw-large-response').then(
-          (response) => {
-            // The body stream needs to be cancelled, else the browser will still download the whole
-            // response even if it is unused.
-            response
-              .body!.getReader()
-              .cancel()
-              .catch((error) => console.log(error))
+        fetch('/throw-large-response')
+          .then((response: Response) => response.clone().blob())
```
BenoitZugmeyer, line 82: **🔨 warning:** ‏the `.clone()` here looks fishy. The user shouldn't need to clone their response... let's see why we need it

packages/core/src/tools/readBytesFromStream.ts
```diff
@@ -0,0 +1,75 @@
+import { monitor } from './monitor'
+import { noop } from './utils'
+
+type Options = {
+  limit: number
```
amortemousque, line 5: **🥜 nitpick:** ‏ ‏Maybe the naming could be more explicit like `bytesLimit`

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -39,6 +39,7 @@ export interface RequestCompleteEvent {
   status: number
   responseType?: string
   startClocks: ClocksState
+  resolveDuration?: Duration
   duration: Duration
```
amortemousque, line 43: **❓ question:** ‏Should it be optional?
```suggestion
  duration?: Duration
```

packages/core/src/browser/fetchObservable.ts
```diff
@@ -96,22 +97,19 @@ function afterSend(
...
-
+    context.resolveDuration = elapsed(startContext.startClocks.timeStamp, timeStampNow())
```
amortemousque, line 100: **❓ question:** ‏Shouldn't it be in `FetchCompleteContext` instead of the `duration` attribute?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -37,7 +41,12 @@ export function startResourceCollection(
   sessionManager: RumSessionManager
 ) {
   lifeCycle.subscribe(LifeCycleEventType.REQUEST_COMPLETED, (request: RequestCompleteEvent) => {
-    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processRequest(request, configuration, sessionManager))
+    waitForResponseToFinish(request, () =>
```
amortemousque, line 44: **❓ question:** ‏Why not waiting for response to finish inside the `requestCollection`?

packages/core/src/browser/fetchObservable.ts
liywjl: I am not sure I follow, do you mean to say that the variable should probably be defined in `beforeSend` where we define the `FetchCompleteContext`?

If so, I do not think this is the right place to put it. 

`reportFetch` gets called only after the `fetch` `.then` has been triggered. Moving this logic to `FetchCompleteContext` (aka `beforeSend`) would mean we incorrectly calculating `resolveDuration`

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
liywjl: Good point, I initially put it here as this was where `PerformanceEntry` logic resided. But bar that there was no real reason for it to be here.

It probably is better for us to move this logic into `requestCollection` as it is closer to the instrumentation and thus will give us a better `duration` estimation

packages/rum-core/src/domain/requestCollection.ts
liywjl: Good point, I will extend this type to make sure we reflect when `RequestCompleteEvent` contains and does not contain duration

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -66,6 +102,13 @@ function processRequest(
   const tracingInfo = computeRequestTracingInfo(request, configuration)
   const indexingInfo = computeIndexingInfo(sessionManager, startClocks)
 
+  const durationDiff = request.resolveDuration
+    ? toServerDuration((request.duration - request.resolveDuration) as Duration)
+    : undefined
+  const durationPercentageDiff = durationDiff
+    ? Math.round((durationDiff / toServerDuration(request.duration)) * 100)
+    : undefined
+
```
BenoitZugmeyer, line 111: **💭 thought:** ‏I'm not sure we need those, as we can do arithmetics with measurement facets directly in Datadog


**💬 suggestion:** ‏if we really need those, I would do something like this:
```ts
let durationDetails
if (request.resolveDuration) {
  durationDetails = {
    resolveDuration: ...,
    durationDiff: ...,
    durationPercentageDiff: ...
  }
}
```

test/e2e/lib/framework/createTest.ts
```diff
@@ -20,7 +20,7 @@ const DEFAULT_RUM_CONFIGURATION = {
   trackLongTasks: true,
   telemetrySampleRate: 100,
   telemetryConfigurationSampleRate: 100,
-  enableExperimentalFeatures: [],
+  enableExperimentalFeatures: ['fetch_duration'],
```
BenoitZugmeyer, line 23: **💬 suggestion:** ‏I'm not sure we want to enable it for all tests, as we want to ensure that the SDK works without this feature too.
* If you added a test that requires this flag to work, you can use `withRum({ enableExperimentalFeatures: ['fetch_duration'] })`
* if you want to make sure an existing test still works with this flag, you can duplicate it and use `withRum({ enableExperimentalFeatures: ['fetch_duration'] })` on one of them.

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
liywjl: > we can do arithmetics with measurement facets

This was done based on @bcaudan 's suggestion. Happy to remove it if you insist, I personally think it's fine and do plan on deleting this section in a future PR, so I do not have super strong feelings either way.

> I would do something like ...

I agree its not very nice looking. I will add `nullish coalescing` and the `if` statement to make it a little nicer. I do plan on removing this section in a future PR shortly

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
amortemousque: If we put it in `requestCollection` and wait For Response To Finish  before triggering `REQUEST_COMPLETED` it will also affect `waitPageActivityEnd`. Conceptually it would make sens to me but might have broader impact. @BenoitZugmeyer wdyt?

BenoitZugmeyer: Makes sense to me too! The UI usually react when the response finishes to load

BenoitZugmeyer (approved):

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -119,23 +127,25 @@ export function trackFetch(lifeCycle: LifeCycle, configuration: RumConfiguration
...
-          startClocks: context.startClocks,
-          status: context.status,
-          traceId: context.traceId,
-          traceSampled: context.traceSampled,
-          type: RequestType.FETCH,
-          url: context.url,
-          response: context.response,
-          init: context.init,
-          input: context.input,
+        waitForResponseToFinish(context, (duration: Duration) => {
```
bcaudan, line 130: **🥜 nitpick:** ‏We could maybe make clearer the lifecycle difference with xhr by renaming `complete` occurrences for fetch into something like `resolve`, wdyt?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -66,6 +66,13 @@ function processRequest(
   const tracingInfo = computeRequestTracingInfo(request, configuration)
   const indexingInfo = computeIndexingInfo(sessionManager, startClocks)
 
+  let durationDiff
+  let durationPercentageDiff
+  if (request.resolveDuration) {
+    durationDiff = toServerDuration((request.duration - request.resolveDuration) as Duration)
+    durationPercentageDiff = Math.round((durationDiff / toServerDuration(request.duration)) * 100)
+  }
```
bcaudan, line 74: **🥜 nitpick:** ‏what about extracting all these infos into a `computeXXXInfo` like other parts of this function?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.spec.ts
```diff
@@ -135,6 +139,9 @@ describe('resourceCollection', () => {
       type: RumEventType.RESOURCE,
       _dd: {
         discarded: false,
+        resolveDuration: (50 * 1e6) as ServerDuration,
+        durationDiff: 50000000 as ServerDuration,
```
bcaudan, line 143: **💬 suggestion:** ‏for consistency sake
```suggestion
        resolveDuration: (50 * 1e6) as ServerDuration,
        durationDiff: (50 * 1e6) as ServerDuration,
```

packages/core/src/tools/readBytesFromStream.spec.ts
```diff
@@ -0,0 +1,93 @@
+import { isIE } from './browserDetection'
+import { readBytesFromStream } from './readBytesFromStream'
+
+describe('readBytesFromStream', () => {
```
bcaudan, line 4: **💭 thought:** ‏some tests in `networkErrorCollection.spec` could be more suited here now

test/e2e/scenario/logs.scenario.ts
```diff
@@ -74,22 +74,15 @@ describe('logs', () => {
...
-        fetch('/throw-large-response').then(
-          (response) => {
-            // The body stream needs to be cancelled, else the browser will still download the whole
-            // response even if it is unused.
-            response
-              .body!.getReader()
-              .cancel()
-              .catch((error) => console.log(error))
+        fetch('/throw-large-response')
+          .then((response: Response) => response.blob())
```
bcaudan, line 82: **💬 suggestion:** ‏adding a comment to explain what is your intent with `response.blob()` could be helpful.

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -194,8 +194,47 @@ describe('rum resources', () => {
...
+
+          fetch('/large-response', { signal }).then(() => {
+            controller.abort()
+            done(undefined)
+          }, console.log)
+        })
+
+        await flushEvents()
+        expect(servers.base.app.getLargeResponseWroteSize()).toBeLessThan(LARGE_RESPONSE_MIN_BYTE_SIZE)
+      })
```
bcaudan, line 214: **❓ question:** is this behavior related to SDK code?

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -148,3 +158,23 @@ function getNextRequestIndex() {
...
+        },
+        {
+          bytesLimit: Number.POSITIVE_INFINITY,
+          collectStreamBody: false,
+        }
+      )
+      return
+    }
+  }
+  callback(elapsed(context.startClocks.timeStamp, timeStampNow()))
```
bcaudan, line 179: **💭 thought:** ‏would it worth having tests for null response/body behaviors?

packages/core/src/tools/readBytesFromStream.spec.ts
bcaudan: Unit tests in `networkErrorCollection.spec` should be specific to network collection logic and tests in `readBytesFromStream.spec` should be specific to byte read logic.
By having a quick look `does not yield an unhandled rejection error if the cancel promise is rejected` seems to be misplaced to me. is it the case? and if yes, could other ones be misplaced as well?

liywjl: All tests under `describe('computeFetchResponseText'` focus on testing logic from `readBytesFromStream` (especially as we do not test our fallback cases). 

Are you suggesting we remove all of them and add the missing ones to `readBytesFromStream.spec`?

bcaudan: I'd like to keep the same level of coverage but have the tests as close as possible of the logic they are testing.
Since you extracted the logic, you should extract and adjust the corresponding tests as well.
Does it make sense?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -78,6 +78,8 @@ describe('logs', () => {
     .withLogs({ forwardErrorsToLogs: true })
     .run(async ({ serverEvents, baseUrl, servers }) => {
       await browserExecuteAsync((done) => {
+        // This example formerly used "response.body!.getReader().cancel()" to cancel the stream midway.
+        // However, as we not longer support ".cancel()" in streams and thus why we use "response.blob()".
```
bcaudan, line 82: **❓ question:** ‏
does using blob cancel the stream? 
if not, why are we using `response.blob` and not calling `done` directly after the fetch resolve?
any reason to use `blob` and not `json` or `text` instead?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -183,6 +178,19 @@ function computeEntryTracingInfo(entry: RumPerformanceResourceTiming, configurat
...
+  return {
+    durationDiff,
+    durationPercentageDiff,
+  }
```
bcaudan, line 191: **💬 suggestion:** ‏what about encapsulating all the response duration info logic/format in this method and return directly:
```ts
return {
  _dd: {
    resolveDuration,
    durationDiff,
    durationPercentageDiff
  }
}
```
it is what we are doing on other methods

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -159,7 +159,7 @@ function getNextRequestIndex() {
   return result
 }
 
-function waitForResponseToFinish(context: RumFetchCompleteContext, callback: (duration: Duration) => void) {
+function waitForResponseToFinish(context: RumFetchResolveContext, callback: (duration: Duration) => void) {
```
bcaudan, line 162: **💬 suggestion:** `waitForResponseToComplete` would avoid to use two terms for the same concept‏

packages/logs/src/domain/logsCollection/networkError/networkErrorCollection.ts
```diff
@@ -27,12 +27,12 @@ export function startNetworkErrorCollection(configuration: LogsConfiguration, li
...
   })
   const fetchSubscription = initFetchObservable().subscribe((context) => {
-    if (context.state === 'complete') {
+    if (context.state === 'resolve') {
       handleCompleteRequest(RequestType.FETCH, context)
     }
   })
 
-  function handleCompleteRequest(type: RequestType, request: XhrCompleteContext | FetchCompleteContext) {
+  function handleCompleteRequest(type: RequestType, request: XhrCompleteContext | FetchResolveContext) {
```
bcaudan, line 35: **💬 suggestion:** ‏`handleCompleteRequest` is not accurate anymore, maybe something like `handleResponse` would be enough

packages/rum-core/src/domain/requestCollection.spec.ts
```diff
@@ -100,6 +100,33 @@ describe('collect fetch', () => {
     })
   })
 
+  describe('with feature flag fetch_duration', () => {
+    beforeEach(() => updateExperimentalFeatures(['fetch_duration']))
+    afterEach(() => resetExperimentalFeatures())
+
+    it('should still function if response is defined', (done) => {
```
bcaudan, line 107: **💬 suggestion:** ‏`still function` is quite vague, looking at other tests something like `should notify when response is defined` would be more consistent

test/e2e/scenario/logs.scenario.ts
liywjl: > does using blob cancel the stream?

Good point, no. I will update the comment.

> why are we using response.blob and not calling done directly after the fetch resolve?

I initially had issues getting this condition to work because we were aborting the call too early to be able to fetch the first couple bytes.

I assumed calling `done` would also abort the stream too early for us to collect the bytes, but that does not seem to be the case. I will therefore not use any response methods (`blob`,`json` and `text`) and simply use `done`

> any reason to use blob and not json or text instead?

No, we do not care about the contents of the response (also we are not going to use any of these methods).

bcaudan (approved):

PR #1816, opened 2022-11-09
Adjust RUM parameter sessionReplaySampleRate to default to 0

ssmiller25:
## Motivation

The Browser SDK, by default, sets `sessionReplaySampleRate` to 100.  However, unless a user **also** goes through the instructions in <https://docs.datadoghq.com/real_user_monitoring/session_replay/>, the result is sending data that cannot be used, but will be billed.  Suggesting to change the default to 0, with documentation to actually set this parameter if a user wishes to enable Session Reply.

## Changes

Parameter `sessionReplaySampleRate` defaults to 0

## Testing

- [ ] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


bcaudan:
Hi @ssmiller25,

Thanks for your suggestion, we are aware of this behavior and since this change would be a breaking change, we are already planning to do it in a future major version.
No ETA on the next major for now.

PR #1824, opened 2022-11-17
✨ Auto Flush for SDK extension

liywjl:
## Motivation

Add Auto Flush checkbox for SDK extensions

Screenshot 2022-11-17 at 14 50 26

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/panel/hooks/usePolling.ts
```diff
@@ -0,0 +1,16 @@
...
+  useEffect(() => {
+    if (condition && !autoFlushIntervalId) {
+      autoFlushIntervalId = setInterval(callback, 5000)
+    }
+    if (!condition && autoFlushIntervalId) {
+      clearInterval(autoFlushIntervalId)
+      autoFlushIntervalId = undefined
+    }
+  }, [condition])
+}
```
BenoitZugmeyer, line 16: That's a bit frightening :D could you use something like [this](https://usehooks-ts.com/react-hook/use-interval) instead?

developer-extension/src/background/store.ts
```diff
@@ -6,6 +6,7 @@ export const store: Store = {
   useDevBundles: false,
   useRumSlim: false,
   blockIntakeRequests: false,
+  autoFlush: true,
```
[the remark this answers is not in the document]
BenoitZugmeyer: no, it's a centralized store I used at the beginning when the extension was an icon next to the URL bar. Now that it is a devtools extension, it makes more sense to not use a centralized store, but keep each tab / devtools instance separated.

developer-extension/src/panel/components/actionsBar.tsx
```diff
@@ -1,45 +1,56 @@
...
+import React, { useEffect, useState } from 'react'
 import { evalInWindow } from '../evalInWindow'
 import { useStore } from '../hooks/useStore'
 
 export function ActionsBar() {
   const [{ useDevBundles, useRumSlim, devServerStatus, blockIntakeRequests }, setStore] = useStore()
+  const [autoFlush, setAutoFlush] = useState<boolean>(true)
+
+  useEffect(flushEvents, [])
+  useInterval(flushEvents, 5000)
```
BenoitZugmeyer, line 12: **🔨 warning:** ‏disabling auto flush won't work, as we don't check for the `autoFlush` variable


**🔨 warning:** ‏usehooks-ts is not compatible with our nodejs version (kind of weird but well...)


**💬 suggestion:** ‏write a more well-suited hook ourselves:

```ts
function useAutoFlushEvents(enabled: boolean) {
  useEffect(() => {
    if (enabled) {
      flushEvents()
      const id = setInterval(flushEvents, FLUSH_EVENTS_INTERVAL)
      return () => clearInterval(id)
    }
  }, [enabled])
}


// then use it in the `ActionBar` component like that:
useAutoFlushEvents(autoFlush)
```

BenoitZugmeyer (approved):

PR #1827, opened 2022-11-17
⚡️ ⚗ [RUMF-1438] Collect feature flags

amortemousque:
## Motivation

Expose a new API to collect feature flags

## Changes

- New API `addFeatureFlags()` under feature flag

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/contexts/viewContexts.ts
```diff
@@ -25,6 +26,9 @@ export function startViewContexts(lifeCycle: LifeCycle): ViewContexts {
     viewContextHistory.add(buildViewContext(view), view.startClocks.relative)
   })
 
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (view) => {
+    viewContextHistory.add(buildViewContext(view), view.startClocks.relative)
+  })
```
BenoitZugmeyer, line 31: **💬 suggestion:** ‏ ‏I'm not sure if it's good thing to add a new history entry on each view update. What about mutating the feature flag object in `trackViews`, so it will be up to date in the context as well?

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.spec.ts
```diff
@@ -131,6 +140,14 @@ describe('error collection', () => {
       })
     })
 
+    // it('should have feature flag of the current view', () => {
```
BenoitZugmeyer, line 143: **💬 suggestion:** ‏remove this commented test?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -223,6 +228,10 @@ export function makeRumPublicApi(
       addTimingStrategy(name, time as RelativeTime | TimeStamp | undefined)
     }),
 
+    addFeatureFlags: monitor((featureFlags: object) => {
+      addFeatureFlagsStrategy(featureFlags as Context)
+    }),
```
BenoitZugmeyer, line 233: **💬 suggestion:** ‏don't add this method if the FF is disabled. Also, make sure it doesn't show up in public TS typings

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -602,6 +603,53 @@ describe('rum public api', () => {
...
+        noopRecorderApi
+      )
+      setupBuilder = setup()
+    })
+
+    afterEach(() => {
+      setupBuilder.cleanup()
+    })
+
+    it('should allow to add feature flags before init', () => {
```
liywjl, line 629:
```suggestion
    it('should collect feature flags from before init', () => {
```

amortemousque: I updated the test to be "API should not be available when ff feature_flags disabled"

BenoitZugmeyer (approved):

liywjl (approved):
LGTM

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -165,6 +176,10 @@ export function trackViews(
       currentView.addTiming(name, time)
       currentView.scheduleUpdate()
     },
+    addFeatureFlags: (featureFlags: Context) => {
+      currentView.addFeatureFlags(featureFlags)
+      currentView.scheduleUpdate()
+    },
```
bcaudan, line 182: **💬 suggestion:** ‏IMO, trackViews has enough responsibility related to view tracking logic. What about extracting feature flag management to a dedicated component? maybe something like `urlContexts`

amortemousque: Yes, I also love standalone module :). With this approach the only complexity I see it to trigger a `view_updated` when a new feature flag is added. 
I could:
- a make a featureFlagObservable so that the view module is notified  
- expose the scheduleUpdate() method to the featureFlagContexts
Wdyt?

amortemousque: I guess it was to be consistent to the timing updates behaviour since the implementation is currently very close. But it might not be needed.

amortemousque: I'll create a follow up task for this update

PR #1837, opened 2022-11-23
🐛 [RUMF-1421] improve counters by filtering child events

BenoitZugmeyer:
## Motivation

We noticed that child event counters were often inaccurate. This first PR goal is to imporve the situation by only taking child events related to the current parent event into account.

## Changes

When computing child event counters, only take events related to the current parent event into account.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
@@ -256,7 +256,6 @@ function newClick(
   const eventCountsSubscription = trackEventCounts({
     lifeCycle,
     predicate: (event) =>
-      event.type !== 'view' &&
       event.action !== undefined &&
       (Array.isArray(event.action.id) ? includes(event.action.id, id) : event.action.id === id),
   })
--- a/packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
@@ -1,5 +1,6 @@
 import type { Duration, ClocksState, RelativeTime, TimeStamp } from '@datadog/browser-core'
 import {
+  includes,
   timeStampNow,
   isExperimentalFeatureEnabled,
   Observable,
@@ -252,7 +253,13 @@ function newClick(
   const id = generateUUID()
   const startClocks = clocksNow()
   const historyEntry = history.add(id, startClocks.relative)
-  const eventCountsSubscription = trackEventCounts(lifeCycle)
+  const eventCountsSubscription = trackEventCounts({
+    lifeCycle,
+    predicate: (event) =>
+      event.type !== 'view' &&
+      event.action !== undefined &&
+      (Array.isArray(event.action.id) ? includes(event.action.id, id) : event.action.id === id),
+  })
   let status = ClickStatus.ONGOING
   let activityEndTime: undefined | TimeStamp
   const frustrationTypes: FrustrationType[] = []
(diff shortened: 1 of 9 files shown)


packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -252,7 +253,13 @@ function newClick(
   const id = generateUUID()
   const startClocks = clocksNow()
   const historyEntry = history.add(id, startClocks.relative)
-  const eventCountsSubscription = trackEventCounts(lifeCycle)
+  const eventCountsSubscription = trackEventCounts({
+    lifeCycle,
+    predicate: (event) =>
+      event.type !== 'view' &&
+      event.action !== undefined &&
```
bcaudan, line 260: **💭 thought:** FMU we are excluding `view` type because of `action.count`‏ attribute, should we directly have a check like:
```ts
event.action?.id !== undefined
```

bcaudan (approved):

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
BenoitZugmeyer: We are excluding `view` because on `RumViewEvent`, the `action` property has type `unknown` ([any undefined property has type `unknown`](https://github.com/DataDog/browser-sdk/blob/911084ad512096420f790f9e25ebc3ba255610e7/packages/rum-core/src/rumEvent.types.ts#L750)). So we can't do `event.action?.id` directly.

TS v4.9 improve ["unknown" type narrowing](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#unlisted-property-narrowing-with-the-in-operator) and we'll be able to do something [like that](https://www.typescriptlang.org/play?ts=5.0.0-dev.20221125#code/JYOwLgpgTgZghgYwgAgEoFcC2A1YEDuAogG4TjIDeAUMrcgNoDWAXMgM5hSgDmAuq+hCMQAe3wgqAXyqhIsRCgyZCUKCKgkyYSjTqIwwESAD8rCsmAATVhy4huyadLABPAA6Ksm8gF40X1XVvbQAffxw8IlJwKioYQQQDI2Q3KAhLYAQ4SAAKCGiwViVggEodWjSwdCgQZFcPERhkfK0AOn1DWp9u5AAiEQAjACsIRN7kADIJ5oL2xM7kAEIekHQAGzXJ6d6rcdAZto7kqYPwOaSQVqtkbr9esBEH3qkgA), but for now the easiest thing is to exclude types that don't define an `action` property

bcaudan (approved):

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -252,7 +253,12 @@ function newClick(
   const id = generateUUID()
   const startClocks = clocksNow()
   const historyEntry = history.add(id, startClocks.relative)
-  const eventCountsSubscription = trackEventCounts(lifeCycle)
+  const eventCountsSubscription = trackEventCounts({
+    lifeCycle,
+    predicate: (event) =>
+      event.action !== undefined &&
+      (Array.isArray(event.action.id) ? includes(event.action.id, id) : event.action.id === id),
```
amortemousque, line 260: **🥜 nitpick:** ‏To be a bit more explicit what do you think of naming the predicate functions. Something like `predicate: isChildEvent`

PR #1839, opened 2022-11-25
♻️ [RUMF-1440] improve feature flag collection implementation

amortemousque:
## Motivation

Extracting feature flag management to a dedicated component.

## Changes
- Remove feature flag logic from trackViews
- Create featureFlagContexts
- Use featureFlagContexts in viewCollection and errorCollection

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
+++ b/packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
@@ -141,10 +141,10 @@ describe('featureFlagContexts', () => {
       featureFlagContexts.addFeatureFlagEvaluation('feature', 'two')
 
       lifeCycle.notify(LifeCycleEventType.VIEW_ENDED, {
-        endClocks: relativeToClocks(30 as RelativeTime),
+        endClocks: relativeToClocks(20 as RelativeTime),
       } as ViewEndedEvent)
       lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
-        startClocks: relativeToClocks(30 as RelativeTime),
+        startClocks: relativeToClocks(20 as RelativeTime),
       } as ViewCreatedEvent)
 
       expect(featureFlagContexts.findFeatureFlagEvaluations(5 as RelativeTime)).toEqual({ feature: 'one' })
--- a/packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
+++ b/packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
@@ -82,7 +82,7 @@ describe('featureFlagContexts', () => {
 
       const featureFlagContext = featureFlagContexts.findFeatureFlagEvaluations()!
 
-      expect(featureFlagContext).toEqual({})
+      expect(featureFlagContext).toBeUndefined()
     })
   })
 
--- a/packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
+++ b/packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
@@ -1,0 +1,154 @@
+import type { RelativeTime } from '@datadog/browser-core'
+import { resetExperimentalFeatures, updateExperimentalFeatures, relativeToClocks } from '@datadog/browser-core'
+import type { TestSetupBuilder } from '../../../test/specHelper'
+import { setup } from '../../../test/specHelper'
+import { LifeCycleEventType } from '../lifeCycle'
+import type { ViewCreatedEvent, ViewEndedEvent } from '../rumEventsCollection/view/trackViews'
+import type { FeatureFlagContexts } from './featureFlagContext'
+import { startFeatureFlagContexts } from './featureFlagContext'
+
+describe('featureFlagContexts', () => {
+  let setupBuilder: TestSetupBuilder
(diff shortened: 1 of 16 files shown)


packages/rum-core/src/domain/contexts/featureFlagContext.ts
```diff
@@ -0,0 +1,38 @@
...
+  })
+
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, ({ startClocks }) => {
+    featureFlagContexts.add({}, startClocks.relative)
+  })
+
+  return {
+    findFeatureFlagEvaluations: (startTime?: RelativeTime) => featureFlagContexts.find(startTime),
+    addFeatureFlagEvaluation: (key: string, value: ContextValue) => {
+      if (isExperimentalFeatureEnabled('feature_flags')) {
```
liywjl, line 29: **💬 suggestion:**
You could check the FF condition first and just return a `noop`  for `findFeatureFlagEvaluations` and `addFeatureFlagEvaluation`. That way you would not need to add any  `lifeCycle.subscribe`

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
```diff
@@ -105,14 +105,15 @@ function processError(
...
   const inForeground = foregroundContexts.isInForegroundAt(error.startClocks.relative)
   if (inForeground) {
     rawRumEvent.view = { in_foreground: inForeground }
   }
 
-  const viewContext = viewContexts.findView(error.startClocks.relative)
-  if (viewContext && !isEmptyObject(viewContext.featureFlagEvaluations)) {
-    rawRumEvent.feature_flags = viewContext.featureFlagEvaluations
+  const featureFlagContext = featureFlagContexts.findFeatureFlagEvaluations(error.startClocks.relative)
+  if (featureFlagContext && !isEmptyObject(featureFlagContext)) {
```
liywjl, line 115: **💬 suggestion:**
You probably do not need to check `!isEmptyObject(featureFlagContext)`

packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
+        startClocks: relativeToClocks(10 as RelativeTime),
+      } as ViewCreatedEvent)
+
+      const featureFlagContext = featureFlagContexts.findFeatureFlagEvaluations()!
+      expect(featureFlagContext).toEqual({})
+    })
+
+    it('should return the feature flag context corresponding to the start time', () => {
+      updateExperimentalFeatures(['feature_flags'])
```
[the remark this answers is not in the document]
amortemousque: the `resetExperimentalFeatures` is called in the parent describe

packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts
amortemousque: The feature flag context with an empty object is created for each `VIEW_CREATED`. If no they are no calls to `addFeatureFlagEvaluation`, we don't want to see the attribute in Datadog. Same logic applied for `customTimings` in `viewCollections`. Wdyt?

packages/rum-core/src/domain/contexts/featureFlagContext.ts
```diff
@@ -0,0 +1,38 @@
...
+        const currentContext = featureFlagContexts.find()
+        if (currentContext) {
+          // mutate the current context to avoid creating a new context history entry to save memory
+          assign(currentContext, { [key]: value })
```
bcaudan: **❓ question:** ‏I don't get the intent here, why not do directly:
```ts
context[key] = value
```
no test seems to cover that

bcaudan: **❓ question:** ‏should not we `deepClone` the `value` like we do for other APIs allowing customers to pass their own objects?

packages/rum-core/src/domain/contexts/featureFlagContext.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+      clock.tick(10)
+      featureFlagContexts.addFeatureFlagEvaluation('feature', 'two')
+
+      lifeCycle.notify(LifeCycleEventType.VIEW_ENDED, {
+        endClocks: relativeToClocks(30 as RelativeTime),
+      } as ViewEndedEvent)
+      lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
+        startClocks: relativeToClocks(30 as RelativeTime),
+      } as ViewCreatedEvent)
```
bcaudan, line 148: **💬 suggestion:** ‏those statements do not seem to be useful for the test

packages/rum-core/src/domain/contexts/featureFlagContext.ts
amortemousque: Indeed good point

bcaudan (approved):

PR #1850, opened 2022-12-02
✨ send data from the SDK to the extension

BenoitZugmeyer:
## Motivation

Make the SDK send its data to the extension as soon as it happens. This will allow the extension to list events as soon as they happen, and integrate with sessior replay data.

## Changes

* The SDK now notify the extension every time some data is collected. To make sure it does not have any impact in production, this is only done in dev bundles. We could change that in the future.
* Improve logging and communication error handling in the extension
* Forward messages from the SDK up to the devtools panel
* Finally, implement the setting to display events directly from the SDK

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/developer-extension/src/common/actions.ts
+++ b/developer-extension/src/common/actions.ts
@@ -1,7 +1,3 @@
-import { createLogger } from './logger'
-
-const logger = createLogger('action')
-
 type ValueOf<T> = T[keyof T]
 
 type Message<Actions extends { [key: string]: any }> = ValueOf<{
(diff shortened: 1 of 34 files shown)


developer-extension/src/panel/hooks/useEvents.ts
```diff
@@ -114,3 +136,11 @@ function listenRequests(callback: (events: StoredEvent[]) => void) {
 
   return () => chrome.devtools.network.onRequestFinished.removeListener(beforeRequestHandler)
 }
+
+function listenEventsFromSdk(events: (events: StoredEvent[]) => void) {
+  return listenSdkMessages((message) => {
+    if (message.type === 'logs' || message.type === 'rum' || message.type === 'telemetry') {
```
bcaudan, line 142: **❓ question:** ‏are you planning to handle records in another PR?

packages/core/src/tools/sendToExtension.ts
```diff
@@ -0,0 +1,14 @@
...
+export function sendToExtension(type: ExtensionMessageType, payload: unknown) {
+  // Only allow this for dev bundles for now, as it might impact performances?
+  if (__BUILD_ENV__SDK_VERSION__ === 'dev') {
+    window.dispatchEvent(
+      new CustomEvent('__ddBrowserSdkMessage', {
+        detail: { type, payload },
+      })
+    )
+  }
+}
```
bcaudan, line 14: **💭 thought:** ‏IMO, for the performance impact alone, we could have an approach allowing us to use it with prod bundle as well by emitting sdk events only after a specific API has been called.
A bit like the debug mode with the `_setDebug` API.
However, that would open a new public api for others to integrate with us which could require a bit more thoughts on this API.
wdyt?

BenoitZugmeyer: Sure, let me try that! I want to make sure that whatever solution we chose, the very early events are captured (ex: those that are collected during `init`). That's why it could be a bit trickier to implement.

developer-extension/src/common/actions.ts
```diff
@@ -26,34 +26,13 @@ export function createListenAction<Actions extends { [key: string]: any }>() {
...
+export function createSendAction<Actions>(onError: (error: { message: string }) => void) {
   function sendAction<K extends keyof Actions>(action: K, payload: Actions[K]) {
     try {
-      chrome.runtime.sendMessage({ action, payload }).catch(handleError)
+      chrome.runtime.sendMessage({ action, payload }).catch(onError)
     } catch (error) {
-      handleError(error)
+      onError(error as { message: string })
     }
   }
```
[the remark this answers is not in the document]
BenoitZugmeyer: One is synchronous, the other is async...

liywjl (approved):

bcaudan (approved):
LGTM

PR #1856, opened 2022-12-05
👷Fix staging reset

bcaudan:
## Motivation

During staging reset, if the staging branch already exists, it is not updated and all subsequent `merge-into-staging` checks are failing. 

## Changes

Delete staging branch if it already exists and create a fresh one.

## Testing

None, we'll see how it will go next week.

---

### DIFF
--- a/scripts/staging-ci/staging-reset.js
+++ b/scripts/staging-ci/staging-reset.js
@@ -31,15 +31,14 @@ async function main() {
   }
 
   const doesStagingExist = await executeCommand(`git ls-remote --heads ${REPOSITORY} ${NEW_STAGING_BRANCH}`)
-  if (!doesStagingExist) {
-    printLog('Creating the new staging branch...')
-    await executeCommand(`git checkout -b ${NEW_STAGING_BRANCH}`)
-    await executeCommand(`git push origin ${NEW_STAGING_BRANCH}`)
-  } else {
-    printLog('New staging branch already exists, updating...')
-    await executeCommand(`git branch -f ${NEW_STAGING_BRANCH}`)
-    await executeCommand(`git push origin ${NEW_STAGING_BRANCH}`)
+  if (doesStagingExist) {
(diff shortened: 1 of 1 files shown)


liywjl (approved):

scripts/staging-ci/staging-reset.js
```diff
@@ -30,13 +30,15 @@ async function main() {
...
+  const isStagingAlreadyExists = await executeCommand(`git ls-remote --heads ${REPOSITORY} ${NEW_STAGING_BRANCH}`)
+  if (!isStagingAlreadyExists) {
     printLog('Creating the new staging branch...')
     await executeCommand(`git checkout -b ${NEW_STAGING_BRANCH}`)
     await executeCommand(`git push origin ${NEW_STAGING_BRANCH}`)
   } else {
-    printLog('New staging branch already created. Skipping.')
+    printLog('New staging branch already exists, updating...')
+    await executeCommand(`git branch -f ${NEW_STAGING_BRANCH}`)
+    await executeCommand(`git push origin ${NEW_STAGING_BRANCH}`)
```
BenoitZugmeyer, line 41: **🔨 warning:** ‏I don't think this will work as intended:
* `git branch -f staging-nn` will move the branch `staging-nn` reference to the current `HEAD` (which is `main` at this point), but it *won't* switch to the `staging-nn` branch, so `git push` will try to push `main` instead of `staging-nn`. I think what you want is `git checkout -B staging-nn` (uppercase B): https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutem-b-Bltnew-branchgtltstart-pointgt
* `git push origin` will fail because the upstream `staging-nn` branch is probably not an ancestor of the pushed branch. You might want to use `--force`.

This could work and I'd be fine with it, but why not just removing old branches by hand?

BenoitZugmeyer (approved):

PR #1860, opened 2022-12-07
🐛 [RUMF-1449]  workaround for Firefox memory leak when using Zone.js

BenoitZugmeyer:
## Motivation

Firefox is leaking memory when using Zone.js + `visualViewport.addEventListener`: https://bugzilla.mozilla.org/show_bug.cgi?id=1804409

## Changes

* Some light refactoring
* Use the same strategy as for MutationObserver: we try to use the original `addEventListener` method instead of the patched one

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


liywjl (approved):

liywjl (approved):

packages/core/src/tools/utils.ts
```diff
@@ -9,39 +10,6 @@ export const ONE_YEAR = 365 * ONE_DAY
 export const ONE_KIBI_BYTE = 1024
 export const ONE_MEBI_BYTE = 1024 * ONE_KIBI_BYTE
 
-export const enum DOM_EVENT {
```
amortemousque, line 12: **👏 praise:** ‏I totally agree on moving it to `/browser` 👍

packages/core/test/specHelper.ts
```diff
@@ -485,3 +486,31 @@ export function interceptRequests() {
     },
   }
 }
+
+export function stubZoneJs() {
```
amortemousque, line 490: **💬 suggestion:** ‏The specHelper.ts is quite big. It could be nice to move the function to a dedicated module

packages/core/src/tools/getZoneJsOriginalValue.ts
```diff
@@ -0,0 +1,33 @@
...
+ * [1]: https://github.com/angular/angular/tree/main/packages/zone.js
+ */
+export function getZoneJsOriginalValue<Target, Name extends keyof Target & string>(
+  target: Target,
+  name: Name
+): Target[Name] {
+  const browserWindow = window as BrowserWindowWithZoneJs
+  let original: Target[Name] | undefined
+  if (browserWindow.Zone) {
+    original = (target as any)[browserWindow.Zone.__symbol__(name)]
```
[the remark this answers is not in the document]
BenoitZugmeyer: Yes this is something we used before: https://github.com/DataDog/browser-sdk/blob/e4f2bad8d6ecaea8e83eaf3da5f06f177bade01c/packages/rum-core/src/browser/domMutationObservable.ts#L47 and it works with all tested Zone.js versions (from old ones used by Angularjs 1 to latest ones)

packages/core/test/specHelper.ts
BenoitZugmeyer: Yes right! Done.

bcaudan (approved):

amortemousque (approved):

PR #1861, opened 2022-12-08
🔊 Collect computed and perf entry durations

amortemousque:
## Motivation

In the context of the investigation of resources with  long durations, I want to assess the performance entry durations correctness.
To do it, I want to colllect xhr and fetch computed durations and corresponding performance entry durations to identify:

- happens when we got a the duration from performance entry, computed duration or both
- compare entry duration with computed duration that could highlight an issue from the browser

## Changes

Adds `resource._dd.computed_duration` and `resource._dd.entry_performance_duration`
Under feature flag resource_durations

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
@@ -68,13 +68,15 @@ function processRequest(
 
   const responseDurationInfo = computeResponseDurationInfo(request)
 
+  const duration = toServerDuration(request.duration)
+
   const resourceEvent = combine(
     {
       date: startClocks.timeStamp,
       resource: {
         id: generateUUID(),
         type,
-        duration: toServerDuration(request.duration),
+        duration,
         method: request.method,
         status_code: request.status,
         url: request.url,
@@ -84,7 +86,15 @@ function processRequest(
     tracingInfo,
     correspondingTimingOverrides,
     indexingInfo,
-    responseDurationInfo
+    responseDurationInfo,
+    {
+      _dd: {
+        computed_duration: duration,
+        performance_entry_duration: correspondingTimingOverrides
+          ? correspondingTimingOverrides.resource.duration
+          : undefined,
+      },
+    }
   )
   return {
     startTime: startClocks.relative,
--- a/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
@@ -70,6 +70,16 @@ function processRequest(
   const responseDurationInfo = computeResponseDurationInfo(request)
 
   const duration = toServerDuration(request.duration)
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -84,7 +87,17 @@ function processRequest(
...
+    isExperimentalFeatureEnabled('resource_durations')
+      ? {
+          _dd: {
+            computed_duration: duration,
+            performance_entry_duration: correspondingTimingOverrides
+              ? correspondingTimingOverrides.resource.duration
+              : undefined,
+          },
+        }
+      : undefined
```
liywjl, line 100: **💬 suggestion:** maybe lift this into a variable before
 `const resourceEvent = combine(` 
WDYT?

liywjl: **💭 thought:** I added a diff and diff percentage variable on another [PR](https://github.com/DataDog/browser-sdk/pull/1810), is that something that you think might be useful?

amortemousque: I lifted it. 
I'd say that without your fix, I could have fallen into wrong conclusions, comparing computed and entry resource duration. So in that sense it helps me doing my investigation :)

liywjl (approved):

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -68,13 +69,25 @@ function processRequest(
...
+  const computedAndEntryDurations = isExperimentalFeatureEnabled('resource_durations')
+    ? {
+        _dd: {
+          computed_duration: duration,
+          performance_entry_duration: correspondingTimingOverrides
+            ? correspondingTimingOverrides.resource.duration
+            : undefined,
+        },
+      }
+    : undefined
```
bcaudan, line 82: **💬 suggestion:** ‏what about extracting this logic in a function like the other computations?

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -68,13 +69,25 @@ function processRequest(
...
   const responseDurationInfo = computeResponseDurationInfo(request)
 
+  const duration = toServerDuration(request.duration)
+  const computedAndEntryDurations = isExperimentalFeatureEnabled('resource_durations')
+    ? {
+        _dd: {
+          computed_duration: duration,
+          performance_entry_duration: correspondingTimingOverrides
+            ? correspondingTimingOverrides.resource.duration
+            : undefined,
```
bcaudan, line 79: **💬 suggestion:** ‏not sure how you plan to analyse those data but collecting directly the difference between the two may be easier to analyse.

BenoitZugmeyer (approved):

bcaudan (approved):

PR #1862, opened 2022-12-09
🐛 [RUMF-1337] Fix incorrect fetch duration

liywjl:
## Motivation

Release the feature intruded in https://github.com/DataDog/browser-sdk/pull/1810 to increase: 
- the precision of `fetch` duration 
- the number of matches for `fetch` resource timings

## How?

Some `fetch` resources, we were under reporting `fetch`'s duration. This was because `fetch` does not resolve when the payload has been downloaded but rather when the response is first received.

The example below only would calculate the duration of the `fetch`'s resolve:
```
startDuration()
fetch('/').then(res => {
  stopDuration()
  // ...
})
```

To fix this, we listen to `fetch`'s [Response.body](https://developer.mozilla.org/en-US/docs/Web/API/Response/body) stream to listen for the completion of the stream before calculation `fetch's` duration ([code](https://github.com/DataDog/browser-sdk/pull/1862/files#diff-e0694bed3d59cee510f2dc21b6cf9141eca03c493fabfebc8454f3d8bd1ad115R163-R177)).

## Fetch resource timings

[PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) are precise timing metrics supplied to us by the browser, that we use to enrich our data. This data is generated after `fetch` has completed.

This fix will thus increase our probability of finding a match; as before we were requesting this information too early.

## Changes

- remove `fetch_duration` feature flag
- remove adjacent tests

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan (approved):

BenoitZugmeyer (commented):
I think you could remove any reference to `resolveDuration` as well, since it's not used anymore

BenoitZugmeyer (approved):

PR #1864, opened 2022-12-09
🐛 [RUMF-1421] keep updating the view event counters after view end 

BenoitZugmeyer:
## Motivation

Improve View child event counters accuracy

## Changes

Keep updating View child event counters after the view ends.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViewEventCounts.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViewEventCounts.ts
@@ -1,0 +1,12 @@
+import type { LifeCycle } from '../../lifeCycle'
+import { trackEventCounts } from '../../trackEventCounts'
+
+export function trackViewEventCounts(lifeCycle: LifeCycle, viewId: string, onChange: () => void) {
+  const { stop, eventCounts } = trackEventCounts({
+    lifeCycle,
+    isChildEvent: (event) => event.view.id === viewId,
+    onChange,
+  })
+
+  return { stop, eventCounts }
+}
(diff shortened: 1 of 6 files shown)


packages/rum-core/src/domain/rumEventsCollection/view/trackViewEventCounts.ts
```diff
@@ -0,0 +1,32 @@
+import { ONE_MINUTE } from '@datadog/browser-core'
+import type { LifeCycle } from '../../lifeCycle'
+import { trackEventCounts } from '../../trackEventCounts'
+
+// Arbitrary delay for stopping event counting after the view ends. Ideally, we would not stop and
+// keep counting events until the end of the session. But this might have a small performance impact
+// if there are many many views: we would need to go through each event to see if the related view
+// matches. So let's have a fairly short delay to avoid impacting performances too much.
```
liywjl, line 8: **💬 suggestion:** It might be good to explain a little more the issue

```suggestion
// Some events are not being counted as they transcend views. To reduce the occurrence; 
// an arbitrary delay is added for stopping event counting after the view ends. 
// 
// Ideally, we would not stop and keep counting events until the end of the session.
// But this might have a small performance impact if there are many many views: 
// we would need to go through each event to see if the related view matches.
// So let's have a fairly short delay to avoid impacting performances too much.
```

packages/rum-core/src/domain/rumEventsCollection/view/trackViewEventCounts.ts
```diff
@@ -0,0 +1,32 @@
...
+export function trackViewEventCounts(lifeCycle: LifeCycle, viewId: string, onChange: () => void) {
+  const { stop, eventCounts } = trackEventCounts({
+    lifeCycle,
+    isChildEvent: (event) => event.view.id === viewId,
+    onChange,
+  })
+
+  return {
+    scheduleStop: () => {
+      setTimeout(stop, KEEP_TRACKING_EVENT_COUNTS_AFTER_VIEW_DELAY)
```
bcaudan, line 28: **💬 suggestion:** ‏missing monitor
```suggestion
      setTimeout(monitor(stop), KEEP_TRACKING_EVENT_COUNTS_AFTER_VIEW_DELAY)
```

bcaudan (approved):

liywjl (approved):

PR #1877, opened 2022-12-14
🐛[RUMF-1435] don't retry opaque response

bcaudan:
## Motivation

If the browser somehow consider the intake response as [opaque](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque), we won't be able to access its true status and may wrongly retry it.

## Changes

- do not retry opaque response
- force [request mode](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode) 'cors'
- add extra debug tag with response type

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/sendWithRetryStrategy.spec.ts
+++ b/packages/core/src/transport/sendWithRetryStrategy.spec.ts
@@ -190,50 +190,59 @@ describe('sendWithRetryStrategy', () => {
     })
   })
   ;[
-    { description: 'when the intake returns error:', status: 500 },
-    { description: 'when the intake returns too many request:', status: 429 },
-    { description: 'when the intake returns request timeout:', status: 408 },
-    { description: 'when network is down:', status: 0 },
-  ].forEach(({ description, status }) => {
+    { expectRetry: true, description: 'when the intake returns error:', status: 500 },
+    { expectRetry: true, description: 'when the intake returns too many request:', status: 429 },
+    { expectRetry: true, description: 'when the intake returns request timeout:', status: 408 },
+    { expectRetry: true, description: 'when network error:', status: 0, type: undefined },
+    { expectRetry: false, description: 'when the intake returns opaque response:', status: 0, type: 'opaque' as const },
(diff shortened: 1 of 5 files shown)


packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -133,7 +134,10 @@ function retryQueuedPayloads(
 }
 
 function shouldRetryRequest(response: HttpResponse) {
-  return response.status === 0 || response.status === 408 || response.status === 429 || response.status >= 500
+  return (
+    response?.type !== 'opaque' &&
```
BenoitZugmeyer, line 138: **💬 suggestion:** ‏
```suggestion
    response.type !== 'opaque' &&
```

packages/core/src/transport/sendWithRetryStrategy.spec.ts
```diff
@@ -190,50 +190,68 @@ describe('sendWithRetryStrategy', () => {
...
+            lastFailureStatus: status,
+            lastFailureType: type,
+          })
+        })
+      } else {
+        it('should not queue the payload for retry', () => {
+          sendRequest()
+          sendStub.respondWith(0, { status, type })
+          expect(state.queuedPayloads.size()).toBe(0)
+        })
```
BenoitZugmeyer, line 253: **💬 suggestion:** ‏Since there is only one case where `expectRetry` is false, and test cases are completely different (no common code executed), what about declaring a separate test case for it?

bcaudan: I was thinking about the next PR that should introduce other cases around `onLine` status to decide if we should retry or not.
So having all the cases described in the same place and running the different expectations depending on if we shold retry or not seemed convenient.
wdyt?

BenoitZugmeyer: Alright, in this case it seems fine!

BenoitZugmeyer (approved):

liywjl (approved):

PR #1878, opened 2022-12-14
⚗️ [RUMF-1345] log first 3 untrusted events

liywjl:
## Motivation

We want to better understand what type of [untrusted events](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) are being recorded.

## Changes

- create `listenerWithTelemetry` to wrap listener function
- only log telemetry when `log_untrusted_events` flag enabled
- only log first 3 instances of `isTrusted` events
- lift `runOnReadyState` out of utils to avoid circular dependency

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


BenoitZugmeyer:
Superseded by #1910

PR #1879, opened 2022-12-14
👷 [RUMF-1287] don't update rum-events-format on postinstall

BenoitZugmeyer:
## Motivation

This is problematic because rum-events-format can change at any time independently of the Browser SDK repository, and it can make the CI fail when running various unit and e2e tests jobs or when publishing to NPM.

## Changes

Don't update the submodule, just initialize it during `postinstall`.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "performances"
   ],
   "scripts": {
-    "postinstall": "scripts/cli update_submodule",
+    "postinstall": "scripts/cli init_submodule",
     "build": "lerna run build --stream",
     "build:bundle": "lerna run build:bundle --stream",
     "format": "prettier --check .",
--- a/scripts/cli
+++ b/scripts/cli
@@ -31,6 +31,10 @@ cmd_lint () {
   eslint "$project_path"
 }
 
+cmd_init_submodule () {
+  git submodule update --init
+}
+
 cmd_update_submodule () {
   git submodule update --init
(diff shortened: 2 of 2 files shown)


liywjl (approved):

package.json
```diff
@@ -8,7 +8,7 @@
     "performances"
   ],
   "scripts": {
-    "postinstall": "scripts/cli update_submodule",
+    "postinstall": "scripts/cli init_submodule",
```
bcaudan, line 11: **💭 thought:** ‏IIRC, the intent was to be able to detect early if some breaking changes were introduced by the schema.
Maybe we could add a specific gitlab job that run the unit tests on the updated schema.
wdyt?

BenoitZugmeyer: What would be the impact when a breaking change happen in the `rum-events-format`? If we were to catch it when updating the schema in one of our tasks, I think it would be fine, no?

bcaudan: The idea would be to detect it sooner than later. If the breaking change is done by the mobile team, it may take us a while before we detect it.

BenoitZugmeyer: We now run SDK tests in `rum-event-format` directly https://github.com/DataDog/rum-events-format/pull/124

bcaudan (approved):

Issue #1881, opened 2022-12-14
No matching version found for @datadog/browser-rum@4.27.1.


just-web-dev:
After updating to version 4.28.0 for both browser-logs and browser-rum, when running npm install I get an npm error of  **No matching version found for @datadog/browser-rum@4.27.1.**

BenoitZugmeyer:
Thank you for opening this issue, and sorry for the inconvenience. Something went wrong during the v4.28.0 release and some versions were not updated. A new release has been made to fix the issue.

PR #1884, opened 2022-12-15
👷  add a CI job to do some checks before the release

BenoitZugmeyer:
## Motivation

Prevents https://github.com/DataDog/browser-sdk/issues/1881
## Changes

Add a CI job to do some checks before doing a release. It checks that:
* the version tag matches the lerna version
* the git HEAD and the version tag are pointing to the same commit
* package.json files have the correct version in their `version` field and their prod/dev/peer dependencies
* packages entry points are pointing to existing files

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/check-licenses.js
+++ b/scripts/check-licenses.js
@@ -3,12 +3,12 @@
 const fs = require('fs')
 const path = require('path')
 const readline = require('readline')
-const { findPackages, printLog, printError, logAndExit } = require('./utils')
+const { printLog, printError, logAndExit, generateMetadataForAllNodePackages } = require('./utils')
 
 const LICENSE_FILE = 'LICENSE-3rdparty.csv'
 
 async function main() {
-  const packages = await findPackages()
+  const packages = await generateMetadataForAllNodePackages()
 
   printLog(
     'Looking for dependencies in:\n',
--- a/scripts/check-licenses.js
+++ b/scripts/check-licenses.js
@@ -3,15 +3,19 @@
 const fs = require('fs')
 const path = require('path')
 const readline = require('readline')
-const { executeCommand, printLog, printError, logAndExit } = require('./utils')
+const { findPackages, printLog, printError, logAndExit } = require('./utils')
 
 const LICENSE_FILE = 'LICENSE-3rdparty.csv'
 
 async function main() {
-  const packageJsonPaths = await findPackageJsonPaths()
+  const packages = await findPackages()
 
-  printLog('Looking for dependencies in:\n', packageJsonPaths, '\n')
-  const declaredDependencies = withoutDuplicates(packageJsonPaths.flatMap(retrievePackageJsonDependencies)).sort()
+  printLog(
+    'Looking for dependencies in:\n',
+    packages.map((pkg) => pkg.relativePath),
+    '\n'
+  )
+  const declaredDependencies = withoutDuplicates(packages.flatMap(retrievePackageDependencies)).sort()
 
   const declaredLicenses = (await retrieveLicenses()).sort()
 
@@ -30,16 +34,9 @@ async function main() {
   printLog('Dependencies check done.')
 }
 
-async function findPackageJsonPaths() {
-  const paths = await executeCommand('find . -path "*/node_modules/*" -prune -o -name "package.json" -print')
-  return paths.trim().split('\n')
-}
-
-function retrievePackageJsonDependencies(packageJsonPath) {
-  const packageJson = require(path.join(__dirname, '..', packageJsonPath))
-
-  return Object.keys(packageJson.dependencies || {})
-    .concat(Object.keys(packageJson.devDependencies || {}))
+function retrievePackageDependencies(pkg) {
+  return Object.keys(pkg.manifest.dependencies || {})
+    .concat(Object.keys(pkg.manifest.devDependencies || {}))
     .filter((dependency) => !dependency.includes('@datadog'))
 }
 
--- a/scripts/check-licenses.js
+++ b/scripts/check-licenses.js
@@ -3,19 +3,19 @@
 const fs = require('fs')
 const path = require('path')
 const readline = require('readline')
-const { printLog, printError, logAndExit, generateMetadataForAllNodePackages } = require('./utils')
+const { printLog, printError, logAndExit, findBrowserSdkPackageJsonFiles } = require('./utils')
 
 const LICENSE_FILE = 'LICENSE-3rdparty.csv'
 
 async function main() {
-  const packages = await generateMetadataForAllNodePackages()
+  const packageJsonFiles = await findBrowserSdkPackageJsonFiles()
 
   printLog(
     'Looking for dependencies in:\n',
-    packages.map((pkg) => pkg.relativePath),
+    packageJsonFiles.map((packageJsonFile) => packageJsonFile.relativePath),
     '\n'
   )
-  const declaredDependencies = withoutDuplicates(packages.flatMap(retrievePackageDependencies)).sort()
+  const declaredDependencies = withoutDuplicates(packageJsonFiles.flatMap(retrievePackageDependencies)).sort()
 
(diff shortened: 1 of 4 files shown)


liywjl (approved):

scripts/utils.js
```diff
@@ -103,6 +104,21 @@ async function fetchWrapper(url, options) {
   return response.text()
 }
 
+async function findPackages() {
```
bcaudan, line 107: **💬 suggestion:** ‏it could be nice to have a naming a bit more specific:

- packages could easily be understood as the content of the `packages` directory
- it is kind of hard to guess what to expect as a result of this function

proposal: `generateMetadataForAllNodePackages()`

scripts/check-licenses.js
```diff
@@ -3,15 +3,19 @@
...
 
 async function main() {
-  const packageJsonPaths = await findPackageJsonPaths()
+  const packages = await findPackages()
 
-  printLog('Looking for dependencies in:\n', packageJsonPaths, '\n')
-  const declaredDependencies = withoutDuplicates(packageJsonPaths.flatMap(retrievePackageJsonDependencies)).sort()
+  printLog(
+    'Looking for dependencies in:\n',
+    packages.map((pkg) => pkg.relativePath),
```
bcaudan, line 15: **🥜 nitpick:** ‏avoid abbreviation
```suggestion
    packages.map((package) => package.relativePath),
```

scripts/check-licenses.js
```diff
@@ -30,16 +34,9 @@ async function main() {
...
-}
-
-function retrievePackageJsonDependencies(packageJsonPath) {
-  const packageJson = require(path.join(__dirname, '..', packageJsonPath))
-
-  return Object.keys(packageJson.dependencies || {})
-    .concat(Object.keys(packageJson.devDependencies || {}))
+function retrievePackageDependencies(pkg) {
+  return Object.keys(pkg.manifest.dependencies || {})
+    .concat(Object.keys(pkg.manifest.devDependencies || {}))
```
bcaudan, line 39: **🥜 nitpick:** ‏avoid abbreviation
```suggestion
function retrievePackageDependencies(package) {
  return Object.keys(package.manifest.dependencies || {})
    .concat(Object.keys(package.manifest.devDependencies || {}))
```

scripts/check-release.js
```diff
@@ -0,0 +1,90 @@
...
+    throw new Error('Git tag not on HEAD')
+  }
+}
+
+async function checkPackages() {
+  const packages = await findPackages()
+
+  printLog(
+    `Checking packages:
+   - ${packages.map((pkg) => pkg.relativePath).join('\n   - ')}
```
bcaudan, line 34: **🥜 nitpick:** ‏avoid abbreviation in this file as well

scripts/check-release.js
```diff
@@ -0,0 +1,90 @@
...
+  } catch (error) {
+    throw new Error(`Failed to find git tag reference: ${error}`)
+  }
+  if (tagRef !== headRef) {
+    throw new Error('Git tag not on HEAD')
+  }
+}
+
+async function checkPackages() {
+  const packages = await findPackages()
```
bcaudan, line 30: **💬 suggestion:** ‏FMU we only want to check browser SDK packages. If so what about renaming the check function to make it clearer and do an early filter after the `findPackages` call.

scripts/check-release.js
```diff
@@ -0,0 +1,90 @@
...
+
+async function main() {
+  await checkGitTag()
+  await checkPackages()
+
+  printLog('Release check done.')
+}
+
+async function checkGitTag() {
+  printLog('Checking git tag')
```
bcaudan, line 16: **💬 suggestion:** ‏
```suggestion
  printLog('Checking release version tag is on HEAD')
```

scripts/check-release.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    const absoluteEntryPointPath = path.resolve(pkg.path, entryPointPath)
+    if (!fs.existsSync(absoluteEntryPointPath)) {
+      throw new Error(
+        `Invalid entry point ${entryPointPath} in ${pkg.relativePath}: ${absoluteEntryPointPath} not found`
+      )
+    }
+  }
+}
```
bcaudan, line 84: **💭 thought:** ‏are you checking entry points of all packages and not only `@datadog` packages in order to be sure to check `internal` packages?
if so, IMO it is a bit weird to check rum-events-format entry point as well.
Should we not consider browser sdk packages as packages under `packages` directory and adjust other checks accordingly?

scripts/check-licenses.js
BenoitZugmeyer: `package` is a reserved keyword it's a syntax error to use it. I usually use `pkg` in this case but I'd be fine with another alternative, like `package_`

scripts/check-release.js
BenoitZugmeyer: I found out that in `check-licenses` we are iterating over `rum-events-format` as well, so I kept this behavior, and since it does not matter that much to check for those packages as well in `check-release`, I didn't filter them out.

But sure, I can remove  packages from `rum-events-format` if you think that it's not needed. I think we should do the same for `check-licenses` then. WDYT?

>  adjust other checks accordingly?

I'm not sure what you mean by that

scripts/check-licenses.js
bcaudan: oh ok, got it. Nevermind then.

scripts/check-release.js
bcaudan: > But sure, I can remove packages from rum-events-format if you think that it's not needed. I think we should do the same for check-licenses then. WDYT?

sounds good to me

>> Should we not consider browser sdk packages as packages under packages directory and adjust other checks accordingly?

> I'm not sure what you mean by that

it seemed to me that some of the checks performed in `checkPackage` would need need to be updated if we change what we consider to be browser sdk packages.

scripts/check-release.js
```diff
@@ -0,0 +1,103 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+const { version: releaseVersion } = require('../lerna.json')
+const { findBrowserSdkPackageJsonFiles, printLog, logAndExit, executeCommand } = require('./utils')
+
+async function main() {
+  // await checkGitTag()
```
bcaudan, line 9: **❓ question:** ‏comment leftover after debug?

BenoitZugmeyer: yep 😄

scripts/check-release.js
```diff
@@ -0,0 +1,103 @@
...
+    isBrowserSdkPublicPackageName(packageJsonFile.content.name) &&
+    packageJsonFile.content.version !== releaseVersion
+  ) {
+    throw new Error(
+      `Invalid version for ${packageJsonFile.relativePath}: expected ${releaseVersion}, got ${packageJsonFile.content.version}`
+    )
+  }
+  checkPackageDependencyVersions(packageJsonFile)
+  checkPackageJsonEntryPoints(packageJsonFile)
+}
```
bcaudan, line 54: **🥜 nitpick:** ‏in order to have a consistent level of abstraction and avoid one wrapping function

```suggestion
  for (const packageJsonFile of packageJsonFiles) {
    checkPackageJsonVersions(packageJsonFile)
    checkPackageDependencyVersions(packageJsonFile)
    checkPackageJsonEntryPoints(packageJsonFile)
  }
}

function checkPackageJsonVersions(packageJsonFile) {
  if (
    isBrowserSdkPublicPackageName(packageJsonFile.content.name) &&
    packageJsonFile.content.version !== releaseVersion
  ) {
    throw new Error(
      `Invalid version for ${packageJsonFile.relativePath}: expected ${releaseVersion}, got ${packageJsonFile.content.version}`
    )
  }
}
```

bcaudan (approved):

PR #1890, opened 2022-12-19
🔊 Collect page lifecycle states in resources

amortemousque:
## Motivation

The investigation on resources with long duration made us think that the[ page lifecycle state](https://developer.chrome.com/blog/page-lifecycle-api/) may have an impact on timings. In order to check this assumption this PR collect each page state happening during the resource duration  (`active`, `passive`, `hidden`, `frozen`, `terminated`).

Format:
```
_dd: {
 page_states: [{state: 'active', startTime: 1}, ...}]
}
```
The implementation logic follows the one from [Google](https://developer.chrome.com/blog/page-lifecycle-api/#observing-state-changes)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/contexts/pageStateContexts.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  it('should return undefined if timings out of timeline bounds', () => {
+    pageStateContexts = startPageStateContexts()
+    expect(pageStateContexts.getPageStates(-10 as RelativeTime, 0 as RelativeTime)).not.toBeDefined()
+  })
+
+  it('should track the number and duration of all page state', () => {
+    const { clock } = setupBuilder.build()
+    resetPageStates()
+
+    clock.tick(0)
```
BenoitZugmeyer, line 48: **🥜 nitpick:** ‏`tick(0)` does nothing

packages/rum-core/src/domain/contexts/pageStateContexts.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    addPageState(PageState.HIDDEN)
+
+    clock.tick(10)
+    addPageState(PageState.FROZEN)
+
+    expect(pageStateContexts.getPageStates(0 as RelativeTime, 50 as RelativeTime)).toEqual({
+      active: { count: 1, duration: 10 as Duration },
+      passive: { count: 1, duration: 10 as Duration },
+      hidden: { count: 1, duration: 10 as Duration },
+      frozen: { count: 1, duration: 20 as Duration },
```
BenoitZugmeyer, line 64: **❓ question:** ‏Not sure why `frozen` duration is 20 in this case?

packages/rum-core/src/domain/contexts/pageStateContexts.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      terminated: { count: 0, duration: 0 as Duration },
```
BenoitZugmeyer, line 65: **💬 suggestion:** ‏add a `terminated` page state so all states are covered?

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,157 @@
...
+      DOM_EVENT.PAGE_HIDE,
+      (ev: PageTransitionEvent) => {
+        // If the event's persisted property is `true` the page is about
+        // to enter the back/forward cache, which is also in the frozen state.
+        // If the event's persisted property is not `true` the page is
+        // about to be unloaded.
+        if (ev.isTrusted) addPageState(ev.persisted ? PageState.FROZEN : PageState.TERMINATED)
+      },
+      listenerOpts
+    )
```
BenoitZugmeyer, line 83: **💬 suggestion:** it might be a bit shorter to use a single `addEventListener` and discriminate over `event.type`

```ts

 addEventListeners(
      window,
      [DOM_EVENT.PAGE_SHOW, DOM_EVENT.FOCUS, DOM_EVENT.BLUR, DOM_EVENT.VISIBILITY_CHANGE, DOM_EVENT.RESUME, DOM_EVENT.FREEZE, DOM_EVENT.PAGE_HIDE],
      (event) => {
        if (!event.isTrusted) return
        if (event.type === DOM_EVENT.FREEZE) {
          addPageState(PageState.FROZEN)
        } else if (event.type === DOM_EVENT.PAGE_HIDE) {
          ...
        } else {
          ...
        }
        
      },
      listenerOpts
    ),
```

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,157 @@
...
+      const pageStateContext: PageStateContext = {
+        active: { count: 0, duration: 0 as Duration },
+        passive: { count: 0, duration: 0 as Duration },
+        hidden: { count: 0, duration: 0 as Duration },
+        frozen: { count: 0, duration: 0 as Duration },
+        terminated: { count: 0, duration: 0 as Duration },
+      }
+
+      let i = findStartIndex(startTime)
+      while (i < pageStateTimeline.length && pageStateTimeline[i].startTime < endTime) {
```
BenoitZugmeyer, line 112: **🥜 nitpick:** ‏a `for(let i = 0; i < ... && ...; i++)` loop might be more idiomatic

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,157 @@
...
+    let i = -1
+    let endTime
+
+    do {
+      i++
+      endTime = i + 1 < pageStateTimeline.length ? pageStateTimeline[i + 1].startTime : Infinity
+    } while (i < pageStateTimeline.length && endTime <= startTime)
+
+    return i
+  }
```
BenoitZugmeyer, line 96: **💭 thought:** ‏this is a bit convoluted, we can probably do something simpler

packages/rum-core/src/domain/contexts/pageStateContexts.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    addPageState(PageState.ACTIVE, limit)
+
+    clock.tick(10)
+    addPageState(PageState.PASSIVE, limit)
+
+    clock.tick(10)
+    addPageState(PageState.ACTIVE, limit)
+
+    expect(pageStateContexts.getPageStates(0 as RelativeTime, 100 as RelativeTime)).toEqual({
+      active: { count: 1, duration: jasmine.any(Number) },
```
BenoitZugmeyer, line 106: **💬 suggestion:** ‏finish on a different state than the first state, so we can verify thats a FIFO

packages/rum-core/src/domain/contexts/pageStateContexts.spec.ts
amortemousque: In the timeline I got:
1. from 0 to 10 => state = active
2. from 10 to 20 => state = passive
3. from 20 to 30 => state = hidden
4. from 30 to infinity => state = frozen

With  `getPageStates(0 as RelativeTime, 50 as RelativeTime))`  from 30 to 50 I get `frozen`

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,157 @@
...
+  noop,
+  elapsed,
+  addDuration,
+  addEventListeners,
+  addEventListener,
+  relativeNow,
+  DOM_EVENT,
+} from '@datadog/browser-core'
+
+export const PAGE_STATE_TIMELINE_MAX_LENGTH = 2500
```
BenoitZugmeyer, line 13: **💭 thought:** ‏Having to loop on 2500 states every time a resource is collected might impact performaces, maybe consider using a smaller limit? `100` could be enough, no?

packages/rum-core/src/domain/contexts/pageStateContexts.spec.ts
BenoitZugmeyer: Ooh right, thank you :)

packages/rum-core/src/domain/contexts/pageStateContexts.ts
amortemousque: I'll try with 500

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,141 @@
...
+      DOM_EVENT.PAGE_SHOW,
+      DOM_EVENT.FOCUS,
+      DOM_EVENT.BLUR,
+      DOM_EVENT.VISIBILITY_CHANGE,
+      DOM_EVENT.RESUME,
+      DOM_EVENT.FREEZE,
+      DOM_EVENT.PAGE_HIDE,
+    ],
+    (event) => {
+      if (!event.isTrusted) return
```
bcaudan, line 63: **❓ question:** ‏what is the intent of this `isTrusted` check?

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,141 @@
...
+
+      if (event.type === DOM_EVENT.FREEZE) addPageState(PageState.FROZEN)
+      else if (event.type === DOM_EVENT.PAGE_HIDE)
+        addPageState((event as PageTransitionEvent).persisted ? PageState.FROZEN : PageState.TERMINATED)
+      else addPageState(getState())
```
bcaudan, line 68: **💭 thought:** ‏I'm a bit surprised with the lack of parenthesis there. IMO, it could be easier to read with consistent line return.
 If it is not just me, should we enforce it?

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export function addPageState(nextState: PageState, pageStateTimelineMaxLength = PAGE_STATE_TIMELINE_MAX_LENGTH) {
+  if (nextState === state) return
+
+  state = nextState
+  const now = relativeNow()
+
+  if (pageStateTimeline.length === pageStateTimelineMaxLength) pageStateTimeline.shift()
+
+  pageStateTimeline.push({ state, startTime: now })
```
bcaudan, line 135: **🥜 nitpick:** ‏inline now
```suggestion
  if (pageStateTimeline.length === pageStateTimelineMaxLength) pageStateTimeline.shift()

  pageStateTimeline.push({ state, startTime: relativeNow() })
```

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,141 @@
...
+      const endTime = addDuration(startTime, duration)
+      if (pageStateTimeline.length === 0 || pageStateTimeline[0].startTime >= endTime) return
+
+      const pageStateContext: PageStateContext = {
+        active: { count: 0, duration: 0 as Duration },
+        passive: { count: 0, duration: 0 as Duration },
+        hidden: { count: 0, duration: 0 as Duration },
+        frozen: { count: 0, duration: 0 as Duration },
+        terminated: { count: 0, duration: 0 as Duration },
+      }
```
bcaudan, line 93: **❓ question:** ‏What is the goal of collecting the count/duration of each states? should not it be more helpful to collect the corresponding timeline directly?

packages/rum-core/src/domain/contexts/pageStateContexts.ts
amortemousque: While testing the page state context locally, I saw a weird behaviour related to the fact that our chrome extension[ trigger custom  'visibilitychange' event ](https://github.com/DataDog/browser-sdk/blob/b604215075db0a483f3b6674e31c69a3fa83194b/developer-extension/src/panel/flushEvents.ts#L11). I want to prevent that.

packages/rum-core/src/domain/contexts/pageStateContexts.ts
amortemousque: If everyone agree, I'll fix it and enforce it
cc @BenoitZugmeyer @yannickadam

packages/rum-core/src/domain/contexts/pageStateContexts.ts
amortemousque: The timeline would be an array of object which is hard to query in Datadog.

packages/rum-core/src/domain/contexts/pageStateContexts.ts
yannickadam: I would also expect curly brackets { } on the else if.

packages/rum-core/src/domain/contexts/pageStateContexts.ts
bcaudan: Oh, I guess that many states were added due to the auto flush.

it could be nice to add a comment explaining the intent.

packages/rum-core/src/domain/contexts/pageStateContexts.ts
bcaudan: oh, you plan to query that, I would have expect to use this debug data only when looking at a resource with long duration.

Not sure what you would expect to get from those queries though.
Could you share your plans around that?

amortemousque: Nothing fancy, my idea was to look at resource with long duration to identify potential pattern and then validate my assumption by queying on one specific state duration for instance to have an idea of volume

amortemousque: Get the raw data

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export interface PageStateContexts {
+  getPageStates: (startTime: RelativeTime, duration: Duration) => PageStateContext | undefined
+  stop: () => void
+}
+
+let pageStateContext: PageStateContext = []
+let state: PageState | undefined
+
+export function startPageStateContexts(): PageStateContexts {
```
bcaudan, line 31: **💬 suggestion:** ‏The behaviour of this component seems quite similar to ContextHistory, if we need to keep it after the investigation, it could be interesting to mutualize some code.
In the mean time, it could be interesting to take inspiration from the naming we have there to be a bit clearer and consistent.
Something like:
```suggestion
export const enum PageState {
  ACTIVE = 'active',
  PASSIVE = 'passive',
  HIDDEN = 'hidden',
  FROZEN = 'frozen',
  TERMINATED = 'terminated',
}

export type PageStateEntry = { pageState: PageState; startTime: RelativeTime }

export interface PageStateHistory {
  findAll: (startTime: RelativeTime, duration: Duration) => PageStateEntry[] | undefined
  stop: () => void
}

let pageStateEntries: PageStateEntry[] = []
let pageState: PageState | undefined

export function startPageStateHistory(): PageStateHistory {
```

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,122 @@
...
+    return PageState.HIDDEN
+  }
+  if (document.hasFocus()) {
+    return PageState.ACTIVE
+  }
+  return PageState.PASSIVE
+}
+
+export function addPageState(nextState: PageState, pageStateContextMaxLength = PAGE_STATE_CONTEXT_MAX_LENGTH) {
+  if (nextState === state) {
```
bcaudan, line 105: **🥜 nitpick:** ‏

- it could be nice to have `currentState` to be more symmetric with `nextState`
- since `PageState` has been shorted in `State` in the variable name, what about only keeping `next` and `current`?

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  state = nextState
+  const now = relativeNow()
+
+  if (pageStateContext.length === pageStateContextMaxLength) {
+    pageStateContext.shift()
+  }
+
+  pageStateContext.push({ state, startTime: now })
```
bcaudan, line 116: **🥜 nitpick:** ‏variable could be inlined

```suggestion
  if (pageStateContext.length === pageStateContextMaxLength) {
    pageStateContext.shift()
  }

  pageStateContext.push({ state, startTime: relativeNow() })
```

packages/rum-core/src/domain/contexts/pageStateContexts.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (document.visibilityState === 'hidden') {
+    return PageState.HIDDEN
+  }
+  if (document.hasFocus()) {
+    return PageState.ACTIVE
+  }
+  return PageState.PASSIVE
+}
+
+export function addPageState(nextState: PageState, pageStateContextMaxLength = PAGE_STATE_CONTEXT_MAX_LENGTH) {
```
bcaudan, line 104: **💭 thought:** ‏overriding the limit from here seems odd, what about providing this limit in the start method instead?

packages/rum-core/src/domain/contexts/pageStateContexts.ts
amortemousque: This comment is a bit conflicting with the previous [one](https://github.com/DataDog/browser-sdk/pull/1890#discussion_r1062589880) so I change the name like so  `currentPageState` and `nextPageState`

packages/rum-core/src/domain/contexts/pageStateContexts.ts
amortemousque: While it may be a bit odd, do you think it's good enough for now? According to your [comment](https://github.com/DataDog/browser-sdk/pull/1890#discussion_r1062589880) if this code has to leave longer than the investigation it will be refactored. So we could wait that moment to refine tests.

bcaudan: Sure, worth to keep it in mind though.

bcaudan (approved):
LGTM

PR #1892, opened 2022-12-20
🔊 add view document_count in non-view events

BenoitZugmeyer:
## Motivation

Help investigate why view event counts are often incorrect.

## Changes

Add `_dd.view.document_count` reflecting the current view document count at the moment when the event is collected. In theory, the final `_dd.document_count` on View events will alway be greater than child events `_dd.view.document_count`, because any child event collected will trigger a scheduled view update. Let's verify that.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/contexts/viewContexts.ts
+++ b/packages/rum-core/src/domain/contexts/viewContexts.ts
@@ -47,7 +47,7 @@ export function startViewContexts(lifeCycle: LifeCycle): ViewContexts {
       version: view.version,
       id: view.id,
       name: view.name,
-      documentVersion: 0,
+      documentVersion: view.documentVersion,
     }
   }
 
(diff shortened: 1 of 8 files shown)


packages/rum-core/src/domain/contexts/viewContexts.ts
```diff
@@ -39,6 +47,7 @@ export function startViewContexts(lifeCycle: LifeCycle): ViewContexts {
       version: view.version,
       id: view.id,
       name: view.name,
+      documentVersion: 0,
```
bcaudan, line 50: **💬 suggestion:** ‏what about also adding `documentVersion` to the `ViewCreatedEvent`? 
it would avoid to duplicate its initialization logic

bcaudan (approved):

PR #1893, opened 2022-12-23
Fix `getInitConfiguration()` behaviour

piotr-oles:
## Motivation
Context: Synthetics Profiler doesn't tag profiles with correct service tag

I need `DD_RUM.getInitConfiguration()` to return init configuration even if it's inside synthetic test. Currently, when I call `DD_RUM.getInitConfiguration()` inside synthetic test, it returns undefined, because `initRum()` function is interrupted by:
```typescript
    if (ignoreInitIfSyntheticsWillInjectRum && willSyntheticsInjectRum()) {
      return
    }
```
which prevents from this code to be executed:
```typescript
    getInitConfigurationStrategy = () => deepClone<InitConfiguration>(initConfiguration)
```

This breaks profile tagging, as we can't infer service name (and because of that, we can't resolve source maps)

I think that `getInitConfiguration()` should return init configuration even if init process was not successful (as this function is not called `getValidInitConfiguration()` or `getInitConfigurationIfInitWasSuccessful()`)

## Changes
Update RUM and Logs API - `getInitConfiguration()` will return init configuration regardless of init process outcome.

## Testing
- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/boot/logsPublicApi.spec.ts
+++ b/packages/logs/src/boot/logsPublicApi.spec.ts
@@ -68,6 +68,11 @@ describe('logs entry', () => {
       expect(startLogs).not.toHaveBeenCalled()
     })
 
+    it("should return init configuration even if it's invalid", () => {
+      LOGS.init(INVALID_INIT_CONFIGURATION)
+      expect(LOGS.getInitConfiguration()).toEqual(INVALID_INIT_CONFIGURATION)
+    })
+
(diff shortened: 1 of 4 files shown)


BenoitZugmeyer (approved):

bcaudan (approved):
LGTM

PR #1910, opened 2023-01-02
🔊 [RUMF-1345] add telemetry logs on untrusted events

BenoitZugmeyer:
## Motivation

We want to better understand what type of [untrusted events](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) are being recorded.

Superseds https://github.com/DataDog/browser-sdk/pull/1878

## Changes

Add a telemetry debug log when an untrusted event is dispatched and caught by our SDK. We limit the number of reported events to one per event type to avoid sending too many logs.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/addEventListener.ts
+++ b/packages/core/src/browser/addEventListener.ts
@@ -99,27 +99,19 @@ export function addEventListeners<E extends Event>(
   }
 }
 
-let untrustedEventsReported: Set<string> | undefined
+const reportedUntrustedEventTypes = new Set<string>()
 
 function reportUntrustedEvent(event: Event) {
-  if (event.isTrusted || (window as any).jasmine) {
-    return
-  }
-
-  if (!untrustedEventsReported) {
-    untrustedEventsReported = new Set()
-  }
-
   const eventType = event.type
 
-  if (untrustedEventsReported.has(eventType)) {
+  if (event.isTrusted || (window as any).jasmine || reportedUntrustedEventTypes.has(eventType)) {
     return
   }
 
-  untrustedEventsReported.add(eventType)
+  reportedUntrustedEventTypes.add(eventType)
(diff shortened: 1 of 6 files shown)


packages/core/src/browser/addEventListener.ts
```diff
@@ -100,3 +98,28 @@ export function addEventListeners<E extends Event>(
...
+    return
+  }
+
+  untrustedEventsReported.add(eventType)
+  addTelemetryDebug('Untrusted event', { event_type: eventType })
+}
+
+export function resetUntrustedEventsCount() {
+  untrustedEventsReported = undefined
+}
```
yannickadam, line 125: Can we add Set as allowed in local-rules/disallow-side-effects? We could then:

```suggestion

const untrustedEventsReported = new Set<string>();

function reportUntrustedEvent(event: Event) {
  
  const eventType = event.type
  if (event.isTrusted || (window as any).jasmine || untrustedEventsReported.has(eventType) ) {
    return
  } 

  untrustedEventsReported.add(eventType)
  addTelemetryDebug('Untrusted event', { event_type: eventType })
}

export function resetUntrustedEventsCount() {
  untrustedEventsReported.clear()
}
```

packages/core/src/browser/addEventListener.ts
```diff
@@ -100,3 +98,28 @@ export function addEventListeners<E extends Event>(
     stop,
   }
 }
+
+let untrustedEventsReported: Set<string> | undefined
```
bcaudan, line 102: **🥜 nitpick:** from the naming, I would have expected a boolean indicating if untrusted events are reported or not‏
```suggestion
let reportedUntrustedEventTypes: Set<string> | undefined
```

bcaudan (approved):

yannickadam (approved):

PR #1911, opened 2023-01-02
🐛 extension: handle internal analytics domain

bcaudan:
## Motivation

Some extension features were not supported on applications using internal analytics domain:

-  Event source `Requests``
- Block intake requests

## Changes

- Move `constants.ts` to `common`
- Use intake domain constants for both `listenEventsFromRequests` and `blockIntakeRequests`
- Add `iam-rum-intake.datadoghq.com` intake domain

## Testing

- [x] Local on app.datadoghq.com and docs.datadoghq.com
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/developer-extension/src/panel/hooks/useEvents.ts
+++ b/developer-extension/src/panel/hooks/useEvents.ts
@@ -1,8 +1,9 @@
 import { useEffect, useRef, useState } from 'react'
-import { generateUUID } from '../../../../packages/core/src/tools/utils'
-import type { RumEvent } from '../../../../packages/rum-core/src/rumEvent.types'
-import type { LogsEvent } from '../../../../packages/logs/src/logsEvent.types'
 import type { TelemetryEvent } from '../../../../packages/core/src/domain/telemetry'
+import { generateUUID } from '../../../../packages/core/src/tools/utils'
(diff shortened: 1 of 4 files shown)


developer-extension/src/panel/hooks/useEvents.ts
```diff
@@ -115,17 +116,18 @@ function listenEventsFromRequests(callback: (events: StoredEvent[]) => void) {
   function beforeRequestHandler(request: chrome.devtools.network.Request) {
     const url = new URL(request.request.url)
 
-    const intake = /^(rum|logs).browser-intake-/.exec(url.hostname)
```
BenoitZugmeyer, line 118: **💬 suggestion:** ‏This was used instead of the hardcoded list of all intake domains because when that was implemented, we chose to keep the staging domain private. Now that it is [part of the codebase](https://github.com/DataDog/browser-sdk/blob/323e0726f07a50c2154bde0863b86c366b9977c1/packages/core/src/domain/configuration/intakeSites.ts#L1), I guess this rule doesn't apply anymore. Let's add it to the `INTAKE_DOMAINS` to make the extension work on staging as well.

BenoitZugmeyer (approved):

PR #1931, opened 2023-01-09
👷 attempt to fix staging-reset

BenoitZugmeyer:
## Motivation

Staging reset failed this morning. It appears to be multiple issues:
* It checks wether the branch exists on the remote, but it failed to delete the existing branch locally.
* When retried, the job failed again because it did not had any change when editing `.gitlab-ci.yml`. The retry was done on the same initial commit, so the CURRENT_BRANCH environment variable was still `staging-01`, but a commit to updathe this environment variable to `staging-02` was already pushed on  `main` branch.
* While investigating, I figured that week number computation was timezone dependent, which is a bit confusing when running parts of the script locally

## Changes

* Don't check whether the branch exists remotely. Instead, only check if it exists locally, and force push (that's what the dogweb script does)
* Read the CURRENT_BRANCH value directly from the CI file instead of the environment variable
* Make week number computation timezone independent

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/staging-ci/staging-reset.js
+++ b/scripts/staging-ci/staging-reset.js
@@ -1,12 +1,20 @@
 'use strict'
 
 const fs = require('fs')
-const { CI_FILE, initGitConfig, executeCommand, printLog, logAndExit, replaceCiVariable } = require('../utils')
+const {
+  CI_FILE,
+  initGitConfig,
+  executeCommand,
+  printLog,
+  logAndExit,
+  replaceCiVariable,
+  readCiVariable,
+} = require('../utils')
 
 const REPOSITORY = process.env.GIT_REPOSITORY
 const MAIN_BRANCH = process.env.MAIN_BRANCH
 
-const CURRENT_STAGING_BRANCH = process.env.CURRENT_STAGING
+const CURRENT_STAGING_BRANCH = readCiVariable('CURRENT_STAGING')
 const NEW_STAGING_NUMBER = getWeekNumber().toString().padStart(2, '0')
 const NEW_STAGING_BRANCH = `staging-${NEW_STAGING_NUMBER}`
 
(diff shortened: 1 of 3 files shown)


amortemousque (approved):

scripts/staging-ci/staging-reset.js
```diff
@@ -30,15 +38,21 @@ async function main() {
...
+  try {
+    await executeCommand(`git show-ref --heads ${NEW_STAGING_BRANCH}`)
+    doesStagingExistLocally = true
+  } catch {
+    doesStagingExistLocally = false
   }
+  if (doesStagingExistLocally) {
+    printLog('New staging branch already exists locally, deleting...')
+    await executeCommand(`git branch -D ${NEW_STAGING_BRANCH}`)
+  }
```
bcaudan, line 51: **🥜 nitpick:** ‏what about directly deleting inside the try/catch?

scripts/utils.js
```diff
@@ -35,6 +36,12 @@ async function initGitConfig(repository) {
   await executeCommand(`git remote set-url origin ${repository}`)
 }
 
+function readCiVariable(variableName) {
+  const regexp = new RegExp(`${variableName}: (.*)`)
+  const ciFileContent = fs.readFileSync(CI_FILE, { encoding: 'utf-8' })
+  return regexp.exec(ciFileContent)?.[1]
+}
+
 async function replaceCiVariable(variableName, value) {
```
bcaudan, line 45: **🥜 nitpick:** `‏readingCiFileVariable`/`replaceCiFileVariable` could be clearer

bcaudan (approved):

yannickadam (approved):

PR #1941, opened 2023-01-11
✨ [RUMF-1467] Collect user data telemetry

amortemousque:
## Motivation

Monitor user usages to understand the user data volume and its impact on our batch strategy:
- Measure the number of events per batch
- Measure contexts bytes volumes per batch (global context, user context, feature flags)

> Worth to review commit by commit 

## Changes
- Move `computeBytesCount` from batch to utils
- Count the context bytes of `ContextManager`
- Count the `FeatureFlagContext` bytes
- Make the `Batch` flush observable
- Collect user data telemetry

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/transport/batch.ts
```diff
@@ -2,15 +2,22 @@ import { display } from '../tools/display'
...
 import { jsonStringify, objectValues } from '../tools/utils'
 import { monitor } from '../tools/monitor'
-import type { Observable } from '../tools/observable'
+import { Observable } from '../tools/observable'
 import type { PageExitEvent } from '../browser/pageExitObservable'
 import type { HttpRequest } from './httpRequest'
 
 // https://en.wikipedia.org/wiki/UTF-8
 // eslint-disable-next-line no-control-regex
 const HAS_MULTI_BYTES_CHARACTERS = /[^\u0000-\u007F]/
```
BenoitZugmeyer, line 11: **💬 suggestion:** ‏use the code that you moved to utils?

amortemousque: Indeed that was intended :)

packages/core/src/tools/contextManager.ts
```diff
@@ -1,45 +1,72 @@
...
+
+  return {
+    compute: (context: Context) => {
+      if (bytesCount === undefined) {
+        bytesCount = !isEmptyObject(context) ? computeBytesCountImpl(jsonStringify(context)!) : 0
+      }
+      return bytesCount
+    },
+    invalidate: () => {
+      bytesCount = undefined
```
bcaudan, line 69: **💬 suggestion:** ‏FMU this component is a simple cache over `computeBytesCount(jsonStringify(x))`, what about keeping things simple and handling that directly inside the `contextManager.getByteCount()` and `featureFlagContexts.getFeatureFlagBytesCount()`?

```ts
    getBytesCount: () => {
      if (bytesCountCache === undefined) {
        bytesCountCache = computeBytesCount(jsonStringify(context))
      }
      return bytesCountCache
    },

    add: (key: string, value: any) => {
      context[key] = value as ContextValue
      bytesCountCache = undefined
    },
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -157,6 +157,8 @@ export function makeRumPublicApi(
         hasReplay: recorderApi.isRecording() ? true : undefined,
       }),
       recorderApi,
+      globalContextManager,
+      userContextManager,
```
bcaudan, line 163: **💬 suggestion:** ‏if startRum has now access to all that, `getCommonContext` could be declared there instead of here

packages/rum-core/src/boot/startRum.ts
```diff
@@ -90,6 +94,18 @@ export function startRum(
       getCommonContext,
       reportError
     )
+
+  if (batch) {
+    startUserDataTelemetry(
```
bcaudan, line 100: **💬 suggestion:** ‏what about starting this component after `startRumBatch` directly?

packages/rum-core/src/domain/startUserDataTelemetry.ts
```diff
@@ -0,0 +1,119 @@
...
+let batchMessagesCount: Measure
+let globalContextBytes: Measure
+let userContextBytes: Measure
+let featureFlagBytes: Measure
+
+let currentBatchGlobalContextBytes: Measure
+let currentBatchUserContextBytes: Measure
+let currentBatchFeatureFlagBytes: Measure
+
+export function startUserDataTelemetry(
```
bcaudan, line 36: **💬 suggestion:** ‏what about using `CustomerData` naming? it would avoid confusion with user context and be consistent with naming used in RFC, jira, ...

packages/rum-core/src/domain/startUserDataTelemetry.ts
```diff
@@ -0,0 +1,119 @@
...
+) {
+  const userDataTelemetryEnabled = telemetry.enabled && performDraw(configuration.userDataTelemetrySampleRate)
+  if (!userDataTelemetryEnabled || !isExperimentalFeatureEnabled('user_data_telemetry')) {
+    return
+  }
+
+  initMeasures()
+  initCurrentBatchMeasures()
+
+  lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event: RumEvent & Context) => {
```
bcaudan, line 53: **💬 suggestion:** ‏it could be helpful to add a comment to warn that measures here are updated for each view updates.
It should not be problematic now but it could be easily missed if we need to collect more stuff.

packages/rum-core/src/domain/startUserDataTelemetry.ts
```diff
@@ -0,0 +1,119 @@
...
+  featureFlagContexts: FeatureFlagContexts,
+  batchFlushObservable: Observable<BatchFlushEvent>
+) {
+  const userDataTelemetryEnabled = telemetry.enabled && performDraw(configuration.userDataTelemetrySampleRate)
+  if (!userDataTelemetryEnabled || !isExperimentalFeatureEnabled('user_data_telemetry')) {
+    return
+  }
+
+  initMeasures()
+  initCurrentBatchMeasures()
```
bcaudan, line 51: **💬 suggestion:** ‏it could be interesting to introduce a concept to differentiate current batch measures from the others, something like `initCurrentPeriodMeasures`.
Then, we could reinforce this period concept by using it across this file.

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: I don't really get the design you have in mind but I thought that we could remove the `getCommonContext` once the `contextManagers` will use context history.

packages/rum-core/src/domain/startUserDataTelemetry.ts
amortemousque: Let me know what you think

packages/core/src/tools/contextManager.ts
```diff
@@ -1,45 +1,72 @@
...
+
+export type ContextBytesCounter = ReturnType<typeof contextBytesCounter>
+
+export function contextBytesCounter(computeBytesCountImpl = computeBytesCount) {
+  let bytesCount: number | undefined
+
+  return {
+    compute: (context: Context) => {
+      if (bytesCount === undefined) {
+        bytesCount = !isEmptyObject(context) ? computeBytesCountImpl(jsonStringify(context)!) : 0
```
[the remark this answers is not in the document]
amortemousque: If the context is empty `computeBytesCount` will return 2 for the empty object "{}" which will pollute our measures. Especially the `min`.

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: I tryied to do as less change as possible but due to this [test](https://github.com/DataDog/browser-sdk/blob/90eee94c1ec89a208a423a4ac59c912a119ad30b/packages/rum-core/src/boot/rumPublicApi.spec.ts#L830), I moved `commonContext` logic to /context. 
Like I suspected, it polluted a bit this PR. Let me know what you think

packages/rum-core/src/domain/startUserDataTelemetry.ts
amortemousque: I don't really get your comment

packages/rum-core/src/domain/startCustomerDataTelemetry.ts
```diff
@@ -0,0 +1,143 @@
...
+    telemetry.enabled && performDraw(configuration.customerDataTelemetrySampleRate)
+  if (!customerDataTelemetrySampleRate || !isExperimentalFeatureEnabled('customer_data_telemetry')) {
+    return
+  }
+
+  initCurrentPeriodMeasures()
+  initCurrentBatchMeasures()
+
+  lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event: RumEvent & Context) => {
+    if (!isEmptyObject(globalContextManager.getContext())) {
```
BenoitZugmeyer, line 63: **💬 suggestion:** ‏`getContext` does a deepClone of the context. This is unneeded in this case, maybe use `.get()` instead? (it is deprecated, but still useful here)

BenoitZugmeyer (approved):

packages/rum-core/src/domain/startUserDataTelemetry.ts
amortemousque: I added this [comment](https://github.com/DataDog/browser-sdk/blob/81d2864f77d6b8c86af72a166587b85c3de2787b/packages/rum-core/src/domain/startCustomerDataTelemetry.ts#L62). Let me know what you think

packages/rum-core/src/domain/startCustomerDataTelemetry.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  initCurrentPeriodMeasures()
+  initCurrentBatchMeasures()
+
+  // We measure the data of every view updates even if there could only be one per batch due to the upsert
+  // It means that contexts bytes count sums can be higher than it really is
+  lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event: RumEvent & Context) => {
+    if (!isEmptyObject(globalContextManager.get())) {
+      updateMeasure(currentBatchMeasures.globalContextBytes, globalContextManager.getBytesCount())
+    }
```
bcaudan, line 67: **❓ question:** ‏In case of a `clearContext` should not we update the measure to take into account the empty context?

packages/rum-core/src/domain/startCustomerDataTelemetry.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      type: 'log',
+      status: 'debug',
+      message: 'Customer data measures',
+      batchCount: 2,
+      batchBytesCount: { min: 1, max: 2, sum: 3 },
+      batchMessagesCount: { min: 1, max: 2, sum: 3 },
+      globalContextBytes: { min: 1, max: 2, sum: 5 },
+      userContextBytes: { min: 1, max: 2, sum: 5 },
+      featureFlagBytes: { min: 1, max: 2, sum: 5 },
+    })
```
bcaudan, line 111: **🥜 nitpick:** ‏ ‏it may be easier to guess what is going on by using different magnitudes
```suggestion
    generateBatch({ eventNumber: 10, contextBytesCount: 10, batchBytesCount: 10 })
    generateBatch({ eventNumber: 1, contextBytesCount: 1, batchBytesCount: 1 })
    clock.tick(MEASURES_PERIOD_DURATION)

    expect(telemetryEvents[0].telemetry).toEqual({
      type: 'log',
      status: 'debug',
      message: 'Customer data measures',
      batchCount: 2,
      batchBytesCount: { min: 1, max: 10, sum: 11 },
      batchMessagesCount: { min: 1, max: 10, sum: 11 },
      globalContextBytes: { min: 1, max: 10, sum: 101 },
      userContextBytes: { min: 1, max: 10, sum: 101 },
      featureFlagBytes: { min: 1, max: 10, sum: 101 },
    })
```

packages/rum-core/src/domain/startCustomerDataTelemetry.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      jasmine.objectContaining({
+        batchCount: 1,
+        batchBytesCount: { min: 1, max: 1, sum: 1 },
+        batchMessagesCount: { min: 1, max: 1, sum: 1 },
+        globalContextBytes: { min: 1, max: 1, sum: 1 },
+        userContextBytes: { min: 1, max: 1, sum: 1 },
+        featureFlagBytes: { min: 1, max: 1, sum: 1 },
+      })
+    )
+  })
```
bcaudan, line 143: **💬 suggestion:** ‏could we focus the test setup/expectation to make clearer what would fail if we collect contexts telemetry of unfinished batches?

packages/rum-core/src/domain/contexts/commonContext.ts
```diff
@@ -0,0 +1,25 @@
...
+  recorderApi?: RecorderApi
+): CommonContext {
+  const commonContext: CommonContext = {
+    context: globalContextManager.getContext(),
+    user: userContextManager.getContext(),
+  }
+
+  if (recorderApi) {
+    commonContext.hasReplay = recorderApi.isRecording() ? true : undefined
+  }
```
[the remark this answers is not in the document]
amortemousque: @BenoitZugmeyer can you confirm?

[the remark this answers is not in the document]
BenoitZugmeyer: Yup!

packages/rum-core/src/domain/contexts/commonContext.ts
```diff
@@ -0,0 +1,25 @@
+import type { Context, ContextManager, User } from '@datadog/browser-core'
+import type { RecorderApi } from '../../boot/rumPublicApi'
+
+export interface CommonContext {
+  user: User
+  context: Context
+  hasReplay?: true
+}
+
+export function getCommonContext(
```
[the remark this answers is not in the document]
amortemousque: Couldn't we say the same for `displayContext`. For all our contexts, we have either `get...` or `find...`. I don't think the  small computation of hasReplay worth using `build...` but of course it's subjective.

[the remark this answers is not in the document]
bcaudan: > Couldn't we say the same for displayContext

sure :)

> For all our contexts, we have either get... or find...

As long as `get` return an existing object and `find` select an object among others, it seems legitimate

> I don't think the small computation of hasReplay worth using build... but of course it's subjective.

we are creating a new object at each call and there is indeed a bit of logic, so `build` would seem more accurate than `get` to me.

That is my reasoning for namings but it is a nitpick 🙂

bcaudan (approved):
Great!

PR #1945, opened 2023-01-12
🔊 [RUMF-1445] Check if the page was discarded before a resource

amortemousque:
## Motivation

To investigate resource with long duration, check if the page has been discarded by the browser.

## Changes

Add the document.wasDiscarded in resources

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -231,9 +231,14 @@ function computeIndexingInfo(sessionManager: RumSessionManager, resourceStart: C
...
 
 function computePageStateInfo(pageStateHistory: PageStateHistory, startClocks: ClocksState, duration: Duration) {
+  if (!isExperimentalFeatureEnabled('resource_page_states')) {
+    return
+  }
+
   return {
     _dd: {
       page_states: pageStateHistory.findAll(startClocks.relative, duration),
+      page_was_discarded: (document as any).wasDiscarded,
```
BenoitZugmeyer, line 241: **💬 suggestion:** ‏maybe `String((document as any).wasDiscarded)` so you can differentiate between true/false/undefined?

BenoitZugmeyer (approved):

packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
amortemousque: I don't see the point of getting undefined. true and false should be okay but maybe I'm missing something?

BenoitZugmeyer: Not all browsers are supporting this API, so when it doesn't, the `page_was_discarded` property would be missing, and it would be a bit less trivial to know whether this property is missing because an older SDK has been used, because the FF was not enabled or because the browser don't support it.

That's why I think it might help if you had an explicit `undefined` value here.

PR #1947, opened 2023-01-13
✨ [RUMF-1469] introduce a new `proxy` initialization parameter

BenoitZugmeyer:
## Motivation

Introduce a new proxy strategy, similar to current one but without providing the intake origin (only the URL path and parameters).

## Changes

* Deprecate the `proxyUrl` initialization parameter
* Introduce a new `proxy` initialization parameter that should be used in place of `proxyUrl`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


amortemousque:
Nice!

packages/core/src/domain/configuration/endpointBuilder.spec.ts
```diff
@@ -36,7 +37,47 @@ describe('endpointBuilder', () => {
...
+        createEndpointBuilder({ ...initConfiguration, proxy: 'https://proxy.io/path' }, 'rum', []).build('xhr')
+      ).toMatch(
+        `https://proxy.io/path\\?ddforward=${encodeURIComponent(
+          `/api/v2/rum?ddsource=(.*)&ddtags=(.*)&dd-api-key=${clientToken}` +
+            '&dd-evp-origin-version=(.*)&dd-evp-origin=browser&dd-request-id=(.*)&batch_time=(.*)'
+        )}`
+      )
+    })
+
+    it('normalizes the proxy url', () => {
```
bcaudan, line 52: **❓ question:** ‏any specific motivation to introduce this behaviour?

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -31,9 +31,8 @@ export function createEndpointBuilder(
 ) {
   const { clientToken } = initConfiguration
 
-  const host = buildEndpointHost(initConfiguration, endpointType)
-  const baseUrl = `https://${host}/api/v2/${INTAKE_TRACKS[endpointType]}`
-  const proxyUrl = initConfiguration.proxyUrl && normalizeUrl(initConfiguration.proxyUrl)
+  const baseUrl = buildEndpointBaseUrl(initConfiguration, endpointType)
+  const proxyUrl = buildProxyUrl(initConfiguration)
```
bcaudan, line 35: **💬 suggestion:** ‏IMO the overall implementation could be easier to follow by being more explicit about the different cases that we are handling.

packages/core/src/domain/configuration/endpointBuilder.spec.ts
BenoitZugmeyer: I didn't introduce it, it was already there, but not tested in this spec. This is used to prevent intake requests from being considered as a RUM resource: because PerformanceTiming resources urls are absolute, we need to make the proxy url also absolute to compare them.

packages/core/src/domain/configuration/endpointBuilder.ts
BenoitZugmeyer: I reworked things a bit. Let me know if it's clearer.

packages/core/src/domain/configuration/endpointBuilder.spec.ts
bcaudan: oh, indeed I missed that

packages/core/src/domain/configuration/endpointBuilder.ts
bcaudan: LGTM!

test/e2e/lib/framework/serverApps/intake.ts
```diff
@@ -53,7 +53,7 @@ function computeIntakeType(
   }
 
   let intakeType: IntakeType
-  const forwardUrl = new URL(ddforward)
+  const forwardUrl = new URL(ddforward, 'http://example.com')
   const endpoint = forwardUrl.pathname.split('/').pop()
```
bcaudan, line 57: **💬 suggestion:** if we only use new proxy, we could assume that ddforward is already the path, no?‏
```suggestion
  const endpoint = ddforward.split('/').pop()
```

bcaudan (approved):

test/e2e/lib/framework/serverApps/intake.ts
BenoitZugmeyer: I thought about it but we also have parameters here, so it would be a bit more work.

bcaudan: What do you mean?

bcaudan: oh, I see, what about:
```suggestion
  // ddforward=/api/v2/rum?key=value
  const endpoint = ddforward.split(/[\/\?]/)[3]
```

bcaudan (approved):

PR #1958, opened 2023-01-18
⚗🐛 [RUMF-1296] use pointerup to trigger click actions 

BenoitZugmeyer:
This ticket is part of an effort to improve dead clicks. 

* https://github.com/DataDog/browser-sdk/pull/1958 (this issue) will discard dead clicks when something happens during the 'pointerup' event (previously only things happening during the 'click' event were taken into account)
* https://github.com/DataDog/browser-sdk/pull/1979 will discard dead clicks when something happens during the 'pointerdown' event

All three fixes are behind the `dead_click_fixes` and will be released together after some dogfooding.

## Motivation

We observed that in some cases, click actions are flagged as "dead click" when the UI updates during "pointerup" or "mouseup" events (happening before the "click" event).

## Changes

This PR fixes the issue by starting a "click action" on `pointerup` instead of `click` events. This slightly changes the start of each action, but this change should be minimal when there is no `pointerup` or `mouseup` event listener.

This is behind a feature flag as I want to assess the impact on org2 before enabling it.

This PR also improve the multitouch situation.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


amortemousque (approved):

bcaudan (approved):

packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -80,17 +81,18 @@ export function trackClickActions(
 
   const { stop: stopActionEventsListener } = listenActionEvents<ClickActionBase>({
     onPointerDown: (pointerDownEvent) => processPointerDown(configuration, history, pointerDownEvent),
-    onClick: (clickActionBase, clickEvent, getUserActivity) =>
-      processClick(
+    onActionStart: (clickActionBase, startEvent, getUserActivity, getClickEventTimeStamp) =>
+      startClickAction(
```
bcaudan, line 85: **💬 suggestion:** ‏`onActionStart` naming seems a bit confusing to me since:

- we are listening action events like pointer down above and here it seems more domain related
- we are actually starting the click action just below

What about `onStartEvent` instead?

bcaudan (approved):

PR #1960, opened 2023-01-20
[REPLAY] Keep Href from link even when importing CSS

ThibautGeriz:
## Motivation

The goal is to leave to the player the opportunity to either use inline CSS or rely on external resource

## Changes

- Keep Href from link even when importing CSS
- Add missing tests on the style serialization

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

We now capture both `href` & `_cssText`

image

---

### DIFF
--- a/packages/core/test/specHelper.ts
+++ b/packages/core/test/specHelper.ts
@@ -506,3 +506,12 @@ export function interceptRequests() {
 export function isAdoptedStyleSheetsSupported() {
   return Boolean((document as any).adoptedStyleSheets)
 }
+
+export function isCSSStyleSheetConstructorSupported() {
+  try {
+    new CSSStyleSheet()
+  } catch (err) {
+    return false
+  }
+  return true
+}
--- a/packages/core/test/specHelper.ts
+++ b/packages/core/test/specHelper.ts
@@ -506,12 +506,3 @@ export function interceptRequests() {
 export function isAdoptedStyleSheetsSupported() {
   return Boolean((document as any).adoptedStyleSheets)
 }
-
-export function isCSSStyleSheetConstructorSupported() {
-  try {
-    new CSSStyleSheet()
-  } catch (err) {
-    return false
-  }
-  return true
-}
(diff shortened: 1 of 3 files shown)


packages/rum/src/domain/record/serialize.spec.ts
```diff
@@ -519,6 +529,60 @@ describe('serializeNodeWithId', () => {
...
+
+    it('serializes style node with dynamic CSS that can be fetched', () => {
+      const linkNode = document.createElement('link')
+      linkNode.setAttribute('rel', 'stylesheet')
+      linkNode.setAttribute('href', 'https://datadoghq.com/some/style.css')
+      isolatedDom.document.head.appendChild(linkNode)
+      const styleSheet = new CSSStyleSheet()
+      styleSheet.insertRule('body { width: 100%; }')
+      spyOnProperty(styleSheet, 'href', 'get').and.returnValue('https://datadoghq.com/some/style.css')
+      spyOnProperty(isolatedDom.document, 'styleSheets', 'get').and.returnValue([styleSheet])
```
ThibautGeriz, line 571: for some reason `styleSheets` is always empty, I resorted to mocking

packages/core/test/specHelper.ts
```diff
@@ -506,3 +506,12 @@ export function interceptRequests() {
...
 }
+
+export function isCSSStyleSheetConstructorSupported() {
+  try {
+    new CSSStyleSheet()
+  } catch (err) {
+    return false
+  }
+  return true
+}
```
N-Boutaib, line 517:
```suggestion
  return typeof CSSStyleSheet !== 'undefined'
}
```

ThibautGeriz: Sadly, this won't work because `CSSStyleSheet`  will be defined, it's a old API but the constructor is relatively recent, see: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet

amortemousque (approved):

acorretti (approved):

BenoitZugmeyer (approved):

N-Boutaib (approved):

PR #1978, opened 2023-01-27
[REPLAY] Add telemetry for shadow dom

ThibautGeriz:
## Motivation

[REPLAY] Add telemetry for shadow dom #1978

## Changes

- add shadow dom content if FF is present (only to be activated on org2 for privacy concerned)

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/shadowRootsController.ts
+++ b/packages/rum/src/domain/record/shadowRootsController.ts
@@ -1,4 +1,4 @@
-import { addTelemetryDebug, DOM_EVENT } from '@datadog/browser-core'
+import { addTelemetryDebug, DOM_EVENT, isExperimentalFeatureEnabled } from '@datadog/browser-core'
 import type { RumConfiguration } from '@datadog/browser-rum-core'
 import { startMutationObserver } from './mutationObserver'
 import { initInputObserver } from './observers'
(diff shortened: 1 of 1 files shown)


packages/rum/src/domain/record/shadowRootsController.ts
```diff
@@ -57,6 +57,8 @@ export const initShadowRootsController = (
         addTelemetryDebug('no shadow root in map', {
           shadowRoot: shadowRoot ? shadowRoot.nodeName : 'no node name',
           childrenLength: shadowRoot ? shadowRoot.childElementCount : '-1',
+          controllerByShadowRootSize: controllerByShadowRoot.size,
+          html: shadowRoot && isExperimentalFeatureEnabled('shadow_dom_debug') ? shadowRoot.innerHTML : undefined,
```
bcaudan, line 61: **💬 suggestion:** ‏it could worth truncating the innerHTML to something like 1-2k to avoid sending huge payload

bcaudan, line 61: **💬 suggestion:** ‏as discussed together, you could add `view.name` here, if it can help you detect some pattern

ThibautGeriz: How can I get that here?

BenoitZugmeyer: Getting the view.name might be non-trivial here (either passing the ViewContext or tracking the current view with LifeCycle events could be a bit overkill for such investigation)

I shared with you a way to make it easier to navigate to the related view for telemetry events, maybe it could be enough?

bcaudan: a dirty way (for temporary debug sake) could be to access it from the `DD_RUM.getInternalContext().view.name` 🙈

ThibautGeriz: honestly I can do without, I would rather merge this PR (on Monday) in its current state than trying to get the perfect attributes

bcaudan (approved):

PR #1979, opened 2023-01-27
⚗🐛 [RUMF-1293] discard dead clicks when activity occurs on pointerdown

BenoitZugmeyer:
This ticket is part of an effort to improve dead clicks. 

* https://github.com/DataDog/browser-sdk/pull/1958 will discard dead clicks when something happens during the 'pointerup' event (previously only things happening during the 'click' event were taken into account)
* https://github.com/DataDog/browser-sdk/pull/1979 (this issue) will discard dead clicks when something happens during the 'pointerdown' event

All three fixes are behind the `dead_click_fixes` and will be released together after some dogfooding.

## Motivation

We observed that in some cases, click actions are flagged as "dead click" when the UI updates during "pointerdown" or "mousedown" events (happening before the "click" event).

## Changes

With this PR, we'll observe page activity occuring during the "pointerdown" event, and will discard the "dead click" frustration if some activity occurs.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
@@ -19,7 +19,7 @@ import type { RumConfiguration } from '../../configuration'
 import type { LifeCycle } from '../../lifeCycle'
 import { LifeCycleEventType } from '../../lifeCycle'
 import { trackEventCounts } from '../../trackEventCounts'
-import { waitPageActivityEnd } from '../../waitPageActivityEnd'
+import { PAGE_ACTIVITY_VALIDATION_DELAY, waitPageActivityEnd } from '../../waitPageActivityEnd'
 import type { ClickChain } from './clickChain'
 import { createClickChain } from './clickChain'
 import { getActionNameFromElement } from './getActionNameFromElement'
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/rumEventsCollection/action/trackClickActions.ts
```diff
@@ -174,7 +203,17 @@ function startClickAction(
...
       } else {
-        click.stop(pageActivityEndEvent.hadActivity ? pageActivityEndEvent.end : undefined)
+        if (pageActivityEndEvent.hadActivity) {
+          click.stop(pageActivityEndEvent.end)
+        } else if (hadActivityOnPointerDown()) {
+          click.stop(
+            // using the click start as activity end, so the click will have some activity but its
+            // duration will be 0 (as the activity started before the click start)
+            click.startClocks.timeStamp
+          )
```
bcaudan, line 213: **💭 thought:** ‏IIUC this case can happen if some activity starts at pointer down but no other activity starts at pointer up. 

- to handle correctly this particular case, it could be nice to not stop tracking the activity from pointer down since we could have pending requests
- more generally, should the action start always be at pointer down?

BenoitZugmeyer: > **💭 thought:** ‏IIUC this case can happen if some activity starts at pointer down but no other activity starts at pointer up.
> 
>     * to handle correctly this particular case, it could be nice to not stop tracking the activity from pointer down since we could have pending requests

Since we are not associating pending requests to the Click Action, I think it might be a bit confusing UI-wise to use them for the click duration. But we can talk about it


> 
>     * more generally, should the action start always be at pointer down?

As previously discussed in the RFC, this was discouraged, but we can still chalenge it. I think it would make this change more noticeable though.

bcaudan (approved):

PR #1985, opened 2023-01-30
🐛[RUMF-1450] stop computing coordinates for focus/blur records

bcaudan:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


amortemousque (approved):

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -328,3 +333,64 @@ describe('initStyleSheetObserver', () => {
...
+
+  afterEach(() => {
+    sandbox.remove()
+    delete (window.visualViewport as any).offsetTop
+    stopObserver()
+  })
+
+  it('should compute x/y coordinates for click record', () => {
+    a.click()
+    expect(mouseInteractionCallbackSpy).toHaveBeenCalled()
```
ThibautGeriz, line 386: **suggestion(not-blocking):** use `toHaveBeenCalledWith` to check that x and y are not present in the payload

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -328,3 +333,64 @@ describe('initStyleSheetObserver', () => {
...
+    expect(coordinatesComputed).toBeTrue()
+  })
+
+  // related to safari issue, see RUMF-1450
+  it('should not compute x/y coordinates for blur record', () => {
+    a.blur()
+    expect(mouseInteractionCallbackSpy).toHaveBeenCalled()
```
ThibautGeriz, line 393: **suggestion(not-blocking):** use `toHaveBeenCalledWith` to check that x and y are present in the payload

ThibautGeriz (approved):
LGTM for replay (I did not QA change)

packages/rum/src/domain/record/observers.spec.ts
bcaudan: I was mostly relying on types for that but since we are lacking tests in this area, it would not hurt 🙂

packages/rum/src/domain/record/observers.ts
```diff
@@ -203,22 +203,20 @@ function initMouseInteractionObserver(
...
+    const id = getSerializedNodeId(target)
+    const type = eventTypeToMouseInteraction[event.type as keyof typeof eventTypeToMouseInteraction]
+
+    let interaction: MouseInteraction
+    if (type !== MouseInteractionType.Blur && type !== MouseInteractionType.Focus) {
+      const { x, y } = computeCoordinates(event)
+      interaction = { id, type, x, y }
+    } else {
+      interaction = { id, type }
     }
```
BenoitZugmeyer, line 215: **💬 suggestion:** ‏we know that MouseEvent/TouchEvents might not have coordinates. While we are working on this, what about checking whether `clientX` / `clientY` are not undefined before proceeding to compute coordinates? I know @ThibautGeriz would be very happy about that :) 

A test could look like this:

```ts

 it('should ignore clicks events without coordinates', () => {
    a.dispatchEvent(new Event('click', { bubbles: true }))
    expect(mouseInteractionCallbackSpy).not.toHaveBeenCalled()
    expect(coordinatesComputed).toBeFalse()
  })
```

packages/rum/src/domain/record/observers.spec.ts
```diff
@@ -328,3 +333,84 @@ describe('initStyleSheetObserver', () => {
...
+  let a: HTMLAnchorElement
+  let coordinatesComputed: boolean
+
+  beforeEach(() => {
+    if (isIE()) {
+      pending('IE not supported')
+    }
+    if (!window.visualViewport) {
+      pending('no visualViewport')
+    }
```
BenoitZugmeyer, line 350: **🥜 nitpick:** ‏those tests should pass also on browsers that lacks visual viewport support. Whether `visualViewport` is used is an implementation detail.

packages/rum/src/domain/record/observers.ts
ThibautGeriz: I was planning to do a separate PR about this if you do not want to add more scope to this one as you want Bastien 🤷

packages/rum/src/domain/record/observers.spec.ts
bcaudan: My intent with those tests was to ensure that blur events was not accessing visualViewort API since it is the one that trigger a forced layout and is related to the safari bug and prevent implementation s where coordinates were computed for those events but dropped before generating the record.
In cases where there is no visualViewport API, there is no issue.

I could clarify the intent in the test by replacing `compute x/y coordinates` references by `force layout`, wdyt?

packages/rum/src/domain/record/observers.ts
bcaudan: I'd be in favour of not handling it in this PR because:

- it would delay the resolution of this issue that prevents the customer raising the issue to use session replay
- it would make sense for the replay team to own this fix and monitor its impact on their telemetry

BenoitZugmeyer: Fine, let's keep things like that for now.

packages/rum/src/domain/record/observers.spec.ts
BenoitZugmeyer: Mmh I would keep general tests on `initMouseInteractionObserver` and add a section specifically for the `force layout` issue that is run only on browsers supporting visualviewport.

bcaudan: Sounds good to me

BenoitZugmeyer (approved):

PR #1993, opened 2023-02-02
[REPLAY] Discard mouse/touch event without x/y position

ThibautGeriz:
## Motivation

We have a bunch of error in the player where we have mouse record without x/y with numeric value. 

The hypothesis is that the event are not generated by the browser but by JS using `dispatchEvent` 

## Changes

- Discard mouse/touch event without x/y position
- Add telemetry with `isTrusted` to validate/invalidate the theory

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/observers.ts
+++ b/packages/rum/src/domain/record/observers.ts
@@ -9,6 +9,7 @@ import {
   addEventListeners,
   addEventListener,
   noop,
+  addTelemetryDebug,
 } from '@datadog/browser-core'
 import type { LifeCycle, RumConfiguration } from '@datadog/browser-rum-core'
 import {
@@ -61,7 +62,7 @@ type GroupingCSSRuleTypes = typeof CSSGroupingRule | typeof CSSMediaRule | typeo
 
 type ListenerHandler = () => void
 
-type MousemoveCallBack = (
+export type MousemoveCallBack = (
   p: MousePosition[],
   source: typeof IncrementalSource.MouseMove | typeof IncrementalSource.TouchMove
 ) => void
@@ -151,23 +152,26 @@ export function initMutationObserver(
   return startMutationObserver(cb, configuration, shadowRootsController, document)
 }
 
-function initMoveObserver(cb: MousemoveCallBack): ListenerHandler {
+export function initMoveObserver(cb: MousemoveCallBack): ListenerHandler {
   const { throttled: updatePosition } = throttle(
     monitor((event: MouseEvent | TouchEvent) => {
       const target = getEventTarget(event)
       if (hasSerializedNode(target)) {
-        const { clientX, clientY } = isTouchEvent(event) ? event.changedTouches[0] : event
+        const { x, y } = computeCoordinates(event)
+        if (!Number.isFinite(x) || !Number.isFinite(y)) {
+          addTelemetryDebug('mouse/touch event without x/y', {
+            isTrusted: event.isTrusted,
+            position: { x, y },
+          })
+          return
+        }
         const position: MousePosition = {
           id: getSerializedNodeId(target),
           timeOffset: 0,
-          x: clientX,
-          y: clientY,
+          x,
+          y,
         }
-        if (window.visualViewport) {
-          const { visualViewportX, visualViewportY } = convertMouseEventToLayoutCoordinates(clientX, clientY)
-          position.x = visualViewportX
-          position.y = visualViewportY
-        }
+
         cb([position], isTouchEvent(event) ? IncrementalSource.TouchMove : IncrementalSource.MouseMove)
       }
     }),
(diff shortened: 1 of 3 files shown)


N-Boutaib (approved):

packages/rum/src/domain/record/observers.ts
```diff
@@ -151,23 +152,26 @@ export function initMutationObserver(
...
     monitor((event: MouseEvent | TouchEvent) => {
       const target = getEventTarget(event)
       if (hasSerializedNode(target)) {
-        const { clientX, clientY } = isTouchEvent(event) ? event.changedTouches[0] : event
+        const { x, y } = computeCoordinates(event)
+        if (!Number.isFinite(x) || !Number.isFinite(y)) {
+          addTelemetryDebug('mouse/touch event without x/y', {
+            isTrusted: event.isTrusted,
+            position: { x, y },
+          })
```
BenoitZugmeyer, line 165: **❓ question:** ‏ What will those telemetry logs used for?


**💬 suggestion:** ‏ If it's to see whether ignoring untrusted events would help, I would suggest emiting a telemetry log only when the event is untrusted. This would reduce the amount of unneeded logs messages.

```suggestion
          if (event.isTrusted) {
            addTelemetryDebug('mouse/touch event without x/y', {
              position: { x, y },
            })
          }
```

packages/rum/src/domain/record/observers.ts
```diff
@@ -151,23 +152,26 @@ export function initMutationObserver(
...
 
-function initMoveObserver(cb: MousemoveCallBack): ListenerHandler {
+export function initMoveObserver(cb: MousemoveCallBack): ListenerHandler {
   const { throttled: updatePosition } = throttle(
     monitor((event: MouseEvent | TouchEvent) => {
       const target = getEventTarget(event)
       if (hasSerializedNode(target)) {
-        const { clientX, clientY } = isTouchEvent(event) ? event.changedTouches[0] : event
+        const { x, y } = computeCoordinates(event)
+        if (!Number.isFinite(x) || !Number.isFinite(y)) {
```
BenoitZugmeyer, line 161: **💬 suggestion:** ‏move this check in `computeCoordinates` to avoit repeating it.

packages/rum/src/domain/record/observers.ts
ThibautGeriz: I disagree, if we have event without x/y that are both trusted and not trusted I would like to know about it

packages/rum/src/domain/record/observers.ts
ThibautGeriz: So you would do something something like that?
```ts
const { x, y, areCoordinatesCorrect } = computeCoordinates(event)
if (areCoordinatesCorrect) {
```

packages/rum/src/domain/record/observers.ts
ThibautGeriz: > What will those telemetry logs used for?

I want to know if the next major release will fix this issue for and we can remove that code

packages/rum/src/domain/record/observers.ts
BenoitZugmeyer: Maybe something like:
```ts
const coordinates = tryComputeCoordinates(event)
if (!coordinates) {
  return
}
// use coordinates without destructuring assignment
```

packages/rum/src/domain/record/observers.ts
BenoitZugmeyer: We know that some untrusted events don't have coordinates. I don't think we need to confirm that anymore, since we can easily reproduce it.

Now, what I'm worried about is the amount of telemetry logs that could be sent. If you still think that we should add a log when an untrusted event don't have coordinates, I would suggest to limit the amount of telemetry debug by only logging the first one, something like:

```ts
let wasEventWithoutCoordinatesLogSent = false

function tryComputeCoordinates(event) {
  ...
  if (!wasEventWithoutCoordinatesLogSent && !Number.isFinite(...) && ...) {
    wasEventWithoutCoordinatesLogSent = true
    addTelemetryDebug('mouse/touch event without x/y', {
      isTrusted: event.isTrusted,
      position: { x, y },
    })
  }
  ...
}
```

packages/rum/src/domain/record/observers.ts
```diff
@@ -151,23 +152,25 @@ export function initMutationObserver(
...
   const { throttled: updatePosition } = throttle(
     monitor((event: MouseEvent | TouchEvent) => {
       const target = getEventTarget(event)
       if (hasSerializedNode(target)) {
-        const { clientX, clientY } = isTouchEvent(event) ? event.changedTouches[0] : event
+        const coordinates = tryComputeCoordinates(event)
+        if (!coordinates) {
+          addTelemetryDebug('mouse/touch event without x/y', {
+            isTrusted: event.isTrusted,
+          })
```
BenoitZugmeyer, line 164: **🥜 nitpick:** ‏ ‏you could also move the telemetry debug in `tryComputeCoordinates`

packages/rum/src/domain/record/observers.ts
ThibautGeriz: Alright, I see your point, I will only send the telemetry if event is trusted

BenoitZugmeyer (approved):

packages/rum/src/domain/record/observers.ts
```diff
@@ -226,13 +229,19 @@ export function initMouseInteractionObserver(
   }).stop
 }
 
-function computeCoordinates(event: MouseEvent | TouchEvent) {
+function tryComputeCoordinates(event: MouseEvent | TouchEvent) {
```
bcaudan, line 232: **🥜 nitpick:** small naming improvement‏
```suggestion
function tryToComputeCoordinates(event: MouseEvent | TouchEvent) {
```

bcaudan (approved):

PR #1998, opened 2023-02-06
🐛 [RUMF-1296][RUMF-1293] fix dead click computation

BenoitZugmeyer:
## Motivation

This PR enables various fixes related to dead click computation:

* https://github.com/DataDog/browser-sdk/pull/1958 will discard dead clicks when something happens during the 'pointerup' event (previously only things happening during the 'click' event were taken into account)
* https://github.com/DataDog/browser-sdk/pull/1979 will discard dead clicks when something happens during the 'pointerdown' event
* https://github.com/DataDog/browser-sdk/pull/1986 will discard dead clicks triggered by manipulating content of `contenteditable` elements

## Changes

* Remove experimental flags
* Adjust tests

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
@@ -1,5 +1,4 @@
-import type { TimeStamp } from '@datadog/browser-core'
-import { addEventListener, DOM_EVENT, isExperimentalFeatureEnabled, timeStampNow } from '@datadog/browser-core'
+import { addEventListener, DOM_EVENT } from '@datadog/browser-core'
 
 export type MouseEventOnElement = PointerEvent & { target: Element }
 
@@ -9,12 +8,7 @@ export interface UserActivity {
 }
 export interface ActionEventsHooks<ClickContext> {
   onPointerDown: (event: MouseEventOnElement) => ClickContext | undefined
-  onStartEvent: (
-    context: ClickContext,
-    event: MouseEventOnElement,
-    getUserActivity: () => UserActivity,
-    getClickEventTimeStamp: () => TimeStamp | undefined
-  ) => void
+  onStartEvent: (context: ClickContext, event: MouseEventOnElement, getUserActivity: () => UserActivity) => void
 }
 
 export function listenActionEvents<ClickContext>({ onPointerDown, onStartEvent }: ActionEventsHooks<ClickContext>) {
@@ -60,24 +54,8 @@ export function listenActionEvents<ClickContext>({ onPointerDown, onStartEvent }
         if (isValidPointerEvent(startEvent) && clickContext) {
           // Use a scoped variable to make sure the value is not changed by other clicks
           const localUserActivity = userActivity
-          let clickEventTimeStamp: TimeStamp | undefined
-          onStartEvent(
-            clickContext,
-            startEvent,
-            () => localUserActivity,
-            () => clickEventTimeStamp
-          )
+          onStartEvent(clickContext, startEvent, () => localUserActivity)
           clickContext = undefined
-          if (isExperimentalFeatureEnabled('dead_click_fixes')) {
-            addEventListener(
-              window,
-              DOM_EVENT.CLICK,
-              () => {
-                clickEventTimeStamp = timeStampNow()
-              },
-              { capture: true, once: true }
-            )
-          }
         }
       },
       { capture: true }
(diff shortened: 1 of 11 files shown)


packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
```diff
@@ -1,20 +1,14 @@
...
 }
 export interface ActionEventsHooks<ClickContext> {
   onPointerDown: (event: MouseEventOnElement) => ClickContext | undefined
-  onStartEvent: (
-    context: ClickContext,
-    event: MouseEventOnElement,
-    getUserActivity: () => UserActivity,
-    getClickEventTimeStamp: () => TimeStamp | undefined
-  ) => void
+  onStartEvent: (context: ClickContext, event: MouseEventOnElement, getUserActivity: () => UserActivity) => void
```
bcaudan, line 11: **💬 suggestion:** ‏should it be `onPointerUp:` now?

packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
```diff
@@ -109,14 +83,11 @@ function isSelectionEmpty(): boolean {
...
   return (
     event.target instanceof Element &&
     // Only consider 'primary' pointer events for now. Multi-touch support could be implemented in
     // the future.
-    // On Chrome, click events are PointerEvent with `isPrimary = false`, but we should still
-    // consider them valid. This could be removed when we enable the `click-action-on-pointerup`
-    // flag, since we won't rely on click events anymore.
-    (event.type === 'click' || (event as PointerEvent).isPrimary !== false)
+    event.isPrimary !== false
   )
```
bcaudan, line 92: **🥜 nitpick:** ‏do we need to wrap the condition by parentheses?

**❓ question:** ‏could we directly use `event.isPrimary` or do we need to explicitly check `event.isPrimary !== false?`

bcaudan (approved):

packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
BenoitZugmeyer: Yes, could be simpler to follow.

packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
BenoitZugmeyer: > **🥜 nitpick:** ‏do we need to wrap the condition by parentheses?

There are parentheses because the condition is on multiple lines and prettier format it like that. I could move the comment above the `return` statement, but since it's only related to `isPrimary` I think it would be better to keep it like that

> 
> **❓ question:** ‏could we directly use `event.isPrimary` or do we need to explicitly check `event.isPrimary !== false?`

we could, but then when emulating clicks (ex: in tests in the SDK or outside of it, in Synthetics test for the demo org...) we would need to always set `isPrimary: true` on those events, which is not a huge deal but still something more to take into account

bcaudan: ok!

packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
```diff
@@ -59,29 +49,13 @@ export function listenActionEvents<ClickContext>({ onPointerDown, onStartEvent }
 
     addEventListener(
       window,
-      isExperimentalFeatureEnabled('dead_click_fixes') ? DOM_EVENT.POINTER_UP : DOM_EVENT.CLICK,
-      (startEvent: MouseEvent) => {
-        if (isValidMouseEvent(startEvent) && clickContext) {
+      DOM_EVENT.POINTER_UP,
+      (startEvent: PointerEvent) => {
+        if (isValidPointerEvent(startEvent) && clickContext) {
```
bcaudan, line 54: **🥜 nitpick:** ‏for symmetry with pointer down
```suggestion
      (event: PointerEvent) => {
        if (isValidPointerEvent(event) && clickContext) {
```

packages/rum-core/src/domain/rumEventsCollection/action/listenActionEvents.ts
```diff
@@ -59,29 +49,13 @@ export function listenActionEvents<ClickContext>({ onPointerDown, onStartEvent }
...
           // Use a scoped variable to make sure the value is not changed by other clicks
           const localUserActivity = userActivity
-          let clickEventTimeStamp: TimeStamp | undefined
-          onStartEvent(
-            clickContext,
-            startEvent,
-            () => localUserActivity,
-            () => clickEventTimeStamp
-          )
+          onPointerUp(clickContext, startEvent, () => localUserActivity)
```
bcaudan, line 57: following previous comment
```suggestion
          onPointerUp(clickContext, event, () => localUserActivity)
```

bcaudan (approved):

PR #2017, opened 2023-02-15
:bug:[RUMF-1493] Avoid infinite loop on `form >  input[name="host"]` element

ThibautGeriz:
## Motivation

Fix freeze in serialization because of infinite loop

## Changes

Make sure to not count form element with child input as name host as a shadow root

## Testing

- [x] Local (on the client website directly)
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/browser/htmlDomUtils.spec.ts
+++ b/packages/rum-core/src/browser/htmlDomUtils.spec.ts
@@ -81,14 +81,14 @@ if (!isIE()) {
       expect(isNodeShadowRoot(shadowRoot)).toBe(true)
     })
 
-    it('should return false for a[href] since it has a host property', () => {
+    it('should return false for a[href] despite it has a host property', () => {
       const link = document.createElement('a')
       link.setAttribute('href', 'http://localhost/some/path')
       expect(link.host).toBeTruthy()
       expect(isNodeShadowRoot(link)).toBe(false)
     })
 
-    it('should return false for form > input[name="host"] since the host property will be added to the form element', () => {
+    it('should return false for a form with an input[name="host"] despite it has a host property', () => {
       const form = document.createElement('form')
       const input = document.createElement('input')
       input.setAttribute('name', 'host')
--- a/packages/rum-core/src/browser/htmlDomUtils.spec.ts
+++ b/packages/rum-core/src/browser/htmlDomUtils.spec.ts
@@ -59,26 +59,41 @@ describe('isElementNode', () => {
 
 if (!isIE()) {
   describe('isShadowRoot', () => {
-    const parent = document.createElement('div')
-    parent.attachShadow({ mode: 'open' })
-    const form = document.createElement('form')
-    const input = document.createElement('input')
-    input.setAttribute('name', 'host')
-    form.appendChild(input)
-    const parameters: Array<[Node, boolean]> = [
-      [parent.shadowRoot!, true],
-      [form, false],
-      [parent, false],
-      [document.body, false],
-      [input, false],
-      [document.createTextNode('hello'), false],
-      [document.createComment('hello'), false],
+    const notShadowDomNodes: Node[] = [
+      document,
+      document.head,
+      document.body,
+      document.createElement('div'),
+      document.createTextNode('hello'),
+      document.createComment('hello'),
     ]
 
-    parameters.forEach(([element, result]) => {
-      it(`should return ${String(result)} for "${String(element)}"`, () => {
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/browser/htmlDomUtils.spec.ts
```diff
@@ -61,10 +61,16 @@ if (!isIE()) {
   describe('isShadowRoot', () => {
     const parent = document.createElement('div')
     parent.attachShadow({ mode: 'open' })
+    const form = document.createElement('form')
+    const input = document.createElement('input')
+    input.setAttribute('name', 'host')
+    form.appendChild(input)
     const parameters: Array<[Node, boolean]> = [
       [parent.shadowRoot!, true],
+      [form, false],
```
ThibautGeriz, line 70: **note:** this one would fail without changing the code

packages/rum-core/src/browser/htmlDomUtils.spec.ts
```diff
@@ -61,10 +61,16 @@ if (!isIE()) {
   describe('isShadowRoot', () => {
     const parent = document.createElement('div')
     parent.attachShadow({ mode: 'open' })
+    const form = document.createElement('form')
+    const input = document.createElement('input')
+    input.setAttribute('name', 'host')
+    form.appendChild(input)
```
bcaudan, line 67: **💬 suggestion:** ‏it took me a bit of time to understand what was going on here, it may be interesting to extract a dedicated test for it:
- we could clarify this case in the test name
- it would isolate this test setup from the simpler ones

packages/rum-core/src/browser/htmlDomUtils.spec.ts
```diff
@@ -61,10 +61,16 @@ if (!isIE()) {
...
     const parameters: Array<[Node, boolean]> = [
       [parent.shadowRoot!, true],
+      [form, false],
       [parent, false],
       [document.body, false],
+      [input, false],
```
bcaudan, line 73: **❓ question:** not sure to get why we should add a check on this one, ‏what was your intent there?

ThibautGeriz: I added mostly because it was easy. I feel we should add test on a bunch of tags. For example `<a href="http://localhost/some/place.html"></a>` have a `host` property but it's not an element therefore it was not failing.
I think we would get a lot of value from property testing, were there discussion in your team about adding it?

bcaudan: Indeed, that sounds like a nice test case to add!
It could definitely be interesting to have a test case for all possible combination of `has host attribute`, `is doc fragment` and `is element`. 

IMO, the current parametrised strategy makes a bit hard to understand what exactly we want to test for each case, it may be interesting to move to dedicated tests to better express what is interesting about each case. 

About property based testing, I don't think we had chat about it but that could definitely be something to explore at some point 👍

packages/rum-core/src/browser/htmlDomUtils.spec.ts
```diff
@@ -59,21 +59,42 @@ describe('isElementNode', () => {
...
     })
+
+    it('should return true for shadow root but not its host', () => {
+      const parent = document.createElement('div')
+      const shadowRoot = parent.attachShadow({ mode: 'open' })
+      expect(isNodeShadowRoot(parent)).toBe(false)
+      expect(isNodeShadowRoot(shadowRoot)).toBe(true)
+    })
+
+    it('should return false for a[href] since it has a host property', () => {
```
bcaudan, line 84: **🥜 nitpick:** I'd say that it returns false since it is not a document fragment
‏
```suggestion
    it('should return false for a[href] despite it has a host property', () => {
```

packages/rum-core/src/browser/htmlDomUtils.spec.ts
```diff
@@ -59,21 +59,42 @@ describe('isElementNode', () => {
...
+      const link = document.createElement('a')
+      link.setAttribute('href', 'http://localhost/some/path')
+      expect(link.host).toBeTruthy()
+      expect(isNodeShadowRoot(link)).toBe(false)
+    })
+
+    it('should return false for form > input[name="host"] since the host property will be added to the form element', () => {
```
bcaudan, line 91: **🥜 nitpick:** ‏same reasoning than the above comment
```suggestion
    it('should return false for a form with an input[name="host"] despite it has a host property', () => {
```

bcaudan (approved):

PR #2020, opened 2023-02-17
[RUMF-1484] ⚗️ use pagehide as unload event

amortemousque:
## Motivation

Use `pagehide` instead of `beforeunlaod` to:

- Prevent from not being bfcache in firefox
- Collect events / DOM mutations happening between `beforeunload` and `pagehide`
- Prevent from being wrongly notified if the unload is canceled by the user. The view won’t be wrongly ended. 

## Assess impact

To assess the impact:

- Enable `collect_flush_reason` ff to collect it for RUM events and replay segments
- Enable `pagehide`  ff for 50% of users and compare both data sets

## Changes

- pageExitObservablt use pagehide instead beforeunload (under `pagehide` ff)
- collect flush reason as a tag (under `collect_flush_reason` ff)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/pageExitObservable.ts
+++ b/packages/core/src/browser/pageExitObservable.ts
@@ -1,10 +1,12 @@
+import { isExperimentalFeatureEnabled } from '../domain/configuration'
 import { Observable } from '../tools/observable'
-import { includes, objectValues } from '../tools/utils'
-import { addEventListener, addEventListeners, DOM_EVENT } from './addEventListener'
+import { includes, noop, objectValues } from '../tools/utils'
+import { addEventListeners, addEventListener, DOM_EVENT } from './addEventListener'
 
 export const PageExitReason = {
   HIDDEN: 'visibility_hidden',
   UNLOADING: 'before_unload',
+  PAGEHIDE: 'pagehide',
   FROZEN: 'page_frozen',
 } as const
 
@@ -16,11 +18,17 @@ export interface PageExitEvent {
 
 export function createPageExitObservable(): Observable<PageExitEvent> {
   const observable = new Observable<PageExitEvent>(() => {
+    const pagehideEnabled = isExperimentalFeatureEnabled('pagehide')
     const { stop: stopListeners } = addEventListeners(
       window,
-      [DOM_EVENT.VISIBILITY_CHANGE, DOM_EVENT.FREEZE],
+      [DOM_EVENT.VISIBILITY_CHANGE, DOM_EVENT.FREEZE, DOM_EVENT.PAGE_HIDE],
       (event) => {
-        if (event.type === DOM_EVENT.VISIBILITY_CHANGE && document.visibilityState === 'hidden') {
+        if (event.type === DOM_EVENT.PAGE_HIDE && pagehideEnabled) {
+          /**
+           * Only event that detect page unload events while being compatible with the back/forward cache (bfcache)
+           */
+          observable.notify({ reason: PageExitReason.PAGEHIDE })
+        } else if (event.type === DOM_EVENT.VISIBILITY_CHANGE && document.visibilityState === 'hidden') {
           /**
            * Only event that guarantee to fire on mobile devices when the page transitions to background state
            * (e.g. when user switches to a different application, goes to homescreen, etc), or is being unloaded.
@@ -37,14 +45,12 @@ export function createPageExitObservable(): Observable<PageExitEvent> {
       { capture: true }
     )
 
-    /**
-     * Safari does not support yet to send a request during:
-     * - a visibility change during doc unload (cf: https://bugs.webkit.org/show_bug.cgi?id=194897)
-     * - a page hide transition (cf: https://bugs.webkit.org/show_bug.cgi?id=188329)
-     */
-    const { stop: stopBeforeUnloadListener } = addEventListener(window, DOM_EVENT.BEFORE_UNLOAD, () => {
-      observable.notify({ reason: PageExitReason.UNLOADING })
-    })
+    let stopBeforeUnloadListener = noop
+    if (!pagehideEnabled) {
+      stopBeforeUnloadListener = addEventListener(window, DOM_EVENT.BEFORE_UNLOAD, () => {
+        observable.notify({ reason: PageExitReason.UNLOADING })
+      }).stop
+    }
 
     return () => {
       stopListeners()
--- a/packages/core/src/browser/pageExitObservable.ts
+++ b/packages/core/src/browser/pageExitObservable.ts
@@ -6,7 +6,7 @@ import { addEventListeners, addEventListener, DOM_EVENT } from './addEventListen
 export const PageExitReason = {
   HIDDEN: 'visibility_hidden',
   UNLOADING: 'before_unload',
-  PAGEHIDE: 'pagehide',
+  PAGEHIDE: 'page_hide',
   FROZEN: 'page_frozen',
 } as const
 
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -1,9 +1,10 @@
-import type { RetryInfo } from '../../transport'
+import type { RetryInfo, FlushReason } from '../../transport'
 import { timeStampNow } from '../../tools/timeUtils'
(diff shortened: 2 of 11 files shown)


packages/core/src/transport/batch.ts
```diff
@@ -70,12 +78,12 @@ export class Batch {
...
     }
     if (this.willReachedBytesLimitWith(messageBytesCount)) {
-      this.flush()
+      this.flush('batch_bytes_limit')
     }
 
     this.push(processedMessage, messageBytesCount, key)
     if (this.isFull()) {
-      this.flush()
+      this.flush('batch_bytes_limit')
```
BenoitZugmeyer, line 86: **💬 suggestion:** ‏We could have two flush reason, one for the bytes limit and the other for messages limit

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -32,8 +33,15 @@ export function createEndpointBuilder(
   const buildUrlWithParameters = createEndpointUrlWithParametersBuilder(initConfiguration, endpointType)
 
   return {
-    build(api: 'xhr' | 'fetch' | 'beacon', retry?: RetryInfo) {
-      const parameters = buildEndpointParameters(initConfiguration, endpointType, configurationTags, api, retry)
+    build(api: 'xhr' | 'fetch' | 'beacon', flushReason?: FlushReason, retry?: RetryInfo) {
```
BenoitZugmeyer, line 36: **❓ question:** ‏In which case the `flushReason` is undefined? Could it be made required?

packages/core/src/transport/batch.ts
```diff
@@ -11,6 +11,14 @@ export interface BatchFlushEvent {
   bufferMessagesCount: number
 }
 
+export type FlushReason =
+  | 'batch_duration_limit'
+  | 'batch_bytes_limit'
+  | 'before_unload'
+  | 'pagehide'
```
BenoitZugmeyer, line 18: **💬 suggestion:** ‏`page_hide`, to match other reasons pattern

packages/core/src/browser/pageExitObservable.ts
```diff
@@ -1,10 +1,12 @@
...
 import { Observable } from '../tools/observable'
-import { includes, objectValues } from '../tools/utils'
-import { addEventListener, addEventListeners, DOM_EVENT } from './addEventListener'
+import { includes, noop, objectValues } from '../tools/utils'
+import { addEventListeners, addEventListener, DOM_EVENT } from './addEventListener'
 
 export const PageExitReason = {
   HIDDEN: 'visibility_hidden',
   UNLOADING: 'before_unload',
+  PAGEHIDE: 'pagehide',
```
BenoitZugmeyer, line 9: **💬 suggestion:** ‏At first I wanted to suggest using `PAGE_HIDE: "page_hide"`, but then I'm a bit mitigated. The `PageExitReason` goal was to abstract a bit over event names (which are implementation detail) and to have two variants, one as a "possible exit" where we just want to flush things and the other as a "definitive exit" where we also want to end things.

Now we have four variants and I'm not sure what we want/need to do with each of them. Maybe it could be clarified with something like this:

```ts
interface PageExitEvent {
  definitive: boolean, // whether we should end things or not
  reason: PageExitReason,
}

```

BenoitZugmeyer (commented):
You should adjust `trackViews` [here](https://github.com/DataDog/browser-sdk/blob/a483e80b210ef0ef1fa918d7557108f1570b382a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts#L132) to end the view on `pagehide`

packages/core/src/transport/batch.ts
amortemousque: True, however for the scope of this investigation (assess pagehide impact) I think it's okay. Wdyt?

packages/core/src/browser/pageExitObservable.ts
amortemousque: To be honest, I have the same filling looking at this observable. I do agree that with that many events, all triggered in different page lifecycle state change, the `exit` naming is unclear. Maybe this observable as too many responsibilities.
I prefer to think of it as a convenient way to share logic for batch and segment flush.

Maybe we could:
- Use this observable only for the batch and segment flush and chose a better name
- No use it to end view and just use a normal window event listener on pagehide

Finally:
To be sure this PR make sense: I would prefer to move forward with this implementation to investigate the `pagehide` impact. Wdyt?

packages/core/src/domain/configuration/endpointBuilder.ts
amortemousque: In the [segment flush](https://github.com/DataDog/browser-sdk/blob/e394296baae484f64241c171b80d6220994261f1/packages/rum/src/domain/segmentCollection/segmentCollection.ts#L138), payload.flushReason is empty (because already set in the metadata).

packages/core/src/transport/batch.ts
BenoitZugmeyer: Alright, I'm fine with that if it's temporary

packages/core/src/browser/pageExitObservable.ts
BenoitZugmeyer: > Your suggestion could make sense but to me it adds another level of complexity where we want to make things easier.

On the contrary, I would argue that it makes things easier as we would have 2 values to consider instead of 4 values (boolean vs enum values).

In your PR, you did not adjust trackViews [here](https://github.com/DataDog/browser-sdk/blob/a483e80b210ef0ef1fa918d7557108f1570b382a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts#L132) to end the view on `pagehide`, so we will lose final view updates. Having a clear boolean instead of relying on the reason value could have been less error-prone.

>  To be sure this PR make sense: I would prefer to move forward with this implementation to investigate the pagehide impact. Wdyt?

I'm fine with keeping the large enum for now, since it is temporary.

amortemousque: > In your PR, you did not adjust trackViews [here](https://github.com/DataDog/browser-sdk/blob/a483e80b210ef0ef1fa918d7557108f1570b382a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts#L132) to end the view on pagehide
Good catch 🙂 

> On the contrary, I would argue that it makes things easier as we would have 2 values to consider instead of 4 values (boolean vs enum values).
It might be just me but since we would keep the pageExitReason, adding this boolean would make more combination to understand 2 * number of reasons. Of course not all these combinations would exist but PageExitEvent don't explicitly tells it.

bcaudan: > the exit naming is unclear

agree

> one as a "possible exit" where we just want to flush things and the other as a "definitive exit" where we also want to end things

I'm wondering if we could align our concepts with the ones from [page lifecycle api](https://developer.chrome.com/blog/page-lifecycle-api/) and then have something like:

```
PageLifecycleEvent {
  state: 'hidden' | 'frozen' | 'terminated',
  event: 'visibilitychange' | 'pagehide' | 'beforeunload' | 'freeze'
}
```

With that we could:
- listen for `hidden` | `frozen` | `terminated` to flush batches/segments.
- listen for `terminated` to end the views
- use the event names as reason for monitoring purposes

would it make sense?

amortemousque: Good idea! I don't want to mix it with this PR so I'll create a followup task

BenoitZugmeyer: I don't think it's particullary simpler than a boolean, and I still think the state and event are implementation details.

BenoitZugmeyer (approved):

packages/rum/src/types/sessionReplay.ts
```diff
@@ -34,6 +34,7 @@ export type CreationReason =
   | 'before_unload'
   | 'visibility_hidden'
   | 'page_frozen'
+  | 'page_hide'
```
bcaudan, line 37: **💭 thought:** Since we are not updating the schema, it could be nice to not update this file nether.
Someone could do a schema update on another topic and be bothered by that.

bcaudan (approved):

PR #2021, opened 2023-02-17
[REPLAY] Fix serialization for checkbox & radio

ThibautGeriz:
## Motivation

When serialise checkbox & radio with mask-input-privacy. It would record the input with `checked: '***'` on all the input regardless or if it was set.

It would be re-created like this in the player
```html
<label>
  <input
    type="radio"
    name="name1"
    value="***"
    checked="***"
  />
  <span>option 1</span>
</label>
<label class="c-radio">
  <input
    type="radio"
    name="name1"
    value="***"
    checked="***"
  />
  <span>option 2</span>
</label>
```

Meaning all the radio & checkbox would be checked.

Let's uncheck all of them when that's the privacy setting

## Changes

Delete the checked attribute instead of the overding it with `***`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum/src/domain/record/serialize.spec.ts
```diff
@@ -359,6 +359,66 @@ describe('serializeNodeWithId', () => {
...
+          attributes: {
+            type: 'radio',
+            value: 'on',
+            checked: true,
+          },
+        })
+      )
+    })
+
+    it('serializes <input type="radio"> elements checked state with mask serialization', () => {
```
BenoitZugmeyer, line 388: **💬 suggestion:**
* this test title is a bit misleading as it actually *does not* serialize the checked state.
* Maybe move it in the `describe('input privacy mode mask-user-input'` to colocate with similar tests

BenoitZugmeyer (approved):

PR #2029, opened 2023-02-21
RUMF-1497 Allow logger APIs to pass an Error parameter

yannickadam:
## Motivation

The goal of this PR is to align capabilities for error monitoring between RUM and LOGS SDKs.

## Changes

Added an optional error parameter: 
`logger.[level](message: string, messageContext?: object, error?: Error)`
which will be serialized as part of an error context, with defined fields (message, kind, stack)

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/domain/logger.ts
+++ b/packages/logs/src/domain/logger.ts
@@ -1,9 +1,10 @@
 import type { Context } from '@datadog/browser-core'
 import {
-  NO_ERROR_STACK_PRESENT_MESSAGE,
+  clocksNow,
+  computeRawError,
+  ErrorHandling,
   PROVIDED_ERROR_MESSAGE_PREFIX,
   computeStackTrace,
-  toStackTraceString,
   deepClone,
   assign,
   combine,
@@ -55,19 +56,25 @@ export class Logger {
     let context: Context
 
     if (status === StatusType.error || (error !== undefined && error !== null)) {
-      // Always add origin if status is error (backward compatibility), or we have an actual error
+      // Always add origin if status is error (backward compatibility - Remove in next major)
       const errorContext = ({ origin: ErrorSource.LOGGER } as LogsEvent['error'])!
 
-      // Extract information from error object if provided
-      if (error instanceof Error) {
-        const stackTrace = computeStackTrace(error)
-        errorContext.kind = stackTrace.name
-        errorContext.message = stackTrace.message
-        errorContext.stack = toStackTraceString(stackTrace)
-        // Serialize other types if provided as error parameter
-      } else if (error !== undefined && error !== null) {
-        errorContext.message = `${PROVIDED_ERROR_MESSAGE_PREFIX} ${JSON.stringify(error, undefined, 2)}`
-        errorContext.stack = NO_ERROR_STACK_PRESENT_MESSAGE
+      if (error !== undefined && error !== null) {
+        const stackTrace = error instanceof Error ? computeStackTrace(error) : undefined
+        const rawError = computeRawError({
+          stackTrace,
+          originalError: error,
+          nonErrorPrefix: PROVIDED_ERROR_MESSAGE_PREFIX,
+          source: ErrorSource.LOGGER,
+          handling: ErrorHandling.HANDLED,
+          startClocks: clocksNow(),
+        })
+
+        errorContext.stack = rawError.stack
+        errorContext.message = rawError.message
+        if (rawError.type) {
+          errorContext.kind = rawError.type
+        }
       }
 
       context = combine({ error: errorContext }, messageContext) as Context
--- a/packages/logs/src/domain/logger.ts
+++ b/packages/logs/src/domain/logger.ts
@@ -53,32 +53,35 @@ export class Logger {
 
   @monitored
   log(message: string, messageContext?: object, status: StatusType = StatusType.info, error?: Error) {
-    let context: Context
+    let errorContext: LogsEvent['error']
 
-    if (status === StatusType.error || (error !== undefined && error !== null)) {
+    if (status === StatusType.error) {
       // Always add origin if status is error (backward compatibility - Remove in next major)
-      const errorContext = ({ origin: ErrorSource.LOGGER } as LogsEvent['error'])!
-
-      if (error !== undefined && error !== null) {
-        const stackTrace = error instanceof Error ? computeStackTrace(error) : undefined
-        const rawError = computeRawError({
-          stackTrace,
-          originalError: error,
-          nonErrorPrefix: PROVIDED_ERROR_MESSAGE_PREFIX,
-          source: ErrorSource.LOGGER,
-          handling: ErrorHandling.HANDLED,
-          startClocks: clocksNow(),
-        })
-
-        errorContext.stack = rawError.stack
-        errorContext.message = rawError.message
-        errorContext.kind = rawError.type
-      }
-
-      context = combine({ error: errorContext }, messageContext) as Context
-    } else {
-      context = deepClone(messageContext) as Context
+      errorContext = { origin: ErrorSource.LOGGER }
     }
+
+    if (error !== undefined && error !== null) {
+      const stackTrace = error instanceof Error ? computeStackTrace(error) : undefined
+      const rawError = computeRawError({
+        stackTrace,
+        originalError: error,
+        nonErrorPrefix: PROVIDED_ERROR_MESSAGE_PREFIX,
+        source: ErrorSource.LOGGER,
+        handling: ErrorHandling.HANDLED,
+        startClocks: clocksNow(),
+      })
+
+      errorContext = {
+        origin: ErrorSource.LOGGER, // Remove in next major
+        stack: rawError.stack,
+        kind: rawError.type,
+        message: rawError.message,
+      }
+    }
+
+    const context = errorContext
+      ? (combine({ error: errorContext }, messageContext) as Context)
+      : (deepClone(messageContext) as Context)
 
     this.handleLogStrategy({ message, context, status }, this)
   }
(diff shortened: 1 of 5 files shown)


packages/core/src/domain/console/consoleObservable.ts
```diff
@@ -6,9 +6,17 @@ import { ConsoleApiName } from '../../tools/display'
 import { callMonitored } from '../../tools/monitor'
 
 export interface ConsoleLog {
+  /* Error message created from the concatenation of console parameters */
```
BenoitZugmeyer, line 9: **👏 praise:** ‏Nice initiative


**💬 suggestion:** ‏to be even more helpful, just use `/** ... */` instead of `/* ... */` so TS picks it up and displays it in your IDE of choice :)

packages/logs/src/domain/logger.ts
```diff
@@ -39,29 +51,47 @@ export class Logger {
...
+      if (error instanceof Error) {
+        const stackTrace = computeStackTrace(error)
+        errorContext.error.kind = stackTrace.name
+        errorContext.error.message = stackTrace.message
+        errorContext.error.stack = toStackTraceString(stackTrace)
+        // Serialize other types if provided as error parameter
+      } else if (error !== undefined && error !== null) {
+        errorContext.error.message = `${PROVIDED_ERROR_MESSAGE_PREFIX} ${JSON.stringify(error, undefined, 2)}`
+        errorContext.error.stack = NO_ERROR_STACK_PRESENT_MESSAGE
+      }
```
BenoitZugmeyer, line 71: **💬 suggestion:** ‏it could be nice to use `computeRawError` here so we have the same logic for serializing errors between RUM and Logs (ex: same empty message fallback, same non-`Error` instance serialization, `error.causes` support...)
* Right now, `computeRawError` requires a `StackTrace` as parameter, but I believe this could be changed and the `StackTrace` could be computed directly within `computeRawError` instead.
* It also requires unneeded things in this context, like `startClocks`. This might be adjusted, maybe with the new intermediary function that requires less parameters.
* The mapping between `RawError` to the logs error context could be done here to handle subtelties like `type -> kind`.

packages/logs/src/domain/logsCollection/console/consoleCollection.ts
```diff
@@ -31,7 +31,9 @@ export function startConsoleCollection(configuration: LogsConfiguration, lifeCyc
           log.api === ConsoleApiName.error
             ? {
                 origin: ErrorSource.CONSOLE, // Todo: Remove in the next major release
-                stack: log.stack,
+                kind: log.kind,
+                stack: log.stack || NO_ERROR_STACK_PRESENT_MESSAGE,
+                message: log.errorMessage || `${PROVIDED_ERROR_MESSAGE_PREFIX} ${log.message}`,
```
BenoitZugmeyer, line 36: **💬 suggestion:** ‏same thing here, we might be able to use the same serialization logic

packages/core/src/domain/console/consoleObservable.ts
```diff
@@ -6,9 +6,17 @@ import { ConsoleApiName } from '../../tools/display'
...
   message: string
+  /* The console API generating this log (debug, info, warn, error, ...) */
   api: ConsoleApiName
+  /* The specific type of the first Error object present in console parameters (SyntaxError, RangeError, ...) */
+  kind?: string
+  /* The stack trace provided by the first Error object present in console parameters */
   stack?: string
+  /* The message included in the first Error object present in console parameters */
+  errorMessage?: string
```
BenoitZugmeyer, line 18: **💭 thought:** ‏I think this might be a breaking change (using error.message instead of the full console message). Maybe we could remove this change from this PR?

packages/logs/src/domain/logger.ts
```diff
@@ -39,29 +51,47 @@ export class Logger {
...
 
   @monitored
-  log(message: string, messageContext?: object, status: StatusType = StatusType.info) {
-    this.handleLogStrategy({ message, context: deepClone(messageContext) as Context, status }, this)
+  log(message: string, messageContext?: object, status: StatusType = StatusType.info, error?: Error) {
+    let context: Context
+
+    if (status === StatusType.error || (error !== undefined && error !== null)) {
+      // Always add origin if status is error (backward compatibility), or we have an actual error
+      const errorContext = { error: { origin: ErrorSource.LOGGER } as Context }
```
bcaudan, line 59: **💬 suggestion:** ‏typings improvements + postponing the `error` field use
```suggestion
      const errorContext = ({ origin: ErrorSource.LOGGER } as LogsEvent['error'])!
      ...
      errorContext.kind = stackTrace.name
      ...
      context = combine({ error: errorContext }, messageContext)
```

packages/logs/src/domain/logger.ts
```diff
@@ -25,6 +34,9 @@ export const HandlerType = {
 export type HandlerType = typeof HandlerType[keyof typeof HandlerType]
 export const STATUSES = Object.keys(StatusType) as StatusType[]
 
+export const NO_ERROR_STACK_PRESENT_MESSAGE = 'No stack, consider using an instance of Error'
+export const PROVIDED_ERROR_MESSAGE_PREFIX = 'Provided:'
```
bcaudan, line 38: **💬 suggestion:** ‏What about mutualising those constants with other usages in the codebase?

packages/core/src/domain/console/consoleObservable.ts
bcaudan: FMU, it still uses the full message in log.message but now use the error message in log.error.message.
However, I'd be in favour of focusing the scope of this PR to only the logger APIs and address other improvements in a following PR to also discuss:

- usage of error.message on other error sources as well
- stacktrace computation of error instance for other console APIs

packages/core/src/domain/console/consoleObservable.ts
yannickadam: 👌

packages/logs/src/domain/logger.ts
yannickadam: 👌

packages/core/src/domain/console/consoleObservable.ts
yannickadam: Removed the commit for consoleObservable - to be handled in another PR.

packages/logs/src/domain/logger.ts
yannickadam: Agreed! Will be done in a separate PR as this needs a full review of the computeRawError test cases.

packages/logs/src/domain/logger.ts
yannickadam: 👌

packages/logs/src/domain/logger.ts
bcaudan: why not doing the exact same thing as in DD_RUM.addError (with logs specific): 
https://github.com/DataDog/browser-sdk/blob/1cc4fc73063d43f3abad1194bf069a992a8003d5/packages/rum-core/src/domain/rumEventsCollection/error/errorCollection.ts#L69-L78
and then build the `errorContext` from the needed `rawError` properties, something like:
```ts
const errorContext = {
  origin: ErrorSource.LOGGER,
  message: rawError.message,
  kind: rawError.type,
  stack: rawError.stack
}
```

packages/logs/src/domain/logger.ts
```diff
@@ -39,29 +52,53 @@ export class Logger {
...
+          source: ErrorSource.LOGGER,
+          handling: ErrorHandling.HANDLED,
+          startClocks: clocksNow(),
+        })
+
+        errorContext.stack = rawError.stack
+        errorContext.message = rawError.message
+        if (rawError.type) {
+          errorContext.kind = rawError.type
+        }
```
bcaudan, line 77: **🥜 nitpick:** ‏we could remove this `if` and let the serialization remove the undefined property

packages/logs/src/domain/logger.ts
```diff
@@ -39,29 +52,53 @@ export class Logger {
...
-  log(message: string, messageContext?: object, status: StatusType = StatusType.info) {
-    this.handleLogStrategy({ message, context: deepClone(messageContext) as Context, status }, this)
+  log(message: string, messageContext?: object, status: StatusType = StatusType.info, error?: Error) {
+    let context: Context
+
+    if (status === StatusType.error || (error !== undefined && error !== null)) {
+      // Always add origin if status is error (backward compatibility - Remove in next major)
+      const errorContext = ({ origin: ErrorSource.LOGGER } as LogsEvent['error'])!
+
+      if (error !== undefined && error !== null) {
```
bcaudan, line 62: **🥜 nitpick:** ‏we could remove of the `if` imbrication and the repetition of the "hasErrorParameter" condition with something like:

```ts
let errorContext: LogsEvent['error'] | undefined;

if (status === StatusType.error) {
  // Always add origin if status is error (backward compatibility - Remove in next major)
  errorContext = { origin: ErrorSource.LOGGER };
}

if (error !== undefined && error !== null) {
  ...
  errorContext = {
    ...
  }
}

const context = errorContext
  ? combine({ error: errorContext }, messageContext) as Context
  : deepClone(messageContext) as Context
```

bcaudan (approved):

amortemousque (approved):

BenoitZugmeyer (approved):

packages/logs/src/domain/logger.ts
yannickadam: 👌

packages/logs/src/domain/logger.ts
yannickadam: 👌

PR #2030, opened 2023-02-22
🐛 [RUMF-1443] fix Zone.js/Angular crash when recording the session

BenoitZugmeyer:
Issue reported to the Kendo-UI project here: https://github.com/telerik/kendo-angular/issues/3930

## Motivation

The issue is related to Zone.js/Angular + using the Kendo UI Grid component

1. open the official Kendo UI grid example: https://www.telerik.com/kendo-angular-ui/components/grid/examples/filter-all-columns/?theme=default-ocean-blue&themeVersion=6.1.0

2. execute the following code in the console (this is similar to what the SDK is doing):
```js
Object.defineProperty(HTMLInputElement.prototype, 'checked', { set() { setTimeout(() => { }) } })
```
3. click on the first checkbox

4. the CPU usage increases to 100%

## Diagnostic

* Angular is using Zone.js to track asynchronous executions (like `setTimeout`), and apparently in some cases triggers a component re-render when a `setTimeout` executes its callback.

* The grid component is likely using something like `input.checked = true` when being rendered.

* The Browser SDK instrumentation of `input.checked` uses `setTimeout` every time this property is set.

Thus, we have an infinite render loop consuming the CPU.

## Changes

Use the original `setTimeout` function when instrumenting property setters

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/instrumentMethod.spec.ts
+++ b/packages/core/src/tools/instrumentMethod.spec.ts
@@ -1,3 +1,4 @@
+import { stubZoneJs } from '../../test/stubZoneJs'
 import type { Clock } from '../../test/specHelper'
 import { mockClock } from '../../test/specHelper'
 import { instrumentMethod, instrumentSetter } from './instrumentMethod'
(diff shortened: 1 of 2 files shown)


yannickadam (approved):

amortemousque:
**💭 thought:**  Could it be interesting to do this check each time we use a setTimeout in the code base? Like we do with `addEventListeners` helper.

BenoitZugmeyer:
> **💭 thought:** Could it be interesting to do this check each time we use a setTimeout in the code base? Like we do with `addEventListeners` helper.

For the record: I'll open a new task for it

amortemousque (approved):

bcaudan (approved):

PR #2031, opened 2023-02-23
🚨 [RUMF-1505] ESLint rule to forbid usage of Zone.js problematic values

BenoitZugmeyer:
## Motivation

As we've seen in the past, integrating the SDK on an app that uses Angular/Zone.js causes critical issues in various ways. See related PRs:

* #376
* #866
* #1530
* #1860
* #2030

## Changes

This PR adds an ESLint rule to forbid usage of some values that are known to cause trouble when they are patched by Zone.js. This PR focuses on introducing the rule and disallow `MutationObserver`. Future PRs will disallow more things.

Screenshot 2023-02-23 at 16 00 53

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -217,6 +217,7 @@ module.exports = {
       excludedFiles: '*.spec.ts',
       rules: {
         'local-rules/disallow-side-effects': 'error',
+        'local-rules/disallow-zone-js-patched-values': 'error',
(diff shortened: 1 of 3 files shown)


bcaudan (approved):
Nice!

amortemousque (approved):

PR #2047, opened 2023-03-01
[REPLAY] Add public function to get the link to current Replay

ThibautGeriz:
## Motivation

Add public function to get the link to current Replay
## Changes

Add public function to get the link to current Replay
## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -138,6 +145,7 @@ export function makeRumPublicApi(
     }
 
     isAlreadyInitialized = true
+    return { configuration }
```
BenoitZugmeyer, line 148: **🔨 warning:** ‏init should keep returning undefined

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -252,6 +261,7 @@ export function makeRumPublicApi(
 
     startSessionReplayRecording: monitor(recorderApi.start),
     stopSessionReplayRecording: monitor(recorderApi.stop),
+    getReplayLink: monitor((subdomain = 'app') => getReplayLinkStrategy(subdomain)),
```
BenoitZugmeyer, line 264: **💬 suggestion:** ‏rename this method as `getSessionReplayLink` to match `startSessionReplayRecording` wording.


**💬 suggestion:** ‏don't set a default subdomain here (see other comment)

packages/rum/src/domain/getReplayLink.ts
```diff
@@ -0,0 +1,20 @@
+import type { RumConfiguration, RumSessionManager, ViewContexts } from '@datadog/browser-rum-core'
+
+export function getReplayLink(
+  configuration: RumConfiguration,
+  sessionManager: RumSessionManager,
+  viewContexts: ViewContexts,
+  subdomain: string
```
BenoitZugmeyer, line 7: **🔨 warning:** Not all site require a subdomain, so this argument should at least be optional


**💬 suggestion:** ‏I think letting the customer specify a subdomain complicates the API a bit. Could we avoid that and use sensible defaults? Ex:

| site |  subdomain |
|---|---|
| datadoghq.com | `app` |
| datad0g.com | `dd` |
| anything else | no subdomain |

BenoitZugmeyer: To clarify: I'm not sure if we should let the customer set the subdomain at all. I think we could start without it, and if there is some need in the future, adjust the API.

packages/rum/src/boot/isBrowserSupported.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Test for Browser features used while recording
+ */
+export function isBrowserSupported() {
+  return (
+    // Array.from is a bit less supported by browsers than CSSSupportsRule, but has higher chances
+    // to be polyfilled. Test for both to be more confident. We could add more things if we find out
+    // this test is not sufficient.
+    typeof Array.from === 'function' && typeof CSSSupportsRule === 'function' && 'forEach' in NodeList.prototype
+  )
+}
```
ThibautGeriz, line 11: **note:** extracted from `recorderApi.ts` to avoid a circular dep

packages/rum/src/domain/getSessionReplayLink.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+          startClocks: {
+            timeStamp: 123456,
+          },
+        }),
+      } as ViewContexts
+      addRecord('view-id-1')
+
+      const link = getSessionReplayLink({ ...DEFAULT_CONFIGURATION, site, subdomain }, sessionManager, viewContexts)
+
+      expect(link).toBe(`https://${host}/rum/replay/sessions/session-id-1?seed=view-id-1&from=123456`)
```
ThibautGeriz, line 51: **question:** this one fails on the browser stack with IE because `error-type=browser-not-supported` gets rightfully added. Should I disable the test on IO or should I re-use the helper `isBrowserSupported` in the unit tests here?

packages/rum/src/domain/getSessionReplayLink.ts
```diff
@@ -0,0 +1,50 @@
...
+  const session = sessionManager.findTrackedSession()
+  const queryParams: Record<string, string> = {}
+  let sessionId: string
+  if (session) {
+    sessionId = session.id
+
+    if (session.sessionReplayAllowed === false) {
+      // possibilities
+      // - replay sampled out
+      queryParams['error-type'] = 'incorrect-session-plan'
```
BenoitZugmeyer, line 20: **💬 suggestion:** ‏Having a separate function to get the error type would be clearer:

```ts
function getErrorType(session: RumSession | undefined, view: ViewContext | undefined) {
  if (!isBrowserSupported()) {
    return 'browser-not-supported'
  }
  if (!session) {
    // possibilities:
    // - rum sampled out
    // - session expired (edge case)
    return 'rum-not-tracked'
  }
  if (!session.sessionReplayAllowed) {
    // possibilities
    // - replay sampled out
    return 'incorrect-session-plan'
  }
  if (view && !getReplayStats(view.id)) {
    return 'replay-not-started'
  }
}
```

packages/rum/src/domain/getSessionReplayLink.ts
```diff
@@ -0,0 +1,50 @@
...
+import { isBrowserSupported } from '../boot/isBrowserSupported'
+import { getReplayStats } from './replayStats'
+
+export function getSessionReplayLink(
+  configuration: RumConfiguration,
+  sessionManager: RumSessionManager,
+  viewContexts: ViewContexts
+): string | undefined {
+  const session = sessionManager.findTrackedSession()
+  const queryParams: Record<string, string> = {}
```
BenoitZugmeyer, line 12: **💬 suggestion:** ‏Similarly to [`endpointBuilder`](https://github.com/DataDog/browser-sdk/blob/c0cd6b589368856db5c43d1aa91cd2a99618220c/packages/core/src/domain/configuration/endpointBuilder.ts#L114), using a string array would be simpler than a record here:

```ts
const parameters = []
...
parameters.push(`seed=${view.id}`)
...
parameters.join('&')
```

packages/rum/src/domain/getSessionReplayLink.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!replayStats && !queryParams['error-type']) {
+      queryParams['error-type'] = 'replay-not-started'
+    }
+  }
+  if (!isBrowserSupported()) {
+    queryParams['error-type'] = 'browser-not-supported'
+  }
+
+  let path = `/rum/replay/sessions/${sessionId}`
+  if (Object.keys(queryParams).length > 0) {
```
BenoitZugmeyer, line 43: **🥜 nitpick:** ‏I don't think we can have zero param in practice? I think even if we can, it's fine to have the question mark in this case, so the code could look like this:

```ts
return `${origin}${path}?${parameters.join('&')}`
```

packages/rum-slim/src/domain/getSessionReplayLink.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  parameters.forEach(([site, subdomain, host]) => {
+    it(`should return ${host} for subdomain "${
+      subdomain ?? 'undefined'
+    }" on "${site}" with query params if view is found`, () => {
+      const link = getSessionReplayLink({ ...DEFAULT_CONFIGURATION, site, subdomain })
+
+      expect(link).toBe(`https://${host}/rum/replay/sessions/session-id?error-type=slim-package`)
+    })
+  })
+})
```
BenoitZugmeyer, line 29: **💬 suggestion:** ‏You have tests for `getDatadogOrigin`. You might not need to test for every case in other spec files. Maybe just keep a single test like:

```suggestion
describe('getReplayLink (slim package)', () => {
  it('should return the replay link with a "slim-package" error type', () => {
      const link = getSessionReplayLink(DEFAULT_CONFIGURATION)
      expect(link).toBe(`https://dd.datad0g.com/rum/replay/sessions/session-id?error-type=slim-package`)
  })
})
```

packages/rum/src/domain/getSessionReplayLink.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    ['datadoghq.com', undefined, 'app.datadoghq.com'],
+    ['datadoghq.com', 'toto', 'toto.datadoghq.com'],
+    ['datad0g.com', undefined, 'dd.datad0g.com'],
+    ['datad0g.com', 'toto', 'toto.datad0g.com'],
+    ['us3.datadoghq.com', undefined, 'us3.datadoghq.com'],
+    ['us3.datadoghq.com', 'toto', 'toto.us3.datadoghq.com'],
+    ['us5.datadoghq.com', undefined, 'us5.datadoghq.com'],
+    ['us5.datadoghq.com', 'toto', 'toto.us5.datadoghq.com'],
+  ]
+
```
BenoitZugmeyer, line 33: **💬 suggestion:** ‏here too you might not need to test for every site/subdomain cases

packages/rum/src/domain/getSessionReplayLink.spec.ts
BenoitZugmeyer: Maybe change the expectation on IE?

```suggestion
      expect(link).toBe(isIE() ? ... : `https://${host}/rum/replay/sessions/session-id-1?seed=view-id-1&from=123456`)
```

BenoitZugmeyer (approved):

packages/rum/src/domain/getSessionReplayLink.ts
```diff
@@ -0,0 +1,55 @@
...
+    return 'rum-not-tracked'
+  }
+  if (!session.sessionReplayAllowed) {
+    // possibilities
+    // - replay sampled out
+    return 'incorrect-session-plan'
+  }
+  if (view && !getReplayStats(view.id)) {
+    return 'replay-not-started'
+  }
```
bcaudan, line 54: **💬 suggestion:** ‏what about using recorder state directly rather than accessing replay stats?

packages/rum/src/domain/getSessionReplayLink.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { RumConfiguration, ViewContexts } from '@datadog/browser-rum-core'
+import { createRumSessionManagerMock } from '../../../rum-core/test'
+import { getSessionReplayLink } from './getSessionReplayLink'
+import { addRecord, resetReplayStats } from './replayStats'
+
+const DEFAULT_CONFIGURATION = {
+  site: 'datad0g.com',
+} as RumConfiguration
+
+describe('getReplayLink', () => {
```
bcaudan, line 11: **💬 suggestion:** ‏Could you add a test for `replay-not-started`?

packages/rum/src/domain/getSessionReplayLink.spec.ts
ThibautGeriz: it seems I cannot read apparently

packages/rum-slim/src/domain/getSessionReplayLink.ts
```diff
@@ -0,0 +1,7 @@
+import type { RumConfiguration } from '@datadog/browser-rum-core'
+import { getDatadogOrigin } from '@datadog/browser-rum-core'
+
+export function getSessionReplayLink(configuration: RumConfiguration): string | undefined {
+  const origin = getDatadogOrigin(configuration)
+  return `${origin}/rum/replay/sessions/session-id?error-type=slim-package`
```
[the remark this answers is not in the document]
ThibautGeriz: something like that? Where should that go? `packages/rum-core/src/domain` ?
```ts
function getSessionReplayUrl(configuration: RumConfiguration, { errorType, seed, ts}: {errorType?: string; seed?: string; ts?: number }) {
  // impl
}
```

packages/rum/src/domain/getSessionReplayLink.ts
ThibautGeriz: I will leave that to @BenoitZugmeyer

BenoitZugmeyer: It's a good idea. You could do something like this: https://gist.github.com/BenoitZugmeyer/b0c6a09367b4a7970211cb66c7eea5d3 .

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

packages/rum-core/src/domain/getDatadogUrl.ts
bcaudan, line 1: **💬 suggestion:** ‏what about renaming this file to `getSessionReplayUrl`?
FMU, `getDatadogSiteUrl` is only exposed for tests

packages/rum-core/src/index.ts
```diff
@@ -23,10 +23,11 @@ export { startRum } from './boot/startRum'
...
 export { ViewContexts, ViewContext } from './domain/contexts/viewContexts'
-export { RumSessionManager, RumSessionPlan } from './domain/rumSessionManager'
+export { RumSessionManager, RumSessionPlan, RumSession } from './domain/rumSessionManager'
 export { getMutationObserverConstructor } from './browser/domMutationObservable'
 export { initViewportObservable, getViewportDimension } from './browser/viewportObservable'
 export { RumInitConfiguration, RumConfiguration } from './domain/configuration'
 export { DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE } from './domain/rumEventsCollection/action/getActionNameFromElement'
 export { STABLE_ATTRIBUTES } from './domain/rumEventsCollection/action/getSelectorFromElement'
 export * from './browser/htmlDomUtils'
+export { getDatadogSiteUrl, getSessionReplayUrl, SessionReplayUrlQueryParams } from './domain/getDatadogUrl'
```
bcaudan, line 33: **💬 suggestion:** it does not seem needed anymore 
‏
```suggestion
export { getSessionReplayUrl, SessionReplayUrlQueryParams } from './domain/getDatadogUrl'
```

packages/rum/src/domain/getSessionReplayLink.ts
```diff
@@ -0,0 +1,53 @@
...
+    queryParams.errorType = errorType
+  }
+
+  const view = viewContexts.findView()
+  if (view) {
+    queryParams.seed = view.id
+    queryParams.from = view.startClocks.timeStamp
+  }
+
+  return getSessionReplayUrl(configuration, sessionId, queryParams)
```
bcaudan, line 32: **💬 suggestion:** ‏what about passing directly `session`, `view` and `errorType` to `getSessionReplayUrl`?
it would avoid to declare query params outside of the link computation and would avoid to have `no-session-id` declared at several spots.

packages/rum-core/src/domain/getSessionReplayUrl.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      const link = getSessionReplayUrl({ site: 'datadoghq.com' } as RumConfiguration, {
+        viewContext,
+        session,
+        errorType,
+      })
+
+      expect(link).toBe(expected)
+    })
+  })
+})
```
bcaudan, line 74: **💬 suggestion:** ‏rework with:

- use object instead of tuple as spec parameter to increase readability and avoid adding cumbersome type 
- document intent of spec case in the spec parameters instead of building a message from the spec parameters
- simplify implementation by providing correct object in spec parameters instead of building them in the test


```suggestion
describe('getSessionReplayUrl', () => {
  const parameters = [
    [{
      testCase: 'session, no view, no error',
      session: { id: 'session-id-1' } as RumSession,
      viewContext: undefined,
      errorType: undefined,
      expected: 'https://app.datadoghq.com/rum/replay/sessions/session-id-1?'
    }],
    [{
      testCase: 'no session, no view, error',
      session: undefined,
      viewContext: undefined,
      errorType: 'toto',
      expected: 'https://app.datadoghq.com/rum/replay/sessions/no-session-id?error-type=toto'
    }],
    [{
      testCase: 'session, view, no error',
      session: { id: 'session-id-2' } as RumSession,
      viewContext: { id: 'view-id-1', startClocks: relativeToClocks(1234 as RelativeTime) },
      errorType: undefined,
      expected: `https://app.datadoghq.com/rum/replay/sessions/session-id-2?seed=view-id-1&from=${getTimeStamp(1234 as RelativeTime)}`,
    }],
    [{
      testCase: 'session, view, error',
      session: { id: 'session-id-3' } as RumSession,
      viewContext: { id: 'view-id-2', startClocks: relativeToClocks(1234 as RelativeTime) },
      errorType: 'titi',
      expected: `https://app.datadoghq.com/rum/replay/sessions/session-id-3?error-type=titi&seed=view-id-2&from=${getTimeStamp(1234 as RelativeTime)}`
    }],
  ]

  parameters.forEach(([{ testCase, session, viewContext, errorType, expected }]) => {
    it(`should build url when ${testCase}`, () => {
      const link = getSessionReplayUrl({ site: 'datadoghq.com' } as RumConfiguration, {
        viewContext,
        session,
        errorType,
      })
      expect(link).toBe(expected)
    })
  })
})
```

bcaudan (approved):

PR #2048, opened 2023-03-01
👷 extend Volta configuration in most package.json

BenoitZugmeyer:
## Motivation

With this change, the right version of Node and Yarn will be used when invoking commands from sub-packages directories. This is particularly useful with "native"/binary dependencies installed for a given Node version which will not run with a different version.

Ex: running `yarn dev` in developer-extension/, or `yarn start` in performances/.

## Changes

Set a volta configuration in most package.json, extending the root configuration (see [Volta doc](https://docs.volta.sh/advanced/workspaces))

I also removed nodenv support, since we should all have migrated to Volta by now.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


mbeaudru (approved):
Nice! I recall having trouble with this a few years ago, it's cool that it's now supported ☺️

bcaudan (approved):

PR #2050, opened 2023-03-01
[RUMF-1508] Provide stack trace for all uncaught exceptions

yannickadam:
## Motivation

As per [documentation](https://docs.datadoghq.com/real_user_monitoring/error_tracking/browser/), errors that do not have a stack trace are not grouped into issues. 
In the case of Uncaught Exceptions, even when we do not receive a parameter which is an `instance of Error`, we can generate a compliant stack trace from the source, line number and column number parameters.

## Changes

If the parameter errorObj is not an instance of `Error`, we stop trying to compute a stack trace. Instead. we use the fallback mechanism based on lineno/colno/source. However, we give priority to the errorObj's content to create the error message. 

Thrown string before:
```
message:"Uncaught "Just a thrown message, no error""
source:"source"
stack:"No stack, consider using an instance of Error"
handling:"unhandled"
source_type:"browser"
```
Thrown string after:
```
message:"Uncaught "Just a thrown message, no error""
source:"source"
stack:"Error: Uncaught "Just a thrown message, no error" at undefined @ http://localhost:8200/polyfills.js:8004:13"
handling:"unhandled"
source_type:"browser"
```

Thrown object before:
```
message:"Uncaught {"msg":"An object thrown","data":"foo"}"
source:"source"
stack:"No stack, consider using an instance of Error"
handling:"unhandled"
source_type:"browser"
```
Thrown object after:
```
message:"Uncaught {"msg":"An object thrown","data":"foo"}"
source:"source"
stack:"Error: Uncaught {"msg":"An object thrown","data":"foo"} at undefined @ http://localhost:8200/polyfills.js:8004:13"
handling:"unhandled"
source_type:"browser"
```

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


amortemousque:
**💭 thought:** ‏It's not directly related to this PR but I'm a bit uncomfortable to have error logic in tools. What about moving `tools/error` in `domain/error`?

packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -55,36 +57,44 @@ export function startUnhandledErrorCollection(callback: UnhandledErrorCallback)
...
+        // If an errorObj is provided, give priority to its content to populate the message of the stack trace
+        if (errorObj) {
+          msg = jsonStringify(errorObj)
         }
 
-        callback(stack, message)
+        const stack: StackTrace = {
+          name,
+          // We only include the `Uncaught` prefix when errorObj is present, to preserve backward compatibility
+          message: errorObj && msg ? `${NonErrorPrefix.UNCAUGHT} ${msg}` : msg,
```
bcaudan, line 87: **💭 thought:** It could be nice to limit `trackit` responsibilities to `Cross-browser collection of unhandled errors` only and not introducing our logic of building error messages here.

packages/core/src/tools/error.types.ts
```diff
@@ -0,0 +1,47 @@
...
+  LOGGER: 'logger',
+  NETWORK: 'network',
+  SOURCE: 'source',
+  REPORT: 'report',
+} as const
+
+export const NonErrorPrefix = {
+  UNCAUGHT: 'Uncaught',
+  PROVIDED: 'Provided',
+} as const
```
BenoitZugmeyer, line 39: **🥜 nitpick:** ‏ you could use an `enum` here

packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -55,36 +57,44 @@ export function startUnhandledErrorCollection(callback: UnhandledErrorCallback)
...
+          stack: [
+            {
+              url,
+              column: columnNo,
+              line: lineNo,
+            },
+          ],
```
BenoitZugmeyer, line 94: **💬 suggestion:** ‏when `url`, `columnNo` and `lineNo` is undefined, it won't provide anything meaningful. Maybe check wether some are defined before setting it? You might go even further and add the stack only if `url` is defined, because a line/column without url is not really helpful.

BenoitZugmeyer (approved):

packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -55,36 +57,44 @@ export function startUnhandledErrorCollection(callback: UnhandledErrorCallback)
...
             msg = groups[2]
           }
         }
 
-        stack = {
-          name,
-          message: typeof msg === 'string' ? msg : undefined,
-          stack: [location],
+        // If an errorObj is provided, give priority to its content to populate the message of the stack trace
+        if (errorObj) {
```
BenoitZugmeyer, line 80: **💬 suggestion:** use an `else if` to be explicit on whether the `if ({}.toString.call(message) === '[object String]') {` condition or the `if(errorObj)` condition takes precedence to define the `msg` variable

bcaudan:
Outdated, closed in favour of #2182

PR #2061, opened 2023-03-07
🐛 fix error when calling `fetch(null)`

BenoitZugmeyer:
## Motivation

When called with anything other than `Request` as a first argument, `fetch` will use the string representation of the first parameter as an URL. For example:

```js
fetch({ toString() { return 'toto' }}) // equivalent of fetch("toto")
fetch(null) // equivalent of fetch("null")
fetch() // equivalent of fetch("undefined")
```

Our fetch instrumentation does not support such usage and crashes with the error:
```
URL constructor: null is not a valid URL.
```

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/core/src/browser/fetchObservable.ts
```diff
@@ -52,7 +52,7 @@ function createFetchObservable() {
 
           const context = callMonitored(beforeSend, null, [observable, input, init])
           if (context) {
-            responsePromise = originalFetch.call(this, context.input, context.init)
+            responsePromise = originalFetch.call(this, context.input as Request | string, context.init)
```
bcaudan, line 55: **💬 suggestion:** ‏we could use the existing type
```suggestion
            responsePromise = originalFetch.call(this, context.input as RequestInfo, context.init)
```

bcaudan (approved):

BenoitZugmeyer:
* main
  * **PR #2061** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2061" target="_blank">Graphite</a>  👈
    * **PR #2094** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2094" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2061?utm_source=stack-comment).

PR #2062, opened 2023-03-07
🔊 [RUMF-1281] Collect more detail on failing URL build telemetry

amortemousque:
## Motivation

URL normalization used for xhr and fetch uses location.origin as a base. In certain cases the normalisation fails because the location.origin seem malformed.  

## Changes

Collect the url and base used for the normalization

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/core/src/tools/urlPolyfill.ts
```diff
@@ -35,7 +35,11 @@ export function getHash(url: string) {
 
 export function buildUrl(url: string, base?: string) {
   if (checkURLSupported()) {
-    return base !== undefined ? new URL(url, base) : new URL(url)
+    try {
+      return base !== undefined ? new URL(url, base) : new URL(url)
+    } catch (error) {
+      throw new Error(`Failed to construct URL. ${jsonStringify({ url, base })!}`)
```
amortemousque, line 41: What did I chose to throw a custom error over `addTelemetryError`:
  - addTelemetryError creates a circular reference
  - using addTelemetryError would also mean
    - to re throw the error and have two monitoring errors for the same thing
    - or to return undefined and handling this case which I don't want

So let's be simple

bcaudan (approved):

BenoitZugmeyer (approved):

PR #2053, opened 2023-03-03
🎨 [RUMF-1002] Split recorder observers in different files

amortemousque:
## Motivation

Split recorder observers in different files to help discoverability and navigation.

## Changes

Move observers in there own files:

- focusObserver.ts
- frustrationObserver.ts
- inputObserver.ts
- mediaInteractionObserver.ts
- mouseInteractionObserver.ts
- moveObserver.ts
- mutationObserver.ts
- scrollObserver.ts
- styleSheetObserver.ts
- viewportResizeObserver.ts

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2066, opened 2023-03-09
🎨 [RUMF-1002] Split recorder serialize in different files

amortemousque:
## Motivation

Following https://github.com/DataDog/browser-sdk/pull/2053, split recorder observers in different files to help discoverability and navigation.

## Changes

Move serialize functions in there own files:

- getAttributesForPrivacyLevel.ts
- serialization.types.ts
- serializationUtils.ts
- serializeAttribute.ts
- serializeDocument.ts
- serializeNode.ts
- serializeStyleSheets.ts

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


BenoitZugmeyer (approved):

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -1,43 +1,36 @@
...
   AST_MASK,
   AST_MASK_USER_INPUT,
-  generateLeanSerializedDoc,
-} from '../../../test/htmlAst'
-import type { ElementNode, SerializedNodeWithId, TextNode } from '../../types'
-import { NodeType } from '../../types'
-import type { IsolatedDom } from '../../../../rum-core/test/createIsolatedDom'
-import { createIsolatedDom } from '../../../../rum-core/test/createIsolatedDom'
+  AST_ALLOW,
+} from '../../../../test/htmlAst'
```
bcaudan, line 19: **💬 suggestion:** ‏it seems that htmlAst is only used in serialization tests, if it is the case what about moving this utility to the serialization package?

packages/rum/src/domain/record/serialization/getAttributesForPrivacyLevel.ts
```diff
@@ -0,0 +1,150 @@
+import { NodePrivacyLevel } from '../../../constants'
+import { shouldMaskNode } from '../privacy'
+import { getElementInputValue, switchToAbsoluteUrl, getValidTagName } from './serializationUtils'
+import type { SerializeOptions } from './serialization.types'
+import { SerializationContextStatus } from './serialization.types'
+import { serializeAttribute } from './serializeAttribute'
+
+export function getAttributesForPrivacyLevel(
```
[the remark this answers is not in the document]
amortemousque: Yes definitely!

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
amortemousque: For test utilities, we are setting them in `/test` folders. Should I create `serialization/utils/test/htmlAst.ts` ?

bcaudan: Mostly for shared one and at the same level than `src` though 🤔
For non shared utilities, most of the time they are in the same file as the spec that use them.
I don't know if we have some cases with test utilities that are in a dedicated file but used only by a single file, it could worth to have a look. 
Instead of adding an extra directory, we could have a file suffix to easily identify those files, something like `htmlAst.specHelper.ts` since we already have a couple of `specHelpers.ts`.
wdyt?

amortemousque: Sounds good 👍

amortemousque: I'll update the code convention

bcaudan (approved):

PR #2074, opened 2023-03-13
🚚 [devext] move devtools status to panel

BenoitZugmeyer:
## Motivation

Having the devtools status in the store was always a bit weird, since it's more a temporary state than something we want to store on the long term.

Also, its refresh strategy was not very good since it was only refreshed when the store changed, so it did not refresh when running `yarn dev` without manipulating the extension settings.

## Changes

Let's put it in the devtools panel directly, as a hook that periodically polls the dev server.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan (approved):
Nice!

PR #2111, opened 2023-03-27
♻️[RUMF-1517] split core specHelper

bcaudan:
## Motivation

Avoid single utils files

## Changes

- move `clearAllCookies` to `forEach.spec`
- move `capturedExceptions` to `tracekit`
- extract from `core/test/specHelper`:
  - `browserChecks`
  - `buildEnv`
  - `createNewEvent`
  - `mockClock`
  - `requests.specHelper`
  - `location.specHelper`
- rename `specHelper` to `simpleTestDoubles` 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/core/test/location.specHelper.ts
amortemousque, line 1: **❓ question:** ‏ What is the reasoning of having the `.specHelper` suffix or not?

bcaudan: My rule of thumb for now:

- if it is among source code, add a suffix to distinguish it from prod code and to avoid to compile it during package build
- if it contains several domain helpers, `[domain].ts` file name seemed a bit too broad so adding the suffix added a bit of clarity on its responsibility.

Totally challengeable, wdyt?

packages/core/test/simpleTestDoubles.ts
amortemousque, line 1: **❓ question:** ‏I don't understand the file name. What does it mean?

bcaudan: yep, I was not super inspired on this one, [test doubles](https://martinfowler.com/bliki/TestDouble.html) for reference.
The idea was to avoid to have dedicated files for all those simple dependency instrumentations and regroup them together.
`simpleDependencyInstrumentations` could be clearer.
wdyt?

packages/core/test/location.specHelper.ts
amortemousque: Okay, I see! 
It's totally opinionated but I'd prefer: 
- If in the test folder: no suffix
- If among source code: suffix

@BenoitZugmeyer @yannickadam wdyt?

BenoitZugmeyer: I agree with @amortemousque , having `.specHelpers.ts` suffix in `test` seems unnecessary.

packages/core/test/simpleTestDoubles.ts
BenoitZugmeyer: I don't think we can have a good name to differentiate those mocks from other things, and whether a mock is "simple" enough to put there or not. Maybe we could split the file into small files. Maybe we could group them in an "emulate" folder (which seems more appropriate than "double" in our case). Something like this:

```
test/
 |-- browserChecks
 |-- buildEnv
 |-- disableJasmineUncaughtErrorHandler
 |-- createNewEvent
 |-- emulate/
    |-- requests
    |-- clock
    |-- location
    |-- eventBridge
    |-- cookie
    |-- navigatorOnLine

bcaudan: if `emulate` is clearer, what about keeping a single `emulations.ts` file?

packages/core/test/location.specHelper.ts
bcaudan: ok, sounds good to me

packages/core/test/simpleTestDoubles.ts
bcaudan: ok, no hard arguments against it, let's go with it

BenoitZugmeyer (approved):

amortemousque (approved):

PR #2113, opened 2023-03-28
♻️ [RUMF-1529] use an enum for experimental features

BenoitZugmeyer:
## Motivation

Improve typecheck, helps to clean flags when enabling them.

## Changes

Use an enum for experimental features

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -112,7 +112,13 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
   }
 
   // Set the experimental feature flags as early as possible, so we can use them in most places
-  updateExperimentalFeatures(initConfiguration.enableExperimentalFeatures)
+  if (Array.isArray(initConfiguration.enableExperimentalFeatures)) {
+    updateExperimentalFeatures(
+      initConfiguration.enableExperimentalFeatures.filter((flag): flag is ExperimentalFeature =>
+        objectHasValue(ExperimentalFeature, flag)
+      )
+    )
```
bcaudan, line 120: **❓ question:** ‏could it be interesting to warn the customer if we filtered out a provided value?

bcaudan (approved):

packages/core/src/domain/configuration/configuration.ts
BenoitZugmeyer: I'm not sure. Like we discussed, in the future `enableExperimentalFeatures` should be less customer facing, so it'll be mostly us who use it. And if we were to show a warning every time a flag does not exist anymore, it will spam our fellow Datadog frontend devs until we remove the flag from consul, which will never happen exactly at the same time.

I would keep things quiet for now.

packages/core/src/domain/configuration/configuration.spec.ts
```diff
@@ -2,18 +2,41 @@ import type { RumEvent } from '../../../../rum-core/src'
...
 import { validateAndBuildConfiguration } from './configuration'
-import { isExperimentalFeatureEnabled, updateExperimentalFeatures } from './experimentalFeatures'
+import { ExperimentalFeature, isExperimentalFeatureEnabled, addExperimentalFeatures } from './experimentalFeatures'
 
 describe('validateAndBuildConfiguration', () => {
   const clientToken = 'some_client_token'
 
   beforeEach(() => {
-    updateExperimentalFeatures([])
+    addExperimentalFeatures([])
```
bcaudan, line 11: **❓ question:** ‏with the implementation changes, do we still need to do that?

yannickadam (approved):

PR #2120, opened 2023-03-29
✨ [RUMF-1510] Warn the user when a heavy context is used

amortemousque:
## Motivation

Raise awareness about customer data impact on the user bandwidth.

## Changes
- For ssr compat: 
  - Move `getGlobalObject` on its own file to avoid circular ref with `timer.ts`
  - Use `getGlobalObject` in `setTimeout` and `setInterval`
- Compute customer data bytes count at API call
- Add a console warning when 3KiB customer data reach for
    - Global context 
    - User
    - Feature flag evaluation
    - Logger context (I made the call since it also uses context manager)
- Warn once per customer data type to avoid spams

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/tools/heavyCustomerDataWarning.ts
```diff
@@ -13,12 +13,19 @@ export const enum CustomerDataType {
   LoggerContext = 'logger context',
 }
 
+let alreadyWarned: { [key: string]: boolean } = {}
```
bcaudan, line 16: **❓ question:** ‏FMU, for a npm setup with both logs and rum, it would mean that we would warn customers once for all SDK instances, is it intentional?

amortemousque: No it was not :) 
I moved the "already warn check" to `contextManager` and `featureFlagContext` to solve it. 
Let me know what you think.

PR #2125, opened 2023-03-30
✨ [RUMF-1530] send replay metadata as json

BenoitZugmeyer:
## Motivation

Send Replay metadata as JSON instead of individual multipart entries to make it easier to extend.

## Changes

Reading metadata from Replay payloads will need to be asynchronous, because we are using a `Blob` to set the entry content type to `application/json`. Thus, a large part of the PR focuses on preparing the tests for this change.

Then, we actually send the metadata as JSON

Finally, we add the `compressed_segment_size` to avoid needing custom backend code.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/rum/test/readReplayPayload.ts
```diff
@@ -0,0 +1,40 @@
...
+
+function readSegmentFromReplayPayload(payload: Payload) {
+  return readJsonBlob((payload.data as FormData).get('segment') as Blob, {
+    decompress: true,
+  }) as Promise<BrowserSegment>
+}
+
+export async function readMetadataFromReplayPayload(payload: Payload) {
+  return readJsonBlob((payload.data as FormData).get('event') as Blob) as Promise<BrowserSegmentMetadataAndSegmentSizes>
+}
```
bcaudan, line 22: **💭 thought:** ‏since we don't use await in this function, we could remove the async keyword.
We could enforce that with [require-await](https://eslint.org/docs/latest/rules/require-await) eslint rule.

bcaudan (approved):

packages/rum/src/domain/segmentCollection/buildReplayPayload.spec.ts
```diff
@@ -1,33 +1,57 @@
...
+    expect(metadata).toEqual(METADATA_AND_SEGMENT_SIZES)
   })
 
-  it('converts values to string', () => {
-    toFormEntries({ foo: 42, bar: null }, callbackSpy)
-    expect(callbackSpy.calls.allArgs()).toEqual([
-      ['foo', '42'],
-      ['bar', 'null'],
-    ])
+  it('returns the approximative byte counts of the request', () => {
```
amortemousque, line 53:
```suggestion
  it('returns the approximate byte counts of the request', () => {
```

BenoitZugmeyer:
* main
  * **PR #2125** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2125" target="_blank">Graphite</a>  👈
    * **PR #2132** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2132" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2125?utm_source=stack-comment).

packages/rum/test/readReplayPayload.ts
```diff
@@ -0,0 +1,60 @@
...
+      creation_reason: formData.get('creation_reason') as CreationReason,
+      raw_segment_size: Number(formData.get('raw_segment_size')),
+      index_in_view: Number(formData.get('index_in_view')),
+      has_full_snapshot: formData.get('has_full_snapshot') === 'true',
+    } satisfies Omit<BrowserSegmentMetadataAndSegmentSizes, 'compressed_segment_size'> & {
+      compressed_segment_size?: number
+    }
+  }
+
+  return readJsonBlob((payload.data as FormData).get('event') as Blob) as Promise<BrowserSegmentMetadataAndSegmentSizes>
```
ThibautGeriz, line 41: **question:** why not cast in the `resolve` function in `readJsonBlob`?

ThibautGeriz (approved):

packages/rum/test/readReplayPayload.ts
BenoitZugmeyer: You mean, something like having something like this?

```ts
function readJsonBlob<T>() {
  return new Promise<T>(...)
}
```

The end result is indeed the same, but the cast is implicit. I rather have a `readJsonBlob(): Promise<unknown>`, which provide no guarantee about the returned type, and then let the caller take the responsibility to ducktype it or cast it. But I agree the syntax is a bit more verbose.

packages/rum/src/domain/segmentCollection/buildReplayPayload.spec.ts
```diff
@@ -1,8 +1,89 @@
...
+      const eventEntry = (payload.data as FormData).get('event')! as File
+      expect(eventEntry.size).toBe(JSON.stringify(METADATA_AND_SEGMENT_SIZES).length)
+      expect(eventEntry.name).toBe('blob')
+      expect(eventEntry.type).toBe('application/json')
+      const { metadata } = await readReplayPayload(payload)
+      expect(metadata).toEqual(METADATA_AND_SEGMENT_SIZES)
+    })
+  })
+
+  describe('without replay_json_payload experimental flag', () => {
```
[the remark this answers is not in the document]
BenoitZugmeyer: All tests rely on a default state. This default state *needs* to be reset at the end of each test, else everything would be impredictable, especially because we are using a randomized test execution order.

packages/rum/test/readReplayPayload.ts
ThibautGeriz: this would not work?
```ts
resolve(deserialized as BrowserSegmentMetadataAndSegmentSizes)
```

BenoitZugmeyer: `readJsonBlob()` is used two times with different return types

PR #2135, opened 2023-04-03
👷 Update @wdio to major version 8.8.x

yannickadam:
## Motivation

Update to Webdriver 8 !

## Changes

In `test/e2e/tsconfig.json`:
- Updated `target` to `ES2015` => @wdio  uses private class members (`#` syntax), that cannot be transpiled to ES5
- With the previous change, `moduleResolution` had to be set to `node` for `resolveJsonModule` to work.
- The order in `types` is **important**! `jasmine` must come before `@wdio/globals/types` for the type check to take into account the proper `expect()`

Converted the config files to TypeScript

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/test/e2e/wdio.base.conf.ts
+++ b/test/e2e/wdio.base.conf.ts
@@ -78,7 +78,6 @@ export const config: OptionsWithLogsPath = {
   beforeSession: (_config, _capabilities, _specs, _cid) => {
     // Expose everything besides expect, as we want to keep the one from Jasmine
     global.browser = browser
-    global.driver = browser
     global.$ = $
     global.$$ = $$
   },
--- a/test/e2e/scenario/recorder/viewports.scenario.ts
+++ b/test/e2e/scenario/recorder/viewports.scenario.ts
@@ -182,7 +182,7 @@ async function pinchZoom(xChange: number) {
       ],
     },
   ]
-  await driver.performActions(actions)
+  await browser.performActions(actions)
 }
 
 async function performSignificantZoom() {
@@ -219,7 +219,7 @@ async function visualScrollVerticallyDown(yChange: number) {
       ],
     },
   ]
-  await driver.performActions(actions)
+  await browser.performActions(actions)
 }
 
 async function buildScrollablePage() {
--- a/test/e2e/wdio.bs.conf.ts
+++ b/test/e2e/wdio.bs.conf.ts
@@ -10,28 +10,28 @@ export const config: Options.Testrunner = {
 
   capabilities: browserConfigurations
     .filter(
-      (browser) =>
-        browser.sessionName !== 'IE' &&
+      (configuration) =>
+        configuration.sessionName !== 'IE' &&
         // Safari mobile on iOS 12.0 does not support
         // the way we flush events on page change
         // TODO check newer version on browserstack
-        browser.sessionName !== 'Safari mobile'
+        configuration.sessionName !== 'Safari mobile'
     )
-    .map((browser) =>
+    .map((configuration) =>
       // See https://www.browserstack.com/automate/capabilities?tag=selenium-4
       // Make sure to look at the "W3C Protocol" tab
       ({
-        browserName: browser.name,
-        browserVersion: browser.version,
+        browserName: configuration.name,
+        browserVersion: configuration.version,
         'bstack:options': {
-          os: browser.os,
-          osVersion: browser.osVersion,
-          deviceName: browser.device,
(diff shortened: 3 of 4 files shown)


test/e2e/wdio.base.conf.js
```diff
@@ -1,73 +0,0 @@
...
-        unlinkSync(logsPath)
-      } catch (e) {
-        if (e.code !== 'ENOENT') {
-          console.log(`Failed to remove previous logs: ${e.message}`)
-        }
-      }
-    }
-  },
-
-  logsPath,
```
BenoitZugmeyer, line 72: **💬 suggestion:** ‏FMU, webdriverio don't support the `logsPath` config anymore (or maybe it never did), so you removed from the configuration.

But this was used in `logger.ts` : https://github.com/DataDog/browser-sdk/pull/2135/files#diff-664ed702d3aa8efafdb821a59408d16dd1dffcea3f155a9b2ee750d12c18bdfcL4
Now that it is always undefined, logs will always be sent to stdout, which is a bit verbose.

Maybe we can restore the original behavior by having an "initLogger" function in `lib/framework/logger` that initializes the `stream`, which would be called in `onPrepare` after creating the logs file:

```ts
function initLogger(logsPath?: string) {
  stream = logsPath ? fs.createWriteStream(logsPath, { flags: 'a' }) : process.stdout
}
```

test/e2e/scenario/rum/init.scenario.ts
```diff
@@ -108,7 +108,7 @@ describe('API calls and events around init', () => {
 describe('beforeSend', () => {
   createTest('allows to edit non-view events context')
     .withRum({
-      beforeSend(event) {
+      beforeSend: (event: any) => {
```
BenoitZugmeyer, line 111: **💬 suggestion:** ‏I believe you've got some "implicit `any`" error because you didn't build packages with `yarn build` before running typechecks.

This was already the case before your PR, this is why [we build before typechecking e2e tests](https://github.com/DataDog/browser-sdk/blob/3ab528b0711a721d17b2c50e757fed5ece92f4c0/.gitlab-ci.yml#L134-L137). I think you can revert your modification in `init.scenario.ts` `logs.scenario.ts`.

Another solution could be to always import from `.../src/...` instead of relying on built packages, but that's a bigger change and a bit out of scope.

yannickadam: Indeed - The implicit `any` was due to a TS server de-sync. 
However - I need to keep the new form: `beforeSend: (event) => {}` For some reason, after the updates to `tsconfig.json` (I suspect target="ES2015"), the test failed with the former code.

test/e2e/scenario/recorder/recorder.scenario.ts
```diff
@@ -7,7 +7,8 @@ import type {
 } from '@datadog/browser-rum/src/types'
 import { NodeType, IncrementalSource, MouseInteractionType } from '@datadog/browser-rum/src/types'
 
-import { FrustrationType } from '@datadog/browser-rum-core'
+// Import from src to have properties of const enums
+import { FrustrationType } from '@datadog/browser-rum-core/src/rawRumEvent.types'
```
bcaudan, line 11: **❓ question:** ‏why are we missing enums properties from the package import?

test/e2e/wdio.base.conf.ts
```diff
@@ -0,0 +1,84 @@
...
+        if ((e as NodeJS.ErrnoException).code !== 'ENOENT') {
+          console.log(`Failed to remove previous logs: ${(e as Error).message}`)
+        }
+      }
+    }
+  },
+  beforeSession: (_config, _capabilities, _specs, _cid) => {
+    // Export everything besides expect, as we want to keep the one from Jasmine
+    global.browser = browser
+    global.driver = browser
```
bcaudan, line 81: **💬 suggestion:** ‏should we update our `driver` usages to use `browser` instead?

BenoitZugmeyer (approved):

test/e2e/scenario/recorder/recorder.scenario.ts
bcaudan: Since `FrustrationType` is exported from `@datadog/browser-rum-core`, what about using:

```ts
export const FrustrationType = {
  RAGE_CLICK: 'rage_click',
  ...
}
```

bcaudan (approved):

PR #2137, opened 2023-04-03
👷[RUMF-1517] Restrict import of other packages into tools

bcaudan:
## Motivation

Impose code structure

## Changes

Configure `no-restricted-import` rule on `tools` files

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -267,5 +267,21 @@ module.exports = {
         'import/no-extraneous-dependencies': 'off',
       },
     },
+    {
+      files: ['packages/core/src/tools/**/*.ts'],
+      rules: {
+        'no-restricted-imports': [
+          'error',
+          {
+            patterns: [
+              {
+                group: ['../**/boot/*', '../**/browser/*', '../**/domain/*', '../**/transport/*'],
+                message: 'tools components should not import from other directories',
+              },
+            ],
+          },
+        ],
+      },
+    },
   ],
 }


.eslintrc.js
```diff
@@ -267,5 +267,21 @@ module.exports = {
...
     },
+    {
+      files: ['packages/core/src/tools/**/*.ts'],
+      rules: {
+        'no-restricted-imports': [
+          'error',
+          {
+            patterns: [
+              {
+                group: ['../**/boot/*', '../**/browser/*', '../**/domain/*', '../**/transport/*'],
```
BenoitZugmeyer, line 278: **💬 suggestion:** [‏In the docs](https://eslint.org/docs/latest/rules/no-restricted-imports#options), it looks like you can allow some glob by starting with `!`. Maybe we can do something like `'../**', '!../tools/**'`

bcaudan: Yes, I struggled for a while to try to make it work with some variations of: `'../**', '!../**/test'` but without success 😕
However, this strategy may not work if we have more directory levels inside `tools` so I settled with the current proposition.

Let me know if you have other ideas.

BenoitZugmeyer: No other idea, it sounds fine to me

BenoitZugmeyer (approved):

PR #2141, opened 2023-04-05
🐛[RUMF-1544] Fix badly polyfilled URL 

amortemousque:
## Motivation

It has been observed that on some customer website, URL is badly polyfilled. It might not be a constructor, and might not provide a URL.createObjectURL function.

## Changes

- Replace `checkURLSupported` by `getSupportedURL` to keep the original URL ref to avoid the `URL is not a constructor` error in the `buildUrl`.
- Check if `URL.createObjectURL` is supported for the recorder to avoid `URL.createObjectURL is not a function` error in the worker 
- Add an ESLint rule to force using `buildURL` instead of `new URL`. Inspired by [no-new-symbol rule](https://github.com/eslint/eslint/blob/main/lib/rules/no-new-symbol.js) 

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/utils/urlPolyfill.spec.ts
+++ b/packages/core/src/tools/utils/urlPolyfill.spec.ts
@@ -30,6 +30,12 @@ describe('isValidUrl', () => {
     expect(isValidUrl('/plop')).toBe(false)
     expect(isValidUrl('')).toBe(false)
   })
+
+  it('should return the same result if the URL as been wrongfully overridden between calls', () => {
+    expect(isValidUrl('http://www.datadoghq.com')).toBe(true)
+    spyOn(window, 'URL').and.throwError('wrong URL override')
+    expect(isValidUrl('http://www.datadoghq.com')).toBe(true)
+  })
 })
 
 describe('getOrigin', () => {
--- a/packages/core/src/tools/utils/urlPolyfill.spec.ts
+++ b/packages/core/src/tools/utils/urlPolyfill.spec.ts
@@ -31,7 +31,7 @@ describe('isValidUrl', () => {
     expect(isValidUrl('')).toBe(false)
   })
 
-  it('should return the same result if the URL as been wrongfully overridden between calls', () => {
+  it('should return the same result if the URL has been wrongfully overridden between calls', () => {
     expect(isValidUrl('http://www.datadoghq.com')).toBe(true)
     spyOn(window, 'URL').and.throwError('wrong URL override')
     expect(isValidUrl('http://www.datadoghq.com')).toBe(true)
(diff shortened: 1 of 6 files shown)


packages/core/src/tools/utils/urlPolyfill.spec.ts
```diff
@@ -30,6 +30,12 @@ describe('isValidUrl', () => {
     expect(isValidUrl('/plop')).toBe(false)
     expect(isValidUrl('')).toBe(false)
   })
+
+  it('should return the same result if the URL as been wrongfully overridden between calls', () => {
```
yannickadam, line 34: :typo: `has been`

yannickadam (approved):

bcaudan (approved):

packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -34,11 +34,14 @@ export function getHash(url: string) {
...
+  const supportedURL = getSupportedUrl()
+  if (supportedURL) {
     try {
-      return base !== undefined ? new URL(url, base) : new URL(url)
+      return base !== undefined ? new supportedURL(url, base) : new supportedURL(url)
     } catch (error) {
-      throw new Error(`Failed to construct URL. ${jsonStringify({ url, base })!}`)
+      throw new Error(
+        `Failed to construct URL. ${jsonStringify({ url, base, error: error instanceof Error ? error.message : '' })!}`
+      )
```
BenoitZugmeyer, line 44: **💬 suggestion:** you could use `String(error)`, is simpler and handles any kind of error:

```suggestion
      throw new Error(
        `Failed to construct URL: ${String(error)} ${jsonStringify({ url, base })!}`
      )
```

packages/rum/src/boot/recorderApi.ts
```diff
@@ -164,6 +164,9 @@ function isBrowserSupported() {
     // Array.from is a bit less supported by browsers than CSSSupportsRule, but has higher chances
     // to be polyfilled. Test for both to be more confident. We could add more things if we find out
     // this test is not sufficient.
-    typeof Array.from === 'function' && typeof CSSSupportsRule === 'function' && 'forEach' in NodeList.prototype
+    typeof Array.from === 'function' &&
+    typeof CSSSupportsRule === 'function' &&
+    typeof URL.createObjectURL === 'function' &&
```
BenoitZugmeyer, line 169: **💬 suggestion:** ‏we could do`typeof getSupportedUrl()?.createObjectURL === 'function'` and use `getSupportedUrl()!.createObjectURL` when building the Worker, no?

BenoitZugmeyer (approved):

PR #2144, opened 2023-04-06
♻️ [RUMF-1533] extract the Flush logic into a reusable component

BenoitZugmeyer:
## Motivation

The flush logic will become a bit more complex in the short term. Let's extract the flush logic from `Batch` to separate concerns. In a future PR, the newly introduced flush controller will also be used for flushing segments.

## Changes

* Implement a flush controller to handle all flush use cases
* Use it in Batch
* Use it for customer data telemetry

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/transport/batch.spec.ts
```diff
@@ -1,150 +1,95 @@
...
-    const clock = sinon.useFakeTimers()
-    batch = new Batch(transport, BATCH_MESSAGES_LIMIT, BATCH_BYTES_LIMIT, MESSAGE_BYTES_LIMIT, 10, pageExitObservable)
-    batch.add({ message: '50 bytes - xxxxxxxxxxxxxxxxxxxxxxxxx' })
-    clock.tick(100)
+  it('should consider separators when replacing messages', () => {
+    batch.add(SMALL_MESSAGE)
+    batch.upsert(SMALL_MESSAGE, 'a')
 
-    expect(sendSpy).toHaveBeenCalled()
+    flushController.willAddMessage.calls.reset()
```
amortemousque, line 69: **❓ question:** ‏Why do you need this?

amortemousque (approved):

packages/core/src/transport/batch.ts
```diff
@@ -49,44 +27,35 @@ export class Batch {
...
-  flush(flushReason: FlushReason, sendFn = this.request.send) {
-    if (this.bufferMessagesCount !== 0) {
-      const messages = this.pushOnlyBuffer.concat(objectValues(this.upsertBuffer))
-      const bytesCount = this.bufferBytesCount
-
-      this.flushObservable.notify({
-        bufferBytesCount: this.bufferBytesCount,
-        bufferMessagesCount: this.bufferMessagesCount,
-      })
+  flush(event: FlushEvent) {
```
bcaudan, line 30: **💬 suggestion:** ‏it could be private now
```suggestion
  private flush(event: FlushEvent) {
```

packages/core/src/transport/flushController.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+  describe('messages limit', () => {
+    it('notifies when the messages limit is reached', () => {
+      for (let i = 0; i < MESSAGES_LIMIT; i += 1) {
+        flushController.willAddMessage(SMALL_MESSAGE_BYTE_COUNT)
+        flushController.didAddMessage()
+      }
+      expect(flushSpy).toHaveBeenCalled()
+    })
+
+    it('flush reason should be "bytes_limit"', () => {
```
bcaudan, line 138: **❓ question:** ‏should not we have a `messages_limit` flush reason?

packages/core/src/transport/batch.spec.ts
BenoitZugmeyer: I need to call `reset` to reset the number of calls after the test setup, so I can use `expect().toHaveBeenCalledOnceWith(...)` (only the last call is relevent to this test case)

bcaudan (approved):

PR #2064, opened 2023-03-08
✨ [RUMF-1518] implement a new API to stop the RUM session

BenoitZugmeyer:
## Motivation

Let SDK users have more control on when RUM sessions should end.

## Changes

* light refactoring
* implement a way to expire the session in the session manager
* expose `stopSession()`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2150, opened 2023-04-11
✨ [RUMF-1533] flush pending data on session expiration

BenoitZugmeyer:
## Motivation

With `stopSession` #2064 now available, it could be useful to use it to flush pending data in some specific integrations (ex: CI integration tests).

## Changes

Flush pending data on session expiration.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/transport/flushController.spec.ts
```diff
@@ -78,6 +81,33 @@ describe('flushController', () => {
...
+    it('does not notify if no message was added', () => {
+      sessionExpireObservable.notify()
+      expect(flushSpy).not.toHaveBeenCalled()
+    })
+
+    it('notifies when the session expires even if no message have been fully added yet', () => {
+      flushController.notifyBeforeAddMessage(SMALL_MESSAGE_BYTE_COUNT)
+      sessionExpireObservable.notify()
+      expect(flushSpy).toHaveBeenCalled()
+    })
```
bcaudan, line 108: **❓ question:** ‏I am wondering if we could not have a chance to have an incorrect state in this case 🤔

| operation | flushController | batch/segment |
|----------| ----------------|----------------|
| notifyBeforeAddMessage() | messageCount = 1 | bufferSize = 0 |
| expire flush | messageCount = 0 | bufferSize = 0 |
| addMessage | messageCount = 0 | bufferSize = 1 |
| notifyAfterAddMessage() | messageCount = 0 | bufferSize = 1 |

In this case a subsequent flush without other messages added would incorrectly not trigger the batch/segment flush, right?

BenoitZugmeyer: This is true if `expire flush` can happen between `notifyBeforeAddMessage` and `addMessage`. In practice, `notifyBeforeAddMessage` and `addMessage` are happening consecutively and nothing can happen between the two.

This invariant cannot be guaranteed with the current design, but I thought that it was fine for the sake of simplicity.
I'll write a bit of documentation to clarify.

bcaudan: > In practice, notifyBeforeAddMessage and addMessage are happening consecutively and nothing can happen between the two.

True, I am wondering if it will still be the case when we will plug the flush controller to segment collection 🤔
Could it worth to do or plan something?

BenoitZugmeyer: Yes it will still be true for segment collection.

The idea is to call `notifyBeforeAddMessage` right before sending the message/record to the webworker, so it cannot be flushed between the two. Only `notifyAfterAddMessage` will be called asynchronously, when we know the message compressed size.

amortemousque (approved):

PR #2165, opened 2023-04-12
♻️ factorize LifeCycle and simplify its types

BenoitZugmeyer:
## Motivation

The `LifeCycle` implementation was duplicated in Logs and RUM, because we used [function overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to have proper typings and those cannot be abstracted away.

The typings around LifeCycle was a bit noisy, as each function signature needed to be repeated for each event type, and the data associated needed to be repeated for each function.

## Changes

Instead of function overloads, use an `EventMap` based implementation, similar to the pattern used by the TypeScript libraries to type event listeners.

This allows to factorize the LifeCycle code between RUM and Logs and simplify typings.

Also I added simple unit tests around LifeCycles.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/tools/abstractLifeCycle.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    const lifeCycle = new LifeCycle()
+    const subscriber1Spy = jasmine.createSpy()
+    const subscriber2Spy = jasmine.createSpy()
+    lifeCycle.subscribe('foo', subscriber1Spy)
+    lifeCycle.subscribe('foo', subscriber2Spy)
+
+    lifeCycle.notify('foo', 'bar')
+
+    expect(subscriber1Spy).toHaveBeenCalledOnceWith('bar')
+    expect(subscriber1Spy).toHaveBeenCalledOnceWith('bar')
```
yannickadam, line 24: subscriber2Spy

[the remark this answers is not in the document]
BenoitZugmeyer: Ah, thanks!

packages/logs/src/domain/logsCollection/console/consoleCollection.ts
```diff
@@ -35,7 +35,7 @@ export function startConsoleCollection(configuration: LogsConfiguration, lifeCyc
               }
             : undefined,
         status: LogStatusForApi[log.api],
-      },
+      } satisfies RawConsoleLogsEvent,
```
amortemousque, line 38: **❓ question:** It seems that the type is inferred thanks to the `origin` attribute . Is it worth having the `satisfies RawConsoleLogsEvent`?‏

yannickadam (approved):

amortemousque (approved):

PR #2166, opened 2023-04-13
✨ [RUMF-1534] send a view update when session is expiring

BenoitZugmeyer:
## Motivation

Tell the backend to end the session early when we detect that the session expired, especially useful when using `stopSession()` manually.

## Changes

With this PR, the SDK will send a view update with `@session.is_active:false` every time it detects that the session expires.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2167, opened 2023-04-13
✨ [RUMF-1534] allow (some) view updates after session expiration

BenoitZugmeyer:
## Motivation

This PR is a preliminary step for notifying the backend that a session has expired via a View update. In other words, the goal is [to tackle this TODO](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/assembly.ts#L92-L95). To do so, we need to send a view update after the session has expired, which is not possible right now.

## Changes

This PR makes sure we don't send too many view updates when the session is expired. In particular, when the session expires, we now:

* end the current view which stops updates from metrics
* stops the session keep alive interval
* prevent any initial view updates
* do not accept new timings

A view can still be updated after the session expires:
* child events that started before session expiration are still collected, we need to update the view counter
* just when the session has expired, we send a last view update

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
@@ -324,7 +324,7 @@ describe('view lifecycle', () => {
       expect(getViewUpdateCount()).toBe(2)
     })
 
-    it('should not start a new view the session expires', () => {
+    it('should not start a new view if the session expired', () => {
       const { lifeCycle } = setupBuilder.build()
       const { getViewCreateCount } = viewTest
 
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
@@ -688,6 +688,21 @@ describe('view custom timings', () => {
     })
     expect(displaySpy).toHaveBeenCalled()
   })
+
+  it('should not add custom timing when the session has expired', () => {
+    const { clock, lifeCycle } = setupBuilder.build()
+    const { getViewUpdateCount, addTiming } = viewTest
+
+    lifeCycle.notify(LifeCycleEventType.SESSION_EXPIRED)
+
+    expect(getViewUpdateCount()).toBe(2)
+
+    addTiming('foo', relativeNow())
+
+    clock.tick(THROTTLE_VIEW_UPDATE_PERIOD)
+
+    expect(getViewUpdateCount()).toBe(2)
+  })
 })
 
 describe('start view', () => {
--- a/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
@@ -18,7 +18,7 @@ import { RumEventType, ViewLoadingType } from '../../../rawRumEvent.types'
 import type { RumEvent } from '../../../rumEvent.types'
 import { LifeCycleEventType } from '../../lifeCycle'
 import type { ViewEvent } from './trackViews'
-import { THROTTLE_VIEW_UPDATE_PERIOD } from './trackViews'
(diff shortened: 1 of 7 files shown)


BenoitZugmeyer:
* main
  * **PR #2167** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2167" target="_blank">Graphite</a>  👈
    * **PR #2166** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2166" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2167?utm_source=stack-comment).

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -190,6 +190,22 @@ describe('initial view', () => {
       expect(getViewUpdate(2).timings).toEqual({})
     })
 
+    it('should not update timings after session expires', () => {
```
bcaudan, line 193: **❓ question:** ‏was there an issue with updating the initial view after session expiration?

BenoitZugmeyer: There was no issue, but also it was not updated after session expiration since the update got discarded in assembly. I chose preserve this behavior, but it's a bit arbitrary.

Updating the initial view after session expiration should be fine: it's an edge case so any potential negative impact would be quite limited, and it would make the code slightly simpler.

bcaudan: it seems that we introduced that after an unexpected bump in session duration https://github.com/DataDog/browser-sdk/pull/1228 🤔
However, we were still tracking events count indefinitely at the time so that was maybe more related to event counts than to the initial view timings.
Maybe we could have something similar to `scheduleStopEventCountsTracking` and schedule the end of the timing updates when the initial view ends instead of doing that on session expiration.
wdyt?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.spec.ts
```diff
@@ -271,6 +256,27 @@ describe('initial view', () => {
         expect(initialView.last.loadingTime).toBe(FAKE_NAVIGATION_ENTRY.loadEventEnd)
       })
     })
+
+    it('should not update timings long after the view ended', () => {
```
bcaudan, line 260: **👏 praise:** ‏Nice!

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -212,6 +171,19 @@ function newView(
...
+    ;({ scheduleStop: scheduleStopInitialViewTimingsTracking } = trackInitialViewTimings(lifeCycle, (newTimings) => {
+      timings = newTimings
+      if (newTimings.loadEvent !== undefined) {
+        setLoadEvent(newTimings.loadEvent)
+      }
+      scheduleViewUpdate()
+    }))
+  } else {
+    scheduleStopInitialViewTimingsTracking = noop
+  }
```
bcaudan, line 185: **🥜 nitpick:** ‏there is a lot going there, it could be interesting to extract a function for the timings update. 
After reducing the complexity, using a ternary there could also remove a bit of boilerplate, something like:

```ts
const { scheduleStop: scheduleStopInitialViewTimingsTracking } = loadingType === ViewLoadingType.INITIAL_LOAD
    ? trackInitialViewTimings(lifeCycle, updateViewTimings)
    : { scheduleStop: noop }

...

function updateViewTimings(...) {
  ...
}
```

bcaudan (approved):

bcaudan (approved):

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: Even better!

yannickadam (approved):

PR #2173, opened 2023-04-17
🐛 Fix exception when using wrong tracing parameters

yannickadam:
## Motivation

When using wrong parameter types in `allowedTracingUrls`, such as:
`allowedTracingUrls: [42]`
The SDK properly warns that this parameter is invalid and will be ignored. However, it fails later when serializing parameters for telemetry, and stops initializing with no warning. 

## Changes

Add a check to ensure the parameter can be parsed for telemetry.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/configuration.spec.ts
+++ b/packages/rum-core/src/domain/configuration.spec.ts
@@ -541,6 +541,14 @@ describe('validateAndBuildRumConfiguration', () => {
           jasmine.arrayWithExactContents(['datadog', 'b3', 'b3multi', 'tracecontext'])
         )
       })
+
+      it('should survive a configuration with wrong parameters', () => {
+        const wrongTracingConfig: RumInitConfiguration = {
+          ...DEFAULT_INIT_CONFIGURATION,
+          allowedTracingUrls: [42 as any, { match: 'test', propagatorTypes: 42 }, undefined, null, {}],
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer (approved):

packages/rum-core/src/domain/configuration.spec.ts
```diff
@@ -541,6 +541,14 @@ describe('validateAndBuildRumConfiguration', () => {
...
         )
       })
+
+      it('should survive a configuration with wrong parameters', () => {
+        const wrongTracingConfig: RumInitConfiguration = {
+          ...DEFAULT_INIT_CONFIGURATION,
+          allowedTracingUrls: [42 as any, { match: 'test', propagatorTypes: 42 }, undefined, null, {}],
+        }
+        expect(serializeRumConfiguration(wrongTracingConfig).selected_tracing_propagators).toEqual([])
+      })
```
bcaudan, line 551: **❓ question:** ‏since the serialization is used for the telemetry, what about sending wrong parameters somehow as well?

bcaudan: ok, nevermind then

bcaudan (approved):

PR #2176, opened 2023-04-19
Create repository.datadog.yml

mimfgg:


repository.datadog.yml
```diff
@@ -0,0 +1,7 @@
+---
+schema-version: v1
+kind: mergequeue
+team: ci-interfaces
```
BenoitZugmeyer, line 4: **❓ question:** I don't see any doc for the `team` option. ‏Does it correspond do the GitHub team like `DataDog/[team]`? Shouldn't the team be `rum-browser`?

repository.datadog.yml
```diff
@@ -0,0 +1,7 @@
...
+enable: true
+merge_method: merge
```
BenoitZugmeyer: **💬 suggestion:** ‏we are enforcing 'squash' merges on the main branch:

```suggestion
merge_method: squash
```

mimfgg: yeah, I removed the option already, the default for the repo will be used.

repository.datadog.yml
mimfgg: good question ... I don't think we use the field for the moment. Let's change to rum-browser

BenoitZugmeyer (approved):

PR #2179, opened 2023-04-20
♻️ [RUMF-1508] refactor error types / constants

bcaudan:
## Motivation

Extract error refactoring from #2050 

## Changes

- extract error.types
- use an enum for Uncaught/Provided error message prefix 
- increase usage of no stack message 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2181, opened 2023-04-20
♻️ [RUMF-1508] reorganise error handling

bcaudan:
## Motivation

Prepare handling of error without stacktrace

## Changes

- add more tests
- reorganise instrumentOnError
- reorganise computeRawError

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/error/error.ts
+++ b/packages/core/src/domain/error/error.ts
@@ -36,7 +36,7 @@ export function computeRawError({
     : !isErrorInstance
     ? `${nonErrorPrefix} ${jsonStringify(sanitize(originalError))!}`
     : 'Empty message'
-  const stack = isValidStackTrace(isErrorInstance, stackTrace)
+  const stack = hasUsableStack(isErrorInstance, stackTrace)
     ? toStackTraceString(stackTrace)
     : NO_ERROR_STACK_PRESENT_MESSAGE
   const causes = isErrorInstance ? flattenErrorCauses(originalError as ErrorWithCause, source) : undefined
@@ -55,7 +55,7 @@ export function computeRawError({
   }
 }
 
-function isValidStackTrace(isErrorInstance: boolean, stackTrace?: StackTrace): stackTrace is StackTrace {
+function hasUsableStack(isErrorInstance: boolean, stackTrace?: StackTrace): stackTrace is StackTrace {
   if (stackTrace === undefined) {
     return false
   }
--- a/packages/core/src/domain/error/error.ts
+++ b/packages/core/src/domain/error/error.ts
@@ -30,33 +30,41 @@ export function computeRawError({
   source,
   handling,
 }: RawErrorParams): RawError {
-  if (!stackTrace || (stackTrace.message === undefined && !(originalError instanceof Error))) {
-    const sanitizedError = isExperimentalFeatureEnabled(ExperimentalFeature.SANITIZE_INPUTS)
-      ? sanitize(originalError)
-      : originalError
-    return {
-      startClocks,
-      source,
-      handling,
-      originalError: sanitizedError,
-      message: `${nonErrorPrefix} ${jsonStringify(sanitizedError)!}`,
-      stack: NO_ERROR_STACK_PRESENT_MESSAGE,
-      handlingStack,
-      type: stackTrace && stackTrace.name,
-    }
-  }
+  const isErrorInstance = originalError instanceof Error
+  const sanitizedError = isExperimentalFeatureEnabled(ExperimentalFeature.SANITIZE_INPUTS)
+    ? sanitize(originalError)
+    : originalError
+
+  const message = stackTrace?.message
+    ? stackTrace.message
+    : !isErrorInstance
+    ? `${nonErrorPrefix} ${jsonStringify(sanitizedError)!}`
+    : 'Empty message'
+  const stack = isValidStackTrace(isErrorInstance, stackTrace)
+    ? toStackTraceString(stackTrace)
+    : NO_ERROR_STACK_PRESENT_MESSAGE
+  const causes = isErrorInstance ? flattenErrorCauses(originalError as ErrorWithCause, source) : undefined
+  const type = stackTrace && stackTrace.name
 
   return {
     startClocks,
     source,
     handling,
-    originalError,
-    message: stackTrace.message || 'Empty message',
-    stack: toStackTraceString(stackTrace),
     handlingStack,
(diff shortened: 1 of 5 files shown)


bcaudan:
* main
  * **PR #2179** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2179" target="_blank">Graphite</a> 
    * **PR #2181** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2181" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2181?utm_source=stack-comment).

packages/core/src/domain/error/error.ts
```diff
@@ -30,35 +30,43 @@ export function computeRawError({
...
 }
 
+function isValidStackTrace(isErrorInstance: boolean, stackTrace?: StackTrace): stackTrace is StackTrace {
+  return (
+    stackTrace !== undefined &&
+    (isErrorInstance ||
+      (stackTrace.stack.length > 0 && (stackTrace.stack.length > 1 || stackTrace.stack[0].url !== undefined)))
+  )
+}
+
```
amortemousque, line 69: **💬 suggestion:** ‏ To make it easier to read (at least for me) :) 

```suggestion
function isValidStackTrace(isErrorInstance: boolean, stackTrace?: StackTrace): stackTrace is StackTrace {
  if (stackTrace === undefined) {
    return false
  }

  if (isErrorInstance) {
    return true
  }

  return stackTrace.stack.length > 0 && (stackTrace.stack.length > 1 || stackTrace.stack[0].url !== undefined)
}
```

packages/core/src/domain/error/error.ts
```diff
@@ -30,35 +30,43 @@ export function computeRawError({
...
+  const message = stackTrace?.message
+    ? stackTrace.message
+    : !isErrorInstance
+    ? `${nonErrorPrefix} ${jsonStringify(sanitizedError)!}`
+    : 'Empty message'
+  const stack = isValidStackTrace(isErrorInstance, stackTrace)
+    ? toStackTraceString(stackTrace)
+    : NO_ERROR_STACK_PRESENT_MESSAGE
+  const causes = isErrorInstance ? flattenErrorCauses(originalError as ErrorWithCause, source) : undefined
+  const type = stackTrace && stackTrace.name
```
amortemousque, line 47: **💬 suggestion:** ‏Why not simply 
```suggestion
  const type = stackTrace?.name
```

packages/core/src/domain/error/trackRuntimeError.spec.ts
```diff
@@ -15,6 +15,14 @@ describe('runtime error tracker', () => {
   let notifyError: jasmine.Spy
   let stopRuntimeErrorTracking: () => void
 
+  function afterOnErrorInstrumentation(callback: () => void) {
+    const instrumentedOnError = window.onerror!
+    window.onerror = function () {
+      instrumentedOnError.apply(window, arguments as any)
+      callback()
+    }
```
BenoitZugmeyer, line 23: **💬 suggestion:** ‏we already have a spy on `onerror`, maybe just use `onErrorSpy.and.callFake(callback)`? Or you could use `collectAsyncCalls`

packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -55,41 +55,36 @@ export function startUnhandledErrorCollection(callback: UnhandledErrorCallback)
...
-        callback(stack, errorObj)
-      } else {
-        const location = {
-          url,
-          column: columnNo,
-          line: lineNo,
-        }
-
-        let name
-        let msg = message
```
BenoitZugmeyer, line 72: **🔨 warning:** ‏I think you missed this: if `message` is a string, it should be the default value in case it does not match `ERROR_TYPES_RE`

packages/core/src/domain/error/error.ts
```diff
@@ -29,30 +29,40 @@ export function computeRawError({
...
+}
+
+function isValidStackTrace(isErrorInstance: boolean, stackTrace?: StackTrace): stackTrace is StackTrace {
+  if (stackTrace === undefined) {
+    return false
+  }
+  if (isErrorInstance) {
+    return true
   }
+  return stackTrace.stack.length > 0 && (stackTrace.stack.length > 1 || stackTrace.stack[0].url !== undefined)
```
BenoitZugmeyer, line 65: **❓ question:** ‏I don't see this check in the original code, why do we need it now?

packages/core/src/domain/error/error.ts
```diff
@@ -29,30 +29,40 @@ export function computeRawError({
...
-    causes: flattenErrorCauses(originalError as ErrorWithCause, source),
+    originalError,
+    type,
+    message,
+    stack,
+    causes,
+  }
+}
+
+function isValidStackTrace(isErrorInstance: boolean, stackTrace?: StackTrace): stackTrace is StackTrace {
```
BenoitZugmeyer, line 58: **💬 suggestion:** ‏ ‏this function name indicates that we can have valid and invalid stack traces. In `computeRawError`, we are not checking whether `stackTrace` is valid everytime we use it, implying that we are using invalid stacktraces in some places. We should use a more explicit function name here.

packages/core/src/domain/error/error.ts
bcaudan: There are cases where `stackTrace.stack = []` or `stackTrace.stack = [{url: undefined, column: undefined, line: undefined}]` and we want to use the `NO_ERROR_STACK_PRESENT_MESSAGE` for those.
In the original code, those cases happened in the first if but it will change in the following PR.

packages/core/src/domain/tracekit/tracekit.ts
bcaudan: With 
```javascript
const ERROR_TYPES_RE =
  /^(?:[Uu]ncaught (?:exception: )?)?(?:((?:Eval|Internal|Range|Reference|Syntax|Type|URI|)Error): )?(.*)$/
```
if message is a string it should always match, no?

BenoitZugmeyer: Ah that's right, so you could consider that `ERROR_TYPES_RE.exec(messageObj as string)` always succeeds then.

BenoitZugmeyer (approved):

PR #2182, opened 2023-04-21
✨[RUMF-1508] Provide stack trace for all uncaught exceptions

bcaudan:
## Motivation

Errors require a stack trace to be automatically grouped by error tracking.
When an uncaught exception is not an error instance, we don't provide a stack trace. 
However, in some cases some info can be retrieved from the [window.onerror handler arguments](https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event#arguments).

## Changes

- Better handle non error instance:
  - Avoid to compute stack trace but build one from from handler arguments
  - Pass to the caller the thrown object instead of the message
- Used computed stack trace message only when it brings extra value (error instance or successfully parsed error message) 

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/error/error.ts
+++ b/packages/core/src/domain/error/error.ts
@@ -31,11 +31,12 @@ export function computeRawError({
 }: RawErrorParams): RawError {
   const isErrorInstance = originalError instanceof Error
 
-  const message = stackTrace?.message
-    ? stackTrace.message
-    : !isErrorInstance
-    ? `${nonErrorPrefix} ${jsonStringify(sanitize(originalError))!}`
-    : 'Empty message'
+  const message =
+    stackTrace?.message && stackTrace?.name
+      ? stackTrace.message
+      : !isErrorInstance
+      ? `${nonErrorPrefix} ${jsonStringify(sanitize(originalError))!}`
+      : 'Empty message'
   const stack = hasUsableStack(isErrorInstance, stackTrace)
     ? toStackTraceString(stackTrace)
     : NO_ERROR_STACK_PRESENT_MESSAGE
--- a/packages/core/src/domain/error/error.ts
+++ b/packages/core/src/domain/error/error.ts
@@ -31,12 +31,7 @@ export function computeRawError({
 }: RawErrorParams): RawError {
   const isErrorInstance = originalError instanceof Error
 
-  const message =
-    stackTrace?.message && stackTrace?.name
-      ? stackTrace.message
-      : !isErrorInstance
-      ? `${nonErrorPrefix} ${jsonStringify(sanitize(originalError))!}`
-      : 'Empty message'
+  const message = computeMessage(stackTrace, isErrorInstance, nonErrorPrefix, originalError)
   const stack = hasUsableStack(isErrorInstance, stackTrace)
     ? toStackTraceString(stackTrace)
     : NO_ERROR_STACK_PRESENT_MESSAGE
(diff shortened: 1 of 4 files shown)


bcaudan:
* main
  * **PR #2179** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2179" target="_blank">Graphite</a> 
    * **PR #2181** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2181" target="_blank">Graphite</a> 
      * **PR #2182** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2182" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2182?utm_source=stack-comment).

packages/core/src/domain/error/error.ts
```diff
@@ -31,11 +31,12 @@ export function computeRawError({
 }: RawErrorParams): RawError {
   const isErrorInstance = originalError instanceof Error
 
-  const message = stackTrace?.message
-    ? stackTrace.message
-    : !isErrorInstance
-    ? `${nonErrorPrefix} ${jsonStringify(sanitize(originalError))!}`
-    : 'Empty message'
+  const message =
+    stackTrace?.message && stackTrace?.name
```
BenoitZugmeyer, line 35: **💬 suggestion:** ‏I don't understand why we check for `stackTrace?.name` here. If this is now a required condition to consider whether we should take `message` into account, we could handle this when creating the `StackTrace` object and don't set a `stackTrace.message` if there is no `name`. WDYT?

packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -67,7 +67,7 @@ function instrumentOnError(callback: UnhandledErrorCallback) {
         message,
         stack,
       }
-      callback(stackTrace, messageObj)
+      callback(stackTrace, errorObj ?? messageObj)
```
BenoitZugmeyer, line 70: **💬 suggestion:** ‏this could be refactored like this:

```ts
let stackTrace: StackTrace
if (errorObj instanceof Error) {
     stackTrace = computeStackTrace(errorObj)
} else {
      const stack = [{ url, column, line }]
      const { name, message } = tryToParseMessage(messageObj)
      stackTrace = {
        name,
        message,
        stack,
      }
}
callback(stackTrace, errorObj ?? messageObj)
```

packages/core/src/domain/error/error.ts
bcaudan: > I don't understand why we check for `stackTrace?.name` here

The idea is that if we did not manage to get a name, it means that we are in a case where the message would not be great either and we already generate our own message for those ones. We need to add this extra check with the new path taken by non error uncaught exception.

> we could handle this when creating the StackTrace object and don't set a stackTrace.message if there is no name. WDYT?

For now, I tried to not impact tracekit logic/design that much but we could reduce several complexity spots reworking some parts and by making it more integrated with our own error logic.
I don't think that is the right time to do it though, I can collect some thoughts about that in a card for future us.
wdyt?

amortemousque (approved):

yannickadam (approved):

packages/core/src/domain/error/error.ts
BenoitZugmeyer: Okay. You could add a comment to give some context here, and maybe add a TODO to explain that we should rework tracekit design in the future.

bcaudan: comments added in https://github.com/DataDog/browser-sdk/pull/2182/commits/48e7bdff517a641f31699c43de4273a548c4bad1

PR #2186, opened 2023-04-25
✨[RUMF-1573] allow to add modifiable field paths with beforeSend

bcaudan:
## Motivation

Prerequisite for error custom fingerprint

## Changes

- specify for each event type which field is modifiable and its type
- allow to add modifiable fields not present with beforeSend

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/limitModification.ts
+++ b/packages/rum-core/src/domain/limitModification.ts
@@ -1,7 +1,7 @@
 import { sanitize, deepClone, getType, objectEntries } from '@datadog/browser-core'
 import type { Context } from '@datadog/browser-core'
 
-export type ModifiableFieldPaths = { [path: string]: 'string' | 'object' }
+export type ModifiableFieldPaths = Record<string, 'string' | 'object'>
 
 /**
  * Current limitation:
@@ -53,10 +53,10 @@ function set(object: unknown, path: string, value: unknown) {
   }
 }
 
-function isValidObject(object: unknown): object is { [key: string]: unknown } {
+function isValidObject(object: unknown): object is Record<string, unknown> {
   return getType(object) === 'object'
 }
 
-function isValidObjectContaining(object: unknown, field: string): object is { [key: string]: unknown } {
+function isValidObjectContaining(object: unknown, field: string): object is Record<string, unknown> {
   return isValidObject(object) && Object.prototype.hasOwnProperty.call(object, field)
 }
--- a/packages/rum-core/src/domain/limitModification.spec.ts
+++ b/packages/rum-core/src/domain/limitModification.spec.ts
@@ -175,7 +175,7 @@ describe('limitModification', () => {
   })
 })
 
-function generateModifiableFieldPathsFrom(object: { [path: string]: string | object }) {
+function generateModifiableFieldPathsFrom(object: Record<string, string | object>) {
   const modifiableFieldPaths: ModifiableFieldPaths = {}
   objectEntries(object).forEach(([key, value]) => {
     modifiableFieldPaths[key] = typeof value as 'object' | 'string'


packages/rum-core/src/domain/assembly.ts
```diff
@@ -68,6 +65,33 @@ export function startRumAssembly(
   buildCommonContext: () => CommonContext,
   reportError: (error: RawError) => void
 ) {
+  MODIFIABLE_FIELD_PATHS_BY_EVENT = {
+    [RumEventType.VIEW]: VIEW_MODIFIABLE_FIELD_PATHS,
+    [RumEventType.ERROR]: assign(
+      {
+        'error.message': 'string',
+        'error.stack': 'string',
+        'error.resource.url': 'string',
```
bcaudan, line 74: From @yannickadam, see if we can retrieve some existing types rather than re-declaring it here

bcaudan:
* main
  * **PR #2186** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2186" target="_blank">Graphite</a>  👈
    * **PR #2189** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2189" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2186?utm_source=stack-comment).

packages/rum-core/src/domain/assembly.ts
bcaudan: Actually, it does not seem so since elsewhere it is only typings and here we need a value.

packages/rum-core/src/domain/assembly.ts
```diff
@@ -68,6 +65,33 @@ export function startRumAssembly(
   buildCommonContext: () => CommonContext,
   reportError: (error: RawError) => void
 ) {
+  MODIFIABLE_FIELD_PATHS_BY_EVENT = {
```
BenoitZugmeyer, line 68: **💬 suggestion:** ‏use camelcase for this variable


**❓ question:** ‏do we really need to detail each field for each event type? Isn't what we did before enough? I feel like this is extra verbose for not much benefit.

packages/rum-core/src/domain/limitModification.ts
```diff
@@ -53,6 +53,10 @@ function set(object: unknown, path: string, value: unknown) {
   }
 }
 
+function isValidObject(object: unknown): object is { [key: string]: unknown } {
+  return typeof object === 'object' && object !== null
```
BenoitZugmeyer, line 57: **💬 suggestion:** ‏this could be replaced with `getType(object) === 'object'`

yannickadam (approved):

packages/rum-core/src/domain/assembly.ts
bcaudan: > Isn't what we did before enough?

I started with it but with the new possibility of adding attributes that are not on the original object, I was concerned that we unexpectedly allow to add a modifiable attribute on the wrong event type.
I tried to express that with this test:  https://github.com/DataDog/browser-sdk/pull/2186/files#diff-0cc30fdbf6d153bf8bacf2743993dbfbe6cfc8441d98ec7cf714a6213614d393R246

wdyt?

packages/rum-core/src/domain/assembly.ts
yannickadam: 👍

packages/rum-core/src/domain/assembly.ts
BenoitZugmeyer: Ah right I understand that there is a higher change of misusing those fields now. Sounds good to me then!

BenoitZugmeyer (approved):

PR #2209, opened 2023-05-04
🐛 fix isolatedModules support

### LINKED ISSUE #2208: TypeScripts changes causes build issues
**Describe the bug**
Ever since version 4.40.0 (and I'm assuming the type changes that went in for it, our build system is broken. Here is the error I receive:

```
Error: node_modules/@datadog/browser-rum-core/cjs/domain/lifeCycle.d.ts(27,6): error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.
```

**To Reproduce**
Not sure I can provide a repro case. My tsconfig for the project is
```
  "compilerOptions": {
    "allowSyntheticDefaultImports": true,
    "baseUrl": "src",
    "composite": true,
    "isolatedModules": true,
    "lib": ["DOM", "DOM.Iterable", "ESNext"],
    "module": "esnext",
    "moduleResolution": "node",
    "noEmit": true,
    "resolveJsonModule": true,
    "skipLibCheck": false,
    "sourceMap": false,
    "target": "ES6"
  },
```

**Expected behavior**
A breaking change like this should be done in a major version although I suspect this is a bug

BenoitZugmeyer:
## Motivation

Support TS projects using `isolatedModules: true` 

This PR fixes #2208.

Bug reported here: https://github.com/microsoft/TypeScript/issues/54152

## Changes

* add an integration test to make sure we are compatible with this option
* use a workaround to fix the TS error

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/lifeCycle.ts
+++ b/packages/rum-core/src/domain/lifeCycle.ts
@@ -47,7 +47,9 @@ export const enum LifeCycleEventType {
 // "declare" statement, it will only be used during typecheck and completely ignored when building
 // JavaScript).
 //
-// See issue https://github.com/DataDog/browser-sdk/issues/2208
+// See issues:
+// * https://github.com/DataDog/browser-sdk/issues/2208
+// * https://github.com/microsoft/TypeScript/issues/54152
 declare const LifeCycleEventTypeAsConst: {
   PERFORMANCE_ENTRIES_COLLECTED: LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED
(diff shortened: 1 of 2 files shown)


yannickadam (approved):
While alternatives exists, they have impacts on the bundle size:
enum => ~ +1800 bytes
const object => ~ + 120 bytes
Since we do not consider that LifeCycleEventType needs to be exposed, this workaround seems to be the most appropriate for now.

PR #2254, opened 2023-05-17
👷🐛 fix build_json2type

BenoitZugmeyer:
## Motivation

With yarn 3, the 'test/' folder is not present, so all built files are put directly in the 'dist/' folder instead of 'dist/src/'.

## Changes

Using an explicit '--rootDir' fixes this issue.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/cli
+++ b/scripts/cli
@@ -50,7 +50,13 @@ cmd_build_json2type () {
   set +e
   npm i
   npm run clean
-  npm run build:server
+
+  # With yarn 3, the 'test/' folder is not present, so all built files are put directly in the
+  # 'dist/' folder instead of 'dist/src/'.
+  #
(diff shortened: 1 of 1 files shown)


scripts/cli
```diff
@@ -50,7 +50,13 @@ cmd_build_json2type () {
   set +e
   npm i
   npm run clean
-  npm run build:server
+
+  # With yarn 3, the 'test/' folder is not present, so all built files are put directly in the
```
bcaudan, line 54: **❓ question:** ‏out of curiosity, how was created this test folder?

bcaudan: ok, makes sense!

PR #2255, opened 2023-05-17
🐛 [RUMF-1499] Don't send duration for resources crossing a page frozen state

amortemousque:
## Motivation

The resource duration for XHR and fetch is computed by the SDK if the corresponding resource timing isn’t found. If during the resource a page frozen state happens, we will wrongfully compute the duration:

image

 In these cases the browser SDK will now send resources without duration.

## Changes

Check if the request crossed a frozen state in resourceCollection.ts

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
+++ b/packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
@@ -25,6 +25,7 @@ import { LifeCycleEventType } from '../../lifeCycle'
 import type { RequestCompleteEvent } from '../../requestCollection'
 import type { RumSessionManager } from '../../rumSessionManager'
 import type { PageStateHistory } from '../../contexts/pageStateHistory'
+import { PageState } from '../../contexts/pageStateHistory'
 import { matchRequestTiming } from './matchRequestTiming'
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/rumEventsCollection/resource/resourceCollection.ts
```diff
@@ -224,3 +225,16 @@ function computePageStateInfo(pageStateHistory: PageStateHistory, startClocks: C
     },
   }
 }
+
+function computeRequestDuration(pageStateHistory: PageStateHistory, startClocks: ClocksState, duration: Duration) {
+  // TODO remove FF in next major
```
bcaudan, line 230: **❓ question:** ‏Do we need to wait for a major?

amortemousque: We could have some customers that rely on before_send and expect resource.duration to be defined

BenoitZugmeyer (approved):

bcaudan (approved):
it should probably be a ⚗️ instead of 🐛

PR #2294, opened 2023-06-14
💥[RUMF-1229] Logs: remove `error.origin` attribute

bcaudan:
## Motivation

Since the introduction of the `origin` attribute on all logs, `error.origin` is redundant.

## Changes

- Drop logs `error.origin` attribute in favour of `origin` attribute.
- Improve logs types

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/domain/logsCollection/console/consoleCollection.spec.ts
+++ b/packages/logs/src/domain/logsCollection/console/consoleCollection.spec.ts
@@ -1,14 +1,13 @@
-import { ErrorSource, noop } from '@datadog/browser-core'
+import { ErrorSource, noop, objectEntries } from '@datadog/browser-core'
 import type { RawConsoleLogsEvent } from '../../../rawLogsEvent.types'
 import { validateAndBuildLogsConfiguration } from '../../configuration'
 import type { RawLogsEventCollectedData } from '../../lifeCycle'
 import { LifeCycle, LifeCycleEventType } from '../../lifeCycle'
-import { StatusType } from '../../logger'
-import { startConsoleCollection } from './consoleCollection'
+import { startConsoleCollection, LogStatusForApi } from './consoleCollection'
 
 describe('console collection', () => {
   const initConfiguration = { clientToken: 'xxx', service: 'service' }
-  let consoleLogSpy: jasmine.Spy
+  let consoleSpies: { [key: string]: jasmine.Spy }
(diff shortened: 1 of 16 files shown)


packages/logs/src/domain/logsCollection/console/consoleCollection.spec.ts
```diff
@@ -58,7 +58,6 @@ describe('console collection', () => {
     console.error('foo', 'bar')
 
     expect(rawLogsEvents[0].rawLogsEvent.error).toEqual({
-      origin: ErrorSource.CONSOLE,
```
yannickadam, line 61: 💭 Maybe we could have one test checking on `rawLogsEvents[0].rawLogsEvent` to ensure `origin` is `ErrorSource.CONSOLE` when using `console.error` ?

bcaudan: I'd say that it is addressed by the test above: https://github.com/DataDog/browser-sdk/blob/991e97418579256bf36d38e3cbdaf77c7d6052c1/packages/logs/src/domain/logsCollection/console/consoleCollection.spec.ts#L31-L46

However, we could have this test for each status and not only for log.
wdyt?

yannickadam: Yes I was thinking that adding `console.error` would be good, but indeed `console.*` is even better.

packages/logs/src/domain/logsCollection/console/consoleCollection.spec.ts
```diff
@@ -20,32 +19,43 @@ describe('console collection', () => {
...
 
-    expect(rawLogsEvents[0].rawLogsEvent).toEqual({
-      date: jasmine.any(Number),
-      message: 'foo bar',
-      status: StatusType.info,
-      origin: ErrorSource.CONSOLE,
-      error: undefined,
-    })
+      /* eslint-disable-next-line no-console */
+      ;(console as BrowserConsole)[api]('foo', 'bar')
```
yannickadam, line 47: 💭 Would that work?

```suggestion
      console[api as keyof typeof LogStatusForApi]('foo', 'bar')
```

yannickadam (approved):

PR #2301, opened 2023-06-21
🐛 Handle local storage replication latency

yannickadam:
## Motivation

When instantiating two browser windows within a short interval (<10ms), there is a high risk that the lock mechanism will not work for local storage. There is a replication latency between browser windows of ~5ms (according to tests).

## Changes

- Postpone reading the lock by 7ms in case local storage is used. This leaves time for the storage to synchronize.
- Remove the lock when we reach max tries. 
--  The chances for this situation to happen with cookies is extremely low as we'd need to kill the browser in the middle of processStoreOperations, with a lock acquired.
-- However, the chances for this to happen are much higher for local storage: around 7 out of 1000 per opened window.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/session/sessionStoreOperations.spec.ts
+++ b/packages/core/src/domain/session/sessionStoreOperations.spec.ts
@@ -4,13 +4,8 @@ import type { CookieOptions } from '../../browser/cookie'
 import { initCookieStrategy } from './storeStrategies/sessionInCookie'
 import { initLocalStorageStrategy } from './storeStrategies/sessionInLocalStorage'
 import type { SessionState } from './sessionState'
-import { expandSessionState, toSessionString } from './sessionState'
-import {
-  processSessionStoreOperations,
-  isLockEnabled,
-  LOCK_MAX_TRIES,
-  LOCK_RETRY_DELAY,
-} from './sessionStoreOperations'
+import { expandSessionState, toSessionString, toSessionState } from './sessionState'
+import { processSessionStoreOperations, LOCK_MAX_TRIES, LOCK_RETRY_DELAY } from './sessionStoreOperations'
 import { SESSION_STORE_KEY } from './storeStrategies/sessionStoreStrategy'
 
 const cookieOptions: CookieOptions = {}
@@ -49,7 +44,7 @@ const cookieOptions: CookieOptions = {}
 
     describe('with lock access disabled', () => {
       beforeEach(() => {
-        isLockEnabled() && pending('lock-access required')
+        sessionStoreStrategy.lockOptions.enabled && pending('lock-access required')
       })
 
       it('should persist session when process returns a value', () => {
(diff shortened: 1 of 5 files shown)


packages/core/src/domain/session/sessionStoreOperations.spec.ts
```diff
@@ -236,18 +252,36 @@ const cookieOptions: CookieOptions = {}
...
+        processSessionStoreOperations({ process: processSpy, after: afterSpy }, sessionStoreStrategy)
+
+        const lockMaxTries = sessionStoreStrategy.lockOptions.enabled ? LOCK_MAX_TRIES : 0
+        const lockRetryDelay = sessionStoreStrategy.lockOptions.enabled ? LOCK_RETRY_DELAY : 0
+
+        clock.tick(lockMaxTries * lockRetryDelay)
+
+        expect(toSessionState(stubStorage.currentValue(storageKey)).lock).toBeUndefined()
+
+        clock.cleanup()
```
BenoitZugmeyer, line 281: **💬 suggestion:** ‏cleanups need to be in an `afterEach` hook (or maybe a `finally` statement?) else if the test is failing midway, it'll interfer with other tests.

packages/core/src/domain/session/sessionStoreOperations.ts
```diff
@@ -10,54 +9,89 @@ type Operations = {
...
+    // Try to acquire our lock
+    if (currentSession.lock === undefined) {
+      currentLock = generateUUID()
+      currentSession.lock = currentLock
+      persistSession(currentSession)
+
+      // Here, typically for local storage, we must WAIT before retrieving the lock. Else, there is a chance another
+      // process can unknowingly write to the storage. This happens because of a latency to replicate the written value.
+      if (lockOptions.synchronizationLatency > 0) {
+        retryLater(operations, sessionStoreStrategy, numberOfRetries, lockOptions.synchronizationLatency)
```
BenoitZugmeyer, line 75: **💬 suggestion:** ‏`retry` is a bit confusing here, as you don't want to retry but actually continue the operation after a short while.

Maybe you could define a function that wraps everything after this `if`, and call it in a timeout:

```ts
      if (lockOptions.synchronizationLatency > 0) {
        setTimeout(onLockWritten, lockOptions.synchronizationLatency)
      } else {
        onLockWritten()
      }

  function onLockWritten() {
    if (lockOptions.enabled) {
      // If there is no latency, we retrieve our session to check we're still owner of our lock.
      // If the lock has changed, postpone the operation
      currentSession = retrieveSession()
      if (currentSession.lock !== currentLock) {
        retryLater(operations, sessionStoreStrategy, numberOfRetries)
        return
      }
    }

    let processedSession = operations.process(currentSession)

    if (lockOptions.enabled) {
      // If the lock has changed after process, retry later
      currentSession = retrieveSession()
      if (currentSession.lock !== currentLock) {
        retryLater(operations, sessionStoreStrategy, numberOfRetries)
        return
      }
    }

    ...
  }
```

Also, with this design, you wouldn't need to store the current lock value in the module scope (which adds complexity)

PR #2306, opened 2023-06-27
🐛 [RUM-231] Fix location.origin is "null" for file: URIs

amortemousque:
## Motivation

On Firefox window.location.origin is "null" for file: URIs, making our url normalisation fail.
(cf: https://bugzilla.mozilla.org/show_bug.cgi?id=878297)

## Changes

Update getLinkElementOrigin to handle "null" case

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/utils/urlPolyfill.spec.ts
+++ b/packages/core/src/tools/utils/urlPolyfill.spec.ts
@@ -1,5 +1,14 @@
 import { isFirefox } from '../../../test'
-import { getHash, getOrigin, getPathName, getSearch, isValidUrl, normalizeUrl, getLocationOrigin } from './urlPolyfill'
+import {
+  getHash,
+  getOrigin,
+  getPathName,
+  getSearch,
+  isValidUrl,
+  normalizeUrl,
+  getLocationOrigin,
+  getLinkElementOrigin,
+} from './urlPolyfill'
 
 describe('normalize url', () => {
   it('should add origin to relative path', () => {
@@ -15,8 +24,13 @@ describe('normalize url', () => {
   })
 
   it('should keep non http url unchanged', () => {
+    expect(normalizeUrl('ftp://foo.com/my/path')).toEqual('ftp://foo.com/my/path')
+  })
+
+  it('should keep file url unchanged', () => {
     if (isFirefox()) {
-      pending('https://bugzilla.mozilla.org/show_bug.cgi?id=1578787')
+      // On firefox, URL host is empty for file URI: 'https://bugzilla.mozilla.org/show_bug.cgi?id=1578787'
+      expect(normalizeUrl('file://foo.com/my/path')).toMatch(/^file:\/\/.*\/my\/path$/)
     }
     expect(normalizeUrl('file://foo.com/my/path')).toEqual('file://foo.com/my/path')
   })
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -12,10 +12,6 @@ export function isValidUrl(url: string) {
   }
 }
 
-export function haveSameOrigin(url1: string, url2: string) {
```
amortemousque, line 15: Not used

packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -77,11 +73,12 @@ export function getLocationOrigin() {
 }
 
 /**
- * IE fallback
- * https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/origin
+ * Fallback
+ * On IE HTMLAnchorElement origin is not supported: https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/origin
+ * On Firefox window.location.origin is "null" for file: URIs: https://bugzilla.mozilla.org/show_bug.cgi?id=878297
```
bcaudan, line 78: **💬 suggestion:** ‏what about adding a test case for file URI?

amortemousque: Hum, there are two issues with Firefox with file URI:
- The URL.origin contain a "null"
- The URL.host is empty

Which is why we have this [check](https://github.com/DataDog/browser-sdk/blob/91ff5786bd99a3ad8059d7f0f677c2ea550351e6/packages/core/src/tools/utils/urlPolyfill.spec.ts#L26) in our tests

My fix was only not to make the normalizeUrl() break on firefox for file URI, however `normalizeUrl("file://foo.bar")` will return
- "file:///" for firefox
- "file://foo.bar" for other browser 

I tried to reflect that in our tests

packages/core/src/tools/utils/urlPolyfill.spec.ts
```diff
@@ -15,8 +24,13 @@ describe('normalize url', () => {
   })
 
   it('should keep non http url unchanged', () => {
+    expect(normalizeUrl('ftp://foo.com/my/path')).toEqual('ftp://foo.com/my/path')
+  })
+
+  it('should keep file url unchanged', () => {
     if (isFirefox()) {
-      pending('https://bugzilla.mozilla.org/show_bug.cgi?id=1578787')
+      // On firefox, URL host is empty for file URI: 'https://bugzilla.mozilla.org/show_bug.cgi?id=1578787'
```
amortemousque, line 32: I made the test for firefox ignoring the host so that if they fix the issue the test still works

packages/core/src/tools/utils/urlPolyfill.spec.ts
```diff
@@ -15,8 +24,13 @@ describe('normalize url', () => {
...
+      expect(normalizeUrl('file://foo.com/my/path')).toMatch(/^file:\/\/.*\/my\/path$/)
```
amortemousque, line 33: I made the test for firefox ignoring the host so that if they fix the issue the test still works

bcaudan: I'd be in favour of not doing that.
It would make the test clearer to have the exact output and when the issue will be solved and after we update the firefox version, we could remove this if instead of keeping it.

packages/core/src/tools/utils/urlPolyfill.spec.ts
```diff
@@ -66,3 +81,9 @@ describe('getHash', () => {
     expect(getHash('http://www.datadoghq.com/foo/bar?a=b#hello')).toBe('#hello')
   })
 })
+
+describe('getLinkElementOrigin', () => {
+  it('should retrieve the url origin if URL origin is "null"', () => {
+    expect(getLinkElementOrigin({ origin: 'null', host: 'foo.bar', protocol: 'file:' } as URL)).toBe('file://foo.bar')
+  })
+})
```
bcaudan, line 89: **💬 suggestion:** Since we already have tests on `getOrigin`, could we test this case with `getOrigin`?

bcaudan (approved):

PR #2308, opened 2023-06-27
💥 [RUMF-1473] Ignore untrusted event

amortemousque:
## Motivation

[Untrusted events](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) (`event.isTrusted: false`) are arbitrarily crafted by the application: some properties expected by the SDK might be missing or might not reflect legitimate data. Therefore untrusted event  are now ignored.

It's important to note that:
- some data may not be collected anymore, some behaviour may not be taken into account
- for untrusted event to be collected by the browser SDK, add `event.__ddIsTrusted: true`

## Changes

Ignore untrusted events

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/browser/addEventListener.spec.ts
```diff
@@ -33,4 +33,27 @@ describe('addEventListener', () => {
       expect(zoneJsPatchedRemoveEventListener).not.toHaveBeenCalled()
     })
   })
+
+  describe('Untrusted event', () => {
+    it('should be ignored if __ddIsTrusted is false', () => {
```
bcaudan, line 38: **💬 suggestion:** ‏what about adding a test for ignored when `__ddIsTrusted` is absent?

packages/rum-core/src/domain/contexts/foregroundContexts.ts
```diff
@@ -65,19 +65,13 @@ export function closeForegroundPeriod() {
...
-      return
-    }
+  return addEventListener(window, DOM_EVENT.FOCUS, () => {
     onFocusChange()
   })
 }
 
 function trackBlur(onBlurChange: () => void) {
-  return addEventListener(window, DOM_EVENT.BLUR, (event) => {
-    if (!event.isTrusted) {
```
bcaudan: **💬 suggestion:** ‏could we also remove this one? https://github.com/DataDog/browser-sdk/blob/7e0d78cd7e98f374427736ff5682def2935e2493/packages/rum/src/domain/record/observers/moveObserver.ts#L55?

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -170,6 +171,21 @@ export function basePage({ header, body }: { header?: string; body?: string }) {
...
+    ${JSON.stringify(Object.values(DOM_EVENT))}.forEach(function (eventName) {
+      window.addEventListener(
+        eventName,
+        function (event) {
+          event.__ddIsTrusted = true
+        },
+        true
+      )
+    })
+  </script>`
```
bcaudan, line 186: Nice!

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -34,7 +35,9 @@ describe('recorder', () => {
         await performSignificantZoom()
 
         await browserExecute(() => {
-          window.dispatchEvent(new Event('resize'))
+          const event = new Event('resize') as TrustableEvent
+          event.__ddIsTrusted = true
+          window.dispatchEvent(event)
```
bcaudan, line 40: **❓ question:** ‏any idea why this one need to be trusted manually?

test/e2e/lib/framework/pageSetups.ts
amortemousque: The credit goes to @BenoitZugmeyer ;)

packages/rum-core/src/domain/contexts/foregroundContexts.ts
amortemousque: I merge my branch with prerelease-v5. This code no longer exist in v5 :)

bcaudan: it is still there in [prerelease-v5](https://github.com/DataDog/browser-sdk/blob/prerelease-v5/packages/rum/src/domain/record/observers/moveObserver.ts#L55), is it planned to be removed elsewhere?

test/e2e/scenario/recorder/viewports.scenario.ts
amortemousque: Ho right. setupTrustedEvents() should be called before the header: https://github.com/DataDog/browser-sdk/blob/6562f01af1f3aa2929e7d795945fbb2eddc09eba/test/e2e/lib/framework/pageSetups.ts#L165-L166
Then no need to do `event.__ddIsTrusted = true` here

packages/rum/test/mockWorker.ts
```diff
@@ -64,41 +65,47 @@ export class MockWorker implements DeflateWorker {
       switch (message.action) {
         case 'init':
           this.listeners.message.forEach((listener) =>
-            listener({
-              data: {
-                type: 'initialized',
-              },
-            })
+            listener(
+              createNewEvent('message', {
```
BenoitZugmeyer, line 69: **❓ question:** ‏Do we need to use real events in the mock worker? types allow us to use simple structures

amortemousque: True. I felt using `createNewEvent` to centralize the `event.__ddIsTrusted = true` quite handy. Do you see an issue in having real events?

BenoitZugmeyer: Not an issue, but it is a bit unnecessary (also `__ddIsTrusted` isn't used here)

BenoitZugmeyer (approved):

packages/rum/test/mockWorker.ts
amortemousque: Well, FMU it is used because of: https://github.com/DataDog/browser-sdk/blob/53ed383dbfd28ece437db0e813d065d539ee045d/packages/rum/src/domain/segmentCollection/segment.ts#L44-L46

BenoitZugmeyer: Aah right nevermind!

packages/rum-core/src/domain/contexts/foregroundContexts.ts
amortemousque: Ho yes, I remember this one. So my reasoning is that we can still have events with `isTrusted: false `if the user uses `__ddIsTrusted` so this check seems still valid. Wdyt?

packages/core/src/browser/addEventListener.ts
```diff
@@ -2,7 +2,11 @@ import { monitor } from '../tools/monitor'
 import { getZoneJsOriginalValue } from '../tools/getZoneJsOriginalValue'
 import type { VisualViewport, VisualViewportEventMap } from './types'
 
-export const enum DOM_EVENT {
+export type TrustableEvent<E extends Event = Event> = E & { __ddIsTrusted?: boolean }
+
+export // We want to use a real enum (i.e. not a const enum) here, to be able to iterate over it to automatically add _ddIsTrusted in e2e tests
+// eslint-disable-next-line no-restricted-syntax
+enum DOM_EVENT {
```
bcaudan, line 9: comments between `export` and `enum` seem a bit odd, is it intentional?

amortemousque: I was IDE auto export :)

packages/core/src/browser/addEventListener.spec.ts
```diff
@@ -33,4 +33,37 @@ describe('addEventListener', () => {
...
     })
   })
+
+  describe('Untrusted event', () => {
+    it('should be ignored if __ddIsTrusted is absent', () => {
+      const listener = jasmine.createSpy()
+      const eventTarget = document.createElement('div')
+      addEventListener(eventTarget, DOM_EVENT.CLICK, listener)
+
+      const event = createNewEvent(DOM_EVENT.CLICK, { __ddIsTrusted: undefined })
```
bcaudan, line 43: **💬 suggestion:** ‏
```suggestion
      const event = createNewEvent(DOM_EVENT.CLICK)
```
or is it needed for some reason?

bcaudan (approved):

Issue #462, opened 2020-07-21
Allow custom tags for Browser and RUM on the initialization


rodrigopinto:
The current applications that we monitor we make extensive use of tags to be able to aggregate and filter the logs in the way the teams get easier debugging.

Running the browser log it is possible to see default tags being added based on the params of passed on the init method. As can be confirmed here https://github.com/DataDog/browser-sdk/blob/3b25ae0f92842f37be0a343dbd5d690fe973889f/packages/core/src/configuration.ts#L192-L196.

I was wondering if would be ok to have an optional parameter to the `interface UserConfiguration` to allow custom tags during the init of the browser logs and RUM.

bcaudan:
Hi @rodrigopinto 

There is ongoing work to merge the attributes and tags concepts so we are not willing to introduce this kind of API for now.
However, as you are using logs, you can configure a [remapper](https://docs.datadoghq.com/logs/processing/processors/?tab=ui#remapper) to set an env tag from the env attribute.

rodrigopinto:
Hi @bcaudan thanks for the reply. And sure, remapper would work, but just out of curiosity this ongoing work you mean internally at DD or on the lib?

bcaudan:
It is an internal reflexion at DD, if we don't carry it out, we could reassess this feature

michal-filip:
For the record, as long as this feature isn't carried out on datadog side, you can "hack" it by including the custom tags into one of the initialization options like this:
```
datadogLogs.init({
  ...,
  env: 'dev,branch:master,commit:abcd12'
})
```

Which of course is a bug in the first place, as the commas in those options should probably get escaped.

g-dx:
I too am very interested in the ability to add custom tags.  Are there are any updates to report on this?

(For reference this is to enable building a custom dashboard which allows the user to select which RUM application to display using a templated variable pointing to the custom tag, rather than using the `service` tag which overloaded with lots of things or `@application.id` which uses GUIDs)

bcaudan:
@g-dx, what about using a global context attribute instead of a tag?

g-dx:
Hey @bcaudan, I wasn’t aware of this functionality. Sounds like it fits my usecase perfectly. 

Thanks.

uluzox:
Just in case you ran into that problem yourself:

[...]

PR #2310, opened 2023-06-27
✨ Allow to pass custom tags

maksimaderikha:
## Motivation

The goal of this PR is to allow users to pass any custom tags along with the built-in ones. See the issue https://github.com/DataDog/browser-sdk/issues/462 for more details.

## Changes

This PR adds optional `customTags` property to the configuration object, so that logger or rum may be configured to send additional tags. This is completely optional configuration option that will not impact existing configurations. The changes also do not allow overwriting the built-in tags.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -47,6 +47,9 @@ export interface InitConfiguration {
   env?: string | undefined
   version?: string | undefined
 
+  // custom tags
+  customTags?: Record<string, string> | undefined
+
   // cookie options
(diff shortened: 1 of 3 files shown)


amortemousque:
Hello @maksimaderikha,
Thanks for the PR. We are still thinking about merging the attributes and tags concepts in the future so we are not willing to introduce this kind of API for now. What about using a [global context attribute](https://docs.datadoghq.com/real_user_monitoring/browser/advanced_configuration/?tab=npm#global-context) instead of a tag?

PR #2317, opened 2023-07-03
🐛 [RUMF-1436] fix crash when instrumenting readonly methods

BenoitZugmeyer:
## Motivation

In some rare cases, we observed that some methods (`window.open`, `XMLHttpRequest.prototype.send`) were defined as readonly, probably because of another thirdparty script. When instrumenting those methods, the Browser SDK crashed with the following telemetry error:

```
TypeError: Cannot assign to read only property 'open' of object '#<Window>'
```

## Changes

Make sure the method is writable before instrumenting it.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/instrumentMethod.spec.ts
+++ b/packages/core/src/tools/instrumentMethod.spec.ts
@@ -54,6 +54,27 @@ describe('instrumentMethod', () => {
     expect(instrumentationSpy).toHaveBeenCalled()
   })
 
+  it('does not throw if the object is frozen', () => {
+    const object = { method: () => 1 }
+    Object.freeze(object)
+
+    expect(() => {
+      instrumentMethod(object, 'method', () => () => 2)
+    }).not.toThrow()
+  })
+
+  it('does not throw if the method is not writable', () => {
+    const object = { method: () => 1 }
+    Object.defineProperty(object, 'method', {
+      value: () => 1,
+      writable: false,
+    })
+
+    expect(() => {
+      instrumentMethod(object, 'method', () => () => 2)
+    }).not.toThrow()
+  })
+
   describe('stop()', () => {
     it('restores the original behavior', () => {
       const object = { method: () => 1 }
--- a/packages/rum-core/src/browser/locationChangeObservable.ts
+++ b/packages/rum-core/src/browser/locationChangeObservable.ts
@@ -38,10 +38,10 @@ export function createLocationChangeObservable(location: Location) {
 }
 
 function trackHistory(onHistoryChange: () => void) {
-  const { stop: stopInstrumentingPushState } = instrumentMethodAndCallOriginal(history, 'pushState', {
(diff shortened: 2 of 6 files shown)


packages/core/src/tools/instrumentMethod.spec.ts
```diff
@@ -54,6 +54,27 @@ describe('instrumentMethod', () => {
...
+  it('does not throw if the object is frozen', () => {
+    const object = { method: () => 1 }
+    Object.freeze(object)
+
+    expect(() => {
+      instrumentMethod(object, 'method', () => () => 2)
+    }).not.toThrow()
+  })
+
+  it('does not throw if the method is not writable', () => {
```
bcaudan, line 66: **💬 suggestion:** ‏should we document the case where the method is readonly but inherited from the prototype?

packages/rum-core/src/browser/locationChangeObservable.ts
```diff
@@ -38,10 +38,10 @@ export function createLocationChangeObservable(location: Location) {
 }
 
 function trackHistory(onHistoryChange: () => void) {
-  const { stop: stopInstrumentingPushState } = instrumentMethodAndCallOriginal(history, 'pushState', {
+  const { stop: stopInstrumentingPushState } = instrumentMethodAndCallOriginal(History.prototype, 'pushState', {
```
bcaudan, line 41: **❓ question:** ‏Could we also have an issue for xhr.onreadystatechange:
https://github.com/DataDog/browser-sdk/blob/d5b4a4cb3ec0dfc44918cb6e6da031b8811fee5a/packages/core/src/browser/xhrObservable.ts#L89

BenoitZugmeyer: Ah yes, true. My approach does not work here, because:
* `onreadystatechange` is defined on the prototype, so we would need to call `instrumentMethodAndCallOriginal` with `XMLHttpRequest.prototype`
* but we need to set the `onreadystatechange` property on the instance, not the prototype, in order to be used as an event listener

Contrary to other methods like `window.open` where it could be set readonly without breaking what it does, setting an event listener property as readonly would break its usage, so in practice I don't think it will happen on purpose. Thus, I think it's fine to leave it like that. WDYT?

bcaudan: > Contrary to other methods like window.open where it could be set readonly without breaking what it does, setting an event listener property as readonly would break its usage, so in practice I don't think it will happen on purpose.

LGTM

packages/core/src/tools/instrumentMethod.spec.ts
BenoitZugmeyer: I added some documentation to the `instrumentMethod` function.

PR #2335, opened 2023-07-11
🔉 monitor reported errors

bcaudan:
## Motivation

Have an idea on how frequently we are reporting SDK errors to customers.
For now, we only report when:

- an event limiter threshold is reached
- retry strategy queue was full

## Changes

Send telemetry debug when reporting an error

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/boot/startLogs.ts
+++ b/packages/logs/src/boot/startLogs.ts
@@ -52,7 +52,7 @@ export function startLogs(
         status: StatusType.error,
       },
     })
-    addTelemetryDebug(`Error reported: ${error.message}`)
+    addTelemetryDebug('Error reported to customer', { 'error.message': error.message })
   }
   const pageExitObservable = createPageExitObservable()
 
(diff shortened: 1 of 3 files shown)


packages/logs/src/boot/startLogs.ts
```diff
@@ -51,6 +52,8 @@ export function startLogs(
         status: StatusType.error,
       },
     })
+    addTelemetryDebug('Error reported to customer', { 'error.message': error.message })
```
BenoitZugmeyer, line 55: **💬 suggestion:** ‏the three added telemetry logs are identical. Maybe add the error source in the context? It could be logs / rum / replay

BenoitZugmeyer: I don't! `service` is fine to distinghuish between RUM and Logs, but not replay. If you think we don't need to run some analytics between RUM and Replay, we can certainly rely on the message to figure out what was the source of the log.

BenoitZugmeyer (approved):

packages/logs/src/boot/startLogs.ts
bcaudan: > If you think we don't need to run some analytics between RUM and Replay, we can certainly rely on the message

That sounds fine for now, let's wait and see if we feel the need to add more context later

PR #2341, opened 2023-07-17
🐛 Avoid setting non-object values for contexts

bcaudan:
## Motivation

Telemetry errors `Cannot convert undefined or null to object` in [rum assembly](https://github.com/DataDog/browser-sdk/blob/c9ec11d/packages/rum-core/src/domain/assembly.ts#L171)

## Changes

If `setXXXContext()` is called with a non-object value, defaults context value to `{}`

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/serialisation/contextManager.ts
+++ b/packages/core/src/tools/serialisation/contextManager.ts
@@ -1,6 +1,7 @@
 import { computeBytesCount } from '../utils/byteUtils'
 import { throttle } from '../utils/functionUtils'
 import { deepClone } from '../mergeInto'
+import { getType } from '../utils/typeUtils'
 import { jsonStringify } from './jsonStringify'
 import { sanitize } from './sanitize'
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/serialisation/contextManager.ts
```diff
@@ -51,8 +52,13 @@ export function createContextManager(customerDataType: CustomerDataType, compute
...
 
     setContext: (newContext: Context) => {
-      context = sanitize(newContext)
-      computeBytesCountThrottled(context)
+      if (getType(newContext) === 'object') {
+        context = sanitize(newContext)
+        computeBytesCountThrottled(context)
+      } else {
+        context = {}
+        bytesCountCache = 0
```
amortemousque, line 60: **💬 suggestion:** Perhaps we can merge this code with `clearContext`

BenoitZugmeyer (approved):

amortemousque (approved):

PR #2344, opened 2023-07-18
🔈Add web vital attribution telemetry debug

amortemousque:
## Motivation

Assess the best strategy to display / collect the Core Web Vital attributed node.
Here are the questions that need to be answered:

With the "XXX attribution" telemetry debug:
- How often do we have the attributed DOM element?
  - Collect `hasNode`: Check if the web vital has an attributed element to collect
- Can we get the serialized node id by the time we process the web vitals
  -  Collect `replayRecording`: is Replay recording
  - Collect `serializedDomNode`: Get the serialized DOM node if available?

With the "XXX attribution recording delay" debug:
- How precise would be a Replay visualization?
  - Collect `recordingDelay`: Get the serialized DOM node if available
  - Collect `hasNode`: Do we still have the node when replay starts
  - Collect `serializedDomNode`: Can we still get the Node id when replay starts

 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan (approved):
LGTM

BenoitZugmeyer (approved):

Issue #2330, opened 2023-07-10
Module resolution error after updating to `@datadog/browser-rum` 4.44.1


derekcicerone:
**Describe the bug**
Just updated to @datadog/browser-rum 4.44.1 and we're seeing this error when running unit tests:
```
Error: Jest: Got error running globalSetup - <redacted>/jestGlobalSetup.ts, reason: Cannot find module '@datadog/browser-worker/string'
Require stack:
- /home/runner/work/wonder/wonder/node_modules/@datadog/browser-rum/cjs/domain/segmentCollection/startDeflateWorker.js
- /home/runner/work/wonder/wonder/node_modules/@datadog/browser-rum/cjs/domain/segmentCollection/index.js
- /home/runner/work/wonder/wonder/node_modules/@datadog/browser-rum/cjs/boot/startRecording.js
- /home/runner/work/wonder/wonder/node_modules/@datadog/browser-rum/cjs/entries/main.js
```

**To Reproduce**
Install the latest release of `@datadog/browser-rum`.

**Expected behavior**
No module resolution errors.

samc:
seeing the same thing with `esm` imports for any projects with a dependency graph in the range of `^4.x` for `@datadog/browser-rum`

ngbrown:
When building a Remix application (which uses esbuild), I too had encountered this problem after updating to 4.44.1:

```
✘ [ERROR] Could not resolve "@datadog/browser-worker/string"

    node_modules/@datadog/browser-rum/esm/domain/segmentCollection/startDeflateWorker.js:2:29:
      2 │ import { workerString } from '@datadog/browser-worker/string';
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Looking at the installed `@datadog/browser-worker` package in `node_modules` it makes no sense to import from `@datadog/browser-worker/string`. There is no folder or file at that location. I see that in the source in this repo, there is a `package.json` in the [`packages/worker/string` directory](https://github.com/DataDog/browser-sdk/tree/2dcd8db0712ceb688ed5157b1d73873ec5caed5b/packages/worker/string), but that isn't translating to anything in the published npm package.

It looks like these changes were introduced by #2319.

cc: @BenoitZugmeyer

JuanGuerreroMp:
Same here in a react app: `

./node_modules/@datadog/browser-rum/esm/domain/segmentCollection/startDeflateWorker.js
--
[...]

PR #2351, opened 2023-07-24
👷 [RUM-245] check package files

BenoitZugmeyer:
## Motivation

Following an architectural change, we recently broke the RUM NPM package because of a misconfiguration (see https://github.com/DataDog/browser-sdk/issues/2330)

## Changes

To avoid this in the future, this PR introduce a new script that ensures all expected files are correctly set in NPM packages.

* Files actually included in the package are listed via `npm pack`.
* Files expected to be included in the package are listed via a set of rules
* Those two lists are compared and the script fails if they differs.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


scripts/check-packages.js
```diff
@@ -0,0 +1,142 @@
...
+const path = require('path')
+const glob = require('glob')
+const { printLog, printError, runMain } = require('./lib/execution-utils')
+const { command } = require('./lib/command')
+
+const PACKAGE_OPTIONS = {
+  'packages/worker': {
+    builtWithTypeScript: false,
+    extraFiles: [
+      // This file is not built with TypeScript and not listed by package.json entry points, so we
```
bcaudan, line 11: **💬 suggestion:** I have struggled a bit to understand why only this file was needed 😅‏
```suggestion
      // This file is not built with TypeScript and not listed by string/package.json entry points, so we
```

bcaudan (approved):

scripts/check-packages.js
```diff
@@ -0,0 +1,142 @@
...
+  // Yarn behavior is a bit different from npm regarding `.npmignore` globs. Since we are publishing
+  // packages using npm through Lerna[1], let's use npm to list files here.
+  //
+  // [1]: Quoting Lerna doc: "Lerna always uses npm to publish packages."
+  // https://lerna.js.org/docs/features/version-and-publish#from-package
+  const output = command`npm pack --dry-run --json`.withCurrentWorkingDirectory(packagePath).run()
+  return JSON.parse(output)[0].files.map((file) => file.path)
+}
+
+function getExpectedPackageFiles(packagePath) {
```
amortemousque, line 82: **❓ question:** ‏Do you think we could rely on the [files](https://docs.npm™js.com/cli/v9/configuring-npm/package-json#files) npm configuration option for the expected file list to check?

scripts/lib/command.js
```diff
@@ -33,10 +34,16 @@ function command(...templateArguments) {
       return this
     },
 
+    withCurrentWorkingDirectory(newCurrentWorkingDirectory) {
+      currentWorkingDirectory = newCurrentWorkingDirectory
+      return this
+    },
```
bcaudan, line 40: 👍

bcaudan (approved):
Nice!

scripts/check-packages.js
BenoitZugmeyer: For the record: I didn't go with the `files` configuration, which is the same as `.npmignore` but reversed. This could be done in the future if we feel the need for it, but does not help this specific PR. But following your advice and after discussions, we decided to implement rules based on `.npmignore` instead of having to list every expected files.

scripts/check-packages.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+  return true
+}
+
+// NPM will [always include some files][1] like `package.json` and `README.md`.
+// [1]: https://docs.npmjs.com/cli/v9/using-npm/developers#keeping-files-out-of-your-package
+const FILES_ALWAYS_INCLUDED_BY_NPM = ['package.json', 'README.md']
+
+function checkNpmIgnore(packagePath, packageFiles) {
```
amortemousque, line 61: Nice 👏 ‏

PR #2353, opened 2023-07-25
👷Use SDK version format supported by backend

bcaudan:
## Motivation

In canary environment, the SDK version (`X.Y.Z+hash`) is not well supported by the backend since:

- [tags don't support](https://docs.datadoghq.com/getting_started/tagging/#define-tags) the `+` character and they are replace by `_`
- version are parsed using [semver regex](https://regex101.com/r/Ly7O1x/3/) that don't support the `_` character

## Changes

In canary environment, send the hash with the version using the prerelease separator instead of the build separator:

```diff
- X.Y.Z+hash
+ X.Y.Z-hash
```

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/lib/build-env.js
+++ b/scripts/lib/build-env.js
@@ -34,7 +34,9 @@ switch (buildMode) {
     break
   case 'canary': {
     const commitSha1 = execSync('git rev-parse HEAD').toString().trim()
-    sdkVersion = `${lernaJson.version}+${commitSha1}`
+    // TODO when tags would allow '+' characters
+    //  use build separator (+) instead of prerelease separator (-)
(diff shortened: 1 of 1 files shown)


BenoitZugmeyer:
> In canary environment, send the hash with the version using the prerelease separator instead of the build separator:
> 
> ```diff
> - X.Y.Z+hash
> + X.Y.Z-hash
> ```

Note: for the prerelease-v5, the version would look like:

```
5.0.0-alpha.0-5bc99408ea1dffffedaf9e13facb23f14c1f0c0f
```

Which is fine, I guess.

BenoitZugmeyer (approved):

PR #2354, opened 2023-07-25
✨ [RUM-233] add workerUrl initialization parameter 

BenoitZugmeyer:
## Motivation

The RUM Browser SDK is using a Web Worker. To facilitate distribution, this worker is built with a "blob" URL, thus requiring the CSP `worker-src blob:;` directive, which might be seen as a security risk.

## Changes

With the new `workerUrl` initialization parameters, users will be able to self-host the worker JavaScript file to workaround CSP restrictions.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


bcaudan (commented):
By playing a bit with it, I think we should be improve what we communicate to customers on error cases. 
Currently:
- for a 404 on the worker URL, we still display the error message related to CSP
- for a worker url that point to a wrong js file, there is no error at all

BenoitZugmeyer:
> By playing a bit with it, I think we should be improve what we communicate to customers on error cases. Currently:
> 
>     * for a 404 on the worker URL, we still display the error message related to CSP

There is no distinctinction between a 404 and a CSP error. We could reword the error message a bit, like:
>  "This could be caused by a CSP directive, see documentation https://..."

> 
>     * for a worker url that point to a wrong js file, there is no error at all

The only thing I can think of is displaying an error in case the worker does not answer with a `initialized` message within a given delay (10 seconds could be fine?).

bcaudan:
> There is no distinctinction between a 404 and a CSP error. We could reword the error message a bit, like:
> 
> > "This could be caused by a CSP directive, see documentation https://..."

When using `workerUrl` could we still have CSP restrictions?
If not, we could remove the CSP mention to avoid to point customers in a wrong direction.

> The only thing I can think of is displaying an error in case the worker does not answer with a `initialized` message within a given delay (10 seconds could be fine?).
> 
> WDYT?

That would be helpful, yes.
No alternative in mind.

BenoitZugmeyer:
> > There is no distinctinction between a 404 and a CSP error. We could reword the error message a bit, like:
> > > "This could be caused by a CSP directive, see documentation https://..."
> 
> When using `workerUrl` could we still have CSP restrictions? If not, we could remove the CSP mention to avoid to point customers in a wrong direction.

Yes the worker script could still be clocked by CSP. I'll adjust the message, you'll let me know

bcaudan (approved):

PR #2355, opened 2023-07-25
⚗️ [RUM-257] Collect INP

amortemousque:
## Motivation

[Interaction to Next Paint (INP)](https://web.dev/inp/) is a Core Web Vital metric that will [replace First Input Delay (FID)](https://web.dev/inp-cwv/) in March 2024. INP assesses a page's overall responsiveness to user interactions by observing the latency of all clicks, taps, and keyboard interactions that occur throughout the lifespan of a user's visit to a page

## Changes

Collect INP under `view.interaction_to_next_paint`

## Testing

- [X] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts
```diff
@@ -93,11 +95,17 @@ export function trackViewMetrics(
     stopCLSTracking = noop
   }
 
+  const { stop: stopINPTracking } = trackInteractionToNextPaint(loadingType, lifeCycle, (interactionToNextPaint) => {
+    viewMetrics.interactionToNextPaint = interactionToNextPaint
+    scheduleViewUpdate()
```
amortemousque, line 100: I will remove it according to the RFC

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,142 @@
...
+  let maxKnownInteractionId = 0
+
+  const observer = new (window as BrowserWindow).PerformanceObserver((entries: PerformanceObserverEntryList) => {
+    entries.getEntries().forEach((e) => {
+      const entry = e as unknown as RumEventTiming
+      if (entry.interactionId) {
+        minKnownInteractionId = Math.min(minKnownInteractionId, entry.interactionId)
+        maxKnownInteractionId = Math.max(maxKnownInteractionId, entry.interactionId)
+
+        interactionCount = maxKnownInteractionId ? (maxKnownInteractionId - minKnownInteractionId) / 7 + 1 : 0
```
bcaudan, line 130: **❓ question:** ‏what does `7` represent?

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+    })
+  })
+  observer.observe({ type: 'event', buffered: viewLoadingType === ViewLoadingType.INITIAL_LOAD, durationThreshold: 0 })
```
bcaudan, line 134: **❓ question:** ‏why `durationThreshold: 0`? 
from [mdn](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe#syntax), the min should be 16 🤔

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,142 @@
...
+) {
+  const { getInteractionCount, stop: stopInteractionCountTracking } = trackInteractionCount(viewLoadingType)
+  let maxInpDuration = 0 as Duration
+
+  const { unsubscribe: stop } = lifeCycle.subscribe(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, (entries) => {
+    for (const entry of entries) {
+      if (
+        (entry.entryType === 'event' && entry.interactionId) ||
+        // Entries of type `first-input` don't currently have an `interactionId`,
+        // so to consider them in INP we have to first check that an existing
```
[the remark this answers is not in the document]
amortemousque: Hum, good question. FMU, first-input is reported as  performance event entry if it is above durationThreshold. Since the durationThreshold is at most 16ms, if the first input is below that value it won't be reported as a performance event entry.
It's true that in our case the durationThreshold is 40 so this scenario will never happen. I guess Google does this check because in their repo the [durationThreshold is an option](https://github.com/GoogleChrome/web-vitals/blob/3806160ffbc93c3c4abf210a167b81228172b31c/src/onINP.ts#L209C32-L209C49) the developer can change which could go bellow 16.

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -121,15 +143,20 @@ export function startPerformanceCollection(lifeCycle: LifeCycle, configuration:
...
       // to get values that could happen before SDK init
       experimentalEntries.forEach((type) => {
-        const observer = new PerformanceObserver(handlePerformanceEntryList)
-        observer.observe({ type, buffered: true })
+        const observer = new (window as BrowserWindow).PerformanceObserver(handlePerformanceEntryList)
+        observer.observe({
+          type,
+          buffered: true,
+          // durationThreshold only impact PerformanceEventTiming entries. INP computation requires threshold at 40 (default is 104ms)
+          durationThreshold: 40,
```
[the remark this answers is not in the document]
amortemousque: Updated

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque: Yes a durationThreshold is capped to 16 but like setTimeout(.., 0) is capped to 4ms I think setting 0 better conveys the idea that we want all the ente entries.

amortemousque: In any case I isolated the interaction count polyfill from Google from the rest of the logic to ease the reading :)

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  // List of longest interactions on the view by duration.
+  const longestInteractions: Array<RumEventTiming | RumFirstInputTiming> = []
+
+  const { getViewInteractionCount } = trackViewInteractionCount(viewLoadingType)
+  let maxInpDuration = -1 as Duration
+
+  const { unsubscribe: stop } = lifeCycle.subscribe(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, (entries) => {
+    for (const entry of entries) {
+      if (entry.entryType === 'event' && entry.interactionId) {
```
amortemousque, line 39: Compared to Goole implementation: No need to track 'first-input' event because our duration threshold is at 40 and so first-input above that duration will also be reported as 'event' entry.

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque, line 1: Compared to Google: This implementation intentionally does not handle some scenarios because they should be considered more broadly for every core web vitals:
- BF Cache restore
- Flushing observer recors at page hidden`PerformanceObserver.takeRecords()` in performance collection
- prerendering

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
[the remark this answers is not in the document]
amortemousque: I've removed this logic since it's not necessary for us

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque: From the [original PR from Google](https://github.com/GoogleChrome/web-vitals/pull/213): "interactionCount is an estimate based on a convention specific to Chrome". I found no more information

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,114 @@
...
+    !isExperimentalFeatureEnabled(ExperimentalFeature.INTERACTION_TO_NEXT_PAINT)
+  ) {
+    return {
+      getInteractionToNextPaint: () => undefined,
+      stop: noop,
+    }
+  }
+
+  // List of longest interactions on the view by duration.
+  const longestInteractions: Array<RumEventTiming | RumFirstInputTiming> = []
```
bcaudan, line 32: **💬 suggestion:** ‏To decouple longest interactions management from the rest of the logic, what about extracting a dedicated method for it?

something like:
```ts
const longestInteractions = trackLongestInteractions(getViewInteractionCount)
longestInteractions.process(event)
longestInteractions.estimateP98Duration()
```

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,114 @@
...
+    const inp = estimateP98LongestInteraction(getViewInteractionCount)
+    if (inp && inp.duration > maxInpDuration) {
+      maxInpDuration = inp.duration
+    }
+  })
+
+  /**
+   * Process the performance entry:
+   * - if its duration is long enough, add the performance entry to the list of worst interactions
+   * - if an entry with the same interaction id exists and but its duration is lower than the new one, then replace it in the list of worst interactions
```
bcaudan, line 53: **💬 suggestion:** ‏or but?
```suggestion
   * - if an entry with the same interaction id exists and its duration is lower than the new one, then replace it in the list of worst interactions
```

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  return {
+    getInteractionToNextPaint: () => {
+      // If the interaction count shows there were interactions but
+      // none were captured by the PerformanceObserver because of the threshold, report a latency of 0.
+      if (maxInpDuration === -1 && getViewInteractionCount()) {
+        return 0 as Duration
+      } else if (maxInpDuration >= 0) {
+        return maxInpDuration
+      }
```
bcaudan, line 97: **💬 suggestion:** ‏we could be more explicit by:
- init maxInpDuration at undefined
- have a `onFirstInteraction` callback and in it set maxInpDuration to 0 if it is still undefined  
- always return the maxInpDuration in this function

packages/rum-core/src/domain/rumEventsCollection/view/interactionCountPolyfill.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * interactionCount polyfill
+ * Reference implementation: https://github.com/GoogleChrome/web-vitals/blob/main/src/lib/polyfills/interactionCountPolyfill.ts
```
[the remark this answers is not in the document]
amortemousque: I added more context. Let me know what you think.

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque: Good idea, done!

packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts
```diff
@@ -105,14 +107,20 @@ export function trackViewMetrics(
...
       stopCLSTracking()
       stopScrollMetricsTracking()
+      stopINPTracking()
     },
     setLoadEvent,
-    viewMetrics,
+    getViewMetrics: () => {
+      viewMetrics.interactionToNextPaint = getInteractionToNextPaint()
+      return viewMetrics
+    },
```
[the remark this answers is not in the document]
amortemousque: `getInteractionToNextPaint` does not really recompute the INP. INP is computed each time a [PerformanceEventTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEventTiming) is collected.

`getInteractionToNextPaint` only serve for the case no INP has been computed but performance.interactionCount > 0, meaning that some interaction did happen. It can occur because of the performanceObserver 40ms threshold. 
In this scenario we need to report an INP at 0 instead of no INP for the page.

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque: Creating a `onFirstInteraction` based on PerformanceEventTiming could miss some interactions due to the threshold (16ms minimum). This is why I rely on performance.interactionCount to check if an interaction occur

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,113 @@
...
+    for (const entry of entries) {
+      if (entry.entryType === 'event' && entry.interactionId) {
+        longestInteractions.process(entry)
+      }
+    }
+
+    const inp = longestInteractions.estimateP98Duration()
+    if (inp && inp.duration > maxInpDuration) {
+      maxInpDuration = inp.duration
+    }
```
BenoitZugmeyer, line 41: **💭 thought:** ‏I'm not questioning your implementation, but using a max of a p98 feels weird to me. I believe using a percentile here helps to get rid of outliers . Imagine the first interaction you have is an outlier and its value is abnormally high, `maxInpDuration` will be set to this value. Then, we have a thousand more interaction with a smaller duration. `maxInpDuration` will still represent the first outlier, defeating the purpose of using a percentile.

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  return {
+    getInteractionToNextPaint: () => {
+      // If no INP duration where captured because of the performanceObserver 40ms threshold
+      // but the view interaction count > 0 then report 0
+      if (maxInpDuration === -1 && getViewInteractionCount()) {
+        return 0 as Duration
+      } else if (maxInpDuration >= 0) {
+        return maxInpDuration
+      }
```
BenoitZugmeyer, line 52: **💬 suggestion:** ‏could be simplified a bit:
```suggestion
      if (maxInpDuration >= 0) {
        return maxInpDuration
      } else if (getViewInteractionCount()) {
        return 0 as Duration
      }
```

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
BenoitZugmeyer: Also the reference implementation seems to have a special case for `first-input` handling: https://github.com/GoogleChrome/web-vitals/blob/a948d3cde1770eabb8a215072cfd9713e940c4d9/src/onINP.ts#L169-L191

packages/rum-core/src/domain/rumEventsCollection/view/interactionCountPolyfill.ts
```diff
@@ -0,0 +1,49 @@
...
+  observer = new (window as BrowserWindow).PerformanceObserver(
+    monitor((entries: PerformanceObserverEntryList) => {
+      entries.getEntries().forEach((e) => {
+        const entry = e as unknown as RumEventTiming
+
+        if (entry.interactionId) {
+          minKnownInteractionId = Math.min(minKnownInteractionId, entry.interactionId)
+          maxKnownInteractionId = Math.max(maxKnownInteractionId, entry.interactionId)
+
+          interactionCountEstimate = maxKnownInteractionId ? (maxKnownInteractionId - minKnownInteractionId) / 7 + 1 : 0
```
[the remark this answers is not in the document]
amortemousque: About the `/ 7 + 1`, from the original Google [PR](https://github.com/GoogleChrome/web-vitals/pull/213) description, we don't have much information besides `interactionCount is an estimate based on a convention specific to Chrome`

Good suggestion :)

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -126,15 +145,21 @@ export function startPerformanceCollection(lifeCycle: LifeCycle, configuration:
...
-        const observer = new PerformanceObserver(handlePerformanceEntryList)
-        observer.observe({ type, buffered: true })
+        const observer = new (window as BrowserWindow).PerformanceObserver(handlePerformanceEntryList)
+        observer.observe({
+          type,
+          buffered: true,
+          // durationThreshold only impact PerformanceEventTiming entries used for INP computation which requires a threshold at 40 (default is 104ms)
+          // cf: https://github.com/GoogleChrome/web-vitals/blob/3806160ffbc93c3c4abf210a167b81228172b31c/src/onINP.ts#L209
+          durationThreshold: 40,
+        })
```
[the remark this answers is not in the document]
amortemousque: I agree on your analyses. The design could clearly be improved I'm gonna track it on Jira.

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque: Yes, we discussed about it with Bastien and removed this logic as we should not be concerned with our implementation. cf https://github.com/DataDog/browser-sdk/pull/2355#discussion_r1274864088

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,113 @@
...
+    },
+  }
+}
+
+export function trackViewInteractionCount(viewLoadingType: ViewLoadingType) {
+  initInteractionCountPolyfill()
+  const previousInteractionCount = viewLoadingType === ViewLoadingType.INITIAL_LOAD ? 0 : getInteractionCount()
+  return {
+    getViewInteractionCount: () => getInteractionCount()! - previousInteractionCount,
+    stop: noop,
```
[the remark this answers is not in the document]
amortemousque: Purely subjective but I prefer to have it here instead of checking it exists in [trackViewMetrics](https://github.com/DataDog/browser-sdk/blob/e4a0329e97bf32d2a25df214bdda30b4fba9a071/packages/rum-core/src/domain/rumEventsCollection/view/trackViewMetrics.ts#L114). Wdyt?

packages/rum-core/src/domain/rumEventsCollection/view/interactionCountPolyfill.ts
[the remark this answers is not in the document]
BenoitZugmeyer: `specific to Chrome` that's not great 😅 For now, this "polyfill" (and the whole INP collection) will only work on Chrome, but if/when other browsers support `interactionId`, this could be an issue. Hopefully they'll implement `performance.interactionCount` at the same time, so we won't need that polyfill.

Could you add a code comment to that PR to justify the approach here?

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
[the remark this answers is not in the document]
BenoitZugmeyer: I'm talking about the `trackViewInteractionCount` `stop` method. I don't think it's used anywhere?

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
BenoitZugmeyer: Ah I wasn't really suggesting to change the implementation, sorry :D I'm not sure what's the right call here, we might want to stick with Google approach.

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Mmh `first-input` entries are not `event` entries (that's two different types). The reference implementation [observes `first-input` separately](https://github.com/GoogleChrome/web-vitals/blob/main/src/onINP.ts#L224), so I guess it is collected regardless of the duration threshold?

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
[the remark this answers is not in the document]
amortemousque: true

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
amortemousque: No it's not the Google approach it's a wrong implementation on my side while trying to make INP for every view

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.ts
```diff
@@ -0,0 +1,112 @@
...
+        (interaction) => entry.interactionId === interaction.interactionId
+      )
+
+      const minLongestInteraction = longestInteractions[longestInteractions.length - 1]
+
+      if (interactionIndex !== -1) {
+        if (entry.duration > longestInteractions[interactionIndex].duration) {
+          longestInteractions[interactionIndex] = entry
+          sortAndTrimLongestInteractions()
+        }
```
bcaudan, line 82: **💬 suggestion:** ‏cf codecov, what about adding a test for this code path?

packages/rum-core/src/domain/rumEventsCollection/view/trackInteractionToNextPaint.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+          duration: index as Duration,
+          interactionId: index,
+        })
+      }
+      expect(getInteractionToNextPaint()).toEqual(98 as Duration)
+    })
+
+    it('should return 0 if no interaction is tracked (because the duration is below 40ms)', () => {
+      setupBuilder.build()
+      interactionCountStub.setInteractionCount(1 as Duration) // assumes an interaction happened but no PERFORMANCE_ENTRIES_COLLECTED have been triggered
```
bcaudan, line 92: **💬 suggestion:** improve test description to avoid the comment‏
```suggestion
    it('should return 0 when an interaction happened without generating a performance event (interaction duration below 40ms)', () => {
      setupBuilder.build()
      interactionCountStub.setInteractionCount(1 as Duration)
```

packages/rum-core/src/domain/rumEventsCollection/view/interactionCountPolyfill.ts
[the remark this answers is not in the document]
amortemousque: I added more detail in the comment section https://github.com/DataDog/browser-sdk/blob/85e1a329850922ef644b1ace9ff959edc546db09/packages/rum-core/src/domain/rumEventsCollection/view/interactionCountPolyfill.ts#L7-L9

packages/rum-core/src/domain/rumEventsCollection/view/interactionCountPolyfill.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * performance.interactionCount polyfill
+ *
+ * The interactionCount is an integer which counts the total number of distinct user interactions,
+ * for which there was a unique interactionId.
+ *
+ * The interactionCount polyfill is an estimate based on a convention specific to Chrome. Cf: https://github.com/GoogleChrome/web-vitals/pull/213
+ * This is currently no an issue as the polyfill is only used for INP which is currently only supported on Chrome.
```
bcaudan, line 8: **🥜 nitpick:** ‏typo
```suggestion
 * This is currently not an issue as the polyfill is only used for INP which is currently only supported on Chrome.
```

bcaudan (approved):

BenoitZugmeyer (approved):

PR #2369, opened 2023-08-08
✨ [extension] filter outdated events

BenoitZugmeyer:
## Motivation

Showing each View update is confusing as it can be seen as separate view events.

## Changes

* Add a checkbox to show only the last view update
* Make sure View updates are collocated when many updates are shown

### Before:

Screenshot 2023-08-24 at 12 05 15

### After:

Outdated view updates are hidden

Screenshot 2023-08-24 at 12 06 23

View updates are collocated

Screenshot 2023-08-24 at 12 06 28

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2370, opened 2023-08-08
✨ [extension] revamp filter UI

BenoitZugmeyer:
## Motivation

The filter UI is quite limited. Typing a filter query works but it is not quite user friendly.

## Changes

Implement a "facet list" similar to the datadog app. The facet list counters is not well implemented, meaning it always show the number of matching events and doesn't update when other facets are hidden, but it's a first step!

Also in this PR I change fonts a bit. I chose to use the same font size as the font in the Chrome devtools, all across the developer extension. This helps to fit more things in the UI.

### After

Screenshot 2023-08-24 at 12 13 47

Note: there is no space between the facet list and the events list! This is fine, a next PR will revamp the event list and the space will be there again.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


BenoitZugmeyer:
* main
  * **PR #2369** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2369" target="_blank">Graphite</a> 
    * **PR #2370** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2370" target="_blank">Graphite</a>  👈
      * **PR #2371** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2371" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2370?utm_source=stack-comment).

developer-extension/src/panel/facets.constants.ts
```diff
@@ -0,0 +1,70 @@
+// The facet id is either:
+//
+// * "$eventSource", in which case it reflects the event source (rum, logs...) and not an actual
+//   field
+//
+// * "$EVENT_SOURCE.EVENT_PATH", where:
+//   * EVENT_SOURCE is an event source (rum, logs, ...)
+//   * EVENT_PATH is the path of the event field.
+export type FacetId = string
```
amortemousque, line 9: **💭 thought:** ‏Instead of crafting an FacetId, I'm wondering if we could use the Facet object ref in the different data structure the ID is used. This way we could have a source and path fields and avoid the regex to find them 
```js
export interface Facet {
  values?: { [value: FacetValue]: { label?: string } }
  label?: string
  parent?: string
  source: 'rum' | 'log'
  path: string
}
```
I don't how we could handle parent though :)

developer-extension/src/panel/hooks/useEvents/facetRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  clear() {
+    this.facetValueCounts = new Map(FACETS.map((facet) => [facet.id, new Map()]))
+  }
+}
+
+export function getAllFields(event: object) {
+  const fields = new Map<FieldPath, FieldMultiValue>()
+
+  getAllFieldsRec(event, undefined)
```
amortemousque, line 61: **💬 suggestion:** ‏The name confused me at first. It could be nice to use the full word like `getAllFieldsRecursively`

developer-extension/src/panel/hooks/useEvents/facetRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  return fields
+
+  function getAllFieldsRec(value: unknown, path: string | undefined) {
```
amortemousque, line 65: **💬 suggestion:** The logic is a bit hard to follow. It might be worth splitting or commenting it‏

developer-extension/src/panel/components/tabs/eventsTab/facetList.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const isSelected = !excludedFacetValues[facet.id] || !excludedFacetValues[facet.id].includes(facetValue)
+  const value = (
+    <Flex justify="space-between" mt={isTopLevel ? 'xs' : SPACE_BETWEEN_CHECKBOX}>
+      <Checkbox
+        label={facet.values?.[facetValue]?.label ?? facetValue}
+        checked={isSelected}
+        onChange={() => {
+          onExcludedFacetValuesChange(toggleExcludedFacetValue(facet, excludedFacetValues, facetValue))
+        }}
+      />
```
bcaudan, line 96: **💭 thought:** ‏Would it be complicated to have the same behaviour than the datadog facet list with "toggle" or "only" depending on if we interact with the checkbox or the label?

developer-extension/src/panel/components/tabs/eventsTab/facetList.tsx
```diff
@@ -0,0 +1,170 @@
...
+      <Text>{facetValueCount}</Text>
+    </Flex>
+  )
+
+  const childFacets = FACETS.filter((otherFacet) => otherFacet.parent === `${facet.id}:${facetValue}`)
+  const children = childFacets.length > 0 && (
+    <Collapse in={isSelected}>
+      <Box
+        ml={isTopLevel ? undefined : 9}
+        pl={isTopLevel ? undefined : 9}
```
bcaudan, line 106: **❓ question:** ‏why 9?

developer-extension/src/panel/facets.constants.ts
BenoitZugmeyer: I removed the `FacetId` concept, but kept the `"$eventSource"` special path so we can declare and handle all facets the same way.

developer-extension/src/panel/components/tabs/eventsTab/facetList.tsx
BenoitZugmeyer: It depends. The particularity here is that we have dynamic facet values, so we don't know the values in advance (this is why I found it easier to store a list of facet values to exclude, instead of having a list of facet value to keep).

We could have a "only" button that exclude all facet existing values, but then some new values can come and it wouldn't be an "only" anymore. This could be fairly easy to implement, and bring enough value for our needs.

An better alternative could be to have a more involved filter strategy that selects only a given value even when other values are coming. It would be a bit more complex to implement.

In any case, it will be part of a future PR :)

developer-extension/src/panel/hooks/useEvents/facetRegistry.ts
BenoitZugmeyer: I added examples so we don't need to read the tests, split it a bit, and added some doc. Let me know if it's clearer.

amortemousque (approved):

BenoitZugmeyer:

* main:
  * **PR #2369**
    * **PR #2370** 👈
      * **PR #2371**
        * **PR #2372**

bcaudan (approved):

PR #2371, opened 2023-08-08
✨ [extension] revamp event description and JSON viz

BenoitZugmeyer:
## Motivation

The current JSON rendering component that we use does not allow to display a proper event description. Also it is a bit limited for future extensions.

## Changes

Implement a new Json component and a better event description similar to the Datadog app

### Before

Screenshot 2023-08-24 at 12 23 11

### After

Screenshot 2023-08-24 at 12 21 50

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2372, opened 2023-08-08
✨ [extension] add columns to the event list

BenoitZugmeyer:
## Motivation

When adding a new field or investigating a specific field value, it is a bit annoying to expand it
on every event.

## Changes

* Use a grid instead of a table for event list (this was not strictly needed but I find it more predictable even if it has a few issues)
* Add support for columns in the event list to display some arbitrary fields.
  * Add columns by clicking on the json fields or by using the "Add column" button
  * Reorder columns by drag and dropping
  * Remove columns by dragging them out of the header

Screenshot 2023-08-24 at 12 25 21

Note: see, the space between the facet list and the event list is back!

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/panel/components/tabs/eventsTab/eventsList.tsx
```diff
@@ -1,19 +1,213 @@
...
+        onColumnsChange(columns.concat(column))
+      }
+    },
+    [columns]
+  )
 
-export function EventsList({ events, filters }: { events: SdkEvent[]; filters: EventFilters }) {
   return (
-    <Table striped verticalSpacing="xs">
-      <tbody>
```
amortemousque, line 11: **❓ question:** ‏Why move from a table to a grid?

bcaudan:
After playing with the UI:

🐛 You can "add column" on an already added column which create a new column with empty values

Screenshot 2023-08-25 at 14 04 59
Screenshot 2023-08-25 at 14 05 09

🐛 the display does not seem to handle well long URL in the message, maybe truncating them?

Screenshot 2023-08-25 at 14 13 38


💬 for removing a column, drag and drop outside of the header seems quite unusual.
What about adding a dropdown on column header with remove column (move left / right could also be added)?  

💬 to indicate the columns can be reorderd by drag and drop, what about displaying "move" cursor on hover?

BenoitZugmeyer:

* main:
  * **PR #2369**
    * **PR #2370**
      * **PR #2371**
        * **PR #2372** 👈

developer-extension/src/panel/components/tabs/eventsTab/eventsList.tsx
BenoitZugmeyer: I reverted back to a table, the grid offers more control over the exact size of columns, but was too much trouble to implement. I want to revisit this one day when I have time :)

BenoitZugmeyer:
> After playing with the UI:
> 
> 🐛 You can "add column" on an already added column which create a new column with empty values
> 

Fixed

> 
> 🐛 the display does not seem to handle well long URL in the message, maybe truncating them?
 
Fixed by reverting to using a Table, so now long strings should wrap properly

> 
> 💬 for removing a column, drag and drop outside of the header seems quite unusual. What about adding a dropdown on column header with remove column (move left / right could also be added)?

Added a button to remove columns.

> 
> 💬 to indicate the columns can be reorderd by drag and drop, what about displaying "move" cursor on hover?

Devtools tabs can be reordered by drag and drop, but don't have a "move" cursor. I think it could be a good idea to follow the same UI, so I didn't change the cursor. I will revisit the drag behavior later to make it closer to the native devtools experience.

bcaudan (approved):

amortemousque (approved):

PR #2377, opened 2023-08-11
✨ [RUM-252] optimistic worker creation

BenoitZugmeyer:
## Motivation

The Deflate Worker initialization is asynchronous: we instantiate the Worker, we send a first
message and wait for a response to make sure the communication works.

The recorder has been designed to wait until this initialization succeeds before actually starting
to record. This has one major benefit for Replay: no `has_replay: true` flag is set in case the
Worker initialization fails. If it was set, a link leading to an unknown replay would be displayed
in the Datadog App.

Because we want to use the Deflate Worker in other places, this asynchronous behavior is a bit
inconvenient as it forces the SDK initialization to be asynchronous too, for little benefit because
the worker initialization should succeeds in most case and we don't have the Replay constraint.

## Changes

- Make the worker start "optimistic", so the deflate Worker instance is returned synchronously.

- Make sure the `has_replay` (and replay stats) is present only if the worker was correctly
  initialized.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/test/unit/karma.bs.conf.js
+++ b/test/unit/karma.bs.conf.js
@@ -1,5 +1,5 @@
 const { browserConfigurations } = require('../browsers.conf')
-const { getBuildInfos, getIp } = require('../envUtils')
+const { getBuildInfos } = require('../envUtils')
 const karmaBaseConf = require('./karma.base.conf')
 
 module.exports = function (config) {
@@ -10,7 +10,6 @@ module.exports = function (config) {
     browsers: browserConfigurations.map((configuration) => configuration.sessionName),
     concurrency: 5,
     browserDisconnectTolerance: 3,
-    hostname: getIp(),
     browserStack: {
       username: process.env.BS_USERNAME,
       accessKey: process.env.BS_ACCESS_KEY,


BenoitZugmeyer:
* main
  * **PR #2376** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2376" target="_blank">Graphite</a> 
    * **PR #2377** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2377" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2377?utm_source=stack-comment).

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -116,25 +106,22 @@ describe('makeRecorderApi', () => {
       expect(startRecordingSpy).not.toHaveBeenCalled()
     })
 
-    it('do not start recording if worker fails to be instantiated', () => {
+    it('do not start recording if worker creation fails', () => {
```
yannickadam, line 109:
```suggestion
    it('does not start recording if worker creation fails', () => {
```

yannickadam (approved):

bcaudan (approved):

amortemousque (approved):

PR #2378, opened 2023-08-11
✨[RUM-265] Store contexts across pages

bcaudan:
## Motivation

Customer contexts ([global](https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm#global-context), [user](https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm#user-session)) are stored in the current page memory so:
- they are not kept after a full reload of the page
- they are not shared across different tabs / windows of the same session

## Changes

Introduce **new configuration parameter** for RUM and logs:

`storeContextsAcrossPages`
Optional
**Type**: Boolean
**Default**: false

When enabled, **global context** and **user context** are stored in local storage:

Screenshot 2023-08-23 at 17 26 57

**Behaviors:**

- Context is preserved after a full reload
- Context is synchronized between tabs opened on the same origin
- Context does not expire with the session 

**Limitations:**
- Setting PII in those contexts is not recommended: data stored in local storage outlives the user navigation 
- Does not support trackSessionAcrossSubdomains options, the contexts will only be shared among the same origin (login.site.com ≠ app.site.com)
- Subject to the local storage 5M limit for the whole application

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/tools/serialisation/storedContextManager.ts
```diff
@@ -0,0 +1,89 @@
...
+const STORAGE_POLL_DELAY = ONE_SECOND
+export function createStoredContextManager(
+  customerDataType: CustomerDataType,
+  computeBytesCountImpl = computeBytesCount
+): ContextManager {
+  const contextManager = createContextManager(customerDataType, computeBytesCountImpl)
+
+  const synchronizeWithStorage = () => contextManager.setContext(read(customerDataType))
+  const dumpToStorage = () => store(customerDataType, contextManager.getContext())
+  const { throttled: synchronizeWithStorageThrottled } = throttle(synchronizeWithStorage, STORAGE_POLL_DELAY, {
```
BenoitZugmeyer, line 19: **💬 suggestion:** ‏You could use the `storage` event instead of polling and reading synchronously the value from the storage.

I think it would be much easier: read the context from the store at init, then update the context every time we receive a `storage` event.

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -87,6 +90,16 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
         return
       }
 
+      if (initConfiguration.storeContextsAcrossPages) {
+        // Note: context API calls before init are dismissed
```
BenoitZugmeyer, line 94: **❓ question:** ‏Why do we dismiss earlier API calls? Could you add a bit more context?

packages/core/src/domain/context/storedContextManager.ts
```diff
@@ -0,0 +1,87 @@
...
+  const contextManager = createContextManager(customerDataType, computeBytesCountImpl)
+
+  synchronizeWithStorage()
+  addEventListener(configuration, window, DOM_EVENT.STORAGE, ({ key }) => {
+    if (storageKey === key) {
+      synchronizeWithStorage()
+    }
+  })
+
+  return {
```
BenoitZugmeyer, line 27: **💬 suggestion:** ‏Instead of implementing every methods, it could be simpler to provide a "onChange" callback to the base `createContextManager` (or return an observer)

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -100,22 +113,22 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
     }),
 
     /** @deprecated: use getGlobalContext instead */
-    getLoggerGlobalContext: monitor(globalContextManager.get),
-    getGlobalContext: monitor(globalContextManager.getContext),
+    getLoggerGlobalContext: monitor(() => globalContextManager.get()),
```
amortemousque, line 116: **❓ question:** ‏What is the intent behind this change?

packages/logs/src/boot/logsPublicApi.ts
bcaudan: The issue is that before init, we provide a fully functioning context manager and the intent of using the storage is only provided at init.

IMO, the right way to solve that would be to buffer context manager API calls and only applying them at init but that would be a breaking change.

In the mean time, there is a priority issue between the context in memory before init and the context in local storage if we want to reconcile them, which one should overwrite the other?
So, the current approach do not reconcile them.

wdyt?

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -100,22 +113,22 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
...
+    setGlobalContextProperty: monitor((key, value) => globalContextManager.setContextProperty(key, value)),
 
     /** @deprecated: use removeGlobalContextProperty instead */
-    removeLoggerGlobalContext: monitor(globalContextManager.remove),
-    removeGlobalContextProperty: monitor(globalContextManager.removeContextProperty),
+    removeLoggerGlobalContext: monitor((key) => globalContextManager.remove(key)),
+    removeGlobalContextProperty: monitor((key) => globalContextManager.removeContextProperty(key)),
 
-    clearGlobalContext: monitor(globalContextManager.clearContext),
+    clearGlobalContext: monitor(() => globalContextManager.clearContext()),
```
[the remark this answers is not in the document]
bcaudan: I tried to explain that in the commit description, let me know if it is not clear
Screenshot 2023-08-25 at 11 04 10

packages/logs/src/boot/logsPublicApi.ts
BenoitZugmeyer: I don't see the point of buffering context manager API calls? We could have a boolean flag to store whether the context did change before init, and if it has, store the current context in local storage. Or am I missing something?

packages/logs/src/boot/logsPublicApi.ts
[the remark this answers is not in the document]
amortemousque: Hum, I don't fully understand. Do you have a future configuration option in mind that would require us to do this?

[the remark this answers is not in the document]
bcaudan: This one:

Screenshot 2023-08-28 at 10 23 43

And you can see the override for example for RUM:

https://github.com/DataDog/browser-sdk/blob/6afa4c4ae72bc43323b4a56c3ced6c7da4dbb05e/packages/rum-core/src/boot/rumPublicApi.ts#L158-L162

packages/logs/src/boot/logsPublicApi.ts
BenoitZugmeyer: Same as after the init call, `{ foo: 'qux' }`?

bcaudan: Discussed IRL, let's combine contexts for now

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

amortemousque (approved):

PR #2382, opened 2023-08-18
[RUM-250] provide the worker string as a build env

BenoitZugmeyer:
## Motivation

This PR changes how the worker string is provided. Previously (PR https://github.com/DataDog/browser-sdk/pull/2319), we used an alternative entry point to the `@datadog/browser-worker` package, which worked well but we came across a dependency loop at the packages level because `worker` depends on `core` and we'll want `core` to depend on the `worker` so all SDKs can use it.

## Changes

Instead of depending on the `worker` package, the worker string is now directly injected at build time, similarly to the SDK version (build env).

This PR removes the worker build from the NPM postinstall hook. The Worker is now built by the build-env script when it is needed. This make the build a bit slower in general, hopefully not impacting the day to day work too much.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


BenoitZugmeyer:
* main
  * **PR #2376** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2376" target="_blank">Graphite</a> 
    * **PR #2377** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2377" target="_blank">Graphite</a> 
      * **PR #2382** <a href="https://app.graphite.dev/github/pr/DataDog/browser-sdk/2382" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/browser-sdk/2382?utm_source=stack-comment).

packages/worker/package.json
```diff
@@ -19,5 +19,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "webpack": "5.88.2"
```
bcaudan, line 24: **❓ question:** ‏do we need to add this dependency in the worker?

bcaudan (approved):

packages/worker/package.json
BenoitZugmeyer: If it's not declared, running `yarn build` from the `packages/worker` (like we do [here](https://github.com/DataDog/browser-sdk/blob/282e3c7aac9dbae47894a88d8e623418f036d2d3/scripts/lib/build-env.js#L43)) folder fails with `command not found: webpack`. It might be since the yarn upgrade. We may want to add it to other package if needed.

amortemousque (approved):

PR #2396, opened 2023-08-29
✨ [RUM-1016] allow to change "view.name"

arichard-info:
## Motivation

It would be very useful to be able to easily define the view.name field rather than letting datadog group the fields. This is the case, for example, if you have complex urls that datadog is unable to group :

```
/amazing-tee-shirt_P765231C097654
/green-hoodie_P209588C068396
/yellow-jacket_P999300C459483
/men_009890
/women_007865
/all_005432
```

Of these 6 urls, datadog groups them in the same way: `/?`, whereas in my application, the first three correspond to one type of page (regex : `^/[\w\-\.]+_P[0-9]+C([0-9]+)[^/]*$`), and the last three to another (regex : `^/([\w\-\.]+_[0-9]+)([^/]*)$`). I'd like to be able to name them explicitly.

As explained in the [documentation](https://docs.datadoghq.com/fr/real_user_monitoring/browser/modifying_data_and_context), currently the only way to change the name of a RUM view is to set `trackViewsManually` to true when initializing the SDK. You then need to call `datadogRum.startView` manually.

On an application already in production, which already uses datadog for the RUM, this represents a major implementation change, and there may be differences in the way it works with the default, and therefore differences in the metrics.

> For example, in the default mode (`trackViewsManually=false`), the `loading_type` is distinguished between `initial_load` and `route_change`. If you call `datadogRum.startView` manually, the value is always `initial_load`

To avoid these side effects, the aim would be to be able to define the name of a view without having to call the startView itself. We keep the native way of working, no need to make a custom implementation on the route.

My change is a proposal: I don't know why it's impossible to edit this field and I don't know exactly what the implications of this addition are.

## Changes

Possibility to set `event.view.name` in the `beforeSend` callback : 
```JavaScript
      beforeSend: (event) => {
        if (event.type === "view") {
          event.view.name = getViewName(event.view?.url);
        }
      },
```

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


BenoitZugmeyer (approved):

amortemousque (approved):

bcaudan:
Released in v4.49.0

Issue #2380, opened 2023-08-12
🐛 RUM: Styles coming from CSS-in-JS (Stitches) are not recorded


sandgraham:
**Describe the bug**

With `@datadog/browser-rum@v4.45.0`, we are seeing that session recordings do not pick up styles which come from a CSS-in-JS library we use: [Stitches](https://github.com/stitchesjs/stitches). We use Stitches along side static compiled CSS and we can see that the static CSS is recorded whereas styles coming from Stitches are not.

Stitches uses JavaScript to manage styles with the CSSOM API. We render a static `<style id="stitches" />` element in the `head` of our document, which Stitches uses for runtime styling injections.

Note, we had similar issues with FullStory. Originally they were only recording styles if added with `insertRule` to `CSSStyleSheets`, but Stitches adds styles to CSSMediaRules which are children of a top level `CSSStyleSheet`. I don't know the exact changes they made but I think they started recording `insertRule` to `CSSMediaRules` as well. Additionally, we had to add a class to the `<style id="stitches">` element to help FullStory's SDK pick up the runtime insertions.

**To Reproduce**

1. Load a web app using Stitches.
2. Observe that the DataDog session recording is not styled as the page is.

**Expected behavior**

I would expect the styles in the recording to reflect the styles applied in the browser. For instance, when using FullStory, all of our styles are picked up properly in session replays.

Here is an example of the un-styled pages we are observing. Note, styles that are picked up come from static CSS:
| Actual | DataDog |
| --- | --- |
| <img width="1424" alt="Screenshot 2023-08-12 at 04 25 09 PM@2x" src="https://github.com/DataDog/browser-sdk/assets/1939257/d9199a15-f158-4e64-abf7-3dc3066a8243"> | ![Screenshot 2023-08-10 at 3 54 46 PM](https://github.com/DataDog/browser-sdk/assets/1939257/e489866a-40a3-446a-b329-f809353ae8f8) |

bcaudan:
Hello @sandgraham,

If I remember correctly, we already did a couple of fixes for stitches.
Could you share a web app that reproduce the issue?

BenoitZugmeyer:
> Originally they were only recording styles if added with insertRule to CSSStyleSheets, but Stitches adds styles to CSSMediaRules which are children of a top level CSSStyleSheet. I don't know the exact changes they made but I think they started recording insertRule to CSSMediaRules as well

[...]

PR #2397, opened 2023-08-29
🐛 fix dynamic style serialization

BenoitZugmeyer:
## Motivation

See #2380. When a website uses a non-empty `<style>` element, edits it with `insertRule` or `deleteRule`, and then starts recording with `startSessionReplayRecording()`, changes made to the style are lost. Example:

>  Let's assume the page contains this piece of HTML:
> ```html
> <style>
> body { color: red }
> </style>
> ```
> If a script is modifying the <style> state:
> ```js
> document.querySelector('style').sheet.insertRule('body { font-size: 10px }')
> ```
>
> And only then we are taking a full snapshot
> ```js
> DD_RUM.startSessionReplayRecording()
> ```
> The inserted rule will be ignored during serialization. The <style> will be serialized like this:
> ```js
> {
>   type: NodeType.Element,
>   tagName: 'style',
>   attributes: {},
>   childNodes: [{
>     type: NodeType.Text,
>     textContent: 'body { color: red }',
>     isStyle: true
>   }]
> }
> ```

This behavior is inherited from rrweb which [still has it](https://github.com/rrweb-io/rrweb/blob/58c9104eddc8b7994a067a97daae5684e42f892f/packages/rrweb-snapshot/src/snapshot.ts#L658). There is no justification on why this condition is here in the commit history.

## Changes

* Always use the `<style>` element state (from its associated StyleSheet) when serializing it.
* Do not serialize `<style>` element children anymore
* Do not define the `isStyle` property on serialized text nodes

> In the above example, the `<style>` will be serialized like this:
> ```js
> {
>   type: NodeType.Element,
>   tagName: 'style',
>   attributes: { _cssText: 'body { font-size: 10px } body {color: red }' },
>   childNodes: []
> }
> ```

This change is compatible with the current implementation of the player.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/privacy.ts
+++ b/packages/rum/src/domain/record/privacy.ts
@@ -181,10 +181,7 @@ export function getTextContent(
   } else if (nodePrivacyLevel === NodePrivacyLevel.HIDDEN) {
     // Should never occur, but just in case, we set to CENSORED_MARK.
     textContent = CENSORED_STRING_MARK
-  } else if (
-    shouldMaskNode(textNode, nodePrivacyLevel)
-    // Style tags are `overruled` (Use `hide` to enforce privacy)
-  ) {
+  } else if (shouldMaskNode(textNode, nodePrivacyLevel)) {
     if (
       // Scrambling the child list breaks text nodes for DATALIST/SELECT/OPTGROUP
       parentTagName === 'DATALIST' ||
--- a/packages/rum/src/domain/record/privacy.ts
+++ b/packages/rum/src/domain/record/privacy.ts
@@ -173,7 +173,6 @@ export function getTextContent(
 
   const nodePrivacyLevel = parentNodePrivacyLevel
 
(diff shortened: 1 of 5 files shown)


packages/rum/src/domain/record/privacy.ts
```diff
@@ -183,9 +182,8 @@ export function getTextContent(
     // Should never occur, but just in case, we set to CENSORED_MARK.
     textContent = CENSORED_STRING_MARK
   } else if (
-    shouldMaskNode(textNode, nodePrivacyLevel) &&
+    shouldMaskNode(textNode, nodePrivacyLevel)
     // Style tags are `overruled` (Use `hide` to enforce privacy)
```
yannickadam, line 186: I guess this comment is now outdated?

ThibautGeriz (approved):

amortemousque (approved):

bcaudan (approved):

PR #2399, opened 2023-08-30
Report scroll metrics when page is resized

hamza-kadiri:
## Motivation

We currently have an issue in scrollmaps,  as you can see in [this widget](https://app.datadoghq.com/s/yB5yjZ/uqi-948-kxe), many users reach a scroll depth of 100% when the page scroll height is way lower than what it should be.

This is because when the user doesn't scroll, the scroll metrics are only reported once: at loading time.

Since loading time can be reached before the page is fully loaded, it means that we collect wrong scroll metrics

This PR aims to fix this issue by decorrelating scroll height reporting from scroll depth reporting.

## Changes

- Use a `ResizeObserver` to report scroll height increase

## Testing

Visit [this page](https://app.datadoghq.com/rum/heatmap?from_ts=1693474900184&live=false&query=%40application.id%3Aac8218cf-498b-4d33-bd44-151095959547+%40view.name%3A%2Anoresult%2A&to_ts=1694079700184) with the dev version of the sdk enabled and don't scroll. Clear the search, you should notice that the scroll height has increased

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackScrollMetrics.ts
```diff
@@ -18,11 +18,28 @@ export function trackScrollMetrics(
...
+
+    const now = relativeNow()
+    const maxDepthTime = elapsed(viewStart.relative, now)
+    maxDepth = scrollDepth
+    callback({
+      maxDepth,
+      maxDepthScrollHeight: scrollHeight,
+      maxDepthTime,
+      maxDepthScrollTop: scrollTop,
+    })
```
amortemousque, line 37: **💬 suggestion:** ‏You could factorise this code with the on in the `handleScrollEvent`

packages/rum-core/src/browser/scrollHeightObservable.ts
```diff
@@ -0,0 +1,15 @@
+import { Observable, setInterval, clearInterval } from '@datadog/browser-core'
+
+export const SCROLL_HEIGHT_OBSERVABLE_INTERVAL_MS = 500
+
+export function createScrollHeightObservable() {
+  const observable = new Observable<number>(() => {
+    const setIntervalId = setInterval(() => {
+      observable.notify(Math.round((document.scrollingElement || document.documentElement).scrollHeight))
```
amortemousque, line 8: **💬 suggestion:** ‏You could factorise `document.scrollingElement || document.documentElement).scrollHeight` with the one in `computeScrollValues()`

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
```diff
@@ -18,38 +28,79 @@ export function trackScrollMetrics(
...
+        passive: true,
+      }
+    )
+    stop = () => {
+      trotthledUpdateScrollMetrics.cancel()
+      resizeObserver.unobserve(observerTarget)
+      eventListener.stop()
+    }
+  } else {
+    const id = setInterval(updateScrollMetrics, throttleDuration)
```
amortemousque, line 98: **❓ question:** ‏Don't we want to update the scrollHeight also for IE? 
```suggestion
    const id = setInterval(() => {
      const scrollHeight = Math.round((document.scrollingElement || document.documentElement).scrollHeight)
      updateScrollMetrics(scrollHeight)
    }, throttleDuration)
```

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
```diff
@@ -18,38 +28,79 @@ export function trackScrollMetrics(
...
+    })
 
-  const { stop } = addEventListener(configuration, window, DOM_EVENT.SCROLL, handleScrollEvent.throttled, {
-    passive: true,
-  })
+    const observerTarget = document.scrollingElement || document.documentElement
+    const resizeObserver = new ResizeObserver(
+      monitor((entries) => {
+        const height = Math.round(entries?.[0]?.borderBoxSize?.[0]?.blockSize ?? 0)
+        if (height > 0) {
```
amortemousque, line 77: #

**❓ question:** ‏Why this check?

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.spec.ts
```diff
@@ -1,23 +1,36 @@
...
+  const newScroll = (scrollParams: { scrollHeight: number; scrollDepth: number; scrollTop: number }) => {
+    mockGetMetrics(scrollParams)
+    window.dispatchEvent(createNewEvent(DOM_EVENT.SCROLL))
+  }
+
+  const increaseHeight = (scrollParams: { scrollHeight: number; scrollDepth: number; scrollTop: number }) => {
+    mockGetMetrics(scrollParams)
+    const node = document.createElement('div')
+    node.style.height = `${scrollParams.scrollHeight}px`
+    document.body.replaceChildren(node)
```
BenoitZugmeyer, line 32: **🔨 warning:** ‏`replaceChildren` isn't supported in IE11

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.spec.ts
```diff
@@ -26,117 +39,85 @@ describe('trackScrollMetrics', () => {
...
+      {
         maxDepth: 500,
         maxDepthScrollTop: 100,
-        maxDepthTime: 1000 as Duration,
-      })
-    })
-
-    it('should update scroll metrics when scroll depth has increased', () => {
-      newScroll({ scrollHeight: 1000, scrollDepth: 500, scrollTop: 100 })
+        maxDepthScrollHeight: 0,
```
BenoitZugmeyer, line 85: **💭 thought:** `maxDepthScrollHeight` being `0` is a bit unexpected: `maxDepthScrollHeight` should always be greater than or equal to `maxDepth` right?

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.spec.ts
```diff
@@ -26,117 +59,61 @@ describe('trackScrollMetrics', () => {
...
-        maxDepth: 600,
-        maxDepthScrollTop: 200,
-        maxDepthTime: 1000 as Duration,
-      })
+    expect(callbackSpy).not.toHaveBeenCalledOnceWith({
+      maxDepth: 700,
+      maxDepthScrollHeight: 1900,
+      maxDepthTime: jasmine.any(Number),
+      maxDepthScrollTop: 100,
     })
```
BenoitZugmeyer, line 100: **💬 suggestion:** ‏remove this assertion. It is useless, since `toHaveBeenCalledOnceWith` already ensures that it has been called only once with the provided argument.

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.spec.ts
```diff
@@ -26,117 +59,61 @@ describe('trackScrollMetrics', () => {
...
-        maxDepth: jasmine.any(Number),
-        maxDepthTime: jasmine.any(Number),
-        maxDepthScrollTop: jasmine.any(Number),
-      })
+    expect(callbackSpy).not.toHaveBeenCalledOnceWith({
+      maxDepth: 600,
+      maxDepthScrollHeight: 2000,
+      maxDepthTime: jasmine.any(Number),
+      maxDepthScrollTop: 0,
     })
```
BenoitZugmeyer, line 117: **💬 suggestion:** ‏Same here, this assertion is not useful

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.spec.ts
```diff
@@ -26,117 +59,61 @@ describe('trackScrollMetrics', () => {
...
-        maxDepthTime: 2000 as Duration,
-      })
+  })
+  it('should update scroll height only if it has increased', () => {
+    scrollObservable.notify({ scrollDepth: 700, scrollHeight: 2000, scrollTop: 100 })
+    scrollObservable.notify({ scrollDepth: 700, scrollHeight: 1900, scrollTop: 100 })
+    expect(callbackSpy).toHaveBeenCalledOnceWith({
+      maxDepth: 700,
+      maxDepthScrollHeight: 2000,
+      maxDepthTime: jasmine.any(Number),
```
BenoitZugmeyer, line 92: **💬 suggestion:** ‏use `mockClock` to do assertions on the timestamp

BenoitZugmeyer (approved):

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
```diff
@@ -67,3 +88,39 @@ export function computeScrollValues() {
...
+export function createScrollValuesObservable(
+  configuration: RumConfiguration,
+  throttleDuration = THROTTLE_SCROLL_DURATION
+): Observable<ScrollValues> {
+  const observable = new Observable<ScrollValues>(() => {
+    function notify() {
+      observable.notify(computeScrollValues())
+    }
+
+    if (window.ResizeObserver) {
```
bcaudan, line 101: **❓ question:** ‏what about IE11? `ResizeObserver` is not supported there

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
```diff
@@ -67,3 +88,39 @@ export function computeScrollValues() {
...
+      const eventListener = addEventListener(configuration, window, DOM_EVENT.SCROLL, throttledNotify.throttled, {
+        passive: true,
+      })
+
+      return () => {
+        throttledNotify.cancel()
+        resizeObserver.unobserve(observerTarget)
+        eventListener.stop()
+      }
+    }
```
bcaudan, line 119: **💬 suggestion:** ‏unit tests are successful even if we remove all this section, could we add some tests on that?

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
hamza-kadiri: For IE11, the fallback solution is to collect data on the scrolling at regular intervals: 

https://github.com/DataDog/browser-sdk/blob/9ddc001adfbe716f2e38b0922d36a11ac661175e/packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts#L121-L122

bcaudan: oh right, so fallback to setInterval for IE 11 🤔
Do you see any pros / cons to have those two implementations rather than just always using setInterval?

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
bcaudan: I missed the fallback intent, let's discard this comment

packages/rum-core/src/domain/view/viewMetrics/trackScrollMetrics.ts
hamza-kadiri: cons: 
    - we have two different implementations, which can be troublesome
pro:
    - even if we have two implementations, the two still produce the same values, the only difference lies in the frequency with which these values are produced. For IE we're going to produce new values every second, for other browsers, we're going to produce them only when the page is scrolled or resized
    - on modern browser, since we produce values less often, we have less changes of causing layout shifts/reflows caused by scroll/viewport calculation

bcaudan: Indeed, the resize observer approach seems more "polite" than recomputing the values every second, I am wondering if it is not even too agressive for IE 😕
It could worth to double check that it does not degrade performances there.

Did you look into approaches taken by some resize observer polyfills?
In case there is some interesting alternate approach that could be used everywhere

hamza-kadiri: Do we have a way to test the SDK on IE? 

I can definitely try an approach with a resize observer polyfill.

bcaudan: > Do we have a way to test the SDK on IE?

Yes, with browserstack

BenoitZugmeyer (approved):

amortemousque (approved):

PR #2400, opened 2023-08-30
⚗️✨ [RUM-253] compress RUM data

BenoitZugmeyer:
## Motivation

To reduce the bandwidth impact when using the RUM Browser SDK, we want to compress data sent to the
backend.

## Changes

This PR implement data compression for the "rum" package/bundle behind an experimental feature flag.
This will stay like this during dogfooding, then we'll introduce a proper initialization parameter +
support it on "rum-slim" and "logs" as well.

## Testing

- [x] Local
- [x] Stagin
- [x] Unit
- [x] End to end (see #2416)

---


BenoitZugmeyer:

* main:
  * **PR #2414**
    * **PR #2415**
      * **PR #2400** 👈

packages/rum/src/domain/deflate/deflateEncoder.spec.ts
```diff
@@ -20,28 +20,175 @@ describe('createDeflateEncoder', () => {
...
+        [
+          {
+            output: new Uint8Array([...ENCODED_FOO, ...TRAILER]),
+            outputBytesCount: 4,
+            rawBytesCount: 3,
+            encoding: 'deflate',
+          },
+        ],
+      ])
+    })
```
bcaudan, line 191: **❓ question:** ‏In this case, is not it risky to rely on the previous finish callback? 
What about:
- discarding the initial finish callback
- return with finishSync:
  - output: all compressed data so far 
  - pendingData: all pending data, regardless of if is has been added after a previous finish

packages/core/src/transport/startBatchWithReplica.ts
```diff
@@ -3,24 +3,36 @@ import type { Context } from '../tools/serialisation/context'
...
-  const primaryBatch = createBatch(configuration, primary.endpoint)
-  const replicaBatch = replica && createBatch(configuration, replica.endpoint)
+  const primaryBatch = createBatch(configuration, primary)
+  const replicaBatch = replica && createBatch(configuration, replica)
 
-  function createBatch(configuration: Configuration, endpointBuilder: EndpointBuilder) {
+  function createBatch(configuration: Configuration, { endpoint, encoder }: BatchConfiguration) {
     return new Batch(
-      createHttpRequest(configuration, endpointBuilder, configuration.batchBytesLimit, reportError),
+      encoder || createIdentityEncoder(),
```
amortemousque, line 34: **💭 thought:** ‏I had a hard time understanding in which case the encoder was undefined and I'm wondering if it could me simplified. FMU we have to look at: 
- Identify that createDeflateEncoder can be undefined : https://github.com/DataDog/browser-sdk/blob/9862d64c4995ecf805f2dac4f357e3bcda387ed6/packages/rum-core/src/transport/startRumBatch.ts#L36
- Because  deflateWorker && createDeflateEncoder can be undefined : https://github.com/DataDog/browser-sdk/blob/9862d64c4995ecf805f2dac4f357e3bcda387ed6/packages/rum-core/src/boot/rumPublicApi.ts#L193-L195
- Then deflateWorker  can be null because of: https://github.com/DataDog/browser-sdk/blob/9862d64c4995ecf805f2dac4f357e3bcda387ed6/packages/rum-core/src/boot/rumPublicApi.ts#L152-L161

Personally I prefer to have a `rumPublicApi` as dump as possible and make the instantiation checks in startRum.
Also, It could be nice to create the encoder in startRum() to remove the logic from `startRumBatch` and `startBatchWithReplica`.  
I don't know all the intricacies of the implement so it might no be feasible 🙂 Let me know what you think.

BenoitZugmeyer: As seen IRL, I changed a bit how encoders are created so the condition on whether we use an "Identity" or "Deflate" encoder is clearer.

amortemousque (approved):

bcaudan (approved):

PR #2402, opened 2023-08-30
♻️✅ prevent event listener leaks in unit tests

bcaudan:
## Motivation

Avoid issues caused by leaked event listeners in tests

## Changes

- add event listener leak detection in unit tests
- set up / use cleanup methods for more components

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/test/leakDetection.ts
+++ b/packages/core/test/leakDetection.ts
@@ -1,4 +1,5 @@
 import { display } from '../src/tools/display'
+import { isIE } from '../src/tools/utils/browserDetection'
 import { getCurrentJasmineSpec } from './getCurrentJasmineSpec'
 
 let originalAddEventListener: typeof EventTarget.prototype.addEventListener
@@ -8,6 +9,9 @@ let wrappedListeners: {
 }
 
 export function startLeakDetection() {
+  if (isIE()) {
+    return
+  }
   wrappedListeners = {}
 
   // eslint-disable-next-line @typescript-eslint/unbound-method
@@ -31,6 +35,9 @@ export function startLeakDetection() {
 }
 
 export function stopLeakDetection() {
+  if (isIE()) {
+    return
+  }
   EventTarget.prototype.addEventListener = originalAddEventListener
   EventTarget.prototype.removeEventListener = originalRemoveEventListener
   wrappedListeners = {}


packages/rum/src/domain/deflate/deflateWorker.ts
```diff
@@ -61,10 +71,11 @@ export function startDeflateWorker(
   switch (state.status) {
     case DeflateWorkerStatus.Loading:
       state.initializationFailureCallbacks.push(onInitializationFailure)
-      return state.worker
+      return { worker: state.worker, stop: state.stop }
     case DeflateWorkerStatus.Initialized:
-      return state.worker
+      return { worker: state.worker, stop: state.stop }
   }
+  return { worker: undefined, stop: noop }
```
BenoitZugmeyer, line 78: **💬 suggestion:** ‏Instead of returning `stop` here, call `stop` in `resetDeflateWorkerState`. The DeflateWorker is a singleton, and if a test stops it without reseting its state, subsequent tests may fail.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -100,6 +101,8 @@ export function trackViews(
       startClocks,
       viewOptions
     )
+    cleanupTasks.push(() => newlyCreatedView.cleanup())
```
BenoitZugmeyer, line 104: **🔨 warning:** ‏this is a memory leak, because `stop()` is never called outside of tests

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackFirstHidden.ts
```diff
@@ -2,40 +2,42 @@ import type { RelativeTime } from '@datadog/browser-core'
...
-let trackFirstHiddenSingleton: { timeStamp: RelativeTime } | undefined
-let stopListeners: (() => void) | undefined
+let trackFirstHiddenSingleton: { timeStamp: RelativeTime; stop: () => void } | undefined
 
 export function trackFirstHidden(configuration: RumConfiguration, eventTarget: Window = window) {
   if (!trackFirstHiddenSingleton) {
     if (document.visibilityState === 'hidden') {
       trackFirstHiddenSingleton = {
         timeStamp: 0 as RelativeTime,
+        stop: reset,
```
BenoitZugmeyer, line 12: **💬 suggestion:** ‏mixing a `stop()` method with singletons scares me a bit, because `stop()` could be called while the singleton might still be used in a totally different place.

In this particular case, we know that it's not the case because everything using the singleton call stop() at the same time, but still, I don't think it is a good design.

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -43,9 +43,11 @@ describe('logs', () => {
   let interceptor: ReturnType<typeof interceptRequests>
   let requests: Request[]
   let handleLog: ReturnType<typeof startLoggerCollection>['handleLog']
+  let stopLogs: () => void
   let logger: Logger
   let consoleLogSpy: jasmine.Spy
   let displayLogSpy: jasmine.Spy
+  let cleanupTasks: Array<() => void>
```
BenoitZugmeyer, line 50: **💭 thought:** ‏I'm still thinking that a test helper like this could be very *very* helpful:

```js
const cleanupTasks = []

export function registerCleanupTask(task: () => void) {
  cleanupTasks.push(task)
}

afterEach(() => {
  cleanupTasks.splice(0).forEach(task => task())
})
```

(I already suggested it a few years(!) ago, and I clearly remember that you were against it because it did not follow jasmine concepts though 😄)

packages/core/src/browser/addEventListener.ts
```diff
@@ -111,26 +112,53 @@ export function addEventListeners<Target extends EventTarget, EventName extends
...
+        display.error(`Leaked listener
+  event names: "${eventNames.join('", "')}"
+  attached with: "${specWhenAdded}"        
+  executed with: "${testWindow._jasmineCurrentSpec}"`)
+      }
+      listener(event)
+    }
+  }
+  return listener
+}
```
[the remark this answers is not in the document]
bcaudan: I did not think about this approach, I'll give it a try

packages/logs/src/boot/startLogs.spec.ts
bcaudan: I am not sure to see what you want to do exactly or the debate you refer to 😕
Could you elaborate on what you would like to achieve?

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
bcaudan: Indeed 🤔
We could probably maintain active views somewhere until they stop tracking metrics and cleanup those active views after each tests.
I'll look into that.

packages/logs/src/boot/startLogs.spec.ts
BenoitZugmeyer: Well, imagine you have the code I wrote above in a `core/test/registerCleanupTask.ts`.

Then, you could refactor those patterns:

```ts
describe('....', () => {
  let cleanupTasks: Array<() => void>

  beforeEach(() => {
    cleanupTasks = []
  })

  afterEach(() => {
    cleanupTasks.forEach((task) => task())
  })

  it('....', () => {
    const { stop } = ...
    cleanupTasks.push(stop)
    ...
  })
```

into:

```ts
import { registerCleanupTask } from "../core/test"

describe('....', () => {
  it('....', () => {
    const { stop } = ...
    registerCleanupTask(stop)
    ...
  })
```

It could also be really helpful to use in other helpers, like `mockClock`:

```ts
import { resetNavigationStart } from '../../src/tools/utils/timeUtils'

export type Clock = ReturnType<typeof mockClock>

export function mockClock(date?: Date) {
  jasmine.clock().install()
  ...
  registerCleanupTask(() => {
      jasmine.clock().uninstall()
  })
  return {
    tick: (ms: number) => jasmine.clock().tick(ms),
    setDate: (date: Date) => jasmine.clock().mockDate(date),
  }
}
```

so tests don't have to worry about cleanup, they can just use the helper when they need to:

```ts
describe('...', () => {
  // no need for an afterEach()

  it('...', () => {
    const clock = mockClock()
  })
})
```

amortemousque: The @

bcaudan: The only concern that I have with this approach is that we would increasing the coupling of our test helpers with jasmine framework.
At least, having it centralised in the `registerCleanupTask` helper seems better than scattering `afterEach` in our helpers.
Let's try that.

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -100,6 +111,11 @@ export function trackViews(
       startClocks,
       viewOptions
     )
+    activeViews.push(newlyCreatedView)
+    newlyCreatedView.stopObservable.subscribe(() => {
+      activeViews.splice(activeViews.indexOf(newlyCreatedView))
```
BenoitZugmeyer, line 116: **💬 suggestion:** ‏Use a `Set` instead of an array for `activeViews`

packages/rum-core/src/domain/rumEventsCollection/view/trackViews.ts
```diff
@@ -262,9 +277,15 @@ function newView(
...
-      scheduleStopEventCountsTracking()
       triggerViewUpdate()
+      setTimeout(() => {
+        this.stop()
+      }, KEEP_TRACKING_AFTER_VIEW_DELAY)
+    },
+    stop() {
+      stopInitialViewMetricsTracking()
+      stopEventCountsTracking()
+      stopObservable.notify()
```
BenoitZugmeyer, line 288: **👏 praise:** ‏love it

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackFirstHidden.ts
```diff
@@ -2,40 +2,36 @@ import type { RelativeTime } from '@datadog/browser-core'
...
-            trackFirstHiddenSingleton!.timeStamp = event.timeStamp as RelativeTime
-            stopListeners!()
-          }
-        },
-        { capture: true }
-      ))
-    }
-  }
+  let timeStamp: RelativeTime
+  let stopListeners: () => void
```
BenoitZugmeyer, line 9: **💬 suggestion:** ‏type should be `(() => void) | undefined`

packages/core/test/leakDetection.ts
```diff
@@ -0,0 +1,53 @@
...
+    return originalRemoveEventListener.call(this, event, wrappedListener || listener, options)
+  }
+}
+
+export function stopLeakDetection() {
+  EventTarget.prototype.addEventListener = originalAddEventListener
+  EventTarget.prototype.removeEventListener = originalRemoveEventListener
+  Object.keys(wrappedListeners).forEach((key) =>
+    wrappedListeners[key].forEach((_, listener) => wrappedListeners[key].delete(listener))
+  )
```
BenoitZugmeyer, line 38: **💬 suggestion:** ‏it could be simpler with:

```suggestion
  Object.values(wrappedListeners).forEach((wrappedListenersForEvent) =>
    wrappedListenersForEvent.clear()
  )
```

or even simpler:

```suggestion
wrappedListeners = {}
```

BenoitZugmeyer (approved):

amortemousque (approved):

PR #2406, opened 2023-09-01
✅ improve E2E intake implementation and registry

BenoitZugmeyer:
## Motivation

Prepare for e2e tests on compression: extend `EventRegistry` to store "requests" information instead of only "events".
Also, do some light refactoring/cleanup. See individual commits.

## Changes

* Adjust namings
* Refactor how we read intake request body, allowing to refactor how we forward them to Datadog
* Store requests instead of "events", so e2e tests can do assertions on them
* Replace the "bridgeRegistry" with assertions on the intake registry requests

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [x] End to end

---


test/e2e/scenario/eventBridge.scenario.ts
```diff
@@ -72,20 +73,25 @@ describe('bridge present', () => {
...
-      expect(intakeRegistry.logsEvents.length).toBe(0)
-      expect(bridgeEvents.logsEvents.length).toBe(1)
+      expect(intakeRegistry.logsEvents.length).toBe(1)
+      expectAllRequestsComeFromBridge(intakeRegistry)
     })
 })
+
+function expectAllRequestsComeFromBridge(intakeRegistry: IntakeRegistry) {
+  expect(intakeRegistry.requests.every((request) => request.isBridge)).toBe(true)
+}
```
bcaudan, line 97: **🥜 nitpick:** similarly to `intakeRegistry.isEmpty()`, ‏what about exposing a method from the intakeRegistry directly?
to have something like:
```ts
expect(intakeRegistry.hasOnlyBridgeRequests()).toBe(true)
```

bcaudan (approved):

PR #2415, opened 2023-09-05
♻️ [RUM-253] adapt transport to send encoded data

BenoitZugmeyer:
## Motivation

We'll need to send the encoding as a new intake URL parameter.

## Changes

- Light refactoring
- Add the ability to set the encoding through "payload" objects.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end (see #2416)

---

### DIFF
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -28,6 +28,7 @@ export interface Payload {
   bytesCount: number
   retry?: RetryInfo
   flushReason?: FlushReason
+  encoding?: string
 }
 
 export interface RetryInfo {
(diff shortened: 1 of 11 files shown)


BenoitZugmeyer:

* main:
  * **PR #2414**
    * **PR #2415** 👈

packages/core/src/transport/httpRequest.ts
```diff
@@ -28,6 +28,7 @@ export interface Payload {
   bytesCount: number
   retry?: RetryInfo
   flushReason?: FlushReason
+  encoding?: string
```
bcaudan, line 31: **💬 suggestion:** ‏how about restricting the encoding value to `'deflate'` directly?

bcaudan (approved):

PR #2418, opened 2023-09-06
⚗️ [RUM-1020] Collect core web vitals target selectors

amortemousque:
## Motivation

Currently the web vitals are reported with a single value representing the vital score. These scores provide valuable insights to customers, allowing them to determine the performance quality of their website. However, to effectively improve them, it is crucial to identify the specific component to which a vital refers.

## Changes

- Move CWV telemetry to their dedicated metric module
- Collect LCP element selector
- Collect FID target selector
- Collect INP target selector
- Collect CLS first element (in document order) of the largest layout shift
- Introduce `appendElement()` test helper
- Do not flatten intermediary metric data structure

| before | after |
|--|--|
|image|image|
|image|image|

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -1,30 +1,53 @@
...
+    lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, [
+      {
+        entryType: 'largest-contentful-paint',
+        startTime: 789 as RelativeTime,
+        size: 10,
+        element: target,
+        ...overrides,
+      },
+    ])
+  }
```
bcaudan, line 35: **💬 suggestion:** ‏how about extracting some kind of performance entry builder in fixtures.ts, like what we have for events?

```ts
createPerformanceEntry<T>(entryType, overrides) {
  switch (entryType) {
    ...
  } 
}
```

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackFirstInputTimings.spec.ts
```diff
@@ -1,23 +1,58 @@
...
+    ])
+  }
 
   beforeEach(() => {
     configuration = {} as RumConfiguration
     fitCallback = jasmine.createSpy()
+
+    target = document.createElement('button')
+    target.setAttribute('id', 'fid-target-element')
+    document.body.appendChild(target)
```
bcaudan, line 54: **💬 suggestion:** ‏What about introducing an `appendElement(tagName)` and use registerCleanupTask to remove it automatically?

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackFirstInputTimings.ts
```diff
@@ -32,15 +42,25 @@ export function trackFirstInputTimings(
     )
     if (firstInputEntry) {
       const firstInputDelay = elapsed(firstInputEntry.startTime, firstInputEntry.processingStart)
+      let firstInputTargetSelector
 
-      webVitalTelemetryDebug.addWebVitalTelemetryDebug('FID', firstInputEntry.target, firstInputEntry.startTime)
+      if (
+        isExperimentalFeatureEnabled(ExperimentalFeature.WEB_VITALS_ATTRIBUTION) &&
+        firstInputEntry.target &&
+        isElementNode(firstInputEntry.target)
```
bcaudan, line 50: **💬 suggestion:** ‏how about adding a test when the target is not an element?

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackCommonViewMetrics.ts
```diff
@@ -86,7 +91,9 @@ export function trackCommonViewMetrics(
     },
     setLoadEvent,
     getCommonViewMetrics: () => {
-      commonViewMetrics.interactionToNextPaint = getInteractionToNextPaint()
+      const { interactionToNextPaint, interactionToNextPaintTargetSelector } = getInteractionToNextPaint() || {}
+      commonViewMetrics.interactionToNextPaint = interactionToNextPaint
+      commonViewMetrics.interactionToNextPaintTargetSelector = interactionToNextPaintTargetSelector
```
bcaudan, line 99: **💬 suggestion:** ‏what about not flattening the intermediate data structure?
it would avoid to update the intermediary layer at each time we want to add extra properties to an already collected metrics

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackLargestContentfulPaint.spec.ts
amortemousque: Good idea! I created it. For now, I only used the fixures for the Core Web Vital but I'll do a followup PR to extend its usage to all the tests.

packages/rum-core/src/domain/rumEventsCollection/view/viewMetrics/trackFirstInputTimings.spec.ts
amortemousque: Yes, I'll extends its usage to all tests in a following PR

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -77,26 +77,26 @@ function processViewUpdate(
...
+      interaction_to_next_paint_target_selector: view.commonViewMetrics.interactionToNextPaint?.targetSelector,
       is_active: view.isActive,
       name: view.name,
-      largest_contentful_paint: toServerDuration(view.initialViewMetrics.largestContentfulPaint),
-      largest_contentful_paint_target_selector: view.initialViewMetrics.largestContentfulPaintTargetSelector,
-      load_event: toServerDuration(view.initialViewMetrics.loadEvent),
+      largest_contentful_paint: toServerDuration(view.initialViewMetrics.largestContentfulPaint?.value),
+      largest_contentful_paint_target_selector: view.initialViewMetrics.largestContentfulPaint?.targetSelector,
+      load_event: toServerDuration(view.initialViewMetrics.navigationTimings?.loadEvent),
       loading_time: discardNegativeDuration(toServerDuration(view.commonViewMetrics.loadingTime)),
```
BenoitZugmeyer: **💭 thought:** ‏this is a *lot* of unminifiable code

packages/rum-core/test/dom.ts
```diff
@@ -28,3 +30,25 @@ export function createIsolatedDom() {
     },
   }
 }
+
+export function appendElement(tagName: string, attributes: { [key: string]: string }) {
```
BenoitZugmeyer, line 34: **💬 suggestion:** ‏I think this kind of utility is great, but it duplicates a bit the use-cases of `createIsolatedDom`. It could be great if we kept one tool for the job, but I don't think the current `appendElement` implementation is practical enough to replace the `createIsolatedDom` usages in the future.

```ts
function useIsolatedDom(html: string) {
  const dom = createIsolatedDom()
  registerCleanupTask(() => dom.clear())
  return dom.append(html)
}
//
const div = useIsolatedDom('<div id="toto">')
const text = useIsolatedDom('toto') // createIsolatedDom could be modified a bit to support this use-case
```

then we could progressively remove `createIsolatedDom` direct usages.

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -337,17 +366,9 @@ function handleRumPerformanceEntries(
...
-      entry.entryType === 'paint' ||
-      entry.entryType === 'longtask' ||
-      entry.entryType === 'largest-contentful-paint' ||
-      entry.entryType === 'first-input' ||
-      entry.entryType === 'layout-shift' ||
-      entry.entryType === 'event'
-  ) as RumPerformanceEntry[]
+  const rumPerformanceEntries = entries.filter((entry): entry is RumPerformanceEntry =>
+    objectHasValue(RumPerformanceEntryType, entry)
+  )
```
BenoitZugmeyer, line 371: **💭 thought:** ‏I wonder if we really need to keep filtering entries like this.

packages/rum-core/src/browser/performanceCollection.ts
```diff
@@ -199,12 +225,15 @@ export function retrieveInitialDocumentResourceTiming(
     let timing: RumPerformanceResourceTiming
 
     const forcedAttributes = {
-      entryType: 'resource' as const,
+      entryType: RumPerformanceEntryType.RESOURCE as const,
```
BenoitZugmeyer, line 228: **🥜 nitpick:** ‏the `as const` is probably unnecessary now

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -37,20 +49,35 @@ export function trackInteractionToNextPaint(viewLoadingType: ViewLoadingType, li
...
+      if (
+        isExperimentalFeatureEnabled(ExperimentalFeature.WEB_VITALS_ATTRIBUTION) &&
+        newInteraction.target &&
+        isElementNode(newInteraction.target)
+      ) {
+        interactionToNextPaintTargetSelector = getSelectorFromElement(
+          newInteraction.target,
+          configuration.actionNameAttribute
+        )
+      }
```
BenoitZugmeyer, line 64: **💬 suggestion:** ‏if the target of a first interaction is an element, then the target of a second interaction is not an element, the selector still points to the first interaction. Reset the selector maybe?
```suggestion
      if (
        isExperimentalFeatureEnabled(ExperimentalFeature.WEB_VITALS_ATTRIBUTION) &&
        newInteraction.target &&
        isElementNode(newInteraction.target)
      ) {
        interactionToNextPaintTargetSelector = getSelectorFromElement(
          newInteraction.target,
          configuration.actionNameAttribute
        )
      } else {
        interactionToNextPaintTargetSelector = undefined
      }
```

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -77,22 +77,26 @@ function processViewUpdate(
...
       error: {
         count: view.eventCounts.errorCount,
       },
       first_contentful_paint: toServerDuration(view.initialViewMetrics.firstContentfulPaint),
-      first_input_delay: toServerDuration(view.initialViewMetrics.firstInputDelay),
-      first_input_time: toServerDuration(view.initialViewMetrics.firstInputTime),
-      interaction_to_next_paint: toServerDuration(view.commonViewMetrics.interactionToNextPaint),
+      first_input_delay: toServerDuration(view.initialViewMetrics.firstInputTimings?.firstInputDelay),
+      first_input_time: toServerDuration(view.initialViewMetrics.firstInputTimings?.firstInputTime),
+      first_input_target_selector: view.initialViewMetrics.firstInputTimings?.firstInputTargetSelector,
```
BenoitZugmeyer, line 92: **💬 suggestion:** ‏
```suggestion
      first_input_delay: toServerDuration(view.initialViewMetrics.firstInput?.delay),
      first_input_time: toServerDuration(view.initialViewMetrics.firstInput?.time),
      first_input_target_selector: view.initialViewMetrics.firstInput?.targetSelector,
```

packages/rum-core/test/dom.ts
amortemousque: It's true. I would also prefer a single API. My only concern is that isolatedDom
- uses `innerHTML`. Ideally I'd like to use this new API also in Replay and that may be a problem. 
- it uses an iframe which isolate the DOM and the event triggered inside it which may cause problem for some tests 
Wdyt?

amortemousque: Like discussed. I'll make a following PR to harmonize the DOM APIs.

packages/rum-core/src/browser/performanceCollection.ts
amortemousque: Interesting 🤔. Do you remember why it was done in the first place?

packages/rum-core/src/browser/performanceCollection.ts
amortemousque: Without it, it seems like typescript complains  
```
Types of property 'entryType' are incompatible.
    Type 'RumPerformanceEntryType' is not assignable to type 'RumPerformanceEntryType.RESOURCE'
```

packages/rum-core/src/domain/view/viewCollection.ts
amortemousque: Yes. Do you have something in mind to improve it?

packages/rum-core/src/browser/performanceCollection.ts
BenoitZugmeyer: Ah right nevermind!

packages/rum-core/src/domain/view/viewCollection.ts
BenoitZugmeyer: One way could be to change the intermediary representation so we can use `combine` and avoid repeating every single fields. For example, `InitialViewMetrics` could be:

```ts
export interface InitialViewMetrics {
  first_contentful_paint?: ServerDuration
  first_byte?: ServerDuration
  ...
}
```

BenoitZugmeyer: Since this PR is rather large already, we might want to postpone this change after thinking a bit more about it.

packages/rum-core/src/browser/performanceCollection.ts
amortemousque: FMU, we still have this case to filter https://github.com/DataDog/browser-sdk/blob/fbe0200005f0581fab050d2f337a0e0dfe07d925/packages/rum-core/src/browser/performanceCollection.ts#L154-L155

BenoitZugmeyer: Ah right that's why :D Ok nevermind then. We can always change things if/when we work on my suggestion here: https://github.com/DataDog/browser-sdk/pull/2355#discussion_r1302738495

BenoitZugmeyer (approved):

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
```diff
@@ -21,13 +16,7 @@ describe('trackCumulativeLayoutShift', () => {
...
   function newLayoutShift(lifeCycle: LifeCycle, overrides: Partial<RumLayoutShiftTiming>) {
     lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, [
-      {
-        entryType: 'layout-shift',
-        startTime: relativeNow(),
-        hadRecentInput: false,
-        value: 0.1,
-        ...overrides,
-      },
+      createPerformanceEntry(RumPerformanceEntryType.LAYOUT_SHIFT, overrides),
```
bcaudan: **🥜 nitpick:** ‏what about removing the `newLayoutShift` helper? it does not seem to bring a lot now

bcaudan (approved):

PR #2428, opened 2023-09-15
♻️ Use performance entry fixtures in tests

amortemousque:
## Motivation

After introducing` createPerformanceEntry()` fixture builder  ([PR](https://github.com/DataDog/browser-sdk/pull/2418)). Let's use it with all tests playing with performance entries.

## Changes

Update tests:
image

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
+++ b/packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
@@ -12,7 +12,7 @@ describe('matchRequestTiming', () => {
     startClocks: relativeToClocks(100 as RelativeTime),
     duration: 500 as Duration,
   }
-  let entries: PerformanceResourceTiming[]
+  let entries: RumPerformanceResourceTiming[]
 
   beforeEach(() => {
     if (isIE()) {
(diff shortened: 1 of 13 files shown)


packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
```diff
@@ -22,63 +23,75 @@ describe('matchRequestTiming', () => {
...
+
+export function createResourceEntry(overrides?: Partial<RumPerformanceResourceTiming>): PerformanceResourceTiming {
+  const rumPerformanceResourceTiming: Partial<PerformanceResourceTiming> = createPerformanceEntry(
+    RumPerformanceEntryType.RESOURCE,
+    overrides
+  )
+  return {
+    ...rumPerformanceResourceTiming,
+    toJSON: () => rumPerformanceResourceTiming,
+  } as PerformanceResourceTiming
```
bcaudan, line 96: **💬 suggestion:** ‏what about having this behaviour for all performance entries?

packages/rum-core/src/domain/resource/resourceCollection.spec.ts
```diff
@@ -42,14 +43,12 @@ describe('resourceCollection', () => {
...
     const { lifeCycle, rawRumEvents } = setupBuilder.build()
-    const performanceEntry = createResourceEntry({
-      duration: 100 as Duration,
-      name: 'https://resource.com/valid',
-      startTime: 1234 as RelativeTime,
-    })
+
+    const performanceEntry = createPerformanceEntry(RumPerformanceEntryType.RESOURCE)
+
     lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, [performanceEntry])
```
bcaudan, line 49: **💬 suggestion:** ‏spaces needed?
```suggestion
    const performanceEntry = createPerformanceEntry(RumPerformanceEntryType.RESOURCE)
    lifeCycle.notify(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, [performanceEntry])
```

bcaudan (approved):

packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
amortemousque: It would mean adding `toJSON` to RumPerformanceResourceTiming type only for this specific case?

packages/rum-core/src/domain/resource/resourceCollection.spec.ts
amortemousque: Sure

packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
bcaudan: oh indeed, we don't have it for all performance entries.
to be noted, it will be added to resource type with the v5 and it is already in long task type.

PR #2429, opened 2023-09-15
allow the use of proxy suffix routing instead of ddforward

hermanbanken:
## Motivation & changes
The page [1] documents how to set up a proxy. This mentions that the `ddforward` parameter needs to be parsed and used as the path in the proxied upstream url. This detail makes it a lot more involved to setup a proxy.

Instead, it would be much simpler if we could setup a virtual host / path like /rum/* which proxies everything after it by stripping /rum and using the path suffix as path instead. This would work with any standard proxy like NGINX, but also in Load Balancer configurations like the Google Global Load Balancer. The latter would require 0 new infrastructure components for us to manage, while the current setup would require a additional deployment to manage.

Given that we could already configure the Load Balancer in the way we desire, the only problem is that the JS/NPM RUM library uses the ddforward parameter, instead of a path suffix. Could the RUM library be extended with an extra option to not use ddforward but use the path as a suffix instead?

1 = https://docs.datadoghq.com/real_user_monitoring/guide/proxy-rum-data/?tab=npm#proxy-setup

## Changes
https://help.datadoghq.com/hc/en-us/requests/1343144

## Testing
Trivial?

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan:
Hi @hermanbanken and @pacome2018, 

If I understand correctly, the idea would be to configure:

```js
proxy: 'https://example.org'
```
and have the data sent to:

```
https://example.com/api/v2/rum?ddsource=browser&...
```
We considered implementing this approach the last time we updated our proxy strategy.
However, we did not implemented it at the time because some privacy lists already block any requests containing `/rum?ddsource=` (cf [EasyPrivacy list](https://github.com/easylist/easylist/blob/997fb6533c719a015c21723b34e0cedefcc0d83d/easyprivacy/easyprivacy_general.txt#L3840)).
Were you considering using a proxy for ad blockers or for another need?

hermanbanken:
We have very strict CORS policies and made exceptions for DataDog. We'd like to have our app only contact our own domain if possible. We also want to avoid any blockers, but it would not be a big deal if we miss a percentage of traffic.

The different way of routing makes it much simpler for us to use. We could also manually rewrite to include `/api/v2/rum` on our end, if that avoids ad (these are not ads 😢) blocking. Only the query-to-path-conversion makes it harder to add to a GCP UrlMap.

bcaudan:
> We could even also manually rewrite to include `/api/v2/rum`

That is probably not something that would work for most cases since this path is different for logs or replay and we may also want to still have control on that if we introduce an api v3 at some point.

A way to be more flexible there could be to allow you to build the proxy url how you want.
We could provide something like:

```js
proxyUrlFn: (path, parameters) => {
  return `https://example.com/${path}?${parameters}`
  // or return `https://example.com/my-prefix/${path}/my-suffix?${parameters}`
}
```

What do you think?

hermanbanken:
@bcaudan that would be great, and the most flexible!

bcaudan:
Closed in favour of #2502, see [corresponding documentation](https://docs.datadoghq.com/real_user_monitoring/guide/proxy-rum-data/?tab=npm#alternate-sdk-setup).

PR #2434, opened 2023-09-21
♻️ [RUM-1051] Harmonize DOM manipulation in tests

amortemousque:
## Motivation

After introducing `appendElement` to ease DOM manipulation in our tests (cf [PR](https://github.com/DataDog/browser-sdk/pull/2418)), let generalise its usage.

## Changes

- IsolatedDom is not isolated anymore (remove the iframe usage)
- Replace appendElement by append (we can also append text nodes)
- Replace isolatedDom by append
- Update sandbox pattern with append

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/test/dom.ts
+++ b/packages/rum-core/test/dom.ts
@@ -1,34 +1,17 @@
 import { registerCleanupTask } from '@datadog/browser-core/test'
 
-export type IsolatedDom = ReturnType<typeof createIsolatedDom>
+export function append<E extends Element>(html: string, container = document.body): E {
+  const tmp = document.createElement('div')
+  tmp.innerHTML = html.trim()
 
-export function createIsolatedDom() {
-  function append<E extends Element>(html: string, container: HTMLElement): E {
-    const tmp = document.createElement('div')
-    tmp.innerHTML = html.trim()
+  const target = tmp.querySelector('[target]') || tmp.childNodes[0]
+  const nodes = Array.from(tmp.childNodes)
 
-    const target = tmp.querySelector('[target]') || tmp.childNodes[0]
-    const nodes = Array.from(tmp.childNodes)
+  nodes.forEach((node) => container.appendChild(node))
 
-    nodes.forEach((node) => container.appendChild(node))
+  registerCleanupTask(() => {
+    nodes.forEach((node) => node.remove())
+  })
 
-    registerCleanupTask(() => {
-      nodes.forEach((node) => node.remove())
-    })
-
-    return target as E
-  }
-
-  return {
-    element(s: TemplateStringsArray) {
-      return append(s[0], document.body)
-    },
-    append: <E extends Element>(html: string): E => append(html, document.body),
-    appendToHead: <E extends Element>(html: string): E => append<E>(html, document.head),
-    clear() {},
-  }
-}
-
-export function append(html: string) {
-  return createIsolatedDom().append(html)
+  return target as E
 }
--- a/packages/rum-core/test/dom.ts
+++ b/packages/rum-core/test/dom.ts
@@ -29,24 +29,6 @@ export function createIsolatedDom() {
   }
 }
 
-export function appendElement(tagName: string, attributes: { [key: string]: string }) {
-  const element = document.createElement(tagName)
-
-  for (const key in attributes) {
-    if (Object.prototype.hasOwnProperty.call(attributes, key)) {
-      element.setAttribute(key, attributes[key])
-    }
-  }
-
-  return append(element)
-}
-
-export function appendTextNode(text: string) {
-  return append(document.createTextNode(text))
-}
-
-function append<T extends Node = Node>(node: T): T {
-  document.body.appendChild(node)
-  registerCleanupTask(() => node.parentNode!.removeChild(node))
-  return node
+export function append(html: string) {
+  return createIsolatedDom().append(html)
 }
--- a/packages/rum-core/test/dom.ts
+++ b/packages/rum-core/test/dom.ts
@@ -1,6 +1,9 @@
 import { registerCleanupTask } from '@datadog/browser-core/test'
 
-export function append<E extends Element>(html: string, container = document.body): E {
+export function append<E extends Element | Text = Element>(
+  html: string,
+  container: Element | ShadowRoot = document.body
+): E {
   const tmp = document.createElement('div')
   tmp.innerHTML = html.trim()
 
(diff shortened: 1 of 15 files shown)


packages/rum/src/domain/record/observers/inputObserver.spec.ts
```diff
@@ -138,7 +133,7 @@ describe('initInputObserver', () => {
     const host = document.createElement('div')
     host.attachShadow({ mode: 'open' })
     const event = createNewEvent('input', { target: host, composed: true })
-    event.composedPath = () => [input, host, sandbox, document.body]
```
BenoitZugmeyer, line 141: **💬 suggestion:** I understand that we don't have the `sandbox` reference anymore, but this was implemented to mimic the browser behavior. Could this work?‏
```ts
event.composedPath = () => [input, host, input.parentElement, document.body]
```

packages/rum/src/domain/record/observers/mutationObserver.spec.ts
```diff
@@ -883,16 +847,16 @@ describe('startMutationCollection', () => {
...
 
         flushMutations()
 
         expect(mutationCallbackSpy).not.toHaveBeenCalled()
       })
 
       it('when mutating a known child node', () => {
-        const textNode = document.createTextNode('function foo() {}')
-        sandbox.appendChild(textNode)
+        const textNode = append<Text>('function foo() {}', sandbox)
```
BenoitZugmeyer, line 858: **💭 thought:** ‏it might have been better to keep `appendText` / `appendElement` so we don't need to "cast" the return like this. We could also imagine using TS to infer this properly: https://www.typescriptlang.org/play?#code/FAFwngDgpgBA8gVxBJAeAKjKAPEUB2AJgM4zEgBOAlvgOYB8MAvMDG5jnkaQAYwAkAbxoAzKBRgUo5AL58A-PCQoQqKeUYAuVuyy4CJGDwA6+IaPGTpIOTEWJkadSC06YHfdyOpz+MRIB9W0UAUQAbKABbAhAYbTZ3PS5DHl9-GCCAbRAAQwpaKBAAXTTLIIUYcKiYuLd2TmBgQigAYzC82BEEfBaQKgB7fBgciGgiDCSDUnJqOnoACgALEEiwzXcASnWHFQx6AG5GlsHyLAjo-FimYdGDeYAiVBB+5-p7jeAAek+EtgA9eSNYDHfCnKDnGIAJmYNzGhAeCSeL36bw+31+MABQJBpzwuBhIzhCLYz2e7y+P1+WKOJ1i4OqlwAzATbkQHqT+jAkc8YNk8gViqiKRisUA

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -586,11 +557,17 @@ describe('serializeNodeWithId', () => {
...
 
     describe('<link rel="stylesheet"> elements', () => {
+      let originalStyleSheets: StyleSheetList
+
+      beforeEach(() => {
+        originalStyleSheets = document.styleSheets
+      })
+      afterEach(() => {
+        Object.defineProperty(document, 'styleSheets', { value: originalStyleSheets, configurable: true })
+      })
```
BenoitZugmeyer, line 567: **💬 suggestion:** ‏the `styleSheets` property is defined on a `document` prototype, that's why you failed to restore it with `const originalDescriptor = Object.getOwnPropertyDescriptor(document, 'styleSheets')`.

You could simply `delete document.styleSheets` to remove the mocked property defined in tests.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -150,11 +134,7 @@ describe('serializeNodeWithId', () => {
     })
 
     it('serializes attributes', () => {
-      const element = document.createElement('div')
-      element.setAttribute('foo', 'bar')
-      element.setAttribute('data-foo', 'data-bar')
-      element.className = 'zog'
-      element.style.width = '10px'
+      const element = append('<div foo="bar" data-foo="data-bar" class="zog" style="width: 10px;"></div>')
```
BenoitZugmeyer, line 137: **🥜 nitpick:** ‏this was ensuring that browsers return the changed attributes even if they are set with specific setters instead of `setAttribute`.

It could be nice to keep it this way:

```ts
const element = append('<div foo="bar" data-foo="data-bar"></div>')
element.className = 'zog'
element.style.width = '10px'
```

packages/rum/src/domain/record/observers/inputObserver.spec.ts
amortemousque: Sure!

packages/rum/src/domain/record/observers/mutationObserver.spec.ts
amortemousque: Okay, why not keeping separating `appendText` and `appendElement` however there are also other cases where we need to cast to specify which HTMLElement we want and I don't know if a smart inferring worth it. How would you handle them? 
cf: https://github.com/DataDog/browser-sdk/blob/8dc1b90601b8bf398d6604ead3cbeb4534f92937/packages/rum/src/domain/record/record.spec.ts#L298

BenoitZugmeyer: I would just cast them `as HTMLDivElement`.

Functions like `function foo<T>(): T` are implicitly casting the return value anyway.

amortemousque: I splitted to have appendText and appendElement

packages/rum-core/test/dom.ts
```diff
@@ -1,10 +1,18 @@
...
+  container.appendChild(textNode)
+
+  registerCleanupTask(() => {
+    container.removeChild(textNode)
+  })
+
+  return textNode
+}
+
+export function appendElement(html: string, container: Element | ShadowRoot = document.body): HTMLElement {
```
BenoitZugmeyer, line 15: Because it uses `tmp.childNotes[0]`, this function can still return a `Text` node. Maybe use `tmp.children[0]` like before? (Note: I noticed some tests where `appendElement` will need to be changed to `appendText`)

BenoitZugmeyer (approved):

packages/rum-core/test/dom.ts
amortemousque: Yes good idea

BenoitZugmeyer (approved):

PR #2450, opened 2023-10-04
♻️ [RUM-1329] merge tracekit into error

amortemousque:
## Motivation

Tracekit is used for uncaught error and stacktraces computation which are part of our error handling. Having a tracekit module and an error module do not really make sense. Let's merge them.

## Changes

Move tracekit in trackRuntimeError module.

## Testing

- [ ] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


packages/core/src/domain/tracekit/tracekit.ts
```diff
@@ -1,96 +0,0 @@
...
- * - Konqueror 3.5+ (untested)
- *
- * Tries to catch all unhandled errors and report them to the
- * callback.
- *
- * Callbacks receive a StackTrace object as described in the
- * computeStackTrace docs.
- *
- * @memberof TraceKit
- * @namespace
```
amortemousque, line 38: This comment is no longer accurate since https://github.com/DataDog/browser-sdk/pull/1335/files#diff-c33db8fef9eab4ff4982984c1dd39c92172aaefd0f88d528f205dcffa82d35cd. Let's remove it

packages/core/src/domain/error/computeStackTrace.ts
```diff
@@ -1,11 +1,34 @@
+/**
+ *
+ * Cross-browser stack trace computation.
```
bcaudan, line 3: **🥜 nitpick:** ‏
```suggestion
/**
 * Cross-browser stack trace computation.
```

bcaudan (approved):

PR #2452, opened 2023-10-05
[RUM-1062] remove message prefix when using the console handler

BenoitZugmeyer:
## Motivation

#2210 brought console log prefixes to messages displayed by the SDK to make things more explicit.

In the process, logs displayed on the console (using the `console` handler) also got that prefix, whichwas not intended. This PR fixes that issue.

## Changes

* keep references to original console methods
* use them directly in the loggerCollection instead of prefixed ones

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/logs/src/domain/logger/loggerCollection.ts
```diff
@@ -51,3 +58,11 @@ export function isAuthorized(status: StatusType, handlerType: HandlerType, logge
...
   )
 }
+
+function displayInConsole(logsMessage: LogsMessage, messageContext: Context | undefined) {
+  const consoleMethod = Object.prototype.hasOwnProperty.call(originalConsoleMethods, logsMessage.status)
+    ? originalConsoleMethods[logsMessage.status]
+    : originalConsoleMethods.log
+
+  consoleMethod.call(globalConsole, logsMessage.message, messageContext)
+}
```
bcaudan, line 68: **💬 suggestion:** ‏what about adding some tests for:
- ensuring that there is no prefix in this case
- console.log is used when the status is not in originalConsoleMethods

BenoitZugmeyer: > **💬 suggestion:** ‏what about adding some tests for:
> 
>     * ensuring that there is no prefix in this case

Added a E2E test for it

> 
>     * console.log is used when the status is not in originalConsoleMethods

I removed this logic, because it cannot happen in practice: all `StatusType` has a matching `ConsoleApiName`, and `isAuthorized` makes sure the logs `status` is valid. [This test](https://github.com/DataDog/browser-sdk/blob/2d4e4caa73fc01970c85dddcadf647fc3e3c084e/packages/logs/src/domain/logger/loggerCollection.spec.ts#L100) makes sure things do not break when using an unknown status type

packages/core/src/tools/display.ts
```diff
@@ -43,10 +35,18 @@ export const display: Display = (api, ...args) => {
...
-display.info = globalConsole.info.bind(globalConsole, PREFIX)
-display.warn = globalConsole.warn.bind(globalConsole, PREFIX)
-display.error = globalConsole.error.bind(globalConsole, PREFIX)
+export const display: Display = {
+  debug: originalConsoleMethods.debug.bind(globalConsole, PREFIX),
+  log: originalConsoleMethods.log.bind(globalConsole, PREFIX),
+  info: originalConsoleMethods.info.bind(globalConsole, PREFIX),
+  warn: originalConsoleMethods.warn.bind(globalConsole, PREFIX),
+  error: originalConsoleMethods.error.bind(globalConsole, PREFIX),
+}
```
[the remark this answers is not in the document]
BenoitZugmeyer: [This e2e test](https://github.com/DataDog/browser-sdk/blob/2d4e4caa73fc01970c85dddcadf647fc3e3c084e/test/e2e/scenario/rum/init.scenario.ts#L14) already ensures logs are displayed with a prefix

bcaudan (approved):

PR #2464, opened 2023-10-12
⚗️🐛 [RUM-1561] Compute selector from a detached element

bcaudan:
## Motivation

Telemetry errors where an element don't have parent during selector computation:

Screenshot 2023-10-12 at 17 53 53

It seems to only occur during web vital attribution. 

## Changes

Behind `detached_element_selector` flag, stop selector algorithm when an element don't have parent and generate a `DETACHED>...` selector.

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/getSelectorFromElement.ts
+++ b/packages/rum-core/src/domain/getSelectorFromElement.ts
@@ -1,4 +1,4 @@
-import { cssEscape } from '@datadog/browser-core'
+import { cssEscape, isExperimentalFeatureEnabled, ExperimentalFeature } from '@datadog/browser-core'
 import { DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE } from './action/getActionNameFromElement'
 
 /**
@@ -52,6 +52,10 @@ export function getSelectorFromElement(targetElement: Element, actionNameAttribu
     )
     if (globallyUniqueSelector) {
       return globallyUniqueSelector
+    }
+
+    if (isExperimentalFeatureEnabled(ExperimentalFeature.DETACHED_ELEMENT_SELECTOR) && !element.parentElement) {
+      return getSelectorFromDetachedElement(element, targetElementSelector)
     }
 
     const uniqueSelectorAmongChildren = findSelector(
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -144,6 +148,10 @@ function getPositionSelector(element: Element): string {
   return `${cssEscape(element.tagName)}:nth-of-type(${elementIndex})`
 }
 
+function getSelectorFromDetachedElement(element: Element, targetElementSelector: string) {
+  return combineSelector(`DETACHED>${cssEscape(element.tagName)}`, targetElementSelector)
```
amortemousque, line 152: **❓ question:** ‏Does this selector allow to select the node with `document.querySelector()`? If this is not the case, I'm wondering if we should just avoid computing the selector.

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -54,6 +54,10 @@ export function getSelectorFromElement(targetElement: Element, actionNameAttribu
       return globallyUniqueSelector
     }
 
+    if (isExperimentalFeatureEnabled(ExperimentalFeature.DETACHED_ELEMENT_SELECTOR) && !element.parentElement) {
```
amortemousque, line 57: **💬 suggestion:** ‏ I don't know if only detached element can have no parent. Maybe it could worth collecting [element.isConnected](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected) to check. Wdyt?

packages/rum-core/src/domain/getSelectorFromElement.ts
bcaudan: Definitely things that we should discuss before enabling that everywhere.
I'd like to test it like that just to have a bit more info when it happens and having the prefix would help.

packages/rum-core/src/domain/getSelectorFromElement.ts
bcaudan: ok, let's only collect some log first

PR #2466, opened 2023-10-13
🐛 [RUM-1360] Cap INP outliers

amortemousque:
## Motivation

Fix 2 issues about INP:
- Some of the [PerformanceEventTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEventTiming) duration are outliers. So now we CAP INP values to 1 min maximum.
- Some of the [PerformanceEventTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEventTiming) reported on one view actually started on the previous view. We now attach them to the previous view.

## Testing

- [ ] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
@@ -6,7 +6,7 @@ import {
   addTelemetryDebug,
   elapsed,
 } from '@datadog/browser-core'
-import type { ClocksState, Duration } from '@datadog/browser-core'
+import type { Duration, RelativeTime } from '@datadog/browser-core'
 import { RumPerformanceEntryType, supportPerformanceTimingEvent } from '../../../browser/performanceCollection'
 import type { RumFirstInputTiming, RumPerformanceEventTiming } from '../../../browser/performanceCollection'
 import { LifeCycleEventType } from '../../lifeCycle'
@@ -15,6 +15,7 @@ import { ViewLoadingType } from '../../../rawRumEvent.types'
 import { getSelectorFromElement } from '../../getSelectorFromElement'
 import { isElementNode } from '../../../browser/htmlDomUtils'
 import type { RumConfiguration } from '../../configuration'
+import type { ViewEndedEvent } from '../trackViews'
 import { getInteractionCount, initInteractionCountPolyfill } from './interactionCountPolyfill'
 
 // Arbitrary value to prevent unnecessary memory usage on views with lots of interactions.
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -164,8 +175,25 @@ function trackLongestInteractions(getViewInteractionCount: () => number) {
...
+      if (!stopped) {
+        computeViewInteractionCount()
+      }
+
+      return viewInteractionCount
+    },
+    stopViewInteractionCount: () => {
+      computeViewInteractionCount()
+      stopped = true
+    },
```
BenoitZugmeyer, line 198: **💬 suggestion:** having a typed "state" would make the code a bit more explicit (the link between `viewInteractionCount` and `stopped` would be clearer and `computeViewInteractionCount` would not have sideeffect):

```ts
  type State = { stopped: false } | { stopped: true, interactionCount: number }
  let state: State = { stopped: false }

  function computeViewInteractionCount() {
    return getInteractionCount()! - previousInteractionCount
  }

  return {
    getViewInteractionCount: () => {
      if (state.stopped) {
        return state.interactionCount
      }
      return computeViewInteractionCount()
    },
    stopViewInteractionCount: () => {
      state = { stopped: true, interactionCount: computeViewInteractionCount() }
    },
```

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -46,7 +47,14 @@ export function trackInteractionToNextPaint(
...
   }
 
-  const { getViewInteractionCount } = trackViewInteractionCount(viewLoadingType)
+  const { getViewInteractionCount, stopViewInteractionCount } = trackViewInteractionCount(viewLoadingType)
+
+  let viewEnd = Infinity as RelativeTime
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_ENDED, (data: ViewEndedEvent) => {
+    viewEnd = data.endClocks.relative
+    stopViewInteractionCount()
+  })
```
BenoitZugmeyer, line 58: **💭 thought:** ‏using the `VIEW_ENDED` event here is a bit confusing, because in some case `trackInteractionToNextPaint` is "driven" by the caller, and in some case it "drives" itself. Maybe a `setViewEnd(RelativeTime)` (to mimic the `setLoadEvent` logic) would be nicer, and we could remove the `LifeCycle` dependency when tackling https://datadoghq.atlassian.net/browse/RUM-958

BenoitZugmeyer (approved):

PR #2470, opened 2023-10-18
[RUM-1596] Replay event bridge POC

amortemousque:
## Motivation

In the context of webview support for session replay the Broser SDK needs to send Replay record through the event bridge

## Changes

Send Replay record through the event bridge

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/logs/src/transport/startLogsBridge.ts
```diff
@@ -8,6 +8,6 @@ export function startLogsBridge(lifeCycle: LifeCycle) {
   const bridge = getEventBridge<'log', LogsEvent>()!
 
   lifeCycle.subscribe(LifeCycleEventType.LOG_COLLECTED, (serverLogsEvent: LogsEvent & Context) => {
-    bridge.send('log', serverLogsEvent)
+    bridge.send('log', serverLogsEvent, serverLogsEvent.view.id)
```
bcaudan, line 11: **❓ question:** ‏do we need to pass view id when the event is not a record?

packages/rum/src/domain/startRecordBridge.ts
```diff
@@ -0,0 +1,17 @@
+import { getEventBridge } from '@datadog/browser-core'
+import type { ViewContexts } from '@datadog/browser-rum-core'
+import type { BrowserRecord } from '../types'
+
+export function startRecordBridge(viewContexts: ViewContexts) {
+  const bridge = getEventBridge<'record', BrowserRecord>()!
+
+  return {
+    addRecord: (record: BrowserRecord) => {
+      const view = viewContexts.findView()
```
bcaudan, line 10: **❓ question:** ‏should not we get the view corresponding to the record timestamp?

packages/rum/src/domain/startRecordBridge.ts
```diff
@@ -0,0 +1,17 @@
...
+      if (!view) {
+        return
+      }
```
bcaudan, line 13: **❓ question:** ‏How can it happen?

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -112,18 +113,34 @@ describe('rum public api', () => {
...
         rumPublicApi.init(invalidConfiguration as RumInitConfiguration)
         expect(rumPublicApi.getInitConfiguration()?.sessionSampleRate).toEqual(100)
       })
 
+      it('should set the default privacy level received from the bridge if the not provided in the init configuration', () => {
+        const hybridInitConfiguration: HybridInitConfiguration = {}
+        rumPublicApi.init(hybridInitConfiguration as RumInitConfiguration)
+        expect((rumPublicApi.getInitConfiguration() as RumInitConfiguration)?.defaultPrivacyLevel).toEqual(
+          PRIVACY_LEVEL_FROM_EVENT_BRIDGE
+        )
```
bcaudan, line 133: **💬 suggestion:** ‏What about passing the bridge privacy level to the `initEventBridgeStub`?
It would avoid to look into the stub implementation to understand where this value come from

packages/rum/src/domain/startRecordBridge.ts
amortemousque: It should never happen. It was to please the typecheck but I can use `viewContexts.findView()!` if you prefer?

packages/rum/src/boot/recorderApi.ts
```diff
@@ -53,7 +53,7 @@ export function makeRecorderApi(
   startRecordingImpl: StartRecording,
   createDeflateWorkerImpl?: CreateDeflateWorker
 ): RecorderApi {
-  if (canUseEventBridge() || !isBrowserSupported()) {
+  if (!isBrowserSupported()) {
```
BenoitZugmeyer, line 56: **❓ question:** ‏Just to confirm, outdated versions of Mobile SDKs bridges will behave correctly when sending records, right? They won't send them as RUM events?

packages/logs/src/transport/startLogsBridge.ts
bcaudan: If we don't need it, let's not do it.

packages/rum/src/boot/recorderApi.ts
amortemousque: Very good question! Marius told me it should be good. But we have to do proper tests (also for IOS). I did not add a feature flag because we should be able to test it soon. But I won't merge before testing ;)

packages/rum/src/domain/startRecordBridge.ts
bcaudan: Yes, in case it happens, I'd prefer to see a telemetry error than to not be aware of it.

packages/core/src/transport/eventBridge.spec.ts
```diff
@@ -45,11 +47,45 @@ describe('getEventBridge', () => {
...
+  let eventBridgeStub: DatadogEventBridge
+  const bridgePrivacyLevel = DefaultPrivacyLevel.MASK
+
+  beforeEach(() => {
+    eventBridgeStub = initEventBridgeStub({ privacyLevel: bridgePrivacyLevel })
+  })
+
+  afterEach(() => {
+    deleteEventBridgeStub()
+  })
```
bcaudan, line 77: **♻️ opportunity:** use registerCleanupTask for the event bridge stub‏

packages/rum/src/boot/startRecording.ts
```diff
@@ -23,14 +25,21 @@ export function startRecording(
...
 
-  const { addRecord, stop: stopSegmentCollection } = startSegmentCollection(
-    lifeCycle,
-    configuration,
-    sessionManager,
-    viewContexts,
-    replayRequest,
-    encoder
-  )
+  let addRecord = (_record: BrowserRecord) => {}
```
bcaudan, line 28: **💬 suggestion:** unused code
‏
```suggestion
  let addRecord
```

packages/rum/src/boot/startRecording.ts
```diff
@@ -23,14 +25,21 @@ export function startRecording(
...
+  let stopSegmentCollection = noop
```
bcaudan, line 29: **🥜 nitpick:** ‏to avoid this noop, we could setup a cleanupTasks array like in startRum

test/e2e/scenario/eventBridge.scenario.ts
```diff
@@ -91,4 +91,14 @@ describe('bridge present', () => {
       expect(intakeRegistry.logsEvents.length).toBe(1)
       expect(intakeRegistry.hasOnlyBridgeRequests).toBe(true)
     })
+
+  createTest('send records to the bridge')
```
bcaudan, line 95: **💬 suggestion:** what about adding a test where to ensure that in bridge mode stop / start recording also work as expected?
‏

packages/rum/src/boot/recorderApi.ts
amortemousque: I've added a feature detection to only send records when the bridge supports it. https://github.com/DataDog/browser-sdk/pull/2470/commits/e05cc186cc50ffebcd1c51c00ca1c45180b254f0
Let me know what you think

packages/rum/src/domain/startRecordBridge.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function startRecordBridge(viewContexts: ViewContexts) {
+  const bridge = getEventBridge<'record', BrowserRecord>()!
+
+  return {
+    addRecord: (record: BrowserRecord) => {
+      // Get the current active view, not at the time of the record, aligning with the segment logic.
+      // This approach could potentially associate the record to an incorrect view, in case the record date is in the past (e.g. frustration records).
+      // However the risk is minimal. We could address the issue when potential negative impact are identified.
+      const view = viewContexts.findView()!
```
bcaudan, line 13: **💭 thought:** ‏In order to align with segment logic, what about:
- maintaining a "current context" for the record bridge, attached to the events
- creating it with the first added record
- clearing it on VIEW_CREATED or PAGE_EXITED

It would avoid to touch the curent logic around view contexts and view lifecycle.

packages/rum/src/boot/startRecording.ts
```diff
@@ -23,26 +27,34 @@ export function startRecording(
...
+  if (!canUseEventBridge()) {
+    ;({ addRecord, stop: stopSegmentCollection } = startSegmentCollection(
+      lifeCycle,
+      configuration,
+      sessionManager,
+      viewContexts,
+      replayRequest,
+      encoder
+    ))
+    cleanupTasks.push(() => stopSegmentCollection())
```
BenoitZugmeyer, line 42: **🥜 nitpick:** ‏
```suggestion
  if (!canUseEventBridge()) {
    const segmentCollection = startSegmentCollection(
      lifeCycle,
      configuration,
      sessionManager,
      viewContexts,
      replayRequest,
      encoder
    ))
    addRecord = segmentCollection.addRecord
    cleanupTasks.push(segmentCollection.stop)
```

packages/rum/src/boot/startRecording.ts
```diff
@@ -23,26 +27,34 @@ export function startRecording(
...
+    ;({ addRecord } = startRecordBridge(viewContexts))
+  }
 
   const { stop: stopRecording } = record({
     emit: addRecord,
     configuration,
     lifeCycle,
     viewContexts,
   })
+  cleanupTasks.push(() => stopRecording())
```
BenoitZugmeyer, line 53: **🥜 nitpick:** ‏
```suggestion
  cleanupTasks.push(stopRecording)
```

test/e2e/scenario/eventBridge.scenario.ts
```diff
@@ -101,4 +101,34 @@ describe('bridge present', () => {
...
+        window.DD_RUM!.stopSessionReplayRecording()
+
+        // trigger a new record
+        document.body.appendChild(document.createElement('li'))
+      })
+
+      await flushEvents()
+
+      const postStopRecordsCount = intakeRegistry.replayRecords.length - preStopRecordsCount
+      expect(postStopRecordsCount).toEqual(0)
```
BenoitZugmeyer, line 132: **❓ question:** ‏I'm not sure to see why we need all this in this test? Isn't something like this enough?

```ts
  createTest('should only send records when the recording is started')
    .withRum()
    .withEventBridge()
    .run(async ({ intakeRegistry }) => {
      await flushEvents()

      expect(intakeRegistry.replayRecords.length).toBeGreaterThan(0)
      expect(intakeRegistry.hasOnlyBridgeRequests).toBe(true)
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -32,6 +32,11 @@ export function getEventBridge<T, E>() {
   }
 }
 
+export function isBridgeForRecordsSupported(): boolean {
+  const bridge = getEventBridgeGlobal()
+  return !!bridge && 'getPrivacyLevel' in bridge
```
BenoitZugmeyer, line 37: **❓ question:** ‏Did we discuss with mobile if it could be worth having a proper solution for that? Ex: exposing a Mobile SDK version, or "capabilities"?
I'm fine with keeping this in that PR / while on beta, but it would be great to have a better approach

packages/rum/src/domain/startRecordBridge.ts
bcaudan: Maybe a "record context" module instead, replacing the segment context and used to get the view information for replay stats and bridge.

test/e2e/scenario/eventBridge.scenario.ts
amortemousque: The test also test that no records are sent through the bridge if session replay is stopped. I used startSessionReplayRecordingManually: true to be more explicite on what is the code doing. But I could simplify like so I guess
```suggestion
    .withRum()
    .withEventBridge()
    .run(async ({ intakeRegistry }) => {
      // wait for recorder to be properly started
      await browser.pause(200)

      expect(intakeRegistry.replayRecords.length).toBeGreaterThan(0)
      expect(intakeRegistry.hasOnlyBridgeRequests).toBe(true)

      const preStopRecordsCount = intakeRegistry.replayRecords.length
      await browserExecute(() => {
        window.DD_RUM!.stopSessionReplayRecording()

        // trigger a new record
        document.body.appendChild(document.createElement('li'))
      })

      await flushEvents()

      const postStopRecordsCount = intakeRegistry.replayRecords.length - preStopRecordsCount
      expect(postStopRecordsCount).toEqual(0)
```
Wdyt?

packages/core/src/transport/eventBridge.ts
amortemousque: The idea would be for mobile to provide a method like `bridgeSupports('feature_name')`?
No it has not been discussed yet, but it is true that it could be nice to anticipate potential future breaking changes.

test/e2e/scenario/eventBridge.scenario.ts
BenoitZugmeyer: Mmh I don't know if it's truly necessary, as this behavior is unrelated to the event bridge?

If you think it is, you could add another test case so it's more explicit

packages/core/src/transport/eventBridge.ts
BenoitZugmeyer: Yes, or `getCapablities().includes("feature_name")`.

test/e2e/scenario/eventBridge.scenario.ts
amortemousque: Since on `stopSessionReplayRecording`  the segmentCollection is stopped and not the bridge I prefer to check it. I'll make a separated test

packages/rum/src/domain/startRecordBridge.ts
amortemousque: Hum, the segment context is a concept shared with Replay that leaves in the [event format](https://github.com/DataDog/rum-events-format/blob/389581be98dcf8efbfcfe7bffaa32d53f960fb6f/lib/cjs/generated/browserSessionReplay.d.ts#L309). I think it might be confusing to replace it with a "record context"

amortemousque: Resolved: https://github.com/DataDog/browser-sdk/pull/2598

BenoitZugmeyer (approved):

packages/rum-core/src/boot/preStartRum.spec.ts
```diff
@@ -115,12 +112,26 @@ describe('preStartRum', () => {
...
       })
 
-      it('init should force session sample rate to 100', () => {
+      it('should force session sample rate to 100', () => {
         const invalidConfiguration: HybridInitConfiguration = { sessionSampleRate: 50 }
         strategy.init(invalidConfiguration as RumInitConfiguration)
         expect(strategy.initConfiguration?.sessionSampleRate).toEqual(100)
       })
 
+      it('should set the default privacy level received from the bridge if the not provided in the init configuration', () => {
```
bcaudan, line 121: **💬 suggestion:** ‏What about adding a test when default privacy level is not provided by the configuration nor the bridge?

packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -65,7 +72,7 @@ export function startRumSessionManager(
 export function startRumSessionManagerStub(): RumSessionManager {
   const session: RumSession = {
     id: '00000000-aaaa-0000-aaaa-000000000000',
-    sessionReplayAllowed: false,
+    sessionReplayAllowed: bridgeSupports(BridgeCapability.RECORDS),
```
bcaudan, line 75: **👏 praise:** ‏nice!

bcaudan (approved):

PR #2477, opened 2023-10-23
🔊 Add CLS target selector telemetry

amortemousque:
## Motivation

The CLS target selector is computed  only when the CLS score increases.  It means that by the time the computation happens the target node might have been removed from the DOM making the selector computation fails.

This PR collects telemetry to assess if we could compute the selector each time the target is updated in the current CLS window.

This PR collects:
- The number of time the CLS target is updated in a CLS window
- The computation time of the target selector (`getSelectorFromElement`)

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
@@ -73,7 +73,11 @@ export function trackCumulativeLayoutShift(
           const clsTarget = window.largestLayoutShiftTarget()
           let cslTargetSelector
 
-          if (isExperimentalFeatureEnabled(ExperimentalFeature.WEB_VITALS_ATTRIBUTION) && clsTarget) {
+          if (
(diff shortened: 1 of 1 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -103,9 +117,18 @@ function slidingSessionWindow() {
         entry.startTime - startTime >= 5 * ONE_SECOND
       if (shouldCreateNewWindow) {
         startTime = endTime = entry.startTime
+        if (startTime !== undefined && maxTargetUpdateTelemetry) {
+          maxTargetUpdateTelemetry--
+          addTelemetryDebug('CLS target updates per CLS window', {
+            dates: targetUpdates,
```
amortemousque, line 123: This array allow to check the frequency of the update within a window

bcaudan (approved):

PR #2480, opened 2023-10-27
🐛 [RUM-1561] Fix CLS selector computation on detached node

amortemousque:
## Motivation

The CLS target selector is computed only when the CLS score increases. It means that by the time the computation happens the target node might have been removed from the DOM making the selector computation fails.

## Changes

- Check that the node is attached before computing the CLS CSS selector
- Remove related telemetry

## Testing

- [X] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
@@ -1,14 +1,5 @@
-import {
-  round,
-  type RelativeTime,
-  find,
-  ONE_SECOND,
-  isExperimentalFeatureEnabled,
-  ExperimentalFeature,
-  noop,
-  addTelemetryDebug,
-  relativeNow,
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -76,19 +63,10 @@ export function trackCumulativeLayoutShift(
           if (
             isExperimentalFeatureEnabled(ExperimentalFeature.WEB_VITALS_ATTRIBUTION) &&
             clsTarget &&
-            clsTarget.parentElement
+            // Check if the CLS target have been removed from the DOM between the time we collect the target reference and when we compute the selector
+            clsTarget.isConnected
```
amortemousque, line 67: The target might have a `parentElement` and not be connected. It can happen if one of the parents has been detached.

bcaudan (approved):

PR #2486, opened 2023-10-30
🐛 [RUM-1863] fix iOS webview detection 

BenoitZugmeyer:
## Motivation

The fix we introduced in #2440 does not apply in iOS webviews as the user agent don't mention "Safari".

## Changes

Adjust `isSafari` to also match iOS webviews.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


bcaudan (approved):
Nice!

PR #2497, opened 2023-11-06
⚡️[RUM-2017] optimize cookie accesses

BenoitZugmeyer:
## Motivation

We observed that the SDK is spending a significant amount of time reading from `document.cookie` during initialization.

## Changes

Introduce a alternative function to `getCookie` that caches cookies once,  and use it where it makes sense.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/cookie.ts
+++ b/packages/core/src/browser/cookie.ts
@@ -26,7 +26,7 @@ let initCookieParsed: Map<string, string> | undefined
 
 /**
  * Returns a cached value of the cookie. Use this during SDK initialization (and whenever possible)
- * to avoid parsing the document.cookie multiple times.
+ * to avoid accessing document.cookie multiple times.
  */
 export function getInitCookie(name: string) {
   if (!initCookieParsed) {
--- a/packages/core/src/browser/cookie.ts
+++ b/packages/core/src/browser/cookie.ts
@@ -1,6 +1,6 @@
 import { display } from '../tools/display'
 import { ONE_MINUTE, ONE_SECOND } from '../tools/utils/timeUtils'
-import { findCommaSeparatedValue, generateUUID } from '../tools/utils/stringUtils'
+import { findCommaSeparatedValue, findCommaSeparatedValues, generateUUID } from '../tools/utils/stringUtils'
 
 export interface CookieOptions {
   secure?: boolean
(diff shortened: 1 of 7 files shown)


packages/core/src/browser/cookie.ts
```diff
@@ -22,6 +22,23 @@ export function getCookie(name: string) {
   return findCommaSeparatedValue(document.cookie, name)
 }
 
+let initCookieParsed: Map<string, string> | undefined
+
+/**
+ * Returns a cached value of the cookie. Use this during SDK initialization (and whenever possible)
+ * to avoid parsing the document.cookie multiple times.
```
bcaudan, line 29: **🥜 nitpick:** ‏
```suggestion
 * to avoid accessing the document.cookie multiple times.
```

packages/core/src/domain/session/oldCookiesMigration.ts
```diff
@@ -17,11 +17,11 @@ export const LOGS_SESSION_KEY = 'logs'
  * to allow older sdk versions to be upgraded to newer versions without compatibility issues.
  */
 export function tryOldCookiesMigration(cookieStoreStrategy: SessionStoreStrategy) {
-  const sessionString = getCookie(SESSION_STORE_KEY)
+  const sessionString = getInitCookie(SESSION_STORE_KEY)
```
bcaudan, line 20: **💭 thought:** ‏‏for a npm setup with logs and RUM, the init cache is shared.
So in this case, by looking at the current session in the init cookie, the migration is now happening for both SDKs.
It seems to be ok but maybe it could worth adding an extra test for it.

packages/core/src/tools/utils/stringUtils.ts
```diff
@@ -9,10 +9,36 @@ export function generateUUID(placeholder?: string): string {
...
+    const match = COMMA_SEPARATED_KEY_VALUE.exec(rawString)
+    if (match) {
+      if (match[1] === name) {
+        return match[2]
+      }
+    } else {
+      break
+    }
+  }
+}
```
bcaudan, line 27: **❓ question:** ‏what was the reasoning for revisiting this function?

BenoitZugmeyer: I didn't want to implement two different ways of parsing CSV, and the initial function implementation didn't allow to parse all values at once. Also it was a bit slower as a new regex needed to be created every time.

bcaudan (approved):

Issue #2471, opened 2023-10-19
🙏 Configuring Datadog RUM Proxy Enviroment


minJaebak:
Hi

I'm trying to configure Datadog RUM Proxy, setting up the RUM SDK in a Vue.js application and configuring an Nginx proxy server to pass RUM data to Datadoghq. I've referred to the RUM Proxy Doc (https://docs.datadoghq.com/ko/real_user_monitoring/guide/proxy-rum-data/?tab=npm), but I'm unable to confirm session collection in the Datadog console.

Environment Setup:

Datadog RUM SDK version: 4.50.1
Nginx version: 1.25.0

`export const datadogRumInit = {
  applicationId: import.meta.env.VITE_DATADOG_APPLICATION_ID,
  clientToken: import.meta.env.VITE_DATADOG_CLIENT_TOKEN,
  site: 'datadoghq.com',
  service: import.meta.env.VITE_DATADOG_SERVICE,
  env: import.meta.env.VITE_DATADOG_ENV,
  version: import.meta.env.VITE_DATADOG_VERSION,
  sessionSampleRate: 100,
  sessionReplaySampleRate: 100,
  trackInteractions: true,
  defaultPrivacyLevel: 'mask-user-input',
  trackUserInteractions: true,
  trackResources: true,
  trackFrustrations: true,
  trackLongTasks: true,
  proxyUrl: <proxy url>,
  allowedTracingOrigins: <xxx>,
};

export const datadogLogInit = {
  clientToken: import.meta.env.VITE_DATADOG_CLIENT_TOKEN,
  site: 'datadoghq.com',
  forwardErrorsToLogs: true,
  sessionSampleRate: 100,
};`

Nginx Configuration (nginx.conf):
`http {

    client_body_buffer_size     10M;
    client_max_body_size        10M;

    #proxy_ssl_trusted_certificate /etc/ssl/certs/ca-bundle.crt;
    # Datadog Proxy
    server {
        listen 3834;

        # Proxy for RUM
        location /health {
            return 200;
        }

        location / {
            resolver 8.8.8.8 valid=300s ipv6=off;

            if (\$request_method = 'OPTIONS') {
                add_header 'Access-Control-Allow-Origin' '*';
                add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS, HEAD';
[...]

PR #2502, opened 2023-11-14
✨[RUM-2158] Allow more flexible proxy URL

bcaudan:
## Motivation

As discussed in #2429 and #2471, the [current proxy strategy](https://docs.datadoghq.com/real_user_monitoring/guide/proxy-rum-data/?tab=npm#proxy-setup) with the `ddforward` parameter, storing the target path and parameters, can be challenging to implement.

A solution allowing to have the path and parameters appended to the provided proxy URL was initially discarded since one of our URL pattern was already blocked, cf [easylist](https://github.com/easylist/easylist/blob/997fb6533c719a015c21723b34e0cedefcc0d83d/easyprivacy/easyprivacy_general.txt#L3840).
However, some customers may want to use a proxy regardless of the privacy blocking or could want to tweak themselves the final proxy url to prevent the blocking.

## Changes

Allow to pass a function to the `proxy` init parameter:
```ts
init({
  ...
  proxy: (options: {path: string, parameters: string}) => 
    `https://proxy.io/prefix${options.path}/suffix?${options.parameters}`
})
```
With:
* path: `/api/vX/product`
* parameters: `xxx=yyy&zzz=aaa`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -32,7 +32,7 @@ export interface InitConfiguration {
   storeContextsAcrossPages?: boolean | undefined
 
   // transport options
-  proxy?: string | undefined
+  proxy?: string | ProxyFn | undefined
   site?: string | undefined
 
   // tag and context options
@@ -49,6 +49,7 @@ export interface InitConfiguration {
   enableExperimentalFeatures?: string[] | undefined
   replica?: ReplicaUserConfiguration | undefined
   datacenter?: string
+  // TODO next major: remove this option and replace usages by proxyFn
   internalAnalyticsSubdomain?: string
 
   telemetryConfigurationSampleRate?: number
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -58,6 +59,12 @@ export interface InitConfiguration {
 // for this option.
 type GenericBeforeSendCallback = (event: any, context?: any) => unknown
 
+/**
+ * path: /api/vX/product
+ * parameters: xxx=yyy&zzz=aaa
+ */
+type ProxyFn = (path: string, parameters: string) => string
```
BenoitZugmeyer, line 66: **💬 suggestion:** ‏A single argument object with `path` and `parameters` properties could be more future-proof

```suggestion
type ProxyFn = (options: { path: string, parameters: string }) => string
```

bcaudan: Do you have changes in mind that would be easier to handle with this approach?

BenoitZugmeyer: It could help if we wanted to have more fields or deprecate some...

bcaudan: hum, looking at our current APIs, it seems that we are mostly using parameters.
We've moved from a parameter to an options object for [startView](https://github.com/DataDog/browser-sdk/pull/1448/files#diff-63186209ba3af0e1cccdc814069d547e89e9ff71e4dbc592b7cd210dbc4081ea) but from my understanding it was more for handling the fact that all parameters are optionals, which is not the case here.

I'd be in favour of being consistent in our API signatures, so if we want to move toward this pattern, it could be nice to give it a bit more thoughts.
Let's discuss it more.

BenoitZugmeyer (approved):

PR #2510, opened 2023-11-23
✨ Add disposition to RawReport to distinguish CSP types

blchelle:
## Motivation

My team recently integrated DD RUM into our frontend react application. Since the integration, we have been bombarded with over 200k `csp_violation` errors per day.

I recently discovered that the violations were firing due to a `Content-Security-Policy-Report-Only` header sent by Cloudflare Page Shield. Here are some [docs](https://developers.cloudflare.com/page-shield/reference/csp-header/) for Page Shield for more context.

We want to ignore csp violations triggered by a `Content-Security-Policy-Report-Only` header, but we want to still want to show violations triggered by a `Content-Security-Policy` header. The `securitypolicyviolation` event exposes a [disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only#disposition) that informs whether the violation was triggered by a `Content-Security-Policy` or a `Content-Security-Policy-Report-Only` header.

We can't use [beforeSend](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/domain/configuration/configuration.ts#L26) for this because [RawReport](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/domain/report/reportObservable.ts#L18-L23) does not expose a `disposition` field. Therefore, we have no way to distinguish between violations of our actual CSP, and violations of the Cloudflare report-only CSP.

## Changes

* Updated [buildRawReportFromCspViolation()](https://github.com/DataDog/browser-sdk/blob/master/packages/core/src/domain/report/reportObservable.ts#L86) to include the disposition of the `securitypolicyviolation` event. 
* Updated the [RawReport](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/domain/report/reportObservable.ts#L18-L23) type to optionally include the `disposition` field, since it will only apply to CSP violations.

## Testing

* Modified an existing unit test for handling csp violations to expect the `disposition` field.

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan:
Hi @blchelle,

 Indeed having a way to filter on `disposition` from `beforeSend` would be convenient.

Instead of attaching this field to the `RawError`, it could be interesting to propagate the original report event to the beforeSend, in case other properties could be valuable.
In any case, there would be some extra mapping to do in order to have it accessible in [RUM beforeSend](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/assembly.ts#L191), you should have a look at [trackReportError](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/error/trackReportError.ts#L9) and [errorCollection]( https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/error/errorCollection.ts#L124).

> Please advise on how I can setup a local/staging environment to validate my changes.

`yarn dev` builds sdk files with your local changes, you can then tweak the [sandbox page on localhost:8080](https://github.com/DataDog/browser-sdk/blob/main/sandbox/index.html) to test your changes or if possible, locally swap the SDK setup on your app with this local build. 
(note to myself, we should add some details about that in our contributing guidelines)

> I briefly looked through the e2e test suite and I couldn't find anywhere this would slot in nicely, please let me know if you have any ideas.

We don't have an e2e test suite for that but it may be interesting to at least have high level unit test for the different beforeSend domain context 🤔
We could add those in a separate PR.

blchelle:
Hi @bcaudan

I think it makes a lot of sense to propagate the original report up to `beforeSend`. Thank you for pointing me in the right direction to add the additional mappings. I've updated the pull request to reflect these changes. Looking forward to your review!

amortemousque:
Hello @blchelle ,
Thanks a lot for the contribution! I took hover the PR and made some adjustment to be ready.

bcaudan (approved):

bcaudan:
Available with v5.5.0.

PR #2515, opened 2023-11-29
🔊[RUM-2324] Telemetry on LCP with startTime to 0

amortemousque:
## Motivation

Collect telemetry on LCP with startTime to 0:
```js
{
  entry // the entry reported by the performanceObserver API
  previousNonZeroLcp,
  navigationStart // check if there is an unexpected navigationStart
  timeOrigin // check if there is an unexpected timeOrigin
  now: performance.now() // check the diff between now and the entry timings,
  visibilityState // check if these reports happens when "hidden" | "visible" | "prerender",
  prerendering: document.prerendering,
}
```

## Changes

Add telemetry on LCP with a startTime at 0

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
@@ -28,6 +28,7 @@ export interface LargestContentfulPaint {
 
 let zeroLcpReported = false
 let previousNonZeroLcp: Omit<PerformanceEntry, 'toJSON'> | undefined
+
 /**
  * Track the largest contentful paint (LCP) occurring during the initial View.  This can yield
  * multiple values, only the most recent one should be used.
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -61,6 +72,22 @@ export function trackLargestContentfulPaint(
           lcpTargetSelector = getSelectorFromElement(lcpEntry.element, configuration.actionNameAttribute)
         }
 
+        if (
+          !zeroLcpReported &&
+          lcpEntry.startTime === 0 &&
+          isExperimentalFeatureEnabled(ExperimentalFeature.ZERO_LCP_TELEMETRY)
+        ) {
+          zeroLcpReported = true
+          addTelemetryDebug('LCP with startTime = 0', {
```
BenoitZugmeyer, line 81: **💭 thought:** ‏I wonder if it could make sense to log whether an anterior LCP entry was notified with a non 0 startTime

BenoitZugmeyer (approved):

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
amortemousque: Yes, it could be nice to know. I added it

PR #2520, opened 2023-12-01
⚗️[Logs] POC dependency injection

bcaudan:
## Motivation

Play with a dependency injection approach to see how it could look like 

## Changes

- bootstrap simple dependency injector
- use it for logs components
- re-use it for unit tests

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/logs/src/boot/startLogs.ts
```diff
@@ -1,155 +1,56 @@
...
-      reportError,
-      pageExitObservable,
-      session.expireObservable
-    )
-    cleanupTasks.push(() => stopLogsBatch())
-  } else {
-    startLogsBridge(lifeCycle)
+  if (!configuration.sessionStoreStrategyType || willSyntheticsInjectRum()) {
+    injector.override(startLogsSessionManager, startLogsSessionManagerStub)
   }
```
BenoitZugmeyer, line 33: Note: those "override" in the wilde could probably be moved in components, ex:

```ts
const startLogsTransport: Component<...> = (injector) => {
    if (canUseEventBridge()) {
       return injector.run(startLogsBridge)
    } else {
       return injector.run(startLogsBatch)
    }
}
```

packages/core/src/tools/injector.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function createInjector(): Injector {
+  const instances = new Map<AnyComponent<any>, any>()
+  const overrides = new Map<AnyComponent<any>, AnyComponent<any>>()
+  overrides.set(getInjector, () => injector)
+
+  const injector: Injector = {
+    run(component) {
+      if (instances.has(component)) {
+        throw new Error(`Component ${component.name} already started`)
```
BenoitZugmeyer, line 27: Note: This assertion and the following ones (in `get` and `override`) are not strictly needed, but helps to debug weird injection cases.

PR #2527, opened 2023-12-11
🐛 [RUM-2280] fix duplicated mutations when using Shadow DOM

BenoitZugmeyer:
## Motivation

Shadow Roots need to be instrumented with a `MutationObserver` and an `input` event listener to properly generate incremental snapshots for changes happening inside of them. We do this instrumentation during serialization.

The issue is, when serializing the same Shadow Root multiple times (ex:  when a route_change view is created and a second Full Snapshot is taken), Shadow Roots are instrumented multiple times, so inner incremental snapshots are generated multiple times whenever a single change is done.

## Changes

Make sure we instrument Shadow Roots only once

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/record.spec.ts
+++ b/packages/rum/src/domain/record/record.spec.ts
@@ -311,6 +311,22 @@ describe('record', () => {
       expect(innerMutationData.isChecked).toBe(true)
     })
 
+    it('should record the change event inside a shadow root only once, regardless if the DOM is serialized multiple times', () => {
+      const radio = appendElement('<input type="radio"/>', createShadow()) as HTMLInputElement
+      startRecording()
+
+      recordApi.takeSubsequentFullSnapshot()
+
+      radio.checked = true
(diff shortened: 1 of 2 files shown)


packages/rum/src/domain/record/record.spec.ts
```diff
@@ -311,6 +311,22 @@ describe('record', () => {
       expect(innerMutationData.isChecked).toBe(true)
     })
 
+    it('should record the change event inside a shadow root only once, regardless if the DOM is serialized multiple times', () => {
```
BenoitZugmeyer, line 314: This test is on `input` events, as it is easier to test. Mutations are asynchronous and since the original "Shadow Root Controller" was replaced by a new one, calling `flushMutations()` did not flush both of them.

Anyway, one test should be enough to avoid regression.

amortemousque (approved):

ThibautGeriz (approved):
LGTM

bcaudan (approved):

PR #2529, opened 2023-12-14
🔉 [RUM-253] adjust hevy customer data warning when compression is enabled

BenoitZugmeyer:
## Motivation

We are currently displaying a warning in the console when a large context (global context, user, log context) or a lot of feature flag evaluations are set. This is to encourage users to keep their setup lean to avoid impacting the bandwidth too much.

In #2500 we introduced a new initialization parameter to enable intake request compression. With this option enabled, we are able to send more data without impacting the bandwidth too much. Thus, we can now adapt the limit we currently use to display that warning.

## Changes

* Factorize our customer data tracking implementation
* Adjust the customer data byte count limit when compression is enabled.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/context/trackCustomerData.ts
amortemousque, line 1: **🥜 nitpick:** ‏ ‏I was expecting something more like `customerDataTracker` for the filename

packages/core/src/domain/context/trackCustomerData.ts
```diff
@@ -0,0 +1,93 @@
...
+  [CustomerDataType.User]: 'user',
+  [CustomerDataType.GlobalContext]: 'global context',
+  [CustomerDataType.LoggerContext]: 'logger context',
+}
+
+function displayCustomerDataLimitReachedWarning(customerDataType: CustomerDataType, bytesCountLimit: number) {
+  display.warn(
+    `The ${CustomerDataLabel[customerDataType]} data exceeds the recommended ${
+      bytesCountLimit / ONE_KIBI_BYTE
+    }KiB threshold. More details: https://docs.datadoghq.com/real_user_monitoring/browser/troubleshooting/#customer-data-exceeds-the-recommended-3kib-warning`
```
amortemousque, line 91: **💬 suggestion:** ‏I don't know it you already have it in mind but we should probably revisit the documentation to take the compression limit into account and also advice customer to enable the compression if they haven't already.

packages/core/src/domain/context/trackCustomerData.ts
```diff
@@ -0,0 +1,93 @@
...
+  let alreadyWarned = false
+
+  // Throttle the bytes computation to minimize the impact on performance.
+  // Especially useful if the user call context APIs synchronously multiple times in a row
+  const { throttled: computeBytesCountThrottled, cancel: cancelComputeBytesCount } = throttle((context: Context) => {
+    bytesCountCache = computeBytesCount(jsonStringify(context)!)
+    maybeWarn()
+  }, BYTES_COMPUTATION_THROTTLING_DELAY)
+
+  function maybeWarn() {
```
amortemousque, line 42:
```suggestion
  function mayWarn() {
```

packages/core/src/domain/context/trackCustomerData.ts
BenoitZugmeyer: Yes, I was planning to update the docs too.

amortemousque: Nice, so the link might also change as it includes the limit "#customer-data-exceeds-the-recommended-**3kib**-warning"

packages/core/src/domain/context/customerDataTracker.ts
```diff
@@ -0,0 +1,93 @@
...
+export type CustomerDataTracker = ReturnType<typeof createCustomerDataTracker>
+export const enum CustomerDataCompressionStatus {
+  Unknown,
+  Enabled,
+  Disabled,
+}
+
+export function createCustomerDataTracker(
+  type: CustomerDataType,
+  compressionStatus = CustomerDataCompressionStatus.Disabled
```
bcaudan, line 30: **💭 thought:** ‏FMU it is the current behaviour but would not it worth having a single count for all customer data rather than one by type?
for example, currently customer data could be FF 2k + User 2k + global 2k and no warn would be generated, right?

packages/core/src/domain/context/contextManager.ts
```diff
@@ -44,6 +42,7 @@ export function createContextManager(customerDataTracker: CustomerDataTracker) {
       changeObservable.notify()
     },
 
+    customerDataTracker,
```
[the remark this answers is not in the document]
BenoitZugmeyer: It was a bit challenging, but I implemented your suggestion.

amortemousque (approved):

bcaudan (approved):

packages/core/src/domain/context/customerDataTracker.ts
```diff
@@ -0,0 +1,134 @@
...
+  let alreadyWarned = false
+  function checkCustomerDataLimit(initialBytesCount = 0) {
+    if (alreadyWarned || compressionStatus === CustomerDataCompressionStatus.Unknown) {
+      return
+    }
+
+    const bytesCountLimit =
+      compressionStatus === CustomerDataCompressionStatus.Disabled
+        ? CUSTOMER_DATA_BYTES_LIMIT
+        : CUSTOMER_COMPRESSED_DATA_BYTES_LIMIT
```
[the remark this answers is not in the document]
BenoitZugmeyer: Mmh "before init"?
* in Logs, before init (and after init), compression status is `disabled`, so we use the lower limit
* in Rum, before init, compression status is `unknown` and no message will be shown

PR #2553, opened 2024-01-05
🐛 [RUM-2689] fix recorder crash when restarted quickly

BenoitZugmeyer:
## Motivation

When restarted quickly, the Session Replay recorder unexpectedly stops.

### To reproduce

* Run the following on an instrumented page with active session replay:

```js
DD_RUM.stopSessionReplayRecording()
DD_RUM.startSessionReplayRecording()
```

* Move the cursor a bit

* Look at the replay in Datadog (the issue is not visible in the Developer Extension because the extension receives the records before they are deflate-encoded)

### Expected

The last cursor movements should be visible

### Actual

The last cursor movements are missing

## Changes

This commit makes sure the recorder keeps using the same DeflateEncoder instance when restarted. This ensures the underlying stream (identified by `DeflateEncoderStreamId.REPLAY`) is emitting DeflateWorkerAction with continuous identifiers.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/boot/recorderApi.spec.ts
+++ b/packages/rum/src/boot/recorderApi.spec.ts
@@ -1,4 +1,4 @@
-import type { DeflateEncoder, DeflateWorker } from '@datadog/browser-core'
+import type { DeflateEncoder, DeflateWorker, DeflateWorkerAction } from '@datadog/browser-core'
 import { display, isIE } from '@datadog/browser-core'
 import type { RecorderApi, ViewContexts, LifeCycle, RumConfiguration } from '@datadog/browser-rum-core'
 import { LifeCycleEventType } from '@datadog/browser-rum-core'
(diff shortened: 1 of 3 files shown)


packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -162,6 +162,21 @@ describe('makeRecorderApi', () => {
...
+
+      recorderApi.stop()
+      recorderApi.start()
+
+      // Types are specified to make sure we are getting the correct arguments
+      const firstCallDeflateEncoder: DeflateEncoder = startRecordingSpy.calls.argsFor(0)[4]
+      const secondCallDeflateEncoder: DeflateEncoder = startRecordingSpy.calls.argsFor(1)[4]
+
+      expect(firstCallDeflateEncoder).toBe(secondCallDeflateEncoder)
+    })
```
bcaudan, line 178: **💬 suggestion:** ‏looking at this test alone, it is hard to understand why we want this behaviour.
Could we express a more high level behaviour?

bcaudan (approved):

amortemousque (approved):

PR #2560, opened 2024-01-09
✨ [RUM-2729] collect connectivity data

bcaudan:
## Motivation

Collect connectivity data to provide more context on end user environment.

## Changes

Collect when available:

- status: `connected` | `not_connected`
- interface: `bluetooth` | `wifi` | ...
- effective_type: `2g` | `3g` | ...

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


BenoitZugmeyer (approved):

tarsojabbes:
Hey, @bcaudan, I was looking for someone using Network Information API and found out Datadog was doing something similar to what I'm working on. Your approach was quite the same as mine, except that I'm having trouble to access `interface` as it always returns `undefined`. Have you ever encountered problems like this with your solution? (Probably related to the experimental nature of this API). Also, it only collects network connectivity information in Chrome and Edge, right?

bcaudan:
Hi @tarsojabbes,

Looking at the [spec](https://wicg.github.io/netinfo/), there is no `interface` attribute, so not sure what you are trying to achieve exactly.
For browser support, you should probably check https://caniuse.com/?search=navigator.connection.

tarsojabbes:
Super thanks for your feedback!

My bad, when I said `interface` I was thinking about the `navigator.connection.type`, to catch if the user was using either wifi, cellular etc 😄 I also checked the link for Can I Use, it's really limited to a few browsers (I didn't know about this website, pretty handy btw).

PR #2565, opened 2024-01-11
🐛[RUM-2752] Replay: generate censored images with custom dimensions

bcaudan:
## Motivation

Default censored image dimension is 1x1 which can cause rendering issue when the image size ratio is used by the browser to display the image.

Screenshot 2024-01-11 at 18 08 34

## Changes

Generate a censored image from the [natural dimensions](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/naturalWidth) of the image, fallback to rendered dimension if not avialable. 

Screenshot 2024-01-11 at 18 10 04

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/serialization/serializeAttribute.ts
+++ b/packages/rum/src/domain/record/serialization/serializeAttribute.ts
@@ -3,6 +3,7 @@ import { STABLE_ATTRIBUTES } from '@datadog/browser-rum-core'
 import type { RumConfiguration } from '@datadog/browser-rum-core'
 import { NodePrivacyLevel, PRIVACY_ATTR_NAME, CENSORED_STRING_MARK, CENSORED_IMG_MARK } from '../../../constants'
 import { MAX_ATTRIBUTE_VALUE_CHAR_LENGTH } from '../privacy'
+import { censoredImageForSize } from './serializationUtils'
 
 export function serializeAttribute(
   element: Element,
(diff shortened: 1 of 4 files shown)


packages/rum/src/domain/record/serialization/serializeAttribute.ts
```diff
@@ -30,12 +31,23 @@ export function serializeAttribute(
...
-      if (attributeName === 'src' || attributeName === 'srcset') {
-        return CENSORED_IMG_MARK
+    if (tagName === 'IMG' && (attributeName === 'src' || attributeName === 'srcset')) {
+      // generate image with similar dimension than the original to have the same rendering behaviour
+      const image = element as HTMLImageElement
+      if (image.naturalWidth > 0) {
+        return censoredImageForSize(image.naturalWidth, image.naturalHeight)
       }
+      const { width, height } = element.getBoundingClientRect()
+      return censoredImageForSize(width, height)
```
BenoitZugmeyer, line 43: **💬 suggestion:** ‏if the image is in `display:none` when serialized, the width and height will be 0. If it then become visible, the censored image won't show up because its size will be 0.

Maybe ensure width/height is > 0, or fallback to CENSORED_IMG_MARK

BenoitZugmeyer (approved):

PR #2575, opened 2024-01-17
♻️ [RUM-2445] split RUM and Logs public APIs modules

BenoitZugmeyer:
## Motivation

The `rumPublicApi` and `logsPublicApi` modules are a bit large. As we want to add more logic before starting RUM and Logs for consent management, it is a good time to extract the "pre start" logic (code executed before actually starting RUM or Logs) to separate modules.

## Changes

In both RUM and Logs public APIs, introduce a `Strategy` interface for behaviors that differs before and after starting the SDK. The `Strategy` replaces the various `xxxStrategy` functions, and is created by a dedicated `createPreStartStrategy` function. The "post start" strategy is returned by `startRum` and `startLogs` functions, as before.

In the RUM case, the `trackViewManually` logic has been revamped a little bit to make it a bit more explicit (the first `startView` call can be ignored and its related View options can be retrieved when starting RUM)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -0,0 +1,169 @@
...
+    startView(options, startClocks = clocksNow()) {
+      let ignore = false
+      if (!firstStartView) {
+        firstStartView = {
+          options,
+          ignore() {
+            ignore = true
+          },
+        }
+        tryStartRum()
```
bcaudan, line 139: **❓ question:** ‏what is happening around this first start view ignore?

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -157,16 +118,16 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
...
-  }
-
-  function canInitLogs(initConfiguration: LogsInitConfiguration) {
-    if (isAlreadyInitialized) {
-      displayAlreadyInitializedError('DD_LOGS', initConfiguration)
-      return false
-    }
-    return true
-  }
+function createActualStrategy(initConfiguration: LogsInitConfiguration, startLogsResult: StartLogsResult): Strategy {
```
bcaudan, line 123: **💬 suggestion:** ‏what about using the same naming between RUM and logs? (actual VS postStart)

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+
+      bufferApiCalls.add((startRumResult) => {
+        if (!ignore) {
+          startRumResult.startView(options, startClocks)
+        }
+      })
+    },
```
amortemousque, line 147: **💬 suggestion:** ‏Maybe we could direcly use the `firstStartView` ref instead of the `ignore` flag and remove the ref from the bufferApiCalls in case of trackViewsManually is used. Something like:

```typescript
startView(options, startClocks = clocksNow()) {
      const bufferedStartView = (startRumResult: StartRumResult) => {
        startRumResult.startView(options, startClocks)
      }
      if (!firstStartView) {
        firstStartViewOption = options
        firstStartView = bufferedStartView
      }

      bufferApiCalls.add(bufferedStartView)
    }
```

and in `tryStartRum`

```typescript
 if (cachedConfiguration.trackViewsManually) {
      bufferApiCalls.remove(firstStartView!)
      initialViewOptions = firstStartViewOption
    }
```

packages/rum-core/src/boot/preStartRum.ts
BenoitZugmeyer: With `trackViewManually: true`, we want to use the first `startView()` call options, and don't actually call `startView`. This is the same behavior as before, but a bit more explicit.

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -163,9 +164,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs) {
 
   function canInitLogs(initConfiguration: LogsInitConfiguration) {
     if (isAlreadyInitialized) {
-      if (!initConfiguration.silentMultipleInit) {
-        display.error('DD_LOGS is already initialized.')
-      }
+      displayAlreadyInitializedError('DD_LOGS', initConfiguration)
```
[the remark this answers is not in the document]
BenoitZugmeyer: `canInitSdk` has a state associated to it (`isAlreadyInitialized`), so it would be more complex than just a plain function and I don't think it's worth it. WDYT?

packages/rum-core/src/boot/preStartRum.ts
bcaudan: ok, I get it now.
What about:
- add an extra comment to also explain it here
- rename `firstStartView` to something like `firstStartViewCall`, to make a bit more explicit that we talk about the API call

packages/logs/src/boot/logsPublicApi.ts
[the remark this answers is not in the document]
amortemousque: Yes, right. You can leave it this way. No problem

packages/rum-core/src/boot/preStartRum.ts
BenoitZugmeyer: Since Aymeric also had some suggestions on that part, I changed the logic a little bit. Let me know WDYT :)

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -331,20 +231,16 @@ export function makeRumPublicApi(
...
-    return assign({}, initConfiguration, {
-      applicationId: '00000000-aaaa-0000-aaaa-000000000000',
-      clientToken: 'empty',
-      sessionSampleRate: 100,
-    })
-  }
+function createPostStartStrategy(initConfiguration: RumInitConfiguration, startRumResult: StartRumResult): Strategy {
+  return assign(
+    {
+      init: (initConfiguration: RumInitConfiguration) => {
```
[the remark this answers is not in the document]
BenoitZugmeyer: For the record: I tried to make some changes, but after a review and discussions with Aymeric we were not quite satisfied with the tradeoffs. It seems to be better to keep "pre start" and "post start" strategies strictly separated.

amortemousque (approved):

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -0,0 +1,168 @@
...
+    if (cachedConfiguration.trackViewsManually) {
+      if (!firstStartViewCall) {
+        return
+      }
+      // An initial view is always created when starting RUM.
+      // When tracking views automatically, any startView call before RUM start creates an extra
+      // view.
+      // When tracking views manually, we use the ViewOptions from the first startView call as the
+      // initial view options, and we remove the actual startView call so we don't create
+      bufferApiCalls.remove(firstStartViewCall.callback)
```
bcaudan, line 58: **💬 suggestion:** unfinished sentence?‏
```suggestion
      // When tracking views manually, we use the ViewOptions from the first startView call as the
      // initial view options, and we remove the actual startView call so we don't create an extra view.
      bufferApiCalls.remove(firstStartViewCall.callback)
```

packages/core/src/tools/boundedBuffer.ts
```diff
@@ -1,17 +1,24 @@
...
     }
   }
 
-  drain() {
-    this.buffer.forEach((callback) => callback())
+  remove(callback: (arg: T) => void) {
+    const index = this.buffer.indexOf(callback)
+    if (index !== -1) {
+      this.buffer.splice(index, 1)
+    }
```
bcaudan, line 17: **🥜 nitpick:** ‏we are doing that at a couple of other places in the code, what about extracting a method in `arrayUtils`?

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -0,0 +1,168 @@
...
+      // Expose the initial configuration regardless of initialization success.
+      cachedInitConfiguration = initConfiguration
+
+      if (cachedConfiguration) {
+        displayAlreadyInitializedError('DD_RUM', initConfiguration)
+        return
+      }
+
+      // If we are in a Synthetics test configured to automatically inject a RUM instance, we want to
+      // completely discard the customer application RUM instance by ignoring their init() call.  But,
```
bcaudan, line 88: **🥜 nitpick:** ‏extra space
```suggestion
      // completely discard the customer application RUM instance by ignoring their init() call. But,
```

bcaudan (approved):

PR #2578, opened 2024-01-25
🐛 [RUM-2720] Send logs without session id when session inactive

amortemousque:
## Motivation

Do not drop logs when session inactive

⚠ _it's worth reviewing commit by commit_

## Changes

- Keep untracked session in sessionContextHistory
- Allow to find expired session
- Send logs when a session is expired
- Only set the session id if the session is active

## Testing

- [X] Local
- [x] Staging
- [X] Unit
- [ ] End to end

---


packages/core/src/tools/valueHistory.spec.ts
```diff
@@ -54,6 +54,15 @@ describe('valueHistory', () => {
...
       expect(valueHistory.find(15 as RelativeTime)).toBeUndefined()
     })
+
+    describe('with predicate', () => {
+      it('should return the value of the entry matching the startTime and the predicate', () => {
+        valueHistory.add('foo', 0 as RelativeTime).close(10 as RelativeTime)
+        valueHistory.add('bar', 20 as RelativeTime)
+
+        expect(valueHistory.find(15 as RelativeTime, (entry) => entry.endTime === 10)).toEqual('foo')
+      })
```
bcaudan, line 64: **❗ warning:** ‏given your strategy with `() => true` predicate for logs session, I would expect:

```
expect(valueHistory.find(15 as RelativeTime, () => true)).toEqual('bar')
```

but currently it returns `'foo'`.

am I missing something?

amortemousque: To me it's correct. It returns 'foo' because the "foo" startTime is 15 which is greater than 0 but lower than 20. 
In the context of logs it means that we return the latest session even if closed.

bcaudan: work as expected with 25, so it looks good 😅

packages/core/src/transport/flushController.spec.ts
```diff
@@ -85,25 +85,25 @@ describe('flushController', () => {
     it('notifies when the session expires', () => {
       flushController.notifyBeforeAddMessage(SMALL_MESSAGE_BYTE_COUNT)
       flushController.notifyAfterAddMessage()
-      sessionExpireObservable.notify()
+      sessionExpireObservable.notify('')
```
bcaudan, line 88: **💬 suggestion:** ‏what about using a type constant instead of an empty string?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -15,6 +15,8 @@ export interface SessionManager<TrackingType extends string> {
     startTime?: RelativeTime,
     predicate?: (entry: ValueHistoryEntry<SessionContext<TrackingType>>) => boolean
   ) => SessionContext<TrackingType> | undefined
+  renewObservable: Observable<string>
+  expireObservable: Observable<string>
```
bcaudan, line 19: **💬 suggestion:** use `TrackingType` instead of `string`
‏

packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -28,17 +28,21 @@ export function startRumSessionManager(configuration: RumConfiguration, lifeCycl
...
+      lifeCycle.notify(LifeCycleEventType.SESSION_EXPIRED)
+    }
   })
 
-  sessionManager.renewObservable.subscribe(() => {
-    lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
+  sessionManager.renewObservable.subscribe((rawTrackingType) => {
+    if (isTypeTracked(rawTrackingType as RumTrackingType)) {
+      lifeCycle.notify(LifeCycleEventType.SESSION_RENEWED)
+    }
```
bcaudan, line 40: **❓ question:** ‏no unit tests require those checks, do we need them?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -114,8 +113,15 @@ export function startSessionStore<TrackingType extends string>(
       sessionState = {}
     }
     if (hasSessionInCache()) {
+      // A session can be outdated if a new session has been generated in another browser tab
+      // if the new session is tracked, only expire the current session to restart after a user activity
+      // if the new session is not tracked, renew the session so that logs sdk stop sending events
```
bcaudan, line 118: **💬 suggestion:** ‏What about adding a test for that?

packages/rum-core/src/domain/rumSessionManager.ts
amortemousque: Yes we do :)

packages/logs/src/domain/logsSessionManager.ts
```diff
@@ -24,11 +25,12 @@ export function startLogsSessionManager(configuration: LogsConfiguration): LogsS
     computeSessionState(configuration, rawTrackingType)
   )
   return {
-    findTrackedSession: (startTime) => {
+    findTrackedSession: (startTime = relativeNow()) => {
       const session = sessionManager.findSession(startTime, isActiveOrInactiveSessionPredicate)
       return session && session.trackingType === LoggerTrackingType.TRACKED
         ? {
             id: session.id,
+            active: (session.endTime || Infinity) > startTime,
```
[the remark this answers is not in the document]
amortemousque: I've exposed [isActiveAt](https://github.com/DataDog/browser-sdk/blob/4a5df2185135494bd4f59c3fa85446ac90186845/packages/logs/src/domain/logsSessionManager.ts#L32). Let me know what you think.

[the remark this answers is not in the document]
bcaudan: Counter-proposal: what about adding an option to the `findTrackedSession` method to return only active session by default but allow to return expire sessions?
something like:
```
findTrackedSession(startTime, { returnExpired: true })
```

so by default it would have the same behaviour than RUM but when we need to also check for expired sessions it would be explicit.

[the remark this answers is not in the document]
amortemousque: Hum, it would mean to have 2 findTrackedSession() calls in the logsAssembly:

- One for the check 
```typescript
const session = sessionManager.findTrackedSession(startTime, { returnExpired: true })
if (!session) {
    return
}
```
- Another to set the session id
```typescript
const log = combine(
{
session_id: sessionManager.findTrackedSession(startTime)?.id,
...
}
```
If that sounds good for you, it's good for me :)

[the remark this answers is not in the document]
bcaudan: Hum, that does not really say what it is doing though 🤔

It would be nice to be be able to express:
```js
{
  session_id: !session.isExpired : session.id : undefined,
  ...
}
```

But then this property would be computed from the `startTime` which is what I found confusing originally 😕

I like the idea of not impacting other places where we are checking for active tracked session and being explicit about what we are doing in the logs assembly.

I'll let you choose the approach that seems best with your understanding 😉

packages/core/src/domain/session/sessionStore.ts
amortemousque: There are already tests covering these behaviours but they are splitting between RUM and Logs (which is logical). I've removed this comment and only added one specific to my change.

packages/logs/src/domain/logsSessionManager.ts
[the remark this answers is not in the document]
amortemousque: Okay, so I chose to have both { returnExpired: true } option and the isActiveAt() method to be more explicit. So basically in the assembly we have

```typescript
const session = sessionManager.findTrackedSession(startTime, { returnExpired: true })
if (!session) {
  return
}
const log = combine(
{
  session_id: session.isActiveAt(startTime) ? session.id : undefined,
  ...
}
```
Let me know what you think

packages/core/src/domain/session/sessionStore.ts
bcaudan: > if the new session is not tracked, renew the session so that logs sdk stop sending events

is there a test for that?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -116,6 +115,11 @@ export function startSessionStore<TrackingType extends string>(
     if (hasSessionInCache()) {
       if (isSessionInCacheOutdated(sessionState)) {
         expireSessionInCache()
+
+        // Only renew if the session is not tracked, because we want tracked session to be renewed on user activity
+        if (sessionState[productKey] && !computeSessionState(sessionState[productKey]).isTracked) {
+          renewSessionInCache(sessionState)
+        }
```
bcaudan, line 122: **💬 suggestion:** ‏on a design perspective, it is quite bothering that we need to add a specific behaviour related to how this component his used.
Could we move this logic up?

packages/core/src/domain/session/sessionStore.ts
bcaudan: This use case `if the new session is not tracked, renew the session so that logs sdk stop sending events` 😅

FMU, It is the intent for some of the sessionStore changes but it is hard to guess it from the sessionStore changes only.

packages/core/src/domain/session/sessionStore.ts
nulrich: Those lines have been removed after discussion.

packages/logs/src/domain/configuration.ts
```diff
@@ -19,6 +19,7 @@ export interface LogsInitConfiguration extends InitConfiguration {
   forwardErrorsToLogs?: boolean | undefined
   forwardConsoleLogs?: ConsoleApiName[] | 'all' | undefined
   forwardReports?: RawReportType[] | 'all' | undefined
+  sendLogsAfterSessionExpiration?: boolean
```
amortemousque, line 22: **💬 suggestion:** ‏Could be nice to add a comment: `// TODO next major: remove this option and make it the default behaviour`

packages/core/src/domain/session/sessionStore.ts
nulrich: This use case is not supported anymore.

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -236,4 +240,68 @@ describe('logs', () => {
...
+        { ...baseConfiguration, sendLogsAfterSessionExpiration: true },
+        () => COMMON_CONTEXT,
+        createTrackingConsentState(TrackingConsent.GRANTED)
+      ))
+      registerCleanupTask(stopLogs)
+
+      handleLog({ status: StatusType.info, message: 'message 1' }, logger)
+
+      // expire session
+      setCookie(SESSION_STORE_KEY, 'isExpired=1', ONE_MINUTE)
```
BenoitZugmeyer, line 266: **🥜 nitpick:** ‏this should be in a test helper, as we're doing it in many places

packages/logs/src/domain/assembly.ts
```diff
@@ -35,7 +37,7 @@ export function startLogsAssembly(
       const log = combine(
         {
           service: configuration.service,
-          session_id: session.id,
+          session_id: session.isActiveAt(startTime) ? session.id : undefined,
```
BenoitZugmeyer, line 40: **💬 suggestion:** ‏I think `isActiveAt` complicates things quite a bit. It might be simpler to just call `findTrackedSession` multiple times:

```ts

if (
  !session &&
  (!configuration.sendLogsAfterSessionExpiration || !sessionManager.findTrackedSession(startTime, { returnInactive: true }))) {
  return
}

// and then

session_id: session?.id
```

BenoitZugmeyer (commented):
I can see there is three ways of getting inactive sessions:
* in `valueHistory`, there is the `AFTER_ENTRY_START` predicate
* in `sessionManager`, there is a different method `findInactiveOrExpiredSession`
* in `logsSessionManager`, there is a `{ returnExpired }` option argument.

I would expect to use the same for all three. I like the option object, but not so much the "expired" name, as we use already use "active", I think we should use "inactive".

So my suggestion:

* in `valueHistory`, have `find(startTime, { returnInactive })`
* in `sessionManager`, introduce a `findSession(startTime, { returnInactive })`, that handles both cases (remove other `find*Session` methods)
* in `logsSessionManager`, have `findTrackedSession(startTime, { returnInactive })`

(`returnInactive` should default to `false` in `valueHistory`, and you can just pass the options down without worrying about the default in other places)

packages/logs/src/domain/assembly.ts
amortemousque: Interesting! It removes the needs to add `endTime` to `SessionContext`. I like it 👍

BenoitZugmeyer (approved):

thomas-lebeau (approved):

thomas-lebeau (approved):

PR #2579, opened 2024-01-26
⚡️ [RUM-2917] optimize getNodePrivacyLevel by adding a cache

BenoitZugmeyer:
## Motivation

A customer reported that `getNodePrivacyLevel` was a big performance bottleneck on their application when the devtools were open. On our side, we didn't observe a huge performance impact caused by this function (with or without devtools open), but still find a way to optimize it, which might help the customer case.

## Changes

Cache the node privacy level when processing mutations. The cache is local to the `proccessMutations` function so we don't have to implement a complex way to invalidate it. Even with this simple cache, we observed a nice speedup while processing mutations (from ~10ms to ~1.5ms in a specific case on the Datadog App)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/observers/mutationObserver.ts
+++ b/packages/rum/src/domain/record/observers/mutationObserver.ts
@@ -14,6 +14,7 @@ import type {
   RemovedNodeMutation,
   TextMutation,
 } from '../../../types'
+import type { NodePrivacyLevelCache } from '../privacy'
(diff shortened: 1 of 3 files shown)


amortemousque (approved):

N-Boutaib (approved):
LGTM ✔️

bcaudan (approved):
Nice!

PR #2586, opened 2024-02-02
🐛 Fix LCP with size < previous LCP

amortemousque:
## Motivation

Fix LCP with size < previous LCP cf: https://bugs.chromium.org/p/chromium/issues/detail?id=1516655

## Changes

Discarded the LCP entry if it has a size inferior to the previous LCP entry

## Testing

- [ ] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -58,6 +58,7 @@ export function trackLargestContentfulPaint(
     { capture: true, once: true }
   )
 
+  let prevLcpSize = 0
```
bcaudan, line 61: **🥜 nitpick:** to be more explicit and consistent with the comment‏
```suggestion
  let biggestLcpSize = 0
```

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -102,4 +100,23 @@ describe('trackLargestContentfulPaint', () => {
 
     expect(lcpCallback).not.toHaveBeenCalled()
   })
+
+  it('should be discarded if it has a size inferior to the previous LCP entry', () => {
```
bcaudan, line 104: **💬 suggestion:** ‏what about adding a test to ensure that we are notifying multiple times when the size is bigger than the previous entry?

bcaudan (approved):

Issue #1008, opened 2021-08-20
Request for destroy/stop method


tyl-paul-lapczynski:
There is a scenario where I've come across where based on user cookie preferences, it would be ideal if datadogRum could be toggled on/off in a browser session. There looks like there is no API for something like `stop`, `destroy`, or `uninitialized`.

This would facilitate a good user experience without an interrupting full page refresh especially in SPAs.

BenoitZugmeyer:
Thank you for your suggestion, we'll consider it.

remkoboschker:
Hi, I am having the same issue. We now have to do a page reload an only initialize datadog when consent has been given. Also in the service that uses datadog to send custum events it would be usefull to be able to check if datadog was initialized. What happens if you call datadogRum.addError without it having been initialized?

BenoitZugmeyer:
@remkoboschker I'm not sure to see how a way to **stop** the SDK from collecting data would help in your use-case. It looks like you rather need a way to **start** collecting data when consent has been given? In this case, I suggest to initialize RUM if/when consent is given. All method calls (like `datadogRum.addError`) are buffered and will be transmitted only once `datadogRum.init` is called. How does it sound?

remkoboschker:
Thanks, good to know all calls are buffered until init is called. People can revoke or change there consent from within the application. After a consent change there is no way to stop sending other than reloading. 

Would buffering all session data impact performance (memory on an older phone) when init is never called?

Thank you for your reply

BenoitZugmeyer:
> People can revoke or change there consent from within the application. After a consent change there is no way to stop sending other than reloading. 

Ah, I get it now, it seems like a good use-case for a `stop()` method. We don't have any plan yet, but we definitely want to implement this API. We'll keep you posted!

> Would buffering all session data impact performance (memory on an older phone) when init is never called?

The buffer [is bound](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/boundedBuffer.ts) so it will not leak to much memory. Only the most recent 500 calls are taken into account during init.

jademh:
+1 - also need this for a similar cookie consent use case

gaiachik:
so +1 for the stop() method

Also on the same topic (cookie preferences management) but different solution:

if the datadogRum had a `privacyMode` which didn't set or read cookies, it could still be run in most cases where cookie consent is not given. Some SDK offer methods to set consent so that they can change the behaviour depending on it - it would be great to have something like this.

[...]

PR #2589, opened 2024-02-05
⚗️✨ [RUM-2445] implement Tracking Consent management

BenoitZugmeyer:
## Motivation

Allows starting/stopping data collection based on user consent (ex: regarding to RGPD rules and such).

Related issue: https://github.com/DataDog/browser-sdk/issues/1008

## Changes

For both RUM and Logs SDKs:

* Introduce a `trackingConsent: "granted"/"not-granted"` initialization parameter (defaults to `"granted"` for backward compatibility)
* Add a `setTrackingConsent` method to set the consent state before or after calling `init()`. If called before `init()`, the initialization parameter will be ignored.
* On init, do not start collecting data until consent is granted
* When the consent is withdrawn, stops the session and clear the cookie. No new session will be created on user interaction.
* This change is behind an experimental flag and should not be used by customers yet.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/core/src/domain/trackingConsent.ts
+++ b/packages/core/src/domain/trackingConsent.ts
@@ -1,0 +1,5 @@
+export const TrackingConsent = {
+  GRANTED: 'granted',
+  NOT_GRANTED: 'not-granted',
+} as const
+export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingConsent]
--- a/packages/core/src/domain/trackingConsent.ts
+++ b/packages/core/src/domain/trackingConsent.ts
@@ -8,8 +8,8 @@ export const TrackingConsent = {
 export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingConsent]
 
 export interface TrackingConsentState {
-  setIfNotDefined: (trackingConsent: TrackingConsent) => void
-  set: (trackingConsent: TrackingConsent) => void
+  tryToInit: (trackingConsent: TrackingConsent) => void
+  update: (trackingConsent: TrackingConsent) => void
   isGranted: () => boolean
   observable: Observable<void>
 }
@@ -18,12 +18,12 @@ export function createTrackingConsentState(currentConsent?: TrackingConsent): Tr
   const observable = new Observable<void>()
 
   return {
-    setIfNotDefined(trackingConsent: TrackingConsent) {
+    tryToInit(trackingConsent: TrackingConsent) {
       if (!currentConsent) {
         currentConsent = trackingConsent
       }
     },
-    set(trackingConsent: TrackingConsent) {
+    update(trackingConsent: TrackingConsent) {
       currentConsent = trackingConsent
       observable.notify()
     },
--- a/packages/core/src/domain/trackingConsent.ts
+++ b/packages/core/src/domain/trackingConsent.ts
@@ -1,5 +1,38 @@
+import { ExperimentalFeature, isExperimentalFeatureEnabled } from '../tools/experimentalFeatures'
+import { Observable } from '../tools/observable'
+
 export const TrackingConsent = {
   GRANTED: 'granted',
   NOT_GRANTED: 'not-granted',
 } as const
 export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingConsent]
+
+export interface TrackingConsentState {
+  setIfNotDefined: (trackingConsent: TrackingConsent) => void
+  set: (trackingConsent: TrackingConsent) => void
+  isGranted: () => boolean
+  observable: Observable<void>
+}
+
+export function createTrackingConsentState(currentConsent?: TrackingConsent): TrackingConsentState {
+  const observable = new Observable<void>()
+
+  return {
+    setIfNotDefined(trackingConsent: TrackingConsent) {
+      if (!currentConsent) {
+        currentConsent = trackingConsent
+      }
+    },
+    set(trackingConsent: TrackingConsent) {
+      currentConsent = trackingConsent
+      observable.notify()
+    },
+    isGranted() {
+      return (
+        !isExperimentalFeatureEnabled(ExperimentalFeature.TRACKING_CONSENT) ||
+        currentConsent === TrackingConsent.GRANTED
+      )
+    },
+    observable,
+  }
+}
--- a/packages/logs/src/boot/logsPublicApi.ts
+++ b/packages/logs/src/boot/logsPublicApi.ts
@@ -1,4 +1,4 @@
-import type { Context, User } from '@datadog/browser-core'
+import type { Context, TrackingConsent, User } from '@datadog/browser-core'
 import {
   CustomerDataType,
   assign,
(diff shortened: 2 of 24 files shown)


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -144,6 +146,8 @@ export function makeRumPublicApi(startRumImpl: StartRum, recorderApi: RecorderAp
   const rumPublicApi = makePublicApi({
     init: monitor((initConfiguration: RumInitConfiguration) => strategy.init(initConfiguration)),
 
+    setTrackingConsent: monitor((trackingConsent: TrackingConsent) => strategy.setTrackingConsent(trackingConsent)),
```
amortemousque, line 149: **💬 suggestion:** ‏‏I don't know If you planned to do it later but it could be nice to add a JS doc. Maybe we will have a related guide?

amortemousque (approved):

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -38,9 +39,11 @@ export function createPreStartStrategy(
 
   let cachedInitConfiguration: RumInitConfiguration | undefined
   let cachedConfiguration: RumConfiguration | undefined
+  const trackingConsentState = createTrackingConsentState()
```
amortemousque, line 42: **💬 suggestion:** ‏
I would have expected to pass the `trackingConsentState` to `doStartRum` to avoid having to recreate the consent state in `startRum`

packages/core/src/domain/trackingConsent.ts
```diff
@@ -0,0 +1,38 @@
...
+export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingConsent]
+
+export interface TrackingConsentState {
+  setIfNotDefined: (trackingConsent: TrackingConsent) => void
+  set: (trackingConsent: TrackingConsent) => void
+  isGranted: () => boolean
+  observable: Observable<void>
+}
+
+export function createTrackingConsentState(currentConsent?: TrackingConsent): TrackingConsentState {
```
amortemousque, line 17: **🥜 nitpick:** To be consistent with our namings I would expect to have‏
```suggestion
export function startTrackingConsent(currentConsent?: TrackingConsent): TrackingConsentState {
```

packages/rum-core/src/boot/preStartRum.ts
BenoitZugmeyer: Sure, it's possible but it's yet another value to pass through. Because it doesn't change anything I chose to create it here, but if you prefer I can do what you are suggesting.

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -32,6 +32,7 @@ const LOGS_STORAGE_KEY = 'logs'
 
 export interface Strategy {
   init: (initConfiguration: LogsInitConfiguration) => void
+  setTrackingConsent: StartLogsResult['setTrackingConsent']
```
bcaudan, line 35: **❗ warning:** ‏the API also needs to be exposed from the global variable.

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -23,9 +24,11 @@ export function createPreStartStrategy(
   const bufferApiCalls = new BoundedBuffer<StartLogsResult>()
   let cachedInitConfiguration: LogsInitConfiguration | undefined
   let cachedConfiguration: LogsConfiguration | undefined
+  const trackingConsentState = createTrackingConsentState()
+  trackingConsentState.observable.subscribe(tryStartLogs)
```
bcaudan, line 28: **❓ question:** ‏is it ok to trigger multiple startLogs if the consent API is called multiple times with the granted value?

packages/rum-core/src/boot/rumPublicApi.ts
BenoitZugmeyer: I'm fine with starting to write JSDocs on new public APIs! Good idea.

packages/core/src/domain/trackingConsent.ts
BenoitZugmeyer: Mmh we have a bunch of `createXXX` function as well, I kind of prefer sticking to the `create` naming when returning the associated type.

packages/logs/src/boot/preStartLogs.ts
BenoitZugmeyer: It is not OK to call `startLogs` multiple time. Do you see any scenario where this can happen?

packages/logs/src/boot/logsPublicApi.ts
BenoitZugmeyer: Yeah I figured it out 😄

packages/logs/src/boot/preStartLogs.ts
bcaudan:
```js
DD_LOGS.setTrackingConsent('granted')
DD_LOGS.setTrackingConsent('granted')
```

Probably not on purpose but instrumentation mistakes can happen (like multiple init).

BenoitZugmeyer: It calls `tryStartLogs` multiple times, but not `startLogs`, right?

bcaudan: oh, it actually not even calls `tryStartLogs` multiple times, since two different trackingConsent components are used before and after start.

packages/logs/src/boot/startLogs.ts
```diff
@@ -35,10 +37,11 @@ export function startLogs(
 
   const reportError = startReportError(lifeCycle)
   const pageExitObservable = createPageExitObservable(configuration)
+  const trackingConsentState = createTrackingConsentState(TrackingConsent.GRANTED)
```
bcaudan, line 40: **💬 suggestion:** ‏could we share the same trackingConsentState instance between preStart and start? or maybe just passing the observable to the start?

BenoitZugmeyer: Here are my thoughts: https://github.com/DataDog/browser-sdk/pull/2589#discussion_r1480002807
Let's have a word tomorrow!

packages/rum-core/src/boot/preStartRum.ts
bcaudan: I would find it easier to follow if we had a single component handling this logic rather than having two components, with one overriding the other.

packages/logs/src/boot/startLogs.ts
```diff
@@ -28,7 +27,10 @@ export type StartLogsResult = ReturnType<StartLogs>
 export function startLogs(
   initConfiguration: LogsInitConfiguration,
   configuration: LogsConfiguration,
-  getCommonContext: () => CommonContext
+  getCommonContext: () => CommonContext,
+
+  // Tracking consent should always be granted when the startLogs is called.
+  trackingConsentState: TrackingConsentState
```
bcaudan, line 33: **💬 suggestion:** ‏I still don't fully understand why you feel the need to add this comment. 
Could you maybe develop a bit the comment to help the reader understand what they should do with this information?

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -19,19 +19,20 @@ import type { StartLogsResult } from './startLogs'
...
-  const trackingConsentState = createTrackingConsentState()
-  trackingConsentState.observable.subscribe(tryStartLogs)
+  const trackingConsentStateSubscription = trackingConsentState.observable.subscribe(tryStartLogs)
 
   function tryStartLogs() {
     if (!cachedConfiguration || !cachedInitConfiguration || !trackingConsentState.isGranted()) {
       return
     }
 
+    trackingConsentStateSubscription.unsubscribe()
```
bcaudan, line 35: **💬 suggestion:** ‏Could we have a test case to cover this unsubscribe?

bcaudan (approved):

amortemousque (approved):

PR #2591, opened 2024-02-05
⚗️[RUM-2889] Bootstrap custom vital APIs

bcaudan:
## Motivation

Initial PR to provide APIs to track custom duration as vital events

## Changes

Behind `custom_vitals` flag, add APIs:

```js
DD_RUM.startDurationVital('foo')
// do some stuff
DD_RUM.stopDurationVital('foo')
```

and support the new event type across the code base:

Screenshot 2024-02-05 at 15 17 25

**Note:** First PR with the simplest API, more features in following PRs (customer context, rate limiter, ...)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/trackEventCounts.spec.ts
+++ b/packages/rum-core/src/domain/trackEventCounts.spec.ts
@@ -27,11 +27,12 @@ describe('trackEventCounts', () => {
     notifyCollectedRawRumEvent({ type: RumEventType.LONG_TASK })
     expect(eventCounts.longTaskCount).toBe(1)
   })
-
-  it("doesn't track views", () => {
-    const { eventCounts } = trackEventCounts({ lifeCycle, isChildEvent: () => true })
-    notifyCollectedRawRumEvent({ type: RumEventType.VIEW })
-    expect(objectValues(eventCounts as unknown as { [key: string]: number }).every((value) => value === 0)).toBe(true)
+  ;[RumEventType.VIEW, RumEventType.VITAL].forEach((eventType) => {
+    it(`doesn't track ${eventType} events`, () => {
+      const { eventCounts } = trackEventCounts({ lifeCycle, isChildEvent: () => true })
+      notifyCollectedRawRumEvent({ type: eventType })
+      expect(objectValues(eventCounts as unknown as { [key: string]: number }).every((value) => value === 0)).toBe(true)
+    })
   })
 
   it('tracks actions', () => {
--- a/packages/rum-core/src/domain/vital/vitalCollection.ts
+++ b/packages/rum-core/src/domain/vital/vitalCollection.ts
@@ -1,4 +1,4 @@
-import type { ClocksState } from '@datadog/browser-core'
+import type { ClocksState, Duration } from '@datadog/browser-core'
 import { elapsed, generateUUID } from '@datadog/browser-core'
 import { LifeCycleEventType } from '../lifeCycle'
 import type { LifeCycle, RawRumEventCollectedData } from '../lifeCycle'
@@ -15,11 +15,11 @@ export interface DurationVitalStop {
   stopClocks: ClocksState
 }
 
-interface Vital {
+interface DurationVital {
   name: string
-  type: VitalType
+  type: VitalType.DURATION
   startClocks: ClocksState
-  value: number
+  value: Duration
 }
 
 export function startVitalCollection(lifeCycle: LifeCycle) {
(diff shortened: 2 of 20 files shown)


packages/rum-core/src/rumEvent.types.ts
```diff
@@ -276,6 +282,96 @@ export type RumErrorEvent = CommonProperties &
         }
         [k: string]: unknown
       }
+      /**
+       * Description of each thread in the process when error happened.
+       */
+      threads?: {
```
N-Boutaib, line 288: ❔ **Question**: wouldn't it be straightforward to type a single thread first, then reuse that type for the `threads` attribute ?

packages/rum-core/src/domain/trackEventCounts.spec.ts
```diff
@@ -27,11 +27,12 @@ describe('trackEventCounts', () => {
     notifyCollectedRawRumEvent({ type: RumEventType.LONG_TASK })
     expect(eventCounts.longTaskCount).toBe(1)
   })
-
-  it("doesn't track views", () => {
-    const { eventCounts } = trackEventCounts({ lifeCycle, isChildEvent: () => true })
-    notifyCollectedRawRumEvent({ type: RumEventType.VIEW })
-    expect(objectValues(eventCounts as unknown as { [key: string]: number }).every((value) => value === 0)).toBe(true)
+  ;[RumEventType.VIEW, RumEventType.VITAL].forEach((eventType) => {
```
N-Boutaib, line 30: ❔ **Question**: Why using the `;` at the start of this line 🤔 ?

packages/rum-core/src/rumEvent.types.ts
bcaudan: Those types are generated automatically from the JSON schema.
Not sure if we can influence the generator to produce types with a certain style.
Do you see any issue with it?

packages/rum-core/src/domain/trackEventCounts.spec.ts
bcaudan: it is actually a syntax requirement.
FMU, when not using trailing semi-colon style, you have to put a semi colon before square brackets to differentiate:
```js
obj[x]

// VS

obj;[x]
```

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -11,9 +15,49 @@ export interface DurationVitalStop {
...
-export function startVitalCollection(lifecycle: LifeCycle) {
+interface Vital {
+  name: string
+  type: VitalType
+  startClocks: ClocksState
+  value: number
+}
+
+export function startVitalCollection(lifeCycle: LifeCycle) {
+  const vitalStartsByName: { [name: string]: DurationVitalStart } = {}
```
BenoitZugmeyer, line 26: **💬 suggestion:** when storing arbitrary keys in a hashmap, always prefer a `Map` over a plain object to avoid prototype polution and other issues when keys match native object properties.

packages/rum-core/src/rumEvent.types.ts
N-Boutaib: > Do you see any issue with it?

Not really, I was just thinking if we ever wanted to use the type of `thread`, it would be straightforward to have a dedicated one instead of extracting it from the parent type.

N-Boutaib (approved):

packages/rum-core/src/rumEvent.types.ts
bcaudan: >  if we ever wanted to use the type of thread

On the browser side, we probably won't 🙂

Actually, I think a thread schema could be extracted to generate a thread type if we need it at some point.

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -0,0 +1,63 @@
...
+  name: string
+  stopClocks: ClocksState
+}
+
+interface Vital {
+  name: string
+  type: VitalType
+  startClocks: ClocksState
+  value: number
+}
```
amortemousque, line 23: **💬 suggestion:** ‏We could explicitly specify an `DurationVital` type
```suggestion
interface DurationVital {
  name: string
  type: VitalType.Duration
  startClocks: ClocksState
  value: Duration
}
```
Wdyt?

bcaudan: I had some thoughts about how much distinction we should make now between Vital, CustomVital, CustomDurationVital.
I settled on decoupling the `processVital` from the exact type of Vital (without handling the logic around custom VS non-custom just yet) which is similar to what we are doing for action for example.

But yes, we could remove some future thinking and just go with what is really handled for now.

amortemousque (approved):

BenoitZugmeyer (approved):

PR #2598, opened 2024-02-09
🐛 [RUM-3039] Fix missing pending mutations at view end

amortemousque:
## Motivation

We are missing records generated during the `VIEW_END` lifecycle callbacks (if no segment is already created) because no view is active at that time:

- [Miss the pending mutations and view end record](https://github.com/DataDog/browser-sdk/blob/c1924aa362a56643992af7c9e8a30afeceb80ec4/packages/rum/src/domain/record/record.ts#L111)
- [Miss pending clicks](https://github.com/DataDog/browser-sdk/blob/c1924aa362a56643992af7c9e8a30afeceb80ec4/packages/rum-core/src/domain/action/trackClickActions.ts#L78)

This PR introduce new `BEFORE_XXX` lifecycle events where the SDK contexts are still active allowing to send last pending records

## Changes

- Introduce new prefixed `BEFORE_XXX` and `AFTER_XXX` lifecycle events used by contexts to ensures the context is available during the non prefixed event callbacks.
- Add view related contexts at `BEFORE_VIEW_CREATED`  and close at `AFTER_VIEW_ENDED`
- Stop the recorder at `PAGE_EXITED` to prevent unexpected records after the view is closed
- Cleanly stop `scrollObserver`, `moveObserver` and `instrumentSetter`

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/session/sessionManager.ts
+++ b/packages/core/src/domain/session/sessionManager.ts
@@ -6,6 +6,7 @@ import { relativeNow, clocksOrigin, ONE_MINUTE } from '../../tools/utils/timeUti
 import { DOM_EVENT, addEventListener, addEventListeners } from '../../browser/addEventListener'
 import { clearInterval, setInterval } from '../../tools/timer'
 import type { Configuration } from '../configuration'
+import type { TrackingConsentState } from '../trackingConsent'
 import { SESSION_TIME_OUT_DELAY } from './sessionConstants'
 import { startSessionStore } from './sessionStore'
 
@@ -29,7 +30,8 @@ let stopCallbacks: Array<() => void> = []
 export function startSessionManager<TrackingType extends string>(
   configuration: Configuration,
   productKey: string,
-  computeSessionState: (rawTrackingType?: string) => { trackingType: TrackingType; isTracked: boolean }
+  computeSessionState: (rawTrackingType?: string) => { trackingType: TrackingType; isTracked: boolean },
+  trackingConsentState: TrackingConsentState
 ): SessionManager<TrackingType> {
   const renewObservable = new Observable<void>()
   const expireObservable = new Observable<void>()
@@ -52,10 +54,24 @@ export function startSessionManager<TrackingType extends string>(
     expireObservable.notify()
   })
 
+  // We expand/renew session unconditionally as tracking consent is always granted when the session
+  // manager is started.
   sessionStore.expandOrRenewSession()
   sessionContextHistory.add(buildSessionContext(), clocksOrigin().relative)
 
-  trackActivity(configuration, () => sessionStore.expandOrRenewSession())
+  trackingConsentState.observable.subscribe(() => {
+    if (trackingConsentState.isGranted()) {
+      sessionStore.expandOrRenewSession()
+    } else {
+      sessionStore.expire()
+    }
+  })
+
+  trackActivity(configuration, () => {
+    if (trackingConsentState.isGranted()) {
+      sessionStore.expandOrRenewSession()
+    }
+  })
   trackVisibility(configuration, () => sessionStore.expandSession())
 
   function buildSessionContext() {
--- a/packages/core/src/domain/session/sessionManager.ts
+++ b/packages/core/src/domain/session/sessionManager.ts
@@ -13,7 +13,6 @@ import { startSessionStore } from './sessionStore'
 export interface SessionManager<TrackingType extends string> {
   findActiveSession: (startTime?: RelativeTime) => SessionContext<TrackingType> | undefined
   renewObservable: Observable<void>
-  beforeExpireObservable: Observable<void>
   expireObservable: Observable<void>
   expire: () => void
 }
@@ -35,7 +34,6 @@ export function startSessionManager<TrackingType extends string>(
 ): SessionManager<TrackingType> {
   const renewObservable = new Observable<void>()
   const expireObservable = new Observable<void>()
-  const beforeExpireObservable = new Observable<void>()
 
   // TODO - Improve configuration type and remove assertion
   const sessionStore = startSessionStore(configuration.sessionStoreStrategyType!, productKey, computeSessionState)
@@ -49,9 +47,8 @@ export function startSessionManager<TrackingType extends string>(
     renewObservable.notify()
   })
   sessionStore.expireObservable.subscribe(() => {
-    beforeExpireObservable.notify()
-    sessionContextHistory.closeActive(relativeNow())
     expireObservable.notify()
+    sessionContextHistory.closeActive(relativeNow())
   })
 
   // We expand/renew session unconditionally as tracking consent is always granted when the session
@@ -85,7 +82,6 @@ export function startSessionManager<TrackingType extends string>(
     findActiveSession: (startTime) => sessionContextHistory.find(startTime),
     renewObservable,
     expireObservable,
-    beforeExpireObservable,
     expire: sessionStore.expire,
   }
 }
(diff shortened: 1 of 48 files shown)


packages/rum-core/src/domain/lifeCycle.ts
```diff
@@ -10,14 +10,22 @@ import type { AutoAction } from './action/actionCollection'
...
 
 export const enum LifeCycleEventType {
+  // BEFORE_XXX events have been introduced to provide hooks just before significant actions or transitions in the application lifecycle.
+  // The following three events allow for performing tasks, such as sending last events, since the SDK contexts (viewContext, etc.) are still active.
+  BEFORE_VIEW_ENDED,
+  BEFORE_SESSION_EXPIRED,
+  BEFORE_PAGE_EXITED,
+  // The BEFORE_VIEW_CREATED event has been introduced to provide a hook just before a view is created,
+  // ensuring that the context related to the view (viewContexts, urlContexts) exists at VIEW_CREATED.
+  BEFORE_VIEW_CREATED,
```
bcaudan, line 20: **💬 suggestion:** Given that most of the listeners should not be related to context management, what about:
- using non-suffixed "XXX starts" / "XXX ends" events for listeners that are not managing contexts
- using suffixed "before XXX starts" / "after XXX ends" events for listeners that are managing contexts  

We would then have: contexts should be created before XXX starts and closed after XXX ends.

Wdyt?

amortemousque: Hum interesting. I'll give it a try

BenoitZugmeyer (approved):

packages/rum/src/domain/record/record.ts
```diff
@@ -44,9 +45,11 @@ export function record(options: RecordOptions): RecordAPI {
   const emitAndComputeStats = (record: BrowserRecord) => {
     emit(record)
     sendToExtension('record', { record })
-    const view = options.viewContexts.findView()
+    const view = options.viewContexts.findView()!
     if (view) {
       replayStats.addRecord(view.id)
+    } else {
+      addTelemetryError('[Record] no active view', { record } as unknown as Context)
```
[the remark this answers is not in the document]
amortemousque: It helped me detect what which observsers were not "properly" stopped: https://github.com/DataDog/browser-sdk/pull/2598/commits/8abb8ee44962fb8f318f4462d14ec96de5905122
But we can remove it now :)

packages/core/src/tools/instrumentMethod.ts
```diff
@@ -125,11 +126,14 @@ export function instrumentSetter<TARGET extends { [key: string]: any }, PROPERTY
...
+  const timeoutIds = new Set<TimeoutId>()
   let instrumentation = (target: TARGET, value: TARGET[PROPERTY]) => {
     // put hooked setter into event loop to avoid of set latency
-    setTimeout(() => {
+    const timeoutId = setTimeout(() => {
+      timeoutIds.delete(timeoutId)
       after(target, value)
     }, 0)
+    timeoutIds.add(timeoutId)
   }
```
BenoitZugmeyer, line 137: **💬 suggestion:** ‏This function is performance sensitive. To avoid manipulating a collection of timeouts, ou could have something like:

```ts
  let stopped = false
  let instrumentation = (target: TARGET, value: TARGET[PROPERTY]) => {
    // put hooked setter into event loop to avoid of set latency
    setTimeout(() => {
      if (!stopped) {
        after(target, value)
      }
    }, 0)
  }
  // ...
  stop() {
    // ...
    stopped = true
  }

File: packages/rum-core/src/domain/lifeCycle.ts (hunk 1 above)
Comment (amortemousque) [reply packages/rum-core/src/domain/lifeCycle.ts]: Let me know what you think ?

File: packages/core/src/domain/session/sessionManager.ts [hunk 4]
```diff
@@ -41,8 +44,10 @@ export function startSessionManager<TrackingType extends string>(
 
   sessionStore.renewObservable.subscribe(() => {
     sessionContextHistory.add(buildSessionContext(), relativeNow())
+    renewObservable.notify()
```
bcaudan, line 47: **💬 suggestion:** ‏what about adding some tests to express the relationship between observables and contexts?

packages/core/src/tools/instrumentMethod.ts
```diff
@@ -125,10 +125,13 @@ export function instrumentSetter<TARGET extends { [key: string]: any }, PROPERTY
...
 
+  const stoppedInstrumentation = noop
   let instrumentation = (target: TARGET, value: TARGET[PROPERTY]) => {
     // put hooked setter into event loop to avoid of set latency
     setTimeout(() => {
-      after(target, value)
+      if (instrumentation !== stoppedInstrumentation) {
+        after(target, value)
+      }
```
bcaudan, line 134: **💬 suggestion:** ‏Could we have a test expressing the purpose of this change?

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -317,6 +317,22 @@ describe('view lifecycle', () => {
       expect(getViewCreateCount()).toEqual(1)
     })
   })
+
+  it('should notify VIEW_CREATED before VIEW_CREATED', () => {
```
bcaudan, line 321: **💬 suggestion:** ‏
```suggestion
  it('should notify BEFORE_VIEW_CREATED before VIEW_CREATED', () => {
```

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -317,6 +317,22 @@ describe('view lifecycle', () => {
...
+    setupBuilder.build()
+
+    expect(notifySpy.calls.argsFor(0)[0]).toEqual(LifeCycleEventType.BEFORE_VIEW_CREATED)
+    expect(notifySpy.calls.argsFor(1)[0]).toEqual(LifeCycleEventType.VIEW_CREATED)
+  })
+
+  it('should notify BEFORE_VIEW_ENDED before VIEW_ENDED', () => {
```
bcaudan, line 328: **💬 suggestion:** ‏need to be updated as well 🙂

packages/rum/src/boot/startRecording.spec.ts
```diff
@@ -173,6 +172,20 @@ describe('startRecording', () => {
     expect(secondSegment.records[0].type).toBe(RecordType.Meta)
   })
 
+  it('send view end record when the segment collection is waiting for the initial record', async () => {
```
bcaudan, line 175: **💬 suggestion:** ‏Could we clarify the intent of this test?

amortemousque: What about: 'flushes pending mutations before ending the view, even after the segment has been flushed'?

packages/core/src/tools/instrumentMethod.ts
amortemousque: https://github.com/DataDog/browser-sdk/blob/b1733892001f9e2baf0e5ee75b7228f6dc3d81bb/packages/core/src/tools/instrumentMethod.spec.ts#L270

packages/core/src/domain/session/sessionManager.ts
amortemousque: https://github.com/DataDog/browser-sdk/blob/b1733892001f9e2baf0e5ee75b7228f6dc3d81bb/packages/core/src/domain/session/sessionManager.spec.ts#L501-L524

bcaudan (approved):

PR #2602, opened 2024-02-12
:recycle: Rename error var

N-Boutaib:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/error/trackRuntimeError.ts
+++ b/packages/core/src/domain/error/trackRuntimeError.ts
@@ -11,7 +11,7 @@ export type UnhandledErrorCallback = (stackTrace: StackTrace, originalError?: an
 
 export function trackRuntimeError(errorObservable: Observable<RawError>) {
   const handleRuntimeError = (stackTrace: StackTrace, originalError?: any) => {
-    const test = computeRawError({
+    const rawError = computeRawError({
       stackTrace,
       originalError,
       startClocks: clocksNow(),
@@ -19,7 +19,7 @@ export function trackRuntimeError(errorObservable: Observable<RawError>) {
       source: ErrorSource.SOURCE,
       handling: ErrorHandling.UNHANDLED,
     })
-    errorObservable.notify(test)
+    errorObservable.notify(rawError)
   }
   const { stop: stopInstrumentingOnError } = instrumentOnError(handleRuntimeError)
   const { stop: stopInstrumentingOnUnhandledRejection } = instrumentUnhandledRejection(handleRuntimeError)


BenoitZugmeyer:
PR looks good, but I think we should collect error causes for other errors collected by the Logs SDK:
* console.error instrumentation -> https://github.com/DataDog/browser-sdk/blob/ed481844e1d44e1d7adf6758f0aa3c4e81f219f7/packages/logs/src/domain/console/consoleCollection.ts#L32
* runtime errors -> https://github.com/DataDog/browser-sdk/blob/ed481844e1d44e1d7adf6758f0aa3c4e81f219f7/packages/logs/src/domain/runtimeError/runtimeErrorCollection.ts#L31

I think that's it?

Maybe a function to format a rawError into log `error.*` properties could be nice, so we have `cause`, `fingerprint` and `kind` properly defined everywhere as well.

amortemousque:
**💭 thought:** ‏ It's not directly related to the PR but looking at how we collect errors in Logs it seems we have some discrepancies:
- error.fingerprint only set in `console.error` but not in `logger.error`
- error.message only set in `logger.error` but not in other error collections
- error.kind not set in `console.error`

It could be nice to have a more centralize ways to format [Error](https://github.com/DataDog/browser-sdk/blob/d650679d93bc6bd8fdbd1cf0db19f7bcf6dd42fa/packages/logs/src/rawLogsEvent.types.ts#L12). You could introduce something in this PR like mentioned by @BenoitZugmeyer (cf [comment](https://github.com/DataDog/browser-sdk/pull/2602#issuecomment-1941516021)) or maybe in a following PR :)

packages/logs/src/domain/logger.ts
amortemousque, line 1: **💬 suggestion:** 
We could improve the typing of `let errorContext: LogsEvent['error']` by 
`let errorContext: RawLoggerLogsEvent['error']`

Then in `rawLogsEvent.types.ts` update the Error type like so:
```typescript
type Error = {
  kind?: string
  stack?: string
  fingerprint?: string
  message?:string
  causes?: Array<{
    message: string
    source: string
    type?: string
    stack?: string
  }>
}
```

The `logsEvent.types.ts` error could also be updated to add the new field and the others that we have forgotten along the way:

```typescript
error?: {
    /**
     * Kind of the error
     */
    kind?: string
    /**
     * Stacktrace of the error
     */
    stack?: string
    /**
     * Fingerprint of the error
     */
    fingerprint?: string
    /**
     * Message of the error
     */
     message?:string
     /**
     * Causes of the error
     */
    causes?: Array<{
      message: string
      readonly type?: string
      stack?: string
      readonly source: 'network' | 'source' | 'console' | 'logger' | 'agent' | 'webview' | 'custom' | 'report'
    }>
    [k: string]: unknown
  }
```

PR #2605, opened 2024-02-12
✨[RUM-3151] Report bundle sizes to logs

RomanGaignault:
## Motivation

To send bundle sizes to logs and check the evolution with dashboard.

## Changes

Added a script to retrieve git commit Id, version, branch and bundle size then send them to logs with fetch request.
image

## Testing

- [x] Local
- [x] Staging
- [x] CI
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,3 @@ browserstack.err
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
-
-.DS_Store
-.vscode/*
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ browserstack.err
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+.DS_Store
+.vscode/*
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -136,6 +136,7 @@ build-and-lint:
     - .test-allowed-branches
   interruptible: true
   script:
+    - echo "Target branch: $CI_MERGE_REQUEST_TARGET_BRANCH_NAME"
     - yarn
     - yarn build
     - yarn lint
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -136,7 +136,7 @@ build-and-lint:
     - .test-allowed-branches
   interruptible: true
   script:
-    - echo "Target branch: $CI_MERGE_REQUEST_TARGET_BRANCH_NAME"
+    - echo ${CI_MERGE_REQUEST_TARGET_BRANCH_NAME}
     - yarn
     - yarn build
     - yarn lint
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -135,10 +135,7 @@ build-and-lint:
     - .base-configuration
     - .test-allowed-branches
   interruptible: true
-  only:
-    - merge_requests
   script:
-    - echo $CI_MERGE_REQUEST_TARGET_BRANCH_NAME
     - yarn
     - yarn build
     - yarn lint
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -135,6 +135,9 @@ build-and-lint:
     - .base-configuration
     - .test-allowed-branches
   interruptible: true
+  only:
+    - merge_requests
+    - branches
   script:
     - echo $CI_MERGE_REQUEST_TARGET_BRANCH_NAME
     - yarn
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -136,7 +136,7 @@ build-and-lint:
     - .test-allowed-branches
   interruptible: true
   script:
-    - echo ${CI_MERGE_REQUEST_TARGET_BRANCH_NAME}
+    - echo $CI_MERGE_REQUEST_TARGET_BRANCH_NAME
     - yarn
     - yarn build
     - yarn lint
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -137,7 +137,6 @@ build-and-lint:
   interruptible: true
   only:
     - merge_requests
-    - branches
   script:
     - echo $CI_MERGE_REQUEST_TARGET_BRANCH_NAME
     - yarn
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -140,6 +140,7 @@ build-and-lint:
     - yarn build
     - yarn lint
     - node scripts/check-packages.js
+    - node scripts/export-bundles-sizes.js
 
 build-bundle:
   extends:
--- a/scripts/export-bundles-sizes.js
+++ b/scripts/export-bundles-sizes.js
@@ -31,7 +31,7 @@ runMain(async () => {
       },
       version: getVersion(),
       commit: getGitInformation('git rev-parse HEAD'),
-      branch: getGitInformation('git rev-parse --abbrev-ref HEAD'),
+      branch: process.env.GITHUB_REF ? process.env.GITHUB_REF.split('/').pop() : null,
     },
   ]
   await postBundleSize(URL, logData)
--- a/scripts/export-bundles-sizes.js
+++ b/scripts/export-bundles-sizes.js
@@ -30,14 +30,14 @@ runMain(async () => {
         worker: getBundleSize(workerPath),
       },
       version: getVersion(),
-      commit: getGitInformation('git rev-parse HEAD'),
-      branch: process.env.CI_MERGE_REQUEST_TARGET_BRANCH_NAME,
+      commit: executeGitCommand('git rev-parse HEAD'),
+      branch: process.env.CI_COMMIT_REF_NAME,
     },
   ]
   await postBundleSize(URL, logData)
 })
 
-function getGitInformation(command) {
+function executeGitCommand(command) {
   try {
     return execSync(command)
       .toString('utf8')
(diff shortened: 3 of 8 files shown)


scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function getBundleSize(pathBundle) {
+  try {
+    const file = fs.statSync(pathBundle)
+    return file.size
+  } catch (error) {
+    console.error('Failed to get bundle size:', error)
+    return null
+  }
```
BenoitZugmeyer, line 68: **💬 suggestion:** ‏logging the error is nice, but the script should abort if an error occurs. What we could to is to leverage the [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) to store the log message and the cause like this:

```suggestion
  } catch (error) {
    throw new Error('Failed to get bundle size', { cause: error })
  }
```

And then in `runMain`, display the causes as well as the error ([here](https://github.com/DataDog/browser-sdk/blob/ed481844e1d44e1d7adf6758f0aa3c4e81f219f7/scripts/lib/execution-utils.js#L29)).

(Same for other try/catch in this file)

scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,77 @@
...
+const logsPath = path.join(__dirname, '../packages/logs/bundle/datadog-logs.js')
+const rumSlimPath = path.join(__dirname, '../packages/rum-slim/bundle/datadog-rum-slim.js')
+const workerPath = path.join(__dirname, '../packages/worker/bundle/worker.js')
+const versionPath = path.join(__dirname, '../packages/rum/package.json')
+
+const URL = 'https://http-intake.logs.datadoghq.com/api/v2/logs'
+const HEADERS = {
+  'DD-API-KEY': getOrg2ApiKey(),
+  'Content-Type': 'application/json',
+}
```
BenoitZugmeyer, line 17: **💬 suggestion:** ‏either inline those values in the function, or use more explicit names, like:

```
const LOG_INTAKE_URL =
const LOG_INTAKE_REQUEST_HEADERS =
```

scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,77 @@
...
+  try {
+    const file = fs.statSync(pathBundle)
+    return file.size
+  } catch (error) {
+    console.error('Failed to get bundle size:', error)
+    return null
+  }
+}
+
+async function postBundleSize(url = '', bundleData = {}) {
```
BenoitZugmeyer, line 71: **💬 suggestion:** ‏rename this function to reflect what it does instead of what it is used for. Ex:

```suggestion
async function sendLogToOrg2(url = '', bundleData = {}) {
```

**💬 suggestion:** ‏for consistency, remove the `url` parameter so it's similar to `HEADERS`: either use the constant directly in the function, or inline its value

scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,77 @@
...
+      ddsource: 'browser-sdk',
+      env: 'ci',
+      bundle_sizes: {
+        rum: getBundleSize(rumPath),
+        logs: getBundleSize(logsPath),
+        rum_slim: getBundleSize(rumSlimPath),
+        worker: getBundleSize(workerPath),
+      },
+      version: getVersion(),
+      commit: executeGitCommand('git rev-parse HEAD'),
```
BenoitZugmeyer, line 33: **💬 suggestion:** ‏use a gitlab environment variable like you do for the branch. As an alternative, use the [`command` helper](https://github.com/DataDog/browser-sdk/blob/ed481844e1d44e1d7adf6758f0aa3c4e81f219f7/scripts/lib/command.js#L19)

scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,77 @@
...
+
+function getVersion() {
+  try {
+    const versionJson = fs.readFileSync(versionPath, 'utf8')
+    return JSON.parse(versionJson).version
+  } catch (error) {
+    console.error('Failed to get version:', error)
+    return null
+  }
+}
```
BenoitZugmeyer, line 59: **💬 suggestion:** ‏it seems we get the version in a few scripts already:
* scripts/release/update-peer-dependency-versions.js
* scripts/release/generate-changelog.js
* scripts/lib/build-env.js

Could you refactor this by creating a `getBrowserSdkVersion()` function in `scripts/lib/browser-sdk-version.js`?

Also, please use `lerna.json` instead of `package.json` as it is the main source of truth.

scripts/lib/execution-utils.js
```diff
@@ -27,6 +27,7 @@ function runMain(mainFunction) {
     .catch((error) => {
       printError('\nScript exited with error:')
       printError(error)
+      printError(error.cause)
```
BenoitZugmeyer, line 30: **💬 suggestion:** ‏I think we can do a bit better, because not all errors have a cause (so your code will print "undefined" most of the time), and error causes can also have a cause (it's recursive). Let's create a new function, `printErrorWithCause(error)` that prints error causes. Usage example:

```ts
const error = new Error('Foo', { cause: new Error('Bar', { cause: new Error('Baz') }) })
printErrorWithCause(error)
// Prints:
// Error: Foo
// Caused by: Error: Bar
// Caused by: Error: Baz
```

N-Boutaib (approved):
Nice one 🎉

.gitignore
```diff
@@ -17,3 +17,6 @@ browserstack.err
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+.DS_Store
+.vscode/*
```
bcaudan, line 22: **💬 suggestion:** ‏you should consider [setup a global .gitignore](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) instead

.gitlab-ci.yml
```diff
@@ -140,6 +140,7 @@ build-and-lint:
     - yarn build
     - yarn lint
     - node scripts/check-packages.js
+    - node scripts/export-bundles-sizes.js
```
bcaudan, line 143: **❓ question:** ‏do we plan to export those metrics on all branches or only on main?

scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,54 @@
+const fs = require('fs')
+const path = require('path')
+const { getBrowserSdkVersion } = require('./lib/browser-sdk-version')
+const { getOrg2ApiKey } = require('./lib/secrets')
+const { runMain, fetch: fetchWrapper } = require('./lib/execution-utils')
```
bcaudan, line 5: **🥜 nitpick:** any reason to not use the exported name directly?
‏
```suggestion
const { runMain, fetch } = require('./lib/execution-utils')
```

scripts/export-bundles-sizes.js
```diff
@@ -0,0 +1,54 @@
...
+      message: 'Browser SDK bundles sizes',
+      service: 'browser-sdk',
+      ddsource: 'browser-sdk',
+      env: 'ci',
+      bundle_sizes: {
+        rum: getBundleSize(rumPath),
+        logs: getBundleSize(logsPath),
+        rum_slim: getBundleSize(rumSlimPath),
+        worker: getBundleSize(workerPath),
+      },
```
bcaudan, line 30: **💭 thought:** ‏could it be interesting to monitor all packages instead of only the bundles?

scripts/lib/browser-sdk-version.js
```diff
@@ -0,0 +1,9 @@
+const lernaJson = require('../../lerna.json')
+
+function getBrowserSdkVersion() {
+  return lernaJson.version
+}
+
+module.exports = {
+  getBrowserSdkVersion,
+}
```
bcaudan, line 9: **🥜 nitpick:** ‏what about exporting the version directly instead of a getter?

.gitlab-ci.yml
RomanGaignault: it's only on main, dashboard will be set to only take metrics when something is merge to main

RomanGaignault: it will be filtered with the branch attribut of the request

scripts/export-bundles-sizes.js
RomanGaignault: I thought it would be clearer to specify the type of fetch.

.gitlab-ci.yml
bcaudan: ok, it is clearer.
Maybe nevermind for now and let's see later with PR reporting what seems to be more convenient between having the branch logic inside the script or in the gitlab configuration 🤷‍♂️

scripts/export-bundles-sizes.js
RomanGaignault: I think you're right It's a bit superfluous, I will change to only fetch

scripts/lib/browser-sdk-version.js
```diff
@@ -0,0 +1,5 @@
+const lernaJson = require('../../lerna.json')
+
+module.exports = {
+  getBrowserSdkVersion: lernaJson.version,
```
bcaudan, line 4: **🥜 nitpick:** since we directly expose the value, it is not really a getter anymore ‏
```suggestion
  browserSdkVersion: lernaJson.version,
```

bcaudan (approved):

bcaudan (approved):

PR #2606, opened 2024-02-12
⚗️[RUM-2889] custom vitals improvements

bcaudan:
## Motivation

Following #2591, add more features around custom vitals

## Changes

- 🐛 Fix multiple vital creation with multiple stop calls
- ✨ Add event limiter for vitals
- ✨ Add new API options:

```js
DD_RUM.startDurationVital('foo', { context, startTime })
// do some stuff
DD_RUM.stopDurationVital('foo', { context, stopTime })
```

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/vital/vitalCollection.ts
+++ b/packages/rum-core/src/domain/vital/vitalCollection.ts
@@ -1,5 +1,5 @@
-import type { ClocksState, Duration } from '@datadog/browser-core'
-import { elapsed, generateUUID } from '@datadog/browser-core'
+import type { ClocksState, Duration, Context } from '@datadog/browser-core'
+import { combine, elapsed, generateUUID } from '@datadog/browser-core'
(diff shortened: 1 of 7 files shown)


packages/rum-core/src/domain/vital/vitalCollection.ts
amortemousque, line 1: Should we have a limitation in case a vital is started but never stopped?

bcaudan: I want to look into that in my next PR, see if we should do something when the session expires or the page is frozen.
Do you have other things in mind?

amortemousque: Your plan sounds good. Let's see in the next PR ;)

amortemousque (approved):

PR #2610, opened 2024-02-16
⚗️[RUM-3234] Discard outdated vitals

bcaudan:
## Motivation

Avoid to report inflated vital values and to keep started vitals indefinitely in memory. 

## Changes

- ♻️ Extract pageHistory.wasInPageStateDuringPeriod
- ✨ discard a vital for which a frozen state happened
- ✨ discard pending vitals on SESSION_RENEWED

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


N-Boutaib (approved):
LGTM

packages/rum-core/src/domain/contexts/pageStateHistory.ts
```diff
@@ -31,6 +31,7 @@ export type PageStateEntry = { state: PageState; startTime: RelativeTime }
 export interface PageStateHistory {
   findAll: (startTime: RelativeTime, duration: Duration) => PageStateServerEntry[] | undefined
   isInActivePageStateAt: (startTime: RelativeTime) => boolean
+  wasInPageStateDuringPeriod: (state: PageState, startTime: RelativeTime, duration: Duration) => boolean
```
amortemousque, line 34: **💬 suggestion:** ‏now that we have `wasInPageStateDuringPeriod` it could be nice to get rid of  `isInActivePageStateAt` wdyt?

bcaudan: using `wasInPageStateDuringPeriod(ACTIVE, startTime, 0)` for those cases may not be super explicit with the `0` and the fact that we don't really care about the period there.

What about exposing:

```js
pageStateHistory = {
  wasInPageStateAt: (state: PageState, startTime: RelativeTime) =>
    pageStateHistory.wasInPageStateDuringPeriod(state, startTime, 0 as Duration),
  ...
}
```

amortemousque: Looks good!

amortemousque (approved):

BenoitZugmeyer (approved):

PR #2621, opened 2024-02-21
React SDK

Miz85:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


index.html
bcaudan, line 1: **❓ question:** what is the purpose of this file? Would it be more suited inside the react or sandbox directory?‏

packages/react/src/entries/main.ts
```diff
@@ -0,0 +1,29 @@
...
+  computeViewName: (routeMatches: RouteMatch[]) => string = _computeViewName
+) => {
+  router.subscribe((state: RouterState) => {
+    const viewName = computeViewName(state.matches)
+    datadogRum.startView(viewName)
+  })
+  return router
+}
+
+export const datadogReactRum = Object.assign({}, datadogRum, {
```
bcaudan, line 17: **💭 thought:** ‏I am wondering if we should not override the global variable DD_RUM in this case.
cf https://github.com/DataDog/browser-sdk/blob/7b87e74044cdd92b34383fd3ce48e953054f6597/packages/rum/src/entries/main.ts#L40

PR #2627, opened 2024-02-29
expose `isAborted` property to `beforeSend` callback

### LINKED ISSUE #362: How to filter out aborted XHR requests from the error logs
We frequently see several logs with this pattern:

Message: `XHR error GET https://url-of-ajax-endpoint`
Stack trace: `Failed to load`

What we discovered is that these are not actually errors but are aborted XHR requests. We use a library called select2 which makes XHR requests as the user types an input. It's not uncommon for a single user interaction to generate several requests as they type their input string.

The select2 library is aborting in-flight XHR requests before it makes a new one as described here https://stackoverflow.com/a/37319150/7691972

I am curious what the best way is to either prevent these from being logged, or to distinguish these from other legitimate errors that may present similarly.

swseverance:
resolves #362

## Motivation

The current event passed to the `beforeSend` callback cannot differentiate (to my knowledge) between a network request that was aborted and a network request that failed (for example, tiimed out). The PR makes the `isAborted` property available to the callback so users can decide if they want to prevent cancelled requests from being sent to the logs.

## Changes

Exposes a new property `isAborted` on `event.http` object

## Testing

- [X] Local
- [X] Staging
- [X] Unit
- [X] End to end

---

### DIFF
--- a/packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
+++ b/packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
@@ -78,6 +78,7 @@ describe('network error collection', () => {
           method: 'GET',
           status_code: 503,
           url: 'http://fake.com/',
+          isAborted: false
         },
       })
       done()
--- a/packages/logs/src/domain/networkError/networkErrorCollection.ts
+++ b/packages/logs/src/domain/networkError/networkErrorCollection.ts
@@ -56,6 +56,7 @@ export function startNetworkErrorCollection(configuration: LogsConfiguration, li
             method: request.method as any, // Cast resource method because of case mismatch cf issue RUMF-1152
             status_code: request.status,
             url: request.url,
+            isAborted: request.isAborted
           },
           status: StatusType.error,
           origin: ErrorSource.NETWORK,
--- a/packages/logs/src/logsEvent.types.ts
+++ b/packages/logs/src/logsEvent.types.ts
@@ -106,6 +106,8 @@ export interface LogsEvent {
      */
     url: string
 
+    isAborted: boolean
+
     [k: string]: unknown
   }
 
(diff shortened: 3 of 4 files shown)


bcaudan:
Hi @swseverance,

Not sure that we want to promote the `isAborted` property to an out of the box attribute just yet.
Another approach could be to expose this property inside a beforeSend context (similar to what is done on the [RUM side](https://docs.datadoghq.com/real_user_monitoring/guide/enrich-and-control-rum-data/?tab=event#context)) in order to allow users to do whatever they want with this information.

It would look like:

```js
  if (event.type === 'error' && event.origin === 'network') {
    event.http.is_aborted = event.context.isAborted
  }
},
```

wdyt?

swseverance:
hi @bcaudan 

Thank  you for looking at my PR. As an end user of the library I would be happy with the approach you described above (adding a second `context` parameter to the `beforeSend` callback). However, implementing such an approach would perhaps require more time and analysis than I can dedicate at the moment. For example, we would need to think "What interface should the new `Context` object implement?".  Looking at the RUM [Context](https://github.com/DataDog/browser-sdk/blob/b43623f43019f3e93e491630f6e62ccf93cb4da8/packages/rum-core/src/domainContext.types.ts#L7) as an example I can see that it is not trivial.

bcaudan:
Indeed, it may require more involvement, let us know if you'd be interested do it.
Otherwise, we'll discuss this change internally and try to address it at some point.

swseverance:
@bcaudan  Unfortunately I do not have the time to make the change myself. I'm sure your team will implement a good solution in the future. Thanks!

bcaudan:
Closing this PR for now then and we'll communicate progress on the related issue.

thomas-lebeau:
fixed in #2682 #2665

PR #2635, opened 2024-03-07
✨ [RUM-2885] Collect CSP disposition

thomas-lebeau:
## Motivation

Collect disposition data for rum error in order to differentiate between blocking csp violation (`enforce`) and report-only (`report`).

## Changes

Add a new property `csp.disposition` to the collected CSP violation errors. (Depends on https://github.com/DataDog/rum-events-format/pull/187)

Screenshot 2024-03-08 at 09 04 20

## Testing

_wip: I'm figuring out how to test on staging and I'll update the steps here._

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/rumEvent.types.ts
```diff
@@ -381,6 +381,16 @@ export type RumErrorEvent = CommonProperties &
...
+      /**
+       * Content Security Violation properties
+       */
+      readonly csp?: {
+        /**
+         * In the context of CSP errors, indicates how the violated policy is configured to be treated by the user agent.
+         */
+        readonly disposition?: 'enforce' | 'report'
+        [k: string]: unknown
+      }
```
bcaudan, line 393: **💬 suggestion:** ‏given that there is no update of the rum-events-format reference in the PR, I am guessing that you have updated this file manually.
You should run `yarn rum-events-format:sync` to update all the event format types with the latest changes.

packages/core/src/index.ts
```diff
@@ -134,5 +134,6 @@ export { ErrorSource } from './domain/error/error.types'
 export { RawError } from './domain/error/error.types'
 export { RawErrorCause } from './domain/error/error.types'
 export { ErrorWithCause } from './domain/error/error.types'
+export { Csp } from './domain/error/error.types'
```
bcaudan, line 137: **🥜 nitpick:** ‏it was already the case but it seems a bit weird to add one export by type, maybe we could enforce the grouping somehow to have:
```js
export { ErrorHandling, ErrorSource, ... } from './domain/error/error.types'
```

test/e2e/scenario/rum/errors.scenario.ts
```diff
@@ -121,35 +121,82 @@ describe('rum errors', () => {
...
+            /^script-src(-elem)?: 'https:\/\/example\.com\/foo\.js' blocked by 'script-src(-elem)?' directive of the policy/,
+            `  at <anonymous> @ ${baseUrl}/:`,
+          ],
+          handling: 'unhandled',
+          csp: {
+            disposition: 'enforce',
+          },
+        })
+        await withBrowserLogs((browserLogs) => {
+          expect(browserLogs.length).toEqual(1)
```
bcaudan, line 159: **🥜 nitpick:** ‏what about adding an expect on the log message or a comment to make more explicit which log we are expecting?

bcaudan (commented):
Nice!

packages/core/src/index.ts
thomas-lebeau: ✅ done! I think this is the kind of things that could be automated. Any thoughts on [eslint/sort-imports](https://eslint.org/docs/latest/rules/sort-imports) rule?

test/e2e/scenario/rum/errors.scenario.ts
thomas-lebeau: hmm, I'm not actually sure there is value to check that the browser logs in this case.

I wrote this test looking at the other ones on this file and I think this check is meant to verify that we do not generate any other errors unintentionally.

Also, error message tend to be widely inconsistent between browser vendor, which will make this test even more complex to make work on Safari and FireFox

bcaudan: > I wrote this test looking at the other ones on this file and I think this check is meant to verify that we do not generate any other errors unintentionally.

Oh I see, IIRC we do those assertions to avoid to have these expected errors logged in the test output (see [test teardown](https://github.com/DataDog/browser-sdk/blob/main/test/e2e/lib/framework/createTest.ts#L217)).

So yes, no need to add more details about the log is generated by the test.

bcaudan (approved):

BenoitZugmeyer (approved):

PR #2639, opened 2024-03-11
[ci-visibility] Implement driver-agnostic integration with CI Visibility

nikita-tkachenko-datadog:
## Motivation
Allow CI Visibility users who use Selenium for testing their RUM-enabled pages to have their RUM sessions linked to their test executions.

## Changes

- Get the ci visibility `test_execution_id` from cookie (it will be set by CI Visibility logic in the Datadog tracers using Selenium integration).
- Create a `cookieObservable` that uses the [cookieStore API](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore) to be notified when the ci visibility cookie is set

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/contexts/ciTestContext.ts
```diff
@@ -5,6 +5,10 @@ export interface CiTestWindow extends Window {
 }
 
 export function getCiTestContext() {
+  if (window['dd_ci_visibility_test_context']) {
+    return window['dd_ci_visibility_test_context'];
```
juan-fernandez, line 9: what we include here is what we end up using in the UI to query for RUM sessions, so I think it'd be better to be explicit about what we're including. Something like:

```javascript 
if (window['dd_ci_visibility_test_context']) {
  return {
    test_execution_id: window['dd_ci_visibility_test_context'].testExecutionId,
  }
}
```

this means that we have a hard contract between selenium and RUM but I think that's actually a good thing 😄

packages/rum-core/src/domain/assembly.ts
```diff
@@ -162,7 +161,7 @@ export function startRumAssembly(
           },
           action: needToAssembleWithAction(rawRumEvent) && actionId ? { id: actionId } : undefined,
           synthetics: syntheticsContext,
-          ci_test: ciTestContext,
+          ci_test: getCiTestContext(),
```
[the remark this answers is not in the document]
nikita-tkachenko-datadog: moved `ciTestContext` declaration inside the `lifeCycle.subscribe` callback

packages/rum-core/src/domain/contexts/ciTestContext.ts
nikita-tkachenko-datadog: My intention was to let us change the context without having to modify this function in the future.
Although it's not a big deal, changes as per your suggestion

packages/rum-core/src/domain/contexts/ciTestContext.ts
```diff
@@ -5,8 +5,13 @@ export interface CiTestWindow extends Window {
 }
 
 export function getCiTestContext() {
-  const testExecutionId = (window as CiTestWindow).Cypress?.env('traceId')
+  if (window['dd_ci_visibility_test_context']) {
+    return {
+      test_execution_id: window['dd_ci_visibility_test_context'].test_execution_id,
+    }
+  }
```
juan-fernandez, line 12:
```suggestion
  if (window.dd_ci_visibility_test_context) {
    return {
      test_execution_id: window.dd_ci_visibility_test_context.test_execution_id,
    }
  }
```

juan-fernandez (approved):
LGTM. Let's see what the rum-browser team thinks :)

packages/rum-core/src/domain/contexts/ciTestContext.ts
```diff
@@ -5,8 +5,13 @@ export interface CiTestWindow extends Window {
 }
```
juan-fernandez, line 5: I think you can simply add this to the interface
```suggestion
  dd_ci_visibility_test_context?: {
    test_execution_id: string
  }
}
```

packages/rum-core/src/domain/contexts/ciTestContext.ts
juan-fernandez: @nikita-tkachenko-datadog I updated the suggestion above

juan-fernandez (commented):
you might also need to modify `packages/rum-core/test/mockCiVisibilityWindowValues.ts` which is used in a couple of tests `packages/rum-core/src/domain/contexts/ciTestContext.spec.ts` and `packages/rum-core/src/domain/assembly.spec.ts`

packages/rum-core/src/browser/cookieObservable.ts
amortemousque, line 1: This observable could be used in to watch the session cookie at some point.

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -0,0 +1,64 @@
...
+function listenToCookieStoreChange(
+  configuration: Configuration,
+  cookieName: string,
+  callback: (event: CookieChange) => void
+) {
+  if (!(window as CookieStoreWindow).cookieStore) {
+    return
+  }
+
+  const listener = addEventListener(configuration, (window as CookieStoreWindow).cookieStore!, 'change', (event) => {
```
bcaudan, line 36: **💬 suggestion:** ‏use existing constant
```suggestion
  const listener = addEventListener(configuration, (window as CookieStoreWindow).cookieStore!, DOM_EVENT.CHANGE, (event) => {
```

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -0,0 +1,64 @@
...
+    event.changed
+      .concat(event.deleted)
+      .filter((change) => change.name === cookieName)
+      .forEach((change) => {
+        callback({
+          name: change.name,
+          value: change.value,
+        })
+      })
+  })
```
bcaudan, line 46: **❓ question:** ‏are we sure that we can't have ordering issues if a single event contains multiple changes of a single cookie?

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const WATCH_COOKIE_INTERVAL_DELAY = ONE_SECOND
+
+function watchCookieFallback(
+  cookieName: string,
+  callback: (event: { name: string; value: string | undefined }) => void
+) {
+  const watchCookieIntervalId = setInterval(() => {
+    callback({ name: cookieName, value: findCommaSeparatedValue(document.cookie, cookieName) })
+  }, WATCH_COOKIE_INTERVAL_DELAY)
```
bcaudan, line 59: **💬 suggestion:** in order to have a more consistent behaviour between the strategies, ‏what about only calling the callback if the value effectively changed?

packages/rum-core/src/browser/cookieObservable.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    subscription = observable.subscribe((cookieChange) => {
+      expect(cookieChange).toEqual({
+        name: COOKIE_NAME,
+        value: 'foo',
+      })
+
+      done()
+    })
+    setCookie(COOKIE_NAME, 'foo', COOKIE_DURATION)
+    clock.tick(STORAGE_POLL_DELAY)
```
bcaudan, line 42: **💬 suggestion:** ‏any reason to be related to session store?
```suggestion
    clock.tick(WATCH_COOKIE_INTERVAL_DELAY)
```

packages/rum-core/src/browser/cookieObservable.ts
amortemousque: I tried with the both JS APIs but I never managed to get the same cookie twice in the same table. In fact, I got one change event for each API calls:
```js
cookieStore.addEventListener('change', (event) => {
  if ((event.changed || []).some((c) => c.name === 'datadog-ci-visibility-test-execution-id')) {
    console.log(event.changed)
   }
})

document.cookie = 'datadog-ci-visibility-test-execution-id=foo; SameSite=None; Secure'
document.cookie = 'datadog-ci-visibility-test-execution-id=bar; SameSite=None; Secure'
document.cookie = 'datadog-ci-visibility-test-execution-id=baz; SameSite=None; Secure'

cookieStore.set({
  name: 'datadog-ci-visibility-test-execution-id',
  value: 'foo'
})

cookieStore.set({
  name: 'datadog-ci-visibility-test-execution-id',
  value: 'bar'
})
```

amortemousque: I could not find a clear answer to this question in the [spec](https://wicg.github.io/cookie-store/). So I asked for [clarification](https://github.com/WICG/cookie-store/issues/226). However I don't think it's a blocker for the current usage of the observable since the `datadog-ci-visibility-test-execution-id` will only be set once. Wdyt?

bcaudan: Sure, it could be nice to add a comment in the implementation though.

amortemousque: Done: https://github.com/DataDog/browser-sdk/blob/61410543b5e89c7e410bcd74dcff9c2d134f7309/packages/rum-core/src/browser/cookieObservable.ts#L37-L38

packages/rum-core/src/domain/contexts/ciVisibilityContext.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(ciVisibilityContext.get()).toEqual({
+      test_execution_id: 'trace_id_value',
+    })
+  })
+
+  it('sets the ci visibility context defined by global cookie', () => {
+    mockCiVisibilityValues('trace_id_value', 'cookies')
+    ciVisibilityContext = startCiVisibilityContext({} as Configuration, cookieObservable)
+
+    cookieObservable.notify({ name: CI_VISIBILITY_TEST_ID_COOKIE_NAME, value: 'trace_id_value' })
```
BenoitZugmeyer, line 34: **💬 suggestion**: design this test to get the trace id from `getInitCookie` instead of the observable (in other words, you shouldn't need to `notify` from a cookie change to get it to work)

packages/rum-core/src/domain/contexts/ciVisibilityContext.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+import { mockCiVisibilityValues } from '../../../test'
+import type { CiVisibilityContext } from './ciVisibilityContext'
+import { CI_VISIBILITY_TEST_ID_COOKIE_NAME, startCiVisibilityContext } from './ciVisibilityContext'
+
+describe('startCiVisibilityContext', () => {
+  let ciVisibilityContext: CiVisibilityContext
+  let cookieObservable: Observable<{
+    name: string
+    value: string | undefined
+  }>
```
BenoitZugmeyer, line 12: **💬 suggestion:** ‏import `CookieObservable` from `cookieObservable.ts`

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export interface CookieStoreWindow extends Window {
+  cookieStore?: CookieStore
+}
+
+export type CookieObservable = ReturnType<typeof createCookieObservable>
+
+export type CookieChange = { name: string; value: string | undefined }
+
+export function createCookieObservable(configuration: Configuration, cookieName: string) {
```
BenoitZugmeyer, line 20: **💬 suggestion:** ‏Since the `CookieObservable` is scoped to a single cookie name, it could simply notify the value (`string|undefined`) instead of an object with its `name`

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -0,0 +1,68 @@
...
+        // However, due to ambiguity in the specification, we asked for clarification: https://github.com/WICG/cookie-store/issues/226
+        event.changed
+          .concat(event.deleted)
+          .filter((change) => change.name === cookieName)
+          .forEach((change) => {
+            callback({
+              name: change.name,
+              value: change.value,
+            })
+          })
```
BenoitZugmeyer, line 47: **💬 suggestion:** ‏Only yield the latest value:
```suggestion
        const changeEvent =
          find(event.changed, event => event.name === cookieName) ||
          find(event.deleted, event => event.name === cookieName)
        if (change) {
          callback({
            name: change.name,
            value: change.value
          })
        }
```

packages/rum-core/src/browser/cookieObservable.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+import { mockClock } from '@datadog/browser-core/test'
+import type { RumConfiguration } from '../domain/configuration'
+import { WATCH_COOKIE_INTERVAL_DELAY, createCookieObservable } from './cookieObservable'
+import type { CookieChange, CookieObservable } from './cookieObservable'
+
+const COOKIE_NAME = 'cookie_name'
+const COOKIE_DURATION = ONE_MINUTE
+
+describe('cookieObservable', () => {
+  let observable: CookieObservable
```
BenoitZugmeyer, line 13: **🥜 nitpick:** ‏`observable` does not need to be global

packages/rum-core/src/browser/cookieObservable.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    setCookie(COOKIE_NAME, 'foo', COOKIE_DURATION)
+    clock.tick(WATCH_COOKIE_INTERVAL_DELAY)
+  })
+
+  it('should notify observers on cookie change when cookieStore is not supported', () => {
+    Object.defineProperty(window, 'cookieStore', { get: () => undefined, configurable: true })
+    observable = createCookieObservable({} as RumConfiguration, COOKIE_NAME)
+
+    let cookieChange: CookieChange | undefined
+    observable.subscribe((change) => (cookieChange = change))
```
BenoitZugmeyer, line 51: **💬 suggestion:** ‏
```suggestion
    subscription = observable.subscribe((change) => (cookieChange = change))
```

packages/rum-core/src/browser/cookieObservable.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  })
+
+  it('should not notify observers on cookie change when the cookie value as not changed when cookieStore is not supported', () => {
+    Object.defineProperty(window, 'cookieStore', { get: () => undefined, configurable: true })
+    observable = createCookieObservable({} as RumConfiguration, COOKIE_NAME)
+
+    setCookie(COOKIE_NAME, 'foo', COOKIE_DURATION)
+
+    let cookieChange: CookieChange | undefined
+    observable.subscribe((change) => (cookieChange = change))
```
BenoitZugmeyer, line 69: **💬 suggestion:** ‏
```suggestion
    subscription = observable.subscribe((change) => (cookieChange = change))
```

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -0,0 +1,68 @@
...
+export type CookieChange = { name: string; value: string | undefined }
+
+export function createCookieObservable(configuration: Configuration, cookieName: string) {
+  const detectCookieChangeStrategy = (window as CookieStoreWindow).cookieStore
+    ? listenToCookieStoreChange(configuration)
+    : watchCookieFallback
+
+  return new Observable<CookieChange>((observable) =>
+    detectCookieChangeStrategy(cookieName, (event) => observable.notify(event))
+  )
```
BenoitZugmeyer, line 27: **🥜 nitpick:** ‏For the sake of readability, I would rewrite this like this:

```suggestion
  return new Observable<CookieChange>((observable) => {
    if ((window as CookieStoreWindow).cookieStore) {
      return listenToCookieStoreChange(configuration, cookieName, observable.notify)
    }
    return watchCookieFallback(cookieName, observable.notify)
  })
```

Reasoning: this removes a few indirections ('Strategy' variable, currying/intermediary function)

amortemousque: Well I made this change because of this comment: https://github.com/DataDog/browser-sdk/pull/2639#discussion_r1544284211

BenoitZugmeyer: Clarity is subjective, I stand by my opinion, but feel free to dismiss 😃  This is just a nitpick

BenoitZugmeyer (approved):

PR #2645, opened 2024-03-12
✨ [RUM-3495] Add Pr Commenter Threshold

RomanGaignault:
## Motivation

We want to have a more precise indication of the difference change. 

## Changes

Added a threshold based on previous PR values, to set a status. If the value is below 5 % then the status is ✅ otherwise the status is a warning and message appear at the end of the Pr Commenter.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/report-bundle-size/report-as-a-pr-comment.js
+++ b/scripts/report-bundle-size/report-as-a-pr-comment.js
@@ -150,7 +150,7 @@ function createMessage(difference, resultsBaseQuery, localBundleSizes) {
 <details>
   <summary>Expand threshold legend</summary>
 
-  | 𝚫 < -0.5% | 𝚫 < -0.3% | 𝚫 < -0.15% | 𝚫 < 0 | 𝚫 > 0 | 𝚫 > 0.15% | 𝚫 > 0.3% | 𝚫 > 0.5% |
+  | 𝚫 < -0.5% | 𝚫 < -0.3% | 𝚫 < -0.15% | 𝚫 < 0 | 𝚫 ≥ 0 | 𝚫 > 0.15% | 𝚫 > 0.3% | 𝚫 > 0.5% |
   | --- | --- | --- | --- | --- | --- | --- | --- |
   | 🤩 | 😃 | 😀 | 🙂 | 😐 | 😠 | 😡 | ⛔ |
 </details>
--- a/scripts/report-bundle-size/report-as-a-pr-comment.js
+++ b/scripts/report-bundle-size/report-as-a-pr-comment.js
@@ -163,15 +163,15 @@ function formatSize(bundleSize) {
 function getEmojiWithThreshold(percentageChange) {
   const absChange = Math.abs(percentageChange)
   const emojiMap = [
-    { threshold: 0.15, positive: '🟢', negative: '🎉' },
-    { threshold: 0.3, positive: '🟠', negative: '🥳' },
-    { threshold: 0.5, positive: '😡', negative: '🤩' },
-    { threshold: Infinity, positive: '⛔', negative: '🏆' },
+    { threshold: 0.15, increase: '😐', decrease: '🙂' },
+    { threshold: 0.3, increase: '😠', decrease: '😀' },
+    { threshold: 0.5, increase: '😡', decrease: '😃' },
+    { threshold: Infinity, increase: '⛔', decrease: '🤩' },
   ]
 
   for (let emoji of emojiMap) {
     if (absChange <= emoji.threshold) {
-      return percentageChange >= 0 ? emoji.positive : emoji.negative
+      return percentageChange >= 0 ? emoji.increase : emoji.decrease
     }
   }
 }
--- a/scripts/report-bundle-size/report-as-a-pr-comment.js
+++ b/scripts/report-bundle-size/report-as-a-pr-comment.js
@@ -137,24 +137,23 @@ async function updateOrAddComment(difference, resultsBaseQuery, localBundleSizes
 }
 
 function createMessage(difference, resultsBaseQuery, localBundleSizes) {
-  let message = '| 📦 Bundle Name| Base Size | Local Size | 𝚫% |\n| --- | --- | --- | --- |\n'
+  let message = '| 📦 Bundle Name| Base Size | Local Size | 𝚫% | Status |\n| --- | --- | --- | --- | --- |\n'
+  let warningFlag = false
   difference.forEach((diff, index) => {
     const baseSize = formatSize(resultsBaseQuery[index].size)
     const localSize = formatSize(localBundleSizes[diff.name])
     const sign = diff.percentageChange > 0 ? '+' : ''
-    const emoji = getEmojiWithThreshold(diff.percentageChange)
-    message += `| ${formatBundleName(diff.name)} | ${baseSize} | ${localSize} | ${sign}${diff.percentageChange}% ${emoji} |\n`
+    let status = '✅'
+    if (diff.percentageChange > 5) {
+      status = '⚠️'
+      warningFlag = true
+    }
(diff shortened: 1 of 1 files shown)


scripts/report-bundle-size/report-as-a-pr-comment.js
```diff
@@ -159,6 +160,22 @@ function formatSize(bundleSize) {
...
 }
 
+function getEmojiWithThreshold(percentageChange) {
+  const absChange = Math.abs(percentageChange)
+  const emojiMap = [
+    { threshold: 0.15, positive: '🟢', negative: '🎉' },
+    { threshold: 0.3, positive: '🟠', negative: '🥳' },
+    { threshold: 0.5, positive: '😡', negative: '🤩' },
+    { threshold: Infinity, positive: '⛔', negative: '🏆' },
+  ]
```
amortemousque, line 170: **💬 suggestion:** ‏The first time I read I got confused thinking that positive and negative were reversed because for instance the emoji 😡 is a negative feeling. Also I found hard to understand the rating especially with the positive emojis.

What do you think about?

```ts
  const emojiMap = [
    { threshold: 0.15, increase: '😐', decrease: '🙂' },
    { threshold: 0.3, increase: '😠', decrease: '😀' },
    { threshold: 0.5, increase: '😡', decrease: '😃' },
    { threshold: Infinity, increase: '👿 ', decrease: '🤩' },
  ]
```

RomanGaignault: Positive and negative was just here to declare if  "is the bundle size difference positive or negative". But it's maybe clearer to just say increase or decrease. I agree with the scale you made but for the last increase don't you think it would be better to just keep ⛔ ? because in that case we don't want that pr with that much increase

amortemousque: ⛔  also works :)

amortemousque (approved):

scripts/report-bundle-size/report-as-a-pr-comment.js
```diff
@@ -159,6 +160,22 @@ function formatSize(bundleSize) {
...
 }
 
+function getEmojiWithThreshold(percentageChange) {
+  const absChange = Math.abs(percentageChange)
+  const emojiMap = [
+    { threshold: 0.15, increase: '😐', decrease: '🙂' },
+    { threshold: 0.3, increase: '😠', decrease: '😀' },
+    { threshold: 0.5, increase: '😡', decrease: '😃' },
+    { threshold: Infinity, increase: '⛔', decrease: '🤩' },
+  ]
```
bcaudan, line 170: **❓ question:** ‏any reasoning on the threshold values and the number of thresholds?

thomas-lebeau:
**💬 suggestion:** what do you think adding a legend to the bot's comment so we can quickly check what are the thresholds if needed?

something like:
<details>
<summary>Threshold legend</summary>

| 𝚫 < -0.5% | 𝚫 < -0.3% | 𝚫 < -0.15% | 𝚫 < 0 | 𝚫 > 0 | 𝚫 > 0.15% | 𝚫 > 0.3% | 𝚫 > 0.5% |
| --------- | --------- | ---------- | ----- | ----- | --------- | -------- | -------- |
| 🤩        | 😃        | 😀         | 🙂    | 😐    | 😠        | 😡       | ⛔       |
</details>
‏

scripts/report-bundle-size/report-as-a-pr-comment.js
RomanGaignault: Sorry, I just updated the values to 1.5 / 3.0 / 4.5. We discussed the values with Alfonso after calculating what 1 %, 3 %... would represent in terms of bytes. Do you have an other approach to decide those ?

bcaudan: IMO, it could be interesting that the signal raised has a clear meaning and push us to do something.
Should we react differently between a 😠 and a 😡?
Maybe having a single threshold for significant increase/decrease could be enough.

About threshold values, I think displaying % is valuable in the report but using % for thresholds is probably not ideal, since it would allow bigger and bigger increases as the package size is growing.

scripts/report-bundle-size/report-as-a-pr-comment.js
```diff
@@ -137,14 +137,24 @@ async function updateOrAddComment(difference, resultsBaseQuery, localBundleSizes
...
-  let message = '| 📦 Bundle Name| Base Size | Local Size | 𝚫% |\n| --- | --- | --- | --- |\n'
+  let message = '| 📦 Bundle Name| Base Size | Local Size | 𝚫% | Status |\n| --- | --- | --- | --- | --- |\n'
+  let warningFlag = false
   difference.forEach((diff, index) => {
     const baseSize = formatSize(resultsBaseQuery[index].size)
     const localSize = formatSize(localBundleSizes[diff.name])
     const sign = diff.percentageChange > 0 ? '+' : ''
-    message += `| ${formatBundleName(diff.name)} | ${baseSize} | ${localSize} | ${sign}${diff.percentageChange}% |\n`
+    let status = '  ✅  '
+    if (diff.percentageChange > 5) {
```
bcaudan, line 147: **💬 suggestion:** ‏what about:
- extracting a constant for this 5 
- add a comment about the reasoning on the value
- reuse this constant in the displayed message

scripts/report-bundle-size/report-as-a-pr-comment.js
```diff
@@ -137,14 +137,24 @@ async function updateOrAddComment(difference, resultsBaseQuery, localBundleSizes
 }
 
 function createMessage(difference, resultsBaseQuery, localBundleSizes) {
-  let message = '| 📦 Bundle Name| Base Size | Local Size | 𝚫% |\n| --- | --- | --- | --- |\n'
+  let message = '| 📦 Bundle Name| Base Size | Local Size | 𝚫% | Status |\n| --- | --- | --- | --- | --- |\n'
+  let warningFlag = false
```
bcaudan, line 141: **💬 suggestion:** ‏it could be more helpful to describe the intent rather than the effect, maybe something like `highIncreaseDetected` instead

scripts/report-bundle-size/report-as-a-pr-comment.js
```diff
@@ -137,14 +137,24 @@ async function updateOrAddComment(difference, resultsBaseQuery, localBundleSizes
...
   difference.forEach((diff, index) => {
     const baseSize = formatSize(resultsBaseQuery[index].size)
     const localSize = formatSize(localBundleSizes[diff.name])
     const sign = diff.percentageChange > 0 ? '+' : ''
-    message += `| ${formatBundleName(diff.name)} | ${baseSize} | ${localSize} | ${sign}${diff.percentageChange}% |\n`
+    let status = '  ✅  '
```
bcaudan, line 146: **❓ question:** ‏any reason to have these spaces around the emoji? should we have some around the warning emoji as well?

scripts/report-bundle-size/report-as-a-pr-comment.js
```diff
@@ -7,6 +7,8 @@ const LOCAL_BRANCH = process.env.CI_COMMIT_REF_NAME
 const PR_COMMENTER_AUTH_TOKEN = command`authanywhere`.run().split(' ')[2].trim()
 const GITHUB_TOKEN = getGithubAccessToken()
 const ONE_DAY_IN_SECOND = 24 * 60 * 60
+const SIZE_INCREASE_THRESHOLD = 5
+// The value is set to 5% as it's around 10 times the average value for small PRs.
```
bcaudan, line 11: **💬 suggestion:** comments are usually above the line they refer to 🤔 
at least for consistency across the repo let's move it above
```suggestion
// The value is set to 5% as it's around 10 times the average value for small PRs.
const SIZE_INCREASE_THRESHOLD = 5
```

bcaudan (approved):

PR #2648, opened 2024-03-14
✨ [RUM-3546] Add support of PCI compliant intake for browser logs 

N-Boutaib:
## Motivation
Forward browser-sdk logs to PCI compliant endpoint.

## Changes
- Add a new param `usePciIntake` in the SDK configuration
- Use PCI compliant Intake when `usePciIntake` is set to `true`
- Display a warning in console when the configuration is wrong (since PCI compliant intake is only available in us1)

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -59,6 +59,8 @@ export interface InitConfiguration {
   internalAnalyticsSubdomain?: string
 
   telemetryConfigurationSampleRate?: number
+
+  usePciIntake?: boolean
```
BenoitZugmeyer, line 63: **💬 suggestion:** ‏Since it is only log related, we should move this to the `LogsInitConfiguration` interface. You could change the endpoint builder code to accept `InitConfiguration & { usePciIntake?: boolean }` as parameter. WDYT?

packages/core/test/coreConfiguration.ts
```diff
@@ -72,6 +74,7 @@ export type MapInitConfigurationKey<Key extends string> =
           | 'internalAnalyticsSubdomain'
           | 'replica'
           | 'enableExperimentalFeatures'
+          | 'usePciIntake'
```
BenoitZugmeyer, line 77: **💬 suggestion:** ‏include this parameter in the Telemetry Configuration as it's not sensitive

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -131,6 +133,12 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     return
   }
 
+  if (initConfiguration.usePciIntake === true && initConfiguration.site !== 'datadoghq.com') {
```
BenoitZugmeyer, line 136: **🔨 warning:** ‏at this point `initConfiguration.site` is whatever the user provided, so `site` has no default yet. The condition should be:
```suggestion
  if (initConfiguration.usePciIntake === true && initConfiguration.site && initConfiguration.site !== 'datadoghq.com') {
```

packages/core/src/domain/configuration/configuration.ts
N-Boutaib: It makes sense since this param is only for logs config.
Done ✔️

packages/core/src/domain/configuration/configuration.ts
N-Boutaib: Done ✔️

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -20,7 +20,9 @@ export interface ReplicaConfiguration {
   rumEndpointBuilder: EndpointBuilder
 }
 
-export function computeTransportConfiguration(initConfiguration: InitConfiguration): TransportConfiguration {
+export function computeTransportConfiguration(
+  initConfiguration: InitConfiguration & { usePciIntake?: boolean }
```
BenoitZugmeyer, line 24: **💬 suggestion:** ‏Maybe we could use `& { usePciIntake?: boolean }` only in the function it's used (`buildEndpointHost`).

It would be less explicit but at the same time a bit cleaner.

An alternative would be to define a different type for this configuration object, but I don't think it's worth it.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -133,7 +133,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     return
   }
 
-  if (initConfiguration.usePciIntake === true && initConfiguration.site !== 'datadoghq.com') {
+  if (initConfiguration.usePciIntake === true && initConfiguration.site && initConfiguration.site !== 'datadoghq.com') {
```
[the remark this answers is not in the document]
N-Boutaib: Done

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -101,7 +99,9 @@ export interface Configuration extends TransportConfiguration {
   messageBytesLimit: number
 }
 
-export function validateAndBuildConfiguration(initConfiguration: InitConfiguration): Configuration | undefined {
+export function validateAndBuildConfiguration(
+  initConfiguration: InitConfiguration & { usePciIntake?: boolean }
```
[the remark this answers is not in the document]
N-Boutaib: Done ✔️

packages/core/src/domain/configuration/transportConfiguration.ts
N-Boutaib: Done ✔️

BenoitZugmeyer (approved):

thomas-lebeau (approved):

amortemousque:
**🔨 warning:** The pci intake request is collected as RUM ressource. We need also need add the pci intake to isIntakeUrl to prevent this. https://github.com/DataDog/browser-sdk/blob/5040e80871b60a8767461da8577f6ae63cdd8525/packages/core/src/domain/configuration/transportConfiguration.ts#L33
‏

PR #2654, opened 2024-03-18
👷 check monitors before deploying to prod

bcaudan:
## Motivation

Automate manual check, ensure that our monitors are green before deploying to avoid to deploy when there is an ongoing issue.
We could later introduce gate steps to automate completely the deployment process.

## Changes

- extract siteByDatacenter 
- introduce new script to check monitors
- add check monitors before CDN deployment

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/deploy/check-monitors.js
+++ b/scripts/deploy/check-monitors.js
@@ -1,0 +1,48 @@
+/**
+ * Check monitors status
+ * Usage:
+ * node check-monitors.js us1,eu1,...
+ */
+const { printLog, runMain, fetch } = require('../lib/execution-utils')
+const { getTelemetryOrgApiKey, getTelemetryOrgApplicationKey } = require('../lib/secrets')
+const { siteByDatacenter } = require('../lib/datadog-sites')
+
+const monitorIdsByDatacenter = {
+  us1: [72055549, 68975047, 110519972],
+  eu1: [5855803, 5663834, 9896387],
+  us3: [164368, 160677, 329066],
+  us5: [22388, 20646, 96049],
+  ap1: [858, 859, 2757030],
+}
+
+const datacenters = process.argv[2].split(',')
+
+runMain(async () => {
+  for (const datacenter of datacenters) {
+    if (!monitorIdsByDatacenter[datacenter]) {
+      printLog(`No monitors configured for datacenter ${datacenter}`)
+      continue
+    }
+    const monitorIds = monitorIdsByDatacenter[datacenter]
+    const site = siteByDatacenter[datacenter]
+    const monitorStatuses = await Promise.all(monitorIds.map((monitorId) => fetchMonitorStatus(site, monitorId)))
+    for (const monitorStatus of monitorStatuses) {
+      printLog(`${monitorStatus.overall_state} - ${monitorStatus.name}`)
+      if (monitorStatus.overall_state !== 'OK') {
+        throw new Error(`Monitor ${monitorStatus.name} is in state ${monitorStatus.overall_state}`)
+      }
+    }
+  }
+})
+
+async function fetchMonitorStatus(site, monitorId) {
+  const response = await fetch(`https://api.${site}/api/v1/monitor/${monitorId}`, {
+    method: 'GET',
+    headers: {
+      Accept: 'application/json',
+      'DD-API-KEY': getTelemetryOrgApiKey(site),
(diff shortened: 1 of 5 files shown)


scripts/deploy/check-monitors.js
```diff
@@ -0,0 +1,48 @@
...
+ * Check monitors status
+ * Usage:
+ * node check-monitors.js us1,eu1,...
+ */
+const { printLog, runMain, fetch } = require('../lib/execution-utils')
+const { getTelemetryOrgApiKey, getTelemetryOrgApplicationKey } = require('../lib/secrets')
+const { siteByDatacenter } = require('../lib/datadog-sites')
+
+const monitorIdsByDatacenter = {
+  us1: [72055549, 68975047, 110519972],
```
amortemousque, line 11: **❓ question:** ‏Is there an API to get the list of monitors for a datacenter?

amortemousque (approved):

scripts/deploy/check-monitors.js
```diff
@@ -0,0 +1,48 @@
...
+      printLog(`No monitors configured for datacenter ${datacenter}`)
+      continue
+    }
+    const monitorIds = monitorIdsByDatacenter[datacenter]
+    const site = siteByDatacenter[datacenter]
+    const monitorStatuses = await Promise.all(monitorIds.map((monitorId) => fetchMonitorStatus(site, monitorId)))
+    for (const monitorStatus of monitorStatuses) {
+      printLog(`${monitorStatus.overall_state} - ${monitorStatus.name}`)
+      if (monitorStatus.overall_state !== 'OK') {
+        throw new Error(`Monitor ${monitorStatus.name} is in state ${monitorStatus.overall_state}`)
```
amortemousque, line 32: **💬 suggestion:** ‏It could be nice to have link to be able to jump to the monitor

scripts/deploy/check-monitors.js
bcaudan: I have not checked, I guess you are wondering if we could query those monitors instead of listing it.
I have not thought about it, I stayed close to the SDP API.
It should not change that often though.
Do you see/forsee any issue?

amortemousque: Not an issue, just less maintenance: It could prevent us from updating this script if we add or remove a monitor at some point.

bcaudan: FMU, It would require us to have a query specific enough to select only the monitors that we want to check before deployment, which we don't have for now.
We could attach a specific tags but it would require to update all monitors and then it could be easily missed or added inadvertently on new monitors 🤔
I think it should be good enough for now and it could be revisited if we start to add new monitors and are bothered by this configuration.

scripts/deploy/check-monitors.js
```diff
@@ -0,0 +1,64 @@
...
+async function fetchMonitorStatus(site, monitorId) {
+  const response = await fetch(`https://api.${site}/api/v1/monitor/${monitorId}`, {
+    method: 'GET',
+    headers: {
+      Accept: 'application/json',
+      'DD-API-KEY': getTelemetryOrgApiKey(site),
+      'DD-APPLICATION-KEY': getTelemetryOrgApplicationKey(site),
+    },
+  })
+  return JSON.parse(response)
```
BenoitZugmeyer, line 49: **💭 thought:** ‏this function called `fetch` is quite confusing (for example it's not possible to use `JSON.parse(response)` like this with the native `fetch` function), I'd like to move away from this name clash one day!

BenoitZugmeyer (approved):

scripts/deploy/check-monitors.js
bcaudan: Actually, do we still to use this fetch wrapper ?
https://github.com/DataDog/browser-sdk/blob/84cd328da65f6957208bc4b2dd2fec21e39de870/scripts/lib/execution-utils.js#L3

BenoitZugmeyer: We definitly can remove `node-fetch`. But we need to see what to do with the wrapper: https://github.com/DataDog/browser-sdk/blob/84cd328da65f6957208bc4b2dd2fec21e39de870/scripts/lib/execution-utils.js#L54-L61
Exporting it as `fetch` is confusing.

bcaudan: I'll look into in a following PR

PR #2655, opened 2024-03-18
🐛 [RUM-3598] Ignore collecting requests to logs PCI intake as RUM resources

N-Boutaib:
## Motivation
Requests to PCI compliant endpoint are captured by RUM as resources, while they shouldn't be.

## Changes
RUM initialisation doesn't have access to Logs config. Thus, PCI endpoint requests are not skipped and are captured as RUM resources.

When the provided site is `datadoghq.com`, push PCI intake url to the intake prefixes used to ignore resources 

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/transportConfiguration.spec.ts
+++ b/packages/core/src/domain/configuration/transportConfiguration.spec.ts
@@ -87,6 +87,7 @@ describe('transportConfiguration', () => {
     ;[
       { site: 'datadoghq.eu', intakeDomain: 'browser-intake-datadoghq.eu' },
       { site: 'datadoghq.com', intakeDomain: 'browser-intake-datadoghq.com' },
+      { site: 'datadoghq.com', intakeDomain: 'pci.browser-intake-datadoghq.com' },
       { site: 'us3.datadoghq.com', intakeDomain: 'browser-intake-us3-datadoghq.com' },
       { site: 'us5.datadoghq.com', intakeDomain: 'browser-intake-us5-datadoghq.com' },
       { site: 'ap1.datadoghq.com', intakeDomain: 'browser-intake-ap1-datadoghq.com' },
@@ -101,13 +102,6 @@ describe('transportConfiguration', () => {
         expect(configuration.isIntakeUrl(`https://${intakeDomain}/api/v2/logs?xxx`)).toBe(true)
         expect(configuration.isIntakeUrl(`https://${intakeDomain}/api/v2/replay?xxx`)).toBe(true)
       })
-    })
-
-    it('should detect PCI intake request of logs for datadoghq.com site', () => {
-      const configuration = computeTransportConfiguration({ clientToken, site: 'datadoghq.com' })
-      expect(configuration.isIntakeUrl('https://pci.browser-intake-datadoghq.com/api/v2/logs?xxx')).toBe(true)
-      expect(configuration.isIntakeUrl('https://pci.browser-intake-datadoghq.com/api/v2/rum?xxx')).toBe(false)
-      expect(configuration.isIntakeUrl('https://pci.browser-intake-datadoghq.com/api/v2/replay?xxx')).toBe(false)
     })
 
     it('should detect internal analytics intake request for datadoghq.com site', () => {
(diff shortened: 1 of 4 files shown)


packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -69,3 +76,29 @@ function computeReplicaConfiguration(
...
+  } else {
+    const domainParts = site.split('.')
+    const extension = domainParts.pop()
+    intakePrefixesRegEx.push(
+      new RegExp(`^https://browser-intake-${domainParts.join('-')}.${extension!}/api/v2/(logs|rum|replay)`)
+    )
+  }
+
+  return intakePrefixesRegEx
+}
```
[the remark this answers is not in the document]
BenoitZugmeyer: I agree, we should probably simplify things.

Simply having the PCI intake (`https://pci.browser-intake-datadoghq.com`) in the `intakeUrlPrefixes` list could solve the issue.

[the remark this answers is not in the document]
N-Boutaib: I agree, that would the simplest way. Doing the change ✔️

packages/core/src/domain/configuration/transportConfiguration.spec.ts
```diff
@@ -87,10 +87,12 @@ describe('transportConfiguration', () => {
     ;[
       { site: 'datadoghq.eu', intakeDomain: 'browser-intake-datadoghq.eu' },
       { site: 'datadoghq.com', intakeDomain: 'browser-intake-datadoghq.com' },
+      { site: 'datadoghq.com', intakeDomain: 'pci.browser-intake-datadoghq.com' },
```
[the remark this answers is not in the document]
N-Boutaib: Done ✔️

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -21,18 +21,24 @@ export interface ReplicaConfiguration {
...
+  const { site = INTAKE_SITE_US1 } = initConfiguration
+
   const tags = buildTags(initConfiguration)
 
   const endpointBuilders = computeEndpointBuilders(initConfiguration, tags)
   const intakeUrlPrefixes = objectValues(endpointBuilders).map((builder) => builder.urlPrefix)
 
+  if (site === INTAKE_SITE_US1) {
+    intakeUrlPrefixes.push(`https://${PCI_INTAKE_HOST_US1}/api/v2/logs?`)
+  }
```
bcaudan, line 33: **🥜 nitpick:** ‏what about extracting a `computeIntakeUrlPrefixes` function?

bcaudan (approved):

packages/core/src/domain/configuration/transportConfiguration.ts
BenoitZugmeyer, line 33: **💬 suggestion:** ‏We don't need to be too strict here. It could be simpler/smaller like this:

```suggestion
  intakeUrlPrefixes.push(`https://${PCI_INTAKE_HOST_US1}/`)
```

thomas-lebeau (approved):

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -21,18 +21,24 @@ export interface ReplicaConfiguration {
 }
 
 export function computeTransportConfiguration(initConfiguration: InitConfiguration): TransportConfiguration {
+  const { site = INTAKE_SITE_US1 } = initConfiguration
```
BenoitZugmeyer, line 24: **💬 suggestion:** ‏`initConfiguration` is directly the object provided by the user. `initConfiguration.site` is typed as `string | undefined`, but in practice it can be anything, like an empty string.

Changing from `initConfiguration.site || INTAKE_SITE_US1` to `{ site = INTAKE_SITE_US1 } = initConfiguration` could be a breaking change in that case.

Suggestion:
```suggestion
  const site = initConfiguration.site || INTAKE_SITE_US1
```

amortemousque (approved):

packages/core/src/domain/configuration/transportConfiguration.ts
N-Boutaib: I think we should keep the path as well. Only logs are supported in the PCI compliant endpoint.

PR #2656, opened 2024-03-18
✨ [RUM-3542] Add trace context injection control in rum configuration

cy-moi:
## Motivation
APM is working on building more granular sampling control on the backend side which can't be applied if the sampling decision is already taken on the client side.

## Changes
Added a field `traceContextInjection` in the configuration to control context injection. 

This new configuration allows more granular sampling at the backend and does not bring any break changes. It is available along with sampling rate configuration.

`traceContextInjection` is an opt-in configuration defaults to "all" which means no behavior changes. If opt-in for the "sampled" option, you would see only sampled requests having trace context attached to them.

## Testing
| Sampling Decision | traceContextInjection | Expected Behavior | 
|--------|--------|--------|
| Sampled | all | Inject trace context with sampling priority 1 | 
| Sampled | sampled | Inject trace context with sampling priority 1 | 
| Not Sampled | all | Inject trace context with sampling priority 0 | 
| Not Sampled | sampled | Do not inject trace context | 

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/configuration.ts
+++ b/packages/rum-core/src/domain/configuration.ts
@@ -13,7 +13,7 @@ import {
 } from '@datadog/browser-core'
 import type { RumEventDomainContext } from '../domainContext.types'
 import type { RumEvent } from '../rumEvent.types'
-import { isTracingOption } from './tracing/tracer'
+import { isTracingOption, TraceContextInjection } from './tracing/tracer'
 import type { PropagatorType, TracingOption } from './tracing/tracer.types'
 
 export const DEFAULT_PROPAGATOR_TYPES: PropagatorType[] = ['tracecontext', 'datadog']
@@ -48,22 +48,6 @@ export interface RumInitConfiguration extends InitConfiguration {
 }
 
 export type HybridInitConfiguration = Omit<RumInitConfiguration, 'applicationId' | 'clientToken'>
-
-/*
-Allows to control the behavior of trace context injection.
-This allows upstream services to inject trace context based on their sampling configuration.
-*/
-export enum TraceContextInjection {
-  /*
-  Default. Inject trace context to all network requests as per the sampling rate.
-  */
-  All,
-
-  /*
-  Inject trace context only if the trace is sampled.
-  */
-  Sampled,
-}
 
 export interface RumConfiguration extends Configuration {
   // Built from init configuration
--- a/packages/rum-core/src/domain/configuration.ts
+++ b/packages/rum-core/src/domain/configuration.ts
@@ -49,6 +49,22 @@ export interface RumInitConfiguration extends InitConfiguration {
 
 export type HybridInitConfiguration = Omit<RumInitConfiguration, 'applicationId' | 'clientToken'>
 
+/*
+Allows to control the behavior of trace context injection.
+This allows upstream services to inject trace context based on their sampling configuration.
+*/
+export enum TraceContextInjection {
+  /*
+  Default. Inject trace context to all network requests as per the sampling rate.
+  */
+  All,
+
+  /*
+  Inject trace context only if the trace is sampled.
+  */
+  Sampled,
+}
+
 export interface RumConfiguration extends Configuration {
   // Built from init configuration
   actionNameAttribute: string | undefined
@@ -68,6 +84,7 @@ export interface RumConfiguration extends Configuration {
   version?: string
   subdomain?: string
   customerDataTelemetrySampleRate: number
+  traceContextInjection?: TraceContextInjection
 }
 
 export function validateAndBuildRumConfiguration(
@@ -127,6 +144,7 @@ export function validateAndBuildRumConfiguration(
         ? initConfiguration.defaultPrivacyLevel
         : DefaultPrivacyLevel.MASK,
       customerDataTelemetrySampleRate: 1,
+      traceContextInjection: TraceContextInjection.All,
     },
     baseConfiguration
   )
--- a/packages/rum-core/src/domain/configuration.ts
+++ b/packages/rum-core/src/domain/configuration.ts
@@ -130,7 +130,9 @@ export function validateAndBuildRumConfiguration(
         ? initConfiguration.defaultPrivacyLevel
         : DefaultPrivacyLevel.MASK,
       customerDataTelemetrySampleRate: 1,
-      traceContextInjection: initConfiguration.traceContextInjection ?? TraceContextInjection.ALL,
+      traceContextInjection: objectHasValue(TraceContextInjection, initConfiguration.traceContextInjection)
+        ? initConfiguration.traceContextInjection
+        : TraceContextInjection.ALL,
     },
     baseConfiguration
   )
--- a/packages/rum-core/src/domain/configuration.ts
+++ b/packages/rum-core/src/domain/configuration.ts
@@ -6,6 +6,7 @@ import {
   serializeConfiguration,
   assign,
   DefaultPrivacyLevel,
+  TraceContextInjection,
   display,
   isPercentage,
   objectHasValue,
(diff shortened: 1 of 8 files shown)


packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -121,7 +137,9 @@ function injectHeadersIfTracingAllowed(
   context.traceId = new TraceIdentifier()
   context.spanId = new TraceIdentifier()
   context.traceSampled = !isNumber(configuration.traceSampleRate) || performDraw(configuration.traceSampleRate)
-  inject(makeTracingHeaders(context.traceId, context.spanId, context.traceSampled, tracingOption.propagatorTypes))
+  if (configuration.traceContextInjection === TraceContextInjection.All) {
+    inject(makeTracingHeaders(context.traceId, context.spanId, context.traceSampled, tracingOption.propagatorTypes))
+  }
```
amortemousque, line 142: FMU, the check should more look like this 

```ts
 if (
    (context.traceSampled && configuration.traceContextInjection === TraceContextInjection.Sampled) ||
    configuration.traceContextInjection === TraceContextInjection.All
  ) {
    inject(makeTracingHeaders(context.traceId, context.spanId, context.traceSampled, tracingOption.propagatorTypes))
  }
```

packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -19,6 +19,22 @@ import type {
...
+  /*
+    Default. Inject trace context to all network requests as per the sampling rate.
+    */
+  All,
+
+  /*
+    Inject trace context only if the trace is sampled.
+    */
+  Sampled,
+}
```
BenoitZugmeyer, line 36: **💬 suggestion:** ‏user-facing enums should be defined as a plain object, like so:

https://github.com/DataDog/browser-sdk/blob/c52216b192f772911a07d0c8dd255a4134a5597b/packages/core/src/domain/configuration/configuration.ts#L17-L22

This allows using either the object (ex: `TraceContextInjection.All`) or the value directly (ex: `"all"`) without facing typing issues, from JS or TS.

In any case, `const enum` should never be customer-facing because non-TS users won't have access to it.

packages/rum-core/src/domain/tracing/tracer.ts
cy-moi: Thank you! I updated this and added tests accordingly.

packages/rum-core/src/domain/tracing/tracer.spec.ts
```diff
@@ -522,6 +567,53 @@ describe('tracer', () => {
...
+      const context: Partial<RumFetchStartContext> = { ...ALLOWED_DOMAIN_CONTEXT }
+      tracer.traceFetch(context)
+
+      expect(context.init!.headers).toContain(jasmine.arrayContaining(['x-datadog-origin']))
+      expect(context.init!.headers).toContain(jasmine.arrayContaining(['x-datadog-parent-id']))
+      expect(context.init!.headers).toContain(jasmine.arrayContaining(['x-datadog-trace-id']))
+      expect(context.init!.headers).toContain(jasmine.arrayContaining(['x-datadog-sampling-priority']))
+    })
+
+    it('should add headers when trace not sampled yet config set to all', () => {
```
amortemousque, line 601:
```suggestion
    it('should add headers when trace not sampled and config set to all', () => {
```

packages/rum-core/src/domain/tracing/tracer.spec.ts
```diff
@@ -207,6 +207,51 @@ describe('tracer', () => {
       expect(xhrStub.headers['X-B3-TraceId']).toBeUndefined()
     })
 
+    it('should not add any headers when trace not sampled and config set to sampled only', () => {
```
amortemousque, line 210:
```suggestion
    it('should not add any headers when trace not sampled and config set to sampled', () => {
```

packages/rum-core/src/domain/tracing/tracer.spec.ts
```diff
@@ -522,6 +567,53 @@ describe('tracer', () => {
       expect(context.init!.headers).not.toContain(jasmine.arrayContaining(['x-datadog-trace-id']))
       expect(context.init!.headers).not.toContain(jasmine.arrayContaining(['X-B3-TraceId']))
     })
+    it('should not add headers when trace not sampled and config set to sample', () => {
```
bcaudan, line 570: **💬 suggestion:** ‏typo
```suggestion
    it('should not add headers when trace not sampled and config set to sampled', () => {
```

packages/rum-core/src/domain/configuration.ts
```diff
@@ -127,6 +130,7 @@ export function validateAndBuildRumConfiguration(
         ? initConfiguration.defaultPrivacyLevel
         : DefaultPrivacyLevel.MASK,
       customerDataTelemetrySampleRate: 1,
+      traceContextInjection: initConfiguration.traceContextInjection ?? TraceContextInjection.ALL,
```
bcaudan, line 133: **💬 suggestion:** ‏what about adding some tests for this configuration option? (like those we have for other configuration options)

packages/rum-core/src/domain/tracing/tracer.spec.ts
```diff
@@ -207,6 +207,51 @@ describe('tracer', () => {
...
+
+      const tracer = startTracer(configurationWithInjectionParam, sessionManager)
+      const context = { ...ALLOWED_DOMAIN_CONTEXT }
+      tracer.traceXhr(context, xhrStub as unknown as XMLHttpRequest)
+
+      expect(xhrStub.headers['x-datadog-trace-id']).toBeUndefined()
+      expect(xhrStub.headers['x-datadog-sampling-priority']).toBeUndefined()
+    })
+
+    it('should add headers when trace sampled and config set to Sampled', () => {
```
bcaudan, line 225: **🥜 nitpick:** ‏case consistency with other test descriptions
```suggestion
    it('should add headers when trace sampled and config set to sampled', () => {
```

packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -121,7 +122,10 @@ function injectHeadersIfTracingAllowed(
   context.traceId = new TraceIdentifier()
   context.spanId = new TraceIdentifier()
   context.traceSampled = !isNumber(configuration.traceSampleRate) || performDraw(configuration.traceSampleRate)
-  inject(makeTracingHeaders(context.traceId, context.spanId, context.traceSampled, tracingOption.propagatorTypes))
+
+  if (context.traceSampled || configuration.traceContextInjection === TraceContextInjection.ALL) {
+    inject(makeTracingHeaders(context.traceId, context.spanId, context.traceSampled, tracingOption.propagatorTypes))
+  }
```
bcaudan, line 128: **💭 thought:** In the case where we don't inject headers, we don't really need to generate trace and span ids, so it may be nice to avoid to generate them.‏

packages/rum-core/src/domain/configuration.ts
```diff
@@ -68,6 +70,7 @@ export interface RumConfiguration extends Configuration {
   version?: string
   subdomain?: string
   customerDataTelemetrySampleRate: number
+  traceContextInjection?: TraceContextInjection
```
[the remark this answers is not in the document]
cy-moi: I think it can be undefined in the `RumInitconfiguration`, we just use the default value. 

But this is a good point. I made it always defined in RumConfiguration.

bcaudan (approved):
Nice!

BenoitZugmeyer (approved):

hong823:
@cy-moi Any idea if this would support when only using W3C propagatorTypes?

```
          allowedTracingUrls: [
            { match: "https://example.com", propagatorTypes: ['tracecontext'] }
          ]
```

From what I can tell, it might not be. Can you confirm?

PR #2659, opened 2024-03-19
♻️✅ Harmonize record observers

amortemousque:
## Motivation

While implementing webview support for Session Replay, I identified inconsistencies in the recorder observers, plus some missing tests.

## Changes
 
- Harmonize record observers to always 
  - return [BrowserRecord](https://github.com/DataDog/browser-sdk/blob/e81ba6ddb33ece8aca6cbdec02606e6870f07c8b/packages/rum/src/types/sessionReplay.ts#L40) in callbacks
  - return a `stop` function
- Add tests for focusObserver, mediaInteractionObserver, scrollObserver, visualViewportResizeObserver
- I moved the [observers initialization up in record()](https://github.com/DataDog/browser-sdk/blob/02065577bb1aeae6ccfffb738dd3132c2f3e16a8/packages/rum/src/domain/record/record.ts#L75) and removed the `initObservers()` wrapper.

## Testing

- [X] Local
- [x] Staging
- [X] Unit
- [ ] End to end

---


packages/rum/src/domain/record/record.ts
```diff
@@ -74,34 +58,18 @@ export function record(options: RecordOptions): RecordAPI {
...
+    mediaInteractionCb: emitAndComputeStats,
+    mouseInteractionCb: emitAndComputeStats,
+    mousemoveCb: emitAndComputeStats,
+    mutationCb: emitAndComputeStats,
+    scrollCb: emitAndComputeStats,
+    styleSheetCb: emitAndComputeStats,
+    viewportResizeCb: emitAndComputeStats,
+    frustrationCb: emitAndComputeStats,
+    focusCb: emitAndComputeStats,
+    visualViewportResizeCb: emitAndComputeStats,
```
thomas-lebeau, line 72: **❓ question:** ‏ if they're all the same, can't we just have one generic `cb` property?

amortemousque: Good point! I'm even wondering if we still need the observer.ts. I'll make some changes and let you know :)

amortemousque: So:
- I moved the [observers initialization up in record()](https://github.com/DataDog/browser-sdk/blob/02065577bb1aeae6ccfffb738dd3132c2f3e16a8/packages/rum/src/domain/record/record.ts#L75) and removed the initObservers() wrapper. 
- Harmonize observers to return a `stop` function 
Let me know what you think

packages/rum/src/domain/record/observers/viewportResizeObserver.ts
```diff
@@ -13,22 +12,25 @@ export type ViewportResizeCallback = (incrementalSnapshotRecord: BrowserIncremen
...
 
 export function initVisualViewportResizeObserver(
   configuration: RumConfiguration,
   visualViewportResizeCb: VisualViewportResizeCallback
-): ListenerHandler {
+) {
   const visualViewport = window.visualViewport
   if (!visualViewport) {
-    return noop
+    return { stop: () => {} }
```
thomas-lebeau, line 33: **🥜 nitpick:** ‏ 

```suggestion
    return { stop: noop }
```

packages/rum/src/domain/record/record.ts
```diff
@@ -53,34 +67,33 @@ export function record(options: RecordOptions): RecordAPI {
...
-    scrollCb: emitAndComputeStats,
-    styleSheetCb: emitAndComputeStats,
-    viewportResizeCb: emitAndComputeStats,
-    frustrationCb: emitAndComputeStats,
-    focusCb: emitAndComputeStats,
-    visualViewportResizeCb: emitAndComputeStats,
-    viewEndCb: (viewEndRecord) => {
+  const recordIds = initRecordIds()
+  const mutationObserver = initMutationObserver(emitAndComputeStats, configuration, shadowRootsController, document)
+  const observers = [
```
thomas-lebeau, line 75: **💬 suggestion:** ‏ what about being explicit with the types here. They are inferred quite well, but I think it could help reading and understanding the code.

```suggestion
  const observers: Observer[] = [
```

with `Observer` something like `type Observer = { stop: () => void }`.

Then you could also be explicit in the return type of all the `init*Observer` methods (Seems like some of them wouldl return `FlushableObserver`, or something similar?)

wdyt?

BenoitZugmeyer: I wouldn't mind naming the type explicitly, although `Observer` might not be great? It comes from RRWeb, but we don't *need* to stick with it.

* We often call those kind of function `trackSomething` (ex: `trackViews`, `trackActions`...), and sometimes `somethingObservable` (ex: `xhrObservable`...)
* We have a larger concept of things that can be "stopped"
* I often confuse `Observer` with `Observable`

Maybe we could ditch the "Observer" naming, and use "track"? (Or "observable", but it we don't use `Observable` objects there yet).

Maybe we could have a `type Stoppable = { stop: () => void }` that we can reuse in other parts of the SDK?

amortemousque: I agree it's confusing to have `Observer` and `Observable`, I like having consistent names "trackSomething()". However I'm not sure about `Stoppable`, what about something more related to the tracking concept like `type TrackingResult = { stop: () => void }`

Let see how the current `const observers: Observer[initObserver, ...]` would look like? 
- `const stoppables: Stoppable[] = [ trackSomething,...`,
- `const trackingResults: TrackingResult[] = [ trackSomething,...`
-  (Maybe) `const trackers: Tracker[] = [ trackSomething,...`

What do you think?
cc @bcaudan

bcaudan: > However I'm not sure about Stoppable, what about something more related to the tracking

👍

> I like having consistent names `trackSomething()`

Looking at the code base, we are using different name patterns for similar usages `trackXXX`, `XXXCollection`, `XXXObserver`, `XXXObservable`, not sure there is clear distinction between all of them.

> - const trackingResults: TrackingResult[] = [ trackSomething,...
> - (Maybe) const trackers: Tracker[] = [ trackSomething,...

Given that the object we are retrieving is "actionable", I would lean toward `tracker` as the `result` suffix would more express to me a simple data structure (without business logic attached).

thomas-lebeau: > result suffix would more express to me a simple data structure (without business logic attached).

I agree! `Tracker` looks good to me, otherwise `Subscription` seems like a good fit too.

> Looking at the code base, we are using different name patterns for similar usages trackXXX, XXXCollection, XXXObserver, XXXObservable, not sure there is clear distinction between all of them.

Yeah, and that makes onboarding to the project difficult. It would be great to standardise these concepts. I tend to prefer well know names such as `Observer`, `Subscription`, `Subject` where possible but as long as it's consistent within the browser SDK it should be fine.

BenoitZugmeyer (approved):

thomas-lebeau (approved):

packages/rum/src/domain/record/record.ts
```diff
@@ -69,50 +67,38 @@ export function record(options: RecordOptions): RecordAPI {
...
   function flushMutations() {
     shadowRootsController.flush()
-    flushMutationsFromObservers()
+    mutationTracker.flush()
   }
 
-  const { stop: stopObservers, flush: flushMutationsFromObservers } = initObservers(configuration, {
-    lifeCycle: options.lifeCycle,
+  const recordIds = initRecordIds()
+  const mutationTracker: FlushableTracker = trackMutation(
```
bcaudan, line 74: **❓ question:** ‏Do we need this `FlushableTracker` abstraction? it seems to be used only for `MutationTracker`

packages/rum/src/domain/record/trackers/recordIds.ts
bcaudan, line 1: **❓ question:** ‏any reason to have `recordIds` under `trackers` but not `elementsScrollPositions` for example?

amortemousque: No it was already under "observers/" which I just renamed. Do you have an improvement in mind?

packages/rum/src/domain/record/record.ts
amortemousque: Following @thomas-lebeau reasoning in his [comment](https://github.com/DataDog/browser-sdk/pull/2659#discussion_r1533720875): the types are inferred quite well, but being more explicit could help reading and understanding the code.
So if we strongly types other trackers, I would expect to also have a type here. Maybe instead of a `FlushableTracker` concept it could be `type MutationTracker = Tracker & { flush: () => void }` collocated within `trackMutation.ts`. Wdyt?

packages/rum/src/domain/record/trackers/recordIds.ts
bcaudan: ok, since it is the only one not being named `trackXXX`, it stands out a bit.
Compared to the others, it does not track browser events or APIs but store an internal state.
Looking at `elementsScrollPositions`, it seems a bit similar.

So I am wondering if both should not be inside or outside `trackers`.

wdyt?

amortemousque: No strong opinion so let's put it outside to have do the minimal amount of change :)

bcaudan (approved):

PR #2662, opened 2024-03-20
🐛 [RUM-3599] do not define undefined instrumented method

BenoitZugmeyer:
## Motivation

Some websites are removing native methods on purpose. Re-defining the method breaks them.

## Changes

Don't define undefined instrumented methods.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/instrumentMethod.spec.ts
+++ b/packages/core/src/tools/instrumentMethod.spec.ts
@@ -28,15 +28,23 @@ describe('instrumentMethod', () => {
     expect(instrumentationSpy).toHaveBeenCalledBefore(originalSpy)
   })
 
-  it('sets a method originally undefined', () => {
+  it('does not set a method originally undefined', () => {
     const object: { method?: () => number } = {}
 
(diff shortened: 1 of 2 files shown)


N-Boutaib (approved):

packages/core/src/tools/instrumentMethod.spec.ts
```diff
@@ -28,15 +28,23 @@ describe('instrumentMethod', () => {
...
-  it('sets a method originally undefined', () => {
+  it('does not set a method originally undefined', () => {
     const object: { method?: () => number } = {}
 
+    instrumentMethod(object, 'method', noop)
+
+    expect(object.method).toBeUndefined()
+  })
+
+  it('sets an event handler even if it was originally undefined', () => {
```
thomas-lebeau, line 39: **❓ question:** ‏ could you explain when this happens? and why we still want to wrap something that was originally undefined?

thomas-lebeau (approved):

PR #2663, opened 2024-03-21
🐛 [RUM-3581] Fix the selector computation due to properties partially supported on IE

amortemousque:
## Motivation

We got telemetry errors when trying to compute the action target CSS selector on IE because some DOM elements (svg, MathML):
- do not support the [classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) property
- do not support [parentElement](https://developer.mozilla.org/fr/docs/Web/API/Node/parentElement) property

## Changes

- Add a `getClassList` polyfill to fallback to `element.getAttribute('class')` in case classList is not supported
- Add a `getParentElement` polyfill to fallback to `node.parentNode` in case parentElement is not supported
- Use the polyfills in action name computation and CSS selector computation

## Testing

- [X] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


packages/rum-core/src/browser/htmlDomUtils.ts
```diff
@@ -43,3 +43,13 @@ export function forEachChildNodes(node: Node, callback: (child: Node) => void) {
...
   return isNodeShadowRoot(node) ? node.host : node.parentNode
 }
+
+export function getClassList(element: Element): DOMTokenList | string[] {
+  // ie 11 does not support classList on SVG and MathMl elements
+  if (element.classList) {
+    return element.classList
+  }
+
+  // Use element.getAttribute('class') because element.className on SVG elements is a SVGAnimatedString
```
N-Boutaib, line 53: 💭 **Thought**: `classList` is not supported on `MathML` elements as well. It would make sense to reflect it in the comment (if the `MathML` case is covered)

amortemousque: It's true that my comments were confusing. I tried to improve them. Let me know what you think

N-Boutaib (approved):

RomanGaignault (approved):

packages/rum-core/src/browser/htmlDomUtils.ts
```diff
@@ -43,3 +43,18 @@ export function forEachChildNodes(node: Node, callback: (child: Node) => void) {
...
 }
+
+/**
+ * Return the classList of an element or an array of classes if classList is not supported
+ *
+ * In cases where classList is not supported, such as in IE11 for SVG and MathML elements,
+ * we fallback to using element.getAttribute('class').
+ * We opt for element.getAttribute('class') over element.className because className returns an SVGAnimatedString for SVG elements.
+ */
+export function getClassList(element: Element): DOMTokenList | string[] {
```
BenoitZugmeyer, line 54: **💬 suggestion:** ‏put it in `polyfills`, as it is more a polyfill than an utility

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -92,9 +93,10 @@ function getClassSelector(element: Element): string | undefined {
   if (element.tagName === 'BODY') {
     return
   }
-  if (element.classList.length > 0) {
-    for (let i = 0; i < element.classList.length; i += 1) {
-      const className = element.classList[i]
+  const classList = getClassList(element)
+  if (classList.length > 0) {
```
BenoitZugmeyer, line 97: **🥜 nitpick:** The `classList.length > 0` condition is useless here, since it's already done in the `for` loop right after. I know it's not your PR that added it, but maybe you can clean it up?

packages/rum-core/src/browser/htmlDomUtils.ts
amortemousque: Do you think about moving it in the `core` package polyfills or to create another polyfills in `rum-core`?
I guess since Logs do need it would be creating one for rum-core?

packages/rum-core/src/domain/getSelectorFromElement.ts
amortemousque: Yep nice :)

packages/rum-core/src/browser/htmlDomUtils.ts
BenoitZugmeyer: Creating a `polyfill` module in `rum-core` would be slightly more mindful when the SDK is consumed without treeshaking abilities!

BenoitZugmeyer (approved):

PR #2669, opened 2024-03-25
🐛 [RUM-96] Ignore frustrations on clicks resulting in scrolls

N-Boutaib:
## Motivation
Currently, click on scrollbars, or more generally clicks resulting in page or element scrolling, are reported as `dead_clicks` or `rage_clicks` (if more than three clicks occurred).
This behaviour is misleading since those clicks did result in a change, which is the scrolling. Therefore, they shouldn't result in a frustration signal.

## Changes
- Add an event listener over scroll and report it to `userActivity`
- Ignore clicks with `scroll` user activity from being reported as `dead_click`s and `rage_click`s 

> [!WARNING]  
> Clicks on scrollbar that do not result in a scroll event will still be reported as frustrations.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/action/computeFrustration.spec.ts
+++ b/packages/rum-core/src/domain/action/computeFrustration.spec.ts
@@ -68,6 +68,13 @@ describe('computeFrustration', () => {
       expect(getFrustrations(clicks[1])).toEqual([])
     })
 
+    it('does not add a dead frustration when clicking to scroll', () => {
+      clicks[0] = createFakeClick({ userActivity: { scroll: true } })
+      clicks[1] = createFakeClick()
+      computeFrustration(clicks, rageClick)
+      expect(getFrustrations(clicks[1])).toEqual([])
+    })
+
     it('adds an error frustration to clicks that have an error', () => {
       clicks[1] = createFakeClick({ hasError: true })
       computeFrustration(clicks, rageClick)
@@ -97,6 +104,12 @@ describe('isRage', () => {
 
   it('does not consider as rage when triple clicking to select a paragraph', () => {
     expect(isRage([createFakeClick(), createFakeClick({ userActivity: { selection: true } }), createFakeClick()])).toBe(
+      false
+    )
+  })
+
+  it('does not consider rage when at least one click is related to a "scroll" event', () => {
+    expect(isRage([createFakeClick(), createFakeClick({ userActivity: { scroll: true } }), createFakeClick()])).toBe(
       false
     )
   })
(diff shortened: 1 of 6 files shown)


packages/rum-core/test/createFakeClick.ts
```diff
@@ -12,7 +12,7 @@ export function createFakeClick({
 }: {
   hasError?: boolean
   hasPageActivity?: boolean
-  userActivity?: { selection?: boolean; input?: boolean }
+  userActivity?: { selection?: boolean; input?: boolean; scroll?: boolean }
```
thomas-lebeau, line 15: **💬 suggestion:** ‏ 

```suggestion
  userActivity?: Partial<UserActivity>
```

thomas-lebeau (approved):

packages/rum-core/test/createFakeClick.ts
N-Boutaib: Done in https://github.com/DataDog/browser-sdk/pull/2669/commits/e2d1a5d15b7f638efb080b74c00b60449658e32d

packages/rum-core/src/domain/action/computeFrustration.ts
```diff
@@ -17,14 +17,17 @@ export function computeFrustration(clicks: Click[], rageClick: Click) {
...
     if (click.hasError) {
       click.addFrustration(FrustrationType.ERROR_CLICK)
     }
     if (
       isDead(click) &&
       // Avoid considering clicks part of a double-click or triple-click selections as dead clicks
-      !hasSelectionChanged
+      !hasSelectionChanged &&
+      // Avoid considering clicks that resulted in a scroll as dead clicks
+      !didScrollOccur
```
BenoitZugmeyer, line 30: **❓ question:** ‏what's the reasoning behind this condition? I'm not sure to get the comment because `isDead` is already checking for `scroll` user activity.

packages/rum-core/src/domain/action/computeFrustration.spec.ts
```diff
@@ -68,6 +68,13 @@ describe('computeFrustration', () => {
       expect(getFrustrations(clicks[1])).toEqual([])
     })
 
+    it('does not add a dead frustration when clicking to scroll', () => {
```
BenoitZugmeyer, line 71: **❓ question:** ‏same question: what's the idea behind this?

packages/rum-core/src/domain/action/computeFrustration.ts
N-Boutaib: You're completely write, I missed that the first thing we check is if the click `isDead`, in which the check for `scroll` activity is done.

packages/rum-core/src/domain/action/computeFrustration.spec.ts
N-Boutaib: I was trying to cover the case where a sequence of clicks was converted to `rage_click`, and that it should not have `dead_click` if a scroll appeared. But in fact, it can't be converted to a `rage_click` 😅.

PR #2670, opened 2024-03-25
👷 [RUM-3709] Fix local build

bcaudan:
## Motivation

Local build is failing with error:

```
@datadog/browser-rum-core: ../core/cjs/domain/error/trackRuntimeError.d.ts(9,32): 
    error TS2694: Namespace '"packages/rum-core/src/index"' has no exported member 'noop'.
```

## Changes

This behaviour has been introduced with the update of typescript from 5.3.3 to 5.4.3, so reverting to last working version while investigating.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "ts-loader": "9.5.1",
     "ts-node": "10.9.2",
     "tsconfig-paths-webpack-plugin": "4.1.0",
-    "typescript": "5.4.3",
+    "typescript": "5.3.3",
     "webdriverio": "8.35.1",
     "webpack": "5.91.0",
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer (approved):

bcaudan:
After further investigation, it seems to be related to this issue https://github.com/microsoft/TypeScript/issues/57956.

Evidences:
- From bisect, issue introduced by the same change: https://github.com/microsoft/TypeScript/pull/55015
- Same path resolution issues observed in the problematic .d.ts files
- Issue seems to only be observed on MacOS

PR #2673, opened 2024-03-26
✨ [RUM-3710] Update session ID handling to support cookie deletion

thomas-lebeau:
## Motivation

Makes session management resilient to clearing the cookie. If a cookie is deleted outside of the browser SDK, we do not create a new session automatically. 

This solves an issue with Adguard that delete the cookie every second. This had for effect to generate a large amount of short lived sessions.

## Changes

- When a session is expired, we set the cookie to `isExpired=1` with an expiration delay of 4h (instead of deleting the cookie)
- When the cookie is altered by a 3rd party, we can now differentiate it from an expired session and won't start a new session automatically after a user action.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/developer-extension/src/panel/components/tabs/infosTab.tsx
+++ b/developer-extension/src/panel/components/tabs/infosTab.tsx
@@ -259,7 +259,7 @@ function formatSessionType(value: string, ...labels: string[]) {
 function endSession() {
   evalInWindow(
     `
-      document.cookie = '_dd_s=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/'
+      DD_RUM.stopSession()
     `
   ).catch((error) => logger.error('Error while ending session:', error))
 }
--- a/developer-extension/src/panel/components/tabs/infosTab.tsx
+++ b/developer-extension/src/panel/components/tabs/infosTab.tsx
@@ -262,7 +262,7 @@ function endSession() {
 
   evalInWindow(
     `
-      document.cookie = '_dd_s=expired=0; expires=${expires}; path=/'
+      document.cookie = '_dd_s=isExpired=1; expires=${expires}; path=/'
     `
   ).catch((error) => logger.error('Error while ending session:', error))
 }
--- a/developer-extension/src/panel/components/tabs/infosTab.tsx
+++ b/developer-extension/src/panel/components/tabs/infosTab.tsx
@@ -259,7 +259,7 @@ function formatSessionType(value: string, ...labels: string[]) {
 function endSession() {
   evalInWindow(
     `
-      DD_RUM.stopSession()
+      document.cookie = '_dd_s=expired=0; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/'
     `
   ).catch((error) => logger.error('Error while ending session:', error))
 }
--- a/developer-extension/src/panel/components/tabs/infosTab.tsx
+++ b/developer-extension/src/panel/components/tabs/infosTab.tsx
@@ -257,9 +257,12 @@ function formatSessionType(value: string, ...labels: string[]) {
 }
 
 function endSession() {
+  const fourHours = 1000 * 60 * 60 * 4
+  const expires = new Date(Date.now() + fourHours).toUTCString()
+
   evalInWindow(
     `
-      document.cookie = '_dd_s=expired=0; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/'
+      document.cookie = '_dd_s=expired=0; expires=${expires}; path=/'
     `
   ).catch((error) => logger.error('Error while ending session:', error))
 }
--- a/packages/core/src/domain/session/README.md
+++ b/packages/core/src/domain/session/README.md
@@ -1,0 +1,182 @@
+## possible states:
+
+| state      | value                                                        |
+| ---------- | ------------------------------------------------------------ |
+| NotStarted | `{}`                                                         |
+| Expired    | `{expired: '1'}` or `expire >  15 minutes` or `created > 4H` |
+| Tracked    | `{id: 'xxxx-xx-xx}`                                          |
+| NotTracked | `{rum: 0}` or `{logs: 0}`                                    |
+
+## session initializion
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+state fork_state2 <<fork>>
+
+[*] --> fork_state
+
+fork_state --> NotStarted
+fork_state --> Expired
+fork_state --> Tracked
+fork_state --> NotTracked
+NotStarted --> Expired: initializeSession()
+
+Expired --> fork_state2
+Tracked --> fork_state2
+NotTracked --> fork_state2
+fork_state2 --> [*]: expandOrRenew()
+```
+
+## On User Activity
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+
+[*] --> fork_state
+
+fork_state --> [*]: expandOrRenew()
+```
+
+## On Visibility Change
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+state visibility <<choice>>
+
+[*] --> visibility: VisibilityChange
+
+visibility --> [*]: hidden
+visibility --> fork_state : visible
+
+fork_state --> [*]: extendOrExpire()
+```
+
+## On resume from frozen state
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+state fork_state2 <<fork>>
+
+[*] --> fork_state: Resume
+
+fork_state --> NotStarted
+fork_state --> Expired
+fork_state --> Tracked
+fork_state --> NotTracked
+
+NotStarted --> Expired: reinitializeSession()
+Expired --> fork_state2
+Tracked --> fork_state2
+NotTracked --> fork_state2
+
+fork_state2 --> [*]: extendOrExpire()
+```
+
+## Watch (every second)
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+
+
+[*] --> fork_state
+fork_state --> [*]: extendOrExpire()
+```
+
+## 3rd party cookie clearing
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+state join_state <<join>>
+NotStarted2: NotStarted
+
+[*] --> fork_state
+
+fork_state --> NotStarted
+fork_state --> Expired
+fork_state --> Tracked
+fork_state --> NotTracked
+
+Expired --> join_state
+Tracked --> join_state
+NotTracked --> join_state
+NotStarted --> join_state
+
+join_state --> NotStarted2 : clearCookies()
+
+NotStarted2 --> [*]
+```
+
+## Expand Or Renew
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+state fork_state2 <<fork>>
+state fork_state3 <<fork>>
+
+
+state expandOrRenew() {
+  [*] --> fork_state: expandOrRenew()
+  Tracked2: Tracked
+  NotTracked2: NotTracked
+
+  fork_state --> NotStarted
+  fork_state --> Expired
+  fork_state --> Tracked
+  fork_state --> NotTracked
+
+  NotStarted --> [*]
+
+  Expired --> fork_state2: renew()
+  Tracked --> fork_state2: extend()
+  NotTracked --> fork_state2: extend()
+
+  fork_state2 --> fork_state3: computeTrackingType()
+
+  fork_state3 --> Tracked2
+  fork_state3 --> NotTracked2
+
+  Tracked2 --> [*]
+  NotTracked2 --> [*]
+}
+```
+
+> [!NOTE]  
+> Because `computeTrackingType()` happens at every `expandOrRenew()`, it is in theory possible to switch from `Tracked` to `NotTracked` and vice versa within an active session. However, this is not expected to happen in practice at this time.
+
+## Extend Or Expire
+
+```mermaid
+stateDiagram-v2
+state fork_state <<fork>>
+
+state extendOrExpire() {
+  [*] --> fork_state : extendOrExpire()
+  Tracked2: Tracked
+  NotTracked2: NotTracked
+  Expired2: Expired
+
+  fork_state --> NotStarted
+  fork_state --> Expired
+  fork_state --> Tracked
+  fork_state --> NotTracked
+
+  Tracked --> Tracked2: extend()
+  NotTracked --> NotTracked2: extend()
+  Expired --> Expired2: expire()
+
+  NotStarted --> [*]
+  Tracked2 --> [*]
+  NotTracked2 --> [*]
+  Expired2 --> [*]
+}
+```
+
+> [!NOTE]  
+> Because the `Expired` state could happens because a session timed out, `expire()` will normalize explicitely the state to `isExpired=1`
--- a/packages/core/src/domain/session/README.md
+++ b/packages/core/src/domain/session/README.md
@@ -7,7 +7,7 @@
 | Tracked    | `{id: 'xxxx-xx-xx}`                                          |
 | NotTracked | `{rum: 0}` or `{logs: 0}`                                    |
 
-## session initializion
+## session initialization
 
 ```mermaid
 stateDiagram-v2
--- a/packages/core/src/domain/session/README.md
+++ b/packages/core/src/domain/session/README.md
@@ -7,7 +7,12 @@
 | Tracked    | `{id: 'xxxx-xx-xx}`                                          |
 | NotTracked | `{rum: 0}` or `{logs: 0}`                                    |
 
-## session initialization
+Other terminology:
+
+- A _started_ session is a session that is either `Expired`, `Tracked`, or `NotTracked`.
+- An _active_ session is a session that is either `Tracked` or `NotTracked`.
+
+## start session
 
 ```mermaid
 stateDiagram-v2
@@ -20,7 +25,7 @@ fork_state --> NotStarted
 fork_state --> Expired
 fork_state --> Tracked
 fork_state --> NotTracked
-NotStarted --> Expired: initializeSession()
+NotStarted --> Expired: start()
 
 Expired --> fork_state2
 Tracked --> fork_state2
@@ -68,7 +73,7 @@ fork_state --> Expired
 fork_state --> Tracked
 fork_state --> NotTracked
 
-NotStarted --> Expired: reinitializeSession()
+NotStarted --> Expired: restartSession()
 Expired --> fork_state2
 Tracked --> fork_state2
 NotTracked --> fork_state2
--- a/packages/core/src/domain/session/README.md
+++ b/packages/core/src/domain/session/README.md
@@ -1,11 +1,17 @@
 ## possible states:
 
-| state      | value                                                        |
-| ---------- | ------------------------------------------------------------ |
-| NotStarted | `{}`                                                         |
-| Expired    | `{expired: '1'}` or `expire >  15 minutes` or `created > 4H` |
-| Tracked    | `{id: 'xxxx-xx-xx}`                                          |
-| NotTracked | `{rum: 0}` or `{logs: 0}`                                    |
+| state                   | value                                                        |
+| ----------------------- | ------------------------------------------------------------ |
+| NotStarted<sup>\*</sup> | `{}`                                                         |
+| Expired                 | `{expired: '1'}` or `expire >  15 minutes` or `created > 4H` |
+| Tracked                 | `{id: 'xxxx-xx-xx}`                                          |
+| NotTracked              | `{rum: 0}` or `{logs: 0}`                                    |
+
+(<sup>\*</sup>) `NotStarted` is a state that can happen in a few different ways:
+
+- First load of the page if there is no cookie present already
+- After the cookie has been deleted by a 3rd party (user, ad blocker, ...)
+- After the cookie has expired (it is deleted by the browser)
 
 Other terminology:
 
@@ -25,7 +31,7 @@ fork_state --> NotStarted
 fork_state --> Expired
 fork_state --> Tracked
 fork_state --> NotTracked
-NotStarted --> Expired: start()
+NotStarted --> Expired: startSession()
 
 Expired --> fork_state2
 Tracked --> fork_state2
--- a/packages/core/src/domain/session/README.md
+++ b/packages/core/src/domain/session/README.md
@@ -184,4 +184,4 @@ state extendOrExpire() {
 ```
 
 > [!NOTE]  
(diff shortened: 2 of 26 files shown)


test/e2e/lib/helpers/browser.ts
```diff
@@ -1,8 +1,10 @@
 import * as os from 'os'
 
-// typing issue for execute https://github.com/webdriverio/webdriverio/issues/3796
-export function browserExecute(fn: any) {
-  return browser.execute(fn)
+export function browserExecute<ReturnValue, InnerArguments extends any[]>(
+  fn: (...args: InnerArguments) => ReturnValue,
+  ...args: InnerArguments
+): Promise<ReturnValue> {
+  return browser.execute(fn, ...args)
```
thomas-lebeau, line 7: **💬 suggestion:** ‏I needed the `InnerArguments`. However, because the linked issue seems to be closed, I propose to refactor (in a separate PR) the tests to use `browser.execute()` directly. wdyt?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -124,9 +132,13 @@ export function startSessionStore<TrackingType extends string>(
   }
 
   function expandOrRenewSessionState(sessionState: SessionState) {
+    if (!sessionState.id) {
+      return false
+    }
```
BenoitZugmeyer, line 137: **💬 suggestion:** ‏use `isSessionInitialized`

test/e2e/lib/helpers/browser.ts
BenoitZugmeyer: Sounds good to me

test/e2e/lib/helpers/session.ts
```diff
@@ -1,16 +1,25 @@
-import { SESSION_STORE_KEY } from '@datadog/browser-core'
-import { deleteAllCookies } from './browser'
+import { SESSION_STORE_KEY, SESSION_TIME_OUT_DELAY } from '@datadog/browser-core'
+import {
+  getInitialSessionState,
+  toSessionState,
+  toSessionString,
+} from '@datadog/browser-core/src/domain/session/sessionState'
```
BenoitZugmeyer, line 6: **💭 thought:** ‏Types and constants are fine, but e2e tests should not rely on library functions.

test/e2e/lib/helpers/session.ts
```diff
@@ -1,16 +1,25 @@
...
+import { setCookie } from './browser'
 
 export async function renewSession() {
   await expireSession()
   const documentElement = await $('html')
   await documentElement.click()
-  expect(await findSessionCookie()).toBeDefined()
+
+  const session = await getSessionFromCookie()
+  expect(session.id).not.toBe('null')
```
BenoitZugmeyer, line 15: **💬 suggestion:** ‏
```suggestion
  expect(await findSessionCookie()).not.toContain('id=null')
```

BenoitZugmeyer (commented):
Setting `id=null` for expired sessions will be problematic on websites migrating to the new Browser SDK version. Ex: a user has two tabs open on the same website. The website upgrades the SDK version. The user refreshes one tab. Now there is two Browser SDK versions active. The up-to-date SDK can write a cookie with `id=null` and the other will pick it up and generate a session with a `null` id.

Setting a different value like `expired` (not `id=expired`, just a plain `expired` string) could be more tolerant for the migration.

(follow-up unrelated to that specific issue): Also it can be useful to explicitly state what ended the session: Is it a session expiration? Is it an explicit `DD_RUM.stopSession()`? Is it a consent withdrawal `DD_RUM.setTrackingConsent('not-granted')`? In the future we might want to use this information to prevent an automatic session renewal based on this information. But for now just having an `expired` value would be enough.

amortemousque:
**🔨 warning:** ‏ I know that I approved the RFC, but a new scenario comes to mind that could be problematic. With this strategy if a user switch to a new tab and the browser decides to [freezes that page](https://developer.chrome.com/docs/web-platform/page-lifecycle-api#state-frozen). When the user goes back after 4 hours to the tab the session will not be renewed.

packages/core/src/domain/session/sessionState.ts
```diff
@@ -15,8 +14,22 @@ export interface SessionState {
...
+  return session.id !== undefined
+}
+
 export function isSessionInExpiredState(session: SessionState) {
-  return isEmptyObject(session)
+  // an expired session is `{id = null}` or `{id = null, lock = whatever}`
+  return (
+    (Object.keys(session).length === 1 && session.id === 'null') ||
+    (Object.keys(session).length === 2 && session.id === 'null' && session.lock !== undefined)
+  )
```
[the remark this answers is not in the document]
thomas-lebeau: Before it was only checking if `session` is an empty object and it was sometimes called with `{lock: 'xxxxxxx'}` or `{ rum: '0' }` which resulted in a returning `false` like the session is **not** expired even though there is no session id.

Now that we initialise the session with `id=null` or `expired=0`, it is never an empty object, however I need to rely on the object key to keep the same behaviour as before. (e.g. `{expired: '0', rum: '0'}` is **not** expired)

I agree this is far from ideal and I'd like to dig into it, but maybe that'd be better as a separate PR as I'm afraid this touches a lot of thing not directly related to this features. wdyt?

test/e2e/lib/helpers/browser.ts
thomas-lebeau: here: https://github.com/DataDog/browser-sdk/pull/2700

packages/core/src/domain/session/sessionState.ts
```diff
@@ -1,22 +1,40 @@
...
+export const enum SessionExpiredReason {
+  UNKNOWN = '0',
+}
+
 export interface SessionState {
   id?: string
   created?: string
   expire?: string
   lock?: string
+  expired?: SessionExpiredReason
```
amortemousque, line 17: **💬 suggestion:** ‏I have few suggestions:
- I found confusing that an expired session is `{expired = '0'}` I would expect `expired=true` or `expired=1`. 
- The name "expired" is quite close to "expire" maybe we could name it "isExpired" to better distinguish them
- Maybe we could avoid having the expire reason since for now we only have UNKNOWN. If we want to anticipate keeping the reason we could use `ìsExpired=1` and introduce the enum with 2, and 3 later

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -23,11 +23,17 @@ export function initCookieStrategy(cookieOptions: CookieOptions): SessionStoreSt
...
+    clearSession: () => setInitialSessionCookie(cookieOptions),
   }
 
   tryOldCookiesMigration(cookieStore)
 
+  const sessionCookie = retrieveSessionCookie()
+
+  if (!isSessionInitialized(sessionCookie)) {
+    setInitialSessionCookie(cookieOptions)
+  }
```
amortemousque, line 35: **🔨 warning:** ‏ FMU the cookieStore methods are ment to be used with the lock mechanism (aka SessionStoreOperations) to prevent cookie concurrent access issues. In here if the `setInitialSessionCookie` fails to write in the cookie, the session won't start. @bcaudan can you confirm or deny what I'm saying 🙂

You could probably  lifted up the `setInitialSessionCookie()` logic in the `sessionStore` and use the cookie `SessionStoreOperations`

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -70,6 +70,7 @@ export function startSessionManager<TrackingType extends string>(
     }
   })
   trackVisibility(configuration, () => sessionStore.expandSession())
+  trackResume(configuration, () => sessionStore.reinitializeSession())
```
amortemousque, line 73: **👏 praise:** ‏Nice 😀

packages/core/src/domain/session/sessionState.ts
```diff
@@ -1,22 +1,40 @@
...
+}
+
 export function isSessionInExpiredState(session: SessionState) {
-  return isEmptyObject(session)
+  // // an expired session is `{expired = '0'}` or `{expired = '0', lock = whatever}`
+  return (
+    (Object.keys(session).length === 1 && session.expired !== undefined) ||
+    (Object.keys(session).length === 2 && session.expired !== undefined && session.lock !== undefined)
+  )
 }
```
bcaudan, line 38: **💭 thought:** ‏ ‏I am a bit confused with the relationship between expired / initial / initialized 😕
Could we clarify that somehow?

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
bcaudan: > FMU the cookieStore methods are ment to be used with the lock mechanism (aka SessionStoreOperations) to prevent cookie concurrent access issues.

+1

> You could probably lifted the setInitialSessionCookie() logic up in the sessionStore and use the cookie SessionStoreOperations

Not sure if it is exactly what you have in mind but doing that in sessionStore instead of doing it in every store strategy would be nice if possible.

test/e2e/lib/helpers/session.ts
```diff
@@ -21,3 +24,9 @@ export async function findSessionCookie() {
   // in this case.
   return cookies[0]?.value || undefined
 }
+
+// export async function getSessionFromCookie() {
+//   const cookies = await browser.getCookies(SESSION_STORE_KEY)
+
+//   return toSessionState(cookies[0]?.value)
+// }
```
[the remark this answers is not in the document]
thomas-lebeau: Ooops, that's a mistake. thanks!

developer-extension/src/panel/components/tabs/infosTab.tsx
```diff
@@ -257,9 +257,12 @@ function formatSessionType(value: string, ...labels: string[]) {
 }
 
 function endSession() {
+  const fourHours = 1000 * 60 * 60 * 4
+  const expires = new Date(Date.now() + fourHours).toUTCString()
+
   evalInWindow(
     `
-      document.cookie = '_dd_s=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/'
+      document.cookie = '_dd_s=expired=0; expires=${expires}; path=/'
```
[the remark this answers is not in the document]
thomas-lebeau: That was my first approach, however, Benoit [pointed out](https://github.com/DataDog/browser-sdk/pull/2673#discussion_r1553248576) that this would not work for Logs?

test/e2e/scenario/trackingConsent.scenario.ts
```diff
@@ -39,7 +39,7 @@ describe('tracking consent', () => {
         await flushEvents()
 
         expect(intakeRegistry.rumActionEvents).toEqual([])
-        expect(await findSessionCookie()).toBeUndefined()
+        expect(await findSessionCookie()).toContain('expired=0')
```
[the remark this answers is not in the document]
thomas-lebeau: There is actually only one that actually check `.toContain('expired=0')`
Other are similar but not quite the same (`.not.toContain('expired=0')`, `.toBe('expired=0')`). 

Also I personally find the jasmine notation easier to read.

developer-extension/src/panel/components/tabs/infosTab.tsx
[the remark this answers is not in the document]
bcaudan: Yes, valid points!

packages/core/src/domain/session/sessionManager.ts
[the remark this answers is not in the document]
thomas-lebeau: 👍 Added!

packages/core/src/domain/session/sessionState.ts
thomas-lebeau: Yeah, I agree this is confusing. I'm not sure how to improve that though:

- initial state: `{expired: 0}`
- initialized: I wish I could call it `!isEmpty(sessionState)`, however, this is not true. There is some cases where the cookie is only `rum=0`
- expired state: here again I wish I could only check if `expired` is defined. But as I [commented here](https://github.com/DataDog/browser-sdk/pull/2673#discussion_r1555974777) this is not true.

The session management is definitively confusing, and would be worth refactoring to make changes easier and safer (Although I'm not totally sure how). 

Do you have any suggestions that would fit in the scope of this PR?

packages/core/src/domain/session/README.md
thomas-lebeau, line 1: tip: use _Display the rich diff_ on this file

packages/core/src/domain/session/sessionState.ts
```diff
@@ -1,22 +1,52 @@
...
+  if (lock) {
+    session.lock = lock
+  }
+
+  return session
+}
+
+export function isSessionStarted(session: SessionState) {
+  return !(isEmptyObject(session) || (Object.keys(session).length === 1 && 'lock' in session))
+}
```
thomas-lebeau, line 37: **🔨 warning:** ‏ This methods need to take `lock` into account because it is leaked to the session state in the `process()` method in [sessionStoreOperations.ts](https://github.com/DataDog/browser-sdk/blob/d0e13f936569bfeb68d6e697baf22ba87b3c828e/packages/core/src/domain/session/sessionStoreOperations.ts#L45)

This is a temporary solution in order to avoid larger changes in this PR.
Ultimately I would like to fix sessionStoreOperations to not leak the `lock` property

packages/core/src/domain/session/README.md
```diff
@@ -0,0 +1,182 @@
+## possible states:
+
+| state      | value                                                        |
+| ---------- | ------------------------------------------------------------ |
+| NotStarted | `{}`                                                         |
+| Expired    | `{expired: '1'}` or `expire >  15 minutes` or `created > 4H` |
+| Tracked    | `{id: 'xxxx-xx-xx}`                                          |
+| NotTracked | `{rum: 0}` or `{logs: 0}`                                    |
+
+## session initializion
```
thomas-lebeau, line 10:
```suggestion
## session initialization
```

test/e2e/scenario/rum/sessions.scenario.ts
```diff
@@ -115,4 +116,26 @@ describe('rum sessions', () => {
         expect(intakeRegistry.replaySegments.length).toBe(1)
       })
   })
+
+  describe('third party cookie clearing', () => {
+    createTest('after a 3rd party clears the cookies, stop the session')
```
bcaudan, line 121: **💭 thought:** ‏I am wondering if this test was not passing before those changes 🤔
FMU, what we want to achieve with this change is not to stop the session but to avoid to create new ones, it is probably what we should test.

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -123,9 +131,30 @@ export function startSessionStore<TrackingType extends string>(
...
+            return getInitialSessionState(sessionState)
+          }
+        },
+        after: (sessionState) => {
+          sessionCache = sessionState
+        },
+      },
+      sessionStoreStrategy
+    )
+  }
```
bcaudan, line 148: **💬 suggestion:** ‏if this operation pass the session from not started to started, should we name it `startSession` instead?
and similarly reinitialize -> restart

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -153,19 +182,10 @@ export function startSessionStore<TrackingType extends string>(
...
-  }
-
-  function isActiveSession(sessionState: SessionState) {
-    // created and expire can be undefined for versions which was not storing them
-    // these checks could be removed when older versions will not be available/live anymore
-    return (
-      (sessionState.created === undefined || dateNow() - Number(sessionState.created) < SESSION_TIME_OUT_DELAY) &&
-      (sessionState.expire === undefined || dateNow() < Number(sessionState.expire))
-    )
+    return getInitialSessionState()
```
bcaudan, line 188: **💬 suggestion:** ‏`retrieveActiveSession` seems a bit odd now, could `retrieveStartedSession` be more accurate?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -65,11 +66,17 @@ export function startSessionStore<TrackingType extends string>(
   const watchSessionTimeoutId = setInterval(watchSession, STORAGE_POLL_DELAY)
   let sessionCache: SessionState = retrieveActiveSession()
 
+  initializeSession()
```
bcaudan, line 69: **💬 suggestion:** ‏it is a bit confusing to me that `retrieveActiveSession` and `initializeSession` are both accessing the store and having logic around the initial state.
Could we init `sessionCache` with the `initialSessionState` and let `initializeSession` handling all the store accesses and initial state logic ?

packages/core/src/domain/session/sessionState.spec.ts
```diff
@@ -35,12 +63,16 @@ describe('session state utilities', () => {
...
+    })
+
+    it('should handle expired session', () => {
       expect(toSessionState(SERIALIZED_EXPIRED_SESSION)).toEqual(EXPIRED_SESSION)
     })
 
     it('should handle invalid session strings', () => {
       const sessionString = '{invalid: true}'
-      expect(toSessionState(sessionString)).toEqual(EXPIRED_SESSION)
+      expect(toSessionState(sessionString)).toEqual({})
```
bcaudan, line 75: **💬 suggestion:** ‏what about extracting `NOT_STARTED_SESSION` constant?

packages/core/src/domain/session/sessionManager.spec.ts
```diff
@@ -49,11 +54,19 @@ describe('startSessionManager', () => {
...
+    expect(sessionManager.findActiveSession()?.isExpired).toBeUndefined()
+
     expect(getCookie(SESSION_STORE_KEY)).toMatch(/id=[a-f0-9-]+/)
+    expect(getCookie(SESSION_STORE_KEY)).not.toContain('isExpired=1')
+  }
+
+  function expectSessionToNotBeDefined(sessionManager: SessionManager<FakeTrackingType>) {
+    expect(sessionManager.findActiveSession()).toBeUndefined()
+    expect(getCookie(SESSION_STORE_KEY)).toContain('isExpired=1')
   }
```
bcaudan, line 66: **💬 suggestion:** more accurate naming‏
```suggestion
  function expectSessionToBeExpired(sessionManager: SessionManager<FakeTrackingType>) {
    expect(sessionManager.findActiveSession()).toBeUndefined()
    expect(getCookie(SESSION_STORE_KEY)).toContain('isExpired=1')
  }
```

packages/core/src/domain/session/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+  NotStarted --> [*]
+  Tracked2 --> [*]
+  NotTracked2 --> [*]
+  Expired2 --> [*]
+}
+```
+
+> [!NOTE]  
+> Because the `Expired` state could happens because a session timed out, `expire()` will normalize explicitely the state to `isExpired=1`
```
hestonhoffman, line 187:
```suggestion
> Because a session time out can result in an `Expired` state, `expire()` explicitly normalizes the state to `isExpired=1`
```

hestonhoffman (approved):
Small edit. Docs look good

packages/core/src/domain/session/sessionState.ts
```diff
@@ -1,31 +1,58 @@
...
+  return {
+    isExpired: EXPIRED,
+  }
+}
+
+export function isSessionInNotStartedState(session: SessionState) {
   return isEmptyObject(session)
 }
 
+export function isSessionStarted(session: SessionState) {
```
amortemousque, line 31: **💬 suggestion:** ‏To be consistent other state functions 
```suggestion
export function isSessionInStartedState(session: SessionState)
```

packages/core/src/domain/session/sessionState.ts
```diff
@@ -1,31 +1,58 @@
...
 }
 
-export function isSessionInExpiredState(session: SessionState) {
+export function getExpiredSessionState(): SessionState {
+  return {
+    isExpired: EXPIRED,
+  }
+}
+
+export function isSessionInNotStartedState(session: SessionState) {
```
amortemousque, line 27: **💬 suggestion:** ‏Could be nice to add a comment to know when this state can happen (first load of the page + potential cookie deletion)

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -123,9 +130,30 @@ export function startSessionStore<TrackingType extends string>(
     return sessionState
   }
 
+  function startSession() {
```
amortemousque, line 133: **💬 suggestion:** ‏I found confusing that `startSession` uses `getExpiredSessionState()` to start the session. Maybe a comment or another name like `initSession` could help 🤷‍♂️?

packages/core/src/domain/session/sessionState.ts
```diff
@@ -1,31 +1,58 @@
...
+const SESSION_ENTRY_REGEXP = /^([a-zA-Z]+)=([a-z0-9-]+)$/
 const SESSION_ENTRY_SEPARATOR = '&'
 
+export const EXPIRED = '1'
+
 export interface SessionState {
   id?: string
   created?: string
   expire?: string
   lock?: string
```
bcaudan, line 15: **❓ question:** with the change around store operations, ‏could we remove `lock` from `sessionState`?

packages/core/src/domain/session/sessionState.ts
thomas-lebeau: yeah, although as I mentioned in the doc, `started` is not really a state, it is either one of the following states: `Expired`, `Tracked`, or `NotTracked`.

packages/core/src/domain/session/sessionStore.ts
thomas-lebeau: `initializeSession` was confusing too. see this comment: https://github.com/DataDog/browser-sdk/pull/2673#discussion_r1568922841

We could rename `NotStarted` to `NotInitialized` and `startSession()` to `initializeSession` could help. However IMHO `init` and `start` are synonyms and it will feel similarly confusing to use `getExpiredSessionState()` here.

I think the state diagrams in the doc should helps to understand that a session starts (or is initialized) in the `Expired` state:

```mermaid
stateDiagram-v2
state fork_state <<fork>>
state fork_state2 <<fork>>

[*] --> fork_state

fork_state --> NotStarted
fork_state --> Expired
fork_state --> Tracked
fork_state --> NotTracked
NotStarted --> Expired: startSession()

Expired --> fork_state2
Tracked --> fork_state2
NotTracked --> fork_state2
fork_state2 --> [*]: expandOrRenew()
```

amortemousque: Fortunately the diagrams are very cool 🙂  Fine by me!

packages/core/src/domain/session/sessionState.ts
amortemousque: It still seems strange to me that "NotStarted" is a state but "Started" is not. Basically having only negated state confuses me, but I won't block the PR for that. This is clearly a tough one 🙂

amortemousque (approved):

Issue #2566, opened 2024-01-11
🐛 Resource Timings are inaccurate when ServiceWorker is involved


huwyca:
**Describe the bug**

NOTE: This was discussed and shown at length in a call with support on 12/12/2023 with traeger.meyer@datadoghq.com and shayne.sebro@datadoghq.com. There should be a recording available of that call. Datadog 1345744 // 1963 - RUM Resource Timing - Customer call

Service worker based resource network requests are being reported improperly by browser-sdk rum resourceCollection. This causes the duration to be incorrect as well as causing other fields to be missing from the RUM entry.

Browser: Chrome (Firefox and Safari don't seem to support workerStart yet)
BrowserSDK versions affected: Seems to affect multiple major versions including v5 and v4

This seems to happen mainly due to `requestStart` being set to `workerStart` and those falling before the `connectionStart`/`connectionEnd` timings when a ServiceWorker. I notice it most commonly when a service worker is intercepting a request over http2 re-using an existing TCP connection to a domain.

When a service worker is involved in a network request it changes the fields of the PerformanceResourceTiming.
In addition to setting the `workerStart` property to be non-zero it also changes the `requestStart` time to be set to the `workerStart` property. This causes a logic check to fail in the `resourceUtils.ts#toValidEntry` function to fail due to the `entry.requestStart` time falling before the other timings in the `areInOrder` check.
https://github.com/DataDog/browser-sdk/blob/43c44563a9b6671255ee4a591197c125a33b3f68/packages/rum-core/src/domain/resource/resourceUtils.ts#L138-L157

This causes the `matchRequestTiming` function to not find a correct matching timing due to it filtering out the timing due to `filter(toValidEntry)` removing the timings from service workers. https://github.com/DataDog/browser-sdk/blob/43c44563a9b6671255ee4a591197c125a33b3f68/packages/rum-core/src/domain/resource/matchRequestTiming.ts#L24-L37

The `matchingTiming` being set to `undefined` from `matchRequestTiming` then results in not calculating any `correspondingTimingOverrides` with `computePerformanceEntryMetrics`. This has a couple of knock on effects:

1. Due to not having the correct entry metrics the duration being reported in the RumEvent is the timestamp of when the datadog rum XHR onLoad handler fires instead of when the resource finished. https://github.com/DataDog/browser-sdk/blob/43c44563a9b6671255ee4a591197c125a33b3f68/packages/rum-core/src/domain/resource/resourceCollection.ts#L76-L78
[...]

PR #2688, opened 2024-04-04
⚗ [RUM-2782] validate resource timings more granularly

BenoitZugmeyer:
## Motivation

Before this PR, we dropped any `PerformanceResourceEntry` that had inconsistent timings. With this commit, when the experimental feature "tolerant_resource_timings" is enabled, the various timings are validated independently.

This should improve the situation on Chrome when a Service Worker is involved. In this case, most of the time, connection timings are unexpectedly set after the `requestStart` timing (see [Chromium issue][1]). Before this change, the whole `PerformanceResourceEntry` would be ignored, and no timings would be included. With this change (and the flag enabled), relevant timings will be correctly defined.

Related issue: #2566

[1]: https://issues.chromium.org/issues/323703325

## Changes

Before/without the flag:
Screenshot 2024-04-04 at 18 24 54

After:
Screenshot 2024-04-04 at 18 24 39

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/resourceUtils.ts
+++ b/packages/rum-core/src/domain/resource/resourceUtils.ts
@@ -2,8 +2,10 @@ import type { RelativeTime, ServerDuration } from '@datadog/browser-core'
 import {
   addTelemetryDebug,
   elapsed,
+  ExperimentalFeature,
   getPathName,
   includes,
+  isExperimentalFeatureEnabled,
   isValidUrl,
   ResourceType,
   toServerDuration,
@@ -19,8 +21,8 @@ export interface PerformanceResourceDetails {
   dns?: PerformanceResourceDetailsElement
   connect?: PerformanceResourceDetailsElement
   ssl?: PerformanceResourceDetailsElement
-  first_byte: PerformanceResourceDetailsElement
-  download: PerformanceResourceDetailsElement
+  first_byte?: PerformanceResourceDetailsElement
+  download?: PerformanceResourceDetailsElement
 }
 
 export const FAKE_INITIAL_DOCUMENT = 'initial_document'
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -113,28 +115,33 @@ export function computePerformanceResourceDetails(
...
     details.redirect = formatTiming(startTime, redirectStart, redirectEnd)
   }
 
   return details
 }
 
 export function toValidEntry(entry: RumPerformanceResourceTiming) {
+  if (isExperimentalFeatureEnabled(ExperimentalFeature.TOLERANT_RESOURCE_TIMINGS)) {
+    return entry
+  }
```
BenoitZugmeyer, line 143: Note: the goal is to remove this function once the flag is enabled

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -113,28 +115,33 @@ export function computePerformanceResourceDetails(
   }
 
   // Make sure a connection occurred
-  if (connectEnd !== fetchStart) {
+  if (fetchStart < connectEnd) {
```
BenoitZugmeyer, line 118: Note: those conditions were changed (ex: from `fetchStart !== connectEnd` to `fetchStart < connectEnd`) and are not behind the flag, but it should not have any impact, because some other conditions are checked in `toValidEntry`). Ex:

Before this PR, we check that `fetchStart <= connectEnd` in `toValidEntry`, then check that `fetchStart !== connectEnd` here in `computePerformanceResourceDetails`.

With this PR, without the flag, we still check that `fetchStart <= connectEnd` in `toValidEntry`, then check that `fetchStart < connectEnd` here in `computePerformanceResourceDetails`, so this is equivalent.

With this PR, with the flag, we won't have `toValidEntry` anymore, we just check that `fetchStart < connectEnd`, which is still the same.

thomas-lebeau (approved):

PR #2690, opened 2024-04-05
✨ [RUM-162] Truncate resources URL containing data URLs

cy-moi:
## Motivation
We're currently collecting the full url data of resources using data: based resource.url . 
This is highly inefficient for the ingestion and indexing side and may contain sensitive informations. This data is always truncated to 24k characters when being indexed, which would result in partial images. 
In this case, we want to truncate the partial data till only keep the useful information (MIME type, encodings). 

## Changes
We truncate data:url to the embedding codec when they exceeds 24k
Examples
Screenshot 2024-04-08 at 18 21 51

Edit (Added truncate indicator [...])
Screenshot 2024-04-10 at 17 21 26

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -215,3 +215,10 @@ export function computeSize(entry: RumPerformanceResourceTiming) {
 export function isAllowedRequestUrl(configuration: RumConfiguration, url: string) {
   return url && !configuration.isIntakeUrl(url)
 }
+
+const DATA_URL_REGEX = /data:([-\w]+\/[-+\w.]+)?(;?\w+=[-\w]+)*(;base64)?/g
```
amortemousque, line 219: **💬 suggestion:** ‏Looking at the [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs) the format of data urls look like this: `data:[<mediatype>][;base64],<data>`.
So the regex could be simplified as follow: `data:(.+)(;base64)?,`

packages/rum-core/src/domain/resource/resourceUtils.spec.ts
```diff
@@ -313,3 +314,33 @@ describe('shouldTrackResource', () => {
     expect(isAllowedRequestUrl(configuration, 'https://my-domain.com/hello?a=b')).toBe(true)
   })
 })
+
+describe('findDataUrlAndTruncate', () => {
```
amortemousque, line 318: **💬 suggestion:** ‏I think it could also be nice to test data url on other exotic [mime types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) like you do for svg images.  Ex : "application/vnd.openxmlformats-officedocument.presentationml.presentation"

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -215,3 +215,10 @@ export function computeSize(entry: RumPerformanceResourceTiming) {
...
+
+export function findDataUrlAndTruncate(url: string): string | null {
```
amortemousque, line 221: **💬 suggestion:** ‏
```suggestion
export function sanitizeDataUrl(url: string): string | null {
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -91,7 +92,7 @@ function processRequest(
         duration,
         method: request.method,
         status_code: request.status,
-        url: request.url,
+        url: findDataUrlAndTruncate(request.url) ?? request.url,
```
amortemousque, line 95: **💬 suggestion:** ‏To be more explicit you could split the concern like so
```suggestion
        url: isDataUrl(request.url) ? sanitizeDataUrl(request.url) : request.url,
```

Also, it could be nice to test this line :)

packages/rum/src/domain/record/serialization/serializeAttribute.ts
```diff
@@ -68,10 +67,6 @@ export function serializeAttribute(
...
   }
 
-  // Minimum Fix for customer.
-  if (attributeValue.length > MAX_ATTRIBUTE_VALUE_CHAR_LENGTH && attributeValue.slice(0, 5) === 'data:') {
-    return 'data:truncated'
-  }
-
-  return attributeValue
+  // Truncate data:url to avoid performance impact
+  return findDataUrlAndTruncate(attributeValue) ?? attributeValue
```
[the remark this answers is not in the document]
cy-moi: Yes, it was intended. But then indeed the truncation would prevent some allowed data to show up as well. So fixing on this right now.

packages/rum-core/src/domain/resource/resourceUtils.ts
cy-moi: This has been simplified

packages/rum/src/domain/record/serialization/serializeAttribute.ts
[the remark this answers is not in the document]
cy-moi: I've added the length check back to the recorder as well as for the resource url.

packages/rum-core/src/domain/resource/resourceUtils.spec.ts
```diff
@@ -313,3 +316,47 @@ describe('shouldTrackResource', () => {
...
+    expect(sanitizeDataUrl(exoticTypeDataUrl)).toEqual(
+      'data:application/vnd.openxmlformats;fileName=officedocument.presentationxml;base64,'
+    )
+  })
+
+  it('returns the original url when the data url is within limit', () => {
+    expect(isDataUrlTooLong(`data:,${longString.substring(5)}`)).toEqual(false)
+  })
+
+  it('returns null when no data url found', () => {
```
thomas-lebeau, line 359: **🥜 nitpick:** ‏ 
```suggestion
  it('returns false when no data url found', () => {
```

packages/rum-core/src/domain/resource/resourceUtils.spec.ts
```diff
@@ -313,3 +316,47 @@ describe('shouldTrackResource', () => {
...
 })
+
+describe('isDataUrl and sanitizeDataUrl', () => {
+  const longString = new Array(MAX_ATTRIBUTE_VALUE_CHAR_LENGTH).join('a')
+  it('returns truncated url when detects data url of json', () => {
+    const longDataUrl = `data:text/json; charset=utf-8,${longString}`
+
+    const expectedUrl = 'data:text/json; charset=utf-8,'
+    expect(isDataUrlTooLong(longDataUrl)).toEqual(true)
+    expect(sanitizeDataUrl(longDataUrl)).toEqual(expectedUrl)
```
thomas-lebeau, line 327: **🥜 nitpick:** ‏ 
```suggestion
    expect(sanitizeDataUrl(longDataUrl)).toEqual('data:text/json; charset=utf-8,')
```

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -215,3 +215,18 @@ export function computeSize(entry: RumPerformanceResourceTiming) {
 export function isAllowedRequestUrl(configuration: RumConfiguration, url: string) {
   return url && !configuration.isIntakeUrl(url)
 }
+
+const DATA_URL_REGEX = /data:(.+)?(;base64)?,/g
+export const MAX_ATTRIBUTE_VALUE_CHAR_LENGTH = 24_000
+
+export function isDataUrlTooLong(url: string): boolean {
```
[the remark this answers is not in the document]
cy-moi: Thanks! I searched the codebase and it seems like we are only using this limit length for data URL, hence making it into separate functions feels redundant. But the naming is indeed not accurate. I have renamed it to `isLongDataUrl`.

N-Boutaib (approved):

thomas-lebeau (approved):

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -215,3 +215,18 @@ export function computeSize(entry: RumPerformanceResourceTiming) {
...
+export function isLongDataUrl(url: string): boolean {
+  if (url.length <= MAX_ATTRIBUTE_VALUE_CHAR_LENGTH) {
+    return false
+  }
+  const match = url.match(DATA_URL_REGEX)
+  return !!match
+}
+
+export function sanitizeDataUrl(url: string): string {
+  return url.match(DATA_URL_REGEX)![0]
```
amortemousque, line 231: **💬 suggestion:**
It could be nice to have something indicating that it the url has been truncated like `data:[<mediatype>] [...]` (the same way we do for[ action names](https://github.com/DataDog/browser-sdk/blob/160d5fef65a055cebae3fbc50539b26d3d1075c9/packages/rum-core/src/domain/action/getActionNameFromElement.ts#L161))

cy-moi: I've added the [...] as in action names to indicate the truncation.

packages/rum-core/src/domain/resource/resourceCollection.ts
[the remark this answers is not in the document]
cy-moi: Notes on our slack discussion, we found that [here](https://www.w3.org/TR/resource-timing/) it says:
> If an HTML IMG element has a [data: URI](https://tools.ietf.org/html/rfc2397) as its source [[RFC2397](https://www.w3.org/TR/resource-timing/#bib-rfc2397)], then this resource will not be included as a [PerformanceResourceTiming](https://www.w3.org/TR/resource-timing/#dom-performanceresourcetiming) object in the [Performance Timeline](https://www.w3.org/TR/performance-timeline/#performance-timeline). By definition [data: URI](https://tools.ietf.org/html/rfc2397) contains embedded data and does not require a [fetch](https://fetch.spec.whatwg.org/#concept-fetch).

So we do not sanitize data url in this case.

amortemousque (approved):

PR #2701, opened 2024-04-09
⚡️ [RUM-3283] Do not fetch the entire git history in report-bundle-size

RomanGaignault:
## Motivation

While running report-bundle-size, we do a git fetch origin to try to find the branch common commit with main.

We should not fetch the entire git history for that: not only it has performance implications, but also produces a huge output, cluttering the CI output. Also, sometimes [it fails](https://gitlab.ddbuild.io/DataDog/browser-sdk/-/jobs/477865434)

## Changes

Use the --depth option to limit the number of commits to fetch (see [doc](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---depthltdepthgt))

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/report-bundle-size/report-as-a-pr-comment.js
+++ b/scripts/report-bundle-size/report-as-a-pr-comment.js
@@ -27,7 +27,7 @@ async function reportBundleSizesAsPrComment(localBundleSizes) {
 
 function getLastCommonCommit(baseBranch) {
   try {
-    command`git fetch origin`.run()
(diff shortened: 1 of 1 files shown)


thomas-lebeau (approved):
👍 looks good, thanks!

amortemousque (approved):

N-Boutaib (approved):

PR #2702, opened 2024-04-10
✨[RUM-3798] Report the cpu impact as a pr comment

RomanGaignault:
## Motivation

We want to know how much the Browser SDK is impacting the performances, and detect regressions early.

## Changes

Using synthetic test we trigger a micro benchmark of actions you can perform on the sdk, measure the time used and retrieves data using logs2metrics.
image

Note : the S3 bucket contains now every built bundle of each PR. As it replaces the previous one after CI (for the same PR), only the built files for last commit before merging will be kept.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


scripts/performance/cpu-performance/synthetic-trigger.js
```diff
@@ -0,0 +1,53 @@
+const { fetchHandlingError } = require('../../lib/execution-utils')
+const { getOrg2ApiKey, getOrg2AppKey } = require('../../lib/secrets')
+const { timeout } = require('../../lib/execution-utils')
+const API_KEY = getOrg2ApiKey()
+const APP_KEY = getOrg2AppKey()
+const TIMEOUT_IN_MS = 10000
+const TEST_PUBLIC_ID = 'vcg-7rk-5av'
+
+async function syntheticTrigger(prNumber, commitId) {
```
bcaudan, line 9: **💬 suggestion:** ‏‏for readability sake, it could be nice to use consistent form / naming: 
- We mostly use `doSomething` form instead of `somethingDo`, so `triggerSynthetic` could be more aligned with other namings
- Looking at the other function `getSyntheticTestResult`, it could be nice to use the same naming for the same thing and so use either `synthetic` or `syntheticTest` in both function names (and in file name)
- looking at product and API name, it is actually synthetics not synthetic

scripts/performance/index.js
```diff
@@ -0,0 +1,17 @@
...
+const { calculateBundleSizes } = require('./bundle-size/bundle-size-calculator')
+const { syntheticTrigger, getSyntheticTestResult } = require('./cpu-performance/synthetic-trigger')
+
+const RETRIES_NUMBER = 6 // Number of retries to get the synthetic test result
+
+runMain(async () => {
+  const prNumber = (await fetchPR(LOCAL_BRANCH)).number
+  const bundleSizes = calculateBundleSizes()
+  const resultId = await syntheticTrigger(prNumber, LOCAL_COMMIT_SHA)
+  await getSyntheticTestResult(resultId, RETRIES_NUMBER) // Waiting for synthetic test to finish
```
bcaudan, line 14: **💬 suggestion:** ‏what about being more expressive in the function name directly?
```suggestion
  await waitForSyntheticTestToFinish(resultId, RETRIES_NUMBER)
```

scripts/performance/cpu-performance/synthetic-trigger.js
```diff
@@ -0,0 +1,53 @@
...
+    body: JSON.stringify(body),
+  })
+  const data = await response.json()
+  return data.results[0].result_id
+}
+
+async function getSyntheticTestResult(resultId, RETRIES_NUMBER) {
+  const url = `https://api.datadoghq.com/api/v1/synthetics/tests/${TEST_PUBLIC_ID}/results/${resultId}`
+  for (let i = 0; i < RETRIES_NUMBER; i++) {
+    const response = await fetch(url, {
```
bcaudan, line 35: **❓ question:** ‏why not handling error for this fetch?

scripts/deploy/lib/deployment-utils.js
```diff
@@ -14,6 +14,10 @@ const buildDatacenterUploadPath = (datacenter, packageName, version, extension =
 // ex: datadog-rum.js
 const buildBundleFileName = (packageName, extension = 'js') => `datadog-${packageName}.${extension}`
 
+// ex: pull-request/2781/datadog-rum.js
+function buildPullRequestUploadPath(packageName, version, extension = 'js') {
+  return `PullRequest/${version}/datadog-${packageName}.${extension}`
```
bcaudan, line 19: **💬 suggestion:** ‏if `PullRequest` is use in an URL, it could be nice to use `pull-request` instead for readability sake.
It is also what is advertised in the comment.

scripts/deploy/deploy.js
```diff
@@ -27,20 +30,23 @@ const AWS_CONFIG = {
 /**
  * Deploy SDK files to CDN
  * Usage:
- * node deploy.js staging|prod staging|canary|vXXX root,us1,eu1,...
+ * node deploy.js staging|prod staging|canary|pull-request|vXXX root,pull-Request,us1,eu1,...
```
bcaudan, line 33: **💬 suggestion:** ‏no?
```suggestion
 * node deploy.js staging|prod staging|canary|pull-request|vXXX root,pull-request,us1,eu1,...
```

scripts/performance/report-as-a-pr-comment.js
```diff
@@ -0,0 +1,225 @@
...
+const LOCAL_COMMIT_SHA = process.env.CI_COMMIT_SHORT_SHA
+const ACTION_NAMES = [
+  'adderror',
+  'addaction',
+  'logmessage',
+  'startview',
+  'startstopsessionreplayrecording',
+  'addtiming',
+  'addglobalcontext',
+]
```
bcaudan, line 19: **💬 suggestion:** ‏what about extracting all the logic related to the performance data from the reporting logic?
it could be easier to follow from main if we had something like
```js
main() {
  const bundleSizes = calculateBundleSizes()
  const performanceData = await computePerformanceData() // run synthetics test under the hood
  await reportToDatadog(bundleSizes)
  await reportAsPrComment(bundleSizes, performanceData)
}
```

scripts/performance/cpu-performance/synthetic-trigger.js
RomanGaignault: With fetchHandlingError it was exiting the test as there was an error when the result was still not found.
"throw new Error(`HTTP Error Response: ${response.status} ${response.statusText}"

bcaudan: oh, I guess we consider 404 as an error.
We could probably catch server error though.

bcaudan (approved):

N-Boutaib (approved):

PR #2707, opened 2024-04-15
✨ [RUM-3902] Add privacy control for action names

cy-moi:
## Motivation
To be consistent with session Replay and ease the customer configuration, we could use the same privacy options as Replay. To avoid breaking changes, we need to add an extra init option that indicates the action names should take into account the defaultPrivacyLevel and [the HTML overrides](https://docs.datadoghq.com/real_user_monitoring/session_replay/browser/privacy_options/#override-an-html-element).

### Desired behavior
When the user opt-in for `enablePrivacyForActionName`, we mask action names when no action name attributes given and no html override.

When the user does not opt-in for this feature, we mask action names when html override is set.

## Changes
Add an opt-in field `enablePrivacyForActionName: boolean` in configuration. Once `enablePrivacyForActionName: true`, the action name computation would behave (taking into account of all privacy configs) as follow:
| Header | Header |
|--------|--------|
| allow | The action names are collected |
| mask | The action names are not collected |
| mask-user-input | The action names are collected since we don’t rely on the user input value to compute the action name |
| hidden | The actions are not collected | 

This also works for html override `data-dd-privacy` accordingly.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -209,21 +213,33 @@ function startClickAction(
...
+
   return {
     type: ActionType.CLICK,
-    target: {
-      width: Math.round(rect.width),
-      height: Math.round(rect.height),
-      selector: getSelectorFromElement(event.target, actionNameAttribute),
-    },
+    target:
+      actionName === ACTION_NAME_PLACEHOLDER && privacyEnabledForActionName ? assign({ masked: true }, target) : target,
```
amortemousque, line 236: **💭 thought:** ‏This check `actionName === ACTION_NAME_PLACEHOLDER` makes me wondering if the privacy logic could not be collocated to the action name computation. Something like
```ts
function getActionNameFromElement(element: Element, configuration: RumConfiguration): { value: string, masked: boolean }
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -6,8 +6,22 @@ import { getParentElement } from '../../browser/polyfills'
...
+export const ACTION_NAME_PLACEHOLDER = 'Masked Element'
+export function getActionNameFromElement(
+  element: Element,
+  userProgrammaticAttribute?: string,
+  privacyEnabledForActionName?: boolean
+): string {
+  // Only get the defined action name if
+  // * privacy is enabled for action name
+  // * and privacy is enabled for the Node or globally
+  if (privacyEnabledForActionName) {
```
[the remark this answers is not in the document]
cy-moi: Maybe I missed something, but I suppose we use the same strategy for `mask` and `hidden`.

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -209,21 +216,42 @@ function startClickAction(
...
+  configuration: RumConfiguration
+): ClickActionBase | undefined {
   const rect = event.target.getBoundingClientRect()
+  const privacyLevel = getNodePrivacyLevel(event.target, configuration.defaultPrivacyLevel)
+
+  // When the node is set to hidden, we do not track click action
+  // Make sure everything has been done before return
+  if (privacyLevel === NodePrivacyLevel.HIDDEN && configuration.enablePrivacyForActionName) {
+    return
+  }
```
amortemousque, line 230: **💬 suggestion:** ‏I think this condition could be done earlier in `processPointerDown`.  The privacy level could be passed to `computeClickActionBase`. This way no need for this check: https://github.com/DataDog/browser-sdk/blob/f0dacd92dfc320ac3b54f0265194e83e62394152/packages/rum-core/src/domain/action/trackClickActions.ts#L91

packages/rum-core/src/domain/action/getActionNameFromElement.ts
amortemousque, line 1: What about action names relying on innerText (cf Privacy control options for action names RFC)?

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -209,21 +216,42 @@ function startClickAction(
...
+
   return {
     type: ActionType.CLICK,
-    target: {
-      width: Math.round(rect.width),
-      height: Math.round(rect.height),
-      selector: getSelectorFromElement(event.target, actionNameAttribute),
-    },
+    target: masked ? assign({ masked }, target) : target,
```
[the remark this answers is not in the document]
cy-moi: I think masked: undefined is semantically meaningless to be in the attributes. So this line is to avoid adding "undefined" masked key.

packages/rum-core/src/domain/action/trackClickActions.ts
cy-moi: I moved it to `processPointerDown` but I do not see how we could get rid of the check of `clickActionBase` because if would still be undefined when we skip `hidden` override.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: This logic should be added when there are html privacy overriding attributes.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -1,26 +1,64 @@
...
  * It can also be retrieved from a user defined attribute.
  */
 export const DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE = 'data-dd-action-name'
+export const ACTION_NAME_PLACEHOLDER = 'Masked Element'
+export function getActionNameFromElement(
+  element: Element,
+  privacyEnabledForActionName?: boolean,
+  userProgrammaticAttribute?: string
+): { name: string; masked?: boolean } {
+  // to determine if we should add placeholder
```
amortemousque, line 17: **💬 suggestion:** ‏The implementation of the function could probably be simplified like so (to be testes :)
```ts
const userDefineActionName =
    getActionNameFromElementProgrammatically(element, DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE) ||
    (userProgrammaticAttribute && getActionNameFromElementProgrammatically(element, userProgrammaticAttribute))
  let automaticActionName: string | undefined
  if (!privacyEnabledForActionName && !userDefineActionName) {
    automaticActionName =
      getActionNameFromElementForStrategies(
        element,
        userProgrammaticAttribute,
        priorityStrategies,
        privacyEnabledForActionName
      ) ||
      getActionNameFromElementForStrategies(
        element,
        userProgrammaticAttribute,
        fallbackStrategies,
        privacyEnabledForActionName
      )
  }
  const placeholder = privacyEnabledForActionName ? PRIVATE_ACTION_NAME_PLACEHOLDER : ''

  return {
    name: userDefineActionName || automaticActionName || placeholder,
    masked: privacyEnabledForActionName && !userDefineActionName,
  }
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -1,26 +1,64 @@
...
  * Get the action name from the attribute 'data-dd-action-name' on the element or any of its parent.
  * It can also be retrieved from a user defined attribute.
  */
 export const DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE = 'data-dd-action-name'
+export const ACTION_NAME_PLACEHOLDER = 'Masked Element'
+export function getActionNameFromElement(
+  element: Element,
+  privacyEnabledForActionName?: boolean,
+  userProgrammaticAttribute?: string
+): { name: string; masked?: boolean } {
```
BenoitZugmeyer, line 16: **Suggestion**: Both `privacyEnabledForActionName` and `userProgrammaticAttribute` are configuration values (or at least they should be, for clarity). To simplify a little bit, you could just take the `RumConfiguration` as a parameter:

```ts

export function getActionNameFromElement(
  element: Element,
  {
    enablePrivacyForActionName,
    actionNameAttribute: userProgrammaticAttribute
  }: RumConfiguration
): { name: string; masked?: boolean } {
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -1,26 +1,64 @@
...
+      element,
+      userProgrammaticAttribute,
+      fallbackStrategies,
+      privacyEnabledForActionName
+    )
+
+  // get the privacy level from current or parent html attribute override
+  // If privacy is not enabled for action name, we assume the default is allowed
+  // to be consistent with current behaviors.
+  const selfPrivacyLevel = getNodePrivacyLevel(
```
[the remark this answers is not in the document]
cy-moi: `privacyEnabledForActionName` is currently a boolean indicates if the privacy enabled for action name, which also means if we apply the placeholder. 

As this is different from the defaultPrivacyLevel, here I need to re-compute the privacy level taking this boolean into account.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: Nice suggestion. Yet, these parameters were optional. I think I could either pass configuration as optional and deconstruct in the function, or keep the deconstruction and make it required with type `RumConfiguration | Record<string, never>`. Maybe there are better ways to handle it? Wdyt

BenoitZugmeyer: These parameters are optional but in practice they are always provided, appart from the spec file. Let's require a `RumConfiguration`, and in the spec file we can fake it like `{} as RumConfiguration`

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -133,7 +137,21 @@ function processPointerDown(
...
+    return undefined
+  }
+  const privacyEnabledForActionName =
+    nodePrivacyLevel === NodePrivacyLevel.MASK && configuration.enablePrivacyForActionName
+
+  const clickActionBase = computeClickActionBase(
+    pointerDownEvent,
+    privacyEnabledForActionName,
+    configuration.actionNameAttribute
+  )
```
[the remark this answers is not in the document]
amortemousque: **💬 suggestion:** ‏Maybe even simpler ;) 

```ts

  const nodePrivacyLevel = configuration.enablePrivacyForActionName
    ? getNodePrivacyLevel(pointerDownEvent.target, configuration.defaultPrivacyLevel)
    : NodePrivacyLevel.ALLOW
  
  if (nodePrivacyLevel === NodePrivacyLevel.HIDDEN) {
    return undefined
  }
  
  const clickActionBase = computeClickActionBase(
    pointerDownEvent,
    nodePrivacyLevel,
    configuration.actionNameAttribute
  )
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -201,6 +246,11 @@ function getTextualContent(element: Element | HTMLElement, userProgrammaticAttri
       removeTextFromElements(`[${userProgrammaticAttribute}]`)
     }
 
+    if (privacyEnabledActionName) {
+      // remove the text of elements with privacy override
+      removeTextFromElements(`[${PRIVACY_ATTR_NAME}=${NodePrivacyLevel.MASK}]`)
```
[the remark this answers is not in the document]
amortemousque: Yes in fact there is nothing to do for "mask-user-input" has the name does not take into account input value

packages/rum-core/src/constants.ts
amortemousque, line 1: **💬 suggestion:** ‏The file name "constants.ts" is too generic. All constants in that file are related to privacy. Maybe it could merged into the "privacy.ts"

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -201,6 +236,16 @@ function getTextualContent(element: Element | HTMLElement, userProgrammaticAttri
...
     }
 
+    if (privacyEnabledActionName) {
+      // remove the text of elements with privacy override
+      removeTextFromElements(
+        `[${PRIVACY_ATTR_NAME}=${NodePrivacyLevel.MASK}],
+        [${PRIVACY_ATTR_NAME}=${NodePrivacyLevel.HIDDEN}],
+        [class=${PRIVACY_CLASS_MASK}],
+        [class=${PRIVACY_CLASS_HIDDEN}]`
+      )
```
amortemousque, line 246: **💬 suggestion:** ‏I think we could extract the logic of having to check both the class and the attribute. Something like:
```ts
  removeTextFromElements(
    getPrivacySelector(NodePrivacyLevel.MASK),
    getPrivacySelector(NodePrivacyLevel.HIDDEN)
  )
```

It could even be used in `getNodeSelfPrivacyLevel` 
```ts
// instead of 
const privAttr = node.getAttribute(PRIVACY_ATTR_NAME)
if (privAttr === PRIVACY_ATTR_VALUE_MASK || elementClasslistContains(node, PRIVACY_CLASS_MASK))

// we could use 
if(node.matches(getPrivacySelector(NodePrivacyLevel.MASK))
```
@BenoitZugmeyer do you see any downside of using [matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches), maybe the perf?

packages/rum-core/src/domain/configuration.ts
```diff
@@ -39,6 +39,7 @@ export interface RumInitConfiguration extends InitConfiguration {
   startSessionReplayRecordingManually?: boolean | undefined
 
   // action options
+  enablePrivacyForActionName?: boolean | undefined
```
amortemousque, line 42: **💬 suggestion:** ‏Could be nice to add a comment saying `// TODO next major: remove this option and make privacy for action name the default behaviour`

packages/rum-core/src/constants.ts
cy-moi: Good point. I merged it in `privacy.ts`.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: I think there is a problem to use `.matches()` function. It does not select with desired precedence. For example, `"dd-privacy-mask"` does not select `class="dd-privacy-mask-user-input dd-privacy-mask"`
Screenshot 2024-06-03 at 15 05 43

amortemousque: I think you forgot the "." in the selector `node.matches(".dd-privacy-mack-user-input")`

cy-moi: Ahh sorry I meant `[class="dd-privacy-mask-user-input"]`. I think we could use matches to replace `(privAttr === PRIVACY_ATTR_VALUE_MASK || elementClasslistContains(node, PRIVACY_CLASS_MASK))`, but I don't see how we share the same function with removeTextFromElements.

cy-moi: I see. I updated the conditions with this and removed the unused constants and the polyfill function.

packages/rum-core/src/domain/configuration.ts
```diff
@@ -129,6 +133,10 @@ export function validateAndBuildRumConfiguration(
       defaultPrivacyLevel: objectHasValue(DefaultPrivacyLevel, initConfiguration.defaultPrivacyLevel)
         ? initConfiguration.defaultPrivacyLevel
         : DefaultPrivacyLevel.MASK,
+      enablePrivacyForActionName:
+        (isExperimentalFeatureEnabled(ExperimentalFeature.ENABLE_PRIVACY_FOR_ACTION_NAME) &&
+          initConfiguration.enablePrivacyForActionName) ??
+        false,
```
amortemousque, line 139: **🔨 warning:** ‏ ‏To ensure we sanitise as boolean you should use: 
```suggestion
      enablePrivacyForActionName:
        isExperimentalFeatureEnabled(ExperimentalFeature.ENABLE_PRIVACY_FOR_ACTION_NAME) &&
          !!initConfiguration.enablePrivacyForActionName
```

amortemousque (approved):

BenoitZugmeyer (approved):

PR #2708, opened 2024-04-15
✨ [RUM-3684] Capture scroll record on shadow dom elements

N-Boutaib:
## Motivation
Add support of scrolls inside of shadow dom elements for session replay.

## Changes
Listen to scroll events on shadow roots

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [x] End to end

---


cy-moi (approved):
Thanks for walking me through this!

packages/rum/src/domain/record/trackers/trackScroll.ts
```diff
@@ -18,7 +18,8 @@ export type ScrollCallback = (incrementalSnapshotRecord: BrowserIncrementalSnaps
 export function trackScroll(
   configuration: RumConfiguration,
   scrollCb: ScrollCallback,
-  elementsScrollPositions: ElementsScrollPositions
+  elementsScrollPositions: ElementsScrollPositions,
+  target: Document | ShadowRoot = document
```
bcaudan, line 22: **💬 suggestion:** ‏instead of having document as default, what about having a unit test to ensure that it works with both types of target?

test/e2e/scenario/recorder/shadowDom.scenario.ts
```diff
@@ -244,6 +281,29 @@ describe('recorder with shadow DOM', () => {
...
+    .withRum({})
+    .withSetup(bundleSetup)
+    .withBody(html`
+      ${scrollableDivShadowDom}
+      <my-scrollable-div id="host" />
+    `)
+    .run(async ({ intakeRegistry }) => {
+      const div = await getNodeInsideShadowDom('my-scrollable-div', 'button')
+
+      await div.click()
```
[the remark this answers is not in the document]
N-Boutaib: Not really, I just wanted to have a way to trigger the scroll once everything is setup (calling `scrollTo` on the element from the test isn't working because the returned `wdio.Element` doesn't allow this)

packages/rum/src/domain/record/record.ts
```diff
@@ -51,10 +51,15 @@ export function record(options: RecordOptions): RecordAPI {
...
-    mutationCb: emitAndComputeStats,
-    inputCb: emitAndComputeStats,
-  })
+  const shadowRootsController = initShadowRootsController(
+    configuration,
+    {
+      mutationCb: emitAndComputeStats,
+      inputCb: emitAndComputeStats,
+      scrollCb: emitAndComputeStats,
+    },
```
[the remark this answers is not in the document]
N-Boutaib: Done ✔️

packages/rum/src/domain/record/trackers/trackScroll.ts
N-Boutaib: Agreed on DM that since the logic is the same for tracking scrolls on document, elements and shadowRoot, a unit test is not needed.
The change is covered by an e2e tests.

bcaudan: > Agreed on DM that since the logic is the same for tracking scrolls on document, elements and shadowRoot, a unit test is not needed.

I just did not had time to answer back 🙂
My point was that since we are introducing a new param, it could be valuable to test that this function behave as expected for different values of this param.

By looking more at it, it could be nice to ensure that with trackScroll on different targets, only the expected tracker call its callback.

 > The change is covered by an e2e tests.

it should not prevent us to add additional unit tests.
To be noted: more browsers are covered with unit tests.

test/e2e/scenario/recorder/shadowDom.scenario.ts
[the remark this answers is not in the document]
bcaudan: Making the test itself trigger a scroll would be easier to follow but if it is not possible, adding a comment that clarify the intent could help

packages/rum/src/domain/record/trackers/trackScroll.ts
N-Boutaib: Added an integration test ✔️

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -334,6 +334,22 @@ describe('record', () => {
...
+
+      div.dispatchEvent(createNewEvent('scroll', { target: div, composed: false }))
+
+      recordApi.flushMutations()
+      const innerMutationData = getLastIncrementalSnapshotData<BrowserMutationData>(
+        getEmittedRecords(),
+        IncrementalSource.Scroll
+      )
+
+      expect(innerMutationData).toBeDefined()
```
bcaudan, line 350: **💬 suggestion:** ‏could we ensure that there is only the scroll on the inner element that is generated?

test/e2e/scenario/recorder/shadowDom.scenario.ts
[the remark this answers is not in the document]
N-Boutaib: ✔️

packages/rum/src/domain/record/record.spec.ts
N-Boutaib: ✔️

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -334,20 +335,26 @@ describe('record', () => {
...
         getEmittedRecords(),
         IncrementalSource.Scroll
       )
 
-      expect(innerMutationData).toBeDefined()
+      const fs = findFullSnapshot({ records: getEmittedRecords() })!
+      const scrollableNode = findElement(
+        fs.data.node,
+        (node) => { return node.attributes['unique-selector'] === "enabled" }
+      )!
```
bcaudan, line 355: **💬 suggestion:** ‏would it make sense to also check that there is no other scroll records emitted for other elements?

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -334,20 +335,26 @@ describe('record', () => {
       expect(inputRecords.length).toBe(1)
     })
 
-    it('should record the scroll event inside a shadow root', () => {
-      const div = appendElement('<div></div>', createShadow()) as HTMLDivElement
+    fit('should record the scroll event inside a shadow root', () => {
```
bcaudan, line 338: **❗ warning:** ‏😬

bcaudan (approved):
LGTM!

PR #2725, opened 2024-04-25
✨ [RUM-1310] Support all log statuses

nulrich:
## Motivation

Datadog supports more log status than console API.

## Changes

Adding missing log status from [processors documentation](https://docs.datadoghq.com/logs/log_configuration/processors/?tab=ui#log-status-remapper)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/logs/src/domain/logger/loggerCollection.ts
```diff
@@ -60,6 +65,22 @@ export function isAuthorized(status: StatusType, handlerType: HandlerType, logge
...
+      display('info')
+      break
+    case 'critical':
+    case 'alert':
+    case 'emerg':
+      display('error')
+      break
+    default:
+      display(status)
+  }
```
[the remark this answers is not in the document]
nulrich: I did put `Console.debug` for "OK" because `Console.log` is Info level and is superior to Verbose.

packages/core/src/tools/utils/arrayUtils.ts
```diff
@@ -12,3 +12,7 @@ export function removeItem<T>(array: T[], item: T) {
     array.splice(index, 1)
   }
 }
+
+export function includesItem<T>(array: readonly T[], item: unknown): item is T {
+  return array.includes(item as T)
+}
```
bcaudan, line 18: **❗ warning:** ‏as raised by `unit-bs`:
```
[IE 11.0 (Windows 10)]
	TypeError: Object doesn't support property or method 'includes'
```
We have a polyfill though: https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/utils/polyfills.ts#L3

packages/logs/src/domain/logger.ts
```diff
@@ -111,6 +157,18 @@ export class Logger {
...
+    this.log(message, messageContext, StatusType.critical, error)
+  }
+
+  alert(message: string, messageContext?: object, error?: Error) {
+    this.log(message, messageContext, StatusType.alert, error)
+  }
+
+  emerg(message: string, messageContext?: object, error?: Error) {
+    this.log(message, messageContext, StatusType.emerg, error)
+  }
```
bcaudan, line 170: **💬 suggestion:**  since we have a generic log method and that those severities are not commonly used in the browser, what about not exposing additional methods for now?

packages/logs/src/domain/logger.ts
```diff
@@ -38,6 +45,37 @@ export const HandlerType = {
...
+export const StatusMapping: { [key in StatusType]: { match: (v: string) => boolean; priority: number } } = {
+  [StatusType.emerg]: { priority: 0, match: (v) => matchList(['f', 'emerg'], v, true) },
+  [StatusType.alert]: { priority: 1, match: (v) => matchList(['a'], v, true) },
+  [StatusType.critical]: { priority: 2, match: (v) => matchList(['c'], v, true) },
+  [StatusType.error]: { priority: 3, match: (v) => matchList(['e'], v, true) && !matchList(['emerg'], v, true) },
+  [StatusType.warn]: { priority: 4, match: (v) => matchList(['w'], v, true) },
+  [StatusType.notice]: { priority: 5, match: (v) => matchList(['n'], v, true) },
+  [StatusType.info]: { priority: 6, match: (v) => matchList(['i'], v, true) },
+  [StatusType.debug]: { priority: 7, match: (v) => matchList(['d', 'trace', 'verbose'], v, true) },
+  [StatusType.OK]: { priority: 8, match: (v) => matchList(['o', 's'], v, true) || matchList(['ok', 'success'], v) },
```
bcaudan, line 57: **❓ question:** ‏From the doc, there is no priority 8 though 🤔
Looking at [syslog condition](https://en.wikipedia.org/wiki/Syslog#Severity_level), I'd consider `ok` to be similar to `info`.
Maybe it could be worth digging a bit more into it.

packages/logs/src/domain/logger.ts
```diff
@@ -38,6 +45,37 @@ export const HandlerType = {
...
+  for (const status of STATUSES) {
+    const { match, priority } = StatusMapping[status]
+    if (typeof rawStatus === 'string') {
+      if (match(rawStatus.toLowerCase())) {
+        return status
+      }
+    } else if (rawStatus >= 0 && rawStatus <= 7 && rawStatus === priority) {
+      return status
+    }
+  }
```
BenoitZugmeyer, line 74: **Suggestion:** ‏I think this approach is quite inefficient. Something link this could be more perfomant, and maybe a bit less verbose:

```ts
const statusByPriority = Object.values(StatusMapping) // we should reorder `StatusMapping` to be in sorted by priority
const statusByPrefix = {
  f: StatusType.emerg,
  emerg: StautsType.emerg,
  ...
}

if (typeof rawStatus === 'string') {
  rawStatus = rawStatus.toLowerCase()
  for (const prefixLength of [7, 5, 1]) { // optim: those are the only prefix lengths
    const prefix = rawStatus.slice(0, prefixLength)
    if (Object.prototype.hasOwnProperty.call(statusByPrefix, prefix)) {
      return statusByPrefix[prefix]
    }
  }
} else {
  return statusByPriority[rawStatus]
}
```

packages/core/src/tools/utils/arrayUtils.ts
nulrich: I missed it, thanks. It's done.

packages/logs/src/domain/logger.ts
nulrich: Yes, we agreed on it. It's done.

packages/logs/src/domain/logger/loggerCollection.ts
```diff
@@ -56,10 +50,24 @@ export function isAuthorized(status: StatusType, handlerType: HandlerType, logge
   const loggerHandler = logger.getHandler()
   const sanitizedHandlerType = Array.isArray(loggerHandler) ? loggerHandler : [loggerHandler]
   return (
-    STATUS_PRIORITIES[status] >= STATUS_PRIORITIES[logger.getLevel()] && includes(sanitizedHandlerType, handlerType)
+    StatusMapping[status] &&
```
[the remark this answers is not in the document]
nulrich: Thanks, I also updated the tests.

packages/logs/src/domain/logger.ts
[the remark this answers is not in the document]
nulrich: It was for the sake of completeness of the rules defined in [Remapper](https://docs.datadoghq.com/logs/log_configuration/processors/?tab=ui#log-status-remapper) 😅 I'll remove it.

packages/logs/src/domain/logger.ts
nulrich: Nice! I've push a version where I still use the `StatusMapping` because there's the case of "error" where we need to ensure that the `rawValue` is not a prefix if "emerg" (which starts with an 'e').

packages/logs/src/domain/logger.ts
nulrich: With the team we decided to go for log method and not doing mapping

thomas-lebeau (approved):

packages/logs/src/domain/logger.ts
```diff
@@ -21,10 +21,15 @@ export interface LogsMessage {
 }
 
 export const StatusType = {
+  OK: 'OK',
```
BenoitZugmeyer, line 24: **🥜 nitpick:** ‏I know this is like this in the documentation, but for consistency I would use lowercase for `ok` (key and value) similarly to other variants.

BenoitZugmeyer (approved):

PR #2730, opened 2024-04-26
✨ [RUM-4013] DD_RUM: add handling stack in beforeSend context

thomas-lebeau:
## Motivation

Create a stack trace and expose it in the `beforeSend`'s `domainContext` argument.
It allow customers to initialise the browser SDK with something like:

```js
const SERVICE_REGEX = /(?:browser-sdk-test-playground\/module-federation\/|webpack:\/\/)(?<service>\w*)\//;
  
DD_RUM.init({
    const service =
      context?.handlingStack?.match(SERVICE_REGEX)?.groups?.service;

    if (service) {
      event.service = service; //TODO: this is not yet implemented
    }
    return true;
  },
});
```

## Changes

support the following APIs

- [x] `DD_RUM.addError()`
- [x] `DD_RUM.addAction()`
- [x] `fetch()` resources
- [x] `xhr` resources
- [x] console errors

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/tools/stackTrace/handlingStack.ts
```diff
@@ -9,14 +9,13 @@ import { computeStackTrace } from './computeStackTrace'
  * - Has to be called at the utmost position of the call stack.
  * - No monitored function should encapsulate it, that is why we need to use callMonitored inside it.
  */
-export function createHandlingStack(): string {
+export function createHandlingStack(framesToSkip = 0): string {
```
BenoitZugmeyer, line 12: **💬 suggestion:** ‏It seems like we always skip 2 frames. Maybe we don't need this parameter and we could just use the previous code?

packages/core/src/tools/instrumentMethod.ts
```diff
@@ -116,6 +99,7 @@ function createInstrumentedMethod<TARGET extends { [key: string]: any }, METHOD
         onPostCall: (callback) => {
           postCallCallback = callback
         },
+        handlingStack: createHandlingStack(2),
```
BenoitZugmeyer, line 102: **💭 thought:** ‏Always computing the stack is useful, it makes the API simple to use. But I wonder if there is a performance cost in generating such stack when we don't use it?

nulrich (approved):

PR #2744, opened 2024-05-06
🐛 [RUM-97] Sanitize tags parameter in configuration

RomanGaignault:
## Motivation

Setting a non-string value for the customer provided tags init option causes the SDK to fail silently.

## Changes

Add validation to ensure that the tags parameters are properly defined as strings.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -108,6 +108,15 @@ export interface Configuration extends TransportConfiguration {
   batchMessagesLimit: number
   messageBytesLimit: number
 }
+
+function checkIfString(tag: any, tagName: string) {
+  if (tag && typeof tag !== 'string') {
+    display.error(`${tagName} must be defined as a string`)
+    return false
+  }
+  return true
+}
+
 function isDatadogSite(site: string) {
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -144,6 +144,11 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     return
   }
 
+  if (initConfiguration.version && typeof initConfiguration.version !== 'string') {
+    display.error('Version must be defined as a string')
```
bcaudan, line 148: **💬 suggestion:** ‏IIUC, the error happens when building tags for the intake requests, so should we do the same kind of validation for other customer provided tags? (i.e. service and env)

bcaudan, line 148: **💭 thought:** ‏Since we are completely blind on those kind of execution issues, it could worth planning some reflexions around it to see if we could do something better

RomanGaignault: 👍 After testing, the same issue occurs for service and env tag parameter. I added them to the check.

BenoitZugmeyer (approved):

PR #2745, opened 2024-05-06
🔊[RUM-4360] monitor more API usages

bcaudan:
## Motivation

Provide more visibility on how some SDK APIs are used

## Changes

Monitor usage of:

- startView
- addAction
- addError
- addFeatureFlagEvaluation
- startDurationVital
- setGlobalContext
- setUser

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/boot/rumPublicApi.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.ts
@@ -124,6 +124,7 @@ export function makeRumPublicApi(startRumImpl: StartRum, recorderApi: RecorderAp
               startClocks: options?.startTime ? timeStampToClocks(options.startTime as TimeStamp) : clocksNow(),
               context: sanitize(options?.context) as Context,
             })
+            addTelemetryUsage({ feature: 'start-duration-vital' })
           }
         )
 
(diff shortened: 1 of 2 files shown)


thomas-lebeau (approved):

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -296,6 +308,7 @@ export function makeRumPublicApi(startRumImpl: StartRum, recorderApi: RecorderAp
      */
     addFeatureFlagEvaluation: monitor((key: string, value: any) => {
       strategy.addFeatureFlagEvaluation(sanitize(key)!, sanitize(value))
+      addTelemetryUsage({ feature: 'add-feature-flag-evaluation' })
```
BenoitZugmeyer, line 311: **💭 thought:** ‏almost all usages of `addTelemetryUsage` is only providing a feature name. It could be shorter to change the signature to something like:
```ts
addTelemetryUsage("feature", { optional_context: true })
```

BenoitZugmeyer (approved):

packages/rum-core/src/boot/rumPublicApi.ts
bcaudan: when introducing the API, I spent some time trying something like this but did not managed to have the typings working.
Don't hesitate to have a shot at it, happy to tweak that in a following PR.

PR #2746, opened 2024-05-07
🔊 Deduplicate telemetry events

bcaudan:
## Motivation

Avoid to loose some telemetry when a single event is spammed, ex:
Screenshot 2024-05-07 at 16 51 42

## Changes

Discard already seen telemetry events

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/telemetry/telemetry.ts
+++ b/packages/core/src/domain/telemetry/telemetry.ts
@@ -64,6 +64,7 @@ let onRawTelemetryEventCollected: ((event: RawTelemetryEvent) => void) | undefin
 export function startTelemetry(telemetryService: TelemetryService, configuration: Configuration): Telemetry {
   let contextProvider: () => Context
   const observable = new Observable<TelemetryEvent & Context>()
+  const alreadySeenEvents = new Set<string>()
 
   const telemetryEnabled =
     !includes(TELEMETRY_EXCLUDED_SITES, configuration.site) && performDraw(configuration.telemetrySampleRate)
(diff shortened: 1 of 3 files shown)


thomas-lebeau:
**❓ question:**‏ Are we not interested in how much a customer is using an API?

packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -76,10 +77,17 @@ export function startTelemetry(telemetryService: TelemetryService, configuration
 
   const runtimeEnvInfo = getRuntimeEnvInfo()
   onRawTelemetryEventCollected = (rawEvent: RawTelemetryEvent) => {
-    if (telemetryEnabledPerType[rawEvent.type!]) {
+    const stringifiedEvent = jsonStringify(rawEvent)!
```
BenoitZugmeyer, line 80: **💭 thought:** ‏I don't love the extra serialization and memory usage (as we store a serialized version of every telemetry event forever), but I don't have a simpler/better idea.

BenoitZugmeyer (approved):

packages/core/src/domain/telemetry/telemetry.ts
bcaudan: > we store a serialized version of every telemetry event forever

only the first `maxEventsPerPage` events though, so 15 for now.

bcaudan: We could avoid to serialize if we know that we won't send the event.
Do you think it worth adding this additional code path?

bcaudan:
> **❓ question:** Are we not interested in how much a customer is using an API?

It could be interesting but it is not in the scope of the current initiative (cf RFC).
The `maxEventsPerPage` threshold would not allow to have something reliable, so it would require more thoughts if we want to do that at some point.

packages/core/src/domain/telemetry/telemetry.ts
BenoitZugmeyer: Hum, maybe not.

While having another look, I thought of a small simplification: we could replace `telemetryConfiguration.sentEventCount` with `alreadySeenEvents.size` (one less state to maintain!)

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):
Sweet!

amortemousque (approved):

PR #2747, opened 2024-05-08
🐛 [RUM-4434] fix timing matching for the same resource requested twice at the same time

thomas-lebeau:
## Motivation

Fixes an issue where resource timing where not matched to a resource when the same resource is requested twice at the same time.

e.g.
```js
Promise.all([
    fetch("http://localhost:8080/"),
    fetch("http://localhost:8080/")
])
```

The issue is that when the second request resolves, we have 2 resource timings matching the request hence we were not able to decide which one is the correct one.

## Changes

When a match is found between a resource timing and a request, we store the timing in a `WeakMap`, that way we can filter it out of the available timings when the 2nd resource resolve.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
+++ b/packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
@@ -20,7 +20,7 @@ describe('matchRequestTiming', () => {
       pending('no full rum support')
     }
     entries = []
-    spyOn(performance, 'getEntriesByName').and.returnValues(entries as unknown as PerformanceResourceTiming[])
+    spyOn(performance, 'getEntriesByName').and.returnValue(entries)
   })
 
   it('should match single timing nested in the request ', () => {
@@ -57,6 +57,28 @@ describe('matchRequestTiming', () => {
     const matchingTiming = matchRequestTiming(FAKE_REQUEST as RequestCompleteEvent)
 
     expect(matchingTiming).toEqual(undefined)
+  })
+
+  it('should match all timings to the same request done at the same time', () => {
+    const entry1 = createPerformanceEntry(RumPerformanceEntryType.RESOURCE, {
+      startTime: 200 as RelativeTime,
+      duration: 300 as Duration,
+    })
+    entries.push(entry1)
+
+    const matchingTiming1 = matchRequestTiming(FAKE_REQUEST as RequestCompleteEvent)
+
+    expect(matchingTiming1).toEqual(entry1.toJSON() as RumPerformanceResourceTiming)
+
+    const entry2 = createPerformanceEntry(RumPerformanceEntryType.RESOURCE, {
+      startTime: 99 as RelativeTime,
+      duration: 502 as Duration,
+    })
+    entries.push(entry2)
+
+    const matchingTiming2 = matchRequestTiming(FAKE_REQUEST as RequestCompleteEvent)
+
+    expect(matchingTiming2).toEqual(entry2.toJSON() as RumPerformanceResourceTiming)
   })
 
   it('should not match two not following timings nested in the request ', () => {
--- a/packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
+++ b/packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
@@ -59,7 +59,7 @@ describe('matchRequestTiming', () => {
     expect(matchingTiming).toEqual(undefined)
   })
 
-  it('should match all timings to the same request done at the same time', () => {
+  it('should discard already matched timings when multiple identical requests are done conurently', () => {
     const entry1 = createPerformanceEntry(RumPerformanceEntryType.RESOURCE, {
       startTime: 200 as RelativeTime,
       duration: 300 as Duration,
--- a/packages/rum-core/src/domain/resource/matchRequestTiming.ts
+++ b/packages/rum-core/src/domain/resource/matchRequestTiming.ts
@@ -2,6 +2,7 @@ import type { Duration, RelativeTime } from '@datadog/browser-core'
 import { addDuration } from '@datadog/browser-core'
 import type { RumPerformanceResourceTiming } from '../../browser/performanceCollection'
 import type { RequestCompleteEvent } from '../requestCollection'
+import { WeakSet } from '../../browser/polyfills'
 import { isValidEntry } from './resourceUtils'
 
 interface Timing {
@@ -9,9 +10,7 @@ interface Timing {
   duration: Duration
 }
 
-// we use a WeakMap because WeakSet is not supported in ie11
-const PLACEHOLDER = 1
-const matchedResourceTimingEntries = new WeakMap<PerformanceEntry, typeof PLACEHOLDER>()
+const alreadyMatchedEntries = new WeakSet<PerformanceEntry>()
 
 /**
  * Look for corresponding timing in resource timing buffer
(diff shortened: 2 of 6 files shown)


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -137,9 +135,9 @@ export function computePerformanceResourceDetails(
   return details
 }
 
-export function toValidEntry(entry: RumPerformanceResourceTiming) {
+export function isValidEntry(entry: RumPerformanceResourceTiming) {
```
thomas-lebeau, line 138: I renamed this as I believe it's more correct because 
- it's not transforming the entry. 
- it is only used in a `if` condition and in a `Array.filter()` method

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -163,8 +161,10 @@ export function toValidEntry(entry: RumPerformanceResourceTiming) {
     : true
 
   if (areCommonTimingsInOrder && areRedirectionTimingsInOrder) {
-    return entry
+    return true
   }
+
+  return false
```
bcaudan, line 167: **🥜 nitpick:** ‏this statement could be simplified
```
if (x) {
  return true     =>    return x
}
return false
```

packages/rum-core/src/domain/resource/matchRequestTiming.spec.ts
```diff
@@ -59,6 +59,28 @@ describe('matchRequestTiming', () => {
     expect(matchingTiming).toEqual(undefined)
   })
 
+  it('should match all timings to the same request done at the same time', () => {
```
bcaudan, line 62: **❓ question:** ‏ ‏I am a bit confused by the test description, FMU the implementation avoid to reuse an entry that has already been used for a previous matching.
Do we expect more than that?

packages/rum-core/src/domain/resource/matchRequestTiming.ts
```diff
@@ -2,13 +2,17 @@ import type { Duration, RelativeTime } from '@datadog/browser-core'
...
-import { toValidEntry } from './resourceUtils'
+import { isValidEntry } from './resourceUtils'
 
 interface Timing {
   startTime: RelativeTime
   duration: Duration
 }
 
+// we use a WeakMap because WeakSet is not supported in ie11
+const PLACEHOLDER = 1
```
bcaudan, line 13: **🥜 nitpick:** ‏we could extract a weakset polyfill doing the weakmap + placeholder trick to remove this concern from this file

packages/rum-core/src/domain/resource/matchRequestTiming.ts
```diff
@@ -2,13 +2,17 @@ import type { Duration, RelativeTime } from '@datadog/browser-core'
...
+const matchedResourceTimingEntries = new WeakMap<PerformanceEntry, typeof PLACEHOLDER>()
```
bcaudan, line 14: **💬 suggestion:** ‏what about renaming it to `alreadyMatchedEntries`?
- `resourceTiming` seems superfluous as it is not used in `sameNameEntries`
- adding the `already` prefix could make the intent clearer

packages/rum-core/src/domain/resource/matchRequestTiming.ts
```diff
@@ -32,18 +36,24 @@ export function matchRequestTiming(request: RequestCompleteEvent) {
   }
 
   const candidates = sameNameEntries
-    .map((entry) => entry.toJSON() as RumPerformanceResourceTiming)
-    .filter(toValidEntry)
+    .filter((entry) => !matchedResourceTimingEntries.has(entry))
```
bcaudan, line 39: **💬 suggestion:** ‏what about updating the function comment explaining the strategy with this additional behaviour?

packages/rum-core/src/domain/resource/matchRequestTiming.ts
```diff
@@ -32,18 +36,24 @@ export function matchRequestTiming(request: RequestCompleteEvent) {
...
-        entry,
+        entry.serialized,
         request.startClocks.relative,
         endTime({ startTime: request.startClocks.relative, duration: request.duration })
       )
     )
 
   if (candidates.length === 1) {
-    return candidates[0]
+    matchedResourceTimingEntries.set(candidates[0].original, PLACEHOLDER)
```
bcaudan, line 54: **❓ question:** any evidence that when two entries are matching the same timing, one entry has already matched another timing?

thomas-lebeau: In the case of fetch, having two timing matching the same resource without being already matched, means that both timings are available before the `then` of the first fetch is called.

This is not the case when doing `Promise.all()` but I don't exclude the possibility of it happening. In that case no timing will be matched.

bcaudan: Would it be worth doing more experiments on browser behaviours around that? or measure the effectiveness of this change somehow?

packages/rum-core/src/domain/resource/resourceUtils.ts
thomas-lebeau: I'm not sure I understand, do you mean:
```diff
  if (areCommonTimingsInOrder && areRedirectionTimingsInOrder) {
-    return true
+    return areCommonTimingsInOrder && areRedirectionTimingsInOrder
  }

  return false
```

I'm not convinced it's a simplification, but I can make the change anyway. Or maybe I didn't understood your suggestion.

bcaudan: I meant:
```diff
-  if (areCommonTimingsInOrder && areRedirectionTimingsInOrder) {
-    return true  
-  }
-
-  return false
+ return areCommonTimingsInOrder && areRedirectionTimingsInOrder
```

packages/rum-core/src/domain/resource/matchRequestTiming.ts
thomas-lebeau: I'm adding a e2e test, however I've noticed that safari doesn't do like everyone and both `PerformanceResourceTiming` might be available at the time the first resource is matching it's timings.

This means that this PR reduce the amount of occurrence of missing resource timing on concurrent event (mainly chrome/FF) but does not prevent it totally (safari)

anyone has a better idea?

packages/rum-core/src/domain/resource/resourceUtils.ts
thomas-lebeau: 🤦 of course!

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -217,6 +217,27 @@ describe('rum resources', () => {
       expect(resourceEvent.resource.redirect!.duration).toBeGreaterThan(0)
     })
 
+  createTest('track concurent fetch to same resource')
```
amortemousque, line 220: **🥜 nitpick:** ‏
```suggestion
  createTest('track concurrent fetch to same resource')```

File: packages/rum-core/src/domain/resource/matchRequestTiming.ts [hunk 9]
```diff
@@ -2,13 +2,16 @@ import type { Duration, RelativeTime } from '@datadog/browser-core'
...
-import { toValidEntry } from './resourceUtils'
+import { WeakSet } from '../../browser/polyfills'
+import { isValidEntry } from './resourceUtils'
 
 interface Timing {
   startTime: RelativeTime
   duration: Duration
 }
 
+const alreadyMatchedEntries = new WeakSet<PerformanceEntry>()
```
amortemousque, line 13: **💬 suggestion:** Using a WeakSet for entries looks weird to me, as we don't know for how long the ref will exists since garbage collection can happen at any point in time. To have a more predictable behaviour we could keep the last X entries. Wdyt?‏

thomas-lebeau: `performance.getEntriesByName()` is already pulled out from [a buffer](https://developer.mozilla.org/en-US/docs/Web/API/Performance/setResourceTimingBufferSize), so they won't be garbage collected until the buffer is emptied.

If the buffer is emptied before `matchRequestTiming()` is called, then `performance.getEntriesByName()` will not return the entries that we wish to filter out in the first place.

If the garbage collection happens between `performance.getEntriesByName()` and `alreadyMatchedEntries.has()` then the entries we're interested in will be held strongly by `sameNameEntries` and not garbage collected.

In conclusion, I think keep the last X entries will not add more predictability while adding the overhead of managing ourselves a separate buffer.

Does this answer your concern?

packages/rum-core/src/domain/resource/matchRequestTiming.ts
thomas-lebeau: as discussed during parking lot, this PR won't solve everything but is improving in some cases. So I've skipped the e2e on non chromium browsers as it's not reliable there.

packages/rum-core/src/domain/resource/matchRequestTiming.ts
amortemousque: Very interesting! It's very smart to rely on the refs held by the browser :)

amortemousque (approved):

nulrich (approved):

PR #2749, opened 2024-05-08
🐛 [RUM-4436] fix detached node memory leak on CLS

thomas-lebeau:
## Motivation

trackCumulativeLayoutShift keep track of the LCP target Element using a sliding window of layout shift events but this window is cleared [5 minutes](https://github.com/DataDog/browser-sdk/blob/b0d638c02da5b9316cb8c8cf4aa3b5e8297f683b/packages/rum-core/src/domain/view/trackViews.ts#L74) after the view ends.

Within these 5 minutes, many new view events might be started meaning an increase in memory consumption du to the reference to the LCP target Element kept and preventing garbage collection of these elements (and potentially large amount of descendent)

## Changes

Hold onto the LCP target element using a weak reference (`WeakRef`) to not prevent it to be garbage-collected.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
@@ -1,5 +1,5 @@
-import { round, find, ONE_SECOND, noop } from '@datadog/browser-core'
-import type { RelativeTime } from '@datadog/browser-core'
+import { round, find, ONE_SECOND, noop, setTimeout, clearTimeout } from '@datadog/browser-core'
+import type { RelativeTime, TimeoutId } from '@datadog/browser-core'
 import { isElementNode } from '../../../browser/htmlDomUtils'
 import type { LifeCycle } from '../../lifeCycle'
 import { LifeCycleEventType } from '../../lifeCycle'
@@ -58,19 +58,19 @@ export function trackCumulativeLayoutShift(
           maxClsValue = window.value()
           const cls = round(maxClsValue, 4)
           const clsTarget = window.largestLayoutShiftTarget()
-          let cslTargetSelector
+          let clsTargetSelector
 
           if (
             clsTarget &&
             // Check if the CLS target have been removed from the DOM between the time we collect the target reference and when we compute the selector
             clsTarget.isConnected
           ) {
-            cslTargetSelector = getSelectorFromElement(clsTarget, configuration.actionNameAttribute)
+            clsTargetSelector = getSelectorFromElement(clsTarget, configuration.actionNameAttribute)
           }
 
           callback({
             value: cls,
-            targetSelector: cslTargetSelector,
+            targetSelector: clsTargetSelector,
           })
         }
       }
@@ -82,7 +82,10 @@ export function trackCumulativeLayoutShift(
   }
 }
 
-function slidingSessionWindow() {
+const MAX_WINDOW_LENGTH = 5 * ONE_SECOND
+const MAX_UPDATE_GAP = ONE_SECOND
+
+export function slidingSessionWindow() {
   let value = 0
   let startTime: RelativeTime
   let endTime: RelativeTime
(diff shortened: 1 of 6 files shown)


cy-moi (approved):

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -82,26 +82,39 @@ export function trackCumulativeLayoutShift(
...
+    value = entry.value
+    largestLayoutShift = 0
+    largestLayoutShiftTarget = undefined
+    largestLayoutShiftTime = entry.startTime
+  }
+
   return {
     update: (entry: RumLayoutShiftTiming) => {
+      clearTimeout(timeoutId)
+      timeoutId = setTimeout(() => resetWindow(entry), MAX_WINDOW_LENGTH)
```
BenoitZugmeyer, line 109: **💬 suggestion:** ‏Since a window cannot be longer than `MAX_WINDOW_LENGTH`, I think it doesn't make sense to set the timeout like that:
* keep the timeout here (reseting at every new entry) and use a timeout duration of `MAX_UPDATE_GAP`
* or move the timeout when creating the new window and keep using `MAX_WINDOW_LENGTH` (I think that would me more efficient)

**💭 thought:** ‏I don't think it makes sense to call `resetWindow` with the `entry` here. `resetWindow` should reset the state without using any entry, since it is to clear up the current window.

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -82,26 +82,39 @@ export function trackCumulativeLayoutShift(
   }
 }
 
-function slidingSessionWindow() {
+const MAX_WINDOW_LENGTH = 5 * ONE_SECOND
```
BenoitZugmeyer, line 85: **🥜 nitpick:** ‏"length" is more usually used for sizes, not time
```suggestion
const MAX_WINDOW_DURATION = 5 * ONE_SECOND
```

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
BenoitZugmeyer: **💭 thought:** ‏Since the PerformanceObserver is reporting timings asynchronously, `entry.startTime` might be in the past. Resetting the window after `MAX_WINDOW_LENGTH` might be too early and we might miss some entries that would normally it in the window. Maybe we can wait a little be longer just to be safe

BenoitZugmeyer (commented):
Thought: maybe a simpler solution could be to avoid storing the target altogether.
In `trackCumulativeLayoutShift`, we could have something like this:

```
const isLargestLayoutShift = window.update(entry)
// ... then:
if (isLargestLayoutShift) {
  const clsTarget = find(
            entry.sources,
            (s): s is { node: HTMLElement } => !!s.node && isElementNode(s.node)
          )?.node;

          if (
            clsTarget &&
            // Check if the CLS target have been removed from the DOM between the time we collect the target reference and when we compute the selector
            clsTarget.isConnected
          ) {
            cslTargetSelector = getSelectorFromElement(clsTarget, configuration.actionNameAttribute)
            clsTargetSelector = getSelectorFromElement(clsTarget, configuration.actionNameAttribute)
          }
  }

}
```

and remove `largestLayoutShiftTarget` from the window target. Also we could do the same for `largestLayoutShiftTime`.

amortemousque:
**🔨 warning:** ‏With this implementation we are computing the CSS selector lot more often:
- For each window even if it's not a maxClsValue
- For each LAYOUT_SHIFT if the value of the entry is superior to the precedent
Since layout shift can be frequently triggered by some UI elements, it can have an impact on the processing.

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -82,47 +79,52 @@ export function trackCumulativeLayoutShift(
   }
 }
 
+function getTargetSelctorFromSource(sources?: Array<{ node?: Node }>) {
```
thomas-lebeau, line 82:
```suggestion
function getTargetFromSource(sources?: Array<{ node?: Node }>) {
```

amortemousque (approved):

PR #2756, opened 2024-05-14
⚡ [RUM-116] On view change, take the full snapshot asynchronously

RomanGaignault:
## Motivation

In the Datadog App, we are starting a new view when the URL changes (via the onChange react-router callback). But at this time, the old page component is still displayed, and we take a full snapshot of the ending view instead of the new view.

## Changes

A solution would be to take the fullsnapshot asynchronously. This would let a bit of time to the application UI to be updated. While waiting for the FS to be taken, we could ignore any emitted record, since we would know that a FS is coming to capture the whole state:

startView() is called

ignore any emitted record

[async] wait a bit

if stopSessionReplayRecording() was not called in the meantime

full snapshot is taken

start collecting emitted record again

For the asynchronous delay, we could even use requestIdleCallback to limit our impact on the application perceptible performance.

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum/src/domain/record/startFullSnapshots.spec.ts
```diff
@@ -29,26 +34,68 @@ describe('startFullSnapshots', () => {
...
+      startClocks: viewStartClock,
+    } as Partial<ViewCreatedEvent> as any)
+
+    lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
+      startClocks: viewStartClock,
+    } as Partial<ViewCreatedEvent> as any)
+
+    triggerIdleCallbacks()
+    expect(cancelIdleCallbackSpy).toHaveBeenCalledTimes(1)
+  })
```
BenoitZugmeyer, line 71: **💬 suggestion:** ‏Ideally, tests should be assertions on the expected behavior of the subject, not how it's implemented. Instead of testing whether the idle callback has been canceled (which is an implementation detail), test that we end up having a single FS

```suggestion
  it('cancels the full snapshot if another view is created before it can it happens', () => {
    const { triggerIdleCallbacks, cancelIdleCallbackSpy } = mockRequestIdleCallback()

    lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
      startClocks: viewStartClock,
    } as Partial<ViewCreatedEvent> as any)

    lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
      startClocks: viewStartClock,
    } as Partial<ViewCreatedEvent> as any)

    triggerIdleCallbacks()
    expect(fullSnapshotCallback).toHaveBeenCalledTimes(2)
  })
```

packages/rum/src/domain/record/startFullSnapshots.spec.ts
```diff
@@ -29,26 +34,68 @@ describe('startFullSnapshots', () => {
...
+
+    const { triggerIdleCallbacks } = mockRequestIdleCallback()
+
+    lifeCycle.notify(LifeCycleEventType.VIEW_CREATED, {
+      startClocks: viewStartClock,
+    } as Partial<ViewCreatedEvent> as any)
+    triggerIdleCallbacks()
+
+    expect(fullSnapshotCallback).toHaveBeenCalledTimes(2)
+  })
```
BenoitZugmeyer, line 100: **💬 suggestion:** move ‏this kind of test in `requestIdleCallback.spec.ts`

packages/rum/src/domain/record/startFullSnapshots.ts
```diff
@@ -64,18 +65,33 @@ export function startFullSnapshots(
...
-    )
+    } else {
+      fullSnapshotCallback(
+        takeFullSnapshot(view.startClocks.timeStamp, {
+          shadowRootsController,
+          status: SerializationContextStatus.SUBSEQUENT_FULL_SNAPSHOT,
+          elementsScrollPositions,
+        })
+      )
+    }
```
BenoitZugmeyer, line 94: **🔨 warning:** ‏`flushMutations()` should be done directly before `takeFullSnapshot()`

**💬 suggestion:** ‏factorize duplicated code

```ts
    function takeSubsequentFullSnapshot() {
      flushMutations()
      fullSnapshotCallback(
        takeFullSnapshot(view.startClocks.timeStamp, {
          shadowRootsController,
          elementsScrollPositions,
        })
      )
    }

    if (isExperimentalFeatureEnabled(ExperimentalFeature.ASYNC_FULL_SNAPSHOT)) {
      if (cancelIdleCallback) {
        cancelIdleCallback()
      }

      cancelIdleCallback = requestIdleCallback(takeSubsequentFullSnapshot)
    } else {
      takeSubsequentFullSnapshot()
    }
```

BenoitZugmeyer (commented):
This looks good, but it has one drawback: now, changes (like DOM mutations, mouse movement...) can happen between the view start and the full snapshot.

This is not expected: full snapshot records (Meta/Focus/FullSnapshot) should be the very first records of the view. 

To work around this issue, `startFullSnapshot` could return a new function like `isFullSnapshotPending(): bool`. Then, in `record.ts` `emitAndComputeStats`, we can check whether `isFullSnapshotPending()` is `false` before emiting the record.

Let me know if you have any question.

packages/rum/src/domain/record/startFullSnapshots.spec.ts
```diff
@@ -11,13 +13,16 @@ describe('startFullSnapshots', () => {
...
+  const originalRequestIdleCallback = window.requestIdleCallback
+  const originalCancelIdleCallback = window.cancelIdleCallback
 
   beforeEach(() => {
     if (isIE()) {
       pending('IE not supported')
     }
 
     lifeCycle = new LifeCycle()
+    addExperimentalFeatures([ExperimentalFeature.ASYNC_FULL_SNAPSHOT])
```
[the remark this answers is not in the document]
RomanGaignault: 👍

packages/rum/src/domain/record/startFullSnapshots.ts
```diff
@@ -65,13 +66,17 @@ export function startFullSnapshots(
     }
     return records
   }
-  fullSnapshotCallback(takeFullSnapshot())
+
+  fullSnapshotPendingCallback()
```
BenoitZugmeyer, line 70: **💬 suggestion:** ‏the snapshot is not pending here, you could remove this call.

packages/rum/src/domain/record/startFullSnapshots.ts
```diff
@@ -65,13 +66,17 @@ export function startFullSnapshots(
...
+  fullSnapshotReadyCallback(takeFullSnapshot())
+
   let cancelIdleCallback: (() => void) | undefined
   const { unsubscribe } = lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (view) => {
     flushMutations()
     function takeSubsequentFullSnapshot() {
       flushMutations()
-      fullSnapshotCallback(
+      fullSnapshotPendingCallback()
+      fullSnapshotReadyCallback(
```
BenoitZugmeyer, line 79: **💬 suggestion:** ‏there is no point in notifying that the snapshot is pending just before it's ready. Move `fullSnapshotPendingCallback()` before `requestIdleCallback(...)`

packages/rum/src/domain/record/record.ts
```diff
@@ -42,11 +42,15 @@ export function record(options: RecordOptions): RecordAPI {
...
   }
 
+  let isFullSnapshotPending = false
+
   const emitAndComputeStats = (record: BrowserRecord) => {
-    emit(record)
-    sendToExtension('record', { record })
-    const view = options.viewContexts.findView()!
-    replayStats.addRecord(view.id)
+    if (!isFullSnapshotPending) {
```
BenoitZugmeyer, line 48: **💬 suggestion:** ‏write one or multiple unit tests to make sure this works as expected. Based on [my comment](https://github.com/DataDog/browser-sdk/pull/2756#pullrequestreview-2122327947), this could look like:

```ts
it("ignores any record while a full snapshot is pending", () => {
  ...
```

packages/rum/src/browser/requestIdleCallBack.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import { LifeCycle, LifeCycleEventType } from '@datadog/browser-rum-core'
+import type { TimeStamp } from '@datadog/browser-core'
+import { noop } from '@datadog/browser-core'
+import { mockRequestIdleCallback } from '@datadog/browser-core/test'
+import type { BrowserRecord } from '../types'
+import { addExperimentalFeatures, ExperimentalFeature } from '../../../core/src/tools/experimentalFeatures'
+import { createElementsScrollPositions, startFullSnapshots } from '../domain/record'
+import type { ShadowRootsController } from '../domain/record'
+
+describe('startFullSnapshots', () => {
```
BenoitZugmeyer, line 11: **💬 suggestion:** ‏this spec file should focus on `requestIdleCallback.ts`, and should not be related to full snapshots in any way.

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -422,6 +422,23 @@ describe('record', () => {
...
+        emit: emitSpy,
+        configuration: { defaultPrivacyLevel: DefaultPrivacyLevel.ALLOW } as RumConfiguration,
+        lifeCycle,
+        viewContexts: {
+          findView: () => ({ id: FAKE_VIEW_ID, startClocks: {} }),
+        } as any,
+        defineSnapshotPending: true,
+      })
+      const records = getEmittedRecords()
+      expect(records.length).toBe(0)
```
BenoitZugmeyer, line 438: **💬 suggestion:** ‏instead of having a `defineSnapshotPending` that is only used for test, you could induce the pending state by mocking the idle callback then triggering a view change, something like:


```ts
mockRequestIdleCallback()
startRecording()
newView()

// this normally creates a "focus" record
window.dispatchEvent(createNewEvent('focus'))

expect(emitSpy).not.toHaveBeenCalled()
```

packages/rum/src/domain/record/startFullSnapshots.spec.ts
```diff
@@ -2,51 +2,90 @@ import type { RumConfiguration, ViewCreatedEvent } from '@datadog/browser-rum-co
...
-      fullSnapshotCallback
+      fullSnapshotPendingCallback,
+      fullSnapshotReadyCallback
     )
   })
 
+  afterEach(() => {
+    window.requestIdleCallback = originalRequestIdleCallback
+    window.cancelIdleCallback = originalCancelIdleCallback
+  })
```
BenoitZugmeyer, line 44: **💬 suggestion:** ‏This shouldn't be needed?

packages/rum/src/browser/requestIdleCallBack.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  let callback: jasmine.Spy
+
+  beforeEach(() => {
+    requestAnimationFrameSpy = spyOn(window, 'requestAnimationFrame').and.callFake((cb) => {
+      cb(0)
+      return 123
+    })
+    cancelAnimationFrameSpy = spyOn(window, 'cancelAnimationFrame')
+    callback = jasmine.createSpy('callback')
+  })
```
BenoitZugmeyer, line 15: **💬 suggestion:** ‏restore `requestIdleCallback` and `cancelIdleCallback` in `afterEach` (like you did before)

packages/rum/src/domain/record/startFullSnapshots.spec.ts
BenoitZugmeyer: TODO: remove this afterEach because we don't change requestIdleCallback in this spec file

packages/rum/src/domain/record/record.spec.ts
```diff
@@ -422,6 +422,18 @@ describe('record', () => {
...
 
+  describe('it should not record when full snapshot is pending', () => {
+    it('ignores any record while a full snapshot is pending', () => {
+      mockRequestIdleCallback()
+      startRecording()
+      newView()
+
+      window.dispatchEvent(createNewEvent('focus'))
+
+      expect(emitSpy).toHaveBeenCalled()
```
BenoitZugmeyer, line 433: **💬 suggestion:** ‏it could be more explicit like this:
```ts
      mockExperimentalFeatures([ExperimentalFeature.ASYNC_FULL_SNAPSHOT])
      mockRequestIdleCallback()
      startRecording()
      newView()

      emitSpy.calls.reset()

      window.dispatchEvent(createNewEvent('focus'))

      expect(getEmittedRecords().find((record) => record.type === RecordType.Focus)).toBeUndefined()
```

PR #2764, opened 2024-05-17
🐛 [RUM-1666] Don't set negative action loading time 

RomanGaignault:
## Motivation

We noticed that we have actions with negative actions negative loading time. This is probably due to the fact that Date.now() is not monotonic and sometimes goes “back in time” 

## Changes

If that's the case, we could just not set the loading time of the action event.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/action/actionCollection.spec.ts
+++ b/packages/rum-core/src/domain/action/actionCollection.spec.ts
@@ -118,7 +118,7 @@ describe('actionCollection', () => {
     })
     expect(rawRumEvents[0].domainContext).toEqual({})
   })
-  it('should discard negative action loading time', () => {
+  it('should not set the loading time field of the action', () => {
     const { lifeCycle, rawRumEvents } = setupBuilder.build()
     const event = createNewEvent('pointerup', { target: document.createElement('button') })
     lifeCycle.notify(LifeCycleEventType.AUTO_ACTION_COMPLETED, {
--- a/packages/rum-core/src/domain/action/actionCollection.spec.ts
+++ b/packages/rum-core/src/domain/action/actionCollection.spec.ts
@@ -1,5 +1,6 @@
 import type { Duration, RelativeTime, ServerDuration, TimeStamp } from '@datadog/browser-core'
 import { createNewEvent } from '@datadog/browser-core/test'
+import type { RawRumActionEvent } from '@datadog/browser-rum-core'
 import type { TestSetupBuilder } from '../../../test'
 import { setup } from '../../../test'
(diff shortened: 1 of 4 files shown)


N-Boutaib (approved):

packages/rum-core/src/domain/action/actionCollection.spec.ts
```diff
@@ -117,4 +118,24 @@ describe('actionCollection', () => {
     })
     expect(rawRumEvents[0].domainContext).toEqual({})
   })
+  it('should discard negative action loading time', () => {
```
cy-moi, line 121: 🥜 nit: shall we specify here we are discarding the field (do not set it) rather than the action when it's negative?

cy-moi (approved):

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -62,7 +62,7 @@ function processAction(
     ? {
         action: {
           id: action.id,
-          loading_time: toServerDuration(action.duration),
+          loading_time: discardNegativeLoadingTime(toServerDuration(action.duration)),
```
amortemousque, line 65: **💬 suggestion:** ‏You can probably reuse `discardNegativeDuration` from `viewCollection` https://github.com/DataDog/browser-sdk/blob/146b4bf68e61b8ce24e563b85bc930432eacf69f/packages/rum-core/src/domain/view/viewCollection.ts#L135

packages/rum-core/src/domain/action/actionCollection.spec.ts
RomanGaignault: yes great idea 👍

BenoitZugmeyer (approved):

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -132,6 +132,6 @@ function processViewUpdate(
   }
 }
 
-function discardNegativeDuration(duration: ServerDuration | undefined): ServerDuration | undefined {
+export function discardNegativeDuration(duration: ServerDuration | undefined): ServerDuration | undefined {
```
amortemousque, line 135: **🥜 nitpick:** ‏Since it is no longer specific to views, you could move the function up in a new `browser/timeUtils.ts` for instance.

amortemousque (approved):

packages/rum-core/src/domain/view/viewCollection.ts
amortemousque: Or maybe in `domain/discardNegativeDuration.ts` :)

PR #2771, opened 2024-05-22
🔮 [HADXVI-53] Browser SDK extension search bar improvement

cy-moi:
## Motivation
The search bar living in the event tab has only minimal support now. We want to improve it for a better user experience. This PR is an experiment on one of the approaches.

## Changes
|  |  |
|--------|--------|
| Search in descriptions | Copy search query from cell value directly |
| Screenshot 2024-05-22 at 11 56 32 | Screenshot 2024-05-22 at 13 57 10 |
| Search with * | Escape whitespace with backlash |
| Screenshot 2024-05-22 at 14 50 37 | Screenshot 2024-05-22 at 14 56 48  |

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/panel/components/json.tsx
```diff
@@ -275,6 +279,15 @@ function JsonText({
   )
 }
 
+function getSearchQuery(descriptor: JsonValueDescriptor) {
+  const { columnPath } = useContext(JsonContext)!
```
thomas-lebeau, line 283: **🔨 warning:** ‏ Does this works? I don't think you can use hooks in a functions like this.

Instead, you could either pass `columnPath` as a function argument or transform this function into a custom hook

developer-extension/src/panel/components/json.tsx
```diff
@@ -314,7 +327,7 @@ function CopyMenuItem({ value, children }: { value: unknown; children: ReactNode
   return (
     <Menu.Item
       onClick={() => {
-        copy(JSON.stringify(value, null, 2))
+        copy(JSON.parse(JSON.stringify(value, null, 2)))
```
thomas-lebeau, line 330: **❓ question:** ‏ why is it parsed and then stringified?

developer-extension/src/panel/hooks/useEvents/eventFilters.ts
```diff
@@ -75,18 +75,42 @@ function filterOutdatedVersions(events: SdkEvent[]): SdkEvent[] {
 
 function parseQuery(query: string) {
   const queryParts = query
-    .split(' ')
+    .split(/(?<!\\)\s/gm) // Hack it to escape whitespace with backslashes
```
thomas-lebeau, line 78: **👏 praise:** ‏ nice!

developer-extension/src/panel/hooks/useEvents/eventFilters.ts
```diff
@@ -75,18 +75,42 @@ function filterOutdatedVersions(events: SdkEvent[]): SdkEvent[] {
...
     .filter((queryPart) => queryPart)
     .map((queryPart) => queryPart.split(':'))
 
   return {
     match: (event: SdkEvent) =>
-      queryParts.every((queryPart) => matchQueryPart(event, queryPart[0], queryPart.length ? queryPart[1] : '')),
+      queryParts.every((queryPart) => {
+        // Hack it to restore the whitespace
+        const searchTerm = queryPart.length > 1 ? queryPart[1].replaceAll(/\\[ ]+/gm, ' ') : ''
```
thomas-lebeau, line 86: **💬 suggestion:** ‏ I think you don't need the `[]` as there is only one character within.

```suggestion
        const searchTerm = queryPart.length > 1 ? queryPart[1].replaceAll(/\\\s+/gm, ' ') : ''
```

developer-extension/src/panel/components/json.tsx
cy-moi: The stringified result has quotation marks which is not supported with our current search terms. Maybe it would be easier to read if I explicitly remove the quotation marks.

developer-extension/src/panel/components/json.tsx
cy-moi: It works, indeed not good practice. I removed this and passed it in as an argument.

developer-extension/src/panel/components/json.tsx
thomas-lebeau: yes, or adding a comment would do too

thomas-lebeau (approved):

developer-extension/src/panel/components/json.tsx
BenoitZugmeyer: This will break other `<CopyMenuItem` usages, because sometimes `value` is an object, so if we use `copy({})` we'll end up copying `[object Object]` (I guess).

```suggestion
        if (typeof value === 'object') {
          copy(JSON.stringify(value, null, 2))
        } else {
          copy(String(value))
        }
```

It should cover your usecase without breaking other ones.

developer-extension/src/panel/components/json.tsx
```diff
@@ -256,6 +259,7 @@ function JsonText({
     menuItems = (
       <>
         <CopyMenuItem value={descriptor.value}>Copy value</CopyMenuItem>
+        <CopyMenuItem value={getSearchQuery(descriptor, columnPath)}>Copy search query</CopyMenuItem>
```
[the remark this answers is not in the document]
cy-moi: I created the item `getCopyMenuItemsForPath` for copying search query, separating the two for the ease of understanding, because column.path and path might be too mixed up.

BenoitZugmeyer (approved):

PR #2775, opened 2024-05-23
📝 [RUM-158] Add jscdoc to public APIs

amortemousque:
## Motivation

JSDoc annotations are visible through the IDE, providing developers with immediate and comprehensive documentation as they integrate with our SDK.

## Changes

- Document all public APIs
- Add links to the public doc in browser SDK warnings when useful
 
## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/tags.ts
+++ b/packages/core/src/domain/configuration/tags.ts
@@ -32,7 +32,9 @@ export function buildTag(key: string, rawValue: string) {
   const valueSizeLimit = TAG_SIZE_LIMIT - key.length - 1
 
   if (rawValue.length > valueSizeLimit || FORBIDDEN_CHARACTERS.test(rawValue)) {
-    display.warn(`${key} value doesn't meet tag requirements and will be sanitized`)
+    display.warn(
+      `${key} value doesn't meet tag requirements and will be sanitized. More details: https://docs.datadoghq.com/getting_started/tagging/#defining-tags`
+    )
   }
 
   // Let the backend do most of the sanitization, but still make sure multiple tags can't be crafted
--- a/packages/core/src/domain/configuration/tags.ts
+++ b/packages/core/src/domain/configuration/tags.ts
@@ -1,4 +1,4 @@
-import { display } from '../../tools/display'
+import { DOCS_ORIGIN, display } from '../../tools/display'
 import type { InitConfiguration } from './configuration'
 
 export const TAG_SIZE_LIMIT = 200
@@ -33,7 +33,7 @@ export function buildTag(key: string, rawValue: string) {
 
   if (rawValue.length > valueSizeLimit || FORBIDDEN_CHARACTERS.test(rawValue)) {
     display.warn(
-      `${key} value doesn't meet tag requirements and will be sanitized. More details: https://docs.datadoghq.com/getting_started/tagging/#defining-tags`
+      `${key} value doesn't meet tag requirements and will be sanitized. More details: ${DOCS_ORIGIN}/getting_started/tagging/#defining-tags`
     )
   }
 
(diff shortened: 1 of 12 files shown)


cy-moi (approved):

packages/core/src/domain/configuration/tags.ts
```diff
@@ -32,7 +32,9 @@ export function buildTag(key: string, rawValue: string) {
   const valueSizeLimit = TAG_SIZE_LIMIT - key.length - 1
 
   if (rawValue.length > valueSizeLimit || FORBIDDEN_CHARACTERS.test(rawValue)) {
-    display.warn(`${key} value doesn't meet tag requirements and will be sanitized`)
+    display.warn(
+      `${key} value doesn't meet tag requirements and will be sanitized. More details: https://docs.datadoghq.com/getting_started/tagging/#defining-tags`
```
BenoitZugmeyer, line 36: **💭 thought:** ‏We now have 5 doc URLs showing up in the RUM bundle. Maybe a constant could save a few bytes:

```ts
const DOCS_ORIGIN = 'https://docs.datadoghq.com'

...
`${DOCS_ORIGIN}/getting_started/tagging/#defining-tags`
```

BenoitZugmeyer (commented):
When:

* running `yarn build`
* then opening `test/app/app.ts` in your IDE
* then hovering `datadogRum.init` or `datadogRum.setUser`

Do you have the documentation showing up? I don't, and I think this is because typescript drops JSDoc comments in the output `.d.ts` files in our case.

I managed to make them appear by declaring `RumPublicApi` explicitely: instead of 

```ts
type RumPublicApi = ReturnType<typeof makeRumPublicApi>
```

we might have to do:

```ts
interface RumPublicApi {
  /**
   * doc
   */
  init(...): void
  // etc.
}
function makeRumPublicApi(...): RumPublicApi {}
```

It might be a bit more verbose as we need to repeat function signatures in that type, but it should offer a nicer developer experience.

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -28,7 +29,135 @@ export interface LoggerConfiguration {
   context?: object
 }
 
-export type LogsPublicApi = ReturnType<typeof makeLogsPublicApi>
+export interface LogsPublicApi extends PublicApi {
```
BenoitZugmeyer, line 32: **💬 suggestion:** ‏Use this type as return type of `makeLogsPublicApi` so we ensure the returned value matches the typing

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -194,37 +397,25 @@ export function makeRumPublicApi(startRumImpl: StartRum, recorderApi: RecorderAp
   const rumPublicApi = makePublicApi({
     init: monitor((initConfiguration: RumInitConfiguration) => strategy.init(initConfiguration)),
```
BenoitZugmeyer, line 398: **💬 suggestion:** ‏With a little bit of tweaking in `makePublicApi`, we could avoid duplicating a bit of typings here.

```ts
export function makePublicApi<T extends PublicApi>(stub: Omit<T, keyof PublicApi>): T {
   ...
  return publicApi as T
}
```

then, no need to specify parameters types anymore:

```suggestion
  const rumPublicApi = makePublicApi({
    init: monitor((initConfiguration) => strategy.init(initConfiguration)),
```

packages/logs/src/boot/logsPublicApi.ts
amortemousque: Oops I forgot 🤦‍♂️

BenoitZugmeyer (approved):

PR #2777, opened 2024-05-23
✨ [RUM-3837] Force Replay recording on sampled-out sessions

N-Boutaib:
## Motivation
Allow session replay recording on sampled out sessions 

## Changes
On the RUM's public API, session replay recording can be forced even if the session is sampled out for replay. This is achieved by passing `{ force: true }` to `DD_RUM.startSessionReplayRecording` function.

When this method is called with `force` option, we need to keep recording until the session ends. To achieve that:
- A flag is added to `_dd_s` cookie
- The cookie is updated to reflect that change when the recording is forced
- Session cache and history context are synchronised across all tabs 

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/domain/session/sessionStore.ts
```diff
@@ -18,8 +19,10 @@ export interface SessionStore {
   restartSession: () => void
   renewObservable: Observable<void>
   expireObservable: Observable<void>
+  forceReplayObservable: Observable<void>
```
nulrich, line 22: **💬 suggestion:** ‏Shouldn't we do a more generic observable/setter. We only have the `forceReplay` setting, but maybe we'll get more settings like this one. We could have `settingsObservable` that notifies sessionManager with the type of setting.

N-Boutaib: Agree, but that could be done later, when we have new use case. Reviews from the previous PR advised to keep exposed methods and logic restricted to forced replay.

thomas-lebeau (approved):

amortemousque:
**💬 suggestion:** ‏We should make the extension support this feature:

- Addind something in the session info

image

- Making the live replay work. For now I see:

image

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -53,6 +55,12 @@ export function startSessionManager<TrackingType extends string>(
     expireObservable.notify()
     sessionContextHistory.closeActive(relativeNow())
   })
+  sessionStore.forceReplayObservable.subscribe(() => {
```
BenoitZugmeyer, line 58: **💭 thought:** ‏RUM logic is leaking into core. Maybe we can think of a way to abstract it, via `computeSessionState`?

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -135,6 +135,13 @@ describe('makeRecorderApi', () => {
       expect(startRecordingSpy).not.toHaveBeenCalled()
     })
 
+    it('should start recording if session is tracked without session replay when forced', () => {
+      setupBuilder.withSessionManager(createRumSessionManagerMock().setTrackedWithoutSessionReplay()).build()
+      rumInit()
+      recorderApi.start({ force: true })
+      expect(startRecordingSpy).toHaveBeenCalledTimes(1)
```
BenoitZugmeyer, line 142: **💬 suggestion:** ‏add an assertion to make sure `setForcedReplay` has been called.

test/e2e/scenario/recorder/recorder.scenario.ts
```diff
@@ -806,6 +806,34 @@ describe('recorder', () => {
...
+      .withRum({ sessionReplaySampleRate: 0 })
+      .withSetup(bundleSetup)
+      .run(async ({ intakeRegistry }) => {
+        await browser.execute(() => {
+          window.DD_RUM!.startSessionReplayRecording({ force: true })
+        })
+
+        await flushEvents()
+
+        expect(intakeRegistry.replaySegments.length).toBe(1)
```
BenoitZugmeyer, line 833: **💬 suggestion:** ‏same here, you could add an assertion to ensure `forceReplay=1` is in the cookie.

packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -58,11 +60,14 @@ export function startRumSessionManager(
       }
       return {
         id: session.id,
-        sessionReplayAllowed: session.trackingType === RumTrackingType.TRACKED_WITH_SESSION_REPLAY,
+        sampledForReplay: session.trackingType === RumTrackingType.TRACKED_WITH_SESSION_REPLAY,
+        sessionReplayAllowed:
+          session.trackingType === RumTrackingType.TRACKED_WITH_SESSION_REPLAY || session.isReplayForced,
```
BenoitZugmeyer, line 65: **💬 suggestion:** ‏Maybe a single property could be enough:
```suggestion
        sessionReplay:
          session.trackingType === RumTrackingType.TRACKED_WITH_SESSION_REPLAY ? 
            SessionReplayState.SAMPLED : 
          session.isReplayForced ?
            SessionReplayState.FORCED :
            SessionReplayState.OFF
```

developer-extension/src/panel/components/tabs/replayTab.tsx
```diff
@@ -43,30 +43,31 @@ function Player() {
...
 }
 
-function generateFullSnapshot() {
+function generateFullSnapshot(isReplayForced: boolean) {
   // Restart to make sure we have a fresh Full Snapshot
+  const startParam = isReplayForced ? '{ force: true }' : ''
   evalInWindow(`
     DD_RUM.stopSessionReplayRecording()
-    DD_RUM.startSessionReplayRecording()
+    DD_RUM.startSessionReplayRecording(${startParam})
```
BenoitZugmeyer, line 70: **❓ question:** ‏Why is it needed? Couldn't we always use `force: true`?

packages/core/src/domain/session/sessionManager.ts
N-Boutaib: Indeed, but that wouldn't be enough because the `sessionContextHistory` entry won't be updated

developer-extension/src/panel/components/tabs/replayTab.tsx
N-Boutaib: Yes we can, I just wanted to have the expected behaviour

N-Boutaib: Done ✔️

packages/rum-core/src/domain/rumSessionManager.ts
N-Boutaib: Nice suggestion. Done ✔️

test/e2e/scenario/recorder/recorder.scenario.ts
N-Boutaib: Done ✔️

packages/rum/src/boot/recorderApi.spec.ts
N-Boutaib: Done ✔️

packages/core/src/domain/session/sessionManager.ts
N-Boutaib: Done ✔️

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -17,12 +18,15 @@ export interface SessionManager<TrackingType extends string> {
...
   expireObservable: Observable<void>
+  sessionStateUpdateObservable: Observable<{ previousState: SessionState; newState: SessionState }>
   expire: () => void
+  updateSessionState: (state: Partial<SessionState>) => void
 }
 
 export interface SessionContext<TrackingType extends string> extends Context {
   id: string
   trackingType: TrackingType
+  isReplayForced: boolean
```
BenoitZugmeyer, line 29: **💭 thought:** ‏Ideally we should remove any notion of "replay" from `@datadog/browser-core`. But I understand this last bit is not trivial to fix, so let's keep it like this for now, and we'll re-explore this later (I opened https://datadoghq.atlassian.net/browse/RUM-4740 for that)

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -53,6 +58,9 @@ export function startSessionManager<TrackingType extends string>(
     expireObservable.notify()
     sessionContextHistory.closeActive(relativeNow())
   })
+  sessionStore.sessionStateUpdateObservable.subscribe(({ previousState, newState }) =>
+    sessionStateUpdateObservable.notify({ previousState, newState })
+  )
```
BenoitZugmeyer, line 63: **💬 suggestion:** ‏no need to create a different `sessionStateUpdateObservable`, you can just return `sessionStore.sessionStateObservable`

test/e2e/scenario/recorder/recorder.scenario.ts
```diff
@@ -806,6 +807,38 @@ describe('recorder', () => {
...
+      .withRum({ sessionReplaySampleRate: 0 })
+      .withSetup(bundleSetup)
+      .run(async ({ intakeRegistry }) => {
+        await browser.execute(() => {
+          window.DD_RUM!.startSessionReplayRecording({ force: true })
+        })
+        const [cookie] = await browser.getCookies([SESSION_STORE_KEY])
+        const cookieForcedReplay = cookie.value.match(/forcedReplay=([\w-]+)/)![1]
+
+        expect(cookieForcedReplay).toBe('1')
```
BenoitZugmeyer, line 834: **🥜 nitpick:** ‏total nitpick, but you could just use something like:
```ts
expect(cookie.value).toContain('forcedReplay=1')
```

BenoitZugmeyer (approved):

amortemousque (approved):

PR #2781, opened 2024-05-24
🐛 [RUM-4629] accept `null` as env/version/service

BenoitZugmeyer:
## Motivation

Some customer pass `null` as version. Now that we validate that parameter, the initialization fails.

See https://github.com/DataDog/dd-sdk-flutter/issues/616
Regression caused by https://github.com/DataDog/browser-sdk/pull/2744

## Changes

Do not reject `null` for tags.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.spec.ts
+++ b/packages/core/src/domain/configuration/configuration.spec.ts
@@ -218,21 +218,27 @@ describe('validateAndBuildConfiguration', () => {
   })
 
   describe('env parameter validation', () => {
-    it('should not validate the env parameter', () => {
+    it('should validate the env parameter', () => {
       validateAndBuildConfiguration({ clientToken, env: false as any })
       expect(displaySpy).toHaveBeenCalledOnceWith('Env must be defined as a string')
     })
   })
 
   describe('service parameter validation', () => {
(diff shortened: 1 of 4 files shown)


thomas-lebeau (approved):

cy-moi (approved):

gscandola:
Hello @BenoitZugmeyer 👋 (😉 ) ,

Introducing the `null` capability induce a mismatch between the typing `InitConfiguration.service` and `ViewOptions.service` (see [here](https://github.com/DataDog/browser-sdk/blob/0783df470ff2f456b9755e3e170129e6b1375752/packages/rum-core/src/domain/view/trackViews.ts#L76)).

Is it intended ?

BenoitZugmeyer:
👋 Hi @gscandola, it's not intended!
We didn't expect `null` to be passed as version/service. In practice it was the case for certain integrations using `init(..)`, so we fixed it, but we are not aware of such issue for `startView()`. In any case, let's try to make it uniform https://github.com/DataDog/browser-sdk/pull/2798

gscandola:
Thanks a lot 👍

PR #2785, opened 2024-05-28
🐛 [RUM-98] Warn when SDK is loaded multiple times

RomanGaignault:
## Motivation

We should warn when the SDK is loaded multiple times as it can lead to unexpected behaviour. 

## Changes

As we don't want a breaking change for now, we can just send a warning if the sdk is already loaded in the define global function.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/boot/init.ts
+++ b/packages/core/src/boot/init.ts
@@ -35,8 +35,8 @@ export function makePublicApi<T>(stub: T): T & { onReady(callback: () => void): 
 
 export function defineGlobal<Global, Name extends keyof Global>(global: Global, name: Name, api: Global[Name]) {
   const existingGlobalVariable = global[name] as { q?: Array<() => void>; version?: string } | undefined
-  if (global[name] && existingGlobalVariable && !existingGlobalVariable.q && existingGlobalVariable.version) {
-    display.warn('SDK is loaded more than once')
+  if (existingGlobalVariable && !existingGlobalVariable.q && existingGlobalVariable.version) {
+    display.warn('SDK is loaded more than once. This is unsupported and might have unexpected behavior.')
   }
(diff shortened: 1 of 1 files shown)


packages/core/src/boot/init.ts
```diff
@@ -33,7 +34,10 @@ export function makePublicApi<T>(stub: T): T & { onReady(callback: () => void):
 }
 
 export function defineGlobal<Global, Name extends keyof Global>(global: Global, name: Name, api: Global[Name]) {
-  const existingGlobalVariable = global[name] as { q?: Array<() => void> } | undefined
+  const existingGlobalVariable = global[name] as { q?: Array<() => void>; version?: string } | undefined
+  if (global[name] && existingGlobalVariable && !existingGlobalVariable.q && existingGlobalVariable.version) {
```
BenoitZugmeyer, line 38: **💬 suggestion:** ‏`global[name] && existingGlobalVariable` is redundant.

```suggestion
  if (existingGlobalVariable && !existingGlobalVariable.q && existingGlobalVariable.version) {
```

packages/core/src/boot/init.ts
```diff
@@ -33,7 +34,10 @@ export function makePublicApi<T>(stub: T): T & { onReady(callback: () => void):
...
+    display.warn('SDK is loaded more than once')
```
BenoitZugmeyer, line 39: **💬 suggestion:** ‏
```suggestion
    display.warn('SDK is loaded more than once. This is unsupported and might have unexpected behavior.')
```

thomas-lebeau (approved):

N-Boutaib (approved):

PR #2786, opened 2024-05-28
✨ [RUM-4014] DD_LOGS: add handling stack in beforeSend context

thomas-lebeau:
## Motivation

Create a stack trace and expose it in the `beforeSend`'s `domainContext` argument.
It allow customers to initialise the browser SDK with something like:

```js
const SERVICE_REGEX = /(?:browser-sdk-test-playground\/module-federation\/|webpack:\/\/)(?<service>\w*)\//;
  
DD_LOGS.init({
    const service =
      context?.handlingStack?.match(SERVICE_REGEX)?.groups?.service;

    if (service) {
      event.service = service; //TODO: this is not yet implemented
    }
    return true;
  },
});
```

## Changes

- support the following APIs
  - [x] `console.*()`
  - [x] `logger.*()`
  - [x] `fetch` console errors?
  - [x] `xhr` console errors?
- moved `isAuthorized` to a new file to avoid cyclic dependencies

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/logs/src/domain/logger/isAuthorized.ts
thomas-lebeau, line 1: refactor: moved here to avoid cyclic dependency

packages/logs/src/domain/networkError/networkErrorCollection.ts
```diff
@@ -37,17 +37,24 @@ export function startNetworkErrorCollection(configuration: LogsConfiguration, li
...
+        )
       }
     }
 
-    function onResponseDataAvailable(responseData: unknown) {
+    function onResponseDataAvailable(responseData: unknown, handlingStack?: string) {
       const domainContext: LogsEventDomainContext<typeof ErrorSource.NETWORK> = {
         isAborted: request.isAborted,
+        handlingStack,
       }
```
BenoitZugmeyer, line 58: **💬 suggestion:** ‏You already have access to `request` in that function, no need to pass `handlingStack` as an extra argument.

test/e2e/scenario/microfrontend.scenario.ts
```diff
@@ -126,8 +139,69 @@ describe('microfrontend', () => {
...
+          expect(event).toBeTruthy()
+          expect(event?.context).toEqual({
+            handlingStack: jasmine.stringMatching(HANDLING_STACK_REGEX),
+          })
+        })
+    })
+  })
+
+  describe('logger apis', () => {
+    const loggerApis = ['log', 'ok', 'debug', 'info', 'notice', 'warn', 'error', 'critical', 'alert', 'emerg'] as const
```
BenoitZugmeyer, line 173: **💬 suggestion:** ‏It takes more than 1 minute to run all microfrontends e2e tests locally. I don't think it's worth covering every log level, and using a single one of them should be sufficient. The code is the same for all of them anyway, and unit tests should have the full coverage.

test/e2e/scenario/microfrontend.scenario.ts
```diff
@@ -126,8 +139,69 @@ describe('microfrontend', () => {
       expect(event?.context?.handlingStack).toMatch(HANDLING_STACK_REGEX)
     })
 
+  describe('console apis', () => {
+    const consoleApis = ['log', 'debug', 'error', 'info', 'warn'] as const
```
BenoitZugmeyer, line 143: **💬 suggestion:** ‏Same here, a single log API should be enough.

packages/logs/src/domain/networkError/networkErrorCollection.ts
thomas-lebeau: 🤦 of course! thanks

packages/logs/src/domain/logger.ts
```diff
@@ -61,79 +49,105 @@ export class Logger {
...
   ok(message: string, messageContext?: object, error?: Error) {
-    this.log(message, messageContext, StatusType.ok, error)
+    let handlingStack: string | undefined
+
+    if (isAuthorized(StatusType.ok, HandlerType.http, this)) {
+      handlingStack = createHandlingStack()
+    }
+
+    this.logImplementation(message, messageContext, StatusType.ok, error, handlingStack)
   }
```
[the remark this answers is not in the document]
thomas-lebeau: > always create the handlingstack

Indeed, this would save ~660 Bytes as per my calculations. However generating a stack trace could be expensive, especially if used a lot.

Another solution is to skip 3 frame instead of 2 for logs. Customers can always increase the stack size if needed.

wdyt?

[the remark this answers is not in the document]
thomas-lebeau: Apply the prototype trick we talked during the daily. +240B vs +909B before 🎉 
Thank you @BenoitZugmeyer for the suggestion

packages/logs/src/domain/console/consoleCollection.spec.ts
```diff
@@ -55,6 +56,25 @@ describe('console collection', () => {
...
+        validateAndBuildLogsConfiguration({ ...initConfiguration, forwardConsoleLogs: 'all' })!,
+        lifeCycle
+      ))
+
+      /* eslint-disable-next-line no-console */
+      console[api as keyof typeof LogStatusForApi]('foo', 'bar')
+
+      expect(rawLogsEvents[0].domainContext).toEqual({
+        handlingStack: jasmine.any(String),
+      })
```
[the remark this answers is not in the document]
BenoitZugmeyer: Ah right! Looks good then

BenoitZugmeyer (approved):

N-Boutaib (approved):

PR #2788, opened 2024-05-30
✨ [RUM-3965] make service and version fields modifiable

thomas-lebeau:
## Motivation

As part of the micro frontend support, we are allowing customer to override `service` and `version` properties of most events

## Changes

Adds `service` and `version` to the list of modifiable properties for `error`, `action` and `resource` event types.
Note: logs events not protected against modifications

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


N-Boutaib (commented):
LGTM
Could you add a small unit test, in `assembly.spec.ts`, to cover that `service` & `version` are editable only if the FF is enabled ?

N-Boutaib (approved):

nulrich (approved):

PR #2795, opened 2024-06-05
♻️ Sends the conf telemetry from preStartRum

amortemousque:
## Motivation

Now that we collect pre start telemetry (cf [PR](https://github.com/DataDog/browser-sdk/pull/2755)) we can move configuration sending to `preStartRum`. This avoid passing the `initConfiguration` down to [startRum](https://github.com/DataDog/browser-sdk/blob/94d4fbc152286d7dbdc4938bde9e83d3bb2e8dac/packages/rum-core/src/boot/startRum.ts#L54).

## Changes

Sends configuration telemetry from preStartRum

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/boot/preStartRum.ts
+++ b/packages/rum-core/src/boot/preStartRum.ts
@@ -9,6 +9,7 @@ import {
   clocksNow,
   assign,
   getEventBridge,
+  addTelemetryConfiguration,
 } from '@datadog/browser-core'
 import type { TrackingConsentState, DeflateWorker } from '@datadog/browser-core'
 import {
@@ -18,6 +19,7 @@ import {
 } from '../domain/configuration'
 import type { CommonContext } from '../domain/contexts/commonContext'
 import type { ViewOptions } from '../domain/view/trackViews'
+import { serializeRumConfiguration } from '../domain/configuration'
 import type { RumPublicApiOptions, Strategy } from './rumPublicApi'
 import type { StartRumResult } from './startRum'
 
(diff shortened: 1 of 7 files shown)


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -383,7 +382,7 @@ export function makeRumPublicApi(
         deflateWorker
       )
 
-      strategy = createPostStartStrategy(initConfiguration, startRumResult)
+      strategy = createPostStartStrategy(strategy.initConfiguration!, startRumResult)
```
BenoitZugmeyer, line 385: Suggestion: keep passing `initConfiguration` in `doStartRum` so you don't need this non-null assertions

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -85,6 +86,7 @@ export function createPreStartStrategy(
 
       // Expose the initial configuration regardless of initialization success.
       cachedInitConfiguration = initConfiguration
+      addTelemetryConfiguration(serializeRumConfiguration(initConfiguration))
```
BenoitZugmeyer, line 89: Suggestion: move this line below , after all early retuns, so we don't collect the configuration if the SDK doesn't actually run

thomas-lebeau (approved):

BenoitZugmeyer (commented):
Suggestion: do the same in logs sor `preStartRum` and `preStartLogs` are similar

packages/rum-core/src/boot/preStartRum.ts
amortemousque: The telemetry is drained inside startRum so we will collect the telemetry only if the SDK actually run, but it can save a bit of memory :)

packages/rum-core/src/boot/rumPublicApi.ts
amortemousque: In fact I think of passing the all preStartStrategy as param cf commit https://github.com/DataDog/browser-sdk/pull/2795/commits/0c3d32fca91e0648cc5545945b1f84d2053b29a6
Wdyt?

PR #2796, opened 2024-06-05
♻️ Set the experimental feature flags as early as possible

amortemousque:
## Motivation

Being able to use experimental features early in the SDK lifecycle

## Changes

Initialize the experimental feature in the `preStartRum` instead of waiting the `startRum`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/experimentalFeatures.ts
+++ b/packages/core/src/tools/experimentalFeatures.ts
@@ -9,6 +9,9 @@
  */
 // We want to use a real enum (i.e. not a const enum) here, to be able to check whether an arbitrary
 // string is an expected feature flag
+
+import { objectHasValue } from './utils/objectUtils'
+
 // eslint-disable-next-line no-restricted-syntax
 export enum ExperimentalFeature {
   WRITABLE_RESOURCE_GRAPHQL = 'writable_resource_graphql',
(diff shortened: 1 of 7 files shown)


packages/core/src/tools/experimentalFeatures.ts
```diff
@@ -19,6 +22,16 @@ export enum ExperimentalFeature {
 
 const enabledExperimentalFeatures: Set<ExperimentalFeature> = new Set()
 
+export function initFeatureFlags(enableExperimentalFeatures?: string[] | undefined) {
+  if (Array.isArray(enableExperimentalFeatures)) {
+    addExperimentalFeatures(
+      enableExperimentalFeatures.filter((flag): flag is ExperimentalFeature =>
+        objectHasValue(ExperimentalFeature, flag)
+      )
```
BenoitZugmeyer, line 30: *Thoughts:* we don't really need this validation, we could save a few bytes by removing it

BenoitZugmeyer (approved):

packages/core/src/tools/experimentalFeatures.ts
amortemousque: Yes, I was also wondering why we had it. I'll remove it

amortemousque: Hum, it seems there is a test made on purpose https://github.com/DataDog/browser-sdk/blob/77596277f4c9e70f55eb4e373323b63e22f7ccc7/packages/core/src/tools/experimentalFeatures.spec.ts#L46

N-Boutaib (approved):

PR #2799, opened 2024-06-07
⚗️[RUM-4780] Remote configuration

amortemousque:
## Motivation

Add remote configuration capability 

## Changes

- Add new init parameter `remoteConfigurationId`
- Wait for the remote configuration in the preStart module
- Overrides the static init configuration with the remote configuration values

## How to test 

```ts
window.DD_RUM.init({
    remoteConfigurationId: "50d8ba4b-2ac1-481f-b16e-f110699ecf66", // only config available
    enableExperimentalFeatures: ['remote_configuration'],
    ...
})
```

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/preStartRum.spec.ts
```diff
@@ -452,6 +502,22 @@ describe('preStartRum', () => {
 
       expect(strategy.initConfiguration).toEqual(initConfiguration)
     })
+
+    it('returns the initConfiguration with the remote configuration when a remoteConfigurationId is provided', (done) => {
```
amortemousque, line 506: I'm not sure about this behaviour. Basically `DD_RUM.getInitConfiguration()` will also have the remote configuration values. It has the advantage to make the browser extension reflect the remote config without any change.
Wdyt?

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -0,0 +1,53 @@
+import type { DefaultPrivacyLevel } from '@datadog/browser-core'
+import { display, addEventListener, assign } from '@datadog/browser-core'
+import type { RumInitConfiguration } from './configuration'
+
+export const REMOTE_CONFIGURATION_URL = 'https://www.datad0g-browser-agent.com/configuration'
```
amortemousque, line 5: For testing purposes, I hosted the RC in the same CDN as the bundles but it will change once the Remote config provides their CDN

packages/core/src/browser/addEventListener.ts
```diff
@@ -90,7 +89,7 @@ type EventMapFor<T> = T extends Window
  * * returns a `stop` function to remove the listener
  */
 export function addEventListener<Target extends EventTarget, EventName extends keyof EventMapFor<Target> & string>(
-  configuration: Configuration,
+  configuration: { allowUntrustedEvents?: boolean | undefined },
```
thomas-lebeau, line 92: **💬 suggestion:** ‏ 
```suggestion
  configuration: { allowUntrustedEvents?: boolean },
```

or 

```suggestion
  configuration: { allowUntrustedEvents: boolean | undefined },
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -28,7 +29,7 @@ export function getEventBridge<T, E>() {
       return JSON.parse(eventBridgeGlobal.getCapabilities?.() || '[]') as BridgeCapability[]
     },
     getPrivacyLevel() {
-      return eventBridgeGlobal.getPrivacyLevel?.()
+      return eventBridgeGlobal.getPrivacyLevel?.() as DefaultPrivacyLevel
```
thomas-lebeau, line 32: **💭 thought:** ‏ Can we make the `DatadogEventBridge`'s type better instead? (line 10)

```diff
-  getPrivacyLevel?(): string
+  getPrivacyLevel?(): DefaultPrivacyLevel

File: packages/rum-core/src/boot/preStartRum.spec.ts (hunk 1 above)
Comment (thomas-lebeau) [reply packages/rum-core/src/boot/preStartRum.spec.ts:506]: **💭 thought:** ‏ Currently `DD_RUM.getInitConfiguration()` returns exactly what the customer called `init()` with. I wouldn't mind it to return what the SDK was actually initialised with, this means the combination of what the customer called `init()` with, the remote config if applicable and the defaults.

I think it would be useful information to debug rum, both for customers, for us through telemetry and even for support.

I'm not sure if it is safe to do that or if this will be a breaking change.

File: packages/rum-core/src/domain/configuration/configuration.spec.ts [hunk 5]
```diff
@@ -453,7 +458,7 @@ describe('serializeRumConfiguration', () => {
         ? `use_${CamelToSnakeCase<Key>}`
         : Key extends 'trackLongTasks'
           ? 'track_long_task' // oops
-          : Key extends 'applicationId' | 'subdomain'
+          : Key extends 'applicationId' | 'subdomain' | 'remoteConfigurationId'
```
thomas-lebeau, line 461: **❓ question:** ‏ Don't we want to track `remoteConfigurationId` in our telemetry? If it's for privacy reason, maybe we can at least have a flag like `useRemoteConfig: boolean`?

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -0,0 +1,53 @@
...
+  remoteInitConfiguration: RumRemoteConfiguration
+) {
+  return assign({}, initConfiguration, remoteInitConfiguration)
+}
+
+export function fetchRemoteConfiguration(
+  configuration: RumInitConfiguration,
+  callback: (remoteConfiguration: RumRemoteConfiguration) => void
+) {
+  const xhr = new XMLHttpRequest()
```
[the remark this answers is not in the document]
amortemousque: Indeed yes 🙂

packages/core/src/browser/addEventListener.ts
amortemousque: I reused the same type definition as the InitConfiguration params https://github.com/DataDog/browser-sdk/blob/f5b9b0667fc72853347c4f59398eb53c62b78ad9/packages/core/src/domain/configuration/configuration.ts#L61-L62.
Basically we want customer to be able to do:
- `{ allowUntrustedEvents: undefined }`
- `{ allowUntrustedEvents: false }`
- `{ }`

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  addEventListener(configuration, xhr, 'load', function () {
+    if (xhr.status === 200) {
+      callback(JSON.parse(xhr.responseText))
+    } else {
+      displayRemoteConfigurationFetchingError()
```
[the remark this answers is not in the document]
amortemousque: If the RC fetching fails, the SDK won't be initialized and so no telemetry data can be sent.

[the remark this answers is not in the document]
thomas-lebeau: 🤦 makes sense! thanks

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -204,6 +205,7 @@ export function validateAndBuildRumConfiguration(
       traceContextInjection: objectHasValue(TraceContextInjection, initConfiguration.traceContextInjection)
         ? initConfiguration.traceContextInjection
         : TraceContextInjection.ALL,
+      storeContextsAcrossPages: initConfiguration.storeContextsAcrossPages,
```
BenoitZugmeyer, line 208: **❓ question:** ‏Why do we need this? This property is defined in `core` no?

packages/rum-core/src/boot/preStartRum.spec.ts
amortemousque: From what I identifyed in [Datadog repos](https://github.com/search?q=org%3ADataDog+getInitConfiguration&type=code), we use the API for
- [Browser SDK extension](https://github.com/DataDog/browser-sdk/blob/b839f8cb52b676241452dba00fa0b8b166cb3bfe/developer-extension/src/panel/hooks/useSdkInfos.ts#L65): to display the conf
- [synthetics-worker](https://github.com/DataDog/synthetics-worker/blob/3735e7b0f1250561272d4527b523ffd58dedcd81/chrome-extension/src/extension/main-world.ts#L71): Use it to get the `applicationId`
- [rum-react-integration-examples](https://github.com/DataDog/rum-react-integration-examples/blob/4fc30362273ac7e126d9767fe67b77a0648ed375/src/Router/RumRouteComponentWrapper.tsx#L58) (old not maintained repo) uses it to get `trackViewsManually`
- [dd-profile-browser](https://github.com/DataDog/dd-profile-browser/blob/e3fd3dcd54ee2970ab74798f3dc7a87571ae458d/src/profiler/rum-configuration-watcher.ts#L19): it is used to collected profiles with `service`, `env`, and `version` tags

Clearly if we get one of fields applicationId, service, env or version, from the RC we will have some issues. But luckily for now they won't be available in RC :).  So the question is, should we anticipate by introducing a new API right away and defining the `getInitConfiguration()` has only returning whats passed to the `init()` API
I'll create an RFC

amortemousque: Seen with @BenoitZugmeyer, since the feature is under FF, lets keep the behaviour as is and revisite later when needed.

BenoitZugmeyer (approved):

thomas-lebeau (approved):

packages/rum-core/src/boot/preStartRum.spec.ts
thomas-lebeau: perfect! Sounds good!

PR #2801, opened 2024-06-10
✨ [RUM-4781] Update Custom Web Vitals API

nulrich:
## Motivation

We would like to be able to report concurrent vitals with the same name and allowing better visualisations.

### Update custom vital structure: 

The custom vital value is now in `duration` field instead of in `custom` object.

### Update custom vital API:

`startDurationVital` now returns a vital instance that is used to stop the vital instead of calling `stopDurationVital`.

Introducing `addDurationVital` to create a vital with a custom duration and startVital values.

Both methods now have a `details` attribute that can be used to differentiate vitals using the same name.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -74,9 +98,9 @@ function processVital(vital: DurationVital, valueComputedBySdk: boolean): RawRum
       id: generateUUID(),
       type: vital.type,
       name: vital.name,
-      custom: {
-        [vital.name]: vital.value,
-      },
+      duration: vital.duration,
+      details: vital.details,
+      custom: {},
```
N-Boutaib, line 103: ❓ **Question**: why not opting for making custom nullable ?

N-Boutaib (commented):
It would interesting to test these changes on staging to have this new events types ingested

amortemousque:
Looks good! 
I think we should think of naming convention at some point though. For instance in the different collections, we have `processXXX()` method that returns `RawRumXXXEvent`. However the input type names follow different convention:
Sometime prefix with `Raw`, sometimes is suffixed with `Event`, sometime no preffix or suffix:
- [processError(error: **Raw**Error)](https://github.com/DataDog/browser-sdk/blob/e7162a5399a0d7d311344b35ce1c07fd3360b7bd/packages/rum-core/src/domain/error/errorCollection.ts#L96)
- [processVital(vital: DurationVital)](https://github.com/DataDog/browser-sdk/blob/e7162a5399a0d7d311344b35ce1c07fd3360b7bd/packages/rum-core/src/domain/vital/vitalCollection.ts#L94)
- [processViewUpdate(view: View**Event**) ](https://github.com/DataDog/browser-sdk/blob/e7162a5399a0d7d311344b35ce1c07fd3360b7bd/packages/rum-core/src/domain/view/viewCollection.ts#L44)
- [processRequest(request: RequestComplete**Event**)](https://github.com/DataDog/browser-sdk/blob/e7162a5399a0d7d311344b35ce1c07fd3360b7bd/packages/rum-core/src/domain/resource/resourceCollection.ts#L60)
- [processAction(action: AutoAction | CustomAction)](https://github.com/DataDog/browser-sdk/blob/e7162a5399a0d7d311344b35ce1c07fd3360b7bd/packages/rum-core/src/domain/action/actionCollection.ts#L68)

packages/rum-core/src/boot/startRum.ts
```diff
@@ -178,6 +179,11 @@ export function startRum(
     urlContexts
   )
 
+  const startDurationVital = (startVital: DurationVitalStart) =>
+    createVitalInstance((vital) => {
+      vitalCollection.addDurationVital(vital)
+    }, startVital)
```
amortemousque, line 185: **💬 suggestion:**  This code could be collocated in vitalCollection module. For instance something like
```ts
export function startVitalCollection(lifeCycle: LifeCycle, pageStateHistory: PageStateHistory) {
  function isValid(vital: DurationVital) {
    return !pageStateHistory.wasInPageStateDuringPeriod(PageState.FROZEN, vital.startClocks.relative, vital.duration)
  }

  function addDurationVital(vitalAdd: DurationVitalAdd) {
    const vital = assign({ type: VitalType.DURATION }, vitalAdd)

    if (isValid(vital)) {
      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processVital(vital, true))
    }
  }

  return {
    addDurationVital,
    startDurationVital: (startVital: DurationVitalStart) =>
      createVitalInstance((vital) => {
        addDurationVital(vital)
      }, startVital),
  }
}
```
‏

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -9,61 +9,85 @@ import { PageState } from '../contexts/pageStateHistory'
...
       if (isValid(vital)) {
         lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processVital(vital, true))
       }
     },
   }
 }
 
-function buildDurationVital(vitalStart: DurationVitalStart, vitalStop: DurationVitalStop) {
+export function createVitalInstance(
+  cb: (vital: DurationVital) => void,
```
amortemousque, line 59: **💬 suggestion:** ‏Could be nice to have an more explicite name 
```suggestion
  stopCallback: (vital: DurationVital) => void,
```

packages/rum-core/src/domain/vital/vitalCollection.ts
amortemousque: Yes, why do we still need `custom`?

nulrich: We don't. I'll remove it.

packages/rum-core/src/boot/startRum.ts
nulrich: That's what I planned. But I has extracted the logic to be able to provide an implementation for `preStartRum`:

```typescript
return createVitalInstance((vital) => {
}, vitalStart)
```

This API function is specific because it returns the vital instance on which can call `stop` (it calls the callback).

amortemousque: I think it is reasonable to have the preStart implementation in preStart and the postStart in vitalCollection. Wdyt?

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -187,11 +188,13 @@ export function createPreStartStrategy(
...
+      return createVitalInstance((vital) => {
+        bufferApiCalls.add((startRumResult) => startRumResult.addDurationVital(vital))
+      }, vitalStart)
     },
 
-    stopDurationVital(vitalStart) {
-      bufferApiCalls.add((startRumResult) => startRumResult.stopDurationVital(vitalStart))
+    addDurationVital: (vital) => {
+      bufferApiCalls.add((startRumResult) => startRumResult.addDurationVital(vital))
     },
```
amortemousque, line 198: **💬 suggestion:** ‏It could be simplied a bit as follow:
```ts
 const addDurationVital = (vital: DurationVitalAdd) => {
  }
  return {
    ...
    addDurationVital,
    startDurationVital(vitalStart) {
      return createVitalInstance((vital) => addDurationVital(vital), vitalStart)
    }
  }
```

packages/rum-core/src/boot/startRum.ts
nulrich: 💯 make sense

developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
```diff
@@ -339,12 +339,16 @@ function LongTaskDescription({ event }: { event: RumLongTaskEvent }) {
...
+  const vitalName = event.vital.name
+  const vitalValue = event.vital.duration
+  const vitalDetails = event.vital.details
   return (
     <>
-      Custom <Emphasis>{event.vital.type}</Emphasis> vital: <Emphasis>{vitalName}</Emphasis> of{' '}
-      <Emphasis>{vitalValue}</Emphasis>
+      Custom <Emphasis>{event.vital.type}</Emphasis> vital:{' '}
+      <Emphasis>
+        {vitalName} - {vitalDetails}
```
amortemousque, line 349: **💬 suggestion:** ‏If `vitalDetails` is empty do not add the separator ` - `

amortemousque (approved):

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -9,61 +9,91 @@ import { PageState } from '../contexts/pageStateHistory'
...
+}
+
+export interface DurationVitalAdd {
   name: string
-  stopClocks: ClocksState
+  startClocks: ClocksState
+  duration: Duration
   context?: Context
+  details?: string
+}
```
BenoitZugmeyer, line 27: **💬 suggestion:** ‏to avoid an intermediary data structure, I would just use `DurationVital` in place of `DurationVitalAdd`. This implies adding the `type` in `rumPublicApi`, but I think it's alright?

packages/rum-core/test/fixtures.ts
```diff
@@ -48,9 +48,8 @@ export function createRawRumEvent(type: RumEventType, overrides?: Context): RawR
             id: generateUUID(),
             type: VitalType.DURATION,
             name: 'timing',
-            custom: {
-              timing: 0 as ServerDuration,
-            },
+            duration: 0 as ServerDuration,
+            custom: {},
```
BenoitZugmeyer, line 52: **💬 suggestion:** ‏remove the `custom` field here

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -9,61 +9,91 @@ import { PageState } from '../contexts/pageStateHistory'
...
+      stopCallback(buildDurationVital(vitalStart, startClocks, vitalStop, stopClocks))
     },
   }
 }
 
-function buildDurationVital(vitalStart: DurationVitalStart, vitalStop: DurationVitalStop) {
+function buildDurationVital(
+  vitalStart: DurationVitalStart,
+  startClocks: ClocksState,
+  stopOptions: DurationVitalStop = {},
```
amortemousque, line 87: **💬 suggestion:** ‏
```suggestion
  vitalStop: DurationVitalStop = {},
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -322,36 +324,53 @@ export function makeRumPublicApi(
...
+         * @param options.startTime epoch timestamp of the start of the custom vital
+         * @param options.duration duration of the custom vital
          * @param options.context custom context attached to the vital
-         * @param options.stopTime epoch timestamp of the stop of the custom vital (if not set, will use current time)
+         * @param options.details  Details of the vital.
          */
-        ;(rumPublicApi as any).stopDurationVital = monitor(
-          (name: string, options?: { context?: object; stopTime?: number }) => {
-            strategy.stopDurationVital({
+        ;(rumPublicApi as any).addDurationVital = monitor(
```
amortemousque, line 350: **💬 suggestion:** To avoid confusion `startDurationVital`, ‏we could add a warning if the no `options` is passed. We could say `You might want to us the startVitalDuration API ..`. I'm saying this because it happens to me :) I used addDurationVital instead of startDurationVital and it silently failed.

PR #2807, opened 2024-06-12
⚡️[RUM-3570] Batch the records for 16ms minimum before processing them

amortemousque:
## Motivation

When recording for Session Replay, [MutationRecord](https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord) objects are batched together (see [mutationBatch](https://github.com/DataDog/browser-sdk/blob/fd44db8f0a9e6b0e08396deb182eebf69a3ef4d3/packages/rum/src/domain/record/mutationBatch.ts#L1)), then deduplicated on processing (ex: [here](https://github.com/DataDog/browser-sdk/blob/d48922fe03946c67412665a2976e4b4f978c28b6/packages/rum/src/domain/record/trackers/trackMutation.ts#L337)).

When the same mutation happens continuously on the same element, for example with JS-driven animations, a lot of MutationRecords can be notified by the browser in a very short amount of time, which can significantly inflate the data sent to Datadog, and have a performance impact both on the recorder side and on the player side (see below). Those mutations could be deduplicated if they were batched together.

## Changes

Batch the records for 16ms minimum before processing them

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/mutationBatch.spec.ts
+++ b/packages/rum/src/domain/record/mutationBatch.spec.ts
@@ -1,28 +1,38 @@
-import { collectAsyncCalls } from '@datadog/browser-core/test'
-import { createMutationBatch } from './mutationBatch'
+import type { Clock } from '@datadog/browser-core/test'
+import { mockClock } from '@datadog/browser-core/test'
+import { MUTATION_PROCESS_MIN_DELAY, createMutationBatch } from './mutationBatch'
 import type { RumMutationRecord } from './trackers'
 
 describe('createMutationBatch', () => {
   let mutationBatch: ReturnType<typeof createMutationBatch>
   let processMutationBatchSpy: jasmine.Spy<(mutations: RumMutationRecord[]) => void>
+  let requestIdleCallbackSpy: jasmine.Spy
+  let clock: Clock
 
   beforeEach(() => {
+    clock = mockClock()
     processMutationBatchSpy = jasmine.createSpy()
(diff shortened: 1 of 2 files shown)


packages/rum/src/domain/record/mutationBatch.spec.ts
```diff
@@ -1,28 +1,38 @@
...
     const mutation = { type: 'childList' } as RumMutationRecord
     mutationBatch.addMutations([mutation])
 
-    collectAsyncCalls(processMutationBatchSpy, 1, (calls) => {
-      expect(calls.mostRecent().args[0]).toEqual([mutation])
-      done()
-    })
+    expect(requestIdleCallbackSpy).toHaveBeenCalled()
+    clock.tick(MUTATION_PROCESS_MIN_DELAY)
+    expect(processMutationBatchSpy).toHaveBeenCalledWith([mutation])
```
thomas-lebeau, line 35: Seems like this would pass even without the throttling, no?

maybe we should test:
```suggestion
    expect(processMutationBatchSpy).not.toHaveBeenCalled()
    clock.tick(MUTATION_PROCESS_MIN_DELAY)
    expect(processMutationBatchSpy).toHaveBeenCalledWith([mutation])
```

amortemousque: Yes, I fixed the test

thomas-lebeau (approved):

nulrich (approved):

PR #2808, opened 2024-06-13
✨ [RUM-4819] Add an experimental 'updateViewName' API

RomanGaignault:
## Motivation

Currently in RUM, when we access a page that contains a redirection path in the router, by using the original (or former) URL, two view events are created for the same page, as RUM does the next assumption:
1 path/URL === 1 view

## Changes

In the Browser SDK manual tracking, we could make it possible to update an already created view’s name by creating an new api called updateViewName
image

image

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/view/trackViews.ts
BenoitZugmeyer, line 1: **💬 suggestion:** ‏line 270, the view `name` is returned as a property. Since the `name` can change now, maybe use a getter:

```ts
{
  get name() { return name },
  ...
}
```

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -897,4 +906,49 @@ describe('view event count', () => {
...
+        return viewTest
+      })
+    })
+
+    afterEach(() => {
+      resetExperimentalFeatures()
+    })
+
+    it('should update an undefined view name if the experimental feature is enabled', () => {
+      addExperimentalFeatures([ExperimentalFeature.UPDATE_VIEW_NAME])
```
BenoitZugmeyer, line 926: **💬 suggestion:** ‏use `mockExperimentalFeatures`, so you don't have to call `resetExperimentalFeatures`

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -478,6 +488,10 @@ export function makeRumPublicApi(
 
     startView,
 
+    updateViewName: (name) => {
+      strategy.updateViewName(name)
+    },
```
BenoitZugmeyer, line 493: **💬 suggestion:** ‏don't expose the method the experimental feature is not enabled, similarly to `startDurationVital` above. Don't expose the type/documentation either.

N-Boutaib (approved):

BenoitZugmeyer (approved):

PR #2809, opened 2024-06-13
✨⚗️ [RUM-4469] introduce a plugin system 

BenoitZugmeyer:
## Motivation

Experiment with a plugin system to provide custom integrations.

This is experimental: the plugin interface might change in any version, be warned!

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -296,6 +305,7 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
       track_user_interactions: configuration.trackUserInteractions,
       track_resources: configuration.trackResources,
       track_long_task: configuration.trackLongTasks,
+      plugins: configuration.plugins?.map((plugin) => assign({ name: plugin.name }, plugin.serializeConfiguration?.())),
```
thomas-lebeau, line 308: **💭 thought:** ‏ What about doing the serialization within rum instead of the plugin? This way we can control how the serialization is done and might avoid duplicating serialization logic across multiple plugins

packages/rum-core/src/domain/plugins.ts
```diff
@@ -0,0 +1,27 @@
+import type { RumPublicApi } from '../boot/rumPublicApi'
+import type { RumInitConfiguration } from './configuration'
+
+export interface RumPlugin {
+  name: string
+  serializeConfiguration?(): Record<string, unknown>
```
amortemousque, line 6: **❓ question:** ‏What is the serializeConfiguration meant for? Is it only to be passed in the telemetry config?

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -140,6 +141,8 @@ export function createPreStartStrategy(
         return
       }
 
+      callHook(initConfiguration.plugins, 'onInit', { initConfiguration, publicApi })
```
amortemousque, line 144: **💬 suggestion:** ‏To anticipate a bit future hooks, it could be nice to have
```suggestion
      registerePlugins(initConfiguration.plugins) // At some point we could have some checks like ensuring no duplicated plugins
      callHook('onInit', { initConfiguration, publicApi })
```

packages/rum-core/src/domain/configuration/configuration.ts
BenoitZugmeyer: Hum, 'serialization' might not be the best term for it. We don't serialize anything per-se, we transform the configuration before sending it as telemetry event. We need to ensure that we don't leak PII (ex: we avoid sending URLs specified in the configuration, ex: `configuration.proxy = 'https://...'` becomes `use_proxy: true`). It's not possible to define a generic way to do it.

Maybe we could rename `serializeConfiguration()` to something like `getConfigurationTelemetry()`?

packages/rum-core/src/domain/plugins.ts
BenoitZugmeyer: Yes, but I agree the name isn't great, see https://github.com/DataDog/browser-sdk/pull/2809#discussion_r1639569044

packages/rum-core/src/boot/preStartRum.ts
BenoitZugmeyer: What `registerPlugins` would do? I don't want to anticipate too much :) even `callHook` is a bit too much. Maybe I could remove this 'hook' concept and just move the code here:

```ts
initConfiguration.plugins?.forEach(plugin => plugin?.onInit({ initConfiguration, publicApi }))
```

amortemousque (approved):

packages/rum-core/src/domain/plugins.ts
```diff
@@ -0,0 +1,27 @@
...
+type HookParameter<HookName extends HookNames> = Parameters<NonNullable<RumPlugin[HookName]>>[0]
+
+export function callHook<HookName extends HookNames>(
+  plugins: RumPlugin[] | undefined,
+  hookName: HookName,
+  parameter: HookParameter<HookName>
+) {
+  if (!plugins) {
+    return
+  }
```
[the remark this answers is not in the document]
BenoitZugmeyer: Sounds good, let's do it for now.

packages/rum-core/src/domain/configuration/configuration.ts
thomas-lebeau: oh, I didn't understood it was for telemetry. `getConfigurationTelemetry` sounds good, even though it's a bit weird to expose this to public plugins, but I guess that's fine for now and we can revisit if needed when we go GA with plugins

BenoitZugmeyer: > it's a bit weird to expose this to public plugins

There is no plan to support public plugins any time soon though. We'll go GA with the react plugin, but this doesn't imply supporting third-party plugins.

packages/rum-core/src/domain/plugins.ts
```diff
@@ -0,0 +1,27 @@
+import type { RumPublicApi } from '../boot/rumPublicApi'
+import type { RumInitConfiguration } from './configuration'
+
+export interface RumPlugin {
+  name: string
```
nulrich, line 5: **💭 thought:** ‏Maybe adding the plugin version, for telemetry and maybe checking compatibility with the SDK.

BenoitZugmeyer: For now installing a plugin with a different version than the SDK won't be supported. Plugins and SDK need to be updated at the same time.

N-Boutaib (approved):

PR #2817, opened 2024-06-18
Proposal: Add an option to control whether or not to mask data attribute

chuganzy:
## Motivation

This PR adds `maskDataAttributes` option to control whether or not to mask `data` attributes when `privacyLevel` is `mask`.

I see more use of data attributes for styling (CSS), and one of the reasons behind this is because headless accessible React components such as React Aria, Radix and Headless UI are getting more popular, which expose data attributes to enable customizing styles per state.

However, since DataDog masks all `data-` attributes when `privacyPolicy` is set to `mask`, the layout can break in Session Replay. 

Currently it is impossible to fix them unless we use lower privacy levels, or stop using data attributes in styles (which is not realistic under the situation)...

## Changes

As put in the title, to solve the challenge described above, this PR adds the option to change whether or not to mask data attributes (`true` by default which enables masking = the current behavior)

## Testing

As this is a proposal still, I have not added new tests yet, but I have updated the existing tests already to reflect the changes and `yarn test` passes locally now!

Please let me know how you feel and I can add some tests as needed!

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


BenoitZugmeyer:
Thank you for your contribution. I wonder if we need to go further and, similarly to other privacy levels, need to support this for parts of the DOM only (ex: `<div data-dd-privacy="allow-data-attributes">...</div>`). Also, maybe we'll need further customization like `maskDataAttributes: (name) => { return name === 'data-my-attribute' }`.

We'll discuss with the team and get back to you!

BenoitZugmeyer:
👋 While we agree this is a valid use-case, we need more reflexion before moving forward on the subject. Unfortunately we don't have the necessary bandwidth to work on this for now. In the meantime, feel free to use your Browser SDK fork on your project!

chuganzy:
@BenoitZugmeyer Got it, thank you for sharing. It is unfortunate to hear that but looking forward to the future updates!

---

Just to share what I have considered before adding this field, there is an allowed list already exist there and I thought adding some advanced options such as `(name: string) => name === 'something'` makes it _inconsistent_.. Also adding something like _pattern_ filter (as well as a heavy filter function) might affect the performance.. 

https://github.com/DataDog/browser-sdk/blob/1792105abbd1b45e62968afee2ab9e7e63c45530/packages/rum-core/src/domain/getSelectorFromElement.ts#L9

I thought maybe we can add a very simple option to just toggle it as the first step, and later on when these challenges are solved, add these _advanced_ filter options as a separated option, or as a union type, but this is jut my two cents!

PR #2818, opened 2024-06-18
♻️ [RUM-958] Use a performance observable instead of the lifecycle

amortemousque:
## Motivation

The performanceCollection has too many concerns and limitations:
- it forces us to set a central configuration, while sometimes we might want to have slightly different options (ex here for durationThreshold)
- we always listen for every performance entries while in some cases it might not be needed (ex for longtask)
- we cannot use takeRecords() because the observer is centralized and it would take records for all types

The goal of this PR is to be able to use separate performance observer for each entry type:
- The first iteration uses a performanceObservable for [performanceResourceTimings](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming)
- The next iteration will use it for other types of performance

## Changes

- Introduce a lightweight `performanceObservable` which removes the [performanceObservable fallback](https://github.com/DataDog/browser-sdk/blob/bafeb7aa15957c1aed2479748e61404e98bd3093/packages/rum-core/src/browser/performanceCollection.ts#L206) 
- Move `retrieveInitialDocumentResourceTiming` from `performanceCollection.ts` in `/domain/resource/retrieveInitialDocumentResourceTiming.ts`
- Use the `performanceObservable` to collect performance resource
- Use  the `performanceObservable` in `waitPageActivityEnd`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/resource/retrieveInitialDocumentResourceTiming.ts
```diff
@@ -0,0 +1,46 @@
+import type { RelativeTime } from '@datadog/browser-core'
+import { assign, runOnReadyState } from '@datadog/browser-core'
+import { supportPerformanceTimingEvent, RumPerformanceEntryType } from '../../browser/performanceObservable'
+import type { RumPerformanceResourceTiming } from '../../browser/performanceObservable'
+import type { RumConfiguration } from '../configuration'
+import { getDocumentTraceId } from '../tracing/getDocumentTraceId'
+import { FAKE_INITIAL_DOCUMENT, computeRelativePerformanceTiming } from './resourceUtils'
+
+export function retrieveInitialDocumentResourceTiming(
```
BenoitZugmeyer, line 9: **💭 thought:** ‏I think we can "clean" this up and don't fake a performance entry like this. Are you planning to do it when we'll migrate the `NAVIGATION` performance entry to the new `createPerformanceObserver`?

packages/rum-core/src/browser/performanceObservable.ts
```diff
@@ -0,0 +1,212 @@
...
+export function createPerformanceObservable<T extends RumPerformanceEntryType>(
+  configuration: RumConfiguration,
+  options: { type: T; buffered?: boolean; durationThreshold?: number }
+) {
+  return new Observable<Array<EntryTypeToReturnType[T]>>((observable) => {
+    const cleanupTasks: Array<() => void> = []
+
+    const observeEntries = (options?: PerformanceObserverInit & { durationThreshold?: number }) => {
+      const observer = new PerformanceObserver((entries) => {
+        const rumResourceEntries = handleRumPerformanceEntries(
```
BenoitZugmeyer, line 150: **🥜 nitpick:** ‏
```suggestion
        const rumPerformanceEntries = handleRumPerformanceEntries(
```

packages/rum-core/test/testSetupBuilder.ts
```diff
@@ -78,6 +81,7 @@ export function setup(): TestSetupBuilder {
   let sessionManager: RumSessionManager = createRumSessionManagerMock().setId('1234')
   const lifeCycle = new LifeCycle()
   const domMutationObservable = new Observable<void>()
+  const performanceResourceObservable = new Observable<RumPerformanceResourceTiming[]>()
```
BenoitZugmeyer, line 84: **💬 suggestion:** ‏I think we should migrate away of the `SetupBuilder`, because it's not so relevent anymore. Maybe you could create observables in the specs that need it instead?

amortemousque: Out of curiosity, what makes you feel it is not relevant anymore. (I also agree we should move away from it :)

BenoitZugmeyer: * It's designed to test high level modules like `startRum`, where it makes kind of sense, but used in lower level components where it brings more complexity than needed as it's not focused enough. Just creating the dependencies / running the function in the spec is generally enough. When this becomes too verbose, introducing a `xxxWithDefaults` function helps ([ex](https://github.com/DataDog/browser-sdk/blob/6102f81637a73a3d825bde5f2c896a67fa90d66a/packages/rum-core/src/domain/rumSessionManager.spec.ts#L217)). This makes the spec self contained and we avoid one indirection.

* Now that we have `registerCleanupTask`, it loses one of the main benefit/use-case, which was conditionally starting something in a test case knowing that it'll be cleaned up with `setup.cleanup()` anyway. For example, I'd like to improve `mockClock` by using `registerCleanupTask`, so we have a single way to mock the clock that can be used across the SDK, not only in RUM 

* We'll need to figure out how we could keep RUM event validation though

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -215,3 +217,20 @@ export function isLongDataUrl(url: string): boolean {
 export function sanitizeDataUrl(url: string): string {
   return `${url.match(DATA_URL_REGEX)![0]}[...]`
 }
+
+export type RelativePerformanceTiming = {
+  -readonly [key in keyof Omit<PerformanceTiming, 'toJSON'>]: RelativeTime
+}
+
+export function computeRelativePerformanceTiming() {
```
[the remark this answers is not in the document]
amortemousque: It is also used in https://github.com/DataDog/browser-sdk/blob/819fcbb088bbf75be8992a9c7f20c533922df8e1/packages/rum-core/src/browser/performanceCollection.ts#L116
This is why I haven't moved it yet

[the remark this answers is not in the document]
BenoitZugmeyer: Ah right I didn't see it was used two times there. Alright then.

packages/rum-core/src/browser/performanceObservable.ts
```diff
@@ -0,0 +1,212 @@
...
+      })
+      observer.observe(options)
+      cleanupTasks.push(() => observer.disconnect())
+    }
+
+    if (window.PerformanceObserver) {
+      try {
+        observeEntries(options)
+      } catch {
+        observeEntries({ entryTypes: [options.type] })
```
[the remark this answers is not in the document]
amortemousque: I agree, I removed the catch fallback. Let see what the telemetry as to say :)

packages/rum-core/test/testSetupBuilder.ts
amortemousque: I removed it from the builder

packages/rum-core/src/domain/resource/retrieveInitialDocumentResourceTiming.ts
amortemousque: I thought of cleaning it but the PR is already quite substantial. Like you mention, let's discuss it in the following NAVIGATION PR ;)

packages/rum-core/src/browser/performanceCollection.spec.ts
```diff
@@ -1,28 +1,55 @@
...
+  let entryCollectedCallback: jasmine.Spy
+  let setupBuilder: TestSetupBuilder
 
   beforeEach(() => {
-    configuration = {} as RumConfiguration
-    setup().beforeBuild(({ lifeCycle, configuration }) => {
-      startPerformanceCollection(lifeCycle, configuration)
+    if (!window.PerformanceObserver) {
+      pending('PerformanceObserver not supported')
+    }
```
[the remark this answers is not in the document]
amortemousque: I removed it

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -43,16 +47,32 @@ export function startResourceCollection(
...
   })
 
-  lifeCycle.subscribe(LifeCycleEventType.PERFORMANCE_ENTRIES_COLLECTED, (entries) => {
+  const performanceResourceSubscription = createPerformanceObservable(configuration, {
+    type: RumPerformanceEntryType.RESOURCE,
+    buffered: true,
+  }).subscribe((entries) => {
     for (const entry of entries) {
-      if (entry.entryType === RumPerformanceEntryType.RESOURCE && !isRequestKind(entry)) {
+      if (!isRequestKind(entry)) {
```
cy-moi, line 55: 👏 So nice to have this

cy-moi (approved):

packages/rum-core/test/emulate/mockPerformanceObserver.ts
```diff
@@ -16,9 +19,12 @@ export function mockPerformanceObserver() {
       disconnect() {
         instances.delete(instance)
       },
-      observe({ entryTypes, type }: PerformanceObserverInit) {
+      observe({ entryTypes, type, buffered }: PerformanceObserverInit) {
         instance.entryTypes = entryTypes || (type ? [type] : [])
         instances.add(instance)
+        if (buffered) {
+          notify(instance, bufferedEntries)
+        }
```
BenoitZugmeyer, line 27: **💬 suggestion:** ‏this is Safari behavior, but I'm not sure this is the standard behavior. We should at least state this as a comment. Maybe this could be done more explicitly by passing a `{ synchronousBufferedEntriesCallback: true }` flag to `mockPerformanceObserver`.

BenoitZugmeyer (approved):

packages/rum-core/test/emulate/mockPerformanceObserver.ts
amortemousque: Hum, the mock always triggers entries synchronously even for non buffered ones. To me, it's more by convenience to avoid playing with the clock each time we use `mockPerformanceObserver`. 
So I'm not sure it's a good idea to add a special behaviour for buffered entries and if we generalise it for any entries adding `synchronousBufferedEntriesCallback: true` everywhere will be a bit cumbersome. Wdyt?

BenoitZugmeyer: Let's not talk about async/sync, but whether the PerformanceObserver callback is called during or after `observe()`. My point is not to make everything asynchronous, just to make the `observe` default behavior closer to the standard.

Again, I'd be fine with just documenting this with a comment so we know why it's implemented like this.

bbrzoska:
After upgrade with this change, we're seeing a new PerformanceObserver instance is created after every user click / action. 
Is that the intention of this change? Can you confirm there's no performance impact due to creating and garbage collecting these objects unnecessarily?

BenoitZugmeyer:
This change should not have a performance impact. Creating a new PerformanceObserver is pretty cheap, and the memory associated to it should be negligeable compared to memory used by the rest of the SDK and/or your webapp.

bbrzoska:
We do see a noticeable increase in memory usage for RUM-enabled customers after upgrading the SDK from version 4 to version 5:

Screenshot 2024-07-25 at 11 21 58 AM

Dotted line is a week before comparison. You can see clearly when we rolled out the new SDK. The top yellow value is pc95.

PR #2824, opened 2024-06-20
⚗️✨ [RUM-4469] introduce an experimental React integration

BenoitZugmeyer:
## Motivation

Easily instrument a React application with the RUM SDK.

## Changes

This PR introduce a new `@datadog/browser-rum-react` package, with error tracking and react-router integration capabilities. More to come in the future.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end: We probably need e2e tests, I will see what I can do in a separate PR

---


alai97 (approved):

packages/rum-react/test/appendComponent.ts
```diff
@@ -0,0 +1,22 @@
...
+  const container = appendElement('<div></div>')
+  const root = createRoot(container, {
+    // Do nothing by default when an error occurs
+    onRecoverableError: noop,
+  })
+  flushSync(() => {
+    root.render(component)
+  })
+  registerCleanupTask(() => {
+    flushSync(() => {
```
amortemousque, line 17: **❓ question:** ‏Why is this flushSync required in the cleanupTask?

packages/rum-react/src/domain/reactRouterV6/useRoutes.ts
```diff
@@ -0,0 +1,19 @@
+import { useRef } from 'react'
+import { matchRoutes, useLocation, useRoutes as originalUseRoutes } from 'react-router-dom'
+import { startReactRouterView } from './startReactRouterView'
+
+export const useRoutes: typeof originalUseRoutes = (routes, locationArg) => {
```
amortemousque, line 5: **💬 suggestion:** ‏This implementation does not allow to have multiple instrumentation of `useRoute`. For instance if Sentry had the same approach the customer could not use both Sentry and Datadog (weird scenario but it is an example :). We could use a similar approach as Sentry using a HOF:
```ts
const useSentryRoutes = Sentry.wrapUseRoutes(useRoutes);
const useDatadogRoutes = Datadog.wrapUseRoutes(useSentryRoutes);
```

packages/rum-react/src/domain/reactRouterV6/registerRouter.ts
```diff
@@ -0,0 +1,16 @@
+import type { createBrowserRouter } from 'react-router-dom'
+import { startReactRouterView } from './startReactRouterView'
+
+type Router = ReturnType<typeof createBrowserRouter>
+
+export function registerRouter(router: Router) {
```
amortemousque, line 6: **💬 suggestion:** ‏Following my reasoning on the [useRoute instrumentation](https://github.com/DataDog/browser-sdk/pull/2824/files#r1648604979), we could use the same approach using a HOF to be consistent:
```ts
const datadogCreateBrowserRouter = Datadog.createBrowserRouter(createBrowserRouter)
```

packages/rum-react/src/domain/error/errorBoundary.ts
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import type { ErrorInfo } from 'react'
+import { addReactError } from './addReactError'
+
+interface Props {
+  fallback: Fallback
+  children: React.ReactNode
+}
+
+export type Fallback = (parameters: { error: Error; resetError: () => void }) => React.ReactNode
```
nulrich, line 10: **💭 thought:** ‏Maybe we could support also Component for fallback.

Miz85 (approved):
LGTM! I tested it on a custom test app that I created for the occasion and everything seems to work as expected!

Miz85:
I just realised that in my test app I'm getting a type error:

```sh
error TS2345: Argument of type 'import("/Users/nazim.saouli/test-react-app/node_modules/@remix-run/router/dist/router").Router' is not assignable to parameter of type 'import("/Users/nazim.saouli/go/src/github.com/DataDog/browser-sdk/node_modules/@remix-run/router/dist/router").Router'.
  Types of property '_internalActiveDeferreds' are incompatible.
    Type 'Map<string, import("/Users/nazim.saouli/test-react-app/node_modules/@remix-run/router/dist/utils").DeferredData>' is not assignable to type 'Map<string, import("/Users/nazim.saouli/go/src/github.com/DataDog/browser-sdk/node_modules/@remix-run/router/dist/utils").DeferredData>'.
      Type 'import("/Users/nazim.saouli/test-react-app/node_modules/@remix-run/router/dist/utils").DeferredData' is not assignable to type 'import("/Users/nazim.saouli/go/src/github.com/DataDog/browser-sdk/node_modules/@remix-run/router/dist/utils").DeferredData'.
        Types have separate declarations of a private property 'pendingKeysSet'.

13 registerRouter(router);
                  ~~~~~~
```

I'm still looking around to try and see if it's on me or if that's something that needs to be fixed in the react sdk itself.

packages/rum-react/test/appendComponent.ts
BenoitZugmeyer: Right, I was thinking `root.unmount()` was asynchronous, but after more tests it seems like it isn't so `flushSync` is probably not needed. I'll remove it.

packages/rum-react/src/domain/reactRouterV6/useRoutes.ts
BenoitZugmeyer: This is possible no?

```ts
import { useRoutes} from '@datadog/browser-rum-react/react-router-6'

const useSentryRoutes = Sentry.wrapUseRoutes(useRoutes)
```

BenoitZugmeyer: Indeed, this might be an issue if the customer uses another library that also provides a patched `useRoute` instead of providing a wrapping utility. I think this might be an edge case, and if a customer ever face that issue, we can always provide a wrapping utility (beside the already-wrapped `useRoutes`).

packages/rum-react/src/domain/reactRouterV6/registerRouter.ts
BenoitZugmeyer: The `Routes` component and `useRoutes` hook don't provide a way to cleanly subscribe to route changes. Given this constraint, wrapping them is kind of a requirement.

On the contrary, router instances have a way to subscribe to changes, so we don't have this constraint. That being said, for consistency, we could consider exposing a `createBrowserRouter` which wraps the original `createBrowserRouter` and just subscribe to it internally. I think that's interesting! Let me check how this could look like.

packages/rum-react/src/domain/error/errorBoundary.ts
BenoitZugmeyer: Done! As a function is also a Component, accepting any component works for both use cases (`fallback={() => ...}` and `fallbacx={Component}`)

packages/rum-react/src/domain/reactRouterV6/useRoutes.ts
amortemousque: Okay works for me 🙂

amortemousque (approved):

amortemousque (approved):

PR #2829, opened 2024-06-25
Caching Stylesheets serialisation

nulrich:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/record/serialization/serializeNode.ts
+++ b/packages/rum/src/domain/record/serialization/serializeNode.ts
@@ -22,12 +22,21 @@ import type {
 } from '../../../types'
 import { NodeType } from '../../../types'
 import { getSerializedNodeId, getValidTagName, setSerializedNodeId } from './serializationUtils'
-import type { SerializeOptions } from './serialization.types'
+import type { SerializeOptions, SerializedNodeCache } from './serialization.types'
 import { serializeStyleSheets } from './serializeStyleSheets'
 import { serializeAttributes } from './serializeAttributes'
 
-export function serializeNodeWithId(node: Node, options: SerializeOptions): SerializedNodeWithId | null {
-  const serializedNode = serializeNode(node, options)
+export function serializeNodeWithId(
+  node: Node,
+  options: SerializeOptions,
+  cache?: SerializedNodeCache
+): SerializedNodeWithId | null {
+  const serializedNodeId = getSerializedNodeId(node)
+  if (serializedNodeId && cache && cache.has(serializedNodeId)) {
+    return cache.get(serializedNodeId) as SerializedNodeWithId
+  }
+
+  const serializedNode = serializeNode(node, options, cache)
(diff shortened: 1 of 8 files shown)


packages/rum/src/domain/record/serialization/serializeNode.ts
```diff
@@ -48,45 +65,54 @@ export function generateNextId(): number {
...
 
-function serializeNode(node: Node, options: SerializeOptions): SerializedNode | undefined {
+function serializeNode(node: Node, options: SerializeOptions, cache?: SerializedNodeCache): SerializedNode | undefined {
   switch (node.nodeType) {
     case node.DOCUMENT_NODE:
-      return serializeDocumentNode(node as Document, options)
+      return serializeDocumentNode(node as Document, options, cache)
     case node.DOCUMENT_FRAGMENT_NODE:
-      return serializeDocumentFragmentNode(node as DocumentFragment, options)
+      return serializeDocumentFragmentNode(node as DocumentFragment, options, cache)
```
BenoitZugmeyer, line 88: **💬 suggestion:** ‏to avoid passing the `cache` everywhere, you could add it as a `SerializationContext` property (similarly to `shadowRootController`)

packages/rum/src/domain/record/serialization/serializeNode.ts
```diff
@@ -22,12 +22,21 @@ import type {
...
-  const serializedNode = serializeNode(node, options)
+export function serializeNodeWithId(
+  node: Node,
+  options: SerializeOptions,
+  cache?: SerializedNodeCache
+): SerializedNodeWithId | null {
+  const serializedNodeId = getSerializedNodeId(node)
+  if (serializedNodeId && cache && cache.has(serializedNodeId)) {
+    return cache.get(serializedNodeId) as SerializedNodeWithId
+  }
```
BenoitZugmeyer, line 37: **💭 thought:** ‏I don't think we should cache serialized nodes, but serialized stylesheets ([this function](https://github.com/DataDog/browser-sdk/blob/52ee5263cfdd87308c1083393bf9f3f7e194b4bc/packages/rum/src/domain/record/serialization/serializeStyleSheets.ts#L3)). The node has attributes that can change, and we don't invalidate the cache in that case. This might be trickier than expected.

Instead, the cache could be `Map<CSSStyleSheet, string>`

packages/rum/src/domain/record/serialization/serializeNode.ts
nulrich: Done!

packages/rum/src/domain/record/serialization/serializeNode.ts
nulrich: As discussed we'll keep caching the serialized node but listening on mutation observer to invalidate cache.

thomas-lebeau (approved):

PR #2832, opened 2024-06-26
👷‍♀️ Skip typechecks during unit tests

cy-moi:
## Motivation
There are misleading TS errors during tests due to the extension using a different `tsconfig` than the default one set in webpack base configuration, which is used by karma. 

## Changes
We use `transpileOnly` option for ts-loader in webpack to avoid type-checking during unit tests. We ignore the warnings caused by this setting. 

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/webpack.base.js
+++ b/webpack.base.js
@@ -6,7 +6,7 @@ const { buildEnvKeys, getBuildEnvValue } = require('./scripts/lib/build-env')
 
 const tsconfigPath = path.join(__dirname, 'tsconfig.webpack.json')
 
-module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables }) => ({
+module.exports = ({ entry, mode, filename, types, ignoreWarnings, keepBuildEnvVariables }) => ({
   entry,
   mode,
   output: {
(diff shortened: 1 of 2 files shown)


webpack.base.js
```diff
@@ -6,7 +6,7 @@ const { buildEnvKeys, getBuildEnvValue } = require('./scripts/lib/build-env')
 
 const tsconfigPath = path.join(__dirname, 'tsconfig.webpack.json')
 
-module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables }) => ({
+module.exports = ({ entry, mode, filename, types, ignoreWarnings, keepBuildEnvVariables }) => ({
```
BenoitZugmeyer, line 9: **💬 suggestion:** ‏with your solution this is not needed, since in `karma.base.conf.js` you don't use `webpackConfig.ignoreWarnings` but define `ignoreWarning` directly instead. You could remove it.

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

PR #2834, opened 2024-06-26
🎨 Instrument fetch and XHR before trying to init consent

N-Boutaib:
## Motivation
When RUM's initialisation is delayed because of consent not being granted initially, some libraries (such as Apollo Client) do store some methods aside to be re-used (mainly `window.fetch`), which leads to the un-instrumented method being used for http requests, even after consent being granted.

## Changes
Instrument `fetch` method on sdk initialisation. The instrumentation process is idempotent.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/boot/preStartLogs.ts
+++ b/packages/logs/src/boot/preStartLogs.ts
@@ -6,6 +6,7 @@ import {
   display,
   displayAlreadyInitializedError,
   initFeatureFlags,
+  initFetchObservable,
   noop,
   timeStampNow,
 } from '@datadog/browser-core'
@@ -66,6 +67,12 @@ export function createPreStartStrategy(
       }
 
       cachedConfiguration = configuration
+      // Instrumuent fetch to track network requests
+      // This is needed in case the consent is not granted and some cutsomer
+      // library (Apollo Client) is storing uninstrumented fetch to be used later
+      // The subscrption is needed so that the instrumentation process is completed
+      initFetchObservable().subscribe(noop)
+
       trackingConsentState.tryToInit(configuration.trackingConsent)
       tryStartLogs()
     },
--- a/packages/rum-core/src/boot/preStartRum.spec.ts
+++ b/packages/rum-core/src/boot/preStartRum.spec.ts
@@ -13,6 +13,7 @@ import {
   resetExperimentalFeatures,
   resetFetchObservable,
   resetXhrObservable,
+  isIE,
 } from '@datadog/browser-core'
 import type { Clock } from '@datadog/browser-core/test'
 import {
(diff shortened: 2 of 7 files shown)


packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -96,6 +98,10 @@ export function createPreStartStrategy(
       return
     }
 
+    // Instrumuent fetch and XHR to track network requests
+    initFetchObservable()
+    initXhrObservable(configuration)
```
BenoitZugmeyer, line 103: **💬 suggestion:** ‏move this a bit below in the function, after all the early returns

**💬 suggestion:** in the comment, ‏add more details on the motivation why we're doing this here instead of lazily when starting `requestCollection`

**💬 suggestion:** ‏add a unit test to ensure `fetch` and `xhr` are instrumented when running `init()` even without consent

thomas-lebeau (commented):
**🔨 warning:** ‏ I think we have the same issue with logs, don't we?


**❓ question:** ‏ What about other instrumented methods? It might be less common for them to be used that way, but we could have the same issue if that happens, no?

BenoitZugmeyer:
> **🔨 warning:** ‏ I think we have the same issue with logs, don't we?
> 
> **❓ question:** ‏ What about other instrumented methods? It might be less common for them to be used that way, but we could have the same issue if that happens, no?

For both subjects: I would wait for requests to do so before acting on it, as we don't want to over-instrument if we don't need to (in other words, it is better to not instrument if we never have user consent).

packages/rum-core/src/boot/preStartRum.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/unbound-method */
```
BenoitZugmeyer, line 1: **💬 suggestion:** ‏don't disable ESLint rules for the whole file. You can wrap your code between comments:
```ts
/* eslint-disable ... */
xxx
/* eslint-enable ... */
```

packages/rum-core/src/boot/preStartRum.spec.ts
```diff
@@ -37,6 +38,10 @@ const AUTO_CONFIGURATION = { ...DEFAULT_INIT_CONFIGURATION }
 const MANUAL_CONFIGURATION = { ...AUTO_CONFIGURATION, trackViewsManually: true }
 const FAKE_WORKER = {} as DeflateWorker
 const PUBLIC_API = {} as RumPublicApi
+const oldFetch = window.fetch
+const oldXHROpen = XMLHttpRequest.prototype.open
+const oldXHRSend = XMLHttpRequest.prototype.send
+const oldXHRAbort = XMLHttpRequest.prototype.abort
```
BenoitZugmeyer, line 44: **💬 suggestion:** ‏move this at the start of your test case.
Some other tests executing earlier could instrument those methods and not properly clean them up.

thomas-lebeau (approved):

BenoitZugmeyer (approved):

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -66,6 +67,12 @@ export function createPreStartStrategy(
       }
 
       cachedConfiguration = configuration
+      // Instrumuent fetch to track network requests
+      // This is needed in case the consent is not granted and some cutsomer
+      // library (Apollo Client) is storing uninstrumented fetch to be used later
+      // The subscrption is needed so that the instrumentation process is completed
+      initFetchObservable().subscribe(noop)
```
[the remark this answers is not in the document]
BenoitZugmeyer: We talked a bit about it during standup. I'll try to sum it up here.

On one side, we all agree that `.subscribe(noop)` is a bit suprising (but not necessarily hacky). On the other side, changing this would make things more complex in `fetchObservable.ts`. 

What you are suggesting is incorrect because if `stop` is called because there is no more subscriber, then the observable is subscribed again, we wouldn't re-instrument `fetch`. In practice this probably does not happen, but for a newcomer reading this code without context this might be seen as a bug.

So we would need another way to clean the instrumentation? Or maybe don't clean the instrumentation? It doesn't feel great either.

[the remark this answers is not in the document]
amortemousque: Well I think `.subscribe(noop)` does not convey the intention of what it does and raise more questions about the implementation for new newcomer reading this code.  
I put the `stop` in the observable only to avoid modifying too much the tests. We could have also extended the `resetFetchObservable` to reset the instrumentation. Since the removal of the instrumentation and the reset of the observable are only done for testing purposes, this approach would have been reasonable. I feel like we're choosing a confusing implementation just to avoid refactoring the tests.

PR #2853, opened 2024-07-09
🐛 Fix update view name api in context history

RomanGaignault:
## Motivation

The API 'update view name' doesn't update the last view in the context history. This mean next events such as action would not be linked to the view with the new name. 

## Changes

Added a subscribe to view_updated, so it update the last view from the context history with the correct name. 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/contexts/viewContexts.spec.ts
+++ b/packages/rum-core/src/domain/contexts/viewContexts.spec.ts
@@ -1,64 +1,15 @@
-import type { RelativeTime, Duration, TimeStamp } from '@datadog/browser-core'
+import type { RelativeTime } from '@datadog/browser-core'
 import { relativeToClocks, CLEAR_OLD_VALUES_INTERVAL } from '@datadog/browser-core'
 import type { TestSetupBuilder } from '../../../test'
 import { setup } from '../../../test'
 import { LifeCycleEventType } from '../lifeCycle'
 import type { ViewCreatedEvent, ViewEvent } from '../view/trackViews'
-import { ViewLoadingType } from '../../rawRumEvent.types'
 import type { ViewContexts } from './viewContexts'
 import { startViewContexts, VIEW_CONTEXT_TIME_OUT_DELAY } from './viewContexts'
 
 describe('viewContexts', () => {
   const FAKE_ID = 'fake'
   const startClocks = relativeToClocks(10 as RelativeTime)
-  const VIEW: ViewEvent = {
-    customTimings: {
-      bar: 20 as Duration,
-      foo: 10 as Duration,
-    },
-    documentVersion: 3,
-    duration: 100 as Duration,
-    eventCounts: {
-      errorCount: 10,
-      longTaskCount: 10,
-      resourceCount: 10,
-      actionCount: 10,
-      frustrationCount: 10,
-    },
-    id: 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee',
-    name: 'Fake Name',
-    isActive: false,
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/domain/contexts/viewContexts.ts
```diff
@@ -30,6 +30,13 @@ export function startViewContexts(lifeCycle: LifeCycle): ViewContexts {
     viewContextHistory.closeActive(endClocks.relative)
   })
 
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (data: ViewEvent) => {
+    const currentView = viewContextHistory.find()
```
amortemousque, line 34: **💬 suggestion:** ‏I don't think two view can be updated in parallel but just in case:
```suggestion
    const currentView = viewContextHistory.find(d(data.startClocks.relative))
```

packages/rum-core/src/domain/contexts/viewContexts.ts
```diff
@@ -30,6 +30,13 @@ export function startViewContexts(lifeCycle: LifeCycle): ViewContexts {
     viewContextHistory.closeActive(endClocks.relative)
   })
 
+  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (data: ViewEvent) => {
```
amortemousque, line 33: **💬 suggestion:** ‏It could clearer with an explicit variable name 🙂:
```suggestion
  lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, (viewUpdate: ViewEvent) => {
```

packages/rum-core/src/domain/contexts/viewContexts.spec.ts
```diff
@@ -1,15 +1,64 @@
...
-import type { ViewCreatedEvent } from '../view/trackViews'
+import type { ViewCreatedEvent, ViewEvent } from '../view/trackViews'
+import { ViewLoadingType } from '../../rawRumEvent.types'
 import type { ViewContexts } from './viewContexts'
 import { startViewContexts, VIEW_CONTEXT_TIME_OUT_DELAY } from './viewContexts'
 
 describe('viewContexts', () => {
   const FAKE_ID = 'fake'
   const startClocks = relativeToClocks(10 as RelativeTime)
+  const VIEW: ViewEvent = {
```
amortemousque, line 14: **💬 suggestion:** ‏You probably don't need a full ViewEvent. Something like this should be enough:
```ts
      lifeCycle.notify(LifeCycleEventType.VIEW_UPDATED, {
        startClocks,
        name: 'Fake Name',
      } as ViewEvent)
```

packages/rum-core/src/domain/contexts/viewContexts.spec.ts
```diff
@@ -105,6 +154,22 @@ describe('viewContexts', () => {
...
+    it('should update the view name for the current context', () => {
+      const { lifeCycle } = setupBuilder.build()
+      lifeCycle.notify(
+        LifeCycleEventType.BEFORE_VIEW_CREATED,
+        buildViewCreatedEvent({
+          id: '0',
+          name: 'foo',
+          startClocks: relativeToClocks(10 as RelativeTime),
+          service: 'test',
+          version: '1',
```
amortemousque, line 167: **❓ question:** ‏Why do you override all these attribute? Could `buildViewCreatedEvent({ name: 'foo' })` be enough?

N-Boutaib (approved):

amortemousque (approved):

PR #2860, opened 2024-07-11
🐛 Handle non-object response and error

nulrich:
## Motivation

We have a lot of telemetry errors concerning Fetch error that are not an Error object but a string or undefined.

## Changes

- Handling the non-object Fetch Error and Response
- Supporting abort status from AbortController

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

### DIFF
--- a/packages/core/src/browser/fetchObservable.spec.ts
+++ b/packages/core/src/browser/fetchObservable.spec.ts
@@ -82,6 +82,23 @@ describe('fetch proxy', () => {
     })
   })
 
+  it('should track fetch aborted by AbortController', (done) => {
+    const controller = new AbortController()
+    void fetch(FAKE_URL, { signal: controller.signal })
+    controller.abort('AbortError')
+
+    mockFetchManager.whenAllComplete(() => {
+      const request = requests[0]
+      expect(request.method).toEqual('GET')
+      expect(request.url).toEqual(FAKE_URL)
+      expect(request.status).toEqual(0)
+      expect(request.isAborted).toBe(true)
+      expect(request.error).toEqual(controller.signal.reason)
+      expect(request.handlingStack).toBeDefined()
+      done()
+    })
+  })
+
   it('should track opaque fetch', (done) => {
     // https://fetch.spec.whatwg.org/#concept-filtered-response-opaque
     fetch(FAKE_URL).resolveWith({ status: 0, type: 'opaque' })
--- a/packages/core/src/browser/fetchObservable.spec.ts
+++ b/packages/core/src/browser/fetchObservable.spec.ts
@@ -162,7 +162,7 @@ describe('fetch proxy', () => {
     })
   })
 
-  it('should keep promise resolved behavior', (done) => {
+  it('should keep promise resolved behavior for Response', (done) => {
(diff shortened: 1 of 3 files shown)


cy-moi (approved):

RomanGaignault (approved):

packages/core/src/browser/fetchObservable.ts
```diff
@@ -96,21 +97,31 @@ function afterSend(
...
+    monitor((response) => {
+      reportFetch({
+        response,
+        responseType: response.type,
+        status: response.status,
+        isAborted: false,
+      })
+    }),
+    monitor((error: Error) => {
+      const isAborted =
```
amortemousque, line 118: **🥜 nitpick:** ‏You could inline it
```ts
    reportFetch({
        status: 0,
        isAborted:
          context.init?.signal?.aborted || (error instanceof DOMException && error.code === DOMException.ABORT_ERR),
        error,
      })
```

packages/core/src/browser/fetchObservable.spec.ts
amortemousque, line 1: It could be nice to have a test with [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) :)

amortemousque:
**👏 praise:** ‏Very nice simplification!!

packages/core/src/browser/fetchObservable.spec.ts
nulrich: Done

packages/core/src/browser/fetchObservable.ts
nulrich: Done

packages/core/test/emulate/mockFetch.ts
```diff
@@ -14,7 +14,7 @@ export function mockFetch() {
     }
   }
 
-  window.fetch = (() => {
+  window.fetch = ((...[, init]) => {
```
amortemousque, line 17: **❓ question:** ‏Just out of curiosity, do you know how this is transpiled in ES5? :)

amortemousque (approved):

packages/core/test/emulate/mockFetch.ts
BenoitZugmeyer: https://www.typescriptlang.org/play/?target=0#code/O4SwdgJg9sB0BmBTALgYwBYAIC8mAUesRA2gDSbgjIC6AlDgHyYDeAvrUA

Though, this syntax is a bit surprising :D 

Suggestion:

```ts
window.fetch = (_input, init) => {
  ..
}
```

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

packages/core/test/emulate/mockFetch.ts
thomas-lebeau: It makes sense when you think about it, but I prefer benoit's suggestion (so I don't have to think about it 😄)

Or event `(input, init) => {` with [args-after-used](https://eslint.org/docs/latest/rules/no-unused-vars#args-after-used)

PR #2875, opened 2024-07-19
⚡ Performance Friday Reduce Bundle Size 

RomanGaignault:
## Motivation

Avoid using useless ?. to reduce bundle size
While those operators are quite nice additions to ES, they can produce an unexpected amount of code. Example:
	f?.stop()
	// bundled as (30 bytes):
	null===f||void 0===f||f.stop()
	
## Changes

Instead, using standard “boring” syntax is usually generating less code:
	if (f) f.stop()
	// bundled as (11 bytes):
	f&&f.stop()

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/serialisation/sanitize.ts
+++ b/packages/core/src/tools/serialisation/sanitize.ts
@@ -59,7 +59,8 @@ export function sanitize(source: unknown, maxCharacterCount = SANITIZE_DEFAULT_M
     containerQueue,
     visitedObjectsWithPath
   )
-  let accumulatedCharacterCount = JSON.stringify(sanitizedData) ? JSON.stringify(sanitizedData).length : 0
+  const serializedSanitizedData = JSON.stringify(sanitizedData)
+  let accumulatedCharacterCount = serializedSanitizedData ? serializedSanitizedData.length : 0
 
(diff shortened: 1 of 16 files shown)


packages/core/src/tools/serialisation/sanitize.ts
```diff
@@ -59,7 +59,8 @@ export function sanitize(source: unknown, maxCharacterCount = SANITIZE_DEFAULT_M
     containerQueue,
     visitedObjectsWithPath
   )
-  let accumulatedCharacterCount = JSON.stringify(sanitizedData)?.length || 0
+  let accumulatedCharacterCount = JSON.stringify(sanitizedData) ? JSON.stringify(sanitizedData).length : 0
```
BenoitZugmeyer, line 62: **💬 suggestion:** ‏don't call JSON.stringify twice. Ex:

```suggestion
  let serializedSanitizedData = JSON.stringify(sanitizedData)
  let accumulatedCharacterCount = serializedSanitizedData ? serializedSanitizedData.length : 0
```

BenoitZugmeyer (approved):

N-Boutaib (approved):

PR #2884, opened 2024-07-22
🧪 Update browser matrix for tests

thomas-lebeau:
## Motivation

Until now we're testing only on somewhat recent version off different browsers. However, as we're defining a new browser support matrix I thought it would be better to test on the actual minimum version of the browsers we're supporting.

This PR only focus on unit tests. I'll investigate separately for e2e tests as it's a bit more complicated.

## Changes

- separate unit and e2e browser matrix
- use new browser matrix for unit tests.
- properly check feature availability before running unit tests
- _refactor_ `onunhandledrejection` to handle the case where it's not supported (Firefox 67 & 68)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/core/src/domain/error/trackRuntimeError.spec.ts
+++ b/packages/core/src/domain/error/trackRuntimeError.spec.ts
@@ -1,4 +1,4 @@
-import { disableJasmineUncaughtExceptionTracking, collectAsyncCalls } from '../../../test'
+import { disableJasmineUncaughtExceptionTracking, collectAsyncCalls, registerCleanupTask } from '../../../test'
 import { Observable } from '../../tools/observable'
 import { isIE } from '../../tools/utils/browserDetection'
 import type { UnhandledErrorCallback } from './trackRuntimeError'
@@ -10,10 +10,7 @@ describe('trackRuntimeError', () => {
 
   let originalOnErrorHandler: OnErrorEventHandler
   let onErrorSpy: jasmine.Spy
-
-  let originalOnUnhandledRejectionHandler: Window['onunhandledrejection']
-  let onUnhandledrejectionSpy: jasmine.Spy
-
+  let onUnhandledrejectionSpy: jasmine.Spy | null
   let notifyError: jasmine.Spy
   let stopRuntimeErrorTracking: () => void
 
@@ -21,11 +18,7 @@ describe('trackRuntimeError', () => {
     originalOnErrorHandler = window.onerror
     onErrorSpy = jasmine.createSpy()
     window.onerror = onErrorSpy
-
-    originalOnUnhandledRejectionHandler = window.onunhandledrejection
-    onUnhandledrejectionSpy = jasmine.createSpy()
-    window.onunhandledrejection = onUnhandledrejectionSpy
-
+    onUnhandledrejectionSpy = setupOnUnhandledrejectionSpy()
     notifyError = jasmine.createSpy()
     const errorObservable = new Observable<RawError>()
     errorObservable.subscribe((e: RawError) => notifyError(e) as void)
@@ -35,7 +28,6 @@ describe('trackRuntimeError', () => {
   afterEach(() => {
     stopRuntimeErrorTracking()
     window.onerror = originalOnErrorHandler
-    window.onunhandledrejection = originalOnUnhandledRejectionHandler
   })
 
   it('should collect unhandled error', (done) => {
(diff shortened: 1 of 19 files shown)


packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
```diff
@@ -183,17 +183,6 @@ describe('computeXhrResponseData', () => {
...
-
-  beforeEach(() => {
-    onunhandledrejectionSpy = jasmine.createSpy()
-    window.onunhandledrejection = onunhandledrejectionSpy
-
-    registerCleanupTask(() => {
-      window.onunhandledrejection = null
-    })
-  })
-
```
thomas-lebeau, line 196: unused

packages/rum-core/src/domain/requestCollection.spec.ts
```diff
@@ -44,14 +44,10 @@ describe('collect fetch', () => {
     ;({ stop: stopFetchTracking } = trackFetch(lifeCycle, configuration, tracerStub as Tracer))
 
     fetch = window.fetch as MockFetch
-    window.onunhandledrejection = (ev: PromiseRejectionEvent) => {
-      throw new Error(`unhandled rejected promise \n    ${ev.reason as string}`)
-    }
```
thomas-lebeau, line 49: unused

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -138,5 +137,5 @@ function slidingSessionWindow() {
  * Check whether `layout-shift` is supported by the browser.
  */
 export function isLayoutShiftSupported() {
-  return supportPerformanceTimingEvent(RumPerformanceEntryType.LAYOUT_SHIFT)
+  return supportPerformanceTimingEvent(RumPerformanceEntryType.LAYOUT_SHIFT) && 'WeakRef' in window
```
thomas-lebeau, line 140: We used `WeakRef` to avoid memory leak when tracking CLS vital and assumed that browsers supporting layout-shift where also supporting WeakRef. Unfortunately that's not true and there is a few browsers versions that does support layout shift but not WeakRef. Unfortunately we were never testing on such browsers :( 

This bug affect the following browsers:
- Chrome 77 to 83
- Edge 79 to 83
- Opera 64 to 69

packages/core/src/domain/error/trackRuntimeError.spec.ts
```diff
@@ -315,3 +310,20 @@ describe('instrumentUnhandledRejection', () => {
     expect(stack).toBeDefined()
   })
 })
+
+function setupOnUnhandledrejectionSpy() {
+  if (!('onunhandledrejection' in window)) {
+    return null
+  }
```
N-Boutaib, line 317: 💭 **Thought**: I think it would be clearer to have this method always return a `jasmine.Spy`, and have a separated check in the `beforeEach`s. wdyt ?

packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
N-Boutaib, line 1: 👏 **Praise**: Thanks for the cleanup

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
N-Boutaib: Not sure how this is communicated, but it's worth mentioning in our documentation I guess

thomas-lebeau: I agree, although currently there is a lot of gaps on how we document the [features supported by browser](https://github.com/DataDog/browser-sdk/blob/main/packages/rum/BROWSER_SUPPORT.md)s and does not includes version.

My RFC around dropping IE11 proposes to remediates to this issue:
> Additionally we can update the feature support by browser to also include specific version numbers.

N-Boutaib (approved):

test/browsers.unit.conf.js
BenoitZugmeyer, line 1: **Suggestion**:
* move `browsers.e2e.conf.js` to `e2e/browsers.conf.js`
* move `browsers.unit.conf.js` to `unit/browsers.conf.js`
* remove unnecessary entries from browsers used by e2e tests (see https://github.com/DataDog/browser-sdk/blob/efe8b8cd27360cde61c9368993188fc42044781e/test/e2e/wdio.bs.conf.ts#L12-L19)

BenoitZugmeyer (approved):

PR #2887, opened 2024-07-23
⚡ [RUM-116] On view change, take the full snapshot asynchronously

rgaignault:
## Motivation

In the Datadog App, we are starting a new view when the URL changes (via the onChange react-router callback). But at this time, the old page component is still displayed, and we take a full snapshot of the ending view instead of the new view.

## Changes

A solution would be to take the fullsnapshot asynchronously. This would let a bit of time to the application UI to be updated. While waiting for the FS to be taken, we could ignore any emitted record, since we would know that a FS is coming to capture the whole state:

startView() is called

ignore any emitted record

[async] wait a bit

if stopSessionReplayRecording() was not called in the meantime

full snapshot is taken

start collecting emitted record again

For the asynchronous delay, we could even use requestIdleCallback to limit our impact on the application perceptible performance.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum/src/domain/record/startFullSnapshots.ts
```diff
@@ -65,17 +67,31 @@ export function startFullSnapshots(
...
+        cancelIdleCallback()
+      }
+      fullSnapshotPendingCallback()
+      cancelIdleCallback = requestIdleCallback(takeSubsequentFullSnapshot)
+    } else {
+      takeSubsequentFullSnapshot()
+    }
   })
 
   return {
```
amortemousque, line 97: **💬 suggestion:** ‏Make sure to call `cancelIdleCallback` in the `stop()` function
```ts
return {
  stop: () => {
    unsubscribe()
    cancelIdleCallback?.()
  }
}
```

packages/rum/src/domain/record/startFullSnapshots.ts
```diff
@@ -65,17 +67,31 @@ export function startFullSnapshots(
     return records
   }
 
-  fullSnapshotCallback(takeFullSnapshot())
+  fullSnapshotReadyCallback(takeFullSnapshot())
 
+  let cancelIdleCallback: (() => void) | undefined
   const { unsubscribe } = lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (view) => {
     flushMutations()
```
amortemousque, line 74: **❓ question:** ‏Is this `flushMutations()` necessary?

packages/rum/src/domain/record/record.ts
```diff
@@ -42,11 +42,15 @@ export function record(options: RecordOptions): RecordAPI {
...
   }
 
+  let isFullSnapshotPending = false
+
   const emitAndComputeStats = (record: BrowserRecord) => {
-    emit(record)
-    sendToExtension('record', { record })
-    const view = options.viewContexts.findView()!
-    replayStats.addRecord(view.id)
+    if (!isFullSnapshotPending) {
```
amortemousque, line 48: **❓ question:** ‏By discarding records during the fullsnapshot are we confident not to miss important ones like scroll, mouseInteraction, move?

packages/rum/src/browser/requestIdleCallBack.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+      cb(0)
+      return 123
+    })
+    cancelAnimationFrameSpy = spyOn(window, 'cancelAnimationFrame')
+    callback = jasmine.createSpy('callback')
+  })
+
+  afterEach(() => {
+    window.requestIdleCallback = originalRequestIdleCallback
+    window.cancelIdleCallback = originalCancelIdleCallback
```
amortemousque, line 21: **💬 suggestion:** ‏No need to do this since you use `spyOn`.

packages/rum/src/browser/requestIdleCallBack.spec.ts
amortemousque, line 1: You could add test when requestIdleCallback is supported and only mock what is necessary like this:
```ts

import { requestIdleCallback } from './requestIdleCallback'

describe('requestIdleCallback', () => {
  let callback: jasmine.Spy
  const originalRequestIdleCallback = window.requestIdleCallback

  beforeEach(() => {
    callback = jasmine.createSpy('callback')
  })

  afterEach(() => {
    if (originalRequestIdleCallback) {
      window.requestIdleCallback = originalRequestIdleCallback
    }
  })

  it('should use requestIdleCallback when supported', () => {
    if (!window.requestIdleCallback) {
      pending('requestIdleCallback not supported')
    }
    spyOn(window, 'requestIdleCallback').and.callFake((cb) => {
      cb({} as IdleDeadline)
      return 123
    })
    spyOn(window, 'cancelIdleCallback')

    const cancel = requestIdleCallback(callback)
    expect(window.requestIdleCallback).toHaveBeenCalled()
    cancel()
    expect(window.cancelIdleCallback).toHaveBeenCalledWith(123)
  })

  it('should use requestAnimationFrame when requestIdleCallback is not supported', () => {
    if (window.requestIdleCallback) {
      window.requestIdleCallback = undefined as any
    }
    spyOn(window, 'requestAnimationFrame').and.callFake((cb) => {
      cb(1)
      return 123
    })
    spyOn(window, 'cancelAnimationFrame')

    const cancel = requestIdleCallback(callback)
    expect(window.requestAnimationFrame).toHaveBeenCalled()
    cancel()
    expect(window.cancelAnimationFrame).toHaveBeenCalledWith(123)
  })
})

```

PR #2908, opened 2024-07-31
💥 remove Object.* Polyfills

thomas-lebeau:
## Motivation

Removed unnecessary polyfills for v6, resulting in a slight improvement in the bundle sizes.

## Changes

- Remove `objectValues` and `objectEntries` polyfills. However I'm keeping the wrapper method because their mangling result in smaller bundle size than using `Object.values()` every time it is used.
- Remove `Object.assign` polyfills in favor of object spread operator.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -161,7 +160,7 @@ type GenericBeforeSendCallback = (event: any, context?: any) => unknown
  */
 type ProxyFn = (options: { path: string; parameters: string }) => string
 
-interface ReplicaUserConfiguration {
+export interface ReplicaUserConfiguration {
```
thomas-lebeau, line 163: looks weird but it's because this some type is inferred as `ReplicaUserConfiguration` in another module (remoteConfiguration.ts) but typescript can't _name_ it.
```
@datadog/browser-rum-core: src/domain/configuration/remoteConfiguration.ts(22,17): error TS4058: Return type of exported function has or is using name 'ReplicaUserConfiguration' from external module "/Users/thomas.lebeau/go/src/github.com/DataDog/browser-sdk/packages/core/cjs/domain/configuration/configuration" but cannot be named.
```

packages/core/src/tools/utils/objectUtils.ts
```diff
@@ -1,7 +1,5 @@
-import { assign } from './polyfills'
-
 export function shallowClone<T>(object: T): T & Record<string, never> {
-  return assign({}, object)
+  return { ...object } as T & Record<string, never>
```
thomas-lebeau, line 2: Somehow, if `T` is `undefined`, then TS will infer `{...object}` as `undefined`, which does not match the JS that will return `{}`

see [TS playground](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVwM4AWAhgDYkQDuAwuSsADwAqAfABQQBGAVsBgFxxGASjgBvAFBw4UYDCRRcouADpVnHhjgBfOOK3jxaHHnhE4AXnzEylGjmCtUAE2AJMdJ0PEB6b1LgAegD8BobG8BwWVqTk1LQOoggQEAIA5BxEUKlaXr7+wQZAA)

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -315,23 +314,21 @@ function newClick(
       }
 
       const { resourceCount, errorCount, longTaskCount } = eventCountsSubscription.eventCounts
-      const clickAction: ClickAction = assign(
-        {
-          type: ActionType.CLICK as const,
```
thomas-lebeau, line 320: `'type' is specified more than once, so this usage will be overwritten. ts(2783)`

spreading `clickActionBase` will always overwrites `type`

nulrich (approved):

N-Boutaib (approved):

PR #2918, opened 2024-08-02
 ✨ [RUM-5712] add error.handling to logs 

BenoitZugmeyer:
**Please review commit by commit**

## Motivation

This PR implements error.handling to logs, similarly to RUM error events.

## Changes

This PR contains a bit of refactoring in order to unify how we define the `error` field in logs. The idea is to format all kind of errors as `RawError` objects, then use `RawError` objects in RUM (to create RUM Error events) and Logs (to define the `error` field). This `RawError` creation logic was already present in RUM, so I just had to move it in `core`. 

**Note:** Network errors are kind of special and only supported in logs, so I didn't use `RawError` there. Also, we can't be sure whether they are handled or not by the customer app, so to avoid any confusion, they don't have `error.handling`.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/error/trackConsoleError.ts
+++ b/packages/rum-core/src/domain/error/trackConsoleError.ts
@@ -1,18 +1,12 @@
 import type { Observable, RawError } from '@datadog/browser-core'
-import { clocksNow, ErrorHandling, initConsoleObservable, ErrorSource, ConsoleApiName } from '@datadog/browser-core'
+import { initConsoleObservable, ConsoleApiName } from '@datadog/browser-core'
 
 export function trackConsoleError(errorObservable: Observable<RawError>) {
-  const subscription = initConsoleObservable([ConsoleApiName.error]).subscribe((consoleError) =>
-    errorObservable.notify({
-      startClocks: clocksNow(),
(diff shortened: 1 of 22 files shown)


packages/rum-core/src/domain/error/trackConsoleError.ts
```diff
@@ -1,18 +1,12 @@
...
-      message: consoleError.message,
-      stack: consoleError.stack,
-      fingerprint: consoleError.fingerprint,
-      source: ErrorSource.CONSOLE,
-      handling: ErrorHandling.HANDLED,
-      handlingStack: consoleError.handlingStack,
-    })
-  )
+  const subscription = initConsoleObservable([ConsoleApiName.error]).subscribe((consoleLog) => {
+    if (consoleLog.error) {
```
thomas-lebeau, line 6: **❓ question:** ‏ Is this condition necessary as we init the console observable with `ConsoleApiName.error`?

thomas-lebeau (approved):

N-Boutaib (approved):

thomas-lebeau (approved):

PR #2938, opened 2024-08-14
👷 add canary deploy job for next major

thomas-lebeau:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


.gitlab-ci.yml
```diff
@@ -6,6 +6,7 @@ variables:
   CI_IMAGE: '$BUILD_STABLE_REGISTRY/ci/$APP:$CURRENT_CI_IMAGE'
   GIT_REPOSITORY: 'git@github.com:DataDog/browser-sdk.git'
   MAIN_BRANCH: 'main'
+  NEXT_MAJOR_BRANCH: 'v6'
```
amortemousque, line 9: **💬 suggestion:** ‏The next major version can be determined from the current version:
https://github.com/DataDog/browser-sdk/blob/fe822e0a4568b04ae85c7d4c756939fb8c117651/.gitlab/deploy-auto.yml#L14

amortemousque (approved):

.gitlab-ci.yml
thomas-lebeau: I think it makes sense, but two things:
- `NEXT_MAJOR_BRANCH` is about the branch name on github
- the `version` field is not yet updated in the v6 branch, I'll do that in another PR

When the version is updated in the v6 branch I think it will make sense to use that field for the filename of the uploaded files. But for the branch name I think I would prefer to keep an hardcoded value.

wdyt?

N-Boutaib (approved):

.gitlab-ci.yml
amortemousque: No strong opinion

PR #2939, opened 2024-08-14
✨ [RUM-5775] Make view local context writable

cy-moi:
## Motivation
Enable users to set the view specific context with existing apis:
- startView({...options, context})
- beforeSend: (e) => {}

## Changes
- Add customer context in view collection
- Add `USER_CUSTOMIZABLE_FIELD_PATHS` for view in event assembly
- Add feature flag `view_specific_context`

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -126,7 +132,9 @@ function processViewUpdate(
       toServerDuration as (duration: Duration) => ServerDuration
     )
   }
+
   return {
+    customerContext: isExperimentalFeatureEnabled(ExperimentalFeature.VIEW_SPECIFIC_CONTEXT) ? view.context : undefined,
```
nulrich, line 137: ❓ question: Do we need to check the FF? ‏The context should already be undefined from the tracking no?

cy-moi: Indeed, I removed it.

nulrich (approved):

packages/rum-core/src/domain/assembly.ts
amortemousque, line 1: **❓ question:** Shouldn't we also include the view context in the actions and errors context?
ex: `serverRumEvent.context = combine(commonContext.context, viewContext, customerContext)`
‏

cy-moi: Do we only do this for actions and errors? I added it for everything that is supposed to have a view field.

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -126,7 +126,9 @@ function processViewUpdate(
       toServerDuration as (duration: Duration) => ServerDuration
     )
   }
+
   return {
+    customerContext: view.context,
```
amortemousque, line 131: **💬 suggestion:** ‏You could simplify the logic a bit:
- remove the customerContext:
```suggestion
```
- Always merge `viewContext.context` in  assembly.ts: 
```ts 
serverRumEvent.context = combine(
     commonContext.context,
     viewContext.context,
     customerContext,
)
```

packages/rum-core/src/domain/assembly.ts
```diff
@@ -182,7 +184,11 @@ export function startRumAssembly(
         }
 
         const serverRumEvent = combine(rumContext as RumContext & Context, rawRumEvent) as RumEvent & Context
-        serverRumEvent.context = combine(commonContext.context, customerContext)
+        serverRumEvent.context = combine(
+          commonContext.context,
+          customerContext,
+          serverRumEvent.type !== 'view' ? viewContext.customerContext : undefined
+        )
```
amortemousque, line 191: **🔨 warning:** `viewContext.customerContext` would override `customerContext`. I think it is better to do the opposite to be consistent with global context.‏

packages/rum-core/src/domain/assembly.spec.ts
amortemousque, line 1: **💬 suggestion:** Could be nice to test the contexts override priority: Ex, for an action, the global context can be overridden by the view context that can be overridden by the customer context.‏

packages/rum-core/src/domain/assembly.ts
cy-moi: This should be fixed.

BenoitZugmeyer (approved):

packages/rum-core/src/domain/contexts/viewContexts.ts
```diff
@@ -45,6 +46,7 @@ export function startViewContexts(lifeCycle: LifeCycle): ViewContexts {
     return {
       service: view.service,
       version: view.version,
+      customerContext: view.context,
```
BenoitZugmeyer, line 49: **💬 suggestion:** ‏‏I'm not a big fan of the `customer` prefix here. Other properties of this interface also come from the customer, so it feels weird to specify `customer` for context.
 
I understand you are using `customerContext` instead of `context` to avoid having a `context` field on an interface that is already called `xxxContext`? I agree that's too many context! But I would rather rename `ViewContext` to something like `ViewHistoryEntry` (as I suggested in the RFC) instead of using `customerContext`.

cy-moi: I'm moving this renaming to another PR.

PR #2950, opened 2024-08-26
[RUM 5088] Reduce INP Null Target

cy-moi:
## Motivation
Save the selector in a map with timestamp as key to mitigate the null target situation.
More information in [RFC](https://docs.google.com/document/d/1FhJPEeyDPXSzpnk33lgrhSOSaSGLt-17Rpl-zVlQenE/edit?usp=sharing).

## Changes
After mitigation, about 75% of null target INP would have a selector.

## Testing

- [x] Local
- [x] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -76,6 +84,20 @@ export function trackInteractionToNextPaint(
...
+      if (
+        !interactionToNextPaintTargetSelector &&
+        isExperimentalFeatureEnabled(ExperimentalFeature.NULL_INP_TELEMETRY)
+      ) {
+        if (interactionSelectorMap.has(newInteraction.startTime)) {
+          interactionToNextPaintTargetSelector = interactionSelectorMap.get(newInteraction.startTime)
+          interactionSelectorMap.delete(newInteraction.startTime)
+        }
+      }
+      interactionSelectorMap.forEach((_, key) => {
```
amortemousque, line 96: **💬 suggestion:** ‏ You could isolate this logic in a dedicated function like `deleteOldInteractionSelectorByTimestamp`

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -61,6 +63,7 @@ type ClickActionIdHistory = ValueHistory<ClickAction['id']>
 // Maximum duration for click actions
 export const CLICK_ACTION_MAX_DURATION = 10 * ONE_SECOND
 export const ACTION_CONTEXT_TIME_OUT_DELAY = 5 * ONE_MINUTE // arbitrary
+export const interactionSelectorMap = new Map<number, string | undefined>() // key is timestamp
```
amortemousque, line 66: **💬 suggestion:** We could probably have a more explicit name. Also if key is a timestamp you could use TimeStamp type
```suggestion
export const interactionSelectorByTimestamp = new Map<TimeStamp, string | undefined>()
```

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -76,6 +84,20 @@ export function trackInteractionToNextPaint(
...
+        if (key > 10 * ONE_SECOND) {
+          interactionSelectorMap.delete(key)
+        }
+      })
```
BenoitZugmeyer, line 100: **💬 suggestion:** ‏Not all browsers support INP, so if we are clearing the map when an interaction is notified, it might leak memory on browsers not supporting it.

Move this in `trackClickAction`? It could be:

```ts
function addSelectorToInteractionMap(timeStamp, selector) {
  interactionSelectorMap.set(timeStamp, selector)
  // then clear the map
}

```

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -76,6 +84,20 @@ export function trackInteractionToNextPaint(
...
+        !interactionToNextPaintTargetSelector &&
+        isExperimentalFeatureEnabled(ExperimentalFeature.NULL_INP_TELEMETRY)
+      ) {
+        if (interactionSelectorMap.has(newInteraction.startTime)) {
+          interactionToNextPaintTargetSelector = interactionSelectorMap.get(newInteraction.startTime)
+          interactionSelectorMap.delete(newInteraction.startTime)
+        }
+      }
+      interactionSelectorMap.forEach((_, key) => {
+        if (key > 10 * ONE_SECOND) {
```
BenoitZugmeyer, line 97: **🔨 warning:** ‏The key being a relative time, with this condition we'll keep only entries that occur within the 10 first seconds of the page. What you are looking for is:

```suggestion
        if (relativeNow() - key > 10 * ONE_SECOND) {
```

BenoitZugmeyer (commented):
Could you add unit tests covering your changes?

packages/rum-core/src/domain/action/trackClickActions.ts
cy-moi: Should be fixed. Though I used RelativeTime suggested below for this type, I can see there has been confusion around this and `TimeStamp` used in legacy cases.

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
cy-moi: Good Catch! 😭

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
cy-moi: I renamed it to `deleteOutdatedInteractionSelectors`. 'Interaction' feels a bit lengthy, maybe we can just call it `deleteOutdatedSelectors`. Wdyt?

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
```diff
@@ -76,6 +84,20 @@ export function trackInteractionToNextPaint(
       } else {
         interactionToNextPaintTargetSelector = undefined
       }
+      if (
```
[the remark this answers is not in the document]
cy-moi: Yes, I agree. I also removed the telemetry feature flag as we no longer need to experiment with it.

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
amortemousque: I don't see this change, normal?

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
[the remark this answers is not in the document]
cy-moi: 👋 I changed this logic again as we need to cover the null target cases.

packages/rum-core/src/domain/action/interactionSelectorCache.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  set: (relativeTimestamp: RelativeTime, selector: string) => {
+    cache.set(relativeTimestamp, selector)
+    cache.forEach((_, relativeTimestamp) => {
+      if (elapsed(relativeTimestamp, relativeNow()) > CLICK_ACTION_MAX_DURATION) {
+        cache.delete(relativeTimestamp)
+      }
+    })
+  },
+}
```
amortemousque, line 23: **💬 suggestion:** ‏
For stateful module, we tend to a export get and set function like in replayStats.ts. So it would look like:

```suggestion
export const CLICK_ACTION_MAX_DURATION = 10 * ONE_SECOND
const interactionSelectorCache = new Map<RelativeTime, string>()

export function getInteractionSelector(relativeTimestamp: RelativeTime) {
  const selector = interactionSelectorCache.get(relativeTimestamp)
  interactionSelectorCache.delete(relativeTimestamp)
  return selector
}

export function setInteractionSelector(relativeTimestamp: RelativeTime, selector: string) {
  interactionSelectorCache.set(relativeTimestamp, selector)
  clearOutdatedInteractionSelectors()
}

function clearOutdatedInteractionSelectors() {
  interactionSelectorCache.forEach((_, relativeTimestamp) => {
    if (elapsed(relativeTimestamp, relativeNow()) > CLICK_ACTION_MAX_DURATION) {
      interactionSelectorCache.delete(relativeTimestamp)
    }
  })
}

```

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -176,6 +174,13 @@ function startClickAction(
   const click = newClick(lifeCycle, history, getUserActivity, clickActionBase, startEvent)
   appendClickToClickChain(click)
 
+  if (clickActionBase.target) {
+    const { selector } = clickActionBase.target
+    if (selector) {
+      interactionSelectorCache.set(startEvent.timeStamp, selector)
+    }
+  }
+
```
amortemousque, line 183: Can be simplified
```suggestion
  if (clickActionBase?.target.selector) {
      interactionSelectorCache.set(startEvent.timeStamp, selector)
  }

```

packages/rum-core/src/domain/view/viewMetrics/trackInteractionToNextPaint.ts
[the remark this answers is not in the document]
amortemousque: Nice!

packages/rum-core/src/domain/action/interactionSelectorCache.ts
cy-moi: `clearOutdatedInteractionSelectors` only used once and I think `set` is a bit misleading if we also clear. So I renamed the `set` function to be `updateInteractionSelector` and included both setting and clearing logic directly.

packages/rum-core/src/domain/action/trackClickActions.spec.ts
```diff
@@ -425,15 +425,44 @@ describe('trackClickActions', () => {
     })
   })
 
+  describe('interactionSelectorMap', () => {
+    it('should add pointer down to the map', () => {
+      startClickActionsTracking()
+      const setInteractionMapSpy = spyOn(Map.prototype, 'set').and.callThrough()
+      emulateClick({ activity: { on: 'pointerdown' } })
+      expect(setInteractionMapSpy).toHaveBeenCalled()
```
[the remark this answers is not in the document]
cy-moi: Thank you! The tests should be fixed with checking the map directly. It indeed is more effective! TIL

packages/rum-core/src/domain/action/interactionSelectorCache.ts
```diff
@@ -0,0 +1,21 @@
+import { elapsed, ONE_SECOND, relativeNow } from '@datadog/browser-core'
+import type { RelativeTime } from '@datadog/browser-core'
+
+// Maximum duration for click actions
+export const CLICK_ACTION_MAX_DURATION = 10 * ONE_SECOND
+export const interactionSelectorCache = new Map<RelativeTime, string>()
+
+export const getInteractionSelector = (relativeTimestamp: RelativeTime) => {
```
amortemousque, line 8: **🥜 nitpick:** ‏For consistency sake, you could use regular function (not arrow) :)

amortemousque (approved):

BenoitZugmeyer (approved):

PR #2958, opened 2024-08-30
🐛 [RUM-94] ignore performance resource timings with negative duration

BenoitZugmeyer:
## Motivation

Starting with Chrome 128, blob URLs are reported as PerformanceResourceTiming entries. Those entries have (always?) negative durations, which is unexpected. The intake rejects RUM Resource events with negative duration already, so sending them is useless and breaks our e2e tests when validating the events against the rum-events-format.

## Changes

Let's ignore resource entries with negative duration.

Note: I opened a [Chromium issue](https://issues.chromium.org/issues/363031537) about this, but in the meantime, there is absolutely no value in sending those entries to the intake.

This PR also rename a bunch of "PerformanceResourceTiming"-related function to have consistent naming.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/resourceUtils.ts
+++ b/packages/rum-core/src/domain/resource/resourceUtils.ts
@@ -135,6 +135,10 @@ export function computePerformanceResourceDetails(
   return details
 }
 
+export function hasValidResourceEntryDuration(entry: RumPerformanceResourceTiming) {
+  return entry.duration >= 0
+}
+
 export function hasValidResourceEntryTimings(entry: RumPerformanceResourceTiming) {
   if (isExperimentalFeatureEnabled(ExperimentalFeature.TOLERANT_RESOURCE_TIMINGS)) {
     return true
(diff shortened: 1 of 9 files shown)


cy-moi (approved):

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -89,7 +89,7 @@ export function computePerformanceResourceDuration(entry: RumPerformanceResource
 export function computePerformanceResourceDetails(
```
amortemousque, line 89: **💬 suggestion:** ‏The name is not consistent with the other functions. The main concept is "Resource Entry" (ex: computeResourceEntryType, computeResourceEntrySize) but here we talk about "Performance resource". 
Maybe something like 
```ts
export function computeResourceEntryTimings(
  entry: RumPerformanceResourceTiming
): PerformanceEntryTimings | undefined 
```

BenoitZugmeyer: Ah, I missed this!

IIRC `Details` was chosen instead of `Timings` to avoid the confusion with `PerformanceResourceTiming`. I would be enclined to keep `Details` (`computeResourceEntryDetails`). What do you think?

amortemousque (approved):

PR #2959, opened 2024-08-30
👷 don't print log messages twice in unit tests

BenoitZugmeyer:
## Motivation

`baseReporterDecorator` implements a bunch of methods that display messages in the console. Since we already have a reporter for that, messages are displayed twice.

## Changes

Remove `baseReporterDecorator` usage from our karma "skipped failed reporter" plugin.  Without `baseReporterDecorator`, `specSkipped` and `specFailure` "hooks" don't exist anymore, so let's use `onSpecComplete` instead.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


N-Boutaib (approved):
Thanks for this change, I was always wondering why two logs are printed 😅

PR #2964, opened 2024-09-03
🐛 [RUM-5785] fix missing navigation timings on Safari

BenoitZugmeyer:
## Motivation

We observed that very often, navigation timings (ex: `@view.load_event`, `@view.dom_complete`) and associated (ex: `@view.loading_time`) were missing, especially on Safari. After investigation, it turns out that Safari does not notify again the `PerformanceNavigationTiming` to `PerformanceObservers` after the page is loaded (contrary to Firefox and Chrome). So the first time the `PerformanceNavigationTiming` is notified, it's ignored because incomplete, but then it's never re-notified.

## Changes

Instead of relying on a PerformanceObserver to get the `PerformanceNavigationTiming` entry, let's wait for the load event to be over then pick the entry from the `window.performance` buffer.

Please review commit by commit to have a list of individual changes.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/view/viewMetrics/trackNavigationTimings.spec.ts
N-Boutaib, line 1: 🎉 **Praise**: Nice & clear tests

N-Boutaib (approved):

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -415,6 +412,7 @@ describe('view metrics', () => {
       }
       const { getViewUpdate, getViewUpdateCount, getViewCreateCount, startView } = viewTest
       startView()
+      clock.tick(0)
```
nulrich, line 415: **❓ question:** ‏Is it because of the implementation of `waitAfterLoadEvent` ?

BenoitZugmeyer: Yes! `trackViews` tests depend on `trackNavigationTimings` logic, so this requirement "spreads". This is the case in quite a few unit test specs: we generally test more than we should, and I'll come up with a plan to reduce this in the feature.

nulrich (approved):

PR #2965, opened 2024-09-03
:sparkles: Discard loading time when page is hidden

N-Boutaib:
## Motivation
When a view created in background, its loading time can be inflated due to the browser pausing the javascript execution, which leads to data inconsistencies. It's better to not send the loading time of those views, rather than sending an erroneous value.

## Changes
If the page is initially hidden, discard attaching `loading_time` to view event

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
@@ -20,13 +20,11 @@ export function trackLoadingTime(
   const firstHidden = trackFirstHidden(configuration)
 
   function invokeCallbackIfAllCandidatesAreReceived() {
-    if (
-      firstHidden.timeStamp !== 0 &&
-      !isWaitingForActivityLoadingTime &&
-      !isWaitingForLoadEvent &&
-      loadingTimeCandidates.length > 0
-    ) {
-      callback(Math.max(...loadingTimeCandidates) as Duration)
+    if (!isWaitingForActivityLoadingTime && !isWaitingForLoadEvent && loadingTimeCandidates.length > 0) {
+      const loadingTime = Math.max(...loadingTimeCandidates)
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
```diff
@@ -16,9 +17,15 @@ export function trackLoadingTime(
...
 
   function invokeCallbackIfAllCandidatesAreReceived() {
-    if (!isWaitingForActivityLoadingTime && !isWaitingForLoadEvent && loadingTimeCandidates.length > 0) {
+    if (
+      firstHidden.timeStamp !== 0 &&
+      !isWaitingForActivityLoadingTime &&
+      !isWaitingForLoadEvent &&
+      loadingTimeCandidates.length > 0
+    ) {
       callback(Math.max(...loadingTimeCandidates) as Duration)
```
amortemousque, line 29: **❓ question:** ‏Shouldn't we check the loading time against the firstHidden timestamp like we do with firstContentFullPaint?
Something like
```suggestion
     const loadingTime = Math.max(...loadingTimeCandidates)
      if (loadingTime < firstHidden.timeStamp) {
            callback(loadingTime as Duration)
      }
```

N-Boutaib: Indeed, that was my first approach which will work in this use-case, and also when loading a page then putting it in background before the loading is completed.

N-Boutaib: Will test this case and do the change if needed

BenoitZugmeyer (approved):

amortemousque (approved):

Issue #2956, opened 2024-08-29
Client side page views seem to no longer be recognized (script tag version)


corbin-mosher:
I am not 100% sure this is the correct place to report this issue, however the JS script version of this package (https://www.datadoghq-browser-agent.com/us1/v5/datadog-rum.js) is now only recording one page view per session for Nextjs React SPA client side routed pages.

This started for us at 7:37 PT, the same time the JS file is marked as last updated. It also seems to correlate with the time this package had a new version release.

Did something occur that broke client side history tracking? I see some commits in the release related to history API?

I can submit a support ticket through our DataDog rep if that is preferable.

corbin-mosher:
It turns out that not EVERY session is 1 page view. But instead just a dramatic drop in recordings of page views.

This shows how the `p70` of page views per session takes a big drop down at 7:37 PT Aug 29th.

![Screenshot 2024-08-30 at 11 12 42 AM](https://github.com/user-attachments/assets/02984e3e-90da-4f30-a469-6d6ed99164f4)

Same with the average:

![Screenshot 2024-08-30 at 11 14 31 AM](https://github.com/user-attachments/assets/578ecc6d-a611-403e-a291-b3bb8d6ffa57)

Where as the median holds stable:

![Screenshot 2024-08-30 at 11 15 09 AM](https://github.com/user-attachments/assets/ef6afe35-2097-4de0-8b38-1472745bdd07)

Was this expected with the changes? Was the information just very inaccurate prior to this update for long tail sessions?

thomas-lebeau:
Hi @corbin-mosher, Thanks for reporting this issue. I'm not able to reproduce, would you mind opening a support ticket through your DataDog rep?

corbin-mosher:
Of course. Thanks for the reply!

corbin-mosher:
This was resolved in version 5.26.0.

Issue #2943, opened 2024-08-20
History API is wrapped wrong


kamtschatka:
**Describe the bug**
The trackHistory function is calling `instrumentMethod(history, 'pushState')`: [Code](https://github.com/DataDog/browser-sdk/blob/6b3cfefd90ecb05a45b840f3447c4882b1811eba/packages/rum-core/src/browser/locationChangeObservable.ts#L36)
`history` does not actually have the `pushState` function, but `History.prototype` does.
This call causes `history` to get a `pushState` function assigned to it, causing the prototype to be ignored, which is causing issues for other vendors, who are also trying to wrap `pushState`.

Original `history`:
![image](https://github.com/user-attachments/assets/6709869e-40ea-4e9a-a0fa-557c9d8cf1ef)

Wrapped `history`:
![image](https://github.com/user-attachments/assets/ac3df8ef-c6f7-42de-99bc-5ed2102ef070)

**To Reproduce**
I am not 100% sure, as I am unfamiliar with DD, but I guess this happens simply by initializing DD?

**Expected behavior**
I believe the correct code would be `instrumentMethod(History.prototype, 'pushState')`. (same for replaceState)

Disclaimer: I work for Dynatrace and found this during a support case. I know we could work around this issue, but I do believe it is better to be fixed at the source. Let me know what you think.

amortemousque:
Hello @kamtschatka,
Good catch! Thanks for reporting it. I'll let you know when it's fixed.

amortemousque:
Hello @kamtschatka, the fix is available in version v5.24.0

kamtschatka:
awesome, thx a lot.

PR #2952, opened 2024-08-28
🎨 [RUM-5100] Move away from testbuilder in test files - Pt 3

N-Boutaib:
## Motivation
Removing the last occurrences of `TestBuilder` and improving spec files by removing casting and using unified default RumConfiguration.

## Changes
- Removing TestBuilder
- Using default RumConfig instead of `{} as RumConfiguration`

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #2968, opened 2024-09-03
🐛 Allow both history and History prototype to be patched by 3rd party

thomas-lebeau:
## Motivation

Since #2944, we broke the instrumentation for NextJs that is patching  `history.pushState` instead of `History.prototype.pushState`. This PR allow both way to be patched (even if incorrect)

## Changes

- wrap `history.pushState` if it is already patched that way, otherwise wrap `History.prototype.pushState`
- Refactor mockLocation because using `spyOn` is effectively the same as wrapping so this make testing difficult.

## Testing

I've added unit test, to cover both #2956 and #2943 

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/core/test/emulate/mockLocation.ts
BenoitZugmeyer, line 1: **Suggestion:** Because we don't mock the location anymore in `mockLocation`, maybe rename this file. (Don't put `buildLocation` into `testSetupBuilder.ts` as it will conflict with https://github.com/DataDog/browser-sdk/pull/2952)

amortemousque (approved):

PR #2984, opened 2024-09-11
✨ Add records caching logic and cache records if session is sampled out

N-Boutaib:
## Motivation
Implementation of Error Mode Replay recorder.

## Changes
- Initialise records caching module
- Start session replay recording as long as session is tracked (no matter what sample state)
- If session is sampled in for replay, start segments collection, otherwise forward records to cache
- When session replay is forced to start, start segment collection from cache

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/boot/startRecording.ts
+++ b/packages/rum/src/boot/startRecording.ts
@@ -1,12 +1,13 @@
 import type { RawError, HttpRequest, DeflateEncoder } from '@datadog/browser-core'
 import { createHttpRequest, addTelemetryDebug, canUseEventBridge } from '@datadog/browser-core'
 import type { LifeCycle, ViewHistory, RumConfiguration, RumSessionManager } from '@datadog/browser-rum-core'
-import { LifeCycleEventType } from '@datadog/browser-rum-core'
+import { LifeCycleEventType, SessionReplayState } from '@datadog/browser-rum-core'
 
 import { record } from '../domain/record'
 import { startSegmentCollection, SEGMENT_BYTES_LIMIT } from '../domain/segmentCollection'
 import type { BrowserRecord } from '../types'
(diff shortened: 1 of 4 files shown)


packages/rum/src/boot/startRecording.ts
amortemousque, line 1: **💬 suggestion:** The logic is a bit complicated to follow due to the indirections and addRecord overrides. I think it could be a bit simplified like this:

```ts
  const flushCachedRecords: () => void = noop
  let addRecord: (record: BrowserRecord) => void

  if (!canUseEventBridge()) {
    const session = sessionManager.findTrackedSession()!
    if (session.sessionReplay === SessionReplayState.OFF) {
      ;({ addRecord } = initSegmentCollection())
    } else {
      ;({ addRecord, flushCachedRecords } = startRecordsCaching())
    }
  } else {
    ;({ addRecord } = startRecordBridge(viewHistory))
  }

  const { stop: stopRecording } = record({
    emit: addRecord,
    configuration,
    lifeCycle,
    viewHistory,
  })
  cleanupTasks.push(stopRecording)

  return {
    flushCachedRecords: () => {
      ;({ addRecord } = initSegmentCollection())
      flushCachedRecords(addRecord)
    },
```
‏

N-Boutaib: Sadly, this approach doesn't work. I did test it as an initial approach but it wasn't working as expected. In fact, the records collected are created using the initial `addRecord`, so changing that afterword doesn't impact `emit` method used in records collection. Thus the need of having a state `shouldCache` and conditionally change the method used.

amortemousque: Just have to do:
```ts
    emit: (record) => addRecord(record) 
```

PR #2985, opened 2024-09-11
💥 [RUM 6075] Save anonymous id in session cookie

cy-moi:
## Motivation
Test saving anonymous id in session cookie on org2. 

## Changes
Add anonymous id to session cookie behind a feature flag.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/browser/cookie.ts
```diff
@@ -9,14 +11,18 @@ export interface CookieOptions {
...
-  const expires = `expires=${date.toUTCString()}`
+  let expires = `expires=${date.toUTCString()}`
   const sameSite = options && options.crossSite ? 'none' : 'strict'
   const domain = options && options.domain ? `;domain=${options.domain}` : ''
   const secure = options && options.secure ? ';secure' : ''
   const partitioned = options && options.partitioned ? ';partitioned' : ''
+  // we cannot read from cookie here because it will affect cookie locking
+  if (isExperimentalFeatureEnabled(ExperimentalFeature.ANONYMOUS_USER_TRACKING) && name === SESSION_STORE_KEY) {
+    expires = `max-age=${ONE_YEAR}`
+  }
```
BenoitZugmeyer, line 25: **💬 suggestion:** you could just provide a bigger `expireDelay` argument when setting the session cookie instead of having this special case

packages/core/src/domain/user/user.ts
```diff
@@ -30,3 +33,41 @@ export function checkUser(newUser: User): boolean {
...
+
+export function retrieveAnonymousId(sessionStoreStrategyType: string): string {
+  let anonymousId = getAnonymousIdFromStorage()
+
+  if (!anonymousId) {
+    anonymousId = generateAnonymousId()
+    setAnonymousIdInStorage(sessionStoreStrategyType, anonymousId)
+  }
+  return anonymousId
+}
```
BenoitZugmeyer, line 69: **💬 suggestion:** ‏I know the current SessionState/SessionStore etc. is a bit hard to grasp, but I really think we should keep a single way of writing and reading the session cookie. Let's look at this together?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -189,6 +190,14 @@ export function startRumAssembly(
             session.sessionReplay === SessionReplayState.SAMPLED
         }
 
+        if (
+          // TODO: remove ff and should always add anonymous user id
+          isExperimentalFeatureEnabled(ExperimentalFeature.ANONYMOUS_USER_TRACKING) &&
+          !commonContext.user.anonymous_id
+        ) {
+          const { type } = configuration.sessionStoreStrategyType ?? { type: 'LocalStorage' }
+          commonContext.user.anonymous_id = retrieveAnonymousId(type)
```
BenoitZugmeyer, line 199: **💬 suggestion:** ‏a bit related to my previous comment, but we should probably be able to do something like:
```suggestion
          commonContext.user.anonymous_id = session.getAnonymousUserId()
```

packages/core/src/domain/user/user.ts
cy-moi: Yes, thank you! I agree on keeping a single way. But I don't think we want to tie this anonymous id logic with session state logics, as the latter one needs to be refreshed oftenly. I tried to do it directly from the `get/set` cookie functions which did not work.

packages/core/src/domain/session/storeStrategies/sessionInLocalStorage.ts
```diff
@@ -34,9 +34,10 @@ function persistInLocalStorage(sessionState: SessionState) {
 
 function retrieveSessionFromLocalStorage(): SessionState {
   const sessionString = localStorage.getItem(SESSION_STORE_KEY)
-  return toSessionState(sessionString)
+  const sessionState = toSessionState(sessionString)
```
nulrich, line 37: **💬 suggestion:** ‏Remove the `const`.

nulrich (approved):

PR #3014, opened 2024-09-18
Init mapping ids to serialised nodes

N-Boutaib:
## Motivation
POC of Error mode replay

#### Adding squashing logic - copied from web-ui
- Apply incremental mutation on FS
- Update Meta, Focus and VisualViewport records
- Batch StyleSheetRules
- Batch MediaInteraction
- Discard all other records
- TODO: Handle element focus and hover from mouse interactions

#### Using Squashing logic
- Periodically squash records, every 30 seconds
- Squash records before flushing cache into segments (when replay is forced)

#### Cache clearing
- On view end, clear the cache
- After records are batched into segment, clear cache

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


BenoitZugmeyer (changes requested):
This PR significantly increase the bundle size, please don't merge too quickly, I need some time to think about it

N-Boutaib:
> This PR significantly increase the bundle size, please don't merge too quickly, I need some time to think about it

Indeed. In any case, I was planing to merge it into the the POC branch.

PR #3021, opened 2024-09-19
🐛 fix to redirect v6-canary bundles urls to local dev bundles

thomas-lebeau:
## Motivation

Enable local dev bundle to be loaded properly even when the `browser_sdk_next_major_version` is enabled

## Changes

Tweak the redirect rules to account for the `v6-canary` bundles loaded when the next major feature flags is enabled

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/developer-extension/src/background/domain/syncRules.ts
+++ b/developer-extension/src/background/domain/syncRules.ts
@@ -62,11 +62,11 @@ function buildRules(
     const devRumUrl = useRumSlim ? DEV_RUM_SLIM_URL : DEV_RUM_URL
     logger.log('add redirect to dev bundles rules')
     rules.push(
-      createRedirectRule(/^https:\/\/.*\/datadog-rum(-v\d|-canary|-staging|-v\d-canary)?\.js$/, { url: devRumUrl }),
-      createRedirectRule(/^https:\/\/.*\/datadog-rum-slim(-v\d|-canary|-staging|-v\d-canary)?\.js$/, {
+      createRedirectRule(/^https:\/\/.*\/datadog-rum(-[\w-]+)?\.js$/, { url: devRumUrl }),
+      createRedirectRule(/^https:\/\/.*\/datadog-rum-slim(-[\w-]+)?\.js$/, {
         url: DEV_RUM_SLIM_URL,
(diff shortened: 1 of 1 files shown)


developer-extension/src/background/domain/syncRules.ts
```diff
@@ -62,11 +62,11 @@ function buildRules(
     const devRumUrl = useRumSlim ? DEV_RUM_SLIM_URL : DEV_RUM_URL
     logger.log('add redirect to dev bundles rules')
     rules.push(
-      createRedirectRule(/^https:\/\/.*\/datadog-rum(-v\d|-canary|-staging)?\.js$/, { url: devRumUrl }),
-      createRedirectRule(/^https:\/\/.*\/datadog-rum-slim(-v\d|-canary|-staging)?\.js$/, {
+      createRedirectRule(/^https:\/\/.*\/datadog-rum(-v\d|-canary|-staging|-v\d-canary)?\.js$/, { url: devRumUrl }),
```
BenoitZugmeyer, line 65: **💬 suggestion:** ‏I think we should simplify the regex to make it future proof: `datatdog-rum(-\w+)?\.js` . It's pretty low-risk.

cy-moi: 💭 Though I think there is an upside of having the explicit rules: it can serve as docs for the versions we have and always updated.

cy-moi (approved):

developer-extension/src/background/domain/syncRules.ts
BenoitZugmeyer: We have better documentation here though (although a bit outdated): https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2523531139/SDK+distribution#CDN

thomas-lebeau: fixed, Mainly for simplification as I'm not totally convinced that it's futureproof, especially with the split bundles initiative, but we will fix it in due time.

I'll update the doc too, thanks

BenoitZugmeyer (approved):

PR #3022, opened 2024-09-19
🚨 enforce `scripts/` files conventions

BenoitZugmeyer:
## Motivation

As time pass, we get more complex scripts in our `scripts/` folders. We have some convention to group common libraries into `lib` folders.

When a script is getting big, like the `scripts/performance`, we want to split it into multiple JS files, but then it is hard to understand what is supposed to be a script and what is a lib.

## Changes

* (preliminary) enforce `import/no-unresolved` on commonjs files. As we are renaming a bunch of files, we want to make sure we don't have any typo in our `require(...)`.

* Exclude `lib/` folders from ESLint scripts rules. This way, we have the same rules for script libs as other source files (in particular naming conventions) but also we can enable more rules targeting scripts specifically. Which leads to...

* Add a custom lint for scripts to ensure all scripts are effectively scripts (= contains `runMain`)

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


amortemousque:
Nice! Maybe could also enforce the  Kebab Case  script file naming.

amortemousque (approved):

PR #3031, opened 2024-09-23
[RUM 4813] Remove feature flag for view specific context

cy-moi:
## Motivation
Make view specific context apis public without experimental feature flag.

## Changes
Removed feature flag and set APIs public by defualt.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.spec.ts
+++ b/packages/rum-core/src/domain/assembly.spec.ts
@@ -197,7 +197,6 @@ describe('rum assembly', () => {
         })
 
         it('should accept modification on context field for view events', () => {
-          mockExperimentalFeatures([ExperimentalFeature.VIEW_SPECIFIC_CONTEXT])
           const { lifeCycle, serverRumEvents } = setupAssemblyTestWithDefaults({
             partialConfiguration: {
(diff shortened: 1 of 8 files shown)


packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -213,22 +212,6 @@ describe('rum assembly', () => {
           expect(serverRumEvents[0].context).toEqual({ foo: 'bar' })
         })
 
-        it('should reject modification on context field for view events', () => {
```
N-Boutaib, line 216: ❓ **Question**: I assume this test is removed because it was testing when the FF is disabled 🤔

N-Boutaib (approved):

thomas-lebeau (approved):

amortemousque (approved):

packages/rum-core/src/domain/assembly.spec.ts
cy-moi: Right. So it was testing the opposite, when the view context is read-only. I kept it while this feature was behind ff to make sure the ff works.

PR #3057, opened 2024-10-07
👷‍♀️ [RUM-6424] Change webpack chunk config to speed up local unit tests

cy-moi:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/test/unit/karma.base.conf.js
+++ b/test/unit/karma.base.conf.js
@@ -56,7 +56,6 @@ module.exports = {
       // with all dependencies shared.  Our test suite does not support sharing dependencies, each
       // spec bundle should include its own copy of dependencies.
       runtimeChunk: false,
-      splitChunks: false,
     },
     ignoreWarnings: [
       // we will see warnings about missing exports in some files


thomas-lebeau (approved):

test/unit/karma.local.conf.js
```diff
@@ -18,6 +18,7 @@ module.exports = function (config) {
     },
     webpack: {
       ...karmaBaseConf.webpack,
+      optimization: { runtimeChunk: false },
```
amortemousque, line 21: Why are we only doing it for local unit tests?

cy-moi: I don't think we run multiple times consecutively for cicd. This only speed it up when we want to make changes and rerun quickly. [BrowserStack unit tests](https://gitlab.ddbuild.io/DataDog/browser-sdk/-/jobs/663886729) would fail without `splitChunk:false`.

BenoitZugmeyer (approved):

cy-moi:
Same as the developer-extension, `rum-react` is also failed to be excluded and we cannot make it compatible with IE. So let's merge this in v6 directly.

PR #3059, opened 2024-10-09
👷[IR-30972] Include older version intake urls matching

cy-moi:
## Motivation
We collect network errors from older version sdks, because the intake urls have different prefixes. 

## Changes
We check `isIntakeUrl` on parameter keywords instead of the full prefix.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/intakeSites.ts
+++ b/packages/core/src/domain/configuration/intakeSites.ts
@@ -5,4 +5,4 @@ export const INTAKE_SITE_EU1 = 'datadoghq.eu'
 export const INTAKE_SITE_US1_FED = 'ddog-gov.com'
 
 export const PCI_INTAKE_HOST_US1 = 'pci.browser-intake-datadoghq.com'
-export const INTAKE_TAGS = ['ddsource', 'ddtags']
+export const INTAKE_URL_PARAMETERS = ['ddsource', 'ddtags']
--- a/packages/core/src/domain/configuration/intakeSites.ts
+++ b/packages/core/src/domain/configuration/intakeSites.ts
@@ -5,3 +5,4 @@ export const INTAKE_SITE_EU1 = 'datadoghq.eu'
 export const INTAKE_SITE_US1_FED = 'ddog-gov.com'
 
 export const PCI_INTAKE_HOST_US1 = 'pci.browser-intake-datadoghq.com'
+export const INTAKE_TAGS = ['ddsource', 'ddtags']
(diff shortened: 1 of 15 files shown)


thomas-lebeau (approved):

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -26,13 +26,13 @@ export function computeTransportConfiguration(initConfiguration: InitConfigurati
   const tags = buildTags(initConfiguration)
 
   const endpointBuilders = computeEndpointBuilders(initConfiguration, tags)
-  const intakeUrlPrefixes = computeIntakeUrlPrefixes(endpointBuilders, site)
```
BenoitZugmeyer, line 29: **💬 suggestion:** ‏Looks like URL prefixes are not used anymore, could you remove them?

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -72,15 +72,7 @@ function computeReplicaConfiguration(
...
-  const intakeUrlPrefixes = objectValues(endpointBuilders).map((builder) => builder.urlPrefix)
-
-  if (site === INTAKE_SITE_US1) {
-    intakeUrlPrefixes.push(`https://${PCI_INTAKE_HOST_US1}/`)
-  }
-
-  return intakeUrlPrefixes
+function hasIntakeParameters(url: string) {
+  // check if tags is present in the query string
+  return url.indexOf('ddsource') !== -1 || url.indexOf('ddtags') !== -1
```
BenoitZugmeyer, line 77: **💬 suggestion:** Be a bit stricter? URLs should have both params, you could use `&&` instead of `||`

**🥜 nitpick:** ‏use `includes()` (from polyfills.ts) instead of `indexOf`. It's a bit annoying because of v6 though...

packages/core/src/domain/configuration/intakeSites.ts
```diff
@@ -5,3 +5,4 @@ export const INTAKE_SITE_EU1 = 'datadoghq.eu'
 export const INTAKE_SITE_US1_FED = 'ddog-gov.com'
 
 export const PCI_INTAKE_HOST_US1 = 'pci.browser-intake-datadoghq.com'
+export const INTAKE_TAGS = ['ddsource', 'ddtags']
```
BenoitZugmeyer, line 8: **🥜 nitpick:** ‏`INTAKE_URL_PARAMETERS` could be more explicit

BenoitZugmeyer (approved):

PR #3077, opened 2024-10-17
🐛 [RUM-6483] Investigate reducing the batch time limit for Replay

RomanGaignault:
## Motivation

There's some missing replay segment, because most likely the browser/javascript execution stop before data is compressed or uploaded.

## Changes

Flush data more often, like every 5 seconds instead of 30 seconds could increase the chances to send Replay data

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/domain/segmentCollection/segmentCollection.ts
+++ b/packages/rum/src/domain/segmentCollection/segmentCollection.ts
@@ -14,8 +14,11 @@ import { buildReplayPayload } from './buildReplayPayload'
 import type { FlushReason, Segment } from './segment'
 import { createSegment } from './segment'
 
-export const SEGMENT_DURATION_LIMIT = () =>
-  isExperimentalFeatureEnabled(ExperimentalFeature.REDUCE_SEGMENT_LIMIT_BATCH_TIME) ? 5 * ONE_SECOND : 30 * ONE_SECOND
+export function getSegmentDurationLimit() {
+  return isExperimentalFeatureEnabled(ExperimentalFeature.REDUCE_SEGMENT_LIMIT_BATCH_TIME)
+    ? 5 * ONE_SECOND
+    : 30 * ONE_SECOND
+}
 /**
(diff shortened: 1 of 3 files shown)


thomas-lebeau (approved):

packages/rum/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -1,13 +1,21 @@
...
 import type { LifeCycle, ViewHistory, RumSessionManager, RumConfiguration } from '@datadog/browser-rum-core'
 import { LifeCycleEventType } from '@datadog/browser-rum-core'
 import type { BrowserRecord, CreationReason, SegmentContext } from '../../types'
 import { buildReplayPayload } from './buildReplayPayload'
 import type { FlushReason, Segment } from './segment'
 import { createSegment } from './segment'
 
-export const SEGMENT_DURATION_LIMIT = 30 * ONE_SECOND
+export const SEGMENT_DURATION_LIMIT = () =>
+  isExperimentalFeatureEnabled(ExperimentalFeature.REDUCE_SEGMENT_LIMIT_BATCH_TIME) ? 5 * ONE_SECOND : 30 * ONE_SECOND
```
BenoitZugmeyer, line 18: * Top level functions should be declared as `function ...(){}`, not arrow functions.
* Function names should be in camel case, only top level constants primitive values should be in upper case.

Let's find some eslint rule to enforce that. In the meantime:

**💬 suggestion:** ‏

```suggestion
export function getSegmentDurationLimit() {
  return isExperimentalFeatureEnabled(
    ExperimentalFeature.REDUCE_SEGMENT_LIMIT_BATCH_TIME) ? 5 * ONE_SECOND : 30 * ONE_SECOND
  )
}
```

BenoitZugmeyer (approved):

BenoitZugmeyer (approved):

PR #3086, opened 2024-10-22
[React RUM] Add a ReactComponentTracker component

Miz85:
## Motivation

Provide a way to measure React components render duration and report them to DataDog as `Custom Vitals`

## Changes

- Create a `performance` domain in the `rum-react` package 
- Provide a `ReactComponentTracker` component that can be used to wrap components that you want to measure render duration for.
- 
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-react/src/domain/performance/getTimer.ts
```diff
@@ -0,0 +1,23 @@
...
+  let duration: number
+  let startTime: number
+
+  function startTimer() {
+    const start = performance.now()
+    startTime = performance.timeOrigin + start
+  }
+
+  function stopTimer() {
+    duration = performance.timeOrigin + performance.now() - startTime
```
BenoitZugmeyer, line 11: **💭 thought:** ‏I was considering to suggest using functions from `timeUtils` (ex: `timeStampNow()` etc), similarly to the rest of the SDK including Duration Vitals. But then we would use `Date.now()` instead of `performance.now()`, which is somewhat less precise. So using `performance.now()` here *might* make sense, but then why not use it in Vitals?

🤔 🤔

packages/rum-react/archive.tgz
BenoitZugmeyer, line 1: **💬 suggestion:** ‏Remove this file

packages/rum-react/package.json
```diff
@@ -13,7 +13,8 @@
   },
   "dependencies": {
     "@datadog/browser-core": "5.34.1",
-    "@datadog/browser-rum-core": "5.34.1"
+    "@datadog/browser-rum-core": "5.34.1",
+    "uuid": "10.0.0"
```
BenoitZugmeyer, line 17: **💬 suggestion:** ‏Remove this dependency

packages/rum-react/package.json
```diff
@@ -36,6 +37,7 @@
   "devDependencies": {
     "@types/react": "18.3.17",
     "@types/react-dom": "18.3.5",
+    "@types/uuid": "^10",
```
BenoitZugmeyer, line 40: **💬 suggestion:** ‏Remove this dependency

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
```diff
@@ -0,0 +1,84 @@
...
+    const effectDuration = effectTimer.getDuration() ?? 0
+    const layoutEffectDuration = layoutEffectTimer.getDuration() ?? 0
+
+    const totalRenderTime = renderDuration + effectDuration + layoutEffectDuration
+
+    addDurationVital(componentName, {
+      startTime: renderTimer.getStartTime() ?? 0,
+      duration: totalRenderTime,
+      context: {
+        s_first_render: isFirstRender.current,
```
BenoitZugmeyer, line 32: **💬 suggestion:** ‏rename to `is_first_render`

packages/rum-react/src/domain/performance/timer.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { mockClock, registerCleanupTask } from '@datadog/browser-core/test'
+import { createTimer } from './timer'
+
+describe('getTimer', () => {
```
BenoitZugmeyer, line 4:
```suggestion
describe('createTimer', () => {
```

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
```diff
@@ -0,0 +1,84 @@
...
+        render_phase_duration: renderDuration,
+        effect_phase_duration: effectDuration,
+        layout_effect_phase_duration: layoutEffectDuration,
+        component_name: componentName,
```
BenoitZugmeyer, line 36: **💭 thought:** ‏I don't think we'll want to have the component name twice (once in the vital name, once in the context). Also we have the `description` field in vitals designed just for this use case. So I'm thinking something like:

```ts
addDurationVital("React component render", {
  description: componentName,
  ...
})
```

and remove `context.component_name`. Does it seems good to you @Miz85 ?

sandbox/react-app/main.tsx
```diff
@@ -3,7 +3,7 @@ import React, { useEffect, useState } from 'react'
 import ReactDOM from 'react-dom/client'
 import { datadogRum } from '@datadog/browser-rum'
 import { createBrowserRouter } from '@datadog/browser-rum-react/react-router-v6'
-import { reactPlugin, ErrorBoundary } from '@datadog/browser-rum-react'
+import { reactPlugin, ErrorBoundary, ReactComponentTracker } from '@datadog/browser-rum-react'
```
BenoitZugmeyer, line 6: **💬 suggestion:** ‏
```suggestion
import { reactPlugin, ErrorBoundary, UNSTABLE_ReactComponentTracker } from '@datadog/browser-rum-react'
```
@RomanGaignault could you verify that it works as expected using `yarn dev` and navigating to `http://localhost:8080/react-app`?

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
Miz85: That sounds good. I originally did it that way in order to make queries more intuitive as `@vital.component_name:xxx` seemed more straightforward than querying `@vital.description` or `@vital.name`.

BenoitZugmeyer (approved):

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+// eslint-disable-next-line
+export const UNSTABLE_ReactComponentTracker = ({
+  name: componentName,
+  children,
+}: {
+  name: string
+  context?: object
+  children?: React.ReactNode
+  burstDebounce?: number
```
BenoitZugmeyer, line 13: **💬 suggestion:**
* `burstDebounce` is not used, you could remove it
* `context` is not used, but you could mergi it in `context` when reporting the vital : ` framework: 'react', ...context }`

packages/rum-react/src/domain/performance/reactComponentTracker.spec.tsx
```diff
@@ -24,4 +24,25 @@ describe('UNSTABLE_ReactComponentTracker', () => {
...
+    const addDurationVitalSpy = jasmine.createSpy()
+    initializeReactPlugin({
+      publicApi: {
+        addDurationVital: addDurationVitalSpy,
+      },
+    })
+
+    appendComponent(
+      // eslint-disable-next-line camelcase
+      <UNSTABLE_ReactComponentTracker name="test component" context={{ framework: 'angular' }}>
```
Miz85, line 38: shouldn't framework be `forced` to `react` when tracking React components?

Miz85: I'm actually starting to think that the whole context prop might be overkill for now. We should probably force the framework to `react` and not expose a way to add extra context for now? And maybe add that when a use case for it arises? wdyt @BenoitZugmeyer @RomanGaignault ?

BenoitZugmeyer: Sounds good to me

RomanGaignault: yep sounds good 👍

BenoitZugmeyer (approved):
I experimented a bit with that PR and ended up adding a couple of commits. Looks good to me now :)

packages/rum-react/src/domain/performance/timer.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+describe('createTimer', () => {
+  it('is able to measure time', () => {
+    const clock = mockClock()
+    registerCleanupTask(clock.cleanup)
+
+    const timer = createTimer()
+    timer.startTimer()
+    clock.tick(1000)
+    timer.stopTimer()
+    expect(timer.getDuration()).toBe(1000)
```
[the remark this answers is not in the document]
BenoitZugmeyer: thanks!

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
```diff
@@ -0,0 +1,82 @@
...
+    isFirstRender.current = false
+  }
+
+  /**
+   * In react, children are rendered sequentially
+   * in the order they are defined. that's why we
+   * can measure perf timings of a component by
+   * starting recordings in the component above
+   * and stopping them in the component below.
+   */
```
sethfowler-datadog, line 48: Nit: comment could, uh, be wider.

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
```diff
@@ -0,0 +1,82 @@
...
+        onEffect={effectTimer.startTimer}
+      />
+      {children}
+      <LifeCycle
+        onRender={renderTimer.stopTimer}
+        onLayoutEffect={layoutEffectTimer.stopTimer}
+        onEffect={() => {
+          effectTimer.stopTimer()
+          onEffectEnd()
+        }}
```
sethfowler-datadog, line 63: This seems like a good approach! It's a bit unfortunate that no timing information for the reconciliation phase is captured (unless I misunderstand something), but that's admittedly a bit harder to capture, since there are no natural callbacks to hook into. At some point in the future, it might be nice to see if we can hook into the profiler hooks that `ReactFiberWorkLoop` [exposes internally](https://github.com/facebook/react/blob/db6513914f99c260090f26f0a547ee1432c934e6/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L1626)...

packages/rum-react/src/domain/performance/reactComponentTracker.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    clock.tick(1)
+
+    flushSync(() => {
+      forceUpdate!()
+    })
+
+    expect(addDurationVitalSpy).toHaveBeenCalledTimes(2)
+    const options = addDurationVitalSpy.calls.mostRecent().args[1]
+    expect(options.startTime).toBe(clock.timeStamp(TOTAL_DURATION + 1))
```
sethfowler-datadog, line 95: Nit: I'd be a bit inclined to just assert against the whole `options` object, unless there's something unstable in there that makes it hard. We're not checking whether `is_first_render` is `false` on the second run here, for example.

sethfowler-datadog (approved):

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
BenoitZugmeyer: Fixed!

packages/rum-react/src/domain/performance/reactComponentTracker.spec.tsx
BenoitZugmeyer: Ah right, `is_first_render` makes sense. Updated!

packages/rum-react/src/domain/performance/reactComponentTracker.tsx
BenoitZugmeyer: Or we could directly use the [`Profiler`](https://react.dev/reference/react/Profiler) :) We'll see.

Miz85: Agreed @sethfowler-datadog that it would be a lot better if we could use the hooks that the dev tools use as they track everything we need.
I'm not against testing the Profiler but I still think that it will be limiting as the docs about how to use it are very poorly written and I think it will bring us support issues about people not being able to make it work properly with their build system.

romellem:
Hello! I currently use Datadog and Datadog RUM to monitor a web application, and was looking to get more fine tuned information on specific component renders.

I came across this PR, but see this is marked as `UNSTABLE_`.

Just curious, how "unstable" is this? Any identified downsides to using this in production today? If the answer is just "this is untested, therefore we don't recommend it be used in critical applications," is there any plan to harden this component so it can be production ready?

Whatever the answer, this looks like cool work!

PR #3096, opened 2024-10-25
✅👷 kill browserstack execution early

BenoitZugmeyer:
## Motivation

e2e-bs tests are still quite flaky, sometimes because of a BrowserStack issue (?). Because we can run a single e2e-bs job at a time, we should detect and stop the jobs as soon as possible so other can be run.

## Changes

When detecting an unrecoverable error in the command output, or if the command doesn't print anything within 5 minutes, let's kill it early, we don't wait for nothing.

For now, we don't implement a retry mechanism, but it could be done later.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/test/bs-wrapper.js
+++ b/scripts/test/bs-wrapper.js
@@ -14,11 +14,13 @@
 // after killing it. There might be a better way of prematurely aborting the test command if we need
 // to in the future.
 
-const { spawnCommand, printLog, runMain, timeout } = require('../lib/executionUtils')
+const spawn = require('child_process').spawn
+const { printLog, runMain, timeout, printError } = require('../lib/executionUtils')
 const { command } = require('../lib/command')
 const { browserStackRequest } = require('../lib/bsUtils')
 
 const AVAILABILITY_CHECK_DELAY = 30_000
+const NO_OUTPUT_TIMEOUT = 5 * 60_000
 const BS_BUILD_URL = 'https://api.browserstack.com/automate/builds.json?status=running'
 
 runMain(async () => {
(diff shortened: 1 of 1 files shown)


scripts/test/bs-wrapper.js
```diff
@@ -43,6 +45,49 @@ async function hasRunningBuild() {
...
+
+    function onOutput(data) {
+      output += data
+
+      if (hasUnrecoverableFailure(output)) {
+        killIt('unrecoverable failure')
+      }
+
+      clearTimeout(timeoutId)
+      setTimeout(() => killIt('no output timeout'), NO_OUTPUT_TIMEOUT)
```
thomas-lebeau, line 78: **🔨 warning:** ‏ 
```suggestion
      timeoutId = setTimeout(() => killIt('no output timeout'), NO_OUTPUT_TIMEOUT)
```

scripts/test/bs-wrapper.js
```diff
@@ -43,6 +45,49 @@ async function hasRunningBuild() {
...
+    }
+
+    function killIt(message) {
+      if (!isKilled) {
```
thomas-lebeau, line 82: **❓ question:** ‏ Do we need this `isKilled` variable? Instead, can we return on line 74? (unless `child.kill('SIGTERM')` does not terminate immediately and somehow some more output is processed.

scripts/test/bs-wrapper.js
BenoitZugmeyer: Aah good catch, that's why it didn't work :D

scripts/test/bs-wrapper.js
BenoitZugmeyer: > unless child.kill('SIGTERM') does not terminate immediately and somehow some more output is processed

"SIGTERM" is typically a "graceful exit" -> the process can catch it and continue its execution to clean things up before exiting. So it is possible to have some output after a SIGTERM.

In practice, I'm not sure it happens in this particular situation (= karma and wdio processes). Sending the signal multiple times shouldn't hurt, so let's simplify the code!

thomas-lebeau (approved):

PR #3102, opened 2024-10-29
⚡️ [RUM-6929] delay resource collection 

BenoitZugmeyer:
## Motivation

We observed that on some condition, processing resources can block the main thread for a significant amount of time.

Screenshot 2024-10-29 at 18 37 44

## Changes

This PR mitigate the issue by processing each resource in a "task queue". Tasks are scheduled to be run in "idle tasks" via `requestIdleCallback`. This will break down those big blocking tasks into smaller tasks with lower priority.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


thomas-lebeau (approved):

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -41,13 +42,11 @@ export function startResourceCollection(
   lifeCycle: LifeCycle,
   configuration: RumConfiguration,
   pageStateHistory: PageStateHistory,
+  taskQueue = createTaskQueue(),
```
amortemousque, line 45: **❓ question:** ‏What is the benefit of the taskQueue instead of directly using requestIdleCallback?

BenoitZugmeyer: Like, calling `requestIdleCallback` for each resource entry? This is a great question, 

* Performance: we have less asynchronous tasks to schedule, with task queues, tasks are run in synchronous batch.

* Scheduling: if each task use `requestIdleCallback` directly, each task will use its own timeout. So if we have a a lot of tasks timeout at the same time, we will still block the main thread for an uncontrolled amount of time. With the taskqueue, we can control how much time we block the main thread.

Ideally, we should have one taskqueue shared across the SDK, and replace`requestIdleRequest` usages with it, in particular in `mutationBatch` but it's not easy as there are other constraints here (= the throttle that you added in #2807), this is why I didn't act on it yet.

packages/core/src/tools/taskQueue.ts
```diff
@@ -0,0 +1,58 @@
...
+  push(task: Task): void
+}
+
+type Task = () => void
+
+export function createTaskQueue(): TaskQueue {
+  const pendingTasks: Task[] = []
+
+  function run(deadline: IdleDeadline) {
+    let timeRemaining: () => number
```
amortemousque, line 30: **💬 suggestion:** ‏
```suggestion
    let executionTimeRemaining: () => number
```

packages/core/src/tools/taskQueue.ts
```diff
@@ -0,0 +1,58 @@
+import { ONE_SECOND } from './utils/timeUtils'
+import { requestIdleCallback } from './requestIdleCallback'
+
+/**
+ * Maximum amount of time to delay the tasks. We don't want to wait too long before running tasks,
```
amortemousque, line 5: **🥜 nitpick:** ‏
```suggestion
 * Maximum delay before starting to execute tasks in the queue. We don't want to wait too long before running tasks,
```

packages/core/src/tools/taskQueue.ts
```diff
@@ -0,0 +1,58 @@
...
+ * as it might hurt reliability (ex: if the user navigates away, we might lose the opportunity to
+ * send some data). We also don't want to run tasks too often, as it might hurt performance.
+ */
+const IDLE_CALLBACK_TIMEOUT = ONE_SECOND
+
+/**
+ * Minimum amount of time to run the task when a timeout is reached. We should not run tasks for
```
amortemousque, line 12:
```suggestion
 * Maximum amount of time allocated to running tasks when `IDLE_CALLBACK_TIMEOUT` is reached. We should not run tasks for
```

packages/rum-core/src/domain/resource/resourceCollection.ts
amortemousque: Interesting! Sound good. I agree that ideally we should use this concept everywhere instead of requestIdleCallback. Maybe we can add it to our backlog.

packages/core/src/tools/taskQueue.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    const task = jasmine.createSpy('task')
+
+    taskQueue.push(task)
+    expect(task).not.toHaveBeenCalled()
+
+    requestIdleCallbackMock.idle()
+    expect(task).toHaveBeenCalled()
+  })
+
+  it('runs as much task as possible in a single idle callback', () => {
```
amortemousque, line 17:
```suggestion
  it('runs as many tasks as possible in a single idle callback', () => {
```

packages/core/src/tools/taskQueue.ts
```diff
@@ -0,0 +1,58 @@
...
+ * too long as it will hurt performance, but we should still run some tasks to avoid postponing them
+ * forever.
+ *
+ * Rational: Running tasks for 30ms every second (IDLE_CALLBACK_TIMEOUT) should be acceptable.
+ */
+export const TIMED_OUT_TIME_REMAINING = 30
```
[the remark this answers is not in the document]
BenoitZugmeyer: Maybe `MAX_EXECUTION_TIME_WHEN_TIMEOUT`?

packages/core/src/tools/taskQueue.ts
BenoitZugmeyer: I followed the native naming here. But I can rename it, no problem.

amortemousque (approved):

PR #3105, opened 2024-10-31
🐛 [RUM-6226][rum-react] improve routes wildcard substitution

BenoitZugmeyer:
## Motivation

In #3023, we implemented wildcard substition in react-router paths. For example, on a route like `/foo/*`, a path `/foo/bar` would produce a view name `/foo/bar`.

It turns out we have an internal application which uses nested routes with wildcards, ex: `{ path: '*', children: [{ path: '*' }] }`. This is probably not popular use case, but still, it exists. The issue with the previous implementation is that it duplicated the path in the view name (a path `/foo` would produce a view name `/foo/foo`).

## Changes

This PR fixes this issue and consolidates the tests with more edge cases related to wildcards.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


thomas-lebeau (approved):

packages/rum-react/src/domain/reactRouterV6/startReactRouterView.spec.ts
```diff
@@ -42,57 +43,64 @@ describe('computeViewName', () => {
...
-        { route: { path: '/foo/' } },
-        { route: { path: 'bar/' } },
-        { route: { path: ':id/' } },
-      ] as RouteMatch[])
-    ).toBe('/foo/bar/:id/')
-  })
+      // Simple paths
+      ['/foo',                '/foo',         '/foo'],
+      ['/foo',                '/bar',         '/foo'], // apparently when the path doesn't match any route, React Router returns the last route as a matching route
+      ['/foo > bar',          '/foo/bar',     '/foo/bar'],
```
amortemousque, line 59: **❓ question:** ‏I don't know about the ">" syntax is it a react-router thing?

packages/rum-react/src/domain/reactRouterV6/startReactRouterView.ts
```diff
@@ -33,13 +35,27 @@ export function computeViewName(routeMatches: RouteMatch[]) {
...
-        viewName += path
-      }
+      viewName += path
     }
   }
 
   return viewName
 }
+
+function substitutePathWildcards(
```
amortemousque, line 45: **💬 suggestion:** ‏I don't fully understand what this function does. Can you add a comment maybe with some path example before and after?

packages/rum-react/src/domain/reactRouterV6/startReactRouterView.spec.ts
BenoitZugmeyer: It's not, it's just quick way to define nested routes for tests, see `.split(' > ')` below

amortemousque (approved):

PR #3108, opened 2024-11-05
✨[RUM-6798] Collect resource content type

RomanGaignault:
## Motivation

Content type is newly available on resource performance entry: [PerformanceResourceTiming: contentType property - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/contentType) 

## Changes

We could add it as a @resource.content_type field

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


cy-moi (approved):

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -186,6 +186,10 @@ export function computeResourceEntryProtocol(entry: RumPerformanceResourceTiming
   return entry.nextHopProtocol === '' ? undefined : entry.nextHopProtocol
 }
 
+export function computeResourceContentType(entry: RumPerformanceResourceTiming) {
```
nulrich, line 189: **💬 suggestion:** ‏Maybe a little comment to explain why we discard empty string.

PR #3114, opened 2024-11-07
✨ - New feature: Set global tags in init configuration.

anthonypruiz:
## Motivation

We have a complex distributed infrastructure segmented into many business and technical components. In the aim to better account for ownership of telemetry data we have a number of required tags that need to be present. This sdk was lacking a way to set global tags. setting attributes with the setGlobalContext in logging for example would not suffice for our business goals. 

## Changes
- Add `ddtags` as an optional string array parameter in the interface `InitConfiguration`.
- Add `ddtags` as an optional configuration parameter to `buildTags` function 
## Testing

Example test for NextJs

```
import { datadogLogs } from '@datadog/browser-logs';
...
  useEffect(() => {
    datadogLogs.init({
      clientToken: process.env.NEXT_PUBLIC_DD_CLIENT_TOKEN || '',
      site: 'datadoghq.com',
      forwardErrorsToLogs: true,
      sessionSampleRate: 100,
      env: process.env.NEXT_PUBLIC_DD_ENV || 'test',
      service: 'my-fancy-service',
      version: process.env.NEXT_PUBLIC_DD_VERSION || '0.0.1',
      forwardConsoleLogs: "all",
      ddtags: [
        'team:my_team',
        'org_id:007',
        'source:nextjs',
        'location_number:007'
      ],
    });
  }
...
```

- [X] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -97,7 +97,10 @@ export interface InitConfiguration {
    * The application’s version, for example: 1.2.3, 6c44da20, and 2020.02.13. Follows the [tag syntax requirements](https://docs.datadoghq.com/getting_started/tagging/#define-tags).
    */
   version?: string | undefined | null
-
+  /**
+   * Tags to add to all events. Follows the [tag syntax requirements](https://docs.datadoghq.com/getting_started/tagging/#define-tags).
+   */
+  ddtags?: string[] | undefined
   // cookie options
   /**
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer:
Hello,
Thank you for your contribution, but custom tags is not something we want to support right now. You can use alternatives like the global context. See https://github.com/DataDog/browser-sdk/issues/462 .

PR #3115, opened 2024-11-07
✨ [RUM-6956] Add action name source

cy-moi:
## Motivation
We want to add a field to indicate from which source the action names are generated.

## Changes
Add a `name_source` field in `@_dd.action` object

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/action/getActionNameFromElement.ts
+++ b/packages/rum-core/src/domain/action/getActionNameFromElement.ts
@@ -9,41 +9,56 @@ import type { RumConfiguration } from '../configuration'
  */
 export const DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE = 'data-dd-action-name'
 export const ACTION_NAME_PLACEHOLDER = 'Masked Element'
+
+type ActionName = {
+  name: string
+  namingSource: string
+}
 export function getActionNameFromElement(
   element: Element,
   { enablePrivacyForActionName, actionNameAttribute: userProgrammaticAttribute }: RumConfiguration,
   nodePrivacyLevel?: NodePrivacyLevel
-): string {
+): ActionName {
   // Proceed to get the action name in two steps:
   // * first, get the name programmatically, explicitly defined by the user.
-  // * then, use strategies that are known to return good results. Those strategies will be used on
-  //   the element and a few parents, but it's likely that they won't succeed at all.
+  // * then, if privacy is set to mask, return a placeholder for the undefined.
+  // * if privacy is not set to mask, use strategies that are known to return good results.
+  //   Those strategies will be used on the element and a few parents, but it's likely that they won't succeed at all.
   // * if no name is found this way, use strategies returning less accurate names as a fallback.
   //   Those are much likely to succeed.
   const defaultActionName =
     getActionNameFromElementProgrammatically(element, DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE) ||
     (userProgrammaticAttribute && getActionNameFromElementProgrammatically(element, userProgrammaticAttribute))
 
-  if (nodePrivacyLevel === NodePrivacyLevel.MASK) {
-    return defaultActionName || ACTION_NAME_PLACEHOLDER
+  if (defaultActionName) {
+    return { name: defaultActionName, namingSource: 'custom_attribute' }
+  } else if (nodePrivacyLevel === NodePrivacyLevel.MASK) {
+    return { name: ACTION_NAME_PLACEHOLDER, namingSource: 'mask_placeholder' }
   }
 
-  return (
-    defaultActionName ||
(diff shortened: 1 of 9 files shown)


packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -9,41 +9,56 @@ import type { RumConfiguration } from '../configuration'
...
-    ) ||
-    ''
+  const standardName = getActionNameFromElementForStrategies(
+    element,
+    userProgrammaticAttribute,
+    priorityStrategies,
+    enablePrivacyForActionName
   )
+  if (standardName) {
+    return { name: standardName, namingSource: 'standard_attribute' }
```
BenoitZugmeyer, line 46: **💬 suggestion:** ‏`priorityStrategies` don't always return standard attribute values though, sometimes it is `text_content`. Maybe strategies can return an `ActionName` object, that we can use to get both the name and correct source.

(Sorry if this is what you asked confirmation for during the standup, I might have misunderstood)

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -9,41 +9,56 @@ import type { RumConfiguration } from '../configuration'
...
+    fallbackStrategies,
+    enablePrivacyForActionName
+  )
+
+  return fallbackName
+    ? {
+        name: fallbackName,
+        namingSource: 'text_content',
+      }
+    : { name: '', namingSource: 'blank_placeholder' }
```
BenoitZugmeyer, line 61: **💬 suggestion:** ‏maybe just `blank` could be simpler than `blank_placeholder`

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: Yes, I have a local version that manually returns different source from each strategy. But I wonder if we could also make `getTextualContent` return `ActionName` object.

BenoitZugmeyer: Returning an ActionName object from getTextualContent sounds great. Maybe rename the function to `getActionNameFromTextContent` or something, it could be more explicit

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -208,6 +213,15 @@ function getElementById(refElement: Element, id: string) {
   return refElement.ownerDocument ? refElement.ownerDocument.getElementById(id) : null
 }
 
+function getActionNameFromTextualContent(
+  element: Element | HTMLElement,
+  userProgrammaticAttribute: string | undefined,
+  privacyEnabledActionName?: boolean
+): ActionName | undefined {
+  const name = getTextualContent(element, userProgrammaticAttribute, privacyEnabledActionName)
+  return name ? { name, nameSource: ActionNameSource.TEXT_CONTENT } : undefined
```
BenoitZugmeyer, line 222: **💬 suggestion:** ‏I don't think returning `undefined` is useful here, since we'll fallback when the name is empty anyway
```suggestion
  return {
    name: getTextualContent(element, userProgrammaticAttribute, privacyEnabledActionName),
    nameSource: ActionNameSource.TEXT_CONTENT
  }
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -176,11 +177,15 @@ function getActionNameFromElementForStrategies(
...
   ) {
     for (const strategy of strategies) {
-      const name = strategy(element, userProgrammaticAttribute, privacyEnabledActionName)
-      if (typeof name === 'string') {
+      const actionName = strategy(element, userProgrammaticAttribute, privacyEnabledActionName)
+      if (actionName) {
+        const { name, nameSource } =
+          typeof actionName === 'string'
+            ? { name: actionName, nameSource: ActionNameSource.STANDARD_ATTRIBUTE }
+            : actionName
```
BenoitZugmeyer, line 185: **💬 suggestion:** ‏I think it would be simpler if strategies didn't return `string`. If you want to factorize attributes, you could have a function like:
```js
function getActionNameFromStandardAttribute(element, attribute) {
  return {
    name: element.getAttribute(attribute),
    nameSource: ActionNameSource.STANDARD_ATTRIBUTE
  }
}

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: This was for the cases when we fallback to `blank` if there is no `text_content`, or we can directly return `text_content` or `blank` here. Wdyt?

BenoitZugmeyer (approved):

nulrich (approved):

PR #3118, opened 2024-11-07
[RUM-6801] Collect resource worker processing time

RomanGaignault:
## Motivation

Resource performent entries have a workerStart property
Worker processing time is defined as workerProcessingTime = entry.fetchStart - entry.workerStart

## Changes

Similarly to @resource.connect etc. we could have @resource.worker where:

@resource.worker.start would be workerStart

@resource.worker.duration would be fetchStart - workerStart

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/resourceUtils.ts
+++ b/packages/rum-core/src/domain/resource/resourceUtils.ts
@@ -155,7 +155,7 @@ export function hasValidResourceEntryTimings(entry: RumPerformanceResourceTiming
   // it.
   const areCommonTimingsInOrder = areInOrder(
     entry.startTime,
-    entry.workerStart,
+    ...(entry.workerStart > 0 ? [entry.workerStart] : []),
     entry.fetchStart,
     entry.domainLookupStart,
     entry.domainLookupEnd,
--- a/packages/rum-core/src/domain/resource/resourceUtils.ts
+++ b/packages/rum-core/src/domain/resource/resourceUtils.ts
@@ -169,6 +169,7 @@ export function hasValidResourceEntryTimings(entry: RumPerformanceResourceTiming
   const areRedirectionTimingsInOrder = hasRedirection(entry)
     ? areInOrder(entry.startTime, entry.redirectStart, entry.redirectEnd, entry.fetchStart)
     : true
+
   return areCommonTimingsInOrder && areRedirectionTimingsInOrder
 }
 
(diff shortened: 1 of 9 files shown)


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -108,6 +110,11 @@ export function computeResourceEntryDetails(entry: RumPerformanceResourceTiming)
     first_byte: formatTiming(startTime, requestStart, responseStart),
   }
 
+  // Make sure a worker processing time is recorded
+  if (workerStart < fetchStart) {
+    details.worker = formatTiming(startTime, workerStart, fetchStart)
```
nulrich, line 115: **💬 suggestion:** ‏Can you check in which case this situation can happen and maybe add a unit test to cover it.

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -148,6 +155,7 @@ export function hasValidResourceEntryTimings(entry: RumPerformanceResourceTiming
   // it.
   const areCommonTimingsInOrder = areInOrder(
     entry.startTime,
+    ...(entry.workerStart > 0 ? [entry.workerStart] : []),
```
BenoitZugmeyer, line 158: **💬 suggestion:** To makes things simpler, I would not check whether `workerStart` is in order here. In the future I want to get rid of this check altogether. An attempt was already made [here](https://github.com/DataDog/browser-sdk/pull/2688) but was never shipped because we observed too many incoherent results, but I still intend to do something about this.

packages/rum-core/src/domain/resource/resourceUtils.ts
BenoitZugmeyer: Yes, unit tests would be nice here.

packages/rum-core/src/domain/resource/resourceUtils.ts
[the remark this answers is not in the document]
RomanGaignault: It was making test failed as workerStart can possibly be equal to 0, so I just discarded when no worker was involved in the resource fetch. 👍

BenoitZugmeyer (approved):

nulrich (approved):

PR #3123, opened 2024-11-12
🐛 allow untrusted event for httpRequest xhr event listeners

thomas-lebeau:
## Motivation

Fixes an issue where a non native XMLHttpRequest override is dispatching synthetic events (i.e. non-trusted) result in the callback not being executed, and breaking session replay ingestion.

## Changes

allow untrusted event in that particular use-case. This is safe as we're not using any properties from the event.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/httpRequest.spec.ts
+++ b/packages/core/src/transport/httpRequest.spec.ts
@@ -1,6 +1,6 @@
 import { collectAsyncCalls, mockEndpointBuilder, interceptRequests } from '../../test'
 import type { Request } from '../../test'
-import type { EndpointBuilder, Configuration } from '../domain/configuration'
+import type { EndpointBuilder } from '../domain/configuration'
 import { createEndpointBuilder } from '../domain/configuration'
 import { noop } from '../tools/utils/functionUtils'
 import { createHttpRequest, fetchKeepAliveStrategy, sendXHR } from './httpRequest'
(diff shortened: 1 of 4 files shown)


packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -187,6 +182,23 @@ describe('httpRequest', () => {
         done()
       }, 100)
     })
+
+    it('should handle synthetic events events', (done) => {
```
BenoitZugmeyer, line 186: **🥜 nitpick:** ‏
```suggestion
    it('should handle synthetic events', (done) => {
```

BenoitZugmeyer (approved):

packages/core/test/emulate/mockXhr.ts
```diff
@@ -49,11 +49,12 @@ class MockEventEmitter {
     this.listeners[name] = this.listeners[name].filter((listener) => listener !== callback)
   }
 
-  protected dispatchEvent(name: string) {
-    if (!this.listeners[name]) {
+  dispatchEvent(evt: Event) {
```
thomas-lebeau, line 52: `dispatchEvent` is normally exposed on the native xhr, so I see no reason to not expose it for mockXHR, especially as I a change the signature to be more standard. 

It's just a tiny bit more verbose where dispatchEvent is called

amortemousque (approved):

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -187,7 +187,7 @@ describe('httpRequest', () => {
       const onResponseSpy = jasmine.createSpy('xhrOnResponse')
 
       interceptor.withMockXhr((xhr) => {
-        const syntheticEvent = new Event('loadend')
+        const syntheticEvent = createNewEvent('loadend', {}, false)
```
BenoitZugmeyer, line 190: **💬 suggestion:** maybe leveraging the `properties` argument like ‏`createNewEvent('loadend', { __ddIsTrusted: false })` could be simpler/more explicit...

sethfowler-datadog (approved):
Nice!

PR #3141, opened 2024-11-18
✅ Fix test cleanup tasks order

amortemousque:
## Motivation

We are transitionning from using `afterEach` to `registerCleanupTask` API to clean our tests. However with `registerCleanupTask`, tasks are executed in the wrong order:

<table>
<tr><th>afterEach</th><th>registerCleanupTask</th></tr>
<tr><td>

```ts
describe('Test', () => {
  beforeEach(() => { 
    console.log('instrument')
  })

  afterEach(() => { 
    console.log('clean instrumentation')
  })

  describe('Nested test', () => {
    beforeEach(() => {
      console.log('instrument 2')
    })

    afterEach(() => { 
      console.log('clean instrumentation 2')
    })
  })
})
```

</td><td>

```ts
describe('Test', () => {
  beforeEach(() => {
    console.log('instrument')
    registerCleanupTask(() => {
      console.log('clean instrumentation')
    })
  })

  describe('Nested test', () => {
    beforeEach(() => {
      console.log('instrument 2')
      registerCleanupTask(() => {
        console.log('clean instrumentation 2 ')
      })
    })
  })
})
```

</td></tr>
<tr><td>

```
LOG LOG: 'instrument'
LOG LOG: 'instrument 2'
LOG LOG: 'clean instrumentation 2'
LOG LOG: 'clean instrumentation'
```

</td><td>
Before

```
LOG LOG: 'instrument'
LOG LOG: 'instrument 2'
LOG LOG: 'clean instrumentation'
LOG LOG: 'clean instrumentation 2'
```

After

```
LOG LOG: 'instrument'
LOG LOG: 'instrument 2'
LOG LOG: 'clean instrumentation 2'
LOG LOG: 'clean instrumentation'
```
</td></tr>
</table>

## Changes

- Fix cleanup tasks order: image

- Fix tests

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/test/emulate/mockReportingObserver.ts
+++ b/packages/core/test/emulate/mockReportingObserver.ts
@@ -50,16 +50,8 @@ export function mockReportingObserver() {
 export type MockCspEventListener = ReturnType<typeof mockCspEventListener>
 
 export function mockCspEventListener() {
-  // eslint-disable-next-line @typescript-eslint/unbound-method
-  const originalAddEventListener = EventTarget.prototype.addEventListener
-  EventTarget.prototype.addEventListener = jasmine
-    .createSpy()
-    .and.callFake((_type: string, listener: EventListener) => {
-      listeners.push(listener)
-    })
-
-  registerCleanupTask(() => {
-    EventTarget.prototype.addEventListener = originalAddEventListener
+  spyOn(document, 'addEventListener').and.callFake((_type: string, listener: EventListener) => {
+    listeners.push(listener)
   })
 
   const listeners: EventListener[] = []
--- a/packages/core/test/emulate/mockReportingObserver.ts
+++ b/packages/core/test/emulate/mockReportingObserver.ts
@@ -50,8 +50,16 @@ export function mockReportingObserver() {
 export type MockCspEventListener = ReturnType<typeof mockCspEventListener>
 
 export function mockCspEventListener() {
-  spyOn(document, 'addEventListener').and.callFake((_type: string, listener: EventListener) => {
-    listeners.push(listener)
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const originalAddEventListener = EventTarget.prototype.addEventListener
+  EventTarget.prototype.addEventListener = jasmine
+    .createSpy()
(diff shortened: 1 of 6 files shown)


packages/core/test/emulate/mockReportingObserver.ts
```diff
@@ -50,8 +50,16 @@ export function mockReportingObserver() {
 export type MockCspEventListener = ReturnType<typeof mockCspEventListener>
 
 export function mockCspEventListener() {
-  spyOn(document, 'addEventListener').and.callFake((_type: string, listener: EventListener) => {
-    listeners.push(listener)
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const originalAddEventListener = EventTarget.prototype.addEventListener
+  EventTarget.prototype.addEventListener = jasmine
+    .createSpy()
```
BenoitZugmeyer, line 56: This seems unrelated to this PR, but in any case:

**❓ question:** ‏`spyOn(EventTarget.prototype, 'addEventListener')` didn't work?

BenoitZugmeyer (approved):

packages/core/test/emulate/mockReportingObserver.ts
amortemousque: I found cases where mocking the same API both manually with `registerCleanup` and with `spyOn` case don't work. I can't really explain why, but it seems reasonable to use the same mocking strategy twice. wdyt?

BenoitZugmeyer: > wdyt?

I guess spies are removed after all `afterEach`, so we have an order issue similar to what you are fixing in this ticket. So, I guess it makes sense.

Your explanation makes sense but it's a bit annoying that we cannot use Jasmine tools for that. This sounds extremely error prone..

One workaround I have in mind is to always attach spies first, before any other instrumentation. Then they will be detached after everything, so it will be fine.

An alternative could be to have our own `spyOn` utility that works with `registerCleanupTask` and forbid the jasmine one.

BenoitZugmeyer (approved):

PR #3144, opened 2024-11-19
🐛 [RUM-87] AddError should support Error instances coming from other JS contexts

RomanGaignault:
## Motivation

using DD_RUM.addError(error) but the error comes from a different context than the document (ex: created insde an iframe or a window.open), the SDK doesn not consider it as an Error instance and serializes it as JSON.

## Changes

Use class object to determine if it's an error or not.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/console/consoleObservable.ts
+++ b/packages/core/src/domain/console/consoleObservable.ts
@@ -1,4 +1,4 @@
-import { flattenErrorCauses, tryToGetFingerprint } from '../error/error'
+import { flattenErrorCauses, isError, tryToGetFingerprint } from '../error/error'
 import { mergeObservables, Observable } from '../../tools/observable'
 import { ConsoleApiName, globalConsole } from '../../tools/display'
 import { callMonitored } from '../../tools/monitor'
@@ -74,7 +74,7 @@ function buildConsoleLog(params: unknown[], api: ConsoleApiName, handlingStack: 
   let error: RawError | undefined
 
   if (api === ConsoleApiName.error) {
-    const firstErrorParam = find(params, (param: unknown): param is Error => param instanceof Error)
+    const firstErrorParam = find(params, (param: unknown): param is Error => isError(param))
 
     error = {
       stack: firstErrorParam ? toStackTraceString(computeStackTrace(firstErrorParam)) : undefined,
@@ -100,7 +100,7 @@ function formatConsoleParameters(param: unknown) {
   if (typeof param === 'string') {
     return sanitize(param)
   }
-  if (param instanceof Error) {
+  if (isError(param)) {
     return formatErrorMessage(computeStackTrace(param))
   }
   return jsonStringify(sanitize(param), undefined, 2)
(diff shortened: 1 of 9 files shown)


packages/core/src/domain/console/consoleObservable.ts
```diff
@@ -74,7 +74,7 @@ function buildConsoleLog(params: unknown[], api: ConsoleApiName, handlingStack:
   let error: RawError | undefined
 
   if (api === ConsoleApiName.error) {
-    const firstErrorParam = find(params, (param: unknown): param is Error => param instanceof Error)
+    const firstErrorParam = find(params, (param: unknown): param is Error => isError(param))
```
BenoitZugmeyer, line 77: **💬 suggestion:** ‏
```suggestion
    const firstErrorParam = find(params, isError)
```

packages/core/src/domain/error/error.ts
```diff
@@ -80,19 +80,21 @@ function hasUsableStack(isErrorInstance: boolean, stackTrace?: StackTrace): stac
...
+  return isError(originalError) && 'dd_fingerprint' in originalError ? String(originalError.dd_fingerprint) : undefined
 }
 
 export function getFileFromStackTraceString(stack: string) {
   return /@ (.+)/.exec(stack)?.[1]
 }
 
+export function isError(error: unknown): error is Error {
+  return Object.prototype.toString.call(error) === '[object Error]'
+}
```
BenoitZugmeyer, line 92: **💬 suggestion:** ‏Add a test case for this. Something like:

```js
const iframe = document.createElement('iframe');
document.body.appendChild(iframe);
registerCleanupTask(() => document.body.removeChild(iframe))
expect(isError(new iframe.contentWindow.Error()).toBe(true)
```

BenoitZugmeyer (approved):

amortemousque:
Should we do the same everytime we check `instanceof Error``
Like in the logger: https://github.com/DataDog/browser-sdk/blob/f90ede411e67ec3cdba264e9f1a8cdef5321088f/packages/logs/src/domain/logger.ts#L67
in the telemetry: https://github.com/DataDog/browser-sdk/blob/f90ede411e67ec3cdba264e9f1a8cdef5321088f/packages/core/src/domain/telemetry/telemetry.ts#L205

packages/logs/src/domain/logger.ts
```diff
@@ -64,7 +65,7 @@ export class Logger {
 
     if (error !== undefined && error !== null) {
       const rawError = computeRawError({
-        stackTrace: error instanceof Error ? computeStackTrace(error) : undefined,
+        stackTrace: isError(error) ? computeStackTrace(error) : undefined,
```
amortemousque, line 68: **💬 suggestion:** Could be nice to have a test about it.‏

BenoitZugmeyer (approved):

PR #3147, opened 2024-11-19
🐛 skip worker timing when no worker is used

BenoitZugmeyer:
## Motivation

PR #3118 introduced `resource.worker.start/duration` properties to reflect the time it took for a potential worker to process the resource. But those properties were defined even if no worker were present.

## Changes

Make sure `workerStart` is defined (greater than 0) before using it.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/resourceUtils.ts
+++ b/packages/rum-core/src/domain/resource/resourceUtils.ts
@@ -111,7 +111,7 @@ export function computeResourceEntryDetails(entry: RumPerformanceResourceTiming)
   }
 
   // Make sure a worker processing time is recorded
-  if (workerStart < fetchStart) {
+  if (0 < workerStart && workerStart < fetchStart) {
     details.worker = formatTiming(startTime, workerStart, fetchStart)
   }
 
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -111,7 +111,7 @@ export function computeResourceEntryDetails(entry: RumPerformanceResourceTiming)
   }
 
   // Make sure a worker processing time is recorded
-  if (workerStart < fetchStart) {
+  if (0 < workerStart && workerStart < fetchStart) {
```
amortemousque, line 114: **🥜 nitpick:** I generally find more readable when the variables precede static values ‏
```suggestion
  if (workerStart > 0 && workerStart < fetchStart) {
```

RomanGaignault (approved):

amortemousque (approved):

packages/rum-core/src/domain/resource/resourceUtils.ts
BenoitZugmeyer: I generally agree, except when we check whether a value is between two other values `a < b && b < c`

PR #3152, opened 2024-11-20
⚡️ [RUM-6813] Lazy load session replay

amortemousque:
## Motivation

Use dynamic imports to lazy load Session Replay, reducing the bundle size by **3KB** (compressed) for users without Session Replay. Transitioning from Webpack, which adds overhead via `__webpack_require__`, to a new bundler could save additional bytes.

## Changes

- Configure ts module to ES2020
- Update test app to target node and ES2018 (fix ssr). I made to webpack config files: 
  - `webpack.ssr.js` targeting `node` and used for the ssr test
  - `webpack.web.js` targeting `web` and used for e2e tests
- Lazy load `startRecording` using dynamic import
- Updated the `deploy.js` script to deploy chunks
- Updated the `upload-source-maps.js `script to upload chunks maps

## Testing

- [x] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


test/app/webpack.config.js
```diff
@@ -2,7 +2,7 @@ const path = require('path')
 
 module.exports = (_env, argv) => ({
   entry: './app.ts',
-  target: ['web', 'es5'],
+  target: ['node', 'es2018'],
```
amortemousque, line 5: The test app used to test if the SDK works for SSR was targeting the web. I fixed it to target node and es2018.

packages/core/src/browser/runOnReadyState.ts
```diff
@@ -14,3 +14,17 @@ export function runOnReadyState(
   const eventName = expectedReadyState === 'complete' ? DOM_EVENT.LOAD : DOM_EVENT.DOM_CONTENT_LOADED
   return addEventListener(configuration, window, eventName, callback, { once: true })
 }
+
+export function asyncRunOnReadyState(
```
amortemousque, line 18: For now, I introduced an async version of `runOnReadyState` but if we agree on this lazy loading approach I'll refactor the original `runOnReadyState` to use promises.

packages/rum/src/boot/lazyLoadRecorder.ts
```diff
@@ -0,0 +1,3 @@
+export function lazyLoadRecorder() {
+  return import(/* webpackChunkName: "recorder" */ './startRecording').then((module) => module.startRecording)
```
thomas-lebeau, line 2: **❓ question:** ‏ I have a doubt, does this means that customer need to use webpack as well? or is this a stupid question?

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -37,11 +37,17 @@ describe('makeRecorderApi', () => {
 
     lifeCycle = new LifeCycle()
     stopRecordingSpy = jasmine.createSpy('stopRecording')
-    startRecordingSpy = jasmine.createSpy('startRecording').and.callFake(() => ({
-      stop: stopRecordingSpy,
-    }))
+    startRecordingSpy = jasmine.createSpy('startRecording')
+
+    // Workaround because using resolveTo(startRecordingSpy) is not was not working
```
thomas-lebeau, line 42:
```suggestion
    // Workaround because using resolveTo(startRecordingSpy) was not working
```

packages/rum/src/boot/recorderApi.ts
```diff
@@ -186,6 +166,38 @@ export function makeRecorderApi(
...
+          sessionManager,
+          viewHistory,
+          deflateEncoder
+        )
+
+        state = {
+          status: RecorderStatus.Started,
+          stopRecording,
+        }
+      }
```
thomas-lebeau, line 199: **💬 suggestion:** ‏ As we're using promise, I'll avoid having state all over the place. maybe something like this would be a bit more clean:


```ts
const doStart = async () => {
  const [startRecordingImpl] = await Promise.all([
    loadRecorder(),
    asyncRunOnReadyState(configuration, 'interactive'),
  ])

  if (state.status !== RecorderStatus.Starting) {
    return
  }

  const deflateEncoder = getOrCreateDeflateEncoder()
  if (!deflateEncoder) {
    throw new Error('deflate encoder not initialized!');
  }

  const { stop: stopRecording } = startRecordingImpl(
    lifeCycle,
    configuration,
    sessionManager,
    viewHistory,
    deflateEncoder
  )
  
  return stopRecording
}
```

And manage the state when you use it:
```ts
 doStart()
  .then(stop => {
    state = {
      status: RecorderStatus.Started,
      stopRecording: stop
    }
  })
  .catch(() => {
    state = {
      status: RecorderStatus.Stopped,
    }
  })
```

test/app/webpack.config.js
thomas-lebeau: > The test app used to test if the SDK works for SSR was targeting the web.

Is it what is tested? As I understand it is used to test the npm setup and it's injected in a html: https://github.com/DataDog/browser-sdk/blob/7ea12b0df73b84690b7a090695b8a0c06e2ccf31/test/e2e/lib/framework/pageSetups.ts#L153

packages/rum/src/boot/lazyLoadRecorder.ts
amortemousque: Not exactly, I used the `webpackChunkName` special comment to have a fixed chunk name to deploy our CDN files. For customers using NPM:
- if they use webpack: the chunk will have a fix name
- If they use another bundler this comment is ignored

packages/rum/src/boot/recorderApi.ts
amortemousque: Interesting! I'll look at it

packages/core/src/browser/runOnReadyState.ts
```diff
@@ -14,3 +14,17 @@ export function runOnReadyState(
...
+  expectedReadyState: 'complete' | 'interactive'
+): Promise<void> {
+  return new Promise((resolve) => {
+    if (document.readyState === expectedReadyState || document.readyState === 'complete') {
+      resolve()
+    } else {
+      const eventName = expectedReadyState === 'complete' ? DOM_EVENT.LOAD : DOM_EVENT.DOM_CONTENT_LOADED
+      addEventListener(configuration, window, eventName, () => resolve(), { once: true })
+    }
+  })
```
BenoitZugmeyer, line 29: **💬 suggestion:** ‏
```suggestion
  return new Promise((resolve) => {
    runOnReadyState(configuration, expectedReadyState, resolve)
  })
```

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -459,51 +535,54 @@ describe('makeRecorderApi', () => {
 
   describe('isRecording', () => {
     it('is false when recording has not been started', () => {
-      setupRecorderApi()
+      setupRecorderApi({ startSessionReplayRecordingManually: true })
```
BenoitZugmeyer, line 538: **❓ question:** ‏Why do we need this change?

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -514,37 +593,37 @@ describe('makeRecorderApi', () => {
     const VIEW_ID = 'xxx'
 
     it('is undefined when recording has not been started', () => {
-      setupRecorderApi()
+      setupRecorderApi({ startSessionReplayRecordingManually: true })
```
BenoitZugmeyer, line 596: **❓ question:** ‏Here too

packages/rum/src/boot/recorderApi.ts
```diff
@@ -155,29 +155,9 @@ export function makeRecorderApi(
...
-          }
-
-          const { stop: stopRecording } = startRecordingImpl(
-            lifeCycle,
-            configuration,
-            sessionManager,
-            viewHistory,
-            deflateEncoder
-          )
+        doStart().catch(() => {
```
BenoitZugmeyer, line 158: **💭 thought:** ‏It doesn't matter too much here, since the `catch` callback probably won't fail, but I think we'll need to `monitor()` Promise callbacks at the end of promise chains

packages/rum/src/boot/recorderApi.ts
BenoitZugmeyer: I'm not a big fan of throwing errors for control flow. Typescript does not have a syntax to tell whether a function can throw.

Having a `Result` type could be nicer ([ex](https://imhoff.blog/posts/using-results-in-typescript)), as it can be typechecked properly.

Or, you can make `doStart` work as a state machine and return the new state:

```ts
     const doStart = async () => {
        const [startRecordingImpl] = await Promise.all([
          loadRecorder(),
          asyncRunOnReadyState(configuration, 'interactive'),
        ])

        if (state.status !== RecorderStatus.Starting) {
          return state
        }

        const deflateEncoder = getOrCreateDeflateEncoder()
        if (!deflateEncoder) {
          return {
            status: RecorderStatus.Stopped,
          }
        }

        const { stop: stopRecording } = startRecordingImpl(
          lifeCycle,
          configuration,
          sessionManager,
          viewHistory,
          deflateEncoder
        )

        return {
          status: RecorderStatus.Started,
          stopRecording,
        }
      }
```

Not sure if it simplifies things though.

scripts/deploy/deploy.js
```diff
@@ -39,30 +39,48 @@ const uploadPathTypes = process.argv[4].split(',')
...
+  }
+  invalidateCloudfront(awsConfig, cloudfrontPathsToInvalidate)
+})
+
+async function uploadPackage(awsConfig, packageName, chunks) {
+  const cloudfrontPathsToInvalidate = []
+  const bundleFolder = buildBundleFolder(packageName)
+
+  for (const uploadPathType of uploadPathTypes) {
+    for (const chunkName of chunks) {
```
BenoitZugmeyer, line 54: **💬 suggestion:** ‏Instead of hardcoding the chunks here, we could use `fs.readdirSync` and upload all files in the `bundles/` folder.

packages/rum/src/boot/lazyLoadRecorder.ts
BenoitZugmeyer: I don't think we should use a fixed name for chunks though. It should include a hash or something to avoid cache issues, because I don't trust cloudfront/the browser to invalidate both files exactly at the same time. 

We can talk about it.

packages/rum/src/boot/recorderApi.ts
amortemousque: I've worked on a refacto for the recorderApi to mimic what we have in RumPublicAPI:
Having preStart and postStart strategies. I'll create a dedicated PR for it.

packages/core/test/collectAsyncCalls.ts
```diff
@@ -28,3 +28,34 @@ export function collectAsyncCalls<F extends jasmine.Func>(
     fail(`Unexpected extra call for spec '${currentSpec!.fullName}'`)
   }
 }
+export function asyncCollectAsyncCalls<F extends jasmine.Func>(
```
BenoitZugmeyer, line 31: **💭 thought:** ‏Not a big fan of that function name. Maybe we could use the same function as above, but also return a Promise

```ts
export function collectAsyncCalls<F extends jasmine.Func>(
  spy: jasmine.Spy<F>,
  expectedCallsCount: number,
  callback?: (calls: jasmine.Calls<F>) => void
): Promise<jasmine.Calls<F>> {
  ...
}
```

and deprecate `callback` usage so we can remove it one day.

amortemousque: Done!

packages/rum/src/boot/recorderApi.ts
amortemousque: Done. Let me know what you think

packages/core/src/browser/runOnReadyState.ts
amortemousque: Refactoring `runOnReadyState` is quite complexe. Let's keep `asyncRunOnReadyState` for now

packages/rum/src/boot/lazyLoadRecorder.ts
amortemousque: Good point! I changed the chunks format to be 
https://github.com/DataDog/browser-sdk/blob/1acd4ac7309093080d6fb6a0d0efaccb1ae443cf/webpack.base.js#L14

test/app/webpack.config.js
amortemousque: I made two webpack config file:
- webpack.ssr.js used for the ssr test
- webpack.web.js used for e2e tests
commit: https://github.com/DataDog/browser-sdk/pull/3152/commits/1e3d8b51a184565f71a154a0d80707a140ca3393

packages/rum/src/boot/recorderApi.ts
amortemousque: https://github.com/DataDog/browser-sdk/pull/3181

scripts/deploy/deploy.js
amortemousque: Good idea! Done.

packages/rum/src/boot/recorderApi.ts
amortemousque: We can use the promise chain as mentioned as it would make two async tasks. One starting the recorder and another changing the status.

amortemousque: With this refacto PR, things should be clearer: https://github.com/DataDog/browser-sdk/pull/3181

.gitlab-ci.yml
```diff
@@ -309,7 +309,7 @@ deploy-feature:
   stage: deploy
   when: manual
   variables:
-    SUFFIX: 'my-feature' #/datadog-[product]-${SUFFIX}.js
+    SUFFIX: 'lazy-replay' #/datadog-[product]-${SUFFIX}.js
```
thomas-lebeau, line 312: **💬 suggestion:** ‏ You can set the suffix in gitlab when you start a manual job

packages/rum/src/boot/recorderApi.spec.ts
thomas-lebeau: **❓ question:** ‏ You resolved this but I don't understand what `is not was not` means if it's not a typo.

packages/rum/src/boot/postStartStrategy.ts
```diff
@@ -75,42 +100,21 @@ export function createPostStartStrategy(
...
-        lifeCycle,
-        configuration,
-        sessionManager,
-        viewHistory,
-        deflateEncoder
-      ))
-
-      status = RecorderStatus.Started
-    })
+    doStart().catch(monitorError)
```
nulrich, line 103: **💬 suggestion:** ‏Maybe document why we have a promise that we don't await.

.gitlab-ci.yml
amortemousque: Nice, good to know :)

scripts/deploy/upload-source-maps.js
```diff
@@ -33,39 +34,53 @@ function getSitesByVersion(version) {
...
+  }
+  printLog('Source maps upload done.')
+})
+
+async function uploadSourceMaps(packageName, service) {
+  const bundleFolder = buildBundleFolder(packageName)
+
+  for (const uploadPathType of uploadPathTypes) {
+    await forEachFile(bundleFolder, (bundlePath) => {
+      if (!bundlePath.endsWith('.js.map')) {
```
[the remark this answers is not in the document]
amortemousque: What do you mean by filter twice? I considered adding an option to forEachFile, but I felt it was better to keep it simple. Using a condition with endsWith already seems straightforward enough. Wdyt?

packages/rum/src/boot/recorderApi.spec.ts
amortemousque: My bad, sry

scripts/deploy/upload-source-maps.js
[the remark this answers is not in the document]
nulrich: Sorry it was not clear 😅 You do the same filter on file extension also in [scripts/deploy/deploy.js](https://github.com/DataDog/browser-sdk/pull/3152/files/105eb72ee76c32faeaeb3a3e8834205378ae2080#diff-64e275d0732d83f9892d21a0e718046d3ba3e1e09afcf5a305d8038c54cdc97c).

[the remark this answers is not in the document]
amortemousque: It's not the same filter as here I'm filtering source maps and their js files, no?

PR #3154, opened 2024-11-21
[RUM-6563] Attribute ressource URL to LCP

RomanGaignault:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/browser/performanceObservable.ts
+++ b/packages/rum-core/src/browser/performanceObservable.ts
@@ -90,6 +90,7 @@ export interface RumLargestContentfulPaintTiming {
   startTime: RelativeTime
   size: number
   element?: Element
+  url?: string
   toJSON(): Omit<RumLargestContentfulPaintTiming, 'toJSON'>
 }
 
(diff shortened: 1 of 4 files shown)


sethfowler-datadog (commented):
Before making these changes, you'll want to add the new field to `rum-events-format`; it goes [here](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_view-performance-schema.json#L115-L135).

sethfowler-datadog (approved):
LGTM!

nulrich (approved):

RomanGaignault:
> @RomanGaignault @nulrich Thanks for adding LCP resource URL tracking. I'm noticing that `@view.performance.lcp.resource_url` is consistently missing in the dashboard despite these changes being merged. Could you help clarify:
> 
> 1. Is there any additional configuration needed to expose this data?
> 2. Are there specific conditions where the LCP resource URL might not be captured?
> 
> Any guidance would be appreciated as this data would be valuable for our performance monitoring 🙏

Hello, we are currently adding it from our side to view and session, it should be available shortly. You don't have to do anything on your side 👍

ramigluzman:
@RomanGaignault @nulrich
Hi, I'm noticing that 96% of views in our RUM dashboard show "N/A" for this data point. Only about 4% of views actually display an LCP resource URL.

Is this expected behavior? I'm wondering if there are specific conditions required for this data to be captured properly, or if this high percentage of N/A values is normal.

Thanks for any insights!

BenoitZugmeyer:
It sounds like in your app, the LCP is not caused by an image. See https://developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint/url

PR #3159, opened 2024-11-25
♻️ use playwright for e2e

thomas-lebeau:
## Motivation

Improve reliability and speed of our e2e tests.

## Changes

We now uses [Playwright](https://playwright.dev/) as a e2e test runner. There is small changes to the API but things remains very similar.

### Speed 🚀
|        | Local  | CI     | BrowserStack<exp>*</exp> |
|--------|--------|--------|--------|
| Before | ~3m50  | ~8m    | ~24m   |
| After  | ~21s   | ~3m75s | ~9m    | 

(<exp>*</exp>) BrowserStack timing does not account for the time spend waiting for BS resources. 

Playwright already allow e2e to be faster, however there is a few other optimizations:
- `unit-bs` and `e2e-bs` jobs run as soon as `unit` and `e2e` jobs are finished respectively. This is so we don't wait for the slower `test-performance` job.
- `e2e-bs` job will retry each test up to two time to be more resilient for flakiness, however, the job will stop as soon as a single test fails in order to free up browserstack resources.

### new useful commands:
- `yarn test:e2e --ui` run playwright locally in [UI mode](https://playwright.dev/docs/test-ui-mode)
- `yarn test:e2e -g "unhandled rejections"` to grep tests by name
- `yarn test:e2e --debug` to playwright in [debug mode](https://playwright.dev/docs/debug#run-in-debug-mode-1)
- `yarn test:e2e --project firefox` to run tests in firefox. available projects: `chromium` (default), `firefox`, `webkit`, and `*` (all)
- `yarn test:e2e --repeat-each 3` to run each test 3 times, useful for catching flaky tests
- `BS_USERNAME=johdoe_abcD BS_ACCESS_KEY=askldjhfn yarn test:e2e:bs` to run the tests in Browserstack

> [!IMPORTANT]
> `yarn test:e2e` does not build the SDK automatically, run `yarn build:bundle` if you have made change to the source code of the SDK.

### new useful method:
- `expect([1, 2, 3]).toHaveLength(3)`, will show a better error message when failing, including the actual array content.
- `await page.pause()` to use with `--debug`

## Testing

Locally, run `yarn test:e2e --ui`

Screenshot 2024-11-25 at 10 05 22

- [x] Local
- [ ] Staging
- [ ] Unit
- [x] End to end

---


.gitlab-ci.yml
```diff
@@ -219,24 +219,17 @@ e2e:
...
   artifacts:
     when: always
-    paths: ['test-report/e2e/specs.log']
+    paths:
+      - 'test-report/e2e/*'
     reports:
       junit: test-report/e2e/*.xml
   script:
     - yarn
+    - yarn playwright install --with-deps
```
thomas-lebeau, line 228: Unfortunately playwright recommends to [NOT cache browsers](https://playwright.dev/docs/ci#caching-browsers)

This is not ideal because it takes about as much time to install dependencies and browsers than running the tests in the CI.
I believe we can get around that by having pre-build Docker image.

test/e2e/lib/framework/createTest.ts
```diff
@@ -182,53 +189,97 @@ function declareTestsForSetups(
...
-    expect(logs.filter((l) => (l as any).level === 'SEVERE')).toEqual([])
+  withBrowserLogs((logs) => {
+    expect(logs.filter((log) => log.level === 'error')).toHaveLength(0)
   })
+}
+
+async function tearDownTest({ flushEvents, deleteAllCookies }: TestContext) {
+  await flushEvents()
   await deleteAllCookies()
 }
```
thomas-lebeau, line 285: Split `tearDownTest` and `tearDownPassedTest` so that when a test fails no more `expect`  than the one failing is executed. This is because if the log shows only the last failing expect, hence if the teardown expect fails, it will show only this log which make debugging failing test difficult.

test/e2e/lib/helpers/browser.ts
```diff
@@ -1,126 +1,74 @@
...
-  source: string
-  timestamp: number
-}
+    if (filteredLogs.length !== this.logs.length) {
+      // FIXME: fix this at the perfomance observer level as it is visible to customers
+      // It used to pass before because it was only happening in Firefox but wdio io did not support console logs for FF
+      test.info().annotations.push({
+        type: 'dd_tags[test.fixme]',
+        description: 'Unnexpected Console log message: "Ignoring unsupported entryTypes: *"',
+      })
```
thomas-lebeau, line 27: This add the property `@test.fixme` in CI visibility, this is very useful to create dashboards

test/e2e/browsers.conf.js
```diff
@@ -1,37 +1,30 @@
...
     version: '100.0',
     os: 'Windows',
     osVersion: '11',
   },
   {
     sessionName: 'Firefox',
-    name: 'Firefox',
-    version: '91.0',
+    name: 'playwright-firefox',
+    version: '119',
```
amortemousque, line 17: **❓ question:** What is the reasoning behind this version bump?‏

thomas-lebeau: Older version of firefox don't work with this version of playwright.

The version we used to test was already chosen arbitrary and did not reflect the oldest version supported.
For unit tests however we [test on the oldest browser version we support](https://github.com/DataDog/browser-sdk/blob/d0c5aac23e297f535acd13d24eceb9e374e978f7/test/unit/browsers.conf.js)

test/e2e/scenario/recorder/viewports.scenario.ts
```diff
@@ -15,38 +18,37 @@ const VIEWPORT_META_TAGS = `
...
+  test.describe('layout viewport properties', () => {
     createTest('getWindowWidth/Height should not be affected by pinch zoom')
       .withRum()
       .withSetup(bundleSetup)
       .withBody(html`${VIEWPORT_META_TAGS}`)
-      .run(async ({ intakeRegistry }) => {
-        await buildScrollablePage()
+      .run(async ({ intakeRegistry, page, flushEvents }) => {
+        const { sessionName } = test.info().project.metadata as BrowserConfiguration
+        test.fixme(sessionName === 'Edge', 'In Edge, the ViewportResize record data is off by almost 20px')
```
BenoitZugmeyer, line 33: **💬 suggestion:** ‏we should have a single way to identify which browser is being used.

Previously, we had the `getBrowserName` function that returned a normalized browser name. It returned a strongly typed constant string, making it easy to understand what to expect. Now we have (at least) three ways of getting the browser name:
* `testContext.browserName === 'chromium'`
* `testContext.browserName.includes("firefox")`
* `test.info().project.metadata.sessionName === 'Edge'`

In the same spirit, I think `testContext.browserName` should:
* be strongly typed, with constant strings instead of arbitrary strings
* support all browsers (firefox, edge, ...)
* normalize the names (ex: `playwright-firefox` should be normalized to `firefox`)

And we should avoid having to pick into `test.info()` to get that information

test/e2e/scenario/rum/s8sInject.scenario.ts
BenoitZugmeyer, line 1: I don't think this *needs* to be within the e2e test suite. If it makes things more complicated, we could imagine moving it somewhere else (in a standalone script).

test/e2e/scenario/rum/tracing.scenario.ts
```diff
@@ -19,19 +19,22 @@ describe('tracing', () => {
...
+            window
+              .fetch('/headers', {
+                headers: [
+                  ['x-foo', 'bar'],
+                  ['x-foo', 'baz'],
+                ],
+              })
+              .then((response) => response.text())
+              .then(resolve)
+              .catch(() => resolve(new Error('Fetch request failed!')))
```
BenoitZugmeyer, line 35: **💬 suggestion:** ‏Don't use `new Promise` when you already have promises:

```ts
const rawHeaders = await page.evaluate(
        () =>
            window
              .fetch('/headers', {
                headers: [
                  ['x-foo', 'bar'],
                  ['x-foo', 'baz'],
                ],
              })
              .then((response) => response.text())
              .catch(new Error('Fetch request failed!'))
      )
```

(Same suggestion applies to other `new Promise` in that file)

test/e2e/lib/framework/createTest.ts
```diff
@@ -182,53 +190,106 @@ function declareTestsForSetups(
...
+      try {
+        cb(browserLogsManager.get())
+      } finally {
+        browserLogsManager.clear()
+      }
+    },
+    flushBrowserLogs: () => browserLogsManager.clear(),
+    flushEvents: () => flushEvents(page),
+    deleteAllCookies: () => deleteAllCookies(browserContext),
+    sendXhr: (url: string, headers?: string[][]) => sendXhr(page, url, headers),
```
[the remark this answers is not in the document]
thomas-lebeau: Good question! Playwright encourage to use fixtures, so adding a function to TestContext fit well with Playwright style.

However, I think there is room for both:
- For heavily used functions, such as `flushEvents`, `withBrowserLogs`, I think it's convenient to not have to pass `page` or `browserContext` all the time.
- For function that are less often used or used only in one test file, then I think it might be better to have standalone function.

I think `deleteAllCookies` is not actually a good example as it's just doing `browserContext.clearCookies()` so we might use that directly.

At the end, it's a bit inconsistent in this PR as multiple people have migrated tests without clear direction.

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -21,10 +21,7 @@ export interface SetupOptions {
 
 export type SetupFactory = (options: SetupOptions, servers: Servers) => string
 
-const isBrowserStack =
-  'services' in browser.options &&
-  browser.options.services &&
-  browser.options.services.some((service) => (Array.isArray(service) ? service[0] : service) === 'browserstack')
+const isBrowserStack = process.env.BS_USERNAME && process.env.BS_ACCESS_KEY
```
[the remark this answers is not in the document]
thomas-lebeau: we certainly could detect browserstack differently here, but `BS_USERNAME` and `BS_ACCESS_KEY` are populated as env variable anyway and used for `unit-bs` job too.

I thought it would be convenient to use that as a signal test are running on BS.

test/e2e/lib/helpers/browser.ts
```diff
@@ -1,126 +1,69 @@
...
+  get() {
+    const filteredLogs = this.logs.filter((log) => !log.message.includes('Ignoring unsupported entryTypes: '))
 
-interface BrowserLog {
-  level: string
-  message: string
-  source: string
-  timestamp: number
-}
+    if (filteredLogs.length !== this.logs.length) {
```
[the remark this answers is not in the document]
thomas-lebeau: Unfortunately, we need to actually filter out these warning message, otherwise a bunch of test are failing because they expect logs to be a certain length.

Ultimately this need to be fixed in the observer as this message is shown to customers (but this seems out of scope for this PR)

test/e2e/scenario/recorder/viewports.scenario.ts
thomas-lebeau: Actually it is normalized ([here](https://github.com/DataDog/browser-sdk/blob/2c1e3f85cb0dd60c3da89794d28db8fc27cf3300/test/e2e/playwright.bs.config.ts#L15)) because `browserName` is a playwright property, and it's type is `'chromium' | 'firefox' | 'webkit'`

About `testContext.browserName.includes("firefox")`, it is a mistake, I'm fixing it to `testContext.browserName === 'firefox'` 👌 

However, Edge browsers have `chromium` for `browserName`, but the viewport issue is really on Edge, not on all chromium browsers, hence this ugly trick. 😞 
I guess we could normalize `browserName` in TestContext differently to playwright's meaning. That would not be too much of a hassle. wdyt?

BenoitZugmeyer: > differently to playwright's meaning
Yes, I think it would be good!

test/e2e/scenario/rum/s8sInject.scenario.ts
thomas-lebeau: I think it's actually more convenient to have it run in the e2e, we avoid setting up a separate job, script, test runner, ...

amortemousque (approved):

test/e2e/lib/framework/pageSetups.ts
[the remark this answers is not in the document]
thomas-lebeau: > Maybe using `process.argv`

I ended up passing a different env variable that is set in `bs-wrapper.js`.
I could not use `process.argv` because playwright spawn separate processes so they don't have access to the original comand's args

test/e2e/lib/framework/createTest.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Sounds good.

> At the end, it's a bit inconsistent in this PR as multiple people have migrated tests without clear direction.

Should we make things more consistent in this PR?

test/e2e/lib/helpers/validation.ts
```diff
@@ -14,8 +15,6 @@ export function validateRumFormat(events: RumEvent[]) {
 
   events.forEach((rumEvent) => {
     void instance.validate('rum-events-schema.json', rumEvent)
-    if (instance.errors) {
-      instance.errors.map((error) => fail(`event/${error.instancePath || ''} ${error.message!}`))
-    }
+    expect(instance.errors).toBe(null)
```
BenoitZugmeyer, line 18: **❓ question:** ‏what was the issue with this? formating error messages helps a bit to understand what's wrong

test/e2e/lib/framework/pageSetups.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Thank you!

test/e2e/lib/helpers/browser.ts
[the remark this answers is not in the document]
BenoitZugmeyer: Oh I understand now why you did it this way. I guess the previous version was fine, sorry. Well, in any case, we should remove this altogether once this warning is fixed.

test/e2e/lib/helpers/validation.ts
thomas-lebeau: Because `fail` is a jasmine thing (or at least not a playwright thing) and does not exist anymore.

But I guess I can find an alternative way of doing the same

BenoitZugmeyer: Ah ok.

Maybe:

```ts
const formattedErrors = instance.errors.map((error) => fail(`event/${error.instancePath || ''} ${error.message!}`)).join('\n')
throw new Error(`Format validation failure:\n${formattedErrors}`)

```

BenoitZugmeyer (approved):

PR #3161, opened 2024-11-25
✨ [RUM-6814] strongly type site parameter

thomas-lebeau:
## Motivation

Strongly type the site parameter, this will allow customer using typescript to have better autocompletion and avoid misconfiguration. This is not a bulletproof solution, especially if user are not using typescript, but it will help the ones that do.

If we were to add new datacenters, we would need to add them to the type definition, release a new minor and have customer update to the latest SDK.

Customer wanting to use the new datacenters without using the new SDK will still have the option to ignore the type error using `// @ts-expect-error` or `as any`

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/intakeSites.ts
+++ b/packages/core/src/domain/configuration/intakeSites.ts
@@ -6,10 +6,8 @@ export type Site =
   | 'ddog-gov.com'
   | 'ap1.datadoghq.com'
 
-export type SiteWithStaging = Site | 'dd0g-gov.com' | 'datad0g.com'
-
-export const INTAKE_SITE_STAGING: SiteWithStaging = 'datad0g.com'
-export const INTAKE_SITE_FED_STAGING: SiteWithStaging = 'dd0g-gov.com'
+export const INTAKE_SITE_STAGING: Site = 'datad0g.com' as Site
+export const INTAKE_SITE_FED_STAGING: Site = 'dd0g-gov.com' as Site
 export const INTAKE_SITE_US1: Site = 'datadoghq.com'
 export const INTAKE_SITE_EU1: Site = 'datadoghq.eu'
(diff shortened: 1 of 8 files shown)


packages/core/src/domain/configuration/intakeSites.ts
```diff
@@ -1,8 +1,16 @@
...
+export type Site =
+  | 'datadoghq.com'
+  | 'us3.datadoghq.com'
+  | 'us5.datadoghq.com'
+  | 'datadoghq.eu'
+  | 'ddog-gov.com'
+  | 'ap1.datadoghq.com'
+
+export const INTAKE_SITE_STAGING: Site = 'datad0g.com' as Site
+export const INTAKE_SITE_FED_STAGING: Site = 'dd0g-gov.com' as Site
```
thomas-lebeau, line 10: I don't want to expose these in the autocompletion, they are not public facing

packages/rum/src/domain/getSessionReplayLink.spec.ts
```diff
@@ -5,7 +5,7 @@ import { getSessionReplayLink } from './getSessionReplayLink'
 import { addRecord, resetReplayStats } from './replayStats'
 
 const DEFAULT_CONFIGURATION = {
-  site: 'datad0g.com',
+  site: 'datadoghq.com',
```
thomas-lebeau, line 8: `datad0g.com` don't satisfies `Site`, We could cast it as `Site` but I don't think there is any reason not to use the production site in the unit tests

BenoitZugmeyer (approved):

packages/core/src/domain/configuration/intakeSites.ts
amortemousque: **👏 praise:** ‏Good catch!

nulrich (approved):

PR #3162, opened 2024-11-25
✨ [RUM-6182] don't start recording automatically when replay sample is 0

BenoitZugmeyer:
## Motivation

We had some customer cases where they have seen some sessions being recorded even if they were specifically set `sessionReplaySampling: 0`. This was caused by another SDK instance (injected via an extension) setting the session type to "replay", so it worked "as expected" but was still pretty confusing to the customer as they specifically didn't want to collect replay (for privacy reasons).

## Changes

Set the default value of `startSessionReplayRecordingManually` to `true` if `sessionReplaySampleRate` is `0`, keep it `false` otherwise.

Technically, this is a small breaking change: a customer could want to record sessions only for users visiting a given page `/foo`. In that case, they could set a `sessionReplaySampleRate: 0`, and only in `/foo` they would run `startSessionReplayRecording({ force: true })`. After that they would expect that subsequent pages are recorded (even with `sessionReplaySampleRate: 0` ).

In practice, it seems this usecase is not used by our customers based on our telemetry.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/boot/rumPublicApi.spec.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.spec.ts
@@ -652,15 +652,15 @@ describe('rum public api', () => {
 
     it('is started with the default startSessionReplayRecordingManually', () => {
       rumPublicApi.init(DEFAULT_INIT_CONFIGURATION)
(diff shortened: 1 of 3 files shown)


nulrich (approved):

sethfowler-datadog (approved):
LGTM!

PR #3166, opened 2024-11-26
✨ [RUM-6799] Add new delivery type property

RomanGaignault:
## Motivation

The “delivery type” has been introduced on resources performance entries: [PerformanceResourceTiming: deliveryType property - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/deliveryType)

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/resource/resourceUtils.ts
+++ b/packages/rum-core/src/domain/resource/resourceUtils.ts
@@ -12,7 +12,7 @@ import {
 
 import type { RumPerformanceResourceTiming } from '../../browser/performanceObservable'
 
-import type { DeliveryType, ResourceEntryDetailsElement } from '../../rawRumEvent.types'
+import type { ResourceEntryDetailsElement, DeliveryType } from '../../rawRumEvent.types'
 
 export interface ResourceEntryDetails {
   worker?: ResourceEntryDetailsElement
@@ -195,11 +195,11 @@ export function computeResourceEntryProtocol(entry: RumPerformanceResourceTiming
 
 /**
  * Handles the 'deliveryType' property to distinguish between supported values ('cache', 'navigational-prefetch'),
- * undefined (unsupported in some browsers), and other cases ('-' for unknown or unrecognized values).
+ * undefined (unsupported in some browsers), and other cases ('other' for unknown or unrecognized values).
  * see: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/deliveryType
  */
 export function computeResourceEntryDeliveryType(entry: RumPerformanceResourceTiming): DeliveryType | undefined {
-  return entry.deliveryType === '' ? 'other' : (entry.deliveryType as DeliveryType)
+  return entry.deliveryType === '' ? 'other' : entry.deliveryType
 }
 
 export function computeResourceEntrySize(entry: RumPerformanceResourceTiming) {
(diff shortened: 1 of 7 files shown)


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -193,6 +193,25 @@ export function computeResourceEntryProtocol(entry: RumPerformanceResourceTiming
   return entry.nextHopProtocol === '' ? undefined : entry.nextHopProtocol
 }
 
+/**
+ * Handles the 'deliveryType' property to distinguish between supported values ('cache', 'navigational-prefetch'),
+ * undefined (unsupported in some browsers), and other cases ('-' for unknown or unrecognized values).
+ * https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/deliveryType
```
thomas-lebeau, line 199: **🥜 nitpick:** 
```suggestion
 * see: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/deliveryType
```

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -193,6 +193,25 @@ export function computeResourceEntryProtocol(entry: RumPerformanceResourceTiming
...
+
+  if (deliveryType === undefined) {
+    return undefined
+  }
+
+  if (deliveryType === 'cache' || deliveryType === 'navigational-prefetch') {
+    return deliveryType
+  }
+
+  return '-'
```
thomas-lebeau, line 212: **❓ question:** ‏ Why using `'-'` instead of the standard `''` ?

RomanGaignault: To indicate the difference between, we don't have the information and undefined(browser doesn't support it). We want to differentiate both cases and not just let empty for both

thomas-lebeau: ok, but Chrome is already differentiating `undefined` with `''`. Why don't we keep `''`?

Any other place we're using `'-'`?

BenoitZugmeyer: We discussed about it, and having a blank field is not very user friendly, as they won't be able to distinguish between an empty string and a missing field in some cases (ex: when showing that field in a column). This is why I suggested `'-'`. An alternative could be `'other'`, maybe it could be more explicit.

BenoitZugmeyer, line 212: **💬 suggestion:** ‏It could be simpler and more future proof:
```suggestion
  return entry.deliveryType === '' ? '-' : entry.deliveryType
```

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -185,6 +186,8 @@ export interface RawRumLongTaskEvent {
   }
 }
 
+export type deliveryType = 'cache' | 'navigational-prefetch' | '-' | undefined
```
BenoitZugmeyer, line 189: **💬 suggestion:** ‏types should be capitalized, and no need to have `undefined`
```suggestion
export type DeliveryType = 'cache' | 'navigational-prefetch' | '-'
```

packages/rum-core/src/domain/resource/resourceUtils.ts
thomas-lebeau: 👍 makes sense

Another alternative could be `'n/a'` for _not available_.

thomas-lebeau (approved):

BenoitZugmeyer (approved):

PR #3181, opened 2024-11-28
♻️ [RUM-6813] Split the recorder API module

amortemousque:
## Motivation

The recorderApi module is a bit large and we want to add more logic before starting to lazy load the recorder. It is a good time to extract the "pre start" logic like we did with RUM and Logs: https://github.com/DataDog/browser-sdk/pull/2575.

## Changes

- Introduce a `Strategy` interface for behaviors that differs before and after starting the SDK.
- Introduce a `createPreStartStrategy` to initialize the strategy before the SDK is started
- Introduce a `createPostStartStrategy` to update the strategy after the SDK is started

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/boot/postStartStrategy.ts
+++ b/packages/rum/src/boot/postStartStrategy.ts
@@ -1,12 +1,12 @@
-import {
-  LifeCycleEventType,
-  SessionReplayState,
-  type LifeCycle,
-  type RumConfiguration,
-  type RumSessionManager,
-  type StartRecordingOptions,
-  type ViewHistory,
+import type {
+  LifeCycle,
+  RumConfiguration,
+  RumSessionManager,
+  StartRecordingOptions,
+  ViewHistory,
+  RumSession,
 } from '@datadog/browser-rum-core'
+import { LifeCycleEventType, SessionReplayState } from '@datadog/browser-rum-core'
 import { PageExitReason, runOnReadyState, type DeflateEncoder } from '@datadog/browser-core'
 import { getSessionReplayLink } from '../domain/getSessionReplayLink'
 import type { startRecording } from './startRecording'
@@ -64,12 +64,12 @@ export function createPostStartStrategy(
 
   function start(options?: StartRecordingOptions) {
     const session = sessionManager.findTrackedSession()
-    if (!session || (session.sessionReplay === SessionReplayState.OFF && (!options || !options.force))) {
+    if (canStartRecording(session, options)) {
       status = RecorderStatus.IntentToStart
       return
     }
 
-    if (status === RecorderStatus.Starting || status === RecorderStatus.Started) {
+    if (isRecordingInProgress(status)) {
       return
     }
 
@@ -97,17 +97,15 @@ export function createPostStartStrategy(
       status = RecorderStatus.Started
     })
 
-    if (options && options.force && session.sessionReplay === SessionReplayState.OFF) {
+    if (shouldForceReplay(session!, options)) {
       sessionManager.setForcedReplay()
     }
   }
 
   function stop() {
-    if (status === RecorderStatus.Stopped) {
-      return
+    if (status !== RecorderStatus.Stopped && status === RecorderStatus.Started) {
+      stopRecording?.()
     }
-
-    stopRecording?.()
 
     status = RecorderStatus.Stopped
   }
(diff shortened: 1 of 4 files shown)


packages/rum/src/boot/postStartStrategy.ts
```diff
@@ -0,0 +1,103 @@
...
+  IntentToStart,
+  // The recorder is starting. It does not record anything yet.
+  Starting,
+  // The recorder is started, it records the session.
+  Started,
+}
+export type RecorderState = {
+  status: RecorderStatus
+  setStatus: (status: RecorderStatus) => void
+  statusEqual: (status: RecorderStatus) => boolean
```
sethfowler-datadog, line 29: Nit: since a `RecorderState` must provide a gettable `status` anyway, would it work just as well to remove `statusEqual` and use `status ===` instead?

packages/rum/src/boot/postStartStrategy.ts
```diff
@@ -0,0 +1,103 @@
...
+        sessionManager.setForcedReplay()
+      }
+    },
+
+    stopStrategy() {
+      if (state.statusEqual(RecorderStatus.Stopped)) {
+        return
+      }
+
+      stopRecording?.()
```
sethfowler-datadog, line 94: The original version of this code did this:
```ts
        if (state.status === RecorderStatus.Stopped) {
          return
        }
          state.stopRecording()
        }
```

Since `RecorderState` was a discriminated union, it was necessary to check that `state.status` was exactly `RecorderStatus.Started` before calling `stopRecording()`, so we were guaranteed to only ever call `stopRecording()` in that state.

Now, we check that `stopRecording` is truthy, but we don't verify that we're in state `RecorderStatus.Started`. We set `stopRecording` when we _enter_ `RecorderStatus.Started`, but we don't clear it when we _exit_ `RecorderStatus.Started`, so checking if it's truthy just tells us that we've been in `RecorderStatus.Started` at some point in the past.

It's possible that the overall flow of the code means that this isn't an issue, but from a belt-and-suspenders perspective it feels like it might be best to reintroduce some sort of check here that we're in the expected state. Perhaps something like this?
```suggestion
      if (state.statusEqual(RecorderStatus.Started)) {
        const stopRecordingNow = stopRecording
        stopRecording = undefined
        stopRecordingNow?.()
      }
```

This way we'd both ensure that we're in the expected state, and clear `stopRecording` so that it's not accidentally reused later or invoked in a reentrant manner. (Similar guarantees to what we got from the discriminated union before, in other words.)

packages/rum/src/boot/recorderApi.ts
```diff
@@ -236,4 +88,81 @@ export function makeRecorderApi(
...
+      state
+    )
+
+    if (state.status === RecorderStatus.IntentToStart) {
+      strategies.startStrategy()
+    }
+  }
+}
+
+export function createPreStartStrategy(state: RecorderState): Strategy {
```
sethfowler-datadog, line 158: Trivial nit: from a code organization perspective, I feel like if I see that `createPostStartStrategy()` lives in `postStartStrategy.ts`, I might expect this function to live in `preStartStrategy.ts`. Moving it there might imply putting `RecorderState`, `RecorderStatus`, and `Strategy` in a separate file as well, but again doing that might make it a bit easier to find them. Definitely not a strongly held view if you prefer this way of organizing things, though!

amortemousque: Done!

packages/rum/src/boot/postStartStrategy.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export function createPostStartStrategy(
+  configuration: RumConfiguration,
+  lifeCycle: LifeCycle,
+  sessionManager: RumSessionManager,
+  viewHistory: ViewHistory,
+  startRecordingImpl: StartRecording,
+  getOrCreateDeflateEncoder: () => DeflateEncoder | undefined,
+  state: RecorderState
```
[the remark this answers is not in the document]
amortemousque: Yes it's nice! I made few changes let me know what you think

packages/rum/src/boot/postStartStrategy.ts
sethfowler-datadog: If I'm reading that correctly, it boils down to:
```ts
if (status === RecorderStatus.Started) {
  stopRecording?.()
}

status = RecorderStatus.Stopped
```

I think that's OK; it's quite similar to the old behavior! We do leave the old `stopRecording` value around, but that's probably OK, since if we reenter `RecorderStatus.Started`, we'll overwrite it.

amortemousque: Oh my 🤦‍♂️. Yes, let's go with 
```ts
if (status === RecorderStatus.Started) {
  stopRecording?.()
}

status = RecorderStatus.Stopped
```

sethfowler-datadog (approved):

packages/rum/src/boot/postStartStrategy.ts
sethfowler-datadog: SGTM!

thomas-lebeau (approved):

PR #3190, opened 2024-11-29
💥 new default for track* initialization options

thomas-lebeau:
## Motivation

`trackResources`, `trackLongTasks`, and `trackUserInteractions` are set by default in the snippet we provide in the Application configuration page and in the [documentation](https://docs.datadoghq.com/real_user_monitoring/browser/setup/client?tab=rum#npm) configuration 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/configuration/configuration.ts
+++ b/packages/rum-core/src/domain/configuration/configuration.ts
@@ -115,6 +115,7 @@ export interface RumInitConfiguration extends InitConfiguration {
   trackViewsManually?: boolean | undefined
   /**
    * Enables collection of resource events.
+   * @default true
    */
   trackResources?: boolean | undefined
   /**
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -115,6 +115,7 @@ export interface RumInitConfiguration extends InitConfiguration {
   trackViewsManually?: boolean | undefined
   /**
    * Enables collection of resource events.
+   * @default true
```
nulrich, line 118: **❓ question:** ‏Why only document this one with a `@default` ?

thomas-lebeau: no reasons, fixed!

nulrich (approved):

PR #3197, opened 2024-12-05
♻️ move traceSampleRate default to config validation

thomas-lebeau:
## Motivation

Two small things to make code easier to read:
- Factorize `injectHeadersIfTracingAllowed` to use less negative expression
- Move setting the default of `traceSampleRate` to `validateAndBuildRumConfiguration` so it's consistent with other default (e.g. `sessionSampleRate`, `sessionReplaySampleRate`, ...) and it's easier to know what is the default value without having to go in the implementation of the tracer.

## Changes

I needed to calculate `rulePsr` at the validation time too because it need to know explicitly if `traceSampleRate` is defined by the customer (i.e. `undefined` vs `100`)

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -122,10 +121,11 @@ function injectHeadersIfTracingAllowed(
     return
   }
   const traceId = createTraceIdentifier()
-  context.traceSampled =
-    !isNumber(configuration.traceSampleRate) || isTraceSampled(traceId, configuration.traceSampleRate)
```
amortemousque, line 126: **❓ question:** ‏Is it okay to remove `!isNumber(configuration.traceSampleRate)`?

thomas-lebeau: Yes it is ok, because it is already done in [configuration.ts](https://github.com/DataDog/browser-sdk/blob/166489ddc05b31339571d57d12d7cbf298bb441a/packages/rum-core/src/domain/configuration/configuration.ts#L173) with `isSampleRate()` which check if it's a valid value for a sample rate or `undefined`.

And later we set the default value to `100` if it was `undefined` where before we would still keep it undefined and sample the trace when `!isNumber(undefined)`

amortemousque (approved):

cy-moi (approved):

PR #3215, opened 2024-12-12
🐛 [RUM-6322] Use window.open observable

nulrich:
## Motivation

`window.open` instrumentation/de-instrumentation in actions collection is stacking instrumented method.

## Changes

Instrument window.open only once when SDK is loading.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/browser/windowOpenObservable.ts
```diff
@@ -0,0 +1,9 @@
+import { instrumentMethod, Observable } from '@datadog/browser-core'
+
+export function createWindowOpenObservable() {
+  const observable = new Observable<void>()
+  instrumentMethod(window, 'open', () => {
+    observable.notify()
+  })
+  return observable
+}
```
thomas-lebeau: **💬 suggestion:** ‏ You solution works, but we could improve it to
  - avoid instrumenting multiple time unnecessarily in case `createWindowOpenObservable` is called more than once
  - stop intrumenting when the SDK stops (we'll have to adjust things in startRum)


```suggestion
let windowOpenObservable;

export function createWindowOpenObservable() {
  if (windowOpenObservable) {
    return windowOpenObservable;
  }
  
  const observable = new Observable<void>()
  const { stop: stopInstrumentation } = instrumentMethod(window, 'open', () => {
    observable.notify()
  })
  
  return { 
    observable, 
    stop() {
      stopInstrumentation()
      windowOpenObservable = undefined;
    }
  }
}
```

packages/rum-core/src/browser/windowOpenObservable.ts
```diff
@@ -0,0 +1,10 @@
+import { instrumentMethod, Observable } from '@datadog/browser-core'
+
+export function createWindowOpenObservable() {
+  return new Observable<void>((observable) => {
+    const { stop } = instrumentMethod(window, 'open', () => {
+      observable.notify()
+    })
+    return stop
+  })
+}
```
thomas-lebeau, line 10: **💭 thought:** ‏ It might fix the issue because we won't instrumentMethod more than one at a time, but it will instrumentMethod and stop intrumenting a lot (at every click)

I think your previous solution, with some improvement would be better:

```suggestion
let windowOpenObservable;

export function createWindowOpenObservable() {
  if (windowOpenObservable) {
    return windowOpenObservable;
  }
  
  const observable = new Observable<void>()
  const { stop: stopInstrumentation } = instrumentMethod(window, 'open', () => {
    observable.notify()
  })
  
  return { 
    observable, 
    stop() {
      stopInstrumentation()
      windowOpenObservable = undefined;
    }
  }
}
```

packages/rum-core/src/domain/view/viewCollection.spec.ts
```diff
@@ -80,13 +80,15 @@ describe('viewCollection', () => {
...
     const domMutationObservable = new Observable<void>()
+    const windowOpenObservable = new Observable<void>()
     const locationChangeObservable = new Observable<LocationChange>()
 
     const collectionResult = startViewCollection(
       lifeCycle,
       mockRumConfiguration(partialConfiguration),
       location,
       domMutationObservable,
+      pageOpenObserable,
```
sethfowler-datadog, line 91: Nit: "Obserable"

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -18,6 +18,7 @@ export function startViewCollection(
   configuration: RumConfiguration,
   location: Location,
   domMutationObservable: Observable<void>,
+  pageOpenObserable: Observable<void>,
```
sethfowler-datadog, line 21: Nit: "Obserable"

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -34,6 +35,7 @@ export function startViewCollection(
     location,
     lifeCycle,
     domMutationObservable,
+    pageOpenObserable,
```
sethfowler-datadog, line 38: Nit: "Obserable"

packages/rum-core/src/domain/view/viewCollection.spec.ts
sethfowler-datadog: Also, shouldn't this be `windowOpenObservable`, which you create above?

packages/rum-core/src/domain/view/viewCollection.ts
sethfowler-datadog: To be consistent with other places in the code, it might be better to name this `windowOpenObservable`.

packages/rum-core/src/domain/waitPageActivityEnd.spec.ts
```diff
@@ -93,7 +99,7 @@ describe('createPageActivityObservable', () => {
   it('emits an activity event when `window.open` is used', () => {
     spyOn(window, 'open')
     startListeningToPageActivities()
-    window.open('toto')
+    windowOpenObservable.notify()
```
sethfowler-datadog, line 102: With this approach, you're no longer testing that the observable is actually observing `window.open`; you're just telling it directly to notify. That might mean that this test will no longer identify some bugs it might have caught before. Is it possible to switch back to the old approach?

test/e2e/scenario/rum/actions.scenario.ts
```diff
@@ -403,4 +403,36 @@ describe('action collection', () => {
...
+        for (let i = 0; i < 2_500; i++) {
+          click()
+        }
+
+        window.open('foo')
+      </script>
+    `)
+    .run(async () => {
+      await withBrowserLogs((logs) => {
+        // A failing test would have a log with message "Uncaught RangeError: Maximum call stack size exceeded"
```
sethfowler-datadog, line 434: Seems like we should link to a Jira ticket or something with the historical context around this issue.

sethfowler-datadog (approved):
LGTM! Some nits below:

packages/rum-core/src/domain/view/viewCollection.spec.ts
nulrich: Done using @thomas-lebeau update

packages/rum-core/src/domain/view/viewCollection.ts
nulrich: Done using @thomas-lebeau update

packages/rum-core/src/domain/view/viewCollection.ts
nulrich: Done using @thomas-lebeau update

packages/rum-core/src/domain/waitPageActivityEnd.spec.ts
nulrich: True, but before `waitPageActivityEnd` was instrumenting itself `window.open` so it made sense to test by calling `window.open`. Now it relies on an observable implemented outside and I don't want to introduce a dependency for this unit test. WDYT?

thomas-lebeau (approved):

packages/rum-core/src/domain/waitPageActivityEnd.spec.ts
sethfowler-datadog: Yeah, that's a fair point. At a minimum you'd need to call `createWindowOpenObservable()` instead of just using `new Observable<void>()`, I suppose. But it's also true that that code is tested elsewhere, so it would indeed be reasonable to keep the test narrowly scoped and just manually invoke `notify()`. Either approach looks good from my perspective.

If you do stick with manually invoking `notify()`, it may be worthwhile to remove the `spyOn()` call from this test. It serves no purpose anymore, right?

nulrich: I removed the `spyOn()`, nice catch!

amortemousque (approved):

PR #3222, opened 2024-12-17
✅ Add e2e test for telemetry usage

thomas-lebeau:
## Motivation

Telemetry usage e2e test were missing

## Changes

- Add missing e2e test for `telemetryUsage`
- Add comment explaining why we setup `*SampleRate: 100` in the default e2e test config. 

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/test/e2e/lib/framework/createTest.ts
+++ b/test/e2e/lib/framework/createTest.ts
@@ -18,19 +18,23 @@ import { createMockServerApp } from './serverApps/mock'
 const DEFAULT_RUM_CONFIGURATION = {
   applicationId: APPLICATION_ID,
   clientToken: CLIENT_TOKEN,
-  sessionReplaySampleRate: 100,
   defaultPrivacyLevel: DefaultPrivacyLevel.ALLOW,
   trackResources: true,
   trackLongTasks: true,
-  telemetrySampleRate: 100,
-  telemetryConfigurationSampleRate: 100,
   enableExperimentalFeatures: [],
   allowUntrustedEvents: true,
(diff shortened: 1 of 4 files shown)


test/e2e/lib/framework/createTest.ts
```diff
@@ -18,19 +18,23 @@ import { createMockServerApp } from './serverApps/mock'
...
   trackLongTasks: true,
-  telemetrySampleRate: 100,
-  telemetryConfigurationSampleRate: 100,
   enableExperimentalFeatures: [],
   allowUntrustedEvents: true,
+  // Force All sample rates to 100% to avoid flakiness
+  sessionReplaySampleRate: 100,
+  telemetrySampleRate: 100,
+  telemetryUsageSampleRate: 100,
+  telemetryConfigurationSampleRate: 100,
```
thomas-lebeau, line 30: I think ideally we should have test that are more resilient to sample rates:
- only set them where needed (when we test a feature related to it)
- explicitly trigger an event instead of relying on telemety event:

for example, in [here](https://github.com/DataDog/browser-sdk/blob/6f25274787c137b7846eb251083171a20fde3382/test/e2e/scenario/trackingConsent.scenario.ts#L96-L96) the logs SDK is considered started because we have a telemetry event for the usage of `setTrackingConsent()` and it works because we force telemetry sample rate to 100%, but that's not very explicit in the test itself

sethfowler-datadog (approved):
LGTM; nice addition!

PR #3228, opened 2024-12-18
🐛 AddError should support all instances of type Error

lightswitch05:
## Motivation

Recently, all errors of type [AxiosError](https://github.com/axios/axios/blob/ce350f1e2163a7c3b7dec874fc9991038f0dc4f9/lib/core/AxiosError.js#L16-L34) stopped logging properly. Instead of the standard log with Error object, it was serializing the error object instead.

After some investigation I tracked down the regression to #3144 - which changes the object type check from `param instanceof Error` to `Object.prototype.toString.call(error) === '[object Error]'`. Unfortunately, the way AxiosError is created, that toString call just produces `[object Object]`. However, the `instanceof` check works just fine.

Image 12-17-24 at 6 03 PM

## Changes

I've addressed the regression by using both the original `instanceof Error` check as well as the new `toString` check. Supporting both checks cover both the old and the new feature requirements.

## Testing

I spent some time trying to get the tests running - but I was not successful. I don't doubt there might be some tweaks required in the PR, but I think the bulk of it is complete. Please feel free to modify the PR as needed. Hopefully this can get fixed quickly and we can all have AxiosError logging working again.

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/error/error.ts
+++ b/packages/core/src/domain/error/error.ts
@@ -88,7 +88,7 @@ export function getFileFromStackTraceString(stack: string) {
 }
 
 export function isError(error: unknown): error is Error {
-  return Object.prototype.toString.call(error) === '[object Error]'
+  return error instanceof Error || Object.prototype.toString.call(error) === '[object Error]'
 }
 
 export function flattenErrorCauses(error: ErrorWithCause, parentSource: ErrorSource): RawErrorCause[] | undefined {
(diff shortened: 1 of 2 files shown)


RomanGaignault (approved):

cy-moi (approved):

RomanGaignault:
@lightswitch05 the pr is merged, we will do a release soon :)

lightswitch05:
image

Just looping back to confirm AxiosError logging is back! Thank you for the quick response! 💪 🥳

PR #3231, opened 2024-12-18
✅ Make tests robust to an instrumentMethod that does not always rewrap

sethfowler-datadog:
## Motivation

We've observed that `instrumentMethod()` is easy to misuse in a way that can trigger a stack overflow. #3214 is a draft PR with more details and one possible fix; we're exploring other approaches to fixing the issue as well. Regardless of the fix we eventually decide to go with, the solution will involve moving to a model where we retain a single, stable observer attached to each global method we instrument, replacing the current approach where we dynamically reinstrument these methods in various places in the code base.

The changes in the draft PR revealed that we have quite a few tests that break if we switch to a single-observer model for these instrumented methods. These tests make assumptions about the implementation details of our instrumentation; in particular, they assume that instrumentation can be removed from methods by overwriting them with a jasmine spy or some other known function, and that a method which is already instrumented will be re-wrapped if the instrumentation code is rerun.

This PR adjusts these tests to handle this situation better. With these changes to the tests, #3214 is unblocked. If we decide on another approach to fix the underlying problem, and describe to discard #3214, these changes will still be helpful, since any fix for the underlying problem is likely to expose similar issues.

## Changes

This is a test-only change; there are no changes to the browser SDK code itself.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/core/src/domain/error/trackRuntimeError.spec.ts
```diff
@@ -1,48 +1,40 @@
...
-  let notifyError: jasmine.Spy
-  let stopRuntimeErrorTracking: () => void
-
-  beforeEach(() => {
-    originalOnErrorHandler = window.onerror
-    onErrorSpy = jasmine.createSpy()
-    window.onerror = onErrorSpy
-
-    onUnhandledrejectionSpy = setupOnUnhandledrejectionSpy()
+  const withTrackRuntimeError = async (callback: () => void): Promise<RawError> => {
```
sethfowler-datadog, line 10: The tests in this file involve some async steps, and getting the ordering right is tricky. I've reworked them with an `async` callback-based approach that (IMO) feels a bit cleaner and makes it easier to tell that things happen in the right order. I know this is a bit of a different pattern than most of the other tests, though!

packages/core/test/instrumentation.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @returns true if target[method] is instrumented by action. (e.g. via
+ * instrumentMethod())
+ */
+export function instrumentationIsAddedTo<TARGET extends { [key: string]: any }, METHOD extends keyof TARGET & string>(
```
sethfowler-datadog, line 5: This helper is super trivial with the existing version of `instrumentMethod()`, but it's nice to have it, because if we rework `instrumentMethod()` to use a longer-lived observer-like pattern, the equivalent condition becomes a little trickier to write. The updated version of #3214 will replace the version you see here with a new one that handles the new `instrumentMethod()` implementation added in that PR.

packages/core/test/wait.ts
```diff
@@ -0,0 +1,5 @@
+export function wait(durationMs: number = 0): Promise<void> {
```
sethfowler-datadog, line 1: This is just generally a nice helper to have for async code; I'm using it in the updated async versions of the tests in `trackRuntimeError.spec.ts`.

packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
```diff
@@ -37,10 +36,8 @@ describe('network error collection', () => {
 
   function startCollection(forwardErrorsToLogs = true) {
     mockFetchManager = mockFetch()
-    ;({ stop: stopNetworkErrorCollection } = startNetworkErrorCollection(
-      { ...CONFIGURATION, forwardErrorsToLogs },
-      lifeCycle
-    ))
+    const { stop } = startNetworkErrorCollection({ ...CONFIGURATION, forwardErrorsToLogs }, lifeCycle)
+    registerCleanupTask(stop)
```
sethfowler-datadog, line 40: This code overwrites `window.fetch` with a mock version, and then adds our fetch instrumentation to it. The change here ensures that when we tear things down, we remove the instrumentation before overwriting `window.fetch` with its original value, so that cleanup happens correctly.

packages/core/src/domain/error/trackRuntimeError.spec.ts
sethfowler-datadog: Now that I've rebased against v6, the delta is a bit smaller here, because these tests already used async/await in the v6 branch. The difference is now just that the flow of events in the tests is managed more explicitly, without the use of `collectAsyncCalls`.

packages/logs/src/boot/preStartLogs.spec.ts
```diff
@@ -217,14 +217,14 @@ describe('preStartLogs', () => {
...
-        const originalFetch = window.fetch
-
-        strategy.init({
-          ...DEFAULT_INIT_CONFIGURATION,
-          trackingConsent: TrackingConsent.NOT_GRANTED,
-        })
-
-        expect(window.fetch).not.toBe(originalFetch)
+        expect(
+          instrumentationIsAddedTo(window, 'fetch', () => {
```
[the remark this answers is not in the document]
sethfowler-datadog: I agree with the problem! I felt like `didAddInstrumentationTo()` could still be read in the unintended way, though, so I went a bit further and switched to a fluent interface. The new API follows this pattern:
```ts
callbackAddsInstrumentation(() => {
  // ... code that's intended to instrument window.foo ...
}).toMethod(window, 'foo').whenCalled()
```
More verbose, but hopefully this will ultimately be clearer.

packages/core/src/domain/error/trackRuntimeError.spec.ts
```diff
@@ -260,55 +274,68 @@ describe('instrumentOnError', () => {
...
-    window.onunhandledrejection = onUnhandledrejectionSpy
-
-    registerCleanupTask(() => {
-      window.onunhandledrejection = originalOnUnhandledRejectionHandler
+/**
+ * Invokes the given callback, which is expected to be a function that throws or generates
+ * an unhandled promise rejection, and returns a promise that resolves after the callback
+ * has finished running and any global error handlers that run as a result are complete.
+ */
+function invokeAndWaitForErrorHandlers(callback: () => void): Promise<void> {
```
sethfowler-datadog, line 325: I had a bug in my original implementation of this code; it lacked the nested `setTimeout`. That caused test failures on Safari in particular. Originally the buggy code was repeated in a few places in this file; I've pulled it into this helper function, fixed the issue, and added some comments to clarify why the function works in the way that it does.

packages/core/src/domain/error/trackRuntimeError.spec.ts
```diff
@@ -61,196 +54,217 @@ describe('instrumentOnError', () => {
...
-    window.onerror = onErrorSpy
-    callbackSpy = jasmine.createSpy()
-    ;({ stop: stopCollectingUnhandledError } = instrumentOnError(callbackSpy))
-  })
-
-  afterEach(() => {
-    window.onerror = originalOnErrorHandler
-    stopCollectingUnhandledError()
-  })
+  const withInstrumentOnError = async (callback: () => void): Promise<jasmine.Spy> => {
```
amortemousque, line 57: **💬 suggestion:** ‏The name is a bit unclear. To better capture the intention what about: `spyAndInstrumentOnError`?

amortemousque (approved):

packages/core/src/domain/error/trackRuntimeError.spec.ts
sethfowler-datadog: Sure, SGTM! I'll make that change.

sethfowler-datadog: I renamed all of the test helpers. After trying to find a naming pattern that worked well for all of them I ended up going with `errorViaTrackRuntimeError`, `spyViaInstrumentOnError`, and `spyViaInstrumentOnUnhandledRejection`.

PR #3242, opened 2024-12-24
Add setAccount API

nulrich:
## Motivation

Adding a standard way to set an account.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/domain/account/account.ts
+++ b/packages/core/src/domain/account/account.ts
@@ -1,0 +1,6 @@
+import type { Context } from '../../tools/serialisation/context'
+import { sanitizeContext } from '../context/contextUtils'
+
+export function sanitizeAccount(newAccount: Context) {
+  return sanitizeContext(newAccount, ['id', 'name'])
+}
--- a/packages/core/src/domain/account/account.types.ts
+++ b/packages/core/src/domain/account/account.types.ts
@@ -1,0 +1,5 @@
+export interface Account {
+  id?: string | undefined
+  name?: string | undefined
+  [key: string]: unknown
+}
--- a/packages/core/src/domain/context/contextUtils.ts
+++ b/packages/core/src/domain/context/contextUtils.ts
@@ -1,0 +1,31 @@
+import type { Context } from '../../tools/serialisation/context'
+import { display } from '../../tools/display'
+import { getType } from '../../tools/utils/typeUtils'
+import { assign } from '../../tools/utils/polyfills'
+
+/**
+ * Clone input data and ensure known context properties
+ * are strings, as defined here:
+ * https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#user-related-attributes
+ */
+export function sanitizeContext(newContext: Context, keys: string[]): Context {
+  // We shallow clone only to prevent mutation of context data.
+  const context = assign({}, newContext)
+  keys.forEach((key) => {
+    if (key in context) {
+      context[key] = String(context[key])
+    }
+  })
+  return context
+}
+
+/**
+ * Simple check to ensure an object is a valid context
+ */
+export function checkContext(newContext: Context): boolean {
+  const isValid = getType(newContext) === 'object'
+  if (!isValid) {
+    display.error('Unsupported context:', newContext)
+  }
+  return isValid
+}
--- a/packages/logs/src/boot/logsPublicApi.spec.ts
+++ b/packages/logs/src/boot/logsPublicApi.spec.ts
@@ -84,6 +84,7 @@ describe('logs entry', () => {
         },
         context: { foo: 'bar' },
         user: {},
+        account: {},
       })
     })
   })
--- a/packages/logs/src/boot/logsPublicApi.ts
+++ b/packages/logs/src/boot/logsPublicApi.ts
@@ -1,4 +1,4 @@
-import type { Context, TrackingConsent, User, PublicApi } from '@datadog/browser-core'
+import type { Account, Context, TrackingConsent, User, PublicApi } from '@datadog/browser-core'
 import {
   addTelemetryUsage,
   CustomerDataType,
@@ -6,7 +6,6 @@ import {
   createContextManager,
   makePublicApi,
   monitor,
-  checkUser,
   sanitizeUser,
   sanitize,
   createCustomerDataTrackerManager,
@@ -14,6 +13,8 @@ import {
   displayAlreadyInitializedError,
   deepClone,
   createTrackingConsentState,
+  checkContext,
+  sanitizeAccount,
 } from '@datadog/browser-core'
 import type { LogsInitConfiguration } from '../domain/configuration'
 import type { HandlerType } from '../domain/logger'
@@ -158,6 +159,39 @@ export interface LogsPublicApi extends PublicApi {
    * See [User context](https://docs.datadoghq.com/logs/log_collection/javascript/#user-context) for further information.
    */
   clearUser: () => void
+
+  /**
+   * Set account information to all events, stored in `@account`
+   *
+   */
+  setAccount: (newAccount: Account & { id: string }) => void
+
+  /**
+   * Get account information
+   *
+   */
+  getAccount: () => Context
+
+  /**
+   * Set or update the account property, stored in `@account.<key>`
+   *
+   * @param key Key of the property
+   * @param property Value of the property
+   *
+   */
+  setAccountProperty: (key: any, property: any) => void
+
+  /**
+   * Remove a account property
+   *
+   */
+  removeAccountProperty: (key: any) => void
+
+  /**
+   * Clear all account information
+   *
+   */
+  clearAccount: () => void
 }
 
 const LOGS_STORAGE_KEY = 'logs'
@@ -175,16 +209,20 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs): LogsPublicApi {
     customerDataTrackerManager.getOrCreateTracker(CustomerDataType.GlobalContext)
   )
   const userContextManager = createContextManager(customerDataTrackerManager.getOrCreateTracker(CustomerDataType.User))
+  const accountContextManager = createContextManager(
+    customerDataTrackerManager.getOrCreateTracker(CustomerDataType.Account)
+  )
   const trackingConsentState = createTrackingConsentState()
 
   function getCommonContext() {
-    return buildCommonContext(globalContextManager, userContextManager)
+    return buildCommonContext(globalContextManager, userContextManager, accountContextManager)
   }
 
   let strategy = createPreStartStrategy(getCommonContext, trackingConsentState, (initConfiguration, configuration) => {
     if (initConfiguration.storeContextsAcrossPages) {
       storeContextManager(configuration, globalContextManager, LOGS_STORAGE_KEY, CustomerDataType.GlobalContext)
       storeContextManager(configuration, userContextManager, LOGS_STORAGE_KEY, CustomerDataType.User)
+      storeContextManager(configuration, accountContextManager, LOGS_STORAGE_KEY, CustomerDataType.Account)
     }
 
     const startLogsResult = startLogsImpl(initConfiguration, configuration, getCommonContext, trackingConsentState)
@@ -240,7 +278,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs): LogsPublicApi {
     getInternalContext: monitor((startTime) => strategy.getInternalContext(startTime)),
 
     setUser: monitor((newUser) => {
-      if (checkUser(newUser)) {
+      if (checkContext(newUser as Context)) {
         userContextManager.setContext(sanitizeUser(newUser as Context))
       }
     }),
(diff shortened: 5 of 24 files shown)


packages/core/src/domain/context/contextUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+      context[key] = String(context[key])
+    }
+  })
+  return context
+}
+
+/**
+ * Simple check to ensure an object is a valid context
+ */
+export function checkContext(newContext: Context): boolean {
```
sethfowler-datadog, line 26: Would it make sense for the signature of this function to be this instead?
```suggestion
export function checkContext(maybeContext: unknown): maybeContext is Context {
```

That way you wouldn't need the `as Context` type assertions you're currently applying at call sites; instead, you could use this function to narrow the type of a value to `Context` when it can be treated as one.

packages/core/src/domain/account/account.ts
```diff
@@ -0,0 +1,6 @@
+import type { Context } from '../../tools/serialisation/context'
+import { sanitizeContext } from '../context/contextUtils'
+
+export function sanitizeAccount(newAccount: Context) {
```
sethfowler-datadog, line 4: The return value of this function is not just a `Context`, but an `Account`, right? It might make sense to update the function signature to capture that.

packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -318,6 +319,144 @@ describe('logs entry', () => {
...
+      })
+
+      it('should not contain original reference to object', () => {
+        const accountDetails: { [key: string]: any } = { name: 'john' }
+        logsPublicApi.setAccount(account)
+        logsPublicApi.setAccountProperty('accountDetails', accountDetails)
+        accountDetails.DOB = '11/11/1999'
+        const accountClone = logsPublicApi.getAccount()
+
+        expect(accountClone.accountDetails).not.toBe(accountDetails)
```
sethfowler-datadog, line 422: Nit: did you mean to assert something about the `DOB` property here too? I ask only because right now it doesn't really have an impact on the test's result.

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -158,6 +159,39 @@ export interface LogsPublicApi extends PublicApi {
    * See [User context](https://docs.datadoghq.com/logs/log_collection/javascript/#user-context) for further information.
    */
   clearUser: () => void
+
+  /**
+   * Set account information to all events, stored in `@account`
+   *
```
sethfowler-datadog, line 165: Nit: There's an unnecessary empty line at the end of this comment. (All of the new JSDoc comments in this file have it, actually; might be a copy paste mistake.)

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -158,6 +159,39 @@ export interface LogsPublicApi extends PublicApi {
...
+   * Set or update the account property, stored in `@account.<key>`
+   *
+   * @param key Key of the property
+   * @param property Value of the property
+   *
+   */
+  setAccountProperty: (key: any, property: any) => void
+
+  /**
+   * Remove a account property
```
sethfowler-datadog, line 185: Nit:
```suggestion
   * Remove an account property
```

packages/logs/src/domain/assembly.ts
```diff
@@ -36,13 +43,25 @@ export function startLogsAssembly(
...
 
       const commonContext = savedCommonContext || getCommonContext()
+
+      let account
+
+      if (!isEmptyObject(commonContext.account)) {
+        if (commonContext.account.id) {
+          account = commonContext.account
+        } else {
+          display.warn("The account object is missing the 'id' property, it will not be sent to the intake.")
```
sethfowler-datadog, line 53: Nit:
```suggestion
          display.warn("The account object is missing the 'id' property; it will not be sent to the intake.")
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -217,6 +219,39 @@ export interface RumPublicApi extends PublicApi {
    */
   clearUser: () => void
 
+  /**
+   * Set account information to all events, stored in `@account`
+   *
```
sethfowler-datadog, line 224: Nit: these JSDoc comments end with an extra blank line, like in the other copy of this code.

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -217,6 +219,39 @@ export interface RumPublicApi extends PublicApi {
...
+   * Set or update the account property, stored in `@account.<key>`
+   *
+   * @param key Key of the property
+   * @param property Value of the property
+   *
+   */
+  setAccountProperty: (key: any, property: any) => void
+
+  /**
+   * Remove a account property
```
sethfowler-datadog, line 244: Nit:
```suggestion
   * Remove an account property
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -525,6 +564,23 @@ export function makeRumPublicApi(
...
+    setAccount: monitor((newAccount) => {
+      if (checkContext(newAccount as Context)) {
+        accountContextManager.setContext(sanitizeUser(newAccount as Context))
+      }
+    }),
+
+    getAccount: monitor(() => accountContextManager.getContext()),
+
+    setAccountProperty: monitor((key, property) => {
+      const sanitizedProperty = sanitizeAccount({ [key]: property })[key]
```
sethfowler-datadog, line 576: It's unfortunate that this logic has to be repeated twice. It'd be nice if the context manager itself could be configured with a sanitizer, and it'd just handle the work internally. Not a blocker for this PR, but might be nice to consider as a followup.

packages/rum-core/src/domain/assembly.ts
```diff
@@ -203,6 +203,15 @@ export function startRumAssembly(
           ;(serverRumEvent.usr as Mutable<RumEvent['usr']>) = commonContext.user as User & Context
         }
 
+        if (!isEmptyObject(commonContext.account)) {
+          if (commonContext.account.id) {
+            ;(serverRumEvent.account as Mutable<RumEvent['account']>) = commonContext.account as Account &
+              Context & { id: string }
```
sethfowler-datadog, line 209: I'm guessing the `as Account` type assertion is losing the type narrowing you just did with the `if (commonContext.account.id)` check, and that's why you need `& { id: string }`. Out of curiosity, does the assignment type check if you just get rid of the whole type assertion?

sethfowler-datadog (approved):
Some minor nits, but everything looks great!

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -217,6 +219,39 @@ export interface RumPublicApi extends PublicApi {
...
+  getAccount: () => Context
+
+  /**
+   * Set or update the account property, stored in `@account.<key>`
+   *
+   * @param key Key of the property
+   * @param property Value of the property
+   *
+   */
+  setAccountProperty: (key: any, property: any) => void
```
amortemousque, line 241: **❓ question:**  `key: string` be more accurate? I know we also have `any` for user API but might worth reconsider.‏

packages/logs/src/boot/logsPublicApi.spec.ts
nulrich: I removed it 👍 We only want to assert references are different.

packages/core/src/domain/account/account.types.ts
```diff
@@ -0,0 +1,5 @@
+export interface Account {
+  id?: string | undefined
```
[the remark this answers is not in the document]
nulrich: In theory yes 😅 But the issue is that Context Manager is manipulating `Context` so we loose the original type when passing the `User` or `Account` object. Until now it was fine because both `Context` and `User` have no required properties.

I could require the `id` property but we would have to cast the context, and maybe getting an account without an `id`.

```typescript
export function buildCommonContext(
  globalContextManager: ContextManager,
  userContextManager: ContextManager,
  acccountContextManager: ContextManager
): CommonContext {
  return {
    view: {
      referrer: document.referrer,
      url: window.location.href,
    },
    context: globalContextManager.getContext(),
    user: userContextManager.getContext(),
    account: acccountContextManager.getContext() as Account,
  }
}
```

That's why I preferred not requiring the Id and checking it in the assembly. Hope it makes sense.

packages/rum-core/src/domain/assembly.ts
nulrich: The type of `commonContext.account` is `Account` so even with the narrowing it is still "just" an `Account` and we must enforce `{ id: string }`.

But because there's overlap between `Account` and `Context`, I could do:

```typescript
if (commonContext.account.id) {
  (serverRumEvent.account as Mutable<RumEvent['account']>) = commonContext.account as Context & { id: string }
} else {
```

packages/core/src/domain/account/account.ts
nulrich: `sanitizeAccount` and `sanitizeUser` are manipulating `Context` and are used by Context Manager functions that expect a `Context`:

```typescript
accountContextManager.setContext(sanitizeAccount(newUser))
```

If I update the signature I will have to cast the result. But maybe I could rename to `sanitizeAccountContext` ?

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -158,6 +159,34 @@ export interface LogsPublicApi extends PublicApi {
    * See [User context](https://docs.datadoghq.com/logs/log_collection/javascript/#user-context) for further information.
    */
   clearUser: () => void
+
+  /**
+   * Set account information to all events, stored in `@account`
+   */
+  setAccount: (newAccount: Account & { id: string }) => void
```
BenoitZugmeyer, line 166: **💬 suggestion:** ‏maybe using a proper type here would be nicer for the user

```suggestion
  setAccount: (newAccount: Account) => void
```

I understand that you don't want `account.id` to be required in the type we use internally, since we cannot enforce it. I think it's fine to use `Context` instead of `Account` internally (see this [patch](https://gist.github.com/BenoitZugmeyer/bc3c32bc0aa81f46bb39c93fdbc68787)).

nulrich: 💯 and we enforce the type just after in the assembly anyway. We should though how the context management could handle sanitization like @sethfowler-datadog proposed, but also typing.

packages/core/src/domain/account/account.types.ts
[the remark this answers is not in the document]
nulrich: Following @BenoitZugmeyer comment, we now use `Context` type in the `CommonContext` for  account property. So now the `Account` type have a required id property.

packages/logs/src/boot/logsPublicApi.ts
nulrich: Done. Context Manager now check required property.

packages/rum-core/src/boot/rumPublicApi.ts
nulrich: Done in https://github.com/DataDog/browser-sdk/pull/3290

packages/core/src/domain/account/account.ts
nulrich: Sanitization is now done Context Manager

packages/rum-core/src/domain/assembly.ts
nulrich: `CommonContext` now use `Context` instead of specific types (User, Account...). We now just need a cast to the needed type.

packages/rum-core/src/domain/assembly.ts
```diff
@@ -203,6 +203,15 @@ export function startRumAssembly(
...
+          } else {
+            display.warn("The account object is missing the 'id' property, it will not be sent to the intake.")
```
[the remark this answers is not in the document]
nulrich: Warn is now in Context Manager.

packages/core/src/domain/context/contextUtils.ts
```diff
@@ -0,0 +1,31 @@
+import type { Context } from '../../tools/serialisation/context'
+import { display } from '../../tools/display'
+import { getType } from '../../tools/utils/typeUtils'
+
+/**
+ * Clone input data and ensure known context properties
+ * are strings, as defined here:
+ * https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#user-related-attributes
+ */
+export function sanitizeContext(newContext: Context, keys: string[]): Context {
```
BenoitZugmeyer, line 10: **❓ question:** ‏this function seems unused?

packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -259,6 +295,22 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs): LogsPublicApi {
...
+
+    getAccount: monitor(() => accountContextManager.getContext()),
+
+    setAccountProperty: monitor((key, property) => {
+      accountContextManager.setContextProperty(key, property)
+    }),
+
+    removeAccountProperty: monitor((key) => accountContextManager.removeContextProperty(key)),
+
+    clearAccount: monitor(() => accountContextManager.clearContext()),
```
BenoitZugmeyer, line 313: **💬 suggestion:** ‏I wonder if we could:

* do `checkContext` in `ContextManager.setContext`
* and then just:

```suggestion
    setAccount: monitor(accountContextManager.setContext),
    getAccount: monitor(accountContextManager.getContext),
    setAccountProperty: monitor(accountContextManager.setContextProperty),
    removeAccountProperty: monitor(accountContextManager.removeContextProperty),
    clearAccount: monitor(accountContextManager.clearContext),
```

packages/core/src/domain/context/contextUtils.ts
nulrich: Nice catch!

packages/logs/src/boot/logsPublicApi.ts
nulrich: I moved `checkContext` in `ContextManager.setContext` but I had to change the type of newContext to `any`.

packages/core/src/domain/context/contextManager.ts
```diff
@@ -52,8 +52,8 @@ export function createContextManager(
   const contextManager = {
     getContext: () => deepClone(context),
 
-    setContext: (newContext: Context) => {
-      if (getType(newContext) === 'object') {
+    setContext: (newContext: any) => {
```
BenoitZugmeyer, line 55: **💬 suggestion:** ‏use `unknown` instead of `any`

BenoitZugmeyer (approved):

amortemousque (approved):

ramigluzman:
@nulrich Could you clarify the proper usage of this new method and when it will be added to your documentation?
Thank you for your help!

PR #3245, opened 2025-01-03
✅ fix leak detection issues

BenoitZugmeyer:
## Motivation

In  the v6 branch, we started seeing a lot of leak detection warnings. We identified that the main issue was brought by https://github.com/DataDog/browser-sdk/pull/3190, switching the default `trackUserInteraction` to `true`: since we are now tracking user interaction every time `startActionCollection` is called, we need to clean it up as well.

Another issue was identified in the leak detection logic: the leak detection cleanup was done too soon, and the `removeEventListener` function was unpatched before some events were removed. [Because leak detection is wrapping the listener function into another "`wrappedListener`" function](https://github.com/DataDog/browser-sdk/blob/8acb3d08836977188a260060709cd7f1609c04f5/packages/core/test/leakDetection.ts#L24), the original function failed to be removed.

## Changes

* properly cleanup action collection in tests
* make sure forEach.spec.ts is run early, so its `afterEach` (and in turn, the leak detection cleanup) is run after every other `afterEach`.
* (bonus) remove `splitChunk: false` again to speed tests again (see https://github.com/DataDog/browser-sdk/pull/3057 an accidental revert https://github.com/DataDog/browser-sdk/commit/56694a12c293e04ced5027aed7240d5f8a9a615d )

## Testing

- [ ] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/test/unit/karma.base.conf.js
+++ b/test/unit/karma.base.conf.js
@@ -18,6 +18,10 @@ if (testReportDirectory) {
 module.exports = {
   basePath: '../..',
   files: [
+    // Make sure 'forEach.spec' is the first file to be loaded, so its `beforeEach` hook is executed
+    // before all other `beforeEach` hooks, and its `afterEach` hook is executed after all other
+    // `afterEach` hooks.
+    'packages/core/test/forEach.spec.ts',
     'packages/*/@(src|test)/**/*.spec.@(ts|tsx)',
     'developer-extension/@(src|test)/**/*.spec.@(ts|tsx)',
     'packages/rum/test/toto.css',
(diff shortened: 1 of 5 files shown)


test/unit/karma.base.conf.js
```diff
@@ -18,6 +18,10 @@ if (testReportDirectory) {
 module.exports = {
   basePath: '../..',
   files: [
+    // Make sure 'forEach.spec' is the first file to be loaded, so its `beforeEach` hook is executed
+    // before all other `beforeEach` hooks, and its `afterEach` hook is executed after all other
+    // `afterEach` hooks.
+    'packages/core/test/forEach.spec.ts',
```
cy-moi, line 24: 😲 👏

cy-moi (approved):

PR #3262, opened 2025-01-08
✨ [RUM-7958] Deprecate setUser without required Id

nulrich:
## Motivation

We would like to enforce usage of the 'id' property on user context.

## Changes

Deprecate `setUser(newUser: User): void` and add `setUser(newUser: User & { id: string }): void`

## Testing

- [X] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


amortemousque:
**❓ question:** ‏Shouldn't the id be marked as "required" [here](https://github.com/DataDog/browser-sdk/blob/3499db1ff24c6d32656eb2f996e5dd838226c911/packages/rum-core/src/boot/rumPublicApi.ts#L411)?

amortemousque:
Also something I missed in the other [PR](https://github.com/DataDog/browser-sdk/pull/3290).

- The account context manager is wrongly named: [here](https://github.com/DataDog/browser-sdk/blob/3499db1ff24c6d32656eb2f996e5dd838226c911/packages/rum-core/src/boot/rumPublicApi.ts#L417)

- I think we should tweak a bit the console message not to include the term "context" for account and user:
**Datadog Browser SDK: The property id of user <del>context</del> is required; context will not be sent to the intake.**

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -192,7 +192,9 @@ export interface RumPublicApi extends PublicApi {
   setUser(newUser: User & { id: string }): void
 
   /**
-   * @deprecated Use `setUser` with a user object that requires an `id`.
+   * Set user information to all events, stored in `@usr`
+   *
+   * @deprecated You must specified a user id
```
thomas-lebeau, line 197:
```suggestion
   * @deprecated You must specify a user id
```

thomas-lebeau (approved):

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -412,10 +418,10 @@ export function makeRumPublicApi(
       email: { type: 'string' },
     },
   })
-  const accountContextManager = createContextManager('user', {
+  const accountContextManager = createContextManager('account', {
     customerDataTracker: customerDataTrackerManager.getOrCreateTracker(CustomerDataType.User),
     propertiesConfig: {
-      id: { type: 'string' },
+      id: { type: 'string', required: true },
```
amortemousque, line 424: **💬 suggestion:** ‏You probably need to do the same for `createContextManager('user', {`

nulrich: For `user` we only deprecate the API, we still allow sending user without id.

amortemousque: Oh right!

amortemousque (approved):

PR #3266, opened 2025-01-08
✨ [RUM-7572] Add get api of view specific context

RomanGaignault:
## Motivation

After adding set view context we need to have a get view context api

## Changes

Add a new api

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/boot/preStartRum.ts
+++ b/packages/rum-core/src/boot/preStartRum.ts
@@ -204,7 +204,7 @@ export function createPreStartStrategy(
       bufferApiCalls.add((startRumResult) => startRumResult.setViewContextProperty(key, value))
     },
 
-    getViewContext: noop as () => undefined,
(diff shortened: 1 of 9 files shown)


packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -53,6 +53,8 @@ export function createPreStartStrategy(
 
   const trackingConsentStateSubscription = trackingConsentState.observable.subscribe(tryStartRum)
 
+  const noopContext: Context = {}
```
thomas-lebeau, line 56: **🥜 nitpick:** ‏`emptyContext` is more correct I think in this case (even if indeed there is no operation too)

thomas-lebeau (approved):

nulrich (approved):

PR #3268, opened 2025-01-08
👷‍♀️ [RUM-7963] Add anonymous user id e2e test and cleanup

cy-moi:
## Motivation
Add e2e tests for anonymous user id and clean up existing tests to distinguish anonymous id from session id in cookies.

## Changes
Added new e2e tests
Changed current tests matching regex

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


test/e2e/scenario/rum/sessions.scenario.ts
```diff
@@ -50,6 +50,58 @@ describe('rum sessions', () => {
...
+        const prevSessionCookie = await findSessionCookie()
+        const anonymousId = prevSessionCookie?.match(/aid=[a-z0-9-]+/)
+        expect(anonymousId).not.toBeNull()
+
+        await expireSession()
+        await flushEvents()
+
+        if (anonymousId) {
+          expect(await findSessionCookie()).toContain(anonymousId[0])
+        }
```
BenoitZugmeyer, line 66: **💬 suggestion:** ‏FMU you have the `if (anonymousId)` to make TS happy. I don't think we should have this condition:
* even if `anonymousId` should always be truthy (since we have the assertion above), it makes the test a bit confusing
* you actually want to make the test fail if it was falsy, and not just silently skip the assertion

I think using `!` is fine here

```suggestion
        expect(await findSessionCookie()).toContain(anonymousId![0])
```

BenoitZugmeyer (approved):

PR #3272, opened 2025-01-09
✨ Collect long animation frames as long task events

thomas-lebeau:
## Motivation

GA Collecting long animation frames

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


test/e2e/lib/helpers/browser.ts
```diff
@@ -69,7 +69,13 @@ export async function withBrowserLogs(fn: (logs: BrowserLog[]) => void) {
   // https://github.com/webdriverio/webdriverio/issues/4275
   if (browser.getLogs) {
     const logs = (await browser.getLogs('browser')) as BrowserLog[]
-    fn(logs)
+
+    const filterdLogs = logs.filter(
+      // Ignore long-animation-frame warning (only happens on Edge)
+      (log) => !log.message.includes("The entry type 'long-animation-frame' does not exist or isn't supported")
+    )
```
thomas-lebeau, line 76: A bunch of e2e are verifying the number browser logs ([example](https://github.com/DataDog/browser-sdk/blob/0eb2a30fb8775a5c9d34d0f637dfd4d457335228/test/e2e/scenario/rum/errors.scenario.ts#L38-L41))

Another solution would be to makes these test for the console log they are actually expecting. But at the same time it is nice to ensure there is no unexpected console logs.

BenoitZugmeyer: **💬 suggestion:** ‏
We should probably adjust `performanceObservable` to skip observing unsupported entries to avoid spamming the console on customer websites. Something like:

```ts
if (!supportPerformanceTimingEvent(options.type)) {
  return
}
````

thomas-lebeau: `Performance.supportedEntryTypes` is not supported everywhere. For example,
Long Task are supported since chrome 58, but `Performance.supportedEntryTypes` only since Chrome 73, while we supposedly support chrome 63+

thomas-lebeau: yes, as finally we're falling back to Long tasks when LoAF is not supported, I'm starting either Loaf collection or Long task collection in `startRum` depending on `Performance.supportedEntryTypes`.

It should work to get rid of the warning because everywhere LoAF are supported `Performance.supportedEntryTypes` is also supported.

BenoitZugmeyer (approved):

PR #3273, opened 2025-01-09
👷 do not rename bundle file when name does not change

thomas-lebeau:
## Motivation

We don't suffix chunks, so we should not try to rename them as this would fail:

```
 Error: Command failed with exit status 1: mv packages/rum/bundle/chunks/recorder-d7628536637b074ddc3b-datadog-rum.js packages/rum/bundle/chunks/recorder-d7628536637b074ddc3b-datadog-rum.js
---- stderr: ----
mv: 'packages/rum/bundle/chunks/recorder-d7628536637b074ddc3b-datadog-rum.js' and 'packages/rum/bundle/chunks/recorder-d7628536637b074ddc3b-datadog-rum.js' are the same file
----
```

We didn't ran into the issue before because `deploy-next-major-canary` don't upload sourcemaps (I guess we should fix that for next time)

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/scripts/deploy/upload-source-maps.js
+++ b/scripts/deploy/upload-source-maps.js
@@ -68,6 +68,10 @@ async function renameFilesWithVersionSuffix(bundleFolder, version) {
   await forEachFile(bundleFolder, (bundlePath) => {
     const uploadPath = buildRootUploadPath(bundlePath, version)
 
+    if (bundlePath === uploadPath) {
+      return
+    }
+
     console.log(`Renaming ${bundlePath} to ${uploadPath}`)
     command`mv ${bundlePath} ${uploadPath}`.run()
   })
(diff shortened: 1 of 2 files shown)


scripts/deploy/upload-source-maps.js
```diff
@@ -68,6 +68,10 @@ async function renameFilesWithVersionSuffix(bundleFolder, version) {
   await forEachFile(bundleFolder, (bundlePath) => {
     const uploadPath = buildRootUploadPath(bundlePath, version)
 
+    if (bundlePath === uploadPath) {
+      return
+    }
+
     console.log(`Renaming ${bundlePath} to ${uploadPath}`)
     command`mv ${bundlePath} ${uploadPath}`.run()
```
BenoitZugmeyer, line 76: **Nitpick** Using `command` to move files is a bit overkill, we could just use [`fs.renameSync`](https://nodejs.org/docs/v22.13.0/api/fs.html#fsrenamesyncoldpath-newpath)

**Nitpick** Also, we could have used `glob` (either from the `glob` package or [the Node.js stdlib](https://nodejs.org/docs/v22.13.0/api/fs.html#fsglobsyncpattern-options)) instead of `forEachFile`.

amortemousque (approved):

PR #3276, opened 2025-01-11
⚗ [RUM-7213] DOM mutation ignoring

jordan-th:
## Motivation

Ability to exclude certain DOM mutations from affecting the page loading time calculation

## Changes

- `attributes` mutations are ignored if the element has the attribute `dd-ignore-mutations`
- `childList` and `characterData` mutations are ignored if the parent element has the attribute `dd-ignore-mutations`

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/browser/domMutationObservable.spec.ts
+++ b/packages/rum-core/src/browser/domMutationObservable.spec.ts
@@ -1,7 +1,7 @@
-import { isIE } from '@datadog/browser-core'
+import { isIE, ExperimentalFeature } from '@datadog/browser-core'
 import type { MockZoneJs } from '@datadog/browser-core/test'
-import { registerCleanupTask, mockZoneJs } from '@datadog/browser-core/test'
-import { createDOMMutationObservable, getMutationObserverConstructor } from './domMutationObservable'
+import { registerCleanupTask, mockZoneJs, mockExperimentalFeatures } from '@datadog/browser-core/test'
+import { createDOMMutationObservable, getMutationObserverConstructor, IGNORE_MUTATIONS_ELEMENT_IDENTIFIER} from './domMutationObservable'
 
 // The MutationObserver invokes its callback in an event loop microtask, making this asynchronous.
 // We want to wait for a few event loop executions to potentially collect multiple mutation events.
@@ -14,10 +14,13 @@ describe('domMutationObservable', () => {
     }
   })
 
-  function domMutationSpec(mutate: (root: HTMLElement) => void, { expectedMutations }: { expectedMutations: number }) {
+  function domMutationSpec(mutate: (root: HTMLElement) => void, { expectedMutations }: { expectedMutations: number }, isRootIgnored?: boolean) {
     return (done: DoneFn) => {
       const root = document.createElement('div')
       root.setAttribute('data-test', 'foo')
+      if (isRootIgnored) {
+        root.setAttribute(IGNORE_MUTATIONS_ELEMENT_IDENTIFIER, '')
+      }
       root.appendChild(document.createElement('span'))
       root.appendChild(document.createTextNode('foo'))
       document.body.appendChild(root)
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/browser/domMutationObservable.spec.ts
```diff
@@ -1,16 +1,28 @@
...
 // The MutationObserver invokes its callback in an event loop microtask, making this asynchronous.
 // We want to wait for a few event loop executions to potentially collect multiple mutation events.
 const DOM_MUTATION_OBSERVABLE_DURATION = 16
 
 describe('domMutationObservable', () => {
-  function domMutationSpec(mutate: (root: HTMLElement) => void, { expectedMutations }: { expectedMutations: number }) {
+  function domMutationSpec(
+    mutate: (root: HTMLElement) => void,
+    { expectedMutations }: { expectedMutations: number },
+    isRootMutationsIgnored?: boolean
```
jordan-th, line 18: This parameter allows us to test the case where the experimental feature _is enabled_ and the `IGNORE_MUTATIONS_ATTRIBUTE` is _not present_.

If you think it would be useful, I am also open to adding tests for the case where the experimental feature is _not enabled_ and the `IGNORE_MUTATIONS_ATTRIBUTE` _is present_

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -7,7 +18,18 @@ export function createDOMMutationObservable() {
     if (!MutationObserver) {
       return
     }
-    const observer = new MutationObserver(monitor(() => observable.notify()))
+
+    let mutationNotifier: MutationNotifier = () => observable.notify()
+    if (isExperimentalFeatureEnabled(ExperimentalFeature.DOM_MUTATION_IGNORING)) {
+      mutationNotifier = (mutations: MutationRecord[]) => {
```
BenoitZugmeyer, line 24: **suggestion:** use `RumMutationRecord` instead of `MutationRecord`. The type is a bit more fine grained than the default TS type.

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -7,7 +18,18 @@ export function createDOMMutationObservable() {
...
+        if (mutations.every((mutation) => isIgnored(mutation))) {
```
BenoitZugmeyer, line 25: **suggestion:** 
```suggestion
        if (mutations.every(isIgnored)) {
```

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -18,6 +40,16 @@ export function createDOMMutationObservable() {
   })
 }
 
+function isIgnored(mutation: MutationRecord): boolean {
+  switch (mutation.type) {
+    case 'attributes':
+    case 'childList':
+      return (mutation.target as Element).hasAttribute(IGNORE_MUTATIONS_ATTRIBUTE) === true
```
BenoitZugmeyer, line 47: **suggestion:** this cast is a bit problematic because `mutation.target` can technically be a `Document` in the `childList` case. We should make sure `mutation.target.nodeType === 1` (use [`isElementNode`](https://github.com/DataDog/browser-sdk/blob/c2f6a224f9652b6de6d84021a53bc1d3a37cc365/packages/rum-core/src/browser/htmlDomUtils.ts#L9))

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -1,4 +1,15 @@
...
+import {
+  monitor,
+  noop,
+  Observable,
+  getZoneJsOriginalValue,
+  ExperimentalFeature,
+  isExperimentalFeatureEnabled,
+} from '@datadog/browser-core'
+
+export const IGNORE_MUTATIONS_ATTRIBUTE = 'dd-ignore-mutations'
```
BenoitZugmeyer, line 10: **suggestion:** 
* the attribute should start with `data-` similarly to the existing ones `data-dd-privacy` and `data-dd-action-name`
* I wonder if we can have something more specific, like `data-dd-ignore-activity-mutations`. We can take inspiration from the existing `expludedActivityUrls` option.

packages/rum-core/src/browser/domMutationObservable.ts
BenoitZugmeyer, line 47: **thought:** I think we should consider any parent that contains the attribute. This is similar to `data-dd-action-name` and `data-dd-privacy`. It could be as trivial as:

```js
mutation.target.closest(`[${IGNORE_MUTATIONS_ATTRIBUTE}]`) !== null
```

amortemousque (approved):

test/e2e/scenario/rum/views.scenario.ts
```diff
@@ -15,6 +15,41 @@ test.describe('rum views', () => {
...
+  createTest('has a loading time')
+    .withRum()
+    .run(async ({ flushEvents, intakeRegistry }) => {
+      await flushEvents()
+      const viewEvent = intakeRegistry.rumViewEvents[0]
+      expect(viewEvent).toBeDefined()
+      expect(viewEvent.view.loading_time).toBeGreaterThan(0)
+    })
+
+  createTest('excludes some dom mutation when computing the loading time')
```
thomas-lebeau, line 27: **❓ question:** ‏ shouldn't we also test that it includes dom mutation in a normal case?

packages/rum-core/src/browser/domMutationObservable.ts
```diff
@@ -1,13 +1,42 @@
 import { monitor, noop, Observable, getZoneJsOriginalValue } from '@datadog/browser-core'
 
+export const IGNORE_MUTATIONS_ATTRIBUTE = 'dd-ignore-mutations'
+
```
bcaudan, line 4: **💬 suggestion:** ‏this does not seemed to be use anymore
```suggestion
```

test/e2e/scenario/rum/views.scenario.ts
```diff
@@ -15,6 +15,61 @@ test.describe('rum views', () => {
       expect(viewEvent.view.load_event).toBeGreaterThan(0)
     })
 
+  test.describe('loading time', () => {
+    // Simple spinner to test dom mutations.
+    const SPINNER = html`
+      <script type="module">
+        document.body.setAttribute('data-dd-excluded-activity-mutations', 'true')
```
thomas-lebeau, line 22: **💬 suggestion:** ‏ this is set twice for the test that uses it
```suggestion
```

BenoitZugmeyer: Thanks, I didn't meant to have it there.

PR #3278, opened 2025-01-13
✅ don't rely on segment counts on e2e tests

thomas-lebeau:
## Motivation

`record input interactions` and `don't record ignored input interactions` are the two most flacky e2e tests. This is because they are ensuring they only collect one segment, however, we have _recently_ [reduce the batch time limit](https://github.com/DataDog/browser-sdk/pull/3077) for replay segments to 5 seconds. It seems this test might takes longer sometimes, generating more segments that initially expected.

## Changes

Stop expecting a fix number of segments and flatmap search for records across all segments when needed.

This only fixes these two tests as they are the only real flaky ones, but we could extend this strategy to all other tests if needed.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/test/segments.ts
+++ b/packages/rum/test/segments.ts
@@ -37,10 +37,10 @@ export function findIncrementalSnapshot(
 
 // Returns all the IncrementalSnapshotRecord of a given source in a Segment, if any.
 export function findAllIncrementalSnapshots(
-  segments: BrowserSegment | BrowserSegment[],
+  { records }: { records: BrowserRecord[] },
   source: IncrementalSource
 ): BrowserIncrementalSnapshotRecord[] {
-  return getAllrecords(segments).filter(
+  return records.filter(
     (record) => record.type === RecordType.IncrementalSnapshot && record.data.source === source
   ) as BrowserIncrementalSnapshotRecord[]
 }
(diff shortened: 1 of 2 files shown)


packages/rum/test/segments.ts
```diff
@@ -59,3 +59,7 @@ export function findMouseInteractionRecords(
     (record) => 'type' in record.data && record.data.type === source
   )
 }
+
+export function getAllrecords(segments: BrowserSegment | BrowserSegment[]): BrowserRecord[] {
+  return Array.isArray(segments) ? segments.flatMap((segment) => segment.records) : segments.records
+}
```
BenoitZugmeyer, line 65: **💬 suggestion:** ‏I don't think we need this helper, as we can already query all records using `intakeRegistry.replayRecords`.

Maybe change `findAllIncrementalSnapshots` like so:

```ts
export function findAllIncrementalSnapshots({ records }: { record: BrowserRecord[] }, ...
```

and use it like `findAllIncrementalSnapshots({ records: intakeRegistry.replayRecords }, ...)`

BenoitZugmeyer (approved):

PR #3283, opened 2025-01-14
✨ [RUM-6581] Add an init parameter to chose feature flags event collection

RomanGaignault:
## Motivation

For now, feature flags are only included in views and errors. This was chosen to limit the feature flags impact on bandwidth, but now that we implemented compression, we could re-evaluate.

## Changes

Instead of enabling feature flag collections on all events at once, we could provide an init parameter to allow customers to chose on which events they want to collect feature flags:
init({
  collectFeatureFlagsOn: ['long_task', 'error', 'vital']
})

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/domain/collectFeatureFlags.ts
```diff
@@ -0,0 +1,19 @@
...
+  trackFeatureFlagsForEvents: FeatureFlagsForEvents[],
+  featureFlagContexts: FeatureFlagContexts,
+  rawRumEvent: { feature_flags?: Context }
+) {
+  if (trackFeatureFlagsForEvents.includes(eventType)) {
+    const featureFlagContext = featureFlagContexts.findFeatureFlagEvaluations(eventStartTime)
+    if (featureFlagContext && !isEmptyObject(featureFlagContext)) {
+      rawRumEvent.feature_flags = featureFlagContext
+    }
+  }
```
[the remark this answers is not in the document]
amortemousque: @RomanGaignault, just for context, the assembly is where we set attributes that apply on multiple event types.
For instance:
- If an attribute applies on only on resources => `resourceCollection.ts`
- If an attribute applies on resource, error etc. => `assembly.ts`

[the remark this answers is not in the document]
amortemousque: You can probably look at what we do [here](https://github.com/DataDog/browser-sdk/blob/54e65ab2678d12c8d47ed14a54cdbde7119accf7/packages/rum-core/src/domain/assembly.ts#L174)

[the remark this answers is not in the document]
RomanGaignault: Great,  I will update that ! Thanks for context 👍

packages/rum-core/src/domain/assembly.ts
```diff
@@ -235,3 +243,21 @@ function needToAssembleWithAction(
...
   return [RumEventType.ERROR, RumEventType.RESOURCE, RumEventType.LONG_TASK].indexOf(event.type) !== -1
 }
+
+function trackFeatureFlags(
+  rawRumEvent: RawRumEvent,
+  eventStartTime: RelativeTime,
+  trackFeatureFlagsForEvents: FeatureFlagsForEvents[],
+  featureFlagContexts: FeatureFlagContexts
+) {
+  const isdefaultTracking = rawRumEvent.type === RumEventType.VIEW || rawRumEvent.type === RumEventType.ERROR
```
BenoitZugmeyer, line 253: **🥜 nitpick:** ‏
```suggestion
  const isTrackingEnforced = rawRumEvent.type === RumEventType.VIEW || rawRumEvent.type === RumEventType.ERROR
```

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -19,6 +19,15 @@ import type { PropagatorType, TracingOption } from '../tracing/tracer.types'
...
 export const DEFAULT_PROPAGATOR_TYPES: PropagatorType[] = ['tracecontext', 'datadog']
 
+export const FeatureFlagsForEvents = {
+  ACTION: 'action',
+  RESOURCE: 'resource',
+  LONG_TASK: 'long_task',
+  VITAL: 'vital',
+} as const
+
+export type FeatureFlagsForEvents = (typeof FeatureFlagsForEvents)[keyof typeof FeatureFlagsForEvents]
```
BenoitZugmeyer, line 29: **💬 suggestion:** ‏I'm not sure we need this object. We never use its properties individually (like an enum).

We mostly use it to validate the configuration, but an array of supported values might be simpler:
```ts
const TRACK_FEATURE_FLAGS_FOR_EVENTS_SUPPORTED_VALUES = [RumEventType.ACTION, ...]
```

But maybe we don't even't need to go this far to validate the input. We could just accept any array without validating its values. It would reduce a bit the bundle size.

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -159,11 +173,22 @@ export interface RumConfiguration extends Configuration {
...
+  const trackFeatureFlagsForEvents: FeatureFlagsForEvents[] = []
+  if (Array.isArray(initConfiguration.trackFeatureFlagsForEvents)) {
+    initConfiguration.trackFeatureFlagsForEvents.forEach((eventType) => {
+      if (objectHasValue(FeatureFlagsForEvents, eventType) && !trackFeatureFlagsForEvents.includes(eventType)) {
+        trackFeatureFlagsForEvents.push(eventType)
+      } else if (!objectHasValue(FeatureFlagsForEvents, eventType)) {
+        display.warn(`Unknown event type '${eventType as any}' in trackFeatureFlagsForEvents configuration.`)
+      }
+    })
+  }
```
BenoitZugmeyer, line 191: **💬 suggestion:** ‏(follow-up of previous comment)
```suggestion
  if (!Array.isArray(initConfiguration.trackFeatureFlagsForEvents)) {
    display.warn(`trackFeatureFlagsForEvents should be an array`)
  }
```

packages/rum-core/src/domain/assembly.ts
```diff
@@ -235,3 +243,21 @@ function needToAssembleWithAction(
 ): event is RawRumErrorEvent | RawRumResourceEvent | RawRumLongTaskEvent {
   return [RumEventType.ERROR, RumEventType.RESOURCE, RumEventType.LONG_TASK].indexOf(event.type) !== -1
 }
+
+function trackFeatureFlags(
```
BenoitZugmeyer, line 247: **💬 suggestion:** might be closer to our existing naming conventions:‏
```suggestion
function findFeatureFlagsContext(
```

packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -873,6 +874,134 @@ describe('rum assembly', () => {
     })
   })
 
+  describe('feature flags', () => {
```
BenoitZugmeyer, line 877: **💬 suggestion:** ‏those tests are quite verbose, maybe we can refactor them like:

```ts
    it('should always include feature flags for view events', () => {
      assertFeatureFlagCollection(RumEventType.VIEW, [], true)
    })

    it('should always include feature flags for error events', () => {
      assertFeatureFlagCollection(RumEventType.ERROR, [], true)
    })

    it('should include feature flags only if "resource" is in config', () => {
      assertFeatureFlagCollection(RumEventType.RESOURCE, ['resource'], true)
      assertFeatureFlagCollection(RumEventType.RESOURCE, [], false)
    })

    it('should include feature flags only if "long_task" is in config', () => {
      assertFeatureFlagCollection(RumEventType.LONG_TASK, ['long_task'], true)
      assertFeatureFlagCollection(RumEventType.LONG_TASK, [], false)
    })

    it('should include feature flags only if "vital" is in config', () => {
      assertFeatureFlagCollection(RumEventType.VITAL, ['vital'], true)
      assertFeatureFlagCollection(RumEventType.VITAL, [], false)
    })

    it('should include feature flags only if "action" is in config', () => {
      assertFeatureFlagCollection(RumEventType.ACTION, ['action'], true)
      assertFeatureFlagCollection(RumEventType.ACTION, [], false)
    })

    function assertFeatureFlagCollection(
      eventType: RumEventType,
      trackFeatureFlagsForEvents: FeatureFlagsForEvents[],
      expectedToTrackFeatureFlags: boolean
    ) {
      const { lifeCycle, serverRumEvents, featureFlagContexts } = setupAssemblyTestWithDefaults({
        partialConfiguration: {
          trackFeatureFlagsForEvents,
        },
      })

      spyOn(featureFlagContexts, 'findFeatureFlagEvaluations').and.returnValue({
        'my-flag': 'enabled',
      })

      notifyRawRumEvent(lifeCycle, {
        rawRumEvent: createRawRumEvent(eventType),
      })

      if (expectedToTrackFeatureFlags) {
        expect(serverRumEvents[0].feature_flags).toEqual({ 'my-flag': 'enabled' })
      } else {
        expect(serverRumEvents[0].feature_flags).toBeUndefined()
      }
    }

```

BenoitZugmeyer (approved):

amortemousque (approved):

PR #3285, opened 2025-01-15
Report original error from `addReactError` instead of fake rendering error

eloytoro:
## Motivation

The ErrorBoundary component will capture whatever error occurs inside of it and send it to datadog, when this happens it renders the fallback component supplied to it. This works similar to a try/catch, where errors that occur within a component tree will not cause the entire application to crash and instead show a fallback.

As of today there’s a big problem with how ErrorBoundary reports these errors to datadog and how that plays into the Error Tracking product.

The original thrown error is not the one being reported, instead a different error is with a different stack trace (the one obtained from react’s error info) and a different error name (or @error.type in datadog). This messes with Error Tracking’s grouping.

This is the anatomy of a react error sent to Datadog today:

image

It would make more sense to report the original error so it has the original stack and name, drop the cause since it wouldn’t be needed anymore and add the the component’s stack as an additional `component_stack` property reported to datadog.

## Changes

In order to achieve this it's necessary to expose the `strategy` object through a new hook in plugins named `onRumStart`. Now additional information can be added to the payload outside of the event's `context`

Only after that the `strategy` object is accessed by the react plugin and used in `addReactError` to send the component's stack

## Testing

Unit tests were added

### DIFF
--- a/packages/rum-core/src/boot/preStartRum.ts
+++ b/packages/rum-core/src/boot/preStartRum.ts
@@ -137,7 +137,7 @@ export function createPreStartStrategy(
     bufferApiCalls.add((startRumResult) => startRumResult.addDurationVital(vital))
   }
 
-  return {
+  const strategy: Strategy = {
     init(initConfiguration, publicApi) {
       if (!initConfiguration) {
         display.error('Missing configuration')
@@ -229,6 +229,8 @@ export function createPreStartStrategy(
 
     addDurationVital,
   }
+
+  return strategy
 }
 
 function overrideInitConfigurationForBridge(initConfiguration: RumInitConfiguration): RumInitConfiguration {
--- a/packages/rum-react/src/domain/error/addReactError.ts
+++ b/packages/rum-react/src/domain/error/addReactError.ts
@@ -3,8 +3,8 @@ import { callMonitored, clocksNow, createHandlingStack } from '@datadog/browser-
 import { onRumStart } from '../reactPlugin'
 
 export function addReactError(error: Error, info: ErrorInfo) {
+  const handlingStack = createHandlingStack()
   onRumStart((strategy) => {
-    const handlingStack = createHandlingStack()
     callMonitored(() => {
       strategy.addError({
         error,
--- a/packages/rum-react/src/domain/error/addReactError.ts
+++ b/packages/rum-react/src/domain/error/addReactError.ts
@@ -1,13 +1,18 @@
 import type { ErrorInfo } from 'react'
-import type { ErrorWithCause } from '@datadog/browser-core'
-import { onReactPluginInit } from '../reactPlugin'
+import { callMonitored, clocksNow, createHandlingStack } from '@datadog/browser-core'
+import { onRumStart } from '../reactPlugin'
 
 export function addReactError(error: Error, info: ErrorInfo) {
-  const renderingError = new Error(error.message)
-  renderingError.name = 'ReactRenderingError'
-  renderingError.stack = info.componentStack ?? undefined
-  ;(renderingError as ErrorWithCause).cause = error
-  onReactPluginInit((_, rumPublicApi) => {
-    rumPublicApi.addError(renderingError, { framework: 'react' })
+  onRumStart((strategy) => {
+    const handlingStack = createHandlingStack()
+    callMonitored(() => {
+      strategy.addError({
+        error,
+        handlingStack,
+        componentStack: info.componentStack ?? undefined,
+        context: { framework: 'react' },
+        startClocks: clocksNow(),
+      })
+    })
   })
 }
--- a/packages/rum-react/src/domain/error/addReactError.ts
+++ b/packages/rum-react/src/domain/error/addReactError.ts
@@ -4,6 +4,7 @@ import { onRumStart } from '../reactPlugin'
 
 export function addReactError(error: Error, info: ErrorInfo) {
   const handlingStack = createHandlingStack()
+  const startClocks = clocksNow()
   onRumStart((strategy) => {
     callMonitored(() => {
       strategy.addError({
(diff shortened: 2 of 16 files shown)


packages/rum-react/src/domain/error/addReactError.ts
```diff
@@ -1,13 +1,18 @@
...
 
 export function addReactError(error: Error, info: ErrorInfo) {
-  const renderingError = new Error(error.message)
-  renderingError.name = 'ReactRenderingError'
-  renderingError.stack = info.componentStack ?? undefined
-  ;(renderingError as ErrorWithCause).cause = error
-  onReactPluginInit((_, rumPublicApi) => {
-    rumPublicApi.addError(renderingError, { framework: 'react' })
+  onRumStart((strategy) => {
+    const handlingStack = createHandlingStack()
```
BenoitZugmeyer, line 7: **💬 suggestion:** ‏move `const handlingStack = createHandlingStack()` directly in `addReactError` scope, so it actually points to where `addReactError` is called.

```suggestion
  const handlingStack = createHandlingStack()
  onRumStart((strategy) => {
```

packages/rum-react/src/domain/reactPlugin.ts
```diff
@@ -1,10 +1,13 @@
...
 let globalPublicApi: RumPublicApi | undefined
 let globalConfiguration: ReactPluginConfiguration | undefined
-type Subscriber = (configuration: ReactPluginConfiguration, rumPublicApi: RumPublicApi) => void
+let globalStrategy: Strategy | undefined
+type InitSubscriber = (configuration: ReactPluginConfiguration, rumPublicApi: RumPublicApi) => void
+type StartSubscriber = (strategy: Strategy) => void
 
-const onReactPluginInitSubscribers: Subscriber[] = []
+const onRumInitSubscribers: InitSubscriber[] = []
+const onRumStartSubscribers: StartSubscriber[] = []
```
BenoitZugmeyer, line 10: **Thoughts**: maybe at some point we should introduce an abstraction to avoid all those module-scoped variables. For now it seems reasonable though?

packages/rum-react/src/domain/error/addReactError.ts
```diff
@@ -1,13 +1,18 @@
...
+    callMonitored(() => {
+      strategy.addError({
+        error,
+        handlingStack,
+        componentStack: info.componentStack ?? undefined,
+        context: { framework: 'react' },
+        startClocks: clocksNow(),
```
BenoitZugmeyer, line 14: **Suggestion**: maybe move `clocksNow()` into `addReactError` scope directly, so the date better represents when the error happens and does not depend on the time it takes to start rum.

```ts
function addReactError(...) {
  const startClocks = clocksNow()
  onRumStart((...) => {
    addError({
       ...,
       startClocks,
     })
  })
}
```

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -137,7 +137,7 @@ export function createPreStartStrategy(
     bufferApiCalls.add((startRumResult) => startRumResult.addDurationVital(vital))
   }
 
-  return {
+  const strategy: Strategy = {
```
amortemousque, line 140: **💬 suggestion:** ‏What about passing the `strategy` in `onInit` hook instead of `publicApi`. This would limit the number of concepts in the React SDK. To be more explicit we could also rename it `startRumStrategy`

BenoitZugmeyer: Passing the strategy in `onInit` is not that useful because it would be the `preStartStrategy` and there is a good chance that it won't be used right after the hook is called. I think it would make things more complex/confusing for little benefit.

eloytoro: + to what benoit just said, the `preStartStrategy` buffer has already been flushed by the time `onInit` is called, which means that any usage of it would be completely ignored

BenoitZugmeyer: @eloytoro and I briefly discussed exposing the `lifeCycle` instead of the `strategy`, which might make things more flexible, but for now `strategy` seems simpler/enough. We can revisit later.

packages/rum-react/src/domain/reactPlugin.ts
eloytoro: yes im just following the existing pattern... not that i like it 😬

packages/rum-react/src/domain/error/addReactError.ts
eloytoro: https://github.com/DataDog/browser-sdk/pull/3285/commits/e2450c9e24d3ab6f204f39a331702f917546116f

packages/rum-react/src/domain/error/addReactError.ts
eloytoro: https://github.com/DataDog/browser-sdk/pull/3285/commits/43b7fd57e8ef3d1e725e8ca412249b27e7bfecfa

packages/rum-react/src/domain/reactPlugin.ts
BenoitZugmeyer: We probably should improve this at some point anyway, like avoid leaking subscribers when RUM is never initialized. We'll see when we get there.

BenoitZugmeyer (approved):

amortemousque (approved):

packages/rum-react/src/domain/reactPlugin.ts
amortemousque: Can we pass the `startRumResult` instead then? At least it is more explicit than 'strategy'

packages/rum-core/src/boot/preStartRum.ts
amortemousque: Can we pass the startRumResult instead then? At least it is more explicit than 'strategy'

eloytoro:
Closing this in favour of https://github.com/DataDog/browser-sdk/pull/3293 because the CI does not like PRs across forks

PR #3289, opened 2025-01-16
♻️ [RUM-8123] Introduce a hook to assemble events

amortemousque:
## Motivation

The overall goal is to create a Modular SDK to simplify contributions and enhance extensibility.

This PR introduces a **hook system**, initially used to `decouple event assembly`. Each collection module can now independently handle the assembly of its own properties.

## Changes

- Introduce a Hooks system
- Use the Assemble hook in viewCollection to manage the assembly of view properties
- Use the Assemble hook in urlContext to manage the assembly of url properties

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -181,8 +181,6 @@ export function startRumAssembly(
           rawRumEvent
         ) as RumEvent & Context
 
-        serverRumEvent.context = combine(serverRumEvent.context, customerContext)
-
         if (!('has_replay' in serverRumEvent.session)) {
           ;(serverRumEvent.session as Mutable<RumEvent['session']>).has_replay = commonContext.hasReplay
         }
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -155,8 +155,6 @@ export function startRumAssembly(
             id: configuration.applicationId,
           },
           date: timeStampNow(),
-          service: configuration.service,
-          version: configuration.version,
           source: 'browser',
           session: {
             id: session.id,
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -22,13 +22,14 @@ import type {
 } from '../rawRumEvent.types'
 import { RumEventType } from '../rawRumEvent.types'
 import type { RumEvent } from '../rumEvent.types'
+import type { Hooks } from '../hooks'
+import { HookNames } from '../hooks'
 import { getSyntheticsContext } from './contexts/syntheticsContext'
 import type { CiVisibilityContext } from './contexts/ciVisibilityContext'
 import type { LifeCycle } from './lifeCycle'
 import { LifeCycleEventType } from './lifeCycle'
 import type { ViewHistory } from './contexts/viewHistory'
 import { SessionReplayState, type RumSessionManager } from './rumSessionManager'
-import type { UrlContexts } from './contexts/urlContexts'
 import type { RumConfiguration } from './configuration'
 import type { ActionContexts } from './action/actionCollection'
 import type { DisplayContext } from './contexts/displayContext'
@@ -67,9 +68,9 @@ type Mutable<T> = { -readonly [P in keyof T]: T[P] }
 export function startRumAssembly(
   configuration: RumConfiguration,
   lifeCycle: LifeCycle,
+  hooks: Hooks,
   sessionManager: RumSessionManager,
   viewHistory: ViewHistory,
-  urlContexts: UrlContexts,
   actionContexts: ActionContexts,
   displayContext: DisplayContext,
   ciVisibilityContext: CiVisibilityContext,
@@ -134,9 +135,9 @@ export function startRumAssembly(
     LifeCycleEventType.RAW_RUM_EVENT_COLLECTED,
     ({ startTime, rawRumEvent, domainContext, savedCommonContext, customerContext }) => {
       const viewHistoryEntry = viewHistory.findView(startTime)
-      const urlContext = urlContexts.findUrl(startTime)
       const session = sessionManager.findTrackedSession(startTime)
-      if (session && viewHistoryEntry && urlContext) {
+
+      if (session && viewHistoryEntry) {
         const commonContext = savedCommonContext || getCommonContext()
         const actionId = actionContexts.findActionId(startTime)
 
@@ -154,8 +155,8 @@ export function startRumAssembly(
             id: configuration.applicationId,
           },
           date: timeStampNow(),
-          service: viewHistoryEntry.service || configuration.service,
-          version: viewHistoryEntry.version || configuration.version,
+          service: configuration.service,
+          version: configuration.version,
           source: 'browser',
           session: {
             id: session.id,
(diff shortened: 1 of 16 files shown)


packages/rum-core/src/domain/assembly.ts
```diff
@@ -134,13 +129,13 @@ export function startRumAssembly(
...
-      const urlContext = urlContexts.findUrl(startTime)
       const session = sessionManager.findTrackedSession(startTime)
-      if (session && viewHistoryEntry && urlContext) {
+
+      if (session && viewHistoryEntry) {
         const commonContext = savedCommonContext || getCommonContext()
         const actionId = actionContexts.findActionId(startTime)
 
-        const rumContext: RumContext = {
+        const rumContext: Partial<CommonProperties> = {
```
amortemousque, line 138: Change unrelated to the PR, but we can remove `RumContext` and directly use the `CommonProperties` type generated by the rum-event-format

packages/rum-core/src/hooks.ts
```diff
@@ -0,0 +1,59 @@
...
+        },
+      }
+    },
+    triggerHook<K extends keyof HookCallbackMap>(
+      hookName: K,
+      param: Parameters<HookCallbackMap[K]>[0]
+    ): ReturnType<HookCallbackMap[K]> {
+      const hookCallbacks = callbacks[hookName] || []
+      const results = hookCallbacks.map((callback) => callback(param))
+      return combine(...(results as [object, object])) as ReturnType<HookCallbackMap[K]>
```
nulrich, line 56: **💬 suggestion:** ‏The result looks nice. I'm just a bit concerned about debugging the Assembly Hook for overridden props if we connect many sources to it because we can't easily know the order of declarations. Maybe we could warn user when a prop is being overridden?

nulrich (approved):

packages/rum-core/src/hooks.ts
sethfowler-datadog: Yeah, I remember we talked about this a bit too, @amortemousque. What do you think of the idea of requiring all hooks to be registered in a single place and specifying them with an array that makes the order clear?

It doesn't look like you actually use the `unregister` callback returned from `register()` anywhere, but if there's a need to dynamically add and remove these hooks individually, an alternative approach that might work would be to support enabling and disabling them. That way, they'd keep their initially-defined order, but you'd still be able to turn them on and off when needed.

packages/rum-core/src/hooks.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Define a partial RUM event type.
+// Ensuring the `type` field is always present improves type checking, especially in conditional logic in hooks (e.g., `if (eventType === 'view')`).
+export type PartialRumEvent = RecursivePartialExcept<RumEvent, 'type'>
+
+// This is a workaround for an issue occurring when the Browser SDK is included in a TypeScript
+// project configured with `isolatedModules: true`. Even if the const enum is declared in this
+// module, we cannot use it directly to define the EventMap interface keys (TS error: "Cannot access
+// ambient const enums when the '--isolatedModules' flag is provided.").
```
sethfowler-datadog, line 22: An alternative would be to use a union of string literal types instead of a `const enum`:
```ts
type HookName = 'assemble' /* | 'someOtherHook' */
```
I'm sure you thought of that, though, so I'm guessing you just prefer the dot syntax enabled by `const enum`!

sethfowler-datadog (commented):
I love the direction this takes us in!

packages/rum-core/src/hooks.ts
amortemousque: I agree! We discussed this earlier with @sethfowler-datadog, and I experimented with centralized registration. However, this approach adds some boilerplate and coupling, as modules would need to expose their hooks. Self-registration offers better decoupling.

In my view, the assemble hook will be used by `contexts` and `collections`, which are explicitly called in `startRum`. This makes the order relatively easy to visualize. For now, I suggest waiting to see if this becomes an issue before making improvements. Wdyt?

sethfowler-datadog: > For now, I suggest waiting to see if this becomes an issue before making improvements. Wdyt?

That'd be fine with me, if you prefer to take that direction!

> ... this approach adds some boilerplate and coupling, as modules would need to expose their hooks. Self-registration offers better decoupling.

FWIW, my perspective is that centralized registration does not add any additional coupling; it just turns _implicit_ dependencies, which are hard to reason about, into _explicit_ dependencies, which are easier to reason about. Another way to think about it is that this isn't coupling at all, but composition: we start with small pieces of functionality that can be reasoned about independently, and compose them together in an explicit way to form larger pieces of functionality. You end up with a hierarchical structure, which is nice for all the reasons that structured programming in general is nice.

That said, programming is all about tradeoffs, and there are pros and cons to every solution; I'm totally fine with a dynamic registration approach if you prefer it. As you say, we can always revisit the issue later.

packages/rum-core/src/domain/assembly.ts
```diff
@@ -165,21 +166,22 @@ export function startRumAssembly(
...
-        const serverRumEvent = combine(rumContext as RumContext & Context, rawRumEvent) as RumEvent & Context
-        serverRumEvent.context = combine(commonContext.context, viewHistoryEntry.context, customerContext)
+        const serverRumEvent = combine(
+          rumContext,
+          hooks.triggerHook(HookNames.Assemble, { eventType: rawRumEvent.type, startTime }) as RumEvent & Context,
+          { context: customerContext },
+          rawRumEvent
+        ) as RumEvent & Context
+
+        serverRumEvent.context = combine(serverRumEvent.context, customerContext)
```
BenoitZugmeyer, line 184: **💬 suggestion:** ‏I don't think we need this line anymore?

packages/rum-core/src/hooks.ts
BenoitZugmeyer: Also, `const enum` takes less space in the final bundle: `HookNames.Assemble` gets compiled to `0` instead of `"assemble"`.

packages/rum-core/src/domain/contexts/urlContexts.ts
```diff
@@ -69,6 +72,18 @@ export function startUrlContexts(
     }
   }
 
+  hooks.register(HookNames.Assemble, ({ startTime, eventType }): PartialRumEvent | undefined => {
```
BenoitZugmeyer, line 75: **🥜 nitpick:** ‏same here

```suggestion
  hooks.register(HookNames.Assemble, ({ startTime, eventType }) => {
```

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -33,11 +38,27 @@ export function startViewCollection(
       processViewUpdate(view, configuration, featureFlagContexts, recorderApi, pageStateHistory)
     )
   )
+
+  hooks.register(HookNames.Assemble, ({ startTime, eventType }): PartialRumEvent | undefined => {
```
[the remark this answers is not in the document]
amortemousque: I don't know why but it is required for the typechecking to work properly.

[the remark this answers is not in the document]
BenoitZugmeyer: Weird. Do you have an example where there is an issue with the return type but not without the return type?

packages/rum-core/src/hooks.ts
sethfowler-datadog: True!

BenoitZugmeyer (approved):

PR #3290, opened 2025-01-17
♻️ Custom sanitizer for Context Manager

nulrich:
## Motivation

When using Context Manager in our Public API, we manually call the sanitazition functions at multiple places before setting the context.

```typescript
setUserProperty: monitor((key, property) => {
  const sanitizedProperty = sanitizeUser({ [key]: property })[key]
  userContextManager.setContextProperty(key, sanitizedProperty)
}),
```

This could be done automatically by the Context Manager itself and just write

```typescript
setUserProperty: monitor((key, property) => {
  userContextManager.setContextProperty(key, property)
}),
```

## Changes

Now we can pass the sanization configuration to `createContextManager`. It will be applied  whenever we update the internal context.

```typescript
const userContextManager = createContextManager('user', {
  customerDataTracker: customerDataTrackerManager.getOrCreateTracker(CustomerDataType.User),
  propertiesConfig: {
    id: { type: 'string' },
    name: { type: 'string' },
    email: { type: 'string' },
  },
})
```

Also the configuration can also warn the user if a required property is missing:

```typescript
const userContextManager = createContextManager('test', {
  customerDataTracker: customerDataTrackerManager.getOrCreateTracker(CustomerDataType.User),
  propertiesConfig: { id: { required: true } },
})
```

```typescript
manager.setContext({ foo: 'bar' })
// warn: The property id of test context is required; context will not be sent to the intake.

```

## Testing

- [X] Local
- [ ] Staging
- [X] Unit
- [ ] End to end

---


packages/core/src/domain/context/contextManager.ts
```diff
@@ -17,6 +23,9 @@ export function createContextManager(customerDataTracker?: CustomerDataTracker)
     setContext: (newContext: Context) => {
       if (getType(newContext) === 'object') {
         context = sanitize(newContext)
+        if (contextSanitizer) {
+          context = contextSanitizer(context)
+        }
```
amortemousque, line 28: **💬 suggestion:** Having both `sanitize` and `contextSanitizer` is confusing. Maybe we could add param to the `sanitize` function to modify its behavior for certain fields. Something in similar to the [replacer of JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer)?

nulrich: Nice! My use case is to warn the user that a property is missing (ex: id). Maybe we could extend to:

```typescript
createContextManager('User', {
  customerDataTracker: customerDataTrackerManager.getOrCreateTracker(),
  ensureProperties: {
    id: { type: 'string', required: true },
    name: { type: 'string' },
    email: { type: 'string' },
  },
})
```

We'll use the `type` property to customize `sanitize()` and the `required` property to simply warn the user that the property is missing.

WDYT?

sethfowler-datadog: I like that; the result feels very readable to me!

amortemousque: Yes nice!

packages/core/src/domain/context/contextManager.ts
```diff
@@ -3,11 +3,48 @@ import { getType } from '../../tools/utils/typeUtils'
...
+  return newContext
+}
+
+function checkRequiredProperties(context: Context, propertiesConfig: PropertiesConfig, name: string) {
+  for (const [key, { required }] of Object.entries(propertiesConfig)) {
+    if (required && !(key in context)) {
+      display.warn(`The property ${key} of ${name} context is required; context will not be sent to the intake.`)
+    }
+  }
+}
```
amortemousque, line 36: **💬 suggestion:** ‏I think it could be beneficial to include these as part part of the `sanitize` module. 

```ts
sanitize(context, {
  ensureProperty: {
    id: ' { type: 'string', required: true },
    name:  { type: 'string' },
    email:  { type: 'string' },
  }
});
```
Also you could merge `enforceTypeProperties` and `checkRequiredProperties` into a single function. This would allow iterating over `propertiesConfig` only once
```ts
function ensureProperties(context, propertiesConfig) {
...
}
```

nulrich: I merged the functions 👍 

Concerning moving the code into `sanitize` module, I found that the type enforcement and requirement are specific to context management and do not help to sanitize the content. 

The type enforcement is more for controlling the content for the intake than preventing illegal data, and property requirement is not really sanitization work.

Also `sanitize` is checking all properties recursively so we would need to pass the path to properties we want to check and not only supporting the first level.

amortemousque (approved):

sethfowler-datadog (approved):
Love it!

PR #3295, opened 2025-01-21
✅ mitigate e2e protocol latency

thomas-lebeau:
## Motivation

two e2e test start failing quite consistently since around a week ago on Chrome android. The root cause is increased latency in the webdriver <> browserstack connection wich makes 3 consecutive clicks happens slower than a second, not triggering the expected `rage_click` frustration.

## Changes

The mitigation involve using synthetic events instead of webdriver's API to guaranty the clicks happens fast enough to be recognized as rage clicks.

This is not ideal and will be revisited in the future when we move away from webdriver protocol.

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/test/e2e/scenario/recorder/recorder.scenario.ts
+++ b/test/e2e/scenario/recorder/recorder.scenario.ts
@@ -683,7 +683,7 @@ describe('recorder', () => {
       })
 
     createTest('should detect a rage click and match it to mouse interaction records')
-      .withRum({ trackUserInteractions: true })
+      .withRum({ trackUserInteractions: true, allowUntrustedEvents: true })
       .withSetup(bundleSetup)
       .withBody(html`
         <div id="main-div" />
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer (approved):

RomanGaignault (approved):

test/e2e/scenario/recorder/recorder.scenario.ts
```diff
@@ -683,7 +683,7 @@ describe('recorder', () => {
       })
 
     createTest('should detect a rage click and match it to mouse interaction records')
-      .withRum({ trackUserInteractions: true })
+      .withRum({ trackUserInteractions: true, allowUntrustedEvents: true })
```
BenoitZugmeyer, line 686: **🥜 nitpick:** ‏https://github.com/DataDog/browser-sdk/blob/86db84245a850341b24b1e977d1d6b4a6d61e3a2/test/e2e/lib/framework/createTest.ts#L28 `allowUntrustedEvents: true` should already be enabled by default

BenoitZugmeyer (approved):

PR #3299, opened 2025-01-24
✨ [RUM-5500] React-router v7 support

RomanGaignault:
## Motivation

Add support for react-router v7 for the rum-react plugin.

## Changes

Reuse as much as possible the code from react-router-6, adapt unit test instead of cloning them

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-react/package.json
```diff
@@ -38,7 +38,8 @@
     "@types/react-dom": "18.3.5",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "react-router-dom": "6.28.2"
+    "react-router-6": "npm:react-router-dom@6.28.2",
+    "react-router-7": "npm:react-router-dom@7.1.3"
```
BenoitZugmeyer, line 42: **💬 suggestion:** ‏
```suggestion
    "react-router-dom-6": "npm:react-router-dom@6.28.2",
    "react-router-dom-7": "npm:react-router-dom@7.1.3"
```

packages/rum-react/src/domain/reactRouterV7/createRouter.ts
```diff
@@ -0,0 +1,28 @@
...
+  router.subscribe((routerState) => {
+    const newPathname = routerState.location.pathname
+    if (location !== newPathname) {
+      startReactRouterView(routerState.matches)
+      location = newPathname
+    }
+  })
+  startReactRouterView(router.state.matches)
+  return router
+}
```
BenoitZugmeyer, line 28: **💬 suggestion:** ‏To help maintenance, we need to avoid code duplication as much as possible.
What about having a structure like this:

```
domain
  reactRouter
    startReactRouterView.ts
    useRoutes.ts
    createRouterts
    routesComponent.ts
entries
  reactRouterV6.ts
  reactRouterV7.ts
```

Entry files would look like:
```ts
import { wrapCreateRouter } from '../domain/reactRouter/createRouter'
import { createRoutesComponent } from '../domain/reactRouter/routesComponent'
import {
  createBrowserRouter as originalCreateBrowserRouter,
  createHashRouter as originalCreateHashRouter,
  createMemoryRouter as originalCreateMemoryRouter,
  useRoutes as originalUseRoutes,
  useLocation,
} from 'react-router-7'

export const createBrowserRouter = wrapCreateRouter(originalCreateBrowserRouter)
export const createHashRouter = wrapCreateRouter(originalCreateHashRouter)
export const createMemoryRouter = wrapCreateRouter(originalCreateMemoryRouter)
export const useRoutes = wrapUseRoutes({ useRoutes: originalUseRoutes, useLocation, ... })
export const Routes = createRoutesComponent(useRoutes)
```

And domain files would look like (types might be a bit harder to get well, let's see):

```ts
export function wrapCreateRouter(originalCreateRouter) {
  return (routes, options) => registerRouter(originalCreateRouter(routes, options))
}
```

packages/rum-react/src/domain/reactRouter/types.ts
```diff
@@ -0,0 +1,14 @@
+export type AnyRouteObject = { path?: string | undefined }
```
BenoitZugmeyer, line 1: **💬 suggestion:** ‏Add a comment at the top of the file to explain those types, like:

```ts
// Those types are used by our instrumentation functions to make them work with any react-router
// version we support. We should not import react-router types as we don't know which version
// will be used by the customer.
//
// Those types should be:
// * compatible with all react-router-dom versions we support
// * include the minimal set of attributes used by our instrumentation functions.
```

packages/rum-react/src/entries/reactRouterV6.ts
Ayc0, line 1: Open question: is there a way in the browser-sdk to mark a file as existing only for tests?
When looking at this file without too much context, it can be hard to understand that this should only be used by spec files / the sandbox react app, and not by code exposed to npm

Miz85 (commented):
LGTM! 

With v7 it looks like you can use react router as a routing library or as a framework. I think this PR covers everything from the library but in the framework part, they introduced a [route modules](https://reactrouter.com/start/framework/routing#route-modules) and a `route` wrapper function. Maybe we could try to see if we can also instrument that? It def. can come in another PR though.

Miz85:
Just tested on a new app using React Router v7 and it seems to work fine! I do have a weird type error though:
image

I built and linked `@datadog/browser-rum-react` from this branch and I installed `@datadog/browser-rum-react` from npm (^6.3.0).

I don't seem to have it when I also link `@datadog/browser-rum` built on this branch so it might be just a version discrepency but thought I'd mention it!

packages/rum-react/src/domain/reactRouter/useRoutes.spec.tsx
```diff
@@ -186,33 +177,26 @@ versions.forEach(({ version, MemoryRouter, useNavigate, useRoutes }) => {
...
+      act(() => {
         navigate!('/foo?bar=baz')
       })
 
       expect(startViewSpy).not.toHaveBeenCalled()
     })
 
     it('does not start a new view if it does not match any route', () => {
-      // Prevent react router from showing a warning in the console when a route does not match
+      // On empêche react-router d'afficher un avertissement dans la console si aucune route ne correspond
```
BenoitZugmeyer, line 194: **🔨 warning:** ‏Why this translation?

scripts/check-licenses.js
```diff
@@ -36,9 +41,16 @@ runMain(async () => {
...
     .concat(Object.keys(packageJsonFile.content.devDependencies || {}))
     .filter((dependency) => !dependency.includes('@datadog'))
+  return deps.map((dependency) => {
+    let packageName = dependency
+    if (aliasMapping[packageName]) {
+      packageName = aliasMapping[packageName]
+    }
+    return packageName
+  })
 }
```
BenoitZugmeyer, line 54: **💬 suggestion:** ‏you can avoid the alias map by extracting the original dependency name from the version string:

```js

function retrievePackageDependencies(packageJsonFile) {
  return Object.entries(packageJsonFile.content.dependencies || {})
    .concat(Object.entries(packageJsonFile.content.devDependencies || {}))
    .map(([dependency, version]) => {
      if (version.startsWith('npm:')) {
        // Extract the original dependency name from the npm protocol version string. Example:
        // npm:react@17  ->  react
        return version.slice(4).split('@')[0]
      }
      return dependency
    })
    .filter((dependency) => !dependency.includes('@datadog'))
}

```

packages/rum-react/src/entries/reactRouterV6.ts
BenoitZugmeyer: * Code that is intended for the user is within `packages/*/src` except `*.spec.*` files.
* Code for tests are `*.spec.*` files and files within `test/` folders
* Sandbox is not user facing

Does it answer your question?

packages/rum-react/src/domain/reactRouter/startReactRouterView.spec.ts
```diff
@@ -11,22 +11,21 @@ import {
...
+    startReactRouterView,
+    computeViewName,
   },
   {
     version: 'react-router-7',
     createMemoryRouter: createMemoryRouterV7,
-    startReactRouterView: ViewV7.startReactRouterView,
-    computeViewName: ViewV7.computeViewName,
+    startReactRouterView,
+    computeViewName,
```
BenoitZugmeyer, line 28: **💬 suggestion:** ‏you could remove those two functions from `routerVersions` and use them directly.

BenoitZugmeyer (approved):

PR #3305, opened 2025-01-29
♻️ [RUM-8319] Use assembly hook for action

amortemousque:
## Motivation

The overall goal is to create a Modular SDK to simplify contributions and enhance extensibility.
This PR uses the assemble hook to decouple action assembly

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -43,6 +46,26 @@ export function startActionCollection(
...
+      return
+    }
+
+    const actionId = actionContexts.findActionId(startTime)
+    if (!actionId) {
+      return
+    }
+
+    return {
+      type: eventType,
```
sethfowler-datadog, line 64: The event will already have a `type` property, right? Assuming that's true, it may make sense to only have the hook return the properties it's adding or changing.

sethfowler-datadog (approved):
LGTM!

packages/rum-core/src/domain/action/actionCollection.ts
amortemousque: You're right! Specifying the type is a bit of an ugly hack to improve type checking :disappointed:. See https://github.com/DataDog/browser-sdk/blob/61fda307c02d80acbe694a9d2cab5646348401e8/packages/rum-core/src/hooks.ts#L16

amortemousque: Happy to discuss a better alternative :)

PR #3306, opened 2025-01-30
👷‍♀️ [RUM-8295] Change anonymous id format to uuid

cy-moi:
## Motivation
Change anonymous id format to uuid so we achieve:
1. Fewer collisions when >100M device concurrencies
2. Aligned with other ids and with mobile sdk

## Changes
Use `generateUUID()` for anonymous id generation
Update e2e tests and unit tests to test effectively.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---


packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
```diff
@@ -103,33 +101,3 @@ describe('session in cookie strategy', () => {
     })
   })
 })
-describe('session in cookie strategy when opt-out anonymous user tracking', () => {
```
thomas-lebeau, line 106: **❓ question:** ‏ Why removing these tests?

cy-moi: Sorry, this one is wrongly removed. I put them back.

test/e2e/scenario/rum/sessions.scenario.ts
```diff
@@ -78,16 +80,6 @@ describe('rum sessions', () => {
 
         expect((await findSessionCookie())?.aid).toEqual(anonymousId)
       })
-
-    createTest('generated when cookie is cleared')
```
cy-moi, line 82: This is removed as `renewSession` is basically putting the id back manually. So the test is not much meaningful. Plus, we should not restart a session when the cookie cleared by third party.

test/e2e/scenario/rum/sessions.scenario.ts
```diff
@@ -56,7 +56,9 @@ describe('rum sessions', () => {
       .run(async () => {
         const anonymousId = (await findSessionCookie())?.aid
 
-        await expireSession()
+        await browser.execute(() => {
+          window.DD_RUM!.stopSession()
+        })
```
cy-moi, line 61: Avoid calling manual `expireSession` here to test the anonymous id logic fully.

thomas-lebeau: I think the expireSession helper could be improve to actually expire the session.
I think with playwright we could use `await page.clock.runFor(SESSION_TIME_OUT_DELAY)`

maybe you can add a not and we can revert this change when we have migrated to playwright fully

thomas-lebeau (approved):

PR #3307, opened 2025-01-30
🐛 fix document resource duration

BenoitZugmeyer:
## Motivation

Before this PR, we used the navigation entry duration as the duration for the document resource. However, that duration includes the time it takes to load the page until the `load` event (including the load of assets declared in the HTML like JS, CSS, images...).

## Changes

This PR fixes this by adjusting the duration to `responseEnd`, which conforms with the duration of other resource entries.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


sethfowler-datadog (approved):
Nice!

thomas-lebeau (approved):

PR #3314, opened 2025-02-03
🐛 [RUM-8353] Throttle view context update

cy-moi:
## Motivation
Updating of view context automatically triggers view updates, which could trigger excessive requests in edge cases. We should throttle view updates while maintaining current view context update behavior.

## Changes
Added a new LifeCycleEvent type so we can update the view context for in viewHistory immediately
Use `scheduleViewUpdate` to throttle the view updates

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/view/trackViews.spec.ts
@@ -924,6 +924,8 @@ describe('view event count', () => {
       const { getViewUpdate, setViewContext } = viewTest
 
       setViewContext({ foo: 'bar' })
+      clock.tick(3000)
+
       expect(getViewUpdate(1).context).toEqual({ foo: 'bar' })
     })
 
@@ -932,6 +934,8 @@ describe('view event count', () => {
       const { getViewUpdate, setViewContextProperty } = viewTest
 
       setViewContextProperty('foo', 'bar')
+      clock.tick(3000)
+
       expect(getViewUpdate(1).context).toEqual({ foo: 'bar' })
     })
 
--- a/packages/rum-core/src/domain/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/view/trackViews.spec.ts
@@ -924,7 +924,7 @@ describe('view event count', () => {
       const { getViewUpdate, setViewContext } = viewTest
 
       setViewContext({ foo: 'bar' })
-      clock.tick(3000)
+      clock.tick(THROTTLE_VIEW_UPDATE_PERIOD)
 
       expect(getViewUpdate(1).context).toEqual({ foo: 'bar' })
     })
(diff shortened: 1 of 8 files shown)


packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -67,6 +67,13 @@ export interface ViewCreatedEvent {
   startClocks: ClocksState
 }
 
+export interface ViewContextEvent {
```
amortemousque, line 70: **💬 suggestion:** ‏
```suggestion
export interface BeforeViewUpdateEvent {
```

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -229,13 +236,9 @@ function newView(
...
 
   // Update the view every time the measures are changing
-  const { throttled: scheduleViewUpdate, cancel: cancelScheduleViewUpdate } = throttle(
-    triggerViewUpdate,
-    THROTTLE_VIEW_UPDATE_PERIOD,
-    {
-      leading: false,
-    }
-  )
+  const { throttled, cancel: cancelScheduleViewUpdate } = throttle(triggerViewUpdate, THROTTLE_VIEW_UPDATE_PERIOD, {
```
thomas-lebeau, line 239: **🔨 warning:** did you rename `scheduleViewUpdate` on purpose?‏ it looks like it is passed to `trackCommonViewMetrics`, `trackInitialViewMetrics` and `trackViewEventCounts`, but you didn't renamed it there.

I think I would prefer keeping `scheduleViewUpdate`.
Or if you prefer to rename it, use something a bit more specific like `throttledViewUpdate` and rename `cancelScheduleViewUpdate` too to be consistend

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -924,6 +924,8 @@ describe('view event count', () => {
       const { getViewUpdate, setViewContext } = viewTest
 
       setViewContext({ foo: 'bar' })
+      clock.tick(3000)
```
thomas-lebeau, line 927: **💬 suggestion:** ‏ Use `THROTTLE_VIEW_UPDATE_PERIOD` so we know where this value comes from and why we need it

packages/rum-core/src/domain/view/trackViews.ts
cy-moi: Hi, I was in the middle of making changes according to previous comments. Some tests needed to be fixed if we rely on `beforeViewUpdate` everywhere. But I do feel like this is somewhat confusing as well. Maybe we talk about this during the parking lot.

amortemousque: I think it's just that now `scheduleViewUpdate` look like this:
```ts
const scheduleViewUpdate = () => { 
  triggerBeforeViewUpdate() 
  throttled() 
}
```
`throttled` should only be used in this function all other dep should use `scheduleViewUpdate`
Wdyt?

thomas-lebeau: 🤦 of course! makes sense, I should have seen that! thanks

thomas-lebeau (approved):

amortemousque (approved):

PR #3320, opened 2025-02-04
✨ [RUM-7459] Collect error context from a custom Error field

RomanGaignault:
## Motivation

When throwing an exception in an App, Browser SDK users sometimes want to provide more context that is only accessible locally when the error is constructed.

## Changes

support error.dd_context as a custom error property

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/logs/src/domain/report/reportCollection.ts
+++ b/packages/logs/src/domain/report/reportCollection.ts
@@ -33,7 +33,6 @@ export function startReportCollection(configuration: LogsConfiguration, lifeCycl
         error,
         status,
       },
-      messageContext: rawError.context,
     })
   })
 
--- a/packages/logs/src/domain/report/reportCollection.ts
+++ b/packages/logs/src/domain/report/reportCollection.ts
@@ -33,6 +33,7 @@ export function startReportCollection(configuration: LogsConfiguration, lifeCycl
         error,
         status,
       },
+      messageContext: rawError.context,
     })
   })
 
--- a/packages/rum-core/src/domain/error/errorCollection.ts
+++ b/packages/rum-core/src/domain/error/errorCollection.ts
@@ -49,7 +49,7 @@ export function startErrorCollection(
 
 export function doStartErrorCollection(lifeCycle: LifeCycle, pageStateHistory: PageStateHistory) {
   lifeCycle.subscribe(LifeCycleEventType.RAW_ERROR_COLLECTED, ({ error, customerContext, savedCommonContext }) => {
-    customerContext = combine(tryToGetErrorContext(error.originalError), customerContext) as Context
+    customerContext = combine(error.context, customerContext)
     lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, {
       customerContext,
       savedCommonContext,
@@ -114,11 +114,4 @@ function processError(error: RawError, pageStateHistory: PageStateHistory): RawR
     startTime: error.startClocks.relative,
     domainContext,
   }
-}
-
-function tryToGetErrorContext(originalError: unknown) {
-  if (originalError !== null && typeof originalError === 'object' && 'dd_context' in originalError) {
-    return originalError.dd_context as Context
-  }
-  return undefined
 }
--- a/packages/rum-core/src/domain/error/errorCollection.ts
+++ b/packages/rum-core/src/domain/error/errorCollection.ts
@@ -9,6 +9,7 @@ import {
   trackRuntimeError,
   NonErrorPrefix,
   isError,
+  combine,
 } from '@datadog/browser-core'
 import type { RumConfiguration } from '../configuration'
 import type { RawRumErrorEvent } from '../../rawRumEvent.types'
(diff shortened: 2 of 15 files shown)


packages/rum-core/src/domain/error/errorCollection.ts
```diff
@@ -97,6 +97,7 @@ function processError(error: RawError, pageStateHistory: PageStateHistory): RawR
       source_type: 'browser',
       fingerprint: error.fingerprint,
       csp: error.csp,
+      context: error.context,
```
amortemousque, line 100: **🔨 warning:** ‏The `context` should be at the root of the event. We query it with `@context.something` not `@error.context.something`. I think this context could be passed as `customerContext` like in the` addError(.., context) API`

RomanGaignault: the context here is a property of the error not the event itself, I am not sure to understand why it should be at the root ?

packages/rum-core/src/domain/error/errorCollection.ts
```diff
@@ -113,3 +115,10 @@ function processError(error: RawError, pageStateHistory: PageStateHistory): RawR
     domainContext,
   }
 }
+
+function tryToGetErrorContext(originalError: unknown) {
+  if (isError(originalError) && 'dd_context' in originalError) {
```
eloytoro, line 120: should we check that it's an `Error`? Currently we allow users to throw just about anything

to allow for more flexibility we really only need for this to be an object

https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEYD2A7AzgF3kmBLA5rilBAKIww4Bc8ArigNYpIDuKA3AFCfLpYAWUNABFhAYVQYQADywBeeAAo0MMDXpNWKAJQ0VYeLjTwA3vGDAA+rymz1jZm3gBfeHIB8neN-gYAngAOIEgAZvD6bnIKAORIAEYAVuAY0fAAZGnhqvAAhAootBAQ6ZnRFtaSMimGKFlgXNyQQsZitJhIALbklDDwVSAowMbdOKZePvggWOU2VYraYz5L8HAYtDC1ZgwgfjTRAG4ktCCpzlxLzpyXPKhoSBAgAHQQSPiK9KAhRCDAADTwghE4kqsnegxAXxQP202i4vDuD2er0UBSK-0BogkKFsGBRhQgMLht3uTxeb2imDwKHw0XRQkxINxFIwVJphJu6BJSPJBQ6cRAMFpAPpwOxcwADOz4VyyYpomA2hhOn0KDghRjRTiUSAWPBWu0uqqYPMpcTEbLYolkuqRVitWYZoyaCZXM5CUA

packages/rum-core/src/domain/error/errorCollection.ts
```diff
@@ -48,6 +49,7 @@ export function startErrorCollection(
 
 export function doStartErrorCollection(lifeCycle: LifeCycle, pageStateHistory: PageStateHistory) {
   lifeCycle.subscribe(LifeCycleEventType.RAW_ERROR_COLLECTED, ({ error, customerContext, savedCommonContext }) => {
+    customerContext = combine(tryToGetErrorContext(error.originalError), customerContext) as Context
```
BenoitZugmeyer, line 52: **💬 suggestion:** ‏extracting the context should be done in `computeRawError` (add a `RawError.context` property). This way, you'll be able to implement it in Logs as well.

BenoitZugmeyer: You could have something like this:
```suggestion
    customerContext = combine(error.context, customerContext) as Context
```

And within logs, something like:

```diff
diff --git packages/logs/src/domain/console/consoleCollection.ts packages/logs/src/domain/console/consoleCollection.ts
index 6aa982921..3eb2ccb25 100644
--- packages/logs/src/domain/console/consoleCollection.ts
+++ packages/logs/src/domain/console/consoleCollection.ts
@@ -31,6 +31,7 @@ export function startConsoleCollection(configuration: LogsConfiguration, lifeCyc
         error: log.error && createErrorFieldFromRawError(log.error),
         status: LogStatusForApi[log.api],
       },
+      messageContext: log.error?.context,
       domainContext: {
         handlingStack: log.handlingStack,
       },
diff --git packages/logs/src/domain/report/reportCollection.ts packages/logs/src/domain/report/reportCollection.ts
index 024b9df82..2002dd918 100644
--- packages/logs/src/domain/report/reportCollection.ts
+++ packages/logs/src/domain/report/reportCollection.ts
@@ -33,6 +33,7 @@ export function startReportCollection(configuration: LogsConfiguration, lifeCycl
         error,
         status,
       },
+      messageContext: rawError.context,
     })
   })
 
diff --git packages/logs/src/domain/runtimeError/runtimeErrorCollection.ts packages/logs/src/domain/runtimeError/runtimeErrorCollection.ts
index 65c04ede0..a6739c250 100644
--- packages/logs/src/domain/runtimeError/runtimeErrorCollection.ts
+++ packages/logs/src/domain/runtimeError/runtimeErrorCollection.ts
@@ -31,6 +31,7 @@ export function startRuntimeErrorCollection(configuration: LogsConfiguration, li
         origin: ErrorSource.SOURCE,
         status: StatusType.error,
       },
+      messageContext: rawError.context,
     })
   })
 
```

BenoitZugmeyer (approved):

packages/logs/src/domain/report/reportCollection.ts
```diff
@@ -33,6 +33,7 @@ export function startReportCollection(configuration: LogsConfiguration, lifeCycl
         error,
         status,
       },
+      messageContext: rawError.context,
```
amortemousque, line 36: **❓ question:** ‏How can a report have a context?

amortemousque (approved):

packages/logs/src/domain/report/reportCollection.ts
RomanGaignault: true, it's not possible. I will remove it

PR #3324, opened 2025-02-05
🐛 [RUM-8428] Make developer extension compatible with lazy loaded recorder module

sethfowler-datadog:
## Motivation

Lazy loading the recorder extension involves dynamically import()ing the module from a separate JS chunk, stored at a separate URL. The developer extension doesn’t know about this URL, so it doesn’t redirect it to the version served by yarn dev appropriately, breaking local development.

## Changes

This PR fixes the problem by making two changes:
1. Updating our webpack configuration to remove the context hash from the recorder chunk filename in local development, so that it gets hosted at a consistent `localhost` URL. (Nothing has changed in production builds.)
2. Updating the developer extension to redirect fetches of the recorder chunk to the appropriate `localhost` URL when local development is enabled.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/webpack.base.js
+++ b/webpack.base.js
@@ -13,7 +13,8 @@ module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins
     filename,
     chunkFilename:
       mode === 'development'
-        ? // Use a fixed name for each chunk during development.
+        ? // Use a fixed name for each chunk during development so that the developer extension
+          // can redirect requests for them reliably.
           `chunks/[name]-${filename}`
         : // Include a content hash in chunk names in production.
           `chunks/[name]-[contenthash]-${filename}`,


webpack.base.js
```diff
@@ -11,7 +11,12 @@ module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins
   mode,
   output: {
     filename,
-    chunkFilename: `chunks/[name]-[contenthash]-${filename}`,
+    chunkFilename:
+      mode === 'development'
+        ? // Use a fixed name for each chunk during development.
```
amortemousque, line 16: **🥜 nitpick:** Maybe elaborate on why do we need a fixed name in dev‏

amortemousque (approved):

RomanGaignault (approved):

PR #3325, opened 2025-02-05
✨ Allow to modify source_url property in Loaf script attributions

thomas-lebeau:
## Motivation

We want to be able to modify `source_url` and `invoker` properties of Loaf script attributions from the `beforeSend` function. This allow customer to scrub urls for potential PII.

## Changes

Refactor `limitModification` to allow arrays to be present in the path. (e.g. `long_task.scripts[].source_url` will allow to modify `source_url` in long tasks (LoaF) events

Allow modification on `long_task.scripts[].source_url` and `long_task.scripts[].invoker`

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/limitModification.ts
+++ b/packages/rum-core/src/domain/limitModification.ts
@@ -4,8 +4,9 @@ import type { Context } from '@datadog/browser-core'
 export type ModifiableFieldPaths = Record<string, 'string' | 'object'>
 
 /**
- * Current limitation:
- * - field path do not support array, 'a.b.c' only
+ * Allows declaring and enforcing modifications to specific fields of an object.
+ * Only supports modifying properties of an object (even if nested in an array).
+ * Does not support array manipulation (adding/removing items).
  */
 export function limitModification<T extends Context, Result>(
   object: T,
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/limitModification.ts
```diff
@@ -14,43 +15,85 @@ export function limitModification<T extends Context, Result>(
...
   const clone = deepClone(object)
   const result = modifier(clone)
+
+  // Validate the modified fields and assign them back to 'original' if they match the expected type.
   objectEntries(modifiableFieldPaths).forEach(([fieldPath, fieldType]) => {
-    const newValue = get(clone, fieldPath)
-    const newType = getType(newValue)
-    if (newType === fieldType) {
-      set(object, fieldPath, sanitize(newValue))
+    const pathSegments = fieldPath.split('.')
```
sethfowler-datadog, line 21: I love the approach! The one suggestion I have is to tweak the implementation strategy slightly. If you change this line to:
```suggestion
    const pathSegments = fieldPath.split(/\.|(?=\[\])/)
```
Then instead of `foo.bar[].baz` being split into `['foo', 'bar[]', 'baz']`, it will be split into `['foo', 'bar', '[]', 'baz']`. This should lead to a cleaner implementation of `getValueFromPath()`, where you can treat `[]` as a special field name that maps over all elements of an array. With the appropriate changes to `getValueFromPath()`, it will also mean that `foo.bar[][].baz` will work as expected with no special effort, so that we can support modifications of nested arrays.

packages/rum-core/src/domain/limitModification.ts
```diff
@@ -60,3 +97,27 @@ function isValidObject(object: unknown): object is Record<string, unknown> {
 function isValidObjectContaining(object: unknown, field: string): object is Record<string, unknown> {
   return isValidObject(object) && Object.prototype.hasOwnProperty.call(object, field)
 }
+
+function isValidType(actualType: string | string[], expectedType: 'string' | 'object'): boolean {
+  if (Array.isArray(actualType)) {
+    return actualType.length > 0 && actualType.every((type) => type === expectedType)
```
sethfowler-datadog, line 103: We don't validate here that the length matches; is that expected? I may be misreading the code, but it seems like this will allow people to remove items from the array while still passing the `isValidType()` check; the result looks like it would be that you'd get `undefined` for some elements in the output array. That may be OK, but I just wanted to make sure it was intentional. (Or that it's handled somehow and I'm just missing it.)

packages/rum-core/src/domain/limitModification.ts
```diff
@@ -14,49 +15,53 @@ export function limitModification<T extends Context, Result>(
...
+
+  if (field === '[]') {
+    if (!Array.isArray(object) || !Array.isArray(clone)) {
       return
     }
-    current = current[field]
+
+    object.forEach((item, i) => setValueAtPath(item, clone[i], restPathSegments, fieldType))
+    return
+  }
```
amortemousque, line 37: **🥜 nitpick:** ‏
```ts
if (field === '[]') {
  if (Array.isArray(object) && Array.isArray(clone)) {
    object.forEach((item, i) => setValueAtPath(item, clone[i], restPathSegments, fieldType));
  }
  return;
}
```

amortemousque, line 37: **🥜 nitpick:** ‏
```ts

if (field === '[]') {
  if (Array.isArray(object) && Array.isArray(clone)) {
    object.forEach((item, i) => setValueAtPath(item, clone[i], restPathSegments, fieldType));
  }
  return;
}
```

amortemousque (approved):

sethfowler-datadog (approved):
Looks good! I really like the final solution you came up with; very clean!

packages/rum-core/src/domain/limitModification.ts
sethfowler-datadog: Yeah, I really like the final result! Nicely done.

PR #3326, opened 2025-02-05
🐛 [RUM-8429] Report an error when lazy loading the recorder module fails

sethfowler-datadog:
## Motivation

Lazy loading the recorder module can fail, either due to a customer-side issue (like a CSP misconfiguration) or due to an issue on our end (like failing to upload the chunk to the CDN). We should report an error when this happens and add telemetry.

The current situation is a bummer since the failure is totally silent; no error is reported anywhere. I spent quite a bit of time over the past couple of days trying to understand why one of my local changes wasn't working. The root cause turned out to be #3324; the problem would've been instantly obvious with the changes in this PR.

## Changes

This PR adapts the current error reporting we do when the compression worker fails to load so that it can be used for both the compression worker and the lazy loaded recorder module. We now use the same code in both places, so our existing infrastructure for detecting CSP-related errors can be reused.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum/src/boot/lazyLoadRecorder.ts
+++ b/packages/rum/src/boot/lazyLoadRecorder.ts
@@ -1,8 +1,23 @@
-export async function lazyLoadRecorder() {
-  try {
+import { reportScriptLoadingError } from '../domain/scriptLoadingError'
+import type { startRecording } from './startRecording'
+
+export function lazyLoadRecorder() {
+  return lazyLoadRecorderFrom(async () => {
     const module = await import(/* webpackChunkName: "recorder" */ './startRecording')
     return module.startRecording
-  } catch {
-    /* Prevent collecting the webpack ChunkLoadError as it is already collected as a RUM resource. */
+  })
+}
+
+export async function lazyLoadRecorderFrom(
+  importer: () => Promise<typeof startRecording>
+): Promise<typeof startRecording | undefined> {
+  try {
+    return await importer()
+  } catch (error: unknown) {
+    reportScriptLoadingError({
+      error,
+      source: 'Recorder',
+      scriptType: 'module',
+    })
   }
 }
--- a/packages/rum/src/boot/lazyLoadRecorder.ts
+++ b/packages/rum/src/boot/lazyLoadRecorder.ts
@@ -1,18 +1,11 @@
 import { reportScriptLoadingError } from '../domain/scriptLoadingError'
 import type { startRecording } from './startRecording'
 
-export function lazyLoadRecorder() {
-  return lazyLoadRecorderFrom(async () => {
-    const module = await import(/* webpackChunkName: "recorder" */ './startRecording')
-    return module.startRecording
-  })
-}
-
-export async function lazyLoadRecorderFrom(
-  importer: () => Promise<typeof startRecording>
+export async function lazyLoadRecorder(
+  importRecorderImpl = importRecorder
 ): Promise<typeof startRecording | undefined> {
   try {
-    return await importer()
+    return await importRecorderImpl()
   } catch (error: unknown) {
     reportScriptLoadingError({
       error,
@@ -20,4 +13,9 @@ export async function lazyLoadRecorderFrom(
       scriptType: 'module',
     })
   }
+}
+
+async function importRecorder() {
+  const module = await import(/* webpackChunkName: "recorder" */ './startRecording')
+  return module.startRecording
 }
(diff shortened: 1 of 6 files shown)


packages/rum/src/boot/lazyLoadRecorder.ts
```diff
@@ -1,8 +1,23 @@
...
+export function lazyLoadRecorder() {
+  return lazyLoadRecorderFrom(async () => {
     const module = await import(/* webpackChunkName: "recorder" */ './startRecording')
     return module.startRecording
-  } catch {
-    /* Prevent collecting the webpack ChunkLoadError as it is already collected as a RUM resource. */
+  })
+}
+
+export async function lazyLoadRecorderFrom(
```
sethfowler-datadog, line 11: This helper is here for testing purposes, so that it's possible to replace the real dynamic import with a version that throws a specific error.

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -261,6 +274,43 @@ describe('makeRecorderApi', () => {
       })
     })
 
+    describe('while lazy loading the recorder module', () => {
+      it('should report an error but no telemetry if CSP blocks the module', async () => {
```
sethfowler-datadog, line 278: Note that the behaviors we're testing for here are the same ones we already have for the compression worker; the lazy loaded recorder module will now report errors in the same way.

packages/rum/src/domain/scriptLoadingError.ts
```diff
@@ -0,0 +1,37 @@
...
+export function reportScriptLoadingError({
+  configuredUrl,
+  error,
+  source,
+  scriptType,
+}: {
+  configuredUrl?: string | undefined
+  error: unknown
+  source: string
+  scriptType: 'module' | 'Worker'
```
sethfowler-datadog, line 12: The casing inconsistency here is a bit funky. I wanted to avoid changing the capitalization of "Worker" in case it would affect our telemetry metrics, but "Module" doesn't really feel right, so I ended up with this compromise.

cy-moi: 🤔 I don't think capitalization would affect our telemetry metrics tbh. Same for the verb. I'm fine if we just use a same verb everyone.

thomas-lebeau (approved):

amortemousque (approved):

packages/rum/src/boot/lazyLoadRecorder.ts
BenoitZugmeyer: Maybe something like this would be simpler to get:

```ts
export async function lazyLoadRecorder(importRecorderImpl = importRecorder) {
  try {
    return await importRecorderImpl()
  } catch {
   reportScriptLoadingError({
      error,
      source: 'Recorder',
      scriptType: 'module',
    })
  }
}

async function importRecorder() {
   const module = await import(/* webpackChunkName: "recorder" */ './startRecording')
   return module.startRecording
}
```

packages/rum/src/boot/recorderApi.spec.ts
BenoitZugmeyer: **💬 suggestion:** ‏ those two tests should be moved to `scriptLoadingError.spec.ts`.

packages/rum/src/domain/scriptLoadingError.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (configuredUrl) {
+      baseMessage = `Please make sure the ${scriptType} URL ${configuredUrl} is correct and CSP is correctly configured.`
+    } else {
+      baseMessage = 'Please make sure CSP is correctly configured.'
+    }
+    display.error(
+      `${baseMessage} See documentation at ${DOCS_ORIGIN}/integrations/content_security_policy_logs/#use-csp-with-real-user-monitoring-and-session-replay`
+    )
+  } else {
+    addTelemetryError(error)
```
BenoitZugmeyer, line 27: **💬 suggestion:** ‏I don't think we should report the error as telemetry, as there is high chances that we won't be able to do anything about it. Ex: the recorder fails to load because of a user connectivity issue.

packages/rum/src/domain/scriptLoadingError.ts
sethfowler-datadog: Done!

packages/rum/src/boot/recorderApi.spec.ts
sethfowler-datadog: Done!

packages/rum/src/boot/lazyLoadRecorder.ts
sethfowler-datadog: Sounds good; I made the switch!

PR #3331, opened 2025-02-07
🎨 Use a more specific chunk name for recorder

romulof:
## Motivation

When RUM SDK async loads recorder, it instructs webpack to use the name `recorder` for the chunk. Declarin a name is nice, as it keep awareness of the purpose of the file, but the name `recorder` might conflict with other people's chunk names.

## Changes

Simply rename recorder chunk name to `datadog-recorder`, making clear its origin and purpose, while avoiding possible conflicts.

## Testing

Build and check the chunk name in build output directory.

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


bcaudan:
Hi @romulof, thanks for your suggestion, it would indeed be better.

As it would be a breaking change, we will track it internally and consider it for the next major release (no ETA for now).

PR #3335, opened 2025-02-12
👷 [RUM-259] update test/app to use local tarballs

thomas-lebeau:
## Motivation

With this PR, all e2e tests now run all npm setup using the packed version of the sdk packages, this will help catch cases were we deploy broken packages to npm.

## Changes

- add `pack` command to all packages' `package.json`
- pack all packages before building `test/app`
- use packed version of sdk in test/app dependencies

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [x] End to end

---

### DIFF
--- a/scripts/check-typescript-compatibility.js
+++ b/scripts/check-typescript-compatibility.js
@@ -10,9 +10,10 @@ const TEST_APP_DIR = path.join(__dirname, '..', 'test', 'app')
 runMain(() => {
   printLog('Building project...')
   command`yarn build`.run()
+  command`lerna run pack --stream`.run()
 
   printLog('Setting up test environment...')
-  command`yarn`.withCurrentWorkingDirectory(TEST_APP_DIR).run()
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "postinstall": "scripts/cli init_submodule",
     "build": "lerna run build --stream",
     "build:bundle": "lerna run build:bundle --stream",
-    "build:app": "cd test/app && rm -rf node_modules && yarn && yarn build",
+    "build:app": "lerna run pack --stream && cd test/app && rm -rf node_modules && yarn install --no-immutable && yarn build",
```
thomas-lebeau, line 14: `--no-immutable` is needed as we're packing the dependencies used in `test/app` in the CI, their checksum is likely to changes between build.

sethfowler-datadog (approved):
Nice! This is a great addition. This'll give us a lot more confidence that our build system is doing the right thing; really timely since there's a lot of interest in making improvements in that area lately.

RomanGaignault (commented):
Nice addition, this would be definitely closer to real user usage

PR #3340, opened 2025-02-13
[Experimental][RUM Profiler] Integrate Profiler into SDK

thomasbertet:
## Motivation

Integrate the [RUM Profiler](https://github.com/DataDog/rum-profiler-poc) into the Browser SDK package. 

## Changes

- new `profilingSampleRate` configuration to enable the RUM Profiler via the Browser SDK.
- new `profilingEndpointBuilder` to create endpoint for profile ingestion.
- new `profilerApi` (similar to `recorderApi`) where the core Profiling logic is lazy loaded on startup (when active)

## How to try it

This feature requires an experimental feature (`profiling`) as well as specifying a `profilingSampleRate` greater than 0. 

Note: 

- Same limitations applies for this version (eg. required Response Header, CORS policy, etc.), compared to previous `rum-profiler-poc` version.
- Same logic for collection: Collect profiles every 1 minutes, and on unload/hide.
- Collecting profiles is not part of the batching mecanism for now. As it's still experimental, we don't want to mess with batch logic just yet. We want to measure internally our usage and data volume before potentially integrating it.

## Testing
- Added a unit test around profiling collection. 
- [x] Local 
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum/src/domain/profiling/transport/transport.ts
+++ b/packages/rum/src/domain/profiling/transport/transport.ts
@@ -6,7 +6,7 @@ interface ProfileEventAttributes {
   application: { id: string }
   session?: { id: string }
   view?: { ids: string[] }
-  context?: { long_task: { id: string[] } }
+  long_task?: { id: string[] }
 }
 interface ProfileEvent extends ProfileEventAttributes {
   attachments: string[]
@@ -136,9 +136,7 @@ function buildProfileEventAttributes(
     .filter((id) => id !== undefined)
 
   if (longTaskIds.length) {
-    attributes.context = {
-      long_task: { id: longTaskIds },
-    }
+    attributes.long_task = { id: longTaskIds }
   }
   return attributes
 }


packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -31,6 +31,10 @@ export function createEndpointBuilder(
       const parameters = buildEndpointParameters(initConfiguration, trackType, configurationTags, api, payload)
       return buildUrlWithParameters(parameters)
     },
+    rawParameters: {
+      initConfiguration,
+      configurationTags,
+    },
```
thomasbertet, line 37: This is controversial and will probably won't make it. 
@aymericDD, happy to discuss other ways to get access to these parameters at the "send" level.
See `exportToJSONIntake` function to see how its used.

packages/rum-core/src/domain/profiling/exporter/emptyPprof.ts
```diff
@@ -0,0 +1,28 @@
...
+    const binary = atob(base64);
+    const len = binary.length;
+    const bytes = new Uint8Array(len);
+
+    for (let i = 0; i < len; i++) {
+        bytes[i] = binary.charCodeAt(i);
+    }
+
+    return new Blob([bytes], { type: 'application/octet-stream' });
+}
```
thomasbertet, line 28: This whole file will probably be removed. It's used temporarily while the Profiling BE is not handling natively the JSON.

packages/rum-core/src/domain/profiling/exporter/exportToJsonIntake.ts
```diff
@@ -0,0 +1,165 @@
...
+    sessionId,
+) => {
+    const event = createProfileEvent(profilerTrace, endpointBuilder, applicationId, sessionId);
+    const payload = createProfilePayload(profilerTrace, event)
+
+    // Sends profile to public profiling intake.
+    const xhrUrl = endpointBuilder.build('xhr', payload)
+
+    // Send everything.
+    return new Promise((resolve) =>  sendXHR(xhrUrl, payload.data, resolve));
```
thomasbertet, line 42: Controversial: using `sendXHR` here seems prohibited (eslint issue, something about side-effects). 
Happy to discuss alternatives.

packages/rum-core/src/domain/profiling/exporter/exportToJsonIntake.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+function createProfileEvent(
+    profilerTrace: RumProfilerTrace, 
+    endpointBuilder: EndpointBuilder, 
+    applicationId: string ,
+    sessionId: string | undefined
+): ProfileEvent {
+    const rawParameters = endpointBuilder.rawParameters;
```
thomasbertet, line 51: As commented previously, this is probably not the right way to get these parameters. It feels off.

packages/rum-core/src/domain/profiling/exporter/exportToJsonIntake.ts
```diff
@@ -0,0 +1,165 @@
...
+    const profileEventTags = configurationTags.concat([
+        `service:${initConfiguration.service}`,
+        `version:${initConfiguration.version}`,
+        `env:${initConfiguration.env || 'unknown'}`,
+        'language:javascript',
+        'runtime:chrome',
+        'family:chrome',
+        'format:json',
+        // TODO: replace with RUM device id in the future
+        `host:${normalizeTag(navigator.userAgent).slice(0, 200)}`,
```
thomasbertet, line 81: Still unsure how to get the host as well here. 
These tags will be sent alongside the Profile, so it's indeed a duplicate from tags set by RUM transport layer, but are useful for Profiling ingestion.

packages/rum-core/src/domain/profiling/exporter/exportToJsonIntake.ts
```diff
@@ -0,0 +1,165 @@
...
+    ]);
+
+    // TODO deobfuscation is not supported yet
+    // if (config.commitHash) {
+    //     eventTags.push(`git.commit.sha:${config.commitHash}`);
+    // }
+    // if (config.repositoryUrl) {
+    //     eventTags.push(`git.repository_url:${config.repositoryUrl}`);
+    // }
```
thomasbertet, line 90: This is something we need to figure out, where to do this de-obfuscation. This code will be removed before merging obviously.

packages/rum-core/src/domain/profiling/lazyLoadProfiler.ts
```diff
@@ -0,0 +1,8 @@
+export async function lazyLoadProfiler() {
```
Miz85, line 1: Does this mean this only work in when the browser SDK is installed through npm?

thomasbertet: To my understanding, I think this lazy loading code will run whatever the installation process, but I can be wrong of course :) I'm no expert. From what I know, that's also how the Session Replay recorder is loaded.

thomas-lebeau: > To my understanding, I think this lazy loading code will run whatever the installation process

You're correct, this works with the CDN and with NPM given it's using a bundler.

Miz85: Perfect! Thank you!

packages/rum-core/src/domain/profiling/profiler.ts
```diff
@@ -0,0 +1,250 @@
...
+      viewId: viewId || '',
+    })
+  }
+
+  function handleProfilerTrace(trace: RumProfilerTrace): void {
+    // Find current session to assign it to the Profile.
+    const sessionId = session.findTrackedSession()?.id
+
+    // Send JSON Profile to intake.
+    exportToJSONIntake(trace, endpointBuilder, applicationId, sessionId, configuration.site).catch(() => undefined)
```
amortemousque, line 188: **💬 suggestion:** Why not passing the entire configuration like so:
‏
```suggestion
    exportToJSONIntake(trace, sessionId, configuration).catch(() => undefined)
```

packages/rum-core/src/domain/profiling/profiler.ts
```diff
@@ -0,0 +1,250 @@
...
+  function start(viewId: string | undefined): void {
+    if (instance.state === 'running') {
+      return
+    }
+    applicationId = configuration.applicationId
+
+    // Setup event listeners
+    observer = new PerformanceObserver(handlePerformance)
+    observer.observe({
+      entryTypes: [getLongTaskEntryType(), 'measure', 'event'],
```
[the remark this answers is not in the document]
thomasbertet: As discussed,  I'll remove them.

packages/rum-core/src/domain/profiling/profilingCollection.ts
```diff
@@ -0,0 +1,58 @@
...
+  lazyLoadProfiler()
+    .then((createRumProfiler) => {
+      if (!createRumProfiler) {
+        return
+      }
+
+      profiler = createRumProfiler({
+        configuration,
+        endpointBuilder,
+        isLongAnimationFrameEnabled,
```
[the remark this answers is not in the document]
thomasbertet: Discussed: Deactivate long tasks correlations and collection when SDK Long Tasks is disabled.

packages/core/src/domain/configuration/endpointBuilder.ts
thomasbertet: Discussed:  remove initConfig, set `tags` directly.

packages/rum-core/src/domain/profiling/exporter/exportToJsonIntake.ts
```diff
@@ -0,0 +1,139 @@
...
+  site
+) => {
+  const event = createProfileEvent(profilerTrace, endpointBuilder, applicationId, sessionId, site)
+  const payload = createProfilePayload(profilerTrace, event)
+
+  // Sends profile to public profiling intake.
+  const xhrUrl = endpointBuilder.build('xhr', payload)
+
+  // Send everything.
+  return new Promise<HttpResponse>((resolve) => sendXHR(xhrUrl, payload.data, resolve))
```
thomasbertet, line 37: discussed: use `fetch()` directly.

packages/rum-core/src/domain/profiling/profiler.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    // Don't wait for data collection to start next instance
+    collectProfilerInstance()
+
+    const profiler = new globalThisProfiler({
+      sampleInterval: SAMPLE_INTERVAL_MS,
+      // Keep buffer size at 1.5 times of minimum required to collect data for a profiling instance
+      maxBufferSize: Math.round((COLLECT_INTERVAL_MS * 1.5) / SAMPLE_INTERVAL_MS),
+    })
```
thomasbertet, line 96: try/catch with a human message around adding Header.

packages/rum-core/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -0,0 +1,37 @@
+import { createLruCache } from './lruCache/lruCache'
+
+type PerformanceEntryStartTimeMs = number
+// Maps PerformanceEntry start-time to the corresponding long task ID
+// We need this to link RUM Long Tasks with RUM Profiler stack traces
+// Given that long task takes at least 50ms and we export profile at least every 60 seconds, we can have up to 1200 entries (60s / 50ms = 1200).
+// eslint-disable-next-line local-rules/disallow-side-effects
+const registry = createLruCache<PerformanceEntryStartTimeMs, string>(1200)
```
thomasbertet, line 8: get rid of LRU cache, use plain Map. It's reset every profile anyway.

packages/rum-core/src/domain/profiling/profiler.ts
thomasbertet: I'd rather keep the signature small for simpler test.

packages/rum-core/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function getLongTaskId(longTaskEntry: PerformanceEntry): string | undefined {
+  // Don't return if it's not enabled or the long task has been reported before the activation
+  if (enabledTime === false || longTaskEntry.startTime < enabledTime) {
+    return undefined
+  }
+
+  const id = registry.get(longTaskEntry.startTime)
+
+  return id
```
piotr-oles, line 35: opt nit:
```suggestion
  return registry.get(longTaskEntry.startTime)
```
I don't think adding `id` variable helps here with readability

packages/rum-core/src/domain/profiling/profilingCollection.ts
```diff
@@ -0,0 +1,72 @@
...
+    // User is not lucky, no profiling!
+    return NOOP_COLLECTOR
+  }
+
+  let profiler: RUMProfiler
+
+  lazyLoadProfiler()
+    .then((createRumProfiler) => {
+      if (!createRumProfiler) {
+        throw new Error('[DD_RUM] Failed to lazy load the RUM Profiler') // will be caught by the catch block below.
```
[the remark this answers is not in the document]
thomasbertet: I see what's the idea, but in a case of a Promise I find it's a valid thing to throw to get to the "catch", and it avoids us repeating ourselves. I don't really see this as control-flow, as it's part of the design of Promise.

packages/rum-core/src/domain/profiling/transport/transport.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+  const longTaskIds: string[] = profilerTrace.longTasks
+    .map((longTask) => getLongTaskId(longTask))
+    .filter((id) => id !== undefined)
+
+  if (longTaskIds.length) {
+    attributes.context = {
+      profile_long_task_id: longTaskIds,
+    }
```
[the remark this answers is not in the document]
thomasbertet: Good idea indeed! I'll update :)

packages/rum-core/src/rumEvent.types.ts
```diff
@@ -987,6 +987,20 @@ export type RumViewEvent = CommonProperties &
...
+        /**
+         * Duration in ns between start of the view and the start of the slow frame
+         */
+        readonly start: number
+        /**
+         * Duration in ns of the slow frame
+         */
+        readonly duration: number
+        [k: string]: unknown
+      }[]
```
thomasbertet, line 1003: @amortemousque not sure what I should do about this ? It was synced when I did `yarn rum-events-format:sync` IIRC.

BenoitZugmeyer: This is fine you can leave it.

packages/rum/src/domain/profiling/profilingCollection.ts
```diff
@@ -0,0 +1,70 @@
...
+      const isLongAnimationFrameEnabled = supportPerformanceTimingEvent(RumPerformanceEntryType.LONG_ANIMATION_FRAME)
+
+      profiler = createRumProfiler({
+        configuration,
+        isLongAnimationFrameEnabled,
+        lifeCycle,
+        session,
+        profilerConfiguration: customProfilerConfiguration,
+      })
+
```
BenoitZugmeyer, line 58: **💬 suggestion:** ‏maybe use separated arguments instead of an object hash to reduce the main bundle size impact

packages/rum/src/domain/profiling/profilingCollection.spec.ts
BenoitZugmeyer, line 1: **💬 suggestion:** this test could be adjusted to test `profiler.ts`, not `profilingCollection.ts`, as there is nothing specific to `profilingCollection` in there. It would simplify `profilingCollection.ts` and reduce the core bundle size impact.

packages/rum/src/boot/profilerApi.ts
```diff
@@ -0,0 +1,65 @@
...
+      // User is not lucky, no profiling!
+      return noop
+    }
+
+    let profiler: RUMProfiler
+
+    lazyLoadProfiler()
+      .then((createRumProfiler) => {
+        if (!createRumProfiler) {
+          addTelemetryDebug('[DD_RUM] Failed to lazy load the RUM Profiler')
```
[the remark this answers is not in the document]
thomasbertet: It saved me already a few times, while testing locally, I think it does not harm either, so I'm in favor of keeping it around yes.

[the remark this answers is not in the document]
BenoitZugmeyer: You have the error here locally: https://github.com/DataDog/browser-sdk/pull/3340/files#diff-5391b04892beb8306f93aba36a0ce8b3d85476e3b0a15a34d636b6bd72c7e6ecR9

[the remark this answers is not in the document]
thomasbertet: Oh right! Forgot about this one. I guess it's fine to keep the telemetry around though so that we can have rought stats of how often the chunk is failing ?

packages/rum/src/domain/profiling/transport/transport.ts
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Send RUM profile as JSON to public profiling intake.
+ */
+const sendProfile: SendProfileFunction = (profilerTrace, endpointBuilder, applicationId, sessionId) => {
+  const event = buildProfileEvent(profilerTrace, endpointBuilder, applicationId, sessionId)
+  const payload = buildProfilingPayload(profilerTrace, event)
+
+  // Create URL, public profiling intake.
+  const profilingIntakeURL = endpointBuilder.build('xhr', payload)
```
thomas-lebeau, line 34:
```suggestion
  const profilingIntakeURL = endpointBuilder.build('fetch', payload)
```

amortemousque (approved):

test/e2e/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "esModuleInterop": true,
     "moduleResolution": "node",
     "resolveJsonModule": true,
-    "target": "ES2015",
+    "module": "commonjs",
+    "target": "ES2020",
```
thomas-lebeau, line 11: **❓ question:** ‏ Just out of curiosity, why is this changed needed? It does not seems you're changing any e2e test, which is what this config applies to. 

(I'm not asking to change, even if it's not needed, it probably doesn't hurt 😄)

thomasbertet: Le lazy loading part was failing in e2e tests, as it did not know the dynamic `import()` syntax. I has to bump a little the target there to make it work like a charm. I tried this though when the "Profiler" lazy-loading was part of the `rum-core`, now that it's only in `rum` it may not be necessary, I'm not sure.

PR #3186, opened 2024-11-28
 ✨ [RUM-6868] implement consistent probabilistic trace sampling 

BenoitZugmeyer:
## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

I have gone over the [contributing](https://github.com/DataDog/browser-sdk/blob/main/CONTRIBUTING.md) documentation.

PR #3352, opened 2025-02-18
 ⚗️✨ [RUM-8622] session consistent trace sampling 

BenoitZugmeyer:
## Motivation

Make sure all requests are sampled for a given session. This replaces #3186 where trace sampling was based on the trace id.

## Changes

* Do some cleanup, as #3186 is droped
* Implement session-consistent trace sampling

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/tracing/sampler.ts
+++ b/packages/rum-core/src/domain/tracing/sampler.ts
@@ -1,7 +1,8 @@
-import { performDraw } from '@datadog/browser-core'
-import type { TraceIdentifier } from './identifier'
+import { ExperimentalFeature, isExperimentalFeatureEnabled, performDraw } from '@datadog/browser-core'
 
-export function isTraceSampled(identifier: TraceIdentifier, sampleRate: number) {
+let sampleDecisionCache: { sessionId: string; sampleRate: number; decision: boolean } | undefined
+
+export function isTraceSampled(sessionId: string, sampleRate: number) {
   // Shortcuts for common cases. This is not strictly necessary, but it makes the code faster for
   // customers willing to ingest all traces.
   if (sampleRate === 100) {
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/domain/tracing/sampler.ts
```diff
@@ -12,13 +13,45 @@ export function isTraceSampled(identifier: TraceIdentifier, sampleRate: number)
...
+  if (!isExperimentalFeatureEnabled(ExperimentalFeature.CONSISTENT_TRACE_SAMPLING)) {
     return performDraw(sampleRate)
   }
 
-  // Offer consistent sampling for the same trace id across different environments. The rule is:
+  if (
+    sampleDecisionCache &&
+    sessionId === sampleDecisionCache.sessionId &&
+    sampleRate === sampleDecisionCache.sampleRate
+  ) {
```
thomas-lebeau, line 24: **🥜 nitpick:** ‏What about making `sampleDecisionCache` a `Map<sessionId, decision>`? That would simplify this condition:
```suggestion
  if (sampleDecisionCache.get(sessionId)) {
```

thomas-lebeau (approved):

cy-moi (approved):

packages/rum-core/src/domain/tracing/sampler.ts
BenoitZugmeyer: I thought about it at first, but then there is the whole question of limiting the cache to avoid leaking memory (even if it's just a bit).

I simplified a bit the cache as in practice the sample rate won't change during the page lifetime.

thomas-lebeau (approved):

xgouchet (approved):

PR #3359, opened 2025-02-20
Define an error type and message for handling stacks

eloytoro:
## Motivation

In Error Tracking we'd like to have the ability to default to the `handling_stack` when the provided error has no stack trace, or it's missing some attributes that are required, for it to be fingerprinted

Today the handling stack looks like this
```
Error:
  at <anonymous> @ https://www.paypalobjects.com/merchant-hf/latest/hf.js:23993:13
```

After this change it's going to look like this
```diff
-Error: 
+HandlingStack: error
   at <anonymous> @ https://www.paypalobjects.com/merchant-hf/latest/hf.js:23993:13
```

Take into consideration the fact that handling stacks are produced for not only errors but for actions and other objects

## Changes

- Changed the signature of `createHandlingStack` so it now received a `type` which determines which type of object it's reporting and it's added to the error message
- Updated references and tests

## Testing

Tests were defined


packages/core/src/tools/stackTrace/handlingStack.ts
```diff
@@ -8,15 +8,18 @@ import { computeStackTrace } from './computeStackTrace'
...
   /**
    * Skip the two internal frames:
    * - SDK API (console.error, ...)
    * - this function
    * in order to keep only the user calls
    */
   const internalFramesToSkip = 2
-  const error = new Error()
+  const error = new Error(`Received ${type}`)
+  error.name = 'SDKHandling'
```
BenoitZugmeyer, line 22: **💬 suggestion:** ‏the wording seems a bit strange to me. Suggestion:

```suggestion
  const error = new Error(type)
  error.name = 'HandlingStack'
```

packages/core/src/tools/stackTrace/handlingStack.spec.ts
```diff
@@ -15,6 +15,6 @@ describe('createHandlingStack', () => {
   it('should create handling stack trace without internal calls', () => {
     userCallOne()
 
-    expect(handlingStack).toMatch(/^Error: \n\s+at userCallTwo @ (.*)\n\s+at userCallOne @ (.*)/)
+    expect(handlingStack).toMatch(/^HandlingStack: error\n\s+at userCallTwo @ (.*)\n\s+at userCallOne @ (.*)/)
```
thomas-lebeau, line 18: **💭 thought:** ‏ what about `Handled error: at ...` so it read more naturally. by extention we would have `Handled action: at ...`, `Handled console error: at ...`, etc.

eloytoro: That would be a bit odd since in stack traces the error type precedes the message (separated by a colon) and it's never the case that the type of the error to contain spaces in it (take for example `TypeError`, `ReferenceError`, etc.

thomas-lebeau: hmm, you're right the style don't match. If we want to be consistent, then what about `ConsoleErrorHandlingStack:`, `ActionHandlingStack`, `ConsoleError`, `ReactError`, ...

thomas-lebeau (approved):

packages/core/src/tools/stackTrace/handlingStack.spec.ts
eloytoro: we could do that but then we'd have to write something in the message too. But also the handling stack isn't really an error, so i think the name should not imply that it's an error. How about something like this
```
${Type}HandlingStack: ${type} object detected by the SDK
  at ...
```
Having different error types might be benefitial because it makes it so our fingerprinting algorithm makes a clear distinction between them

PR #3365, opened 2025-02-21
⚡️[RUM-8728] Eliminate the fixed ReplayStats history entry limit

sethfowler-datadog:
## Motivation

We currently have a fixed limit for the number of `ReplayStats` entries we retain.

Prior to #3318, the limit was 10 entries; this caused us to drop views from the `ReplayStats` entries list before they were completely flushed to the intake on pages which generate view changes extremely quickly. Because these views had no `ReplayStats`, they wrongly got marked as having no session replay data, which ended up breaking playback.

After #3318, the limit has been raised to a very large number, eliminating the original problem. However, we now potentially keep many thousands of `ReplayStats` entries around unnecessarily, wasting resources on end-user machines.

## Changes

This PR fixes this problem in a cleaner way by removing the fixed `ReplayStats` history entry limit altogether. Instead, `ReplayStats` history entries are removed when the view they correspond to is destroyed (and only then!).

The natural way to handle this is to have the code that manages the `ReplayStats` history listen for a lifecycle event. Unfortunately, the existing `VIEW_ENDED`/`AFTER_VIEW_ENDED` lifecycle events are insufficient for this purpose. These events indicate the _conceptual_ end of the view, but we still accept trailing events for a view for a certain period of time after the view ends, so removing the `ReplayStats` history for a view in response to these events would remove it too early. Instead of that approach, I've added a new `VIEW_DESTROYED` lifecycle event that's fired at the correct time.

As part of these changes, I've updated the comments in `lifeCycle.ts` to clarify the meaning of the lifecycle events a bit. I particularly wanted to make sure that the behavior I described above with `VIEW_ENDED` and `AFTER_VIEW_ENDED` was documented, since it was surprising to me, but I added comments for all of the view-related lifecycle events for completeness.

This PR is overall a bit larger than you might expect because the requirement that the `ReplayStats` history code have access to a `LifeCycle` object means that it can no longer be a singleton; there's now a `ReplayStatsHistory` object that needs to be passed around to the code that needs it. That's a good thing, I think; the fewer globals, the better.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum/src/domain/record/record.ts
```diff
@@ -47,7 +47,7 @@ export function record(options: RecordOptions): RecordAPI {
     emit(record)
     sendToExtension('record', { record })
     const view = options.viewHistory.findView()!
-    replayStats.addRecord(view.id)
+    options.replayStatsHistory.addRecord(view.id)
```
amortemousque, line 50: **🥜 nitpick:** For consistency, use object destructuring for `replayStatsHistory`.
```ts 
const { emit, configuration, lifeCycle, replayStatsHistory } = options
```

packages/rum/src/boot/recorderApi.ts
```diff
@@ -76,7 +79,9 @@ export function makeRecorderApi(
       getDeflateWorkerStatus() === DeflateWorkerStatus.Initialized && strategy.isRecording(),
 
     getReplayStats: (viewId) =>
-      getDeflateWorkerStatus() === DeflateWorkerStatus.Initialized ? getReplayStatsImpl(viewId) : undefined,
+      getDeflateWorkerStatus() === DeflateWorkerStatus.Initialized
+        ? replayStatsHistory?.getReplayStats(viewId)
+        : undefined,
```
amortemousque, line 84: **💬 suggestion:** ‏To be consistent with other APIs, we could have the `getReplayStats` has a strategy?
```ts
getReplayStats: (viewId) => strategy.getReplayStats(viewId)
```
Also  `startReplayStatsHistory` could be called in `startRecording` instead of `recorderAPI` and then in `postStartStrategy`: 

```ts
;({ stop: stopRecording, replayStatsHistory  } = startRecordingImpl(
  lifeCycle,
  configuration,
  sessionManager,
  viewHistory,
  deflateEncoder
))
```

packages/rum/src/domain/record/record.ts
sethfowler-datadog: Sure, I can make that change. I'll update the invocation of `viewHistory` to follow that pattern as well.

packages/rum/src/boot/recorderApi.ts
sethfowler-datadog: I'll do my best!

PR #3377, opened 2025-02-24
CLS tracking: change source attribution logic and include `devicePixelRatio` inside `performance.cls` namespace

anediaz:
## Motivation
Implementation of small adjustments in Browser SDK for CLS attribution:
- Include `devicePixelRatio` attribute: 
   - Recent discoveries showed that Layout Instability API does not report viewPort coordinates for `prevRect` and `currentRect`.
   - We'd need to include this data in the CLS attribution object. See [this document](https://datadoghq.atlassian.net/wiki/spaces/~614b8bf2071141006ad0cebd/pages/4757587959/Bug+discovered+on+Google+Layout+Instability+API) for more context

- Adapt fix “most shifted“ element calculation: 
   - The current calculation does not take into account the sources size attached to a CLS entry, it returns always the first one
   - It’d be more relevant returning the biggest element (among those having the same shift), since it could be more impactful than a smaller element with the same shift
   - Change inspired on [this snippet](https://github.com/GoogleChromeLabs/web-vitals-report/blob/71b0879334798c732f460945ded5267cab5a36bf/src/js/analytics.js#L104)

## Changes
- Include new attribute `devicePixelRatio` in the CLS attribution
- Adapt cls tracking logic to pick the source having the biggest area among those attributed to the CLS entry

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
@@ -79,6 +79,7 @@ describe('trackCumulativeLayoutShift', () => {
       targetSelector: undefined,
       previousRect: undefined,
       currentRect: undefined,
+      devicePixelRatio: jasmine.any(Number),
     })
   })
 
@@ -143,6 +144,7 @@ describe('trackCumulativeLayoutShift', () => {
       targetSelector: undefined,
       previousRect: undefined,
       currentRect: undefined,
+      devicePixelRatio: jasmine.any(Number),
     })
   })
 
@@ -164,6 +166,7 @@ describe('trackCumulativeLayoutShift', () => {
       targetSelector: undefined,
       previousRect: undefined,
       currentRect: undefined,
+      devicePixelRatio: jasmine.any(Number),
     })
   })
 
@@ -218,6 +221,7 @@ describe('trackCumulativeLayoutShift', () => {
       targetSelector: undefined,
       previousRect: undefined,
       currentRect: undefined,
+      devicePixelRatio: jasmine.any(Number),
     })
   })
 
@@ -355,6 +359,7 @@ describe('trackCumulativeLayoutShift', () => {
         targetSelector: '#span-element',
         previousRect: { x: 0, y: 0, width: 40, height: 40 },
         currentRect: { x: 50, y: 50, width: 40, height: 40 },
+        devicePixelRatio: jasmine.any(Number),
       })
     })
   })
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
@@ -310,9 +310,8 @@ describe('trackCumulativeLayoutShift', () => {
     it('should get the target element, time, and rects of the largest layout shift', () => {
       startCLSTracking()
       const divElement = appendElement('<div id="div-element"></div>')
-      const spanElement = appendElement('<span id="span-element"></span>')
+      const largerElement = appendElement('<span id="larger-element"></span>')
 
-      // first session window:  { value: 0.5, time: 1, targetSelector: '#div-element' }
       notifyPerformanceEntries([
         createPerformanceEntry(RumPerformanceEntryType.LAYOUT_SHIFT, { value: 0.1, startTime: 0 as RelativeTime }),
       ])
@@ -327,7 +326,7 @@ describe('trackCumulativeLayoutShift', () => {
               currentRect: DOMRectReadOnly.fromRect({ x: 50, y: 50, width: 10, height: 10 }),
             },
             {
-              node: spanElement,
+              node: largerElement,
               previousRect: DOMRectReadOnly.fromRect({ x: 0, y: 0, width: 40, height: 40 }),
               currentRect: DOMRectReadOnly.fromRect({ x: 50, y: 50, width: 40, height: 40 }),
             },
@@ -356,7 +355,7 @@ describe('trackCumulativeLayoutShift', () => {
       expect(clsCallback.calls.mostRecent().args[0]).toEqual({
         value: 0.5,
         time: 1 as RelativeTime,
-        targetSelector: '#span-element',
+        targetSelector: '#larger-element',
         previousRect: { x: 0, y: 0, width: 40, height: 40 },
         currentRect: { x: 50, y: 50, width: 40, height: 40 },
         devicePixelRatio: jasmine.any(Number),
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
@@ -306,6 +306,7 @@ describe('trackCumulativeLayoutShift', () => {
     it('should get the target element, time, and rects of the largest layout shift', () => {
       startCLSTracking()
       const divElement = appendElement('<div id="div-element"></div>')
+      const spanElement = appendElement('<span id="span-element"></span>')
 
       // first session window:  { value: 0.5, time: 1, targetSelector: '#div-element' }
       notifyPerformanceEntries([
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -108,12 +110,20 @@ export function trackCumulativeLayoutShift(
...
+function getBiggestElementAttribution(
   sources: RumLayoutShiftAttribution[]
 ): (RumLayoutShiftAttribution & { node: Element }) | undefined {
-  return sources.find(
+  const elementNodeSources = sources.filter(
     (source): source is RumLayoutShiftAttribution & { node: Element } => !!source.node && isElementNode(source.node)
   )
+  if (elementNodeSources.length > 0) {
+    return elementNodeSources.reduce(function (a, b) {
+      return a.node && a.previousRect?.width * a.previousRect?.height > b.previousRect?.width * b.previousRect?.height
```
thomas-lebeau, line 121: **💬 suggestion:** ‏ I think we don't need to check `a.node` again as it's seems `elementNodeSources` is already filtering out object without the `node` property, no?

```suggestion
      return a.previousRect?.width * a.previousRect?.height > b.previousRect?.width * b.previousRect?.height
```

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -108,12 +110,20 @@ export function trackCumulativeLayoutShift(
...
 
-function getFirstElementAttribution(
+function getBiggestElementAttribution(
   sources: RumLayoutShiftAttribution[]
 ): (RumLayoutShiftAttribution & { node: Element }) | undefined {
-  return sources.find(
+  const elementNodeSources = sources.filter(
     (source): source is RumLayoutShiftAttribution & { node: Element } => !!source.node && isElementNode(source.node)
   )
+  if (elementNodeSources.length > 0) {
```
thomas-lebeau, line 119: **🥜 nitpick:** ‏use early returns, it makes it easier to read.

 ```suggestion
   if (elementNodeSources.length <= 0) {
     return
   }
```

**Comment by thomas-lebeau** [comment]
Thanks @anediaz, Could you please update [DataDog/rum-events-format](https://github.com/DataDog/rum-events-format) to add the new property?
Here is an example: https://github.com/DataDog/rum-events-format/pull/247. When this is merge, you will need to sync the repos by running `rum-events-format:sync` in the browser-sdk

File: packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts [hunk 3]
```diff
@@ -306,6 +310,7 @@ describe('trackCumulativeLayoutShift', () => {
     it('should get the target element, time, and rects of the largest layout shift', () => {
       startCLSTracking()
       const divElement = appendElement('<div id="div-element"></div>')
+      const spanElement = appendElement('<span id="span-element"></span>')
 
       // first session window:  { value: 0.5, time: 1, targetSelector: '#div-element' }
```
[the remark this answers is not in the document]
anediaz: yeah definitely! I'm removing it

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -74,10 +75,10 @@ export function trackCumulativeLayoutShift(
         continue
       }
 
-      const { cumulatedValue, isMaxValue } = window.update(entry)
+      const { cumulatedValue, isMaxValue, devicePixelRatio } = window.update(entry)
```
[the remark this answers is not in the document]
anediaz: the user can change the screen within the same view, and the `devicePixelRatio` can vary depending on the screen.
This is why I wanted to attach it to the moment where the CLS maxValue is reported.
I'd be happy to hear from your suggestions if you think it is a better way of doing it 😄

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
anediaz: true!

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.spec.ts
```diff
@@ -306,6 +310,7 @@ describe('trackCumulativeLayoutShift', () => {
     it('should get the target element, time, and rects of the largest layout shift', () => {
       startCLSTracking()
       const divElement = appendElement('<div id="div-element"></div>')
+      const spanElement = appendElement('<span id="span-element"></span>')
```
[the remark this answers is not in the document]
anediaz: yes, good idea! change made [here](https://github.com/DataDog/browser-sdk/pull/3377/commits/207ab54098667f097512ac4ea99fbd9bfb7e2ee9)

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
anediaz: done [here](https://github.com/DataDog/browser-sdk/pull/3377/commits/207ab54098667f097512ac4ea99fbd9bfb7e2ee9)

anediaz:
> Thanks @anediaz, Could you please update [DataDog/rum-events-format](https://github.com/DataDog/rum-events-format) to add the new property? Here is an example: [DataDog/rum-events-format#247](https://github.com/DataDog/rum-events-format/pull/247). When this is merge, you will need to sync the repos by running `rum-events-format:sync` in the browser-sdk

Thanks for the explanation!
I've created a PR for this. Here you are: [[RUM-8744]Add property device_pixel_ratio to CLS](https://github.com/DataDog/rum-events-format/pull/257)

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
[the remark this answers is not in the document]
thomas-lebeau: 🤔 true, but it's not calculated depending on previous values from the sliding window. 

I think what you want is to store `devicePixelRatio` at the time of the max value (i.e. in the `if (isMaxValue)` condition), the same way it is done for `attribution = getBiggestElementAttribution()`

I's adding some suggestion comments, so you see what I mean.

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -96,6 +97,7 @@ export function trackCumulativeLayoutShift(
           time: biggestShift?.time,
           previousRect: biggestShift?.previousRect ? asRumRect(biggestShift.previousRect) : undefined,
           currentRect: biggestShift?.currentRect ? asRumRect(biggestShift.currentRect) : undefined,
+          devicePixelRatio,
```
thomas-lebeau, line 100:
```suggestion
          devicePixelRatio = biggestShift?.devicePixelRatio,
```

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -74,10 +75,10 @@ export function trackCumulativeLayoutShift(
...
 
-      const { cumulatedValue, isMaxValue } = window.update(entry)
+      const { cumulatedValue, isMaxValue, devicePixelRatio } = window.update(entry)
 
       if (isMaxValue) {
-        const attribution = getFirstElementAttribution(entry.sources)
+        const attribution = getBiggestElementAttribution(entry.sources)
         biggestShift = {
           target: attribution?.node ? new WeakRef(attribution.node) : undefined,
           time: elapsed(viewStart, entry.startTime),
```
thomas-lebeau, line 84:
```suggestion
        biggestShift = {
          target: attribution?.node ? new WeakRef(attribution.node) : undefined,
          time: elapsed(viewStart, entry.startTime),
          previousRect: attribution?.previousRect,
          currentRect: attribution?.currentRect,
          devicePixelRatio: window.devicePixelRatio,
```

PR #3378, opened 2025-02-25
🐛 Fix empty string for resource url in LCP 

RomanGaignault:
## Motivation

The property url report an empty string if the value is not available, we shouldn't report it in this case.

## Changes

Add a check  function to discard empty string value

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


sethfowler-datadog (approved):
LGTM!

cy-moi (approved):

thomas-lebeau (approved):

PR #3389, opened 2025-03-03
🐛 [RUM-8744] CLS tracking: provide cls `devicePixelRatio` to adjust cls rect scale

anediaz:
## Motivation
https://datadoghq.atlassian.net/browse/RUM-8744

Implementation of small adjustments in Browser SDK for CLS attribution:
- Include `devicePixelRatio` attribute: 
   - Recent discoveries showed that Layout Instability API does not report viewPort coordinates for `prevRect` and `currentRect`.
   - We'd need to include this data in `_dd.cls.device_pixel_ratio`. See [this document](https://datadoghq.atlassian.net/wiki/spaces/~614b8bf2071141006ad0cebd/pages/4757587959/Bug+discovered+on+Google+Layout+Instability+API) for more context

image

## Changes
- Include new attribute `devicePixelRatio` in the CLS attribution

Tested in local:
image

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewCollection.ts
+++ b/packages/rum-core/src/domain/view/viewCollection.ts
@@ -65,10 +65,19 @@ function processViewUpdate(
   recorderApi: RecorderApi
 ): RawRumEventCollectedData<RawRumViewEvent> {
   const replayStats = recorderApi.getReplayStats(view.id)
+  const { cls_device_pixel_ratio: clsDevicePixelRatio, ...viewPerformanceData } = computeViewPerformanceData(
+    view.commonViewMetrics,
+    view.initialViewMetrics
+  )
   const viewEvent: RawRumViewEvent = {
     _dd: {
       document_version: view.documentVersion,
       replay_stats: replayStats,
+      cls: clsDevicePixelRatio
+        ? {
+            device_pixel_ratio: clsDevicePixelRatio,
+          }
+        : undefined,
       configuration: {
         start_session_replay_recording_manually: configuration.startSessionReplayRecordingManually,
       },
@@ -109,7 +118,7 @@ function processViewUpdate(
       long_task: {
         count: view.eventCounts.longTaskCount,
       },
-      performance: computeViewPerformanceData(view.commonViewMetrics, view.initialViewMetrics),
+      performance: viewPerformanceData,
       resource: {
         count: view.eventCounts.resourceCount,
       },
@@ -178,5 +187,6 @@ function computeViewPerformanceData(
       target_selector: largestContentfulPaint.targetSelector,
       resource_url: largestContentfulPaint.resourceUrl,
     },
+    cls_device_pixel_ratio: cumulativeLayoutShift && cumulativeLayoutShift.devicePixelRatio,
   }
 }
--- a/packages/rum-core/src/domain/view/viewCollection.ts
+++ b/packages/rum-core/src/domain/view/viewCollection.ts
@@ -65,10 +65,7 @@ function processViewUpdate(
   recorderApi: RecorderApi
 ): RawRumEventCollectedData<RawRumViewEvent> {
   const replayStats = recorderApi.getReplayStats(view.id)
-  const { cls_device_pixel_ratio: clsDevicePixelRatio, ...viewPerformanceData } = computeViewPerformanceData(
-    view.commonViewMetrics,
-    view.initialViewMetrics
-  )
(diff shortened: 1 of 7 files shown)


packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -65,10 +65,19 @@ function processViewUpdate(
   recorderApi: RecorderApi
 ): RawRumEventCollectedData<RawRumViewEvent> {
   const replayStats = recorderApi.getReplayStats(view.id)
+  const { cls_device_pixel_ratio: clsDevicePixelRatio, ...viewPerformanceData } = computeViewPerformanceData(
+    view.commonViewMetrics,
+    view.initialViewMetrics
+  )
```
BenoitZugmeyer, line 71: **💬 suggestion:** it would be simpler to avoid getting through `computeViewPerformanceData`.

```ts
const clsDevicePixelRatio = view.commonViewMetrics?.clsDevicePixelRatio
```

Advantages:
* Smaller impact on the bundle size
* No back and forth between snake_case and camelCase

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -83,6 +85,7 @@ export function trackCumulativeLayoutShift(
           time: elapsed(viewStart, entry.startTime),
           previousRect: attribution?.previousRect,
           currentRect: attribution?.currentRect,
+          devicePixelRatio: window.devicePixelRatio,
```
sethfowler-datadog, line 88: This will be the same for every shift, right? I think there's no need to include it in `LayoutShiftInstance`; it should be fine to just consult `window.devicePixelRatio` directly where you invoke `callback()` below.

packages/rum-core/src/domain/view/viewCollection.ts
anediaz: thanks for the recommendation 👍 
indeed the code is much cleaner now. Here's [the commit](https://github.com/DataDog/browser-sdk/pull/3389/commits/78d3da57d2dd0eac3d046162ceb4e68b5caac7e0)

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
thomas-lebeau: I guess it can change when you move the browser window to another screen that have a different pixel ratio. (and maybe moving to the new screen triggers layout shift if viewport changes too)

thomas-lebeau (approved):

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
sethfowler-datadog: This is post-processing code; the cumulative layout shift has already happened. All of these instances (which are particular elements affected by a single cumulative layout shift "incident") are being processed synchronously, so the `devicePixelRatio` will be the same for all of them.

anediaz: the user can switch from one screen to another during the same session and cls is continuously calculated.

the only way to guarantee we use the right `devicePixelRatio` is to include it to every reported attribution data of a shift (with prevRect and currentRect).

sethfowler-datadog: @anediaz Looks like we posted a response at the same time! Please take a look at my comment; does that clarify the situation?

sethfowler-datadog (approved):
LGTM!

BenoitZugmeyer (approved):

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -175,6 +178,7 @@ export interface ViewPerformanceData {
     target_selector?: string
     resource_url?: string
   }
+  cls_device_pixel_ratio?: number
```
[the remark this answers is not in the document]
anediaz: indeed! once again, thanks for the catch!
here's [the commit](https://github.com/DataDog/browser-sdk/pull/3389/commits/cf36092f48932a29537e19f63bdc5a9dbdf080fb) that removes it

PR #3391, opened 2025-03-04
🐛 [RUM-8606] Track First Hidden before init

RomanGaignault:
## Motivation

When a page is in background, the browser can pause or slow down its execution to preserve resources. This is known to skew page-load related metrics (ex: vitals like FCP/FID and other like `@view.loading_time`): when the user opens a page in background (ex: middle click), those timings can be arbitrary long (as long as the tab stays in background). This is why we try to detect that a page was hidden/in background and discard those timings in this case.

But we only do so once init is called: we don’t know whether a page was in background before init is called, or even before the SDK was loaded.

## Changes

Chrome is providing Performance entries called visibility-state that are notified whenever a page gets hidden/visible: [VisibilityStateEntry - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/VisibilityStateEntry) . It is possible to query past entries using performange.getEntriesByType('visibility-state'). When the browser support this, it could be used as initial state in trackFirstHidden .

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
@@ -11,7 +11,9 @@ export function trackFirstHidden(configuration: RumConfiguration, eventTarget: W
   }
 
   if (supportPerformanceTimingEvent(RumPerformanceEntryType.VISIBILITY_STATE)) {
-    const firstHiddenEntry = performance.getEntriesByType('visibility-state').find((entry) => entry.name === 'hidden')
+    const firstHiddenEntry = performance
+      .getEntriesByType(RumPerformanceEntryType.VISIBILITY_STATE)
+      .find((entry) => entry.name === 'hidden')
     if (firstHiddenEntry) {
       return { timeStamp: firstHiddenEntry.startTime as RelativeTime, stop: noop }
     }
--- a/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
@@ -5,11 +5,23 @@ import type { RumConfiguration } from '../../configuration'
 export type FirstHidden = ReturnType<typeof trackFirstHidden>
 
 export function trackFirstHidden(configuration: RumConfiguration, eventTarget: Window = window) {
-  let timeStamp: RelativeTime
+  let timeStamp: RelativeTime = Infinity as RelativeTime
   let stopListeners: () => void | undefined
-
+  let earliestHidden = Infinity
+  if (typeof performance !== 'undefined' && 'getEntriesByType' in performance) {
+    const visibilityEntries = performance.getEntriesByType('visibility-state')
+    if (visibilityEntries && visibilityEntries.length > 0) {
+      for (const entry of visibilityEntries) {
+        if (entry.name === 'hidden') {
+          earliestHidden = Math.min(earliestHidden, entry.startTime)
+        }
+      }
+    }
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
```diff
@@ -5,11 +5,24 @@ import type { RumConfiguration } from '../../configuration'
 export type FirstHidden = ReturnType<typeof trackFirstHidden>
 
 export function trackFirstHidden(configuration: RumConfiguration, eventTarget: Window = window) {
-  let timeStamp: RelativeTime
+  let timeStamp: RelativeTime = Infinity as RelativeTime
   let stopListeners: () => void | undefined
+  let earliestHidden = Infinity
 
+  if (typeof performance !== 'undefined' && 'getEntriesByType' in performance && supportsVisibilityStateEntries()) {
```
BenoitZugmeyer, line 12: **💬 suggestion:** ‏
You can use `supportPerformanceTimingEvent` from https://github.com/DataDog/browser-sdk/blob/c83b7076a447cea3eec53c9535a1a63a592ef187/packages/rum-core/src/browser/performanceObservable.ts#L297

```suggestion
  if (supportPerformanceTimingEvent('visibility-state')) {
```

No need to test for performance being defined, as all browsers we support support it.

packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
```diff
@@ -5,11 +5,24 @@ import type { RumConfiguration } from '../../configuration'
...
+    const visibilityEntries = performance.getEntriesByType('visibility-state')
+    if (visibilityEntries && visibilityEntries.length > 0) {
+      for (const entry of visibilityEntries) {
+        if (entry.name === 'hidden') {
+          earliestHidden = Math.min(earliestHidden, entry.startTime)
+        }
+      }
+    }
```
BenoitZugmeyer, line 20: **💬 suggestion:** ‏entries are ordered in the buffer so you could just get the first one

```ts
    performance.getEntriesByType('visibility-state').find(entry => entry.name === 'hidden')?.startTime
```

packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
```diff
@@ -5,11 +5,24 @@ import type { RumConfiguration } from '../../configuration'
...
+  }
   if (document.visibilityState === 'hidden') {
     timeStamp = 0 as RelativeTime
+  } else if (earliestHidden < Infinity) {
+    timeStamp = earliestHidden as RelativeTime
   } else {
     timeStamp = Infinity as RelativeTime
```
BenoitZugmeyer, line 27: **💬 suggestion:** maybe change the function design like the following to make it clearer:

```ts
function trackFirstHidden(configuration: RumConfiguration, eventTarget: Window = window): { timeStamp: TimeStamp, stop: () => void } {
    return { timeStamp: 0, stop: noop }
  }

  if (supportPerformanceTimingEvent('visibility-state')) {
    const firstHiddenEntry = performance.getEntriesByType('visibility-state').find(entry => entry.name === 'hidden')
    if (firstHiddenEntry) {
      return { timeStamp: firstHiddenEntry.startTime, stop: noop }
    }
  }
  
  let timeStamp = Infinity
  const { stop } = addEventListeners(...)
  return { get timeStamp() { return timeStamp }, stop }
}
```

thomas-lebeau (approved):

PR #3393, opened 2025-03-05
⚡️ Remove customer data track for perf and bundle size

cy-moi:
## Motivation
Customer data tracker can cause performance overhead and we want to decrease bundle size.

## Changes
Remove customer data tracker. 

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/contexts/featureFlagContext.ts
```diff
@@ -70,9 +67,7 @@ export function startFeatureFlagContexts(
       const currentContext = featureFlagContexts.find()
       if (currentContext && currentContext[key] !== value) {
         currentContext[key] = value
-        customerDataTracker.updateCustomerData(currentContext)
       }
```
BenoitZugmeyer, line 70: **💬 suggestion:** ‏the `currentContext[key] !== value`  condition isn't useful anymore, I think we could remove it

packages/core/src/domain/context/contextManager.spec.ts
```diff
@@ -1,42 +1,40 @@
 import { display } from '../../tools/display'
-import { noop } from '../../tools/utils/functionUtils'
 import type { PropertiesConfig } from './contextManager'
 import { createContextManager } from './contextManager'
-import { createCustomerDataTracker } from './customerDataTracker'
 
-function createNoopCustomerDataTracker(propertiesConfig?: PropertiesConfig) {
-  return createContextManager('test', { customerDataTracker: createCustomerDataTracker(noop), propertiesConfig })
+function mockContextManager(propertiesConfig?: PropertiesConfig) {
```
BenoitZugmeyer, line 5: **🥜 nitpick:** ‏it's not really a mock we create here. It's the real thing. I usually suffix this kind of function in tests with `WithDefaults`, like `createContextManagerWithDefaults`

amortemousque (approved):

PR #3394, opened 2025-03-05
🐛 Handle illegal invocation in addEventListener

thomas-lebeau:
## Motivation

Fix an issue where when both a Custom XMLHttpRequest override and zone.js are used on a website, we are trying to call the original (a.k.a. non-zone.js patched) method with the context of the override XMLHttpRequest instance resulting in a `Illegal Invocation` error.

We are prioritizing calling the non-patched method as zone.js methods are prone to [memory leaks](https://github.com/DataDog/browser-sdk/pull/1860)

## Changes

Try to invoke the non-patched method and fallback to the patched method in case it was an illegal invocation.

Note that in case this fails, the errors will be caught by our telemetry

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/browser/addEventListener.ts
+++ b/packages/core/src/browser/addEventListener.ts
@@ -127,19 +127,57 @@ export function addEventListeners<Target extends EventTarget, EventName extends 
 
   const options = passive ? { capture, passive } : capture
 
-  // Use the window.EventTarget.prototype when possible to avoid wrong overrides (e.g: https://github.com/salesforce/lwc/issues/1824)
-  const listenerTarget =
-    window.EventTarget && eventTarget instanceof EventTarget ? window.EventTarget.prototype : eventTarget
-
-  const add = getZoneJsOriginalValue(listenerTarget, 'addEventListener')
-  eventNames.forEach((eventName) => add.call(eventTarget, eventName, listenerWithMonitor, options))
+  eventNames.forEach((eventName) =>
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/addEventListener.ts
```diff
@@ -127,19 +127,57 @@ export function addEventListeners<Target extends EventTarget, EventName extends
...
+}
+
+function withOriginalOrZoneJsPatchedMethod<Target extends EventTarget, MethodName extends keyof EventTarget>(
+  eventTarget: Target,
+  methodName: MethodName,
+  cb: (method: (typeof eventTarget)[MethodName]) => void
+) {
+  // Use the window.EventTarget.prototype when possible to avoid wrong overrides (e.g: https://github.com/salesforce/lwc/issues/1824)
+  const listenerTarget =
+    window.EventTarget && eventTarget instanceof EventTarget ? window.EventTarget.prototype : eventTarget
```
sethfowler-datadog, line 168: Out of curiosity, does this situation ever occur if `eventTarget instanceof EventTarget`? I'm wondering if it would be enough to simply check for this and:
* When `eventTarget instanceof EventTarget`, use `originalMethod`.
* When `!(eventTarget instanceof EventTarget)`, use `eventTarget[methodName]`.

BenoitZugmeyer: This sounds interesting indeed.

PR #3395, opened 2025-03-05
✨ [RUM-7387] Add user id and account id to trace header

cy-moi:
## Motivation
Send user id and account id in trace header so they could propagate.
[[RFC] - Adding User and Account Id
](https://docs.google.com/document/d/1W1rrr3tMscxTGmP-jaxrub6PSFyUGn8_fYssbL382q8/edit?tab=t.0)

## Changes
Passed common context to tracer and add user id and account id in trace context.
Encode in base64 as the field was already in base64. ([Initial RFC on usr.id field](https://docs.google.com/document/d/1T3qAE5nol18psOaHESQ3r-WRiZWss9nyGmroShug8ao/edit?tab=t.0#heading=h.3wmduzc8mwe1))

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---


packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -158,11 +187,23 @@ function makeTracingHeaders(
...
       case 'tracecontext': {
+        let userAccountTraceHeader: string | undefined
+        if (isExperimentalFeatureEnabled(ExperimentalFeature.USER_ACCOUNT_TRACE_HEADER)) {
+          /**
+           * We should only enable this feature after the decoding service is available
+           */
+          const userIdEncoded = encodeToUtf8Base64(getCommonContext().user.id?.toString())
+          // we use Context type for account internally but account id should not be an object
+          /* eslint-disable-next-line @typescript-eslint/no-base-to-string */
+          const accountIdEncoded = encodeToUtf8Base64(getCommonContext().account.id?.toString())
```
thomas-lebeau, line 198: **💭 thought:** ‏ I don't know if the performance overhead is significant, but as user.id and account.id are most probably the same for all traces, this could be easily cached/memoized. wdyt?

cy-moi: Good point. They are from context manager, so I assume we already do some caching here? 

There is also the Kiosk use case where these fields are expected to change frequently.

packages/logs/src/rawLogsEvent.types.ts
```diff
@@ -68,5 +68,5 @@ export interface CommonContext {
   }
   context: Context
   user: User
-  account: Context
+  account: Context // use Context for internal Account type
```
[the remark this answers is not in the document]
thomas-lebeau: hmm, ok. Maybe you can add _why_ in the comment, so the next person to read that will understand too.

Or maybe naming `AccountWithId` would help

packages/rum-core/src/domain/tracing/tracer.ts
thomas-lebeau: I meant caching the encoded values, so we don't run the encoding algorithm for every requests.
Even in the case of the Kiosk, there is most likely multiple request in a row with the same ids

BenoitZugmeyer (approved):

PR #3405, opened 2025-03-07
🐛 fix `@datadog/browser-rum-react/react-router-vX` import

BenoitZugmeyer:
## Motivation

Currently, importing `@datadog/browser-rum-react/react-router-v6` or `@datadog/browser-rum-react/react-router-v7` from an external project fails because those entry points are importing `react-router-dom-6` and `react-router-dom-7`, which are not actual packages.

## Changes

This PR changes the react-router entrypoints to import `react-router-dom` instead of aliases. Since both entry points are now identical, they are merged into a single module. We keep the two distinct user-facing entry points (`@datadog/browser-rum-react/react-router-v6` and `-v7`) to be able to potentially offer different functionalities in the future, and keep the established scheme.

In future PRs, we'll:

* add a lint rule to make sure we don't import dev dependencies in prod files
* add an integration test with an actual rum-react project similar to `test/app`

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-react/src/entries/reactRouter.ts
```diff
@@ -8,7 +8,7 @@ import {
   useLocation,
   matchRoutes,
   createRoutesFromChildren,
-} from 'react-router-dom-6'
+} from 'react-router-dom'
```
Ayc0, line 11: 👌

Ayc0 (approved):
This sounds good to me!
Thanks for the quick fix!

cy-moi (approved):

PR #3406, opened 2025-03-07
🐛 [RUM-99] don't stop RUM/Replay on beforeunload

BenoitZugmeyer:
## Motivation

`beforeunload` events are triggered when "navigating" to URLs with schemes that are not handled by the browser, like `mailto:toto@tata.com` or `mycustomescheme://toto`.

Currently, the SDK is stopping RUM and Replay collection on `beforeunload`.

This is problematic for some customers that have this kind of URLs.

## Changes

This PR makes sure we don't stop anything on `beforeunload`. We'll keep flushing data on `beforeunload` to make sure we don't impact reliability too much, as any data produced after `beforeunload` has no guarantee to be sent to datadog.

Now, during `beforeunload`:

* We'll simply not stop Replay recording. No "ViewEnd" record will be sent on page exit, which might have a small impact on the Player as this record is used to "garbage collect" iframes. After a discussion with @mquentin, this risk seems acceptable, and if it ever becomes problematic we have multiple alternative strategies to garbage collect iframes.

* We'll keep marking the current view as inactive. If we didn't do that, views would never be marked as inactive on page exit, which can have a significant impact (ex: metrics are generated when a view is marked inactive). Sending View updates and other events is possible even after the View is marked as inactive, so it shouldn't be an issue (seen with @joaooliveira-datadog )

* We'll stop the current click action "click chain". This was done on view end, but since there is no view end anymore on page exit, we'll need to stop it so pending click actions gets collected immediately.

## Testing

This can be tested by running `location.assign("foo:bar")` in the devtools console. RUM and Replay should continue collecting data after doing so.

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


amortemousque (approved):

sethfowler-datadog (approved):
LGTM!

PR #3410, opened 2025-03-10
📝 [RUM-7705] Add js doc about option that need to be aligned

RomanGaignault:
## Motivation

If options are not aligned between RUM and LOGS the SDK could not initialised correctly.

## Changes

Add Js doc to highlight this better

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -28,18 +28,6 @@ export const TraceContextInjection = {
 
 export type TraceContextInjection = (typeof TraceContextInjection)[keyof typeof TraceContextInjection]
 
-/**
- * Important, if you are using the RUM and Logs Browser SDKs, the following options
- * must be configured with identical values :
- *
- * - sessionPersistence
- * - trackSessionAcrossSubdomains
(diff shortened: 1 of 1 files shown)


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -28,6 +28,18 @@ export const TraceContextInjection = {
...
+ * must be configured with identical values :
+ *
+ * - sessionPersistence
+ * - trackSessionAcrossSubdomains
+ * - useSecureSessionCookie
+ * - usePartitionedCrossSiteSessionCookie
+ * - allowFallbackToLocalStorage (deprecated)
+ *
+ * Mismatching these options between RUM and Logs can cause inconsistent session behavior.
+ */
```
cy-moi, line 42: 💬 Suggestion: Maybe we add this to `RumInitConfiguration` and `LogsInitConfiguration` so it gets picked up by the interface users would call?

cy-moi (approved):
LGTM
Maybe we can also add line breaks for the important lines if it works.

PR #3417, opened 2025-03-12
♻️ [RUM-8795] Replace XHR with Fetch for HTTP requests

thomas-lebeau:
## Motivation

Over time we've added a fixes to our xhr strategy to account for bad implementations/overrides (IE content type header, multiple onResponse callbacks, untrusted events...)

Instead of yet another fix to account for the combination of dynatrace + zonejs making xhr request to fail, and as we don't support IE anymore, we can use plain fetch instead of xhr.

closes #3394 

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/transport/httpRequest.ts
+++ b/packages/core/src/transport/httpRequest.ts
@@ -1,8 +1,6 @@
 import type { EndpointBuilder } from '../domain/configuration'
-import { addTelemetryError } from '../domain/telemetry'
 import type { Context } from '../tools/serialisation/context'
-import { monitor } from '../tools/monitor'
-import { addEventListener } from '../browser/addEventListener'
+import { monitor, monitorError } from '../tools/monitor'
 import type { RawError } from '../domain/error/error.types'
 import { newRetryState, sendWithRetryStrategy } from './sendWithRetryStrategy'
 
@@ -72,8 +70,7 @@ function sendBeaconStrategy(endpointBuilder: EndpointBuilder, bytesLimit: number
     }
   }
 
-  const xhrUrl = endpointBuilder.build('xhr', payload)
-  sendXHR(xhrUrl, payload.data)
+  fetchStrategy(endpointBuilder, payload)
 }
 
 let hasReportedBeaconError = false
@@ -81,7 +78,7 @@ let hasReportedBeaconError = false
 function reportBeaconError(e: unknown) {
   if (!hasReportedBeaconError) {
     hasReportedBeaconError = true
-    addTelemetryError(e)
+    monitorError(e)
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/transport/httpRequest.ts
```diff
@@ -72,16 +70,15 @@ function sendBeaconStrategy(endpointBuilder: EndpointBuilder, bytesLimit: number
...
+  fetchStrategy(endpointBuilder, payload)
 }
 
 let hasReportedBeaconError = false
 
 function reportBeaconError(e: unknown) {
   if (!hasReportedBeaconError) {
     hasReportedBeaconError = true
-    addTelemetryError(e)
+    monitorError(e)
```
thomas-lebeau, line 81: `monitorError()` will send telemetry but will also log debug message in the console  (when `DD_RUM._setDebug(true)`)

packages/core/src/transport/httpRequest.ts
```diff
@@ -92,22 +89,30 @@ export function fetchKeepAliveStrategy(
...
+export function fetchStrategy(
+  endpointBuilder: EndpointBuilder,
+  payload: Payload,
+  onResponse?: (r: HttpResponse) => void
+) {
+  const fetchUrl = endpointBuilder.build('fetch', payload)
+
+  fetch(fetchUrl, { method: 'POST', body: payload.data, mode: 'cors' })
+    .then(monitor((response: Response) => onResponse?.({ status: response.status, type: response.type })))
+    .catch(() => onResponse?.({ status: 0 }))
```
thomas-lebeau, line 113: We do not monitor failed fetch request, as with the previous xhr implementation.

There is a lot of reason a request can fail, (no network, adblockers, ...). Filtering these reliably is difficult, and even if we could, monitoring the remaining errors does not give us actionable items

packages/core/test/interceptRequests.ts
```diff
@@ -15,7 +19,7 @@ export function mockEndpointBuilder(url: string) {
 }
 
 export interface Request {
-  type: 'xhr' | 'sendBeacon' | 'fetch'
+  type: 'sendBeacon' | 'fetch' | 'fetch-keepalive' // TODO remove xhr
```
BenoitZugmeyer, line 22: **🥜 nitpick:** ‏TODO: remove TODO

packages/core/test/interceptRequests.ts
```diff
@@ -26,27 +30,50 @@ export function interceptRequests() {
...
+      throw new Error('No fetch mock provided')
+    }
+
+    return fetchPromise()
+      .then((response) => {
+        requests.push({
+          type: config?.keepalive ? 'fetch-keepalive' : 'fetch',
+          url: url as string,
+          body: config!.body as string,
+        })
```
BenoitZugmeyer, line 60: **💬 suggestion:** ‏push the request before calling `fetchPromise()`. I find it confusing that we don't push the request if fetchPromise is rejected.

packages/core/src/transport/httpRequest.ts
BenoitZugmeyer, line 113: **💬 suggestion:** ‏
```suggestion
    .catch(monitor(() => onResponse?.({ status: 0 })))
```

BenoitZugmeyer (approved):

PR #3418, opened 2025-03-12
👷 Enable the Terser unsafe option

sethfowler-datadog:
## Motivation

This is one of a series of PRs experimenting with enabling more aggressive options for Terser, with the goal of reducing the size of the browser SDK bundles.

## Changes

This PR enables the following options:
* `unsafe: true`

Here's what the Terser docs have to say about `unsafe: true`:

> It enables some transformations that might break code logic in certain contrived cases, but should be fine for most code. It assumes that standard built-in ECMAScript functions and classes have not been altered or replaced. You might want to try it on your own code; it should reduce the minified size. Some examples of the optimizations made when this option is enabled:
> * new Array(1, 2, 3) or Array(1, 2, 3) → [ 1, 2, 3 ]
> * Array.from([1, 2, 3]) → [1, 2, 3]
> * new Object() → {}
> * String(exp) or exp.toString() → "" + exp
> * new Object/RegExp/Function/Error/Array (...) → we discard the new
> * "foo bar".substr(4) → "bar"

It's a bit unfortunate that these transformations don't seem to have been split out into their own, isolated `unsafe_*` options, which would make it a bit easier to evaluate them one by one. (There are other `unsafe_*` options, but there don't seem to be any that control the transformations controlled by `unsafe: true`.)

In general, these transformations strike me as reasonable, in that they seem semantically correct unless someone has been replacing built-in functions/prototypes/etc with the goal of giving them behavior contrary to the originals, and in such a situation, I suspect a general purpose library like this one would have problems with or without these transformations. That said, great caution is certainly warranted here; I'll test extensively on staging before merging.

## Testing

- [x] Local
- [ ] Staging
- [x] Unit
- [ ] End to end

---

### DIFF
--- a/webpack.base.js
+++ b/webpack.base.js
@@ -62,6 +62,7 @@ module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins
           module: true,
           compress: {
             passes: 4,
+            unsafe: true,
             unsafe_methods: true,
           },
         },


sethfowler-datadog:
Looks OK on staging!

BenoitZugmeyer:
I was a bit curious about the size reduction, and the main change in rum and rum-slim is: Terser is able to see that the options passed to [`readBytesFromStream`](https://github.com/DataDog/browser-sdk/blob/13870449d24f75a46c2fb6a522eb146ef676d04a/packages/core/src/tools/readBytesFromStream.ts#L16)
 are always `{  bytesLimit: Number.POSITIVE_INFINITY,  collectStreamBody: false }` in RUM, so it optimizes `readBytesFromStream`.

Before:

```js

function f(t, e, n) {
  const r = t.getReader(),
    o = [];
  let s = 0;
  function a() {
    let t, a;
    if ((r.cancel().catch(i.l), n.collectStreamBody)) {
      let e;
      if (1 === o.length) e = o[0];
      else {
        e = new Uint8Array(s);
        let t = 0;
        o.forEach((n) => {
          e.set(n, t), (t += n.length);
        });
      }
      (t = e.slice(0, n.bytesLimit)), (a = e.length > n.bytesLimit);
    }
    e();
  }
  !(function t() {
    r.read().then(
      S.dm((e) => {
        e.done
          ? a()
          : (n.collectStreamBody && o.push(e.value),
            (s += e.value.length),
            s > n.bytesLimit ? a() : t());
      }),
      S.dm((t) => e()),
    );
  })();
```

With your PR:

```js
function e(t, e) {
  const n = t.getReader();
  let r = 0;
  function o() {
    n.cancel().catch(i.l), e();
  }
  !(function t() {
    n.read().then(
      k.dm((e) => {
        e.done ? o() : ((r += e.value.length), r > 1 / 0 ? o() : t());
      }),
      k.dm((t) => e()),
    );
  })();
}
```

It's kind of amazing.

amortemousque (approved):

sethfowler-datadog:
> It's kind of amazing.

Yeah, it's really impressive what Terser can do! "Minification" undersells what Terser does, I think; it's a source-to-source optimizing compiler for JavaScript.

PR #3420, opened 2025-03-12
👷 [RUM-105] use local dev server for local e2e tests

thomas-lebeau:
## Motivation

Improve e2e test development workflow, by using bundles served by the local dev server instead of the built bundles. This mean it is no longer necessary to run `yarn build` (or `yarn build:bundle` before e2e test on your local dev environment, and that any changes to the SDK will be picked up in the next e2e run, speeding up development workflows.

Playwright will automatically launch `yarn dev` if the local dev server is not already running, otherwise it will otherwise it will use the one running.

> [!NOTE]
> This change only affect local dev workflow, nothing changes on the CI 

## Changes

- Get Rid of `withSetup()` as it's only ever used to force some test to run exclusively the `bundle` setup. It give the false impression that these test should not work on other setup, but they do. When implementing e2e test for framework specific, we might need somthing similar, but we might want to be more specific (i.e. `withReactTestApp()`, `withFooTestApp()` , ...)
- refactor the route handlers in server app to be more generic as we already follow a pattern. This will help as more module are being lazy loaded, and more test app are used, we don't need to export a constant for each bundle.

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [x] End to end

---


test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -116,34 +116,27 @@ export function createMockServerApp(servers: Servers, setup: string): MockServer
...
+      res.sendFile(getSdkBundlePath('worker', req.originalUrl))
+    } else {
+      forwardToDevServer(req.originalUrl, res)
+    }
   })
 
-  app.get('/chunks/:name-:hash-app.js', (req, res) => {
-    const { name, hash } = req.params
-    res.sendFile(sdkBuilds.npmBundleChunks(name, hash))
+  app.get(/(?<appName>app).js$/, (req, res) => {
```
thomas-lebeau, line 137: @RomanGaignault, I think you'll just need to add `react-app` to the regex:
```suggestion
  app.get(/(?<appName>app|react-app).js$/, (req, res) => {
```

test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -152,3 +145,27 @@ export function createMockServerApp(servers: Servers, setup: string): MockServer
     },
   })
 }
+
+// We fetch and pipe the file content instead of redirecting to avoid creating different behavior between CI and local dev
+// This way both environments serve the files from the same origin with the same CSP rules
+function forwardToDevServer(originalUrl: string, res: ServerResponse) {
+  const url = `http://localhost:8080${originalUrl}`
```
amortemousque, line 152: **💬 suggestion:** ‏Can we store the origin in a constant and use it here as well as in `playwright.local.config.ts`?

sethfowler-datadog (approved):
Nicely done; not only should this make the local development experience smoother, but the code also feels cleaner!

PR #3432, opened 2025-03-18
♻️ [RUM-9126] Decouple common contexts from rumPublicApi 

amortemousque:
## Motivation

"Decouple the common context from rumPublicApi. The common contexts (user, account, global) are currently tied to rumPublicApi to ensure that saving and retrieving context work before initialization.

To prevent any breaking changes, this PR:
- Buffers the context API calls to ensure saving context before initialization works
- Maintains a 'fake' context in the pre-init logic so that DD_RUM.getGlobalContext() also works

(Review per commit)

## Changes

- Create dedicated module for account, user and global context.
- Replace common context in rumPublicApi.
- Remove savedCommonContext in collections and assembly

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -202,12 +213,81 @@ export function createPreStartStrategy(
...
+
+    removeUserProperty(key) {
+      userContextManager.removeContextProperty(key)
+      bufferApiCalls.add((startRumResult) => startRumResult.removeUserProperty(key))
+    },
+
+    clearUser() {
+      userContextManager.clearContext()
+      bufferApiCalls.add((startRumResult) => startRumResult.clearUser())
+    },
```
BenoitZugmeyer, line 260: **💬 suggestion:** ‏finally, you could refactor all those methods with something like this:

```ts
  function defineContextMethod<
    MethodName extends 'getContext' | 'setContext' | 'setContextProperty' | 'removeContextProperty' | 'clearContext',
  >(
    contextName: 'userContext' | 'globalContext' | 'accountContext',
    methodName: MethodName,
    usage?: 'set-user' | 'set-global-context'
  ): ContextManager[MethodName] {
    return monitor((...args: any[]) => {
      if (usage) {
        addTelemetryUsage({ feature: usage })
      }
      return strategy[contextName][methodName](...args)
    })
  }
```

then (we could probably have enums or constants to avoid string repetitions)

```ts
    setGlobalContext: defineContextMethod('globalContext', 'setContext', 'set-global-context'),
    getGlobalContext: defineContextMethod('globalContext', 'getContext'),
    setGlobalContextProperty: defineContextMethod('globalContext', 'setContextProperty', 'set-global-context'),
    removeGlobalContextProperty: defineContextMethod('globalContext', 'removeContextProperty'),
    clearGlobalContext: defineContextMethod('globalContext', 'clearContext'),
```

packages/rum-core/src/domain/contexts/userContext.ts
```diff
@@ -0,0 +1,29 @@
+import type { CustomerDataTrackerManager } from '@datadog/browser-core'
+import { createContextManager, CustomerDataType, storeContextManager } from '@datadog/browser-core'
+import type { RumConfiguration } from '../configuration'
+
+// export type UserContext = ReturnType<typeof startUserContext>
```
sethfowler-datadog, line 5: Nit: leftover commented-out code?

packages/rum-core/src/domain/contexts/userContext.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { mockRumConfiguration } from '../../../test'
+import { startUserContext } from './userContext'
+
+describe('user context', () => {
+  let userContext: ContextManager
+  let customerDataTrackerManager: CustomerDataTrackerManager
+
+  beforeEach(() => {
+    customerDataTrackerManager = createCustomerDataTrackerManager(CustomerDataCompressionStatus.Disabled)
+    userContext = startUserContext(customerDataTrackerManager, mockRumConfiguration({ trackAnonymousUser: false }))
```
sethfowler-datadog, line 17: Does `{ trackAnonymousUser: false }` influence the behavior of these tests at all? It's not obvious to me where it gets read; perhaps inside `storeContextManager()`?

packages/rum-core/src/domain/contexts/accountContext.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+  let displaySpy: jasmine.Spy
+
+  beforeEach(() => {
+    const customerDataTrackerManager = createCustomerDataTrackerManager(CustomerDataCompressionStatus.Disabled)
+    displaySpy = spyOn(display, 'warn')
+
+    accountContext = startAccountContext(customerDataTrackerManager, mockRumConfiguration())
+  })
+
+  it('should get account', () => {
```
sethfowler-datadog, line 23: The tests are quite similar for each of the context types! It's not a big deal, but I wonder if it's possible to eliminate some of the redundancy by restructuring the tests a little.

sethfowler-datadog (approved):
I like the new approach! Definitely cleaner.

packages/rum-core/src/domain/contexts/userContext.spec.ts
amortemousque: It's a leftover after I split my PR. Good catch!

packages/rum-core/src/domain/contexts/accountContext.spec.ts
amortemousque: You're right. In fact, I'm wondering if I'm testing the underlying APIs of `CustomerDataTrackerManager` and `storeContextManager`, which are already tested. Maybe I can remove most of the tests 🤔
I have a follow-up PR to use the assembly hook in these modules. I'll refine the tests in that one.

thomas-lebeau (approved):

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -591,31 +564,63 @@ export function makeRumPublicApi(
...
+    removeAccountProperty: defineContextMethod(
+      CustomerContextKey.accountContext,
+      ContextManagerMethod.removeContextProperty,
+      'remove-account-property'
+    ),
+    clearAccount: defineContextMethod(
+      CustomerContextKey.accountContext,
+      ContextManagerMethod.clearContext,
+      'clear-account'
+    ),
```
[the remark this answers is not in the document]
amortemousque: I considered this approach but didn't want to complicate the logic too much, at least in this PR :) . However, it might be worth revisiting the usage telemetry event to adopt camelCase formatting. We could wrap the API to automatically send usage telemetry, like so:

```ts
{
sendUsage(addError)
...
}
```
I'll create a card to track it. Wdyt?

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -39,6 +43,17 @@ export function createPreStartStrategy(
   ) => StartRumResult
 ): Strategy {
   const bufferApiCalls = createBoundedBuffer<StartRumResult>()
+  const customerDataTrackerManager = createCustomerDataTrackerManager(CustomerDataCompressionStatus.Unknown)
+
+  // TODO next major: remove the globalContextManager, userContextManager and accountContextManager from preStartStrategy and use an empty context instead
```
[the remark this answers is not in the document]
amortemousque: Right, I'll create the Jira card.

packages/rum-core/src/boot/rumPublicApi.ts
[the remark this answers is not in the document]
thomas-lebeau: > didn't want to complicate the logic too much

It's a bit less readable but it might be a somewhat significant bundle size reduction. We can then add a comment listing all the methods created for readability (and searchability)

> it might be worth revisiting the usage telemetry event to adopt camelCase formatting

I like the  `sendUsage`, but the formating could be computed automatically if we want.

All said, yes, let's track these possible improvements!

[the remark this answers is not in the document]
amortemousque: More than the complexity, the issue I see with your suggestion is the lack of type checking. However, it’s a good point, so let’s put more thought into it. I’ll create a card to track it. 
Also, in the next quarter, I know @benoit plans to revisit the prestart logic, and we’ll make sure this kind of improvement is on the table 😉

PR #3439, opened 2025-03-23
Expose Site type

ludwigbacklund:
## Motivation

Without this type it's not possible to create intermediary types for Datadog configuration in my application.

## Changes

It exposes the Site type from the core package.


thomas-lebeau:
Hi @ludwigbacklund, aren't you able to use `RumInitConfiguration['site']` instead?

ludwigbacklund:
> Hi @ludwigbacklund, aren't you able to use `RumInitConfiguration['site']` instead?

Ah yeah that works! I think the Site type would make sense to expose though, but it's up to you.

PR #3444, opened 2025-03-26
support declaration maps

quisido:
## Motivation

When I Ctrl-Click into the Datadog SDK from VS Code,
I see the TypeScript definition file.
However, I expect to see the source code.

## Changes

I enabled declaration maps in the base TSConfig file. This should generate a `.d.ts.map` file in addition to `.d.ts`, which will allow tools to map the declaration file to the source code, which is already included in the package (I only checked the `core` and `logs` packages).

## Testing

YOLO

---

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
+    "declarationMap": true,
     "experimentalDecorators": true,
     "esModuleInterop": true,
     "importHelpers": false,


bcaudan:
Hi @quisido,

This would be a nice addition but we currently have two concerns about it:

- The additional impact on the size of the npm packages
- The future support of this feature if/when we change our build setup

So we'd prefer to properly evaluate this addition when we'll look into improving our build setup (no ETA).
Closing this PR for now but we'll ping here if we make any progress on it.

PR #3448, opened 2025-03-26
🐛 fix TypeScript 4 compatibility

BenoitZugmeyer:
## Motivation

`export type *` is supported in TypeScript v5+. For some reason, using `export type *` in v3.8.2 does not produce any error so our typescript compatibility check did not catch this issue.

TypeScript v4 on the other hand fails with:

```
  ../../packages/core/cjs/domain/account/index.d.ts(1,1): error TS1383: Only named exports may use 'export type'.
  ../../packages/core/cjs/domain/telemetry/index.d.ts(4,1): error TS1383: Only named exports may use 'export type'.
  ../../packages/core/cjs/domain/user/index.d.ts(1,1): error TS1383: Only named exports may use 'export type'.
```

## Changes

* Convert the ts compat script to Node.js
* Add ts 4 compat check
* Remove problematic `export type *`

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


scripts/lib/executionUtils.js
```diff
@@ -22,8 +22,7 @@ function runMain(mainFunction) {
     // callback that will catch both thrown errors and rejected promises
     .then(() => mainFunction())
     .catch((error) => {
-      printError('\nScript exited with error:')
-      printErrorWithCause(error)
```
BenoitZugmeyer, line 26: Node 23 now prints error cause by itself, no need to do it ourself else it'll be printed twice

scripts/check-typescript-compatibility.js
```diff
@@ -0,0 +1,72 @@
...
+
+  const checks = [
+    {
+      title: 'TypeScript 3.8.2 compatibility',
+      version: '3.8.2',
+    },
+    {
+      title: 'TypeScript isolated modules compatibility',
+      compilerOptions: { isolatedModules: true },
+      version: '3.8.2',
```
BenoitZugmeyer, line 25: We used to test `isolatedModules` with 3.8.2, I didn't want to change this without a good reason.

scripts/check-typescript-compatibility.js
```diff
@@ -0,0 +1,72 @@
...
+    {
+      title: 'exactOptionalPropertyTypes compatibility',
+      version: 'latest', // Not available in 3.8.2
+      compilerOptions: { exactOptionalPropertyTypes: true },
+    },
+    {
+      title: 'ESNext compatibility',
+      version: 'latest',
+      compilerOptions: { lib: ['ESNext', 'DOM'] },
+    },
```
BenoitZugmeyer, line 44: All those tests were only one in the past. Keeping theme separated makes it easier to understand what fails.

thomas-lebeau (approved):

PR #3450, opened 2025-03-28
[RUM Profiler] Improvements around Views and Long Tasks

thomasbertet:
## Motivation

- Attribute `view.ids` had a `s` that was not coherent with other attributes.
- Views are missing from the exported JSON.
- Long Tasks are missing the  RUM Long Task `id` (when available)
- Long Tasks are full `PerformanceEntry` objects, and we don't need most of it.

## Changes
- Fix how we capture views so that we now report them properly. 
- Add name to view entry so that Profiling BE can do aggregation using that name. In the UI it would ultimately use the `viewId` to get the latest name.
- Make sure attribute fields are coherent.
- Add `id` to the Long Task entry, and make it a lighter object than the full performance entry.

## Testing
- Tested locally and in staging that everything is now reported. 
- In staging, you can : 
   - Open any page, 
   - Stay on a single page from the initial page load
   - Assert that the profile sent has a unique view. 
   - Assert Long tasks now have `id` when applicable.
   - 
- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -53,8 +53,9 @@ export function createRumProfiler(
     }
 
     // Add initial view
+    // Note: `viewEntry.name` is only filled when users use manual view creation via `startView` method.
     lastViewEntry = viewEntry
-      ? { startTime: performance.now(), viewId: viewEntry.id, viewName: viewEntry.name }
+      ? { startTime: viewEntry.startClocks.relative, viewId: viewEntry.id, viewName: viewEntry.name }
       : undefined
 
     // Start profiler instance
@@ -113,7 +114,8 @@ export function createRumProfiler(
 
     // Whenever the View is updated, we add a views entry to the profiler instance.
     const viewUpdatedSubscription = lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (view) => {
-      collectViewEntry({ viewId: view.id, viewName: view.name, startTime: performance.now() })
+      // Note: `view.name` is only filled when users use manual view creation via `startView` method.
+      collectViewEntry({ viewId: view.id, viewName: view.name, startTime: view.startClocks.relative })
     })
     cleanupTasks.push(viewUpdatedSubscription.unsubscribe)
 
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -8,7 +8,7 @@ import {
   getGlobalObject,
 } from '@datadog/browser-core'
 
-import type { LifeCycle, RumConfiguration, RumSessionManager } from '@datadog/browser-rum-core'
+import type { LifeCycle, RumConfiguration, RumSessionManager, ViewHistoryEntry } from '@datadog/browser-rum-core'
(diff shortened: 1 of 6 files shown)


packages/rum/src/domain/profiling/profiler.ts
```diff
@@ -31,15 +43,19 @@ export function createRumProfiler(
...
-  function start(viewId: string | undefined): void {
+  function start(viewEntry: ViewHistoryEntry | undefined): void {
     if (instance.state === 'running') {
       return
     }
 
     // Add initial view
-    collectViewEntry(viewId)
+    lastViewEntry = viewEntry
+      ? { startTime: performance.now(), viewId: viewEntry.id, viewName: viewEntry.name }
```
amortemousque, line 57: The view name is computed in the backend. It's only available in the frontend when the customer wants to override it using `startView` API.

thomasbertet: What do you suggest we use ? In practice it seems to be defined and sensical, cf. [link](https://dd.datad0g.com/profiling/explorer?query=service%3Aweb-ui&agg_m=%40prof_chrome_cpu_cores&agg_m_source=base&agg_q=service&agg_q_source=base&agg_t=sum&event=AwAAAZX14CS3A0TO-AAAABhBWlgxNENVRkFBQ3dOcmlfajNHUXVBQUEAAAA1MDE5NWY1ZTMtYzU1ZS00MTYyLWJhZTctOWQ3ZTczZDlmMDZlLXN5bnRoZXRpYy1saXN0LTAAAuSA&fromUser=false&my_code=disabled&profileId=AZX14CUFAACwNri_j3GQuAAA&top_n=100&top_o=top&viz=stream&x_missing=true&start=1743572462631&end=1743586862631&paused=true)

amortemousque: also for the startTime you should probably do:
```suggestion
      ? { startTime: viewEntry.startClock.startClocks.relative, viewId: viewEntry.id, viewName: viewEntry.name }
```

thomasbertet: Oh so re-reading this, you mean the `viewName` is actually set for us because we are using the `startView` API, but could be empty for customers not using that API ? 
I believe that's fine. We are using the `viewName` to better categorize the samples we get from the Profiler, and thus creating filters per "views" like the following: 
Screenshot 2025-04-02 at 17 05 48

I'm happy to change this, but at the same time I'm fine dealing with empty values in our processing, and use it whenever it's available as I find it to be valuable.

amortemousque: Yes at Datadog we are using `startView`, this is why it works for us. However `startView` API is used by less than 10% of customers, so if categorizing by viewName is valuable, we should consider a different strategy. I'm fine merging this for internal dogfooding, but we should put more thought into it before GA.

thomasbertet: Updated code to use `viewEntry.startClocks.relative` instead of `performance.now()`. I confirmed it was similar (except for initial view but that's even better)

Also, as agreed together, let's keep usage of `view.name` for dogfooding and start discussion around better support for more generic view name handling in Profiling.

amortemousque (approved):

PR #3456, opened 2025-03-31
♻️ Refactor cookieObservable to not rely on cookieStore

thomas-lebeau:
## Motivation

The test `should notify observers on cookie change` is very flaky and it does not seems to be cause by the test setup or mocks.
In the other hand the fallback does not seems to suffer the same problems, so let's only use the fallback

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/rum-core/src/browser/cookieObservable.spec.ts
+++ b/packages/rum-core/src/browser/cookieObservable.spec.ts
@@ -37,7 +37,7 @@ describe('cookieObservable', () => {
     clock.tick(WATCH_COOKIE_INTERVAL_DELAY)
   })
 
-  it('should not notify observers on cookie change when the cookie value as not changed', () => {
+  it('should not notify observers on cookie change when the cookie value has not changed', () => {
     const observable = createCookieObservable(COOKIE_NAME)
 
     setCookie(COOKIE_NAME, 'foo', COOKIE_DURATION)
--- a/packages/rum-core/src/browser/cookieObservable.spec.ts
+++ b/packages/rum-core/src/browser/cookieObservable.spec.ts
@@ -2,7 +2,6 @@ import type { Subscription } from '@datadog/browser-core'
 import { ONE_MINUTE, deleteCookie, setCookie } from '@datadog/browser-core'
 import type { Clock } from '@datadog/browser-core/test'
 import { mockClock } from '@datadog/browser-core/test'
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/browser/cookieObservable.spec.ts
```diff
@@ -38,22 +37,8 @@ describe('cookieObservable', () => {
...
-    setCookie(COOKIE_NAME, 'foo', COOKIE_DURATION)
-    clock.tick(WATCH_COOKIE_INTERVAL_DELAY)
-
-    expect(cookieChange).toEqual('foo')
-  })
-
-  it('should not notify observers on cookie change when the cookie value as not changed when cookieStore is not supported', () => {
-    Object.defineProperty(window, 'cookieStore', { get: () => undefined, configurable: true })
-    const observable = createCookieObservable(mockRumConfiguration(), COOKIE_NAME)
+  it('should not notify observers on cookie change when the cookie value as not changed', () => {
```
sethfowler-datadog, line 40: Nit: looks like there was a typo in the original name of the test:
```suggestion
  it('should not notify observers on cookie change when the cookie value has not changed', () => {
```

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -1,54 +1,25 @@
...
 export type CookieObservable = ReturnType<typeof createCookieObservable>
 
-export function createCookieObservable(configuration: Configuration, cookieName: string) {
-  const detectCookieChangeStrategy = (window as CookieStoreWindow).cookieStore
-    ? listenToCookieStoreChange(configuration)
-    : watchCookieFallback
+export function createCookieObservable(cookieName: string) {
+  // NOTE: we don't use cookiestore.addEventListner('change', handler) as it seems to me more prone to bugs
+  // and cause our tests to be flaky (and probably in production as well)
+  const detectCookieChangeStrategy = watchCookieStrategy
```
sethfowler-datadog, line 13: Nit: you could even remove the indirection of `detectCookieChangeStrategy` here and just directly call `watchCookieStrategy` in the `Observable` callback, since there's no other strategy anymore.

sethfowler-datadog (approved):
Nice!

RomanGaignault (approved):

PR #3459, opened 2025-04-02
[RUM Profiler] Fix collection of profiles after visibility change

thomasbertet:
## Motivation

- Whenever the visibility change to `hidden` (eg. user switch away to another tab), the profiler is paused. 
- Whenever users come back, the profiler is not re-started properly and does not collect profiles anymore.

## Problem

- On pause, we are cleaning-up tasks including the listeners to `DOM_EVENT.VISIBILITY_CHANGE`. 
- When user comes back, nothing is listening to `DOM_EVENT.VISIBILITY_CHANGE` value and thus the profiler is never restarted.

## Changes
- Move "global" listeners outside of the instance (eg. visibility change and unload handlers). Global listeners are not associated to a single instance and now live outside.
- Add unit test.

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [x] End to end

---

### DIFF
--- a/packages/rum/src/domain/profiling/profiler.spec.ts
+++ b/packages/rum/src/domain/profiling/profiler.spec.ts
@@ -1,5 +1,6 @@
 import { LifeCycle } from '@datadog/browser-rum-core'
 import { relativeNow, timeStampNow } from '@datadog/browser-core'
+import { setPageVisibility, restorePageVisibility, createNewEvent } from '@datadog/browser-core/test'
 import { createRumSessionManagerMock, mockPerformanceObserver, mockRumConfiguration } from '../../../../rum-core/test'
 import { mockProfiler } from '../../../test'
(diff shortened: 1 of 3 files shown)


amortemousque:
Naive question: Why are we pausing the profiler when the page is hidden?

thomasbertet:
> Naive question: Why are we pausing the profiler when the page is hidden?

My naive response would be to save resources altogether... but to be fair, I'm not so sure if there is such benefits. We have a bunch of listeners that would probably not be called anyway. On the profiler side, I'm unsure how the JS Self-Profiling API behaves when the page is hidden. The [spec](https://wicg.github.io/js-self-profiling/) does not really mention it.

Do you see usefulness for running the Profiler while the page is hidden ? Otherwise I'd think we can keep the pause mecanism for now.

amortemousque (approved):

PR #3480, opened 2025-04-14
♻️trackCumulativeLayoutShift: Sort sources in attribution data before picking the 1st one

anediaz:
## Motivation
We’ve recently identified an issue on how the Browser SDK reports the shifted element data for CLS.

The way it is reported today, from the list of elements having contributed to the biggest shift, always the first one is returned by the Browser SDK, while the list is not sorted. 
From what we can see in the CLS Inspect page, we suspect this is not indeed the most impactful element:
image

In the Optimize Inspect page, this can lead us to confusing situations, where for a significant CLS, the element we’re showing as the most relevant one is considerably small (comparing to the CLS score).

By using Chrome developer extension's "Performance" tab, we confirm the order of sources in the attribution list corresponds to how elements are painted on the screen. Though we are sure that we have the top 5 elements with greatest impact areas, we are certain that it is not guaranteed they will be sorted by the size of the impact area. 

Take a look at [CLS attribution and score under the hood](https://datadoghq.atlassian.net/wiki/spaces/~712020c333b3f5accb48879f3dc07a036dc63d/pages/4953804246) for more information

## Changes
- Created a new method `getClsSortedSources` to sort sources by their impacted area in descending order
image

## Testing

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
```diff
@@ -115,7 +116,8 @@ export function trackCumulativeLayoutShift(
 function getFirstElementAttribution(
```
amortemousque, line 116: **🥜 nitpick:** ‏Maybe you could rename this function to better describe the intent
```suggestion
function getTopImpactedElement(
```

amortemousque (approved):

packages/rum-core/src/domain/view/viewMetrics/trackCumulativeLayoutShift.ts
anediaz: sure! done [here](https://github.com/DataDog/browser-sdk/pull/3480/commits/ecd0299ce3229a064e23bdccfa726fe5b645efa9)

packages/rum-core/src/domain/view/viewMetrics/getClsSortedSources.ts
```diff
@@ -0,0 +1,59 @@
+import type { RumLayoutShiftAttribution } from '../../../browser/performanceObservable'
+
+/**
+ * Calculates the area of a rectangle given its width and height
+ */
+const calculateArea = (width: number, height: number): number => width * height
```
BenoitZugmeyer, line 6: **🥜 nitpick:** ‏not sure if we need a function for this

packages/rum-core/src/domain/view/viewMetrics/getClsSortedSources.ts
```diff
@@ -0,0 +1,59 @@
...
+    source,
+    impactedArea: getImpactedArea(source),
+  }))
+
+  // Sort by impacted area in descending order
+  sourcesWithImpactedAreas.sort((a, b) => b.impactedArea - a.impactedArea)
+
+  // Return just the sorted sources
+  return sourcesWithImpactedAreas.map(({ source }) => source)
+}
```
BenoitZugmeyer, line 59: **💬 suggestion:** I think that:
* we can mutate the array in-place, we don't really need to make a copy
* pre-computing the `impactedArea` is premature optimization, and probably isn't worth it as the sources array is likely to be small (< 10 items)

With that in mind, we could simplify the function:

```ts
export function sortClsSources(sources: RumLayoutShiftAttribution[]) {
  sources.sort((a, b) => getImpactedArea(b) - getImpactedArea(a))
}
```

Although after reflexion we don't even need to sort the array, just get the source with the bigest area and a source.node element. So it could be something like:

```ts
function getTopImpactedElement(
  sources: RumLayoutShiftAttribution[]
) {
  let topImpactedSource: (RumLayoutShiftAttribution & { node: Element }) | undefined
  for (const source of sources) {
    if (
      source.node &&
        isElementNode(source.node) &&
        (!topImpactedSource || getImpactedArea(topImpactedSource) < getImpactedArea(source))
      ) {
        topImpactedSource = source
    }
  }
  return topImpactedSource
}
```

BenoitZugmeyer (approved):

Issue #3479, opened 2025-04-14
🐛 Safari crash reports missing Wasm frames in stack traces


kateinoigakukun:
**Describe the bug**

We are observing that WebAssembly (Wasm) crash reports from Safari have incomplete stack trace information when sent via the Datadog Browser SDK. This affects both `@datadog/browser-logs` and `@datadog/browser-rum`.

After a closer look, it seems that [the stack trace parser](https://github.com/DataDog/browser-sdk/blob/4266bcd42b3a3c020658ef7578c9b52e7b8def57/packages/core/src/tools/stackTrace/computeStackTrace.ts#L29) fails to parse Wasm frames and silently discards them. I couldn’t find [any test cases](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/stackTrace/computeStackTrace.spec.ts) that cover Wasm stack traces, so it may currently be out of awareness.

**To Reproduce**

You can serve and open an HTML on Safari: https://gist.github.com/kateinoigakukun/282b3e30ffae6c63751ea9b3af80e64b

| Safari | Chrome | Firefox | 
|:--:|:---:|:---:|
| <img width="1005" alt="Image" src="https://github.com/user-attachments/assets/0331ef67-d79d-4f9e-b7cb-b317f57f4165" /> | <img width="1049" alt="Image" src="https://github.com/user-attachments/assets/80b8e155-b4c9-4f53-b6d2-676fd28db3b3" /> | <img width="1009" alt="Image" src="https://github.com/user-attachments/assets/951de510-944d-4911-a79a-d7ee349cb2d6" /> |

**Expected behavior**

Wasm stack frames should be included in crash reports, especially on Safari which surfaces them natively.

RomanGaignault:
Hello @kateinoigakukun,
Thanks for your feedback ! We will look into it and keep you updated

kateinoigakukun:
Thank you for fixing this! Do you have a plan to have a patch release including this?

BenoitZugmeyer:
Release is coming (if not today, probably tomorrow)

BenoitZugmeyer:
It's released in v6.6.4!

PR #3481, opened 2025-04-14
🐛 [RUM-9465] Add support for parsing Safari Wasm stack trace

RomanGaignault:
## Motivation

Add support for parsing correctly stack trace in a safari wasm context. Cf https://github.com/DataDog/browser-sdk/issues/3479

## Changes

Modify computeStackTrace function / add unit test

## Testing

- [ ] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
+++ b/packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
@@ -90,11 +90,7 @@ Error: foo
   })
 
   it('should handle a native error object stack from Chrome', () => {
-    const mockErr = {
-      message: 'foo',
-      name: 'Error',
-      stack: stackStr,
-    }
+    const mockErr = { message: 'foo', name: 'Error', stack: stackStr }
     const stackFrames = computeStackTrace(mockErr)
 
     expect(stackFrames.stack[0].url).toEqual('<anonymous>')
@@ -173,20 +169,8 @@ Error: foo
     const stackFrames = computeStackTrace(CapturedExceptions.SAFARI_7)
 
     expect(stackFrames.stack.length).toEqual(3)
-    expect(stackFrames.stack[0]).toEqual({
-      args: [],
-      column: 22,
-      func: '?',
-      line: 48,
-      url: 'http://path/to/file.js',
-    })
-    expect(stackFrames.stack[1]).toEqual({
-      args: [],
-      column: 15,
-      func: 'foo',
-      line: 52,
-      url: 'http://path/to/file.js',
-    })
+    expect(stackFrames.stack[0]).toEqual({ args: [], column: 22, func: '?', line: 48, url: 'http://path/to/file.js' })
+    expect(stackFrames.stack[1]).toEqual({ args: [], column: 15, func: 'foo', line: 52, url: 'http://path/to/file.js' })
     expect(stackFrames.stack[2]).toEqual({
       args: [],
       column: 107,
@@ -200,20 +184,8 @@ Error: foo
     const stackFrames = computeStackTrace(CapturedExceptions.SAFARI_8)
 
     expect(stackFrames.stack.length).toEqual(3)
-    expect(stackFrames.stack[0]).toEqual({
-      args: [],
-      column: 22,
-      func: '?',
-      line: 47,
-      url: 'http://path/to/file.js',
-    })
-    expect(stackFrames.stack[1]).toEqual({
-      args: [],
-      column: 15,
-      func: 'foo',
-      line: 52,
-      url: 'http://path/to/file.js',
-    })
+    expect(stackFrames.stack[0]).toEqual({ args: [], column: 22, func: '?', line: 47, url: 'http://path/to/file.js' })
+    expect(stackFrames.stack[1]).toEqual({ args: [], column: 15, func: 'foo', line: 52, url: 'http://path/to/file.js' })
     expect(stackFrames.stack[2]).toEqual({
       args: [],
       column: 23,
@@ -236,13 +208,7 @@ Error: foo
       line: undefined,
       url: '[native code]',
     })
-    expect(stackFrames.stack[1]).toEqual({
-      args: [],
-      column: 21,
-      func: 'foo',
-      line: 58,
-      url: 'http://path/to/file.js',
-    })
+    expect(stackFrames.stack[1]).toEqual({ args: [], column: 21, func: 'foo', line: 58, url: 'http://path/to/file.js' })
     expect(stackFrames.stack[2]).toEqual({
       args: [],
       column: 91,
@@ -393,20 +359,8 @@ Error: foo
     const stackFrames = computeStackTrace(CapturedExceptions.FIREFOX_31)
 
     expect(stackFrames.stack.length).toEqual(3)
-    expect(stackFrames.stack[0]).toEqual({
-      args: [],
-      column: 13,
-      func: 'foo',
-      line: 41,
-      url: 'http://path/to/file.js',
-    })
-    expect(stackFrames.stack[1]).toEqual({
-      args: [],
-      column: 1,
-      func: 'bar',
-      line: 1,
-      url: 'http://path/to/file.js',
-    })
+    expect(stackFrames.stack[0]).toEqual({ args: [], column: 13, func: 'foo', line: 41, url: 'http://path/to/file.js' })
+    expect(stackFrames.stack[1]).toEqual({ args: [], column: 1, func: 'bar', line: 1, url: 'http://path/to/file.js' })
     expect(stackFrames.stack[2]).toEqual({
       args: [],
       column: 1,
@@ -434,13 +388,7 @@ Error: foo
       line: 15,
       url: 'file:///path/to/file.js',
     })
-    expect(stackFrames.stack[2]).toEqual({
-      args: [],
-      column: 3,
-      func: 'bar',
-      line: 20,
-      url: 'file:///path/to/file.js',
-    })
+    expect(stackFrames.stack[2]).toEqual({ args: [], column: 3, func: 'bar', line: 20, url: 'file:///path/to/file.js' })
     expect(stackFrames.stack[3]).toEqual({
       args: [],
       column: 1,
@@ -485,13 +433,7 @@ Error: foo
       line: 128,
       url: 'std_interceptors.js',
     })
-    expect(stackFrames.stack[2]).toEqual({
-      args: [],
-      column: 29,
-      func: 'eval',
-      line: 132,
-      url: 'std_interceptors.jsx',
-    })
+    expect(stackFrames.stack[2]).toEqual({ args: [], column: 29, func: 'eval', line: 132, url: 'std_interceptors.jsx' })
   })
 
   it('should not include error message into stacktrace ', () => {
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -153,6 +153,16 @@ function parseGeckoLine(line: string): StackFrame | undefined {
...
+const SAFARI_WASM_RE = /^(\S+)\[(\S+)\]@\[wasm code\]$/i
+
+function parseSafariWasmLine(line: string): StackFrame | undefined {
+  const match = SAFARI_WASM_RE.exec(line.trim())
+  if (!match) {
+    return
+  }
+  return { func: `${match[1]}.${match[2]}`, url: '[wasm code]', args: [] }
+}
+
```
BenoitZugmeyer, line 165: **suggestion:**  for better or worse, on safari, stack frames are parsed by the `GECKO_LINE_RE`. We could add support for the `[wasm code]` URL by changing that regex:

```diff
 const GECKO_LINE_RE =
-  /^\s*(.*?)(?:\((.*?)\))?(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle)(?::(\d+))?(?::(\d+))?\s*$/i
+  /^\s*(.*?)(?:\((.*?)\))?(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?\s*$/i
```

It's more concise than having a dedicated function for it (although we might want to clean that regex one day :D).

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,36 @@ Error: foo
       url: 'capacitor://localhost/media/dist/bundle.js',
     })
   })
+
+  it('should parse Safari WebAssembly stack frames', () => {
```
BenoitZugmeyer, line 874: **suggestion:** as mentioned during standup, it could be nice to add wasm stack traces tests for Chrome and Firefox as well

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,36 @@ Error: foo
...
+      args: [],
+    })
+
+    expect(stackFrames.stack[2]).toEqual({
+      func: '?',
+      url: 'https://example.com/app.js',
+      line: 150,
+      column: 30,
+      args: [],
+    })
```
BenoitZugmeyer, line 903: **suggestion:** Remove assertions on non-wasm stack frames, as it's unrelated to the test

```suggestion
```

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,36 @@ Error: foo
       url: 'capacitor://localhost/media/dist/bundle.js',
     })
   })
+
+  it('should parse Safari WebAssembly stack frames', () => {
+    const wasmStack = `
+  RuntimeError: Out of bounds memory access
+      myModule.wasm-function[crashHandler]@[wasm code]
```
sethfowler-datadog, line 877: It might be worth including a stack frame that would test the case where no debug info was available for the wasm function. In that situation, I think that you'd get a stack frame that looks roughly like this on Safari:
```
myModule.wasm-function[42]@[wasm code]
```
In other words, you'd have the function's numeric index instead of its name.

I think both your regex and Benoît's proposed regex handle this just fine; it'd just be good to lock in that behavior with a test.

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,65 @@ Error: foo
...
+    `
+    const mockErr = { message: 'Wasm Error', name: 'Error', stack: wasmStack }
+    const stackFrames = computeStackTrace(mockErr)
+
+    expect(stackFrames.stack.length).toBe(2)
+
+    expect(stackFrames.stack[0]).toEqual({
+      args: [],
+      func: 'wasm-function[42]',
+      url: 'http://example.com/my-module.wasm:wasm-function[42]:0x1a3b',
```
sethfowler-datadog, line 937: This is a mis-parse; the URL here should just be:
```suggestion
      url: 'http://example.com/my-module.wasm',
```

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,65 @@ Error: foo
...
+    at bar (http://example.com/script.js:10:5)
+    `
+    const mockErr = { message: 'Wasm Error', name: 'Error', stack: wasmStack }
+    const stackFrames = computeStackTrace(mockErr)
+
+    expect(stackFrames.stack.length).toBe(2)
+    expect(stackFrames.stack[0]).toEqual({
+      args: [],
+      func: 'foo',
+      url: '<anonymous>:wasm-function[42]:0x1a3b',
```
sethfowler-datadog, line 917: Same here; the URL is actually just `<anonymous>`:
```suggestion
      url: '<anonymous>',
```
We might want to represent it as the empty string instead, though.

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,73 @@ Error: foo
...
+    `
+    const mockErr = { message: 'Out of bounds memory access', name: 'RuntimeError', stack: wasmStack }
+    const stackFrames = computeStackTrace(mockErr)
+
+    expect(stackFrames.stack.length).toBe(4)
+
+    expect(stackFrames.stack[0]).toEqual({
+      args: [],
+      func: 'myModule.wasm-function[crashHandler]',
+      url: '[wasm code]',
```
sethfowler-datadog, line 890: We took this example from the reporter's test case, but the reporter's test case did the WASM equivalent of `eval()` (it embedded the WASM file as an `ArrayBuffer` of bytes), so there's no meaningful URL for the WASM module. It'd be nice to test what happens when you load the WASM module from a file, so that it _does_ have a URL. We should test that we parse that URL properly when it's present.

I suspect that `[wasm code]` is the equivalent of `<anonymous>` in the Chrome test case -- it may just be a stand-in for "no URL". It might be better to represent this case at the empty string, but whatever we do, we should align it across browsers so that stack traces from different browsers can be correctly merged.

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,92 @@ Error: foo
       url: 'capacitor://localhost/media/dist/bundle.js',
     })
   })
+
+  it('should parse Safari WebAssembly stack frames', () => {
+    const wasmStack = `
+  RuntimeError: Out of bounds memory access
+      myModule.wasm-function[crashHandler]@[wasm code]
+      myModule.wasm-function[42]@[wasm code]
```
BenoitZugmeyer, line 878: **issue:** this stack frame is probably incorrect, it should be
```suggestion
      <?>.wasm-function[42]@[wasm code]
```

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,92 @@ Error: foo
...
+      url: '[wasm code]',
+      column: undefined,
+      line: undefined,
+    })
+  })
+
+  it('should parse Safari Wasm stack frames from file URL', () => {
+    const wasmStack = `
+  RuntimeError: Some Wasm Error
+      myModule.wasm-function[crashFunc]@http://example.com/myModule.wasm
```
BenoitZugmeyer, line 907: **issue:** this stack frame is probably incorrect, from my test Safari never includes the .wasm URL

```suggestion
      myModule.wasm-function[crashFunc]@[wasm code]
```

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,92 @@ Error: foo
...
+      url: 'http://example.com/myModule.wasm',
+      line: undefined,
+      column: undefined,
+    })
+  })
+
+  it('should parse Chrome WebAssembly stack frames', () => {
+    const wasmStack = `
+  Error: Wasm Error
+    at foo (<anonymous>:wasm-function[42]:0x1a3b)
```
BenoitZugmeyer, line 926: **issue:** this stack frame is probably incorrect, it should be something like:

```suggestion
    at myModule.foo (http://example.com/myModule.wasm:wasm-function[42]:0x1a3b)
```

When compiled without debug symbols, it becomes:

```
  at http://example.com/myModule.wasm:wasm-function[42]:0x1a3b
```

packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
```diff
@@ -1012,4 +870,92 @@ Error: foo
...
+      url: '<anonymous>',
+      line: undefined,
+      column: undefined,
+    })
+  })
+
+  it('should parse Firefox WebAssembly stack frames', () => {
+    const wasmStack = `
+  Error: Wasm Error
+    wasm-function[42]@http://example.com/my-module.wasm:wasm-function[42]:0x1a3b
```
BenoitZugmeyer, line 945: **issue:** this stack frame is probably incorrect, it should be

```
myModule.foo@http://example.com/my-module.wasm:wasm-function[42]:0x1a3b
```

or without debug info:

```
@http://example.com/my-module.wasm:wasm-function[42]:0x1a3b
```

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -183,3 +181,16 @@ function tryToParseMessage(messageObj: unknown) {
...
+function tryToParseWasmUrl(frame: StackFrame): void {
+  if (frame.url) {
+    const wasmUrlMatch = WASM_STANDARD_URL_RE.exec(frame.url)
+    if (wasmUrlMatch) {
+      frame.url = wasmUrlMatch[1]
+      frame.line = undefined
+      frame.column = undefined
+    }
+  }
+}
```
BenoitZugmeyer, line 196: **suggestion:** we shouldn't be too opinionated and avoid hiding information that could be helpful to the customer. In this case:
* the `wasm-function` mention is valuable as it helps to understand the stack frame comes from wasm
* the function index is valuable when there is no debug information to help identify which function was called
* the address is valuable to identify the location of the stack frame within the function

So, I suggest keeping the whole "URL".

BenoitZugmeyer (commented):
I experimented with WASM stack frames and this is what I got:

| Browser | With debug info                                         | Without debug info                     |
| ------- | ------------------------------------------------------- | -------------------------------------- |
| Chrome  | `at MODULE.FUNCTION (URL:wasm-function[INDEX]:ADDRESS)` | `at URL:wasm-function[INDEX]:ADDRESS`  |
| Firefox | `MODULE.FUNCTION@URL:wasm-function[INDEX]:ADDRESS`      | `@URL:wasm-function[INDEX]:ADDRESS`    |
| Safari  | `MODULE.wasm-function[FUNCTION]@[wasm-code]`            | `<?>.wasm-function[INDEX]@[wasm-code]` |

BenoitZugmeyer (approved):

PR #3482, opened 2025-04-14
✨ Allow to retrieve records in the developer-extension Replay tab

yannickadam:
## Motivation

- A simple way to capture Replay Records locally. 

## Changes

- Upgraded the Replay Sandbox to the current latest (0.116.0)
- Added the possibility to force a full snapshot at any time during the session
- Keep a buffer of the record applied, and allow to download them using a minimal `segment` format
- Allow to exclude mouse movements (to lower the number of records stored in the buffer)
image

## Testing

To test - you need to install the developer extension unpacked (as per readme) and try it locally. 

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---


developer-extension/src/panel/components/tabs/replayTab.tsx
```diff
@@ -35,13 +36,74 @@ export function ReplayTab() {
...
+
+  useEffect(() => {
+    startSessionReplayPlayer(
+      frameRef.current!,
+      setPlayerStatus,
+      setRecordCount,
+      (getRecordsFn) => setGetRecords(() => getRecordsFn),
+      excludeMouseMovementsRef
+    )
   }, [])
```
BenoitZugmeyer, line 57: **💬 suggestion:** ‏
* store the record count in the player status so you just have one state to maintain
* store the result of `startSessionReplayPlayer` in a ref
* use it directly
* remove all of those dirty hooks


```ts
const [playerState, setPlayerState] = useState<SessionReplayPlayerState>({ status: 'loading', recordCount: 0 })
const playerRef = useRef(null)

useEffect(() => {
  playerRef.current = startSessionReplayPlayer(
    frameRef.current!,
    setPlayerStatus,
  )
}, [])
```

then

```ts
<Checkbox
              label="Exclude mouse movements"
              checked={excludeMouseMovements}
              onChange={(event) => playerRef.current.setExcludeMouseMovements(event.currentTarget.checked)}
 />
```

```ts
const downloadRecords = () => {
    if (!getRecords) {
      return
    }

    const records = playerRef.current.getRecords()
```
etc.

developer-extension/src/panel/sessionReplayPlayer/startSessionReplayPlayer.ts
```diff
@@ -12,19 +12,15 @@ export type SessionReplayPlayerStatus = 'loading' | 'waiting-for-full-snapshot'
 const sandboxOrigin = 'https://session-replay-datadoghq.com'
 // To follow web-ui development, this version will need to be manually updated from time to time.
 // When doing that, be sure to update types and implement any protocol changes.
-const sandboxVersion = '0.68.0'
+const sandboxVersion = '0.116.0'
```
BenoitZugmeyer, line 15: **👏 praise:** ‏thanks!

developer-extension/src/panel/components/tabs/replayTab.tsx
yannickadam: 🫡

BenoitZugmeyer (approved):

PR #3488, opened 2025-04-17
✨ Add publicPath configuration to webpack setups

yannickadam:
## Motivation

When using the `Use development bundles` feature of the developer extension, with the `On NPM setup` experimental option, I would receive the following message from the application (Chrome 135):
Pasted Graphic

## Changes

- Force the publicPath in webpack config to be `http://localhost:8080` when in development mode. 
The browser SDK extension now shows the version as `dev` as expected, however, the following warning is issued:
Pasted Graphic 1
Functions updated in the development SDK seems to be called properly.

## Testing

Take the PR locally, run `yarn dev` and find a website using NPM to bundle the SDK. Try to override with your local bundles using: 
image

- [x] Local
- [ ] Staging
- [ ] Unit
- [ ] End to end

---

### DIFF
--- a/webpack.base.js
+++ b/webpack.base.js
@@ -5,8 +5,7 @@ const TerserPlugin = require('terser-webpack-plugin')
 const { buildEnvKeys, getBuildEnvValue } = require('./scripts/lib/buildEnv')
 
 const tsconfigPath = path.join(__dirname, 'tsconfig.webpack.json')
-
-module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins }) => ({
+module.exports = ({ entry, mode, publicPath, filename, types, keepBuildEnvVariables, plugins }) => ({
   entry,
   mode,
   output: {
@@ -19,6 +18,8 @@ module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins
         : // Include a content hash in chunk names in production.
           `chunks/[name]-[contenthash]-${filename}`,
     path: path.resolve('./bundle'),
+    // Set explicit publicPath for development mode when using the developer extension with NPM package override
+    publicPath: mode === 'development' ? publicPath : 'auto',
   },
   target: ['web', 'es2018'],
   devtool: false,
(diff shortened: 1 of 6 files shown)


webpack.base.js
```diff
@@ -5,8 +5,7 @@ const TerserPlugin = require('terser-webpack-plugin')
 const { buildEnvKeys, getBuildEnvValue } = require('./scripts/lib/buildEnv')
 
 const tsconfigPath = path.join(__dirname, 'tsconfig.webpack.json')
-
-module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins }) => ({
+module.exports = ({ entry, mode, publicPath, filename, types, keepBuildEnvVariables, plugins }) => ({
```
BenoitZugmeyer, line 8: **suggestion:** to simplify a bit, could you pass `argv` to this function?

```suggestion
module.exports = ({ entry, argv, filename, types, keepBuildEnvVariables, plugins }) => ({
```

webpack.base.js
```diff
@@ -19,6 +18,8 @@ module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins
         : // Include a content hash in chunk names in production.
           `chunks/[name]-[contenthash]-${filename}`,
     path: path.resolve('./bundle'),
+    // Set explicit publicPath for development mode when using the developer extension with NPM package override
+    publicPath: mode === 'development' ? publicPath : 'auto',
```
BenoitZugmeyer, line 22: **suggestion:** I don't think we need this condition

```suggestion
    publicPath: argv.publicPath,
```

Also, the comment could be moved to `dev-server.js`

webpack.base.js
yannickadam: I did in latest commit. However, I feel that we gained readability in some places, lost in others.

yannickadam: I removed the previous commit - in the end, this introduced more noise in karma config, react-app config, ...

BenoitZugmeyer (approved):

PR #3489, opened 2025-04-17
♻️ [RUM-9549] Decouple the session context from assembly using hooks

amortemousque:
## Motivation

Introduce a new session context module and use assembly hooks. 
Add the ability to discard an event from the assembly hooks.

## Changes

- Moved `session.id`, `session.type`, `session.has_replay`, `session.sampled_for_replay` from `assembly` to `sessionContext`
- Moved `session.has_replay`, `session.is_active` from `viewCollection` to `sessionContext`
- Move discard logic into their respective context modules (sessionContext, urlContext, viewCollection)

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


packages/rum-core/src/domain/contexts/sessionContext.ts
```diff
@@ -0,0 +1,45 @@
...
+  recorderApi: RecorderApi,
+  viewHistory: ViewHistory
+) {
+  hooks.register(HookNames.Assemble, ({ eventType, startTime }): PartialRumEvent | undefined => {
+    const session = sessionManager.findTrackedSession(startTime)
+    const view = viewHistory.findView(startTime)
+
+    if (!session || !view) {
+      return
+    }
```
BenoitZugmeyer, line 21: **thought:** without the properties returned by this hook, the event would be invalid because it's missing required fields. I understand this can't happen now because we have the `if (session && viewHistoryEntry && urlContext) {` condition in assembly.ts, but this might not be the most explicit.

Do you have some thoughts about this?

* Maybe we could have the hook returns like `false` to prevent sending the event?
* Maybe the `view` and `session` history entries could be passed as hook parameters, so we know it's defined (and we don't need to retrieve them again)?

At least we should document this.

BenoitZugmeyer (commented):
Kind of unrelated to your PR, but Do you think `urlContext` in `assembly.ts` is still needed?

packages/rum-core/src/domain/contexts/sessionContext.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    isRecordingSpy.and.returnValue(false)
+    const eventWithoutHasReplay = hooks.triggerHook(HookNames.Assemble, {
+      eventType: 'action',
+      startTime: 0 as RelativeTime,
+    })
+
+    expect(getReplayStatsSpy).not.toHaveBeenCalled()
+    expect(isRecordingSpy).toHaveBeenCalled()
+    expect(eventWithHasReplay?.session?.has_replay).toEqual(true)
+    expect(eventWithoutHasReplay?.session?.has_replay).toBeUndefined()
```
BenoitZugmeyer, line 72: **suggestion:** using `!.` would make the value we are testing more explicit (same for other `?.` usages below)
 
```suggestion
    expect(eventWithHasReplay!.session!.has_replay).toEqual(true)
    expect(eventWithoutHasReplay!.session!.has_replay).toBeUndefined()
```

packages/rum-core/src/domain/contexts/sessionContext.ts
amortemousque: Yes, good point. I’ve been thinking about it too. I don’t have a design I’m fully happy with yet.

- For option one, I wanted the assembly hook to only enrich events (not discard), to keep its responsibility limited.
- About your option two: having view and session history entries as parameters means we are enriching the data with hook parameters. Originally, I wanted the hook to decouple the context (session and view) from the assembly, so this approach defeats that purpose.

BenoitZugmeyer (approved):

PR #3498, opened 2025-04-23
♻️ [RUM-9614] Decouple default context from assembly using hooks

amortemousque:
## Motivation

Decouple the default context from the assembly. The 'default context' includes the following data:

```ts
 _dd: {
  format_version: 2,
  drift: currentDrift(),
  configuration: {
    session_sample_rate: round(configuration.sessionSampleRate, 3),
    session_replay_sample_rate: round(configuration.sessionReplaySampleRate, 3),
  },
  browser_sdk_version: canUseEventBridge() ? __BUILD_ENV__SDK_VERSION__ : undefined,
},
application: {
  id: configuration.applicationId,
},
date: timeStampNow(),
source: 'browser',
```

## Changes

- Create a defaultContext module that uses assemble hook
- Remove default context from assembly

## Testing

- [x] Local
- [x] Staging
- [x] Unit
- [ ] End to end

---


BenoitZugmeyer (approved):
Amazing

thomas-lebeau (approved):

packages/rum-core/src/domain/assembly.ts
```diff
@@ -110,39 +103,17 @@ export function startRumAssembly(
...
-        date: timeStampNow(),
-        source: 'browser',
-      }
-
       const assembledEvent = hooks.triggerHook(HookNames.Assemble, {
         eventType: rawRumEvent.type,
         startTime,
         duration,
-      })
+      })!
```
bcaudan, line 110: **❓ question:** ‏  IIUC, we need to do that because the return type is `PartialRumEvent | undefined | DISCARDED`. However, we don't seem to have a case where the assemble returns `undefined` 🤔 
Am I missing something or could we remove the `undefined` from the return type and this `!`?

amortemousque: You get it right!  In fact I'm hesitating. At this point in the code, `triggerHook` should always return `PartialRumEvent | DISCARDED`.
However, if hooks haven't been registered correctly, we might get `undefined`.
Maybe I can add something like this instead:
```ts
if (assembledEvent === undefined) {
  // Should never happen as a assembly hook must always return the required fields
  addTelemetryError("Assembly hook misconfigured")
  return
}
```
Wdyt?

bcaudan: If it is not register correctly, should it not get caught by the type-checking?
If it should never happen, I'd be in favor of not handling it and if against all odds it happens we should see a telemetry error and see what to do from here.
If you see some cases where it can legitimately happen, it could be nice to prevent them.

packages/rum-core/src/hooks.ts
```diff
@@ -54,18 +58,22 @@ export function createHooks() {
...
+          return DISCARDED
+        }
+        if (result === SKIPPED) {
+          continue
         }
         results.push(result)
       }
-      return combine(...(results as unknown as [object, object])) as ReturnType<HookCallbackMap[K]>
+
+      return (results.length > 0 ? combine(...(results as unknown as [object, object])) : {}) as PartialRumEvent
```
BenoitZugmeyer, line 76: **thought:**  The non-null assertion we wanted to avoid seemed better than this `{} as PartialRumEvent` cast.

With the former, we acknowledge‏ that although `triggerHook` can return `undefined` in theory, it can't happen in `assembly.ts`. We also assume the `PartialRumEvent` is in fact a valid RUM event with all the required properties defined.

With the latter, we return an empty object, which is not only incorrect from a type perspective (`PartialRumEvent` expect to have a `type` property), but also doesn't improve the situation in `assembly.ts` as we'll still need to implicitly assume that it won't return an empty object.

BenoitZugmeyer (approved):

packages/rum-core/src/hooks.ts
```diff
@@ -54,18 +58,24 @@ export function createHooks() {
...
+        if (result === SKIPPED) {
+          continue
         }
         results.push(result)
       }
-      return combine(...(results as unknown as [object, object])) as ReturnType<HookCallbackMap[K]>
+
+      if (results.length !== 0) {
+        return combine(...(results as unknown as [object, object])) as DefaultRumEventAttributes
+      }
```
bcaudan, line 78: **💬 suggestion:** ‏the check seems superfluous, we could also add a unit test for the no registered hooks case
```suggestion
      return combine(...(results as unknown as [object, object])) as DefaultRumEventAttributes
```

bcaudan (approved):

PR #3499, opened 2025-04-24
♻️ [RUM-9667] refactor stack trace computation

BenoitZugmeyer:
## Motivation

Small refactors around errors to reduce maintenance, reduce bundle size, and potentially simplify generating error events from the `rum-react` plugin (cf "Enable web targets of mobile platforms" project)

## Changes

* compute the error stack trace in `computeRawError`, so we don’t have to compute it in most places
* use `computeRawError` everywhere (the only place missing was in `consoleObservable`)
* make sure we don’t generate unusable stack traces (from onerror)

## Test instructions

As this is mostly a refactor, the functional changes are limited, although we should have the error type for errors collected from console.error now.

Using `yarn dev` sandbox:

```js
DD_RUM.init({
  clientToken: 'xxx',
  applicationId: 'xxx',
})
DD_LOGS.init({
  clientToken: 'xxx',
  forwardErrorsToLogs: true,
})

console.error('This is an error message')
// should generate a RUM error with:
// * error.message "This is an error message"
// * no error.stack or error.type
// should generate a Logs error with:
// * message "This is an error message"
// * no error.stack or error.kind

console.error('foo', new Error('Error message'))
// should generate a RUM error with:
// * error.message "foo Error: Error message"
// * a error.stack
// * error.type "Error"
// should generate a Logs error with:
// * message "foo Error: Error message"
// * a error.stack
// * error.kind "Error"

setTimeout(() => {
  throw 'toto'
})
// should generate a RUM error with:
// * error.message 'Uncaught "toto"'
// * a error.stack
// * no error.type
// should generate a Logs error with:
// * message 'Uncaught "toto"'
// * a error.stack
// * no error.kind

```

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/core/src/domain/console/consoleObservable.ts
+++ b/packages/core/src/domain/console/consoleObservable.ts
@@ -1,13 +1,13 @@
-import { flattenErrorCauses, isError, tryToGetFingerprint, tryToGetErrorContext } from '../error/error'
+import { isError, computeRawError } from '../error/error'
 import { mergeObservables, Observable } from '../../tools/observable'
 import { ConsoleApiName, globalConsole } from '../../tools/display'
 import { callMonitored } from '../../tools/monitor'
 import { sanitize } from '../../tools/serialisation/sanitize'
 import { jsonStringify } from '../../tools/serialisation/jsonStringify'
 import type { RawError } from '../error/error.types'
-import { ErrorHandling, ErrorSource } from '../error/error.types'
+import { ErrorHandling, ErrorSource, NonErrorPrefix } from '../error/error.types'
 import { computeStackTrace } from '../../tools/stackTrace/computeStackTrace'
-import { createHandlingStack, toStackTraceString, formatErrorMessage } from '../../tools/stackTrace/handlingStack'
+import { createHandlingStack, formatErrorMessage } from '../../tools/stackTrace/handlingStack'
 import { clocksNow } from '../../tools/utils/timeUtils'
 
 export type ConsoleLog = NonErrorConsoleLog | ErrorConsoleLog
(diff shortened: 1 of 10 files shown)


packages/core/src/domain/console/consoleObservable.ts
```diff
@@ -70,31 +70,41 @@ function createConsoleObservable(api: ConsoleApiName) {
...
       startClocks: clocksNow(),
-      message,
       source: ErrorSource.CONSOLE,
       handling: ErrorHandling.HANDLED,
+      nonErrorPrefix: NonErrorPrefix.PROVIDED,
+
+      // if no good stack is computed from the error, let's not use the fallback stack message
+      // advising the user to use an instance of Error, as console.error is commonly used without an
+      // Error instance.
+      useFallbackStack: false,
```
amortemousque, line 88: **❓ question:** Why not recommend using an instance of Error? While it's common not to, I believe it's good practice, especially if they want to take full advantage of RUM features.

packages/core/src/domain/error/trackRuntimeError.ts
```diff
@@ -35,19 +35,15 @@ export function trackRuntimeError(errorObservable: Observable<RawError>) {
 export function instrumentOnError(callback: UnhandledErrorCallback) {
   return instrumentMethod(window, 'onerror', ({ parameters: [messageObj, url, line, column, errorObj] }) => {
     let stackTrace
-    if (isError(errorObj)) {
-      stackTrace = computeStackTrace(errorObj)
-    } else {
+    if (!isError(errorObj)) {
       stackTrace = computeStackTraceFromOnErrorMessage(messageObj, url, line, column)
```
amortemousque, line 39: **❓ question:** ‏Not directly related to the PR but do we still need `computeStackTraceFromOnErrorMessage` now that we've dropped IE?

packages/core/src/domain/error/error.ts
amortemousque, line 1: Nit: We could inline like this:

```ts

export function computeRawError({
  stackTrace,
  originalError,
  handlingStack,
  componentStack,
  startClocks,
  nonErrorPrefix,
  useFallbackStack = true,
  source,
  handling,
}: RawErrorParams): RawError {
  const isErrorInstance = isError(originalError)
  if (!stackTrace && isErrorInstance) {
    stackTrace = computeStackTrace(originalError)
  }

  return {
    startClocks,
    source,
    handling,
    handlingStack,
    componentStack,
    originalError,
    type: stackTrace?.name,
    message: computeMessage(stackTrace, isErrorInstance, nonErrorPrefix, originalError),
    stack: stackTrace ? toStackTraceString(stackTrace) : useFallbackStack ? NO_ERROR_STACK_PRESENT_MESSAGE : undefined,
    causes: isErrorInstance ? flattenErrorCauses(originalError as ErrorWithCause, source) : undefined,
    fingerprint: tryToGetFingerprint(originalError),
    context: tryToGetErrorContext(originalError),
  }
}
```

packages/core/src/domain/console/consoleObservable.ts
BenoitZugmeyer: Because there is not always an error to be logged. Sometimes you just want to display an error message.

In any case, I don't want to change the behavior in this PR

packages/core/src/domain/error/trackRuntimeError.ts
BenoitZugmeyer: This is not specific to IE, it works in all browsers. We (always?) get the location via extra onerror parameters, even if it's not an error instance, which is pretty nice!

amortemousque: Same for the `unknown` messageObj we try to parse (`tryToParseMessage`)?

amortemousque (approved):

packages/core/src/domain/error/trackRuntimeError.ts
BenoitZugmeyer: This I'm not sure. I don't know how to reproduce this with actual errors. [This is](https://github.com/csnover/TraceKit/commit/31142a1e0c5802b0c4a91be93ef556f1ca591ebd) the commit that introduced it in TraceKit. The tests are [explicitly calling `onerror`](https://github.com/csnover/TraceKit/commit/31142a1e0c5802b0c4a91be93ef556f1ca591ebd#diff-e776927ac8f426eced5cb25f81448322ca37a3e2c8c453725df237d54647e077R166) with a crafted error message and without 5th argument, I'm not sure if this browser behavior is still relevant or not.

PR #3510, opened 2025-04-30
[RUM Profiler] Use clocks for start/end time

thomasbertet:
## Motivation

We are not using the exact same clocks when collecting long tasks (& LoAF) in both the Browser SDK and the RUM Profiler. 
Let's have the RUM Profiler use the same clocks.

## Changes

- Relative time should now use `RelativeTime` types.
- Make use of clocks helpers to get corrected relative and absolute time.
- Save the drift for debugging purposes.

## Test instructions

- No behavior change expected. Only start/end time should be slightly updated, as they now take the drift into account.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -296,7 +296,7 @@ export function createRumProfiler(
 
       // Store Long Task entry, which is a lightweight version of the PerformanceEntry
       instance.longTasks.push({
-        id: getLongTaskId({ startClocks }),
+        id: getLongTaskId(startClocks.relative),
         duration: entry.duration as Duration,
         entryType: entry.entryType,
         startClocks,
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -1,3 +1,4 @@
+import type { RelativeTime } from '@datadog/browser-core'
 import {
   addEventListener,
   clearTimeout,
@@ -6,6 +7,8 @@ import {
   monitorError,
   display,
   getGlobalObject,
+  relativeNow,
+  relativeToClocks,
 } from '@datadog/browser-core'
 
 import type { LifeCycle, RumConfiguration, RumSessionManager, ViewHistoryEntry } from '@datadog/browser-rum-core'
@@ -168,7 +171,7 @@ export function createRumProfiler(
     // Kick-off the new instance
     instance = {
       state: 'running',
-      startTime: performance.now(),
+      startTime: relativeNow(),
       profiler,
       timeoutId: setTimeout(startNextProfilerInstance, profilerConfiguration.collectIntervalMs),
       longTasks: [],
@@ -206,7 +209,7 @@ export function createRumProfiler(
     await lastInstance.profiler
       .stop()
       .then((trace) => {
-        const endTime = performance.now()
+        const endTime = relativeNow()
 
         const hasLongTasks = longTasks.length > 0
         const isBelowDurationThreshold = endTime - startTime < profilerConfiguration.minProfileDurationMs
@@ -291,7 +294,7 @@ export function createRumProfiler(
         id: getLongTaskId(entry),
         duration: entry.duration,
         entryType: entry.entryType,
-        startTime: entry.startTime,
+        startTime: relativeToClocks(entry.startTime as RelativeTime).relative,
       })
     }
   }
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -127,7 +127,7 @@ export function createRumProfiler(
     // Whenever the View is updated, we add a views entry to the profiler instance.
     const viewUpdatedSubscription = lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (view) => {
       // Note: `view.name` is only filled when users use manual view creation via `startView` method.
-      collectViewEntry({ viewId: view.id, viewName: view.name, startTime: view.startClocks.relative })
+      collectViewEntry({ viewId: view.id, viewName: view.name, startClocks: view.startClocks })
     })
     cleanupTasks.push(viewUpdatedSubscription.unsubscribe)
 
@@ -205,7 +205,7 @@ export function createRumProfiler(
     const { startClocks, longTasks, views } = lastInstance
 
     // Capturing when we stop the profiler so we use this time as a reference to clean-up long task registry, eg. remove the long tasks that we collected already
-    const collectTime = performance.now()
+    const collectClocks = clocksNow()
 
     // Stop current profiler to get trace
     await lastInstance.profiler
@@ -235,7 +235,7 @@ export function createRumProfiler(
         )
 
         // Clear long task registry, remove entries that we collected already (eg. avoid slowly growing memory usage by keeping outdated entries)
-        deleteLongTaskIdsBefore(collectTime)
+        deleteLongTaskIdsBefore(collectClocks)
       })
       .catch(monitorError)
   }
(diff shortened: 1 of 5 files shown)


packages/rum/src/domain/profiling/profiler.ts
```diff
@@ -168,7 +171,7 @@ export function createRumProfiler(
     // Kick-off the new instance
     instance = {
       state: 'running',
-      startTime: performance.now(),
+      startTime: relativeNow(),
```
bcaudan, line 174: **💬 suggestion:** ‏what about using our wrappers for other performance API calls as well? (cf `collectTime` and `timeOrigin`)

packages/rum/src/domain/profiling/types/rumProfiler.types.ts
```diff
@@ -1,10 +1,10 @@
-import type { TimeoutId } from '@datadog/browser-core'
+import type { TimeoutId, RelativeTime } from '@datadog/browser-core'
 import type { ViewHistoryEntry } from '@datadog/browser-rum-core'
 import type { ProfilerTrace, Profiler } from './profilerApi.types'
 
 export interface RumViewEntry {
   /** Detected start time of view */
-  readonly startTime: DOMHighResTimeStamp
+  readonly startTime: RelativeTime
```
bcaudan, line 7: **💬 suggestion:** It may not apply to your use case but IIRC we generally tend to work with `ClocksState` objects for internal interfaces and avoid to use relative time to compute durations.‏

packages/rum/src/domain/profiling/types/rumProfiler.types.ts
```diff
@@ -34,9 +34,9 @@ export interface RumProfilerEnrichmentData {
 
 export interface RumProfilerTrace extends ProfilerTrace, RumProfilerEnrichmentData {
   /** High resolution time when profiler trace started, relative to the profiling session's time origin */
-  readonly startTime: DOMHighResTimeStamp
+  readonly startTime: RelativeTime
   /** High resolution time when profiler trace ended, relative to the profiling session's time origin */
-  readonly endTime: DOMHighResTimeStamp
+  readonly endTime: RelativeTime
   /** Time origin of the profiling session */
   readonly timeOrigin: number
```
bcaudan, line 41: **💬 suggestion:** ‏what about using our date/time types for all time related properties?

packages/rum/src/domain/profiling/transport/transport.ts
```diff
@@ -55,19 +64,22 @@ function buildProfileEvent(
...
+  const profilerStartClocks = relativeToClocks(profilerTrace.startTime)
+  const profilerEndClocks = relativeToClocks(profilerTrace.endTime)
 
   const profileEvent: ProfileEvent = {
     ...profileAttributes,
     attachments: ['wall-time.json'],
-    start: start.toISOString(),
-    end: end.toISOString(),
+    start: new Date(profilerStartClocks.timeStamp).toISOString(),
+    end: new Date(profilerEndClocks.timeStamp).toISOString(),
```
bcaudan, line 74: **🥜 nitpick:** ‏we could extract specific type / helper for this format

packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -6,40 +9,40 @@ type PerformanceEntryStartTimeMs = number
...
-  registry.set(performanceEntryStartTime, longTaskId)
+export function setLongTaskId(longTaskId: string, startTime: RelativeTime) {
+  registry.set(startTime, longTaskId)
 }
 
-export function getLongTaskId(longTaskEntry: { startTime: DOMHighResTimeStamp }): string | undefined {
+export function getLongTaskId(longTaskEntry: { startClocks: ClocksState }): string | undefined {
   // Don't return if it's not enabled or the long task has been reported before the activation
-  if (enabledTime === false || longTaskEntry.startTime < enabledTime) {
+  if (enabledClocks === false || longTaskEntry.startClocks.timeStamp < enabledClocks.timeStamp) {
```
BenoitZugmeyer, line 32: **suggestion:** To compare two dates, using relative time is better as it's monotonic.
```suggestion
  if (enabledClocks === false || longTaskEntry.startClocks.relative < enabledClocks.relative) {
```

packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -6,40 +9,40 @@ type PerformanceEntryStartTimeMs = number
...
   }
 
-  return registry.get(longTaskEntry.startTime)
+  return registry.get(longTaskEntry.startClocks.relative)
 }
 
-export function deleteLongTaskIdsBefore(collectionTime: number) {
-  if (enabledTime === false || collectionTime < enabledTime) {
+export function deleteLongTaskIdsBefore(collectionClocks: ClocksState) {
+  if (enabledClocks === false || collectionClocks.timeStamp < enabledClocks.timeStamp) {
```
BenoitZugmeyer, line 40: **suggestion:** 
```suggestion
  if (enabledClocks === false || collectionClocks.relative < enabledClocks.relative) {
```

packages/rum/src/domain/profiling/types/rumProfiler.types.ts
```diff
@@ -1,10 +1,10 @@
-import type { TimeoutId } from '@datadog/browser-core'
+import type { TimeoutId, ClocksState, Duration } from '@datadog/browser-core'
 import type { ViewHistoryEntry } from '@datadog/browser-rum-core'
 import type { ProfilerTrace, Profiler } from './profilerApi.types'
 
 export interface RumViewEntry {
   /** Detected start time of view */
-  readonly startTime: DOMHighResTimeStamp
+  readonly startClocks: ClocksState
```
BenoitZugmeyer, line 7: **thought:** so you're exposing both clocks to the backend. We usually consider this implementation details and send only one of the two. But I guess it's fine!

packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -6,40 +9,40 @@ type PerformanceEntryStartTimeMs = number
...
  * Store the long task ID in the registry for the Profiler to link it with the corresponding Profile.
  */
-export function setLongTaskId(longTaskId: string, performanceEntryStartTime: number) {
-  registry.set(performanceEntryStartTime, longTaskId)
+export function setLongTaskId(longTaskId: string, startTime: RelativeTime) {
+  registry.set(startTime, longTaskId)
 }
 
-export function getLongTaskId(longTaskEntry: { startTime: DOMHighResTimeStamp }): string | undefined {
+export function getLongTaskId(longTaskEntry: { startClocks: ClocksState }): string | undefined {
```
BenoitZugmeyer, line 30: **nitpick:** I think it would be nicer to have the same argument as `setLongTaskId`, like:

```ts
setLongTaskId(startTime, longTaskId)
getLongTaskId(startTime)
```

packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -1,3 +1,6 @@
+import type { ClocksState, RelativeTime } from '@datadog/browser-core'
+import { clocksNow } from '@datadog/browser-core'
+
 type PerformanceEntryStartTimeMs = number
```
BenoitZugmeyer, line 4: **suggestion:** 
```suggestion
type PerformanceEntryStartTime = RelativeTime
```

packages/rum/src/domain/profiling/types/rumProfiler.types.ts
thomasbertet: Yes I figured it will be useful to have both when we deal with this entry in the FE.

thomasbertet:
@BenoitZugmeyer Thank you for the review! 
I followed your advice and update the logic to use `RelativeTime` and have a nicer API for set/get the ids. Let me know if you'd like other things to change!

BenoitZugmeyer (approved):

PR #3511, opened 2025-04-30
✨ Add developer extension support for replay sandbox local development

sethfowler-datadog:
## Motivation

The development experience for session replay is less than ideal right now. When you're working on a session replay feature that involves changes to both the browser SDK and the session replay sandbox, there's no easy way to preview the effect of those changes on a live web page. This limits the pace at which we can make progress.

## Changes

This PR adds the ability to use the local development version of the session replay sandbox in the Live Replay tab of the developer extension, using an interface similar to the one we use now for overriding the browser SDK bundle. By overriding both the browser SDK bundle and the session replay sandbox simultaneously, you can preview the end-to-end behavior of new session replay features on a live web page.

To support this functional change, I've reworked the UI of the Settings tab of the developer extension a bit. The left column is now called "Overrides", and it contains an accordion UI with three sections: one for the browser SDK, one for the live replay feature, and one for requests to the intake. Here's what the new version of the Settings tab looks like when you first open it:

Screenshot 2025-04-30 at 17 41 22

Here's what the live replay section of the accordion looks like when you expand it:

Screenshot 2025-04-30 at 17 41 42

Here's the intake requests section:

Screenshot 2025-04-30 at 17 57 43

Note that both the browser SDK and live replay sections have an enhanced version of the badge we use to report the presence of the development server. In addition to reporting when the development server is unavailable or available, there's now a new state that clearly indicates when overriding is active. The intake requests section also now has a badge that reports whether requests to the intake are blocked.

When the live replay sandbox is overridden, a small "dev" indicator is also displayed in the upper right corner of the Live Replay tab, as you can see in this screenshot:

Screenshot 2025-04-30 at 17 40 55

I made one other small functional change: instead of explaining in the settings tab that using the RUM Slim bundle is only possible with the network request redirection strategy (i.e., the old "CDN" strategy), I made it work with the other strategy too. The change was trivial, and I think it simplifies the UI slightly to make the choice of how to override the browser SDK bundle separate from the choice of which bundle to replace it with.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/developer-extension/src/panel/components/tabs/replayTab.tsx
+++ b/developer-extension/src/panel/components/tabs/replayTab.tsx
@@ -1,4 +1,4 @@
-import { Button, Flex, Checkbox } from '@mantine/core'
+import { Button, Flex, Checkbox, Badge } from '@mantine/core'
 import React, { useEffect, useRef, useState } from 'react'
 import { TabBase } from '../tabBase'
 import type { SessionReplayPlayerState } from '../../sessionReplayPlayer/startSessionReplayPlayer'
@@ -7,6 +7,7 @@ import { evalInWindow } from '../../evalInWindow'
 import { createLogger } from '../../../common/logger'
 import { Alert } from '../alert'
 import { useSdkInfos } from '../../hooks/useSdkInfos'
+import { useSettings } from '../../hooks/useSettings'
 import * as classes from './replayTab.module.css'
 
 const logger = createLogger('replayTab')
@@ -33,6 +34,7 @@ export function ReplayTab() {
 }
 
 function Player() {
+  const [{ useDevReplaySandbox }] = useSettings()
   const frameRef = useRef<HTMLIFrameElement | null>(null)
   const [playerState, setPlayerState] = useState<SessionReplayPlayerState>({
     status: 'loading',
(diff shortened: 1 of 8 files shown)


developer-extension/src/panel/components/tabs/replayTab.tsx
```diff
@@ -42,8 +44,8 @@ function Player() {
   const playerRef = useRef<ReturnType<typeof startSessionReplayPlayer>>(null)
 
   useEffect(() => {
-    playerRef.current = startSessionReplayPlayer(frameRef.current!, setPlayerState)
-  }, [])
+    playerRef.current = startSessionReplayPlayer(frameRef.current!, setPlayerState, useDevReplaySandbox)
+  }, [useDevReplaySandbox])
```
BenoitZugmeyer, line 48: **suggestion:** stop the previous instance

```suggestion
    playerRef.current = startSessionReplayPlayer(frameRef.current!, setPlayerState, useDevReplaySandbox)
    return () => playerRef.current?.stop()
  }, [useDevReplaySandbox])
```

sethfowler-datadog: Good catch! Fixed.

sethfowler-datadog:
Per internal discussion, I've added a checkbox to the Settings page to control visibility of features that require access to Datadog-internal infrastructure. Since local development of the replay sandbox is such a feature, that part of the accordion is now hidden when that checkbox is unchecked. Here's what the checkbox looks like:

Screenshot 2025-05-05 at 11 45 15

BenoitZugmeyer (approved):

PR #3527, opened 2025-05-13
✨[RUM-260] Support bfcache restore for web vitals

rgaignault:
## Motivation

Add webvitals for bfCache as it is not currently supported in the sdk. Additionally provide a new logic for those view as no events are generated yet. 

## Changes

- Support (FID / LCP / FCP and other metrics for bfCache)
- Create a new view when bfCache is triggered
- Add a new loading_type
- Add a new opt-in parameter called 'trackBfcacheView'

## Checklist

- [x] Tested locally
- [x] Tested on staging


packages/rum-core/src/domain/view/bfCacheSupport.ts
BenoitZugmeyer, line 1: **suggestion:** add unit tests for this module

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -1029,3 +1029,36 @@ describe('service and version', () => {
...
+  })
+
+  it('should create a new "bf_cache" view when restoring from the BFCache', () => {
+    const { getViewCreateCount, getViewEndCount, getViewUpdate, getViewUpdateCount } = viewTest
+
+    expect(getViewCreateCount()).toBe(1)
+    expect(getViewEndCount()).toBe(0)
+
+    const event = createNewEvent('pageshow', { __ddIsTrusted: true }) as PageTransitionEvent
+    Object.defineProperty(event, 'persisted', { value: true })
```
BenoitZugmeyer, line 1056: **suggestion:** 
```suggestion
    const event = createNewEvent('pageshow', { persisted: true })
```

* no need to define `__ddIsTrusted`, it's already set by `createNewEvent`.
* you can just use the second argument to define `persisted`
* adjust `createNewEvent` signature so `'pageshow'` returns a `PageTransitionEvent`, so you don't have to cast it

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -117,6 +119,21 @@ export function trackViews(
...
+      currentView.end()
+      currentView = startNewView(ViewLoadingType.BF_CACHE)
+
+      const { stop: stopBfcacheMetricsTracking } = trackBfcacheMetrics(
+        configuration,
+        pageshowEvent,
+        currentView.initialViewMetrics,
+        currentView.scheduleViewUpdate
+      )
+      currentView.setStopBfcacheMetricsTracking(stopBfcacheMetricsTracking)
```
BenoitZugmeyer, line 133: **suggestion:** move this in `newView`, similarly to `trackInitialViewMetrics`

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  let originalPerformanceNow: () => number
+  let clock: Clock
+  let stopTrackBfcacheMetrics: () => void
+
+  beforeEach(() => {
+    originalRAF = window.requestAnimationFrame
+    window.requestAnimationFrame = (cb: FrameRequestCallback): number => {
+      cb(performance.now())
+      return 0
+    }
```
BenoitZugmeyer, line 20: **suggestion:** use `spyOn(window, 'requestAnimationFrame')`

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    clock = mockClock()
+    originalPerformanceNow = performance.now.bind(performance)
```
BenoitZugmeyer, line 22: **suggestion:** you probably don't need to mock `performance.now()` as it is already mocked by `mockClock()`

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    stopTrackBfcacheMetrics()
+    window.requestAnimationFrame = originalRAF
+    performance.now = originalPerformanceNow
+    clock.cleanup()
+  })
+
+  function createPageshowEvent(timeStamp: number): PageTransitionEvent {
+    const event = createNewEvent('pageshow', { __ddIsTrusted: true }) as PageTransitionEvent
+    Object.defineProperty(event, 'timeStamp', { value: timeStamp })
+    return event
```
BenoitZugmeyer, line 36: **suggestion:** 
```suggestion
    return createNewEvent('pageshow', { timeStamp })
```

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  it('should compute FCP and LCP from the next frame after BFCache restore', () => {
+    const pageshow = createPageshowEvent(100)
+
+    const metrics: InitialViewMetrics = {}
+    const scheduleSpy = jasmine.createSpy('schedule')
+
+    performance.now = () => 150
+
+    const { stop } = trackBfcacheMetrics(mockRumConfiguration(), pageshow, metrics, scheduleSpy)
+    stopTrackBfcacheMetrics = stop
```
BenoitZugmeyer, line 48: **suggestion:** 
```suggestion
    registerCleanupTask(stop)
```

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -117,6 +119,21 @@ export function trackViews(
...
+    })
+  }
+
```
amortemousque, line 136: **💬 suggestion:** ‏Could be moved in `startViewLifeCycle` function

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+// BFCache keeps a full in-memory snapshot of the DOM. When the page is restored, nothing needs to be fetched, so the whole
+// viewport repaints in a single frame. Consequently, LCP almost always equals FCP.
+// (See: https://github.com/GoogleChrome/web-vitals/pull/87)
+
+function measureRestoredPaintTime(pageshowEvent: PageTransitionEvent, callback: (paintTime: Duration) => void): void {
+  requestAnimationFrame(() => {
+    requestAnimationFrame(() => {
```
BenoitZugmeyer, line 41: **suggestion:** could you document why we have those two rAF?

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
```diff
@@ -0,0 +1,60 @@
...
+  measureRestoredPaintTime(pageshowEvent, (paintTime) => {
+    metrics.firstContentfulPaint = paintTime
+    metrics.largestContentfulPaint = { value: paintTime as RelativeTime }
+    scheduleViewUpdate()
+  })
+
+  const { stop: stopMeasureRestoredFID } = measureRestoredFID(configuration, pageshowEvent, ({ delay, time }) => {
+    metrics.firstInput = {
+      delay,
+      time: time as RelativeTime,
```
BenoitZugmeyer, line 24: **question:** I see `time` is first casted to duration then relative time. Is it one or the other?

packages/rum-core/src/domain/view/trackViews.ts
amortemousque, line 133: **💬 suggestion:** Instead of having a separate `trackBfcacheMetrics` module, I would expect each metrics module to handle the bfCache case directly:
- `trackFirstHidden`
- `trackCumulativeLayoutShift`
- `trackLargestContentfulPaint`
- `trackFirstContentfulPaint`

You can propagate `loadingType` to these modules to identify when it’s a bfCache case.

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+  })
+}
+
+function measureRestoredFID(
+  configuration: RumConfiguration,
+  pageshowEvent: PageTransitionEvent,
+  callback: (fid: { delay: Duration; time: Duration }) => void
+): { stop: () => void } {
+  const { stop } = retrieveFirstInputTiming(configuration, (entry) => {
```
BenoitZugmeyer, line 53: **question:** using the polyfill here is unexpected: it's not used as a polyfill but as first class code, so in practice it's not a polyfill anymore.

Do we *need* to report FID for bfcache views?

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
```diff
@@ -6,6 +6,9 @@ import type { LifeCycle } from '../../lifeCycle'
 import { ViewLoadingType } from '../../../rawRumEvent.types'
 import { trackFirstHidden } from './trackFirstHidden'
 
+// For a BFCache-restored view, the regular load event never fires (the document is already in memory).
+// Therefore trackLoadingTime can only emit a loadingTime if waitPageActivityEnd detects some post-restore activity
+// If nothing happens after the pageshow, no candidate is recorded and loadingTime stays undefined.
```
BenoitZugmeyer, line 11: **nitpick:** 
* This also applies to route change views no? It's a bit strange to have a comment focusing on bfcache here. If you want to add documentation here, maybe document how it works in general, then mention bfcache.
* Function comments should be `/**` comments

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
RomanGaignault: time is relative time, I am gonne remove the cast 👌

packages/rum-core/src/domain/view/trackViews.ts
RomanGaignault: Those metrics are computed via initialViewMetrics file, so we would need to go through this file to compute bfCache metrics, I find it a bit more complicated and less clear logic. wdyt

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    return event
+  }
+
+  it('should compute FCP and LCP from the next frame after BFCache restore', () => {
+    const pageshow = createPageshowEvent(100)
+
+    const metrics: InitialViewMetrics = {}
+    const scheduleSpy = jasmine.createSpy('schedule')
+
+    performance.now = () => 150
```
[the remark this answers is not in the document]
RomanGaignault: The trick doesn't seem to work

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
RomanGaignault: From the test the fid is never equal to 0 even if it's way smaller, so It might be interesting to report it

packages/rum-core/src/domain/view/bfCacheSupport.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    const stop = onBFCacheRestore(configuration, callback)
+
+    window.dispatchEvent(createNewEvent('pageshow', { persisted: false } as Partial<PageTransitionEvent>))
+    expect(callback).not.toHaveBeenCalled()
+
+    window.dispatchEvent(createNewEvent('pageshow', { persisted: true } as Partial<PageTransitionEvent>))
+    expect(callback).toHaveBeenCalledTimes(1)
+
+    registerCleanupTask(stop)
```
BenoitZugmeyer, line 18: **suggestion:** move this on line 11, right after when `stop` is defined

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -1029,3 +1029,35 @@ describe('service and version', () => {
...
   })
 })
+
+describe('BFCache views', () => {
+  const lifeCycle = new LifeCycle()
+  let clock: Clock
+  let viewTest: ViewTest
+
+  beforeEach(() => {
+    clock = mockClock()
```
BenoitZugmeyer, line 1039: **suggestion:** it looks like you don't use the clock in your test, maybe remove it?

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
```diff
@@ -0,0 +1,63 @@
...
+ * BFCache keeps a full in-memory snapshot of the DOM. When the page is restored, nothing needs to be fetched, so the whole
+ * viewport repaints in a single frame. Consequently, LCP almost always equals FCP.
+ * (See: https://github.com/GoogleChrome/web-vitals/pull/87)
+ */
+function measureRestoredPaintTime(pageshowEvent: PageTransitionEvent, callback: (paintTime: Duration) => void): void {
+  // Uses two requestAnimationFrame calls to measure FCP after a bfcache restore,
+  // as this gives a more accurate timestamp for the frame following the pageshow event.
+  requestAnimationFrame(() => {
+    requestAnimationFrame(() => {
+      callback(elapsed(pageshowEvent.timeStamp as RelativeTime, relativeNow()))
```
BenoitZugmeyer, line 46: **suggestion:** the timing should be relative to the view start. If you think the view starts at `pageshowEvent.timeStamp`, you should pass a correct `startClocks` to `newView()`, and use those clocks here, instead of the event timestamp (similarly to `trackCommonViewMetrics`)

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -358,6 +389,7 @@ function newView(
       name = updatedName
       triggerViewUpdate()
     },
+    scheduleViewUpdate,
```
BenoitZugmeyer, line 392: **suggestion:** remove this as it's not used anymore.

packages/rum-core/src/domain/view/bfCacheSupport.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  it('should invoke the callback only for BFCache restoration and stop listening when stopped', () => {
+    const configuration = mockRumConfiguration()
+    const callback = jasmine.createSpy('callback')
+
+    const stop = onBFCacheRestore(configuration, callback)
+
+    window.dispatchEvent(createNewEvent('pageshow', { persisted: false } as Partial<PageTransitionEvent>))
+    expect(callback).not.toHaveBeenCalled()
+
+    window.dispatchEvent(createNewEvent('pageshow', { persisted: true } as Partial<PageTransitionEvent>))
```
BenoitZugmeyer, line 15: **suggestion:** those casts are not useful
```suggestion
    window.dispatchEvent(createNewEvent('pageshow', { persisted: false }))
    expect(callback).not.toHaveBeenCalled()

    window.dispatchEvent(createNewEvent('pageshow', { persisted: true }))
```

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+import { noop } from '@datadog/browser-core'
+import type { Clock } from '@datadog/browser-core/test'
+import { createNewEvent, mockClock, registerCleanupTask } from '@datadog/browser-core/test'
+import { mockRumConfiguration } from '../../../../test'
+import { trackBfcacheMetrics } from './trackBfcacheMetrics'
+import type { InitialViewMetrics } from './trackInitialViewMetrics'
+
+describe('trackBfcacheMetrics', () => {
+  let clock: Clock
+  let stopTrackBfcacheMetrics: () => void
```
BenoitZugmeyer, line 11: **suggestion:** `stopTrackBfcacheMetrics` isn't used anymore, you could remove it

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.spec.ts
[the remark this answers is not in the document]
BenoitZugmeyer:
```diff
@@ -24,17 +24,17 @@ describe('trackBfcacheMetrics', () => {
     clock.cleanup()
   })
 
-  function createPageshowEvent(timeStamp: number) {
-    return createNewEvent('pageshow', { timeStamp })
+  function createPageshowEvent() {
+    return createNewEvent('pageshow', { timeStamp: performance.now() })
   }
 
   it('should compute FCP and LCP from the next frame after BFCache restore', () => {
-    const pageshow = createPageshowEvent(100) as PageTransitionEvent
+    const pageshow = createPageshowEvent() as PageTransitionEvent
 
     const metrics: InitialViewMetrics = {}
     const scheduleSpy = jasmine.createSpy('schedule')
 
-    performance.now = () => 150
+    clock.tick(50)
 
     const { stop } = trackBfcacheMetrics(mockRumConfiguration(), pageshow, metrics, scheduleSpy)
     registerCleanupTask(stop)
@@ -48,7 +48,7 @@ describe('trackBfcacheMetrics', () => {
     const metrics: InitialViewMetrics = {}
     const scheduleViewUpdate = jasmine.createSpy('scheduleViewUpdate')
 
-    const pageshow = createPageshowEvent(0) as PageTransitionEvent
+    const pageshow = createPageshowEvent() as PageTransitionEvent
 
     trackBfcacheMetrics(mockRumConfiguration(), pageshow, metrics, scheduleViewUpdate)

```

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -117,7 +119,19 @@ export function trackViews(
     locationChangeSubscription = renewViewOnLocationChange(locationChangeObservable)
   }
 
-  function startNewView(loadingType: ViewLoadingType, startClocks?: ClocksState, viewOptions?: ViewOptions) {
+  if (configuration.trackBfcacheViews) {
```
BenoitZugmeyer, line 122: **thought:** we probably should not track bfcache views when `trackViewsManually` is enabled. WDYT?

packages/rum-core/src/domain/view/viewMetrics/trackBfcacheMetrics.ts
BenoitZugmeyer: Let's discuss about this.

packages/rum-core/src/domain/view/trackViews.ts
rgaignault: Sounds good to me ! I inserted bfCacheView logic to the check that was applied on locationChange

packages/rum-core/src/domain/view/trackViews.ts
amortemousque: If we want to track BFCache metrics like we do in `trackInitialViewMetrics`, then let’s follow the same structure:

```ts

export interface BfCacheViewMetrics {
  firstContentfulPaint?: Duration
  largestContentfulPaint?: LargestContentfulPaint
  firstInput?: FirstInput
}

export function trackBfcacheMetrics(
  viewStart: ClocksState,
  metrics: InitialViewMetrics,
  scheduleViewUpdate: () => void
): { stop: () => void } {
  const bfcacheViewMetrics: BfCacheViewMetrics = {}

  const { stop: stopFCPTracking } = trackRestoredFirstContentfulPaint(
    configuration,
    viewStart.relative,
    (firstContentfulPaint) => {
      bfcacheViewMetrics.firstContentfulPaint = firstContentfulPaint
      scheduleViewUpdate()
    }
  )

  const { stop: stopLCPTracking } = trackRestoredLargestContentfulPaint(
    configuration,
    viewStart.relative,
    (largestContentfulPaint) => {
      bfcacheViewMetrics.largestContentfulPaint = largestContentfulPaint
      scheduleViewUpdate()
    }
  )

  const { stop: stopFCPTracking } = trackRestroredFirstContentfulPaint(
    configuration,
    viewStart.relative,
    (firstContentfulPaint) => {
      initialViewMetrics.firstContentfulPaint = firstContentfulPaint
      scheduleViewUpdate()
    }
  )

  return {
    bfcacheViewMetrics,
    stop: () => {
      stopFCPTracking()
      stopLCPTracking()
      stopMeasureRestoredFID()
      firstHidden.stop()
    },
  }
}
``` 
And:
- Move `trackRestoredFirstContentfulPaint` to the `trackFirstContentfulPaint` file
- Move `trackRestoredLargestContentfulPaint` to the `trackLargestContentfulPaint` file
- Move `trackRestroredFirstContentfulPaint` to the `trackFirstContentfulPaint` file

PR #3529, opened 2025-05-14
👷 add _dd.was_hidden_during_loading to view events

thomas-lebeau:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
@@ -22,7 +22,7 @@ describe('trackLoadingTime', () => {
   let clock: Clock
   let domMutationObservable: Observable<void>
   let windowOpenObservable: Observable<void>
-  let loadingTimeCallback: jasmine.Spy<(loadingTime: Duration) => void>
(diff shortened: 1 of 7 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -127,14 +127,19 @@ describe('trackLoadingTime', () => {
...
   })
 
   it('should discard loading time if page is hidden before activity', () => {
     setPageVisibility('hidden')
     startLoadingTimeTracking()
 
-    expect(loadingTimeCallback).not.toHaveBeenCalled()
+    clock.tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    domMutationObservable.notify()
+    clock.tick(AFTER_PAGE_ACTIVITY_END_DELAY)
```
thomas-lebeau, line 141: If there is no activity, `loadingTimeCallback` is not expected to be called anyway (this is the same test as [line 56](https://github.com/DataDog/browser-sdk/blob/3c1829843c1213b913386335db310b1784bf5741/packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts#L56-L59) and setting the visibility or not has no effect!

BeltranBulbarellaDD (approved):

PR #3530, opened 2025-05-14
✨ Allow overriding service and version for views events in beforeSend

sombrerosheep:
## Motivation

* This change allows you to override the `service` and `version` on`view` events in the `beforeSend` callback.
* This feature exists for errors, resources and actions as part of the original microfrontend support.
* This provides further microfrontend support for changing the view when using custom view tracking is not feasible.

## Changes

* Includes `ROOT_MODIFIABLE_FIELD_PATHS` (`service`/`version`) for `RumEventType.VIEW`.
* Add unit test reassigning `service` and `version` with a `view` event.
* Add e2e test reassigning `service` and `version` with a `view` event.
* I did not yet, but can add this as an experimental feature if desired.
* I am unaware of any other implications this could create but will make updates as the discussion advances.

## Test instructions

* Configure a `beforeSend` hook to update service/version for a view event.
* Unit and e2e tests were added.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/assembly.spec.ts
+++ b/packages/rum-core/src/domain/assembly.spec.ts
@@ -70,7 +70,7 @@ describe('rum assembly', () => {
 
           expect((serverRumEvents[0].view as any).performance.lcp.resource_url).toBe('modified_url')
         })
-        
+
         describe('view type modifyable fields', () => {
           it('service and version should be modifiable', () => {
             const extraConfigurationOptions = { service: 'default service', version: 'default version' }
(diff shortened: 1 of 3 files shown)


sombrerosheep:
@bcaudan - Honestly, one of our problems is trying to work around messy routing logic where manual view tracking would be too monumental at this time. We have 20+ existing UI routes and 6+ microfrontend routes (which is slowly catching up). Turning on  manual view tracking would require us to find and apply every route and is unfortunately not that simple right now.

Obtaining the microfrontend's version is another tricky piece. We originally went with Datadog's guidance to provide the microfrontend's name and version as part of our bundled file names. Since we use our git-sha as part of the version, this had significant impacts on our bundle's caching. We have moved to our microfrontends providing a method so the host application can dynamically get the version. We are using this process to successfully apply the version to actions, errors and resources. Aside from instrumenting all our views manually, the MFE's are not always loaded to obtain this version. Using `beforeSend` helps us make sure we have the value; otherwise the host applications version is used.

We are also sending a number of performance metrics using RUM actions that are more in depth than the [custom performance metrics](https://docs.datadoghq.com/real_user_monitoring/browser/monitoring_page_performance/#create-custom-performance-metrics) feature can help us with. Having those actions associated to an appropriate view will help us using and analyzing that data since those events are all associated back to a view. This gives us better context.

## Usage

We would be using this to match off the URL for specific routes and then reassign the service view. An example is below.

```JavaScript
if (event.type === 'view' && event.view.url.includes('microfrontendRoute')) {
    event.service = 'microfrontendRoute';
    event.version = getMicrofrontendVersion();
}

return true;
```

PR #3537, opened 2025-05-16
♻️ [RUM-9990] Use hooks for Logs SDK to decouple rumInternalContext

amortemousque:
## Motivation

Use hooks for Logs SDK to decouple rumInternalContext

## Changes

- Create `abstractHooks` (in /core) and update all the refs
- Use `hooks` in `rumInternalContext`
- Remove `getRUMInternalContext` from Logs assembly

## Checklist

- [x] Tested locally
- [x] Tested on staging


packages/logs/src/domain/contexts/rumInternalContext.ts
```diff
@@ -15,32 +18,48 @@ interface BrowserWindow {
...
+    if (!rumContext) {
+      if (willSyntheticsInjectRumResult && !logsSentBeforeRumInjectionTelemetryAdded) {
+        logsSentBeforeRumInjectionTelemetryAdded = true
+        addTelemetryDebug('Logs sent before RUM is injected by the synthetics worker', {
+          testId: getSyntheticsTestId(),
+          resultId: getSyntheticsResultId(),
+        })
+      }
+      return
+    }
```
thomas-lebeau, line 48: **💬 suggestion:** ‏ what about using early return to simplify a bit?

```suggestion
    if (rumContext || logsSentBeforeRumInjectionTelemetryAdded || !willSyntheticsInjectRumResult ) {
      return rumContext
    }

    logsSentBeforeRumInjectionTelemetryAdded = true
    addTelemetryDebug('Logs sent before RUM is injected by the synthetics worker', {
      testId: getSyntheticsTestId(),
      resultId: getSyntheticsResultId(),
    })
```

packages/rum-core/src/domain/contexts/connectivityContext.spec.ts
```diff
@@ -1,12 +1,15 @@
...
+import type { Hooks } from '../hooks'
+import { createHooks } from '../hooks'
 import { startConnectivityContext } from './connectivityContext'
 
 describe('startConnectivityContext', () => {
   describe('assemble hook', () => {
-    let hooks: Hooks
+    type NewType = Hooks
+
+    let hooks: NewType
```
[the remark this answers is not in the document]
amortemousque: Probably too much vibe :)

thomas-lebeau (approved):

packages/logs/src/domain/assembly.ts
```diff
@@ -54,7 +67,7 @@ export function startLogsAssembly(
           view: commonContext.view,
         },
         commonContext.context,
-        getRUMInternalContext(startTime),
+        defaultLogsEventAttributes,
```
bcaudan, line 70: **❓ question:** ‏do we plan to move other stuff to the logs assembly? if not, is the "defaultLogsEventAttributes" naming as suited here as in the rum context?

packages/logs/src/domain/contexts/rumInternalContext.spec.ts
```diff
@@ -63,8 +88,12 @@ describe('getRUMInternalContext', () => {
     })
 
     it('adds the telemetry debug event only once', () => {
-      getRUMInternalContext()
-      getRUMInternalContext()
+      hooks.triggerHook(HookNames.Assemble, {
+        startTime: 0 as RelativeTime,
```
bcaudan, line 92: **💬 suggestion:** ‏since `getRumInternalContext` is now exposed from the `startRumInternalContext`, would it worth having a couple of tests also targeting it?

packages/logs/src/domain/assembly.ts
amortemousque: Like for RUM the idea is to use hooks for all attributes in the assembly to only have:
```ts
      const log = combine(defaultLogsEventAttributes, rawLogsEvent)
```
So I think the name is still relevant. Wdyt?

packages/logs/src/domain/contexts/rumInternalContext.spec.ts
amortemousque: Also, once I'll introduce a hook to assemble telemetry event, we will be able to remove `getRUMInternalContext`

packages/logs/src/domain/assembly.ts
bcaudan: ok, it will be relevant when we'll reach that state 👍️

packages/logs/src/domain/contexts/rumInternalContext.spec.ts
amortemousque: I added a test

bcaudan (approved):
Nice!

PR #3549, opened 2025-05-20
🐛 [RUM-10002] Verify the document was not hidden while loading

mormubis:
## Motivation

This PR addresses two issues with the current implementation of loading time when the page is hidden during loading.

**Issue 1**  
During the initial load, we check the `PerformanceObserver` to determine if the page was hidden or visible while loading. We do this because the SDK loads asynchronously, so we only have access to future events from that point onward.

The previous implementation had a problem: the performance observer was accessed even during client-side navigation. 

This condition will always be true, potentially leading to an entry in the observer that is lower than the loading time.

**Issue 2**  
This issue is related to Issue 1. The time compared during client-side navigation was always based on the time origin instead of the loading view. 

This means that if the page undergoes client-side navigation while hidden, the timestamp used in the visibility-change event is based on the time origin rather than the view start reference. As a result, it is almost always higher than expected.

## Changes

This PR changes the `trackFirstHidden` function adding options to only check the performance observer when needed. Additionally, it modifies the condition to check if the page was hidden while loading by replacing the reference point to view start.

| Before | Header |
|--------|--------|
| https://github.com/user-attachments/assets/dbbadff0-422d-49c6-a66a-745e327f4037 | https://github.com/user-attachments/assets/0cad4558-3c9e-415f-9adb-73a02999b0ba | 

## Test instructions

**Case 1**

Using the sandbox, modify the home page to extend the loading time. Like adding a fetch to the home page:

```
fetch('https://upload.wikimedia.org/wikipedia/commons/thumb/3/3f/Fronalpstock_big.jpg/1599px-Fronalpstock_big.jpg')
```

1. Go to http://localhost:8080/react-app/user/42
2. Open dev tools
3. Throttle the network for easy testing
4. Reload the page and quickly change the active tab
5. Go back to the sandbox tab
6. _Check there is no loading time for the current view_
7. Click the home link
8. _Check there is loading time for the home view_

> [!NOTE]
> Previously, this fails as it finds a performance observer entry with a lower time than the loading time.

**Case 2**

Following the same approach as the case above:

1. Go to http://localhost:8080/react-app/user/42
2. Open dev tools
3. Throttle the network for easy testing
7. Click the home link
8. Quickly change the active tab and go back
9. _Check there is **no** loading time for the home view_

> [!NOTE]
> Previously, it reported a wrongful time. 

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.spec.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.spec.ts
@@ -85,7 +85,7 @@ describe('trackFirstHidden', () => {
     })
   })
 
-  xdescribe('using visibilityState entries', () => {
+  describe('using visibilityState entries', () => {
     let originalSupportedEntryTypes: string[] | undefined
     beforeEach(() => {
       performanceBufferMock = mockGlobalPerformanceBuffer()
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.spec.ts
```diff
@@ -113,6 +113,38 @@ describe('trackFirstHidden', () => {
       } as PerformanceEntry)
 
       firstHidden = trackFirstHidden(configuration)
+      expect(firstHidden.timeStamp).toBe(Infinity as RelativeTime)
```
mormubis, line 116: We only need to check the performance observer when we don't have access to events. 

By default, `firstHidden` will only listen to visibility changes and it doesn't check the performance observer.

PR #3551, opened 2025-05-20
👷‍♀️ [RUM-9987] Bump attribute data url limit to 1Mb

cy-moi:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -227,14 +227,15 @@ export function isAllowedRequestUrl(url: string) {
...
-export function isLongDataUrl(url: string): boolean {
-  if (url.length <= MAX_ATTRIBUTE_VALUE_CHAR_LENGTH) {
+export function isLongDataUrl(url: string, maxAttributeLength?: number): boolean {
+  const longUrlLimit = maxAttributeLength || MAX_RESOURCE_VALUE_CHAR_LENGTH
+  if (url.length <= longUrlLimit) {
     return false
   } else if (url.substring(0, 5) === 'data:') {
     // Avoid String.match RangeError: Maximum call stack size exceeded
-    url = url.substring(0, MAX_ATTRIBUTE_VALUE_CHAR_LENGTH)
+    url = url.substring(0, longUrlLimit)
```
bcaudan, line 238: **❓ question:** what is the intent with this line? the local url variable is not used 😕
no test breaks without it
‏

cy-moi: Nice catch. I think it was meant to be in the `sanitizeDataUrl`, where we do `match` 💦 .

sethfowler-datadog (commented):
Suggestion: I think it might be better to straight up _remove_ the limit for image MIME types. If we just bump the limit, we are likely to run into this problem again at some point and spend more resources investigating the issue. Customers will always perceive a truncated image data URL as a bug, so I don't think we ever benefit from doing it. (I do think the limit has value for non-image MIME types, though.)

packages/rum-core/src/domain/resource/resourceUtils.ts
```diff
@@ -227,19 +227,18 @@ export function isAllowedRequestUrl(url: string) {
...
 
 export function sanitizeDataUrl(url: string): string {
-  return `${url.match(DATA_URL_REGEX)![0]}[...]`
+  // Avoid String.match RangeError: Maximum call stack size exceeded
+  const truncatedUrl = url.substring(0, 100).match(DATA_URL_REGEX)
+  if (!truncatedUrl) {
+    return url
+  }
+  return `${truncatedUrl[0]}[...]`
 }
```
bcaudan, line 244: **💬 suggestion:** ‏that seems a bit confusing to me 😕
IIUC, we want to collect only the data type when we have a long data url, so not sure why we need to truncate in this function or when we could not have a match.
No test is covering the `return url` case.

Could it be worth merging the function `isLongDataUrl` inside `sanitizeDataUrl`?
to have something like
```
sanitizeDataUrl(url) {
  if (islongDataUrl) {
    return dataType
  }
  return url
}
```

cy-moi: Mainly I think when the url is too long, regex functions such as `match` will crash. 
> Could it be worth merging the function ...

They were in the same function, but as we have to also check if it is a data url fist, naming it `sanitizeWhenDataUrl` was also a bit confusing. I'm fine with merging them together though.

bcaudan:
**💬 suggestion:** ‏PR title
```diff
- 👷‍♀️ [RUM-9987] Bump attribute data url limit to 1Mb
+ 🐛 [RUM-9987] Bump attribute data url limit to 1Mb
```

PR #3555, opened 2025-05-22
fix: skip unwanted stackframe on custom errors for faulty browsers

leomelki:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/tools/stackTrace/computeStackTrace.ts
+++ b/packages/core/src/tools/stackTrace/computeStackTrace.ts
@@ -48,10 +48,10 @@ export function computeStackTrace(ex: unknown): StackTrace {
     })
   }
 
-  if (stack.length > 0 && BADLY_REPORTING_CUSTOM_ERRORS) { // if we are wrongly reporting custom errors
+  if (stack.length > 0 && WRONGLY_REPORTING_CUSTOM_ERRORS) { // if we are wrongly reporting custom errors
     if (ex instanceof Error && isErrorCustomError(ex)) { // if the element is a custom error
       const firstStackFrame = stack[0];
-      const errorConstructorName = Object.getPrototypeOf(ex).constructor?.name;
+      const errorConstructorName = Object.getPrototypeOf(ex)?.constructor?.name;
       if (firstStackFrame?.func === errorConstructorName) { // if the first stack frame is the custom error constructor
         stack.shift(); // remove it
       }
@@ -207,10 +207,10 @@ function tryToParseMessage(messageObj: unknown) {
 
 function isErrorCustomError(error: Error) {
   let errorProto = Object.getPrototypeOf(error);
-  return errorProto.constructor?.toString().startsWith('class ');
+  return errorProto?.constructor?.toString().startsWith('class ');
 }
 
-function isBadlyReportingCustomErrors() {
+function isWronglyReportingCustomErrors() {
   // Should not be minified during compilation.
   class _DatadogTestCustomError extends Error {
       constructor() {
@@ -226,4 +226,4 @@ function isBadlyReportingCustomErrors() {
   return customErrorStack.includes('_DatadogTestCustomError');
 }
 
-const BADLY_REPORTING_CUSTOM_ERRORS = isBadlyReportingCustomErrors();
+const WRONGLY_REPORTING_CUSTOM_ERRORS = isWronglyReportingCustomErrors();
--- a/packages/core/src/tools/stackTrace/computeStackTrace.ts
+++ b/packages/core/src/tools/stackTrace/computeStackTrace.ts
@@ -48,11 +48,10 @@ export function computeStackTrace(ex: unknown): StackTrace {
     })
   }
 
-  if (stack.length > 0 && BADLY_REPORTING_CUSTOM_ERRORS) { // if we are badly reporting custom errors
+  if (stack.length > 0 && BADLY_REPORTING_CUSTOM_ERRORS) { // if we are wrongly reporting custom errors
     if (ex instanceof Error && isErrorCustomError(ex)) { // if the element is a custom error
       const firstStackFrame = stack[0];
       const errorConstructorName = Object.getPrototypeOf(ex).constructor?.name;
-      console.log("custom error", firstStackFrame?.func, errorConstructorName, firstStackFrame)
       if (firstStackFrame?.func === errorConstructorName) { // if the first stack frame is the custom error constructor
         stack.shift(); // remove it
       }
--- a/packages/core/src/tools/stackTrace/computeStackTrace.ts
+++ b/packages/core/src/tools/stackTrace/computeStackTrace.ts
@@ -48,6 +48,17 @@ export function computeStackTrace(ex: unknown): StackTrace {
     })
   }
 
+  if (stack.length > 0 && BADLY_REPORTING_CUSTOM_ERRORS) { // if we are badly reporting custom errors
+    if (ex instanceof Error && isErrorCustomError(ex)) { // if the element is a custom error
+      const firstStackFrame = stack[0];
+      const errorConstructorName = Object.getPrototypeOf(ex).constructor?.name;
+      console.log("custom error", firstStackFrame?.func, errorConstructorName, firstStackFrame)
+      if (firstStackFrame?.func === errorConstructorName) { // if the first stack frame is the custom error constructor
+        stack.shift(); // remove it
+      }
+    }
+  }
+
   return {
     message: tryToGetString(ex, 'message'),
     name: tryToGetString(ex, 'name'),
@@ -190,3 +201,30 @@ function tryToParseMessage(messageObj: unknown) {
   }
   return { name, message }
 }
+
+// Custom error stacktrace fix 
+// Some browsers (safari/firefox) add the error constructor as a frame in the stacktrace
+// In order to normalize the stacktrace, we need to remove it
+
+function isErrorCustomError(error: Error) {
+  let errorProto = Object.getPrototypeOf(error);
+  return errorProto.constructor?.toString().startsWith('class ');
+}
+
+function isBadlyReportingCustomErrors() {
+  // Should not be minified during compilation.
+  class _DatadogTestCustomError extends Error {
+      constructor() {
+          super();
+          this.name = 'TestError';// different name than the constructor name
+      }
+  }
+
+  let customError = new _DatadogTestCustomError();
+  let customErrorStack = customError.stack?.toString() ?? "";
+
+  // If the stack trace includes the custom error class name, it means that the constructor is added to the stacktrace
+  return customErrorStack.includes('_DatadogTestCustomError');
+}
+
+const BADLY_REPORTING_CUSTOM_ERRORS = isBadlyReportingCustomErrors();
--- a/packages/core/src/tools/stackTrace/computeStackTrace.ts
+++ b/packages/core/src/tools/stackTrace/computeStackTrace.ts
@@ -226,4 +226,8 @@ function isWronglyReportingCustomErrors() {
   return customErrorStack.includes('_DatadogTestCustomError');
 }
 
-const WRONGLY_REPORTING_CUSTOM_ERRORS = isWronglyReportingCustomErrors();
+let WRONGLY_REPORTING_CUSTOM_ERRORS = isWronglyReportingCustomErrors();
+
+export function _setWRONGLY_REPORTING_CUSTOM_ERRORS(value: boolean) {
+  WRONGLY_REPORTING_CUSTOM_ERRORS = value;
+}
--- a/packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
+++ b/packages/core/src/tools/stackTrace/computeStackTrace.spec.ts
@@ -1,6 +1,6 @@
 import { isSafari } from '../utils/browserDetection'
 import * as CapturedExceptions from './capturedExceptions.specHelper'
-import { computeStackTrace } from './computeStackTrace'
+import { _setWRONGLY_REPORTING_CUSTOM_ERRORS, computeStackTrace } from './computeStackTrace'
 
 describe('computeStackTrace', () => {
   it('should not remove anonymous functions from the stack', () => {
@@ -958,4 +958,31 @@ Error: foo
       column: undefined,
     })
   })
+
+  it('should parse Firefox custom errors', () => {
+    class _DatadogTestCustomError extends Error {
+      constructor() {
+        super();
+        this.name = 'TestError';
+        this.stack = `_DatadogTestCustomError@http://localhost:8080/:35:13
+throwCustomError@http://localhost:8080/:39:26
+`;
+      }
+    }
+
+    _setWRONGLY_REPORTING_CUSTOM_ERRORS(true);
+    
+    const customError = new _DatadogTestCustomError();
+    const stackFrames = computeStackTrace(customError);
+
+    expect(stackFrames.stack.length).toBe(1);
+    expect(stackFrames.stack[0]).toEqual({
+      args: [],
+      func: 'throwCustomError',
+      url: 'http://localhost:8080/',
+      line: 39,
+      column: 26,
+    })
+  })
+  
 })


packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +203,39 @@ function tryToParseMessage(messageObj: unknown) {
...
+  const errorProto = Object.getPrototypeOf(error) as { constructor?: () => Error } | undefined
+  return errorProto?.constructor?.toString?.().startsWith('class ')
+}
+
+let WRONGLY_REPORTING_CUSTOM_ERRORS: boolean | undefined
+
+function isWronglyReportingCustomErrors() {
+  if (WRONGLY_REPORTING_CUSTOM_ERRORS !== undefined) {
+    return WRONGLY_REPORTING_CUSTOM_ERRORS
+  }
```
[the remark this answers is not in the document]
leomelki: I tried but the linter doesn't like it : 

```CallExpression can have side effects when the module is evaluated. Maybe move it in a function declaration?eslint(local-rules/disallow-side-effects)```

File: packages/core/src/tools/stackTrace/computeStackTrace.ts [hunk 2]
```diff
@@ -190,3 +203,36 @@ function tryToParseMessage(messageObj: unknown) {
...
+      super()
+      this.name = 'TestError' // different name than the constructor name
+    }
+  }
+
+  const customError = new _DatadogTestCustomError()
+  const customErrorStack = customError.stack?.toString() ?? ''
+
+  // If the stack trace includes the custom error class name, it means that the constructor is added to the stacktrace
+  WRONGLY_REPORTING_CUSTOM_ERRORS = customErrorStack.includes('_DatadogTestCustomError')
```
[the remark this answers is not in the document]
leomelki: This could not work with some minifier that would create classes without names, or could lead to false positives if there's another element with the same name than the class in the stacktrace.
However, I think we can detect it by comparing a normal stacktrace to the custom error stacktrace.
I'll try implementing something like this

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +203,36 @@ function tryToParseMessage(messageObj: unknown) {
...
+  if (WRONGLY_REPORTING_CUSTOM_ERRORS !== undefined) {
+    return WRONGLY_REPORTING_CUSTOM_ERRORS
+  }
+
+  // This class name should be unique and not minified during compilation (so that it remains unique in the stacktrace).
+  /* eslint-disable no-restricted-syntax */
+  class _DatadogTestCustomError extends Error {
+    constructor() {
+      super()
+      this.name = 'TestError' // different name than the constructor name
```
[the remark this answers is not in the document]
leomelki: It impacts some browsers stacktraces, and it's to ensure this does not break our detection or anything else, if something is modified at a later point

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -48,6 +48,19 @@ export function computeStackTrace(ex: unknown): StackTrace {
     })
   }
 
+  if (stack.length > 0 && isWronglyReportingCustomErrors()) {
+    // if we are wrongly reporting custom errors
+    if (ex instanceof Error && isErrorCustomError(ex)) {
+      // if the element is a custom error
+      const firstStackFrame = stack[0]
+      const errorConstructorName = Object.getPrototypeOf(ex)?.constructor?.name || UNKNOWN_FUNCTION
+      if (firstStackFrame?.func === errorConstructorName) {
```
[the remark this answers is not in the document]
leomelki: It's better to ensure that we did not make any mistake (it's free additional safety)

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -48,6 +48,19 @@ export function computeStackTrace(ex: unknown): StackTrace {
...
+        // if the first stack frame is the custom error constructor
+        stack.shift() // remove it
+      }
+    }
```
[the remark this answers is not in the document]
leomelki: That's a very good idea ! Totally forgot custom errors inheritance

packages/core/src/tools/stackTrace/computeStackTrace.ts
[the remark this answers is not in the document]
BenoitZugmeyer: It's not free though, bundle size is a big concern for us 🙂 Also it won't work if for some reason we fail to parse the function name.

test/e2e/scenario/rum/errors.scenario.ts
```diff
@@ -101,7 +116,7 @@ test.describe('rum errors', () => {
       })
     })
 
-  createTest('send custom errors')
+  createTest('send errors from custom source')
```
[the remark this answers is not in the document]
leomelki: done ✅

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +203,36 @@ function tryToParseMessage(messageObj: unknown) {
...
+  /* eslint-disable no-restricted-syntax */
+  class _DatadogTestCustomError extends Error {
+    constructor() {
+      super()
+      this.name = 'TestError' // different name than the constructor name
+    }
+  }
+
+  const customError = new _DatadogTestCustomError()
+  const customErrorStack = customError.stack?.toString() ?? ''
```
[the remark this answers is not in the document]
leomelki: done ✅

webpack.base.js
```diff
@@ -63,6 +63,7 @@ module.exports = ({ entry, mode, filename, types, keepBuildEnvVariables, plugins
             unsafe: true,
             unsafe_methods: true,
           },
+          keep_classnames: /^_/,
```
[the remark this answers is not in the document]
leomelki: Removed this as the check is now comparing stacktraces instead of looking for a specific symbol.

test/e2e/scenario/rum/errors.scenario.ts
```diff
@@ -15,6 +15,21 @@ function createBody(errorGenerator: string) {
...
+        })('oh snap')
+      }
+      function customErrorWithName() {
+        return new (class CustomTestError extends Error {
+          constructor(e) {
+            super(e)
+            this.name = 'CustomTestError'
+          }
+        })('oh snap')
+      }
```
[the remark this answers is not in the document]
leomelki: done ✅

packages/core/src/tools/stackTrace/computeStackTrace.ts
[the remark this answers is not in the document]
leomelki: Changed this to a stacktrace comparaison. Both a custom and a normal error are instantiated from the exact same symbol in the code, and both stacktraces are compared to see if they are different or not.

packages/core/src/tools/stackTrace/computeStackTrace.ts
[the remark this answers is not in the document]
leomelki: Changed this test to also check for error classes inheritance so it's more useful

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +203,36 @@ function tryToParseMessage(messageObj: unknown) {
...
+// Custom error stacktrace fix
+// Some browsers (safari/firefox) add the error constructor as a frame in the stacktrace
+// In order to normalize the stacktrace, we need to remove it
+
+function isErrorCustomError(error: Error) {
+  const errorProto = Object.getPrototypeOf(error) as { constructor?: () => Error } | undefined
+  return errorProto?.constructor?.toString?.().startsWith('class ')
+}
+
+let WRONGLY_REPORTING_CUSTOM_ERRORS: boolean | undefined
```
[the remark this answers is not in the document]
leomelki: done ✅

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +203,36 @@ function tryToParseMessage(messageObj: unknown) {
...
   return { name, message }
 }
+
+// Custom error stacktrace fix
+// Some browsers (safari/firefox) add the error constructor as a frame in the stacktrace
+// In order to normalize the stacktrace, we need to remove it
+
+function isErrorCustomError(error: Error) {
+  const errorProto = Object.getPrototypeOf(error) as { constructor?: () => Error } | undefined
+  return errorProto?.constructor?.toString?.().startsWith('class ')
```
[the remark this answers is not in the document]
leomelki: done ✅

packages/core/src/tools/stackTrace/computeStackTrace.ts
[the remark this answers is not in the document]
leomelki: Done ! It now goes through all the constructors and compare them to the stacktrace in order. Also added an e2e test to test for it with a 3 classes inheritance (with named and unnamed classes).

packages/core/src/tools/stackTrace/computeStackTrace.ts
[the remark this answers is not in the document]
leomelki: I tried to minimise the build size cost, tell me if it works for you. I think this check is still important as stacktrace reporting is critical, and it's better to check for as many things as we can as it's code running in an unknown user environment.

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +217,45 @@ function tryToParseMessage(messageObj: unknown) {
...
+
+  /* eslint-disable no-restricted-syntax */
+  class DatadogTestCustomError extends Error {
+    constructor() {
+      super()
+      this.name = 'Error' // set name to Error so that no browser would default to the constructor name
+    }
+  }
+
+  const [customError, normalError] = [DatadogTestCustomError, Error].map((errConstructor) => new errConstructor()) // so that both errors should exactly have the same stacktrace
```
BenoitZugmeyer: **praise:** smart!

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -48,6 +48,33 @@ export function computeStackTrace(ex: unknown): StackTrace {
     })
   }
 
+  if (stack.length > 0 && isWronglyReportingCustomErrors()) {
+    // if we are wrongly reporting custom errors
+    if (ex instanceof Error && isErrorCustomError(ex)) {
```
BenoitZugmeyer: **suggestion:** the `isErrorCustomError(ex)` check should not be needed here
**suggestion:** merge the two `if`

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -48,6 +48,33 @@ export function computeStackTrace(ex: unknown): StackTrace {
...
+      // traverse the stacktrace in reverse order because the stacktrace starts with the last inherited constructor
+      // we check constructor names to ensure we remove the correct frame (and there isn't a weird unsupported environment behavior)
+      for (let i = constructors.length - 1; i >= 0; i--) {
+        if (stack[0]?.func === constructors[i]) {
+          // if the first stack frame is the custom error constructor
+          stack.shift() // remove it
+        } else {
+          break
+        }
+      }
```
BenoitZugmeyer: **suggestion:** 
Can we have something like this?

```ts
  if (isWronglyReportingCustomErrors() && ex instanceof Error) {
    let currentPrototype = Object.getPrototypeOf(ex)
    while (
      currentPrototype &&
      isNonNativeClassPrototype(currentPrototype) &&
      stack.length > 0 &&
      stack[0].func === (currentPrototype.constructor.name || UNKNOWN_FUNCTION)
    ) {
      stack.shift()
      currentPrototype = Object.getPrototypeOf(currentPrototype)
    }
  }
```

with

```ts
function isNonNativeClassPrototype(prototype: object) {
  return String(prototype.constructor).startsWith('class ')
}
```

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +217,45 @@ function tryToParseMessage(messageObj: unknown) {
...
+    isWronglyReportingCustomErrorsCache = false
+    return isWronglyReportingCustomErrorsCache
+  }
+
+  const customErrorStack = customError.stack
+
+  // If the stack trace includes the custom error class name, it means that the constructor is added to the stacktrace
+  // If the browser is correctly reporting the stacktrace, the normal error stacktrace should be the same as the custom error stacktrace
+  isWronglyReportingCustomErrorsCache =
+    String(customErrorStack).includes(DatadogTestCustomError.name) && normalError.stack !== customErrorStack
```
BenoitZugmeyer: **question:** is `String(customErrorStack).includes(DatadogTestCustomError.name)` useful? Isn't the stack comparison enough?

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +217,45 @@ function tryToParseMessage(messageObj: unknown) {
...
+  class DatadogTestCustomError extends Error {
+    constructor() {
+      super()
+      this.name = 'Error' // set name to Error so that no browser would default to the constructor name
+    }
+  }
+
+  const [customError, normalError] = [DatadogTestCustomError, Error].map((errConstructor) => new errConstructor()) // so that both errors should exactly have the same stacktrace
+
+  if (!isErrorCustomError(customError)) {
```
BenoitZugmeyer: **suggestion:** With my suggestion above, you could use:

```suggestion
  if (!isNonNativeClassPrototype(DatadogTestCustomError.prototype)) {
```

test/e2e/scenario/rum/errors.scenario.ts
```diff
@@ -15,6 +15,32 @@ function createBody(errorGenerator: string) {
...
+          constructor(e) {
+            super(e)
+          }
+        }
+        return new (class extends CustomTestError2 {
+          constructor(e) {
+            super(e)
+            this.name = 'CustomTestError3'
+          }
+        })('oh snap')
```
BenoitZugmeyer, line 33: **suggestion:** for legibility, use a class statement:

```suggestion
        class CustomTestError3 extends CustomTestError2 {
          constructor(e) {
            super(e)
            this.name = 'CustomTestError3'
          }
        }
        return new CustomTestError3('oh snap')
```

test/e2e/scenario/rum/errors.scenario.ts
```diff
@@ -15,6 +15,32 @@ function createBody(errorGenerator: string) {
...
+        class CustomTestError extends Error {
+          constructor(e) {
+            super(e)
+          }
+        }
+        class CustomTestError2 extends CustomTestError {
+          constructor(e) {
+            super(e)
+          }
+        }
```
BenoitZugmeyer: **suggestion:** constructor aren't really needed here
```suggestion
      function customError() {
        class CustomTestError extends Error {}
        return new CustomTestError('oh snap')
      }
      function customErrorWithInheritance() {
        class CustomTestError extends Error {}
        class CustomTestError2 extends CustomTestError {}
```

packages/core/src/tools/stackTrace/computeStackTrace.ts
BenoitZugmeyer, line 1: **suggestion:** add unit tests for those changes

test/e2e/scenario/rum/errors.scenario.ts
leomelki: Not using a class statement here was done on purpose, because it's an anonymous class (without a name). I don't think there's another better way to create this. (i did this to also test when a constructor has no name, and then when the parser can't find the function name)

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: moving the currentPrototype assignment in the while loop would increase the build size slightly as we would have to initialize the variable. Is this necessary ?

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: done ✅  (with the iserrorcustomerror to isnonnativeclassprototype function change)

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: just removed it. it was checking for unexpected environment behaviour but this test is almost useless once minified.

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: done ✅

test/e2e/scenario/rum/errors.scenario.ts
leomelki: done ✅

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: done ✅  however, they'll only be useful when running in bs because it'll not test anything when running on chromium.

leomelki:
Just found an INSANE behaviour from WebKit :
When a custom error class does not have a constructor defined, it is not shown in the stacktrace.

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -156,7 +152,7 @@ function parseWinLine(line: string): StackFrame | undefined {
 }
 
 const GECKO_LINE_RE =
-  /^\s*(.*?)(?:\((.*?)\))?(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?\s*$/i
+  /^\s*(.*?)(?:\((.*?)\))?(?:^|@)(?:((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?)?\s*$/i
```
leomelki, line 155: Made a fix of this issue in https://github.com/DataDog/browser-sdk/pull/3555/commits/dfd486b804f1c7faf2a949fb38bbf26ea69c75f9

Tell me what you think about this @BenoitZugmeyer .

The gecko regex seems to be used for Webkit stacktrace parsing. And WebKit can print stackframes without origin, so i changed it to be optional.

Here is an example of a WebKit stacktrace containing stackframes without origin :

```
**CustomTestError@**
**CustomTestError2@**
@http://localhost:8080/:38:18
throwCustomError@http://localhost:8080/:41:11
onclick@http://localhost:8080/:30:17
```

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -156,7 +152,7 @@ function parseWinLine(line: string): StackFrame | undefined {
 }
 
 const GECKO_LINE_RE =
-  /^\s*(.*?)(?:\((.*?)\))?(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?\s*$/i
+  /^\s*(.*?)(?:\((.*?)\))?(?:(?:(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?)|@)\s*$/i
```
leomelki, line 151: Made a fix of [this issue](https://github.com/DataDog/browser-sdk/pull/3555#issuecomment-2935203455) in https://github.com/DataDog/browser-sdk/pull/3555/commits/79239b788e77cea57df31e8951ff72a7bbd975c5

Tell me what you think about this @BenoitZugmeyer .

The gecko regex seems to be used for Webkit stacktrace parsing. And WebKit can print stackframes without origin, so i changed it to be optional.
It's checking for either `@file` or just `@`

Here is an example of a WebKit stacktrace containing stackframes without origin :

```
CustomTestError@        <-
CustomTestError2@       <-
@http://localhost:8080/:38:18
throwCustomError@http://localhost:8080/:41:11
onclick@http://localhost:8080/:30:17
```

packages/core/src/tools/stackTrace/computeStackTrace.ts
BenoitZugmeyer: Ok, WDYT of:
```ts
  if (isWronglyReportingCustomErrors() && ex instanceof Error) {
    let currentPrototype = ex
    while (
      (currentPrototype = Object.getPrototypeOf(currentPrototype)) &&
      isNonNativeClassPrototype(currentPrototype) &&
      stack.length > 0 &&
      stack[0].func === (currentPrototype.constructor.name || UNKNOWN_FUNCTION)
    ) {
      stack.shift()
    }
  }
```
It's a bit more cryptic but... I'd be fine with it.

packages/core/src/tools/stackTrace/computeStackTrace.ts
BenoitZugmeyer: * Can you add a unit test covering this case?
* Would this work? (slightly simpler)
```suggestion
  /^\s*(.*?)(?:\((.*?)\))?(?:(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?)@?\s*$/i
```

packages/core/src/tools/stackTrace/computeStackTrace.ts
```diff
@@ -190,3 +217,40 @@ function tryToParseMessage(messageObj: unknown) {
...
+  if (!isNonNativeClassPrototype(Object.getPrototypeOf(customError))) {
+    // This was built with ES5 as target, converting the class to a normal object.
+    // Thus, error constructors will be reported on all browsers, which is the expected behavior.
+    isWronglyReportingCustomErrorsCache = false
+    return isWronglyReportingCustomErrorsCache
+  }
+
+  // If the browser is correctly reporting the stacktrace, the normal error stacktrace should be the same as the custom error stacktrace
+  isWronglyReportingCustomErrorsCache = nativeError.stack !== customError.stack
+  return isWronglyReportingCustomErrorsCache
```
BenoitZugmeyer: **suggestion:** merge the two condition:
```ts
  isWronglyReportingCustomErrorsCache =
    // This was built with ES5 as target, converting the class to a normal object.
    // Thus, error constructors will be reported on all browsers, which is the expected behavior.
    isNonNativeClassPrototype(Object.getPrototypeOf(customError)) &&
    
    // If the browser is correctly reporting the stacktrace, the normal error stacktrace should be the same as the custom error stacktrace
    nativeError.stack !== customError.stack
  return isWronglyReportingCustomErrorsCache
```

test/e2e/scenario/rum/errors.scenario.ts
BenoitZugmeyer: Can you add a comment to clarify this?

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: I first went for such a solution, but there's an issue which makes it imperfect (so it's unsuitable).

Doing it this way would result in `/^\s*(.*?)(?:\((.*?)\))?(?:(?:^|@)((?:file|https?|blob|chrome|webpack|resource|capacitor|\[native).*?|[^@]*bundle|\[wasm code\])(?::(\d+))?(?::(\d+))?)@?\s*$/i` with the file group being optional (`?`). However, with both the file group and the `@` being optional, it makes blank strings a valid stackframe, which it is not.

Thus, the only good solution I found was making them exclusive using a OR statement. (just `@` is also a valid stackframe, but ` ` is not)

packages/core/src/tools/stackTrace/computeStackTrace.ts
leomelki: done  in https://github.com/DataDog/browser-sdk/pull/3555/commits/b38bc1c733c7462dfd05bcb6d9f9e6671da85224 ✅

packages/core/src/tools/stackTrace/computeStackTrace.ts
BenoitZugmeyer: Mmh ok, could you add unit tests to cover all those cases? (including 'blank strings should not be a valid stackframe')

leomelki: It's already tested in other tests (that's what made me think about this)
[Here](https://github.com/DataDog/browser-sdk/pull/3555/files#diff-9d902337d5c641e88ea4ad622f36287463b434de980258e1b1768b0614bfaa04R947) for instance, a new line is added at the end of the stacktrace, which would be considered a valid stackframe with the other regex

leomelki: I just added a [test](https://github.com/DataDog/browser-sdk/pull/3555/files#diff-9d902337d5c641e88ea4ad622f36287463b434de980258e1b1768b0614bfaa04R991) specifically for this, so that it's not just a side effect of other tests. It's testing for :
- new lines
- new lines with spaces
- new lines with a tab (\t)

amortemousque (approved):

PR #3558, opened 2025-05-23
✨ [FFL-16] implement precompute client onto openfeature flagging inte…

leoromanovsky:
…rface

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/flagging/src/openfeature/provider.ts
```diff
@@ -5,57 +5,68 @@
...
     _logger: Logger
   ): ResolutionDetails<string> {
     return {
-      value: defaultValue,
+      value: this.precomputeClient?.getStringAssignment(flagKey, defaultValue) ?? defaultValue,
       reason: StandardResolutionReasons.DEFAULT,
     }
   }
 
   resolveNumberEvaluation(
```
leoromanovsky, line 46: OF doesn't differentiate between integer and floating point types. We will need to handle this in a future interation.

packages/flagging/src/openfeature/provider.ts
```diff
@@ -5,57 +5,68 @@
...
   metadata = {
     name: 'datadog',
   } as const
 
+  private precomputeClient: PrecomputeClient | undefined
+
+  constructor(precomputeClient?: PrecomputeClient) {
+    this.precomputeClient = precomputeClient
+  }
+
```
leoromanovsky, line 21: The API here is still rough and needs refinement as we add network backed precompute loading mode.

packages/flagging/src/openfeature/provider.ts
```diff
@@ -5,57 +5,68 @@
...
     _context: EvaluationContext,
     _logger: Logger
   ): ResolutionDetails<T> {
+    const value =
+      typeof defaultValue === 'object' && defaultValue !== null && this.precomputeClient
+        ? (this.precomputeClient.getJSONAssignment(flagKey, defaultValue as object) as T)
+        : defaultValue
     return {
-      value: defaultValue,
+      value,
```
leoromanovsky, line 69: openfeature wants to return a `JsonValue` but I don't want to inject the type into the precompute client. the mapping is rough.

packages/flagging/src/openfeature/provider.ts
aarsilv: Interesting. We've seen some requests for write-time validation. Perhaps we ditch integer flags but offer integers as one of those validations. Or push to expand the OF spec 🤔

packages/flagging/src/precomputeClient.ts
```diff
@@ -0,0 +1,183 @@
+/* eslint-disable local-rules/disallow-side-effects */
```
aarsilv, line 1: 👀  What was triggering this warning?

packages/flagging/src/precomputeClient.ts
```diff
@@ -0,0 +1,183 @@
...
+      entityId: null,
+    }
+
+    try {
+      const transformedValue =
+        result.variation?.value !== undefined ? valueTransformer(result.variation.value) : defaultValue
+      // TODO: add logging or open feature hook
+      // if (result?.doLog) {
+      //   this.logAssignment(result)
+      // }
```
aarsilv, line 123: 👍 good to be thinking about this.
I imagine int he future the logger will also be passed the allocation type, so finer grained logging or not could be controlled by client vs. doLog. I can see a world where we rid ourselves of doLog entirely.

packages/flagging/src/precomputeClient.ts
```diff
@@ -0,0 +1,183 @@
...
+ */
+export function offlinePrecomputedInit(config: IPrecomputedClientConfigSync): PrecomputeClient | null {
+  let configurationWire: IConfigurationWire
+  try {
+    configurationWire = JSON.parse(config.precomputedConfiguration)
+    if (!configurationWire.precomputed) {
+      throw new Error('Invalid precomputed configuration wire: missing precomputed field')
+    }
+  } catch {
+    const errorMessage = 'Invalid precomputed configuration wire'
```
aarsilv, line 155: any reason for this higher-scoped intermediate variable? Could pull down inside the `if` or just inline it.

packages/flagging/src/precomputeClient.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (config.throwOnFailedInitialization) {
+      throw new Error(errorMessage)
+    }
+    return null
+  }
+
+  const { subjectKey, subjectAttributes, response } = configurationWire.precomputed
+  const parsedResponse: IPrecomputedConfigurationResponse = JSON.parse(response)
```
aarsilv, line 163: should we have a try-catch here in case response is bonked?

Same for defensively checking `flags` exists.

aarsilv (approved):
Starting to look like a real client! 📈 
Great stuff, left some comments you are free to take or leave as you see fit!

packages/flagging/src/precomputeClient.ts
leoromanovsky: Error

packages/flagging/src/openfeature/provider.ts
leoromanovsky: Yea .. I am thinking the former. Although types like `double` exist I don't see a ton of reasons why customers would care to create a flag with one type and not the other, especially since we don't allow type changes. I bet its pretty frustrating to have a value like `5`, want another variation `5.5` and be blocked from creating it

PR #3567, opened 2025-05-26
♻️ [RUM-10005] Exposes `addEvent` instead of `strategy` to plugins

thomas-lebeau:
## Motivation

Refactor plugin integration to use `addEvent` instead of `strategy`. This gives plugins author the ability to better integrate with native the RUM events (e.g. click actions)

## Changes

- Replace `strategy` object by `addEvent` method in plugin's `onInit()` method.
  - `addEvent` accept a subset of `RumEvent` as defined by rum-event-format and the `domainContext` to be passed to `beforeSend`.
  - `addEvent` will then send the event to the lifecycle `RAW_RUM_EVENT_COLLECTED` to be assembled and send.
- Use the new `addEvent` in the react plugin

## Test instructions

- open the react-app sandbox's error boundaries test page (http://localhost:8080/react-app/test-error-boundary)
- throw and error
- look at the error in the developer extension

Additionally you can compare the error event before and after the refactor, they should be mostly identical

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/event/eventCollection.ts
```diff
@@ -0,0 +1,51 @@
...
+      domainContext: RumEventDomainContext,
+      duration?: Duration
+    ) => {
+      if (!allowedEventTypes.includes(event.type)) {
+        return
+      }
+
+      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, {
+        startTime,
+        rawRumEvent: event as RawRumEvent,
```
amortemousque: **💬 suggestion:** ‏We can avoid casting and passing a partial raw event like `Pick<RawRumActionEvent, CommonProperties | 'action'>`. The only missing field in `RawRumEvent` seems to be `context`, so I think it’s cleaner to update `RawRumEvent` directly to clarify the expected properties during assembly.

test/e2e/scenario/plugin.scenario.ts
```diff
@@ -0,0 +1,141 @@
...
+function createPlugin(): RumPlugin {
+  return {
+    name: 'test-plugin',
+    onRumStart: ({ addEvent }) => {
+      window.TEST_PLUGIN = { addEvent }
+    },
+  }
+}
+
+const mockPartiaEvents: AllowedRawRumEvent[] = [
```
amortemousque: **🥜 nitpick:** ‏
```suggestion
const mockPartialEvents: AllowedRawRumEvent[] = [
```

packages/rum-core/src/domain/event/eventCollection.ts
```diff
@@ -0,0 +1,53 @@
...
+  | RawRumActionEvent
+  | RawRumVitalEvent
+) & { context?: Context }
+
+export function startEventCollection(lifeCycle: LifeCycle) {
+  return {
+    addEvent: (
+      startTime: RelativeTime,
+      event: AllowedRawRumEvent,
+      domainContext: RumEventDomainContext,
```
amortemousque, line 38: **❓ question:** Are we sure we will have domain context? ‏

amortemousque (approved):

packages/rum-core/src/domain/plugins.ts
```diff
@@ -1,11 +1,12 @@
...
+import type { RumPublicApi } from '../boot/rumPublicApi'
+import type { StartRumResult } from '../boot/startRum'
 import type { RumInitConfiguration } from './configuration'
 
 export interface RumPlugin {
   name: string
   getConfigurationTelemetry?(): Record<string, unknown>
   onInit?(options: { initConfiguration: RumInitConfiguration; publicApi: RumPublicApi }): void
-  onRumStart?(options: { strategy: Strategy }): void
+  onRumStart?(options: { addEvent: StartRumResult['addEvent'] }): void
```
BenoitZugmeyer: **suggestion:** keep `strategy` for the time being, but mark it as "deprecated".

We are using a fixed rum-react NPM package version in web-ui, so deploying this PR in Canary without `strategy` will break web-ui integration. In that spirit, I think removing `strategy` might be a breaking change and we could schedule to remove it in the next major.

test/e2e/scenario/plugin.scenario.ts
```diff
@@ -0,0 +1,141 @@
...
+          ({ startClocks, event }) => {
+            window.TEST_PLUGIN.addEvent(startClocks.relative, event, {})
+          },
+          { startClocks, event: partialEvent }
+        )
+        await flushEvents()
+        const event = intakeRegistry.rumEvents.filter(
+          (e) => e.type === partialEvent.type && (e as any)[eventType]?.id === (partialEvent as any)[eventType].id
+        )
+        console.log('event', event)
```
BenoitZugmeyer: **suggestion:** 
```suggestion
```

test/e2e/scenario/plugin.scenario.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        expect(event.length).toBeGreaterThan(0)
```
BenoitZugmeyer: **question:** Shouldn't it be 1?
```suggestion
        expect(event.length).toBe(1)
```

PR #3569, opened 2025-05-26
[RUM Profiler] Simplify & fix logic for Long Task registry

thomasbertet:
## Motivation

- We sometimes capture Long Tasks using the RUM Profiler without their RUM Long Task IDs. 
- The logic to clean-up and capture Long Task ids (the Registry) is flawed.

## Changes

- Remove the need for an `enabledClocks` alltogether. Whenever we get a Long Task, we store its id in the registry, and whenever we want to capture the id to send in the Profile, we use that registry. 
- Whenever the RUM Profiler is stopped, we clean-up the registry. 

The resulting logic is simpler and still avoids cluttering the memory.

## Test instructions

- Look at RUM Profiles collected, and especially all the Long Tasks reported (or use the UI in the RUM Long Task panel), and assert they all have an ID associated.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
+++ b/packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
@@ -1,4 +1,4 @@
-import type { ClocksState, RelativeTime } from '@datadog/browser-core'
+import type { RelativeTime } from '@datadog/browser-core'
 
 type PerformanceEntryStartTime = RelativeTime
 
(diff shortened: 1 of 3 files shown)


packages/rum/src/domain/profiling/utils/longTaskRegistry.ts
```diff
@@ -8,42 +7,31 @@ type PerformanceEntryStartTime = RelativeTime
...
-    if (performanceEntryStartTime < collectionClocks.relative) {
-      registry.delete(performanceEntryStartTime)
+/**
+ * Delete the Long Task from the registry once we have collected it.
+ *
+ * @param collectionClocks The clocks of the current collection
+ */
+export function cleanupLongTaskRegistryAfterCollection(collectionClocks: ClocksState) {
+  for (const performanceStartTime of registry.keys()) {
+    if (performanceStartTime < collectionClocks.relative) {
```
BenoitZugmeyer, line 32: **nitpick:** Looks like only the relative time is used, maybe use `RelativeTime` instead of `ClocksState`

BenoitZugmeyer (commented):
It would probably be worth covering this logic with unit tests

PR #3573, opened 2025-05-27
🐛 [RUM-10002] Verify the document was not hidden while loading

mormubis:
## Motivation

This PR addresses two issues with the current implementation of loading time when the page is hidden during loading.

**Issue 1**  
During the initial load, we check the `PerformanceObserver` to determine if the page was hidden or visible while loading. We do this because the SDK loads asynchronously, so we only have access to future events from that point onward.

The previous implementation had a problem: the performance observer was accessed even during client-side navigation. 

This condition will always be true, potentially leading to an entry in the observer that is lower than the loading time.

**Issue 2**  
This issue is related to Issue 1. The time compared during client-side navigation was always based on the time origin instead of the loading view. 

This means that if the page undergoes client-side navigation while hidden, the timestamp used in the visibility-change event is based on the time origin rather than the view start reference. As a result, it is almost always higher than expected.

## Changes

This PR changes the `trackFirstHidden` function adding options to only check the performance observer when needed. Additionally, it modifies the condition to check if the page was hidden while loading by replacing the reference point to view start.

| Before | Header |
|--------|--------|
| https://github.com/user-attachments/assets/dbbadff0-422d-49c6-a66a-745e327f4037 | https://github.com/user-attachments/assets/0cad4558-3c9e-415f-9adb-73a02999b0ba | 

## Test instructions

**Case 1**

Using the sandbox, modify the home page to extend the loading time. Like adding a fetch to the home page:

```
fetch('https://upload.wikimedia.org/wikipedia/commons/thumb/3/3f/Fronalpstock_big.jpg/1599px-Fronalpstock_big.jpg')
```

1. Go to http://localhost:8080/react-app/user/42
2. Open dev tools
3. Throttle the network for easy testing
4. Reload the page and quickly change the active tab
5. Go back to the sandbox tab
6. _Check there is no loading time for the current view_
7. Click the home link
8. _Check there is loading time for the home view_

> [!NOTE]
> Previously, this fails as it finds a performance observer entry with a lower time than the loading time.

**Case 2**

Following the same approach as the case above:

1. Go to http://localhost:8080/react-app/user/42
2. Open dev tools
3. Throttle the network for easy testing
7. Click the home link
8. Quickly change the active tab and go back
9. _Check there is **no** loading time for the home view_

> [!NOTE]
> Previously, it reported a wrongful time. 

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
@@ -1,5 +1,5 @@
 import type { ClocksState, RelativeTime } from '@datadog/browser-core'
-import { clocksOrigin, addEventListeners, DOM_EVENT, noop } from '@datadog/browser-core'
+import { addEventListeners, DOM_EVENT, noop } from '@datadog/browser-core'
 import type { RumConfiguration } from '../../configuration'
 import { supportPerformanceTimingEvent, RumPerformanceEntryType } from '../../../browser/performanceObservable'
 
@@ -11,8 +11,8 @@ export type Options = {
 
 export function trackFirstHidden(
   configuration: RumConfiguration,
-  eventTarget: Window = window,
-  { viewStart = clocksOrigin() }: Partial<Options> = {}
+  viewStart: ClocksState,
+  eventTarget: Window = window
 ) {
   if (document.visibilityState === 'hidden') {
     return { timeStamp: 0 as RelativeTime, stop: noop }
--- a/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
@@ -22,7 +22,7 @@ export function trackFirstHidden(
     const firstHiddenEntry = performance
       .getEntriesByType(RumPerformanceEntryType.VISIBILITY_STATE)
       .filter((entry) => entry.name === 'hidden')
-      .find((entry) => entry.startTime > viewStart.relative)
+      .find((entry) => entry.startTime >= viewStart.relative)
 
     if (firstHiddenEntry) {
       return { timeStamp: firstHiddenEntry.startTime as RelativeTime, stop: noop }
--- a/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
@@ -1,11 +1,19 @@
-import type { RelativeTime } from '@datadog/browser-core'
-import { addEventListeners, DOM_EVENT, noop } from '@datadog/browser-core'
+import type { ClocksState, RelativeTime } from '@datadog/browser-core'
+import { clocksOrigin, addEventListeners, DOM_EVENT, noop } from '@datadog/browser-core'
 import type { RumConfiguration } from '../../configuration'
 import { supportPerformanceTimingEvent, RumPerformanceEntryType } from '../../../browser/performanceObservable'
 
 export type FirstHidden = ReturnType<typeof trackFirstHidden>
 
-export function trackFirstHidden(configuration: RumConfiguration, eventTarget: Window = window) {
+export type Options = {
+  viewStart: ClocksState
+}
+
+export function trackFirstHidden(
+  configuration: RumConfiguration,
+  eventTarget: Window = window,
+  { viewStart = clocksOrigin() }: Partial<Options> = {}
+) {
   if (document.visibilityState === 'hidden') {
     return { timeStamp: 0 as RelativeTime, stop: noop }
   }
@@ -13,7 +21,9 @@ export function trackFirstHidden(configuration: RumConfiguration, eventTarget: W
   if (supportPerformanceTimingEvent(RumPerformanceEntryType.VISIBILITY_STATE)) {
     const firstHiddenEntry = performance
       .getEntriesByType(RumPerformanceEntryType.VISIBILITY_STATE)
-      .find((entry) => entry.name === 'hidden')
+      .filter((entry) => entry.name === 'hidden')
+      .find((entry) => entry.startTime > viewStart.relative)
+
     if (firstHiddenEntry) {
       return { timeStamp: firstHiddenEntry.startTime as RelativeTime, stop: noop }
     }
--- a/packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
@@ -30,7 +30,7 @@ export function trackInitialViewMetrics(
     scheduleViewUpdate()
   })
 
-  const firstHidden = trackFirstHidden(configuration, viewStart, window)
+  const firstHidden = trackFirstHidden(configuration, viewStart)
   const { stop: stopFCPTracking } = trackFirstContentfulPaint(configuration, firstHidden, (firstContentfulPaint) => {
     initialViewMetrics.firstContentfulPaint = firstContentfulPaint
     scheduleViewUpdate()
--- a/packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
@@ -1,4 +1,5 @@
-import type { Duration } from '@datadog/browser-core'
+import type { ClocksState, Duration } from '@datadog/browser-core'
+import { clocksOrigin } from '@datadog/browser-core'
 import type { RumConfiguration } from '../../configuration'
 import { trackFirstContentfulPaint } from './trackFirstContentfulPaint'
 import type { FirstInput } from './trackFirstInput'
@@ -19,7 +20,8 @@ export interface InitialViewMetrics {
 export function trackInitialViewMetrics(
   configuration: RumConfiguration,
   setLoadEvent: (loadEnd: Duration) => void,
-  scheduleViewUpdate: () => void
+  scheduleViewUpdate: () => void,
+  viewStart: ClocksState = clocksOrigin()
 ) {
   const initialViewMetrics: InitialViewMetrics = {}
 
@@ -29,7 +31,7 @@ export function trackInitialViewMetrics(
     scheduleViewUpdate()
   })
 
-  const firstHidden = trackFirstHidden(configuration)
+  const firstHidden = trackFirstHidden(configuration, window, { viewStart })
   const { stop: stopFCPTracking } = trackFirstContentfulPaint(configuration, firstHidden, (firstContentfulPaint) => {
     initialViewMetrics.firstContentfulPaint = firstContentfulPaint
     scheduleViewUpdate()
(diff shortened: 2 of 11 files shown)


packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -26,15 +34,31 @@ describe('trackLoadingTime', () => {
...
+  beforeEach(() => {
+    performanceBufferMock = mockGlobalPerformanceBuffer()
+    if (typeof PerformanceObserver !== 'undefined') {
+      originalSupportedEntryTypes = PerformanceObserver.supportedEntryTypes as string[]
+      Object.defineProperty(PerformanceObserver, 'supportedEntryTypes', {
+        get: () => [...(originalSupportedEntryTypes || []), 'visibility-state'],
+        configurable: true,
+      })
+    }
+  })
```
thomas-lebeau, line 49: **❓ question:** ‏ Do we need this? I guess if visibility state is not supported we should skip the test instead of mocking the support, no?

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -136,6 +160,53 @@ describe('trackLoadingTime', () => {
...
 
+    clock.tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    domMutationObservable.notify([createMutationRecord()])
+    clock.tick(AFTER_PAGE_ACTIVITY_END_DELAY)
+
+    expect(loadingTimeCallback).not.toHaveBeenCalled()
+  })
+
+  it('should not discard loading time if page was hidden before the view start', () => {
+    clock.tick(clock.relative(RANDOM_VIEW_START))
```
thomas-lebeau, line 171: **💬 suggestion:** ‏ I think you don't need to use `clock.relative()` here
```suggestion
    clock.tick(RANDOM_VIEW_START)
```

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -136,6 +160,53 @@ describe('trackLoadingTime', () => {
...
+      entryType: 'visibility-state',
+      name: 'hidden',
+      startTime: clock.relative(RANDOM_VIEW_START - 3000) as number,
+    } as PerformanceEntry)
+
+    startLoadingTimeTracking(ViewLoadingType.ROUTE_CHANGE, clocksNow())
+
+    clock.tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    domMutationObservable.notify([createMutationRecord()])
+    clock.tick(AFTER_PAGE_ACTIVITY_END_DELAY)
```
thomas-lebeau, line 183: **💬 suggestion:** ‏ what about moving this in a function such as `emulatePageActivityDuringViewLoading()`. It would make the tests more readable.

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -136,6 +160,53 @@ describe('trackLoadingTime', () => {
     setPageVisibility('hidden')
     startLoadingTimeTracking()
 
+    clock.tick(BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY)
+    domMutationObservable.notify([createMutationRecord()])
+    clock.tick(AFTER_PAGE_ACTIVITY_END_DELAY)
```
thomas-lebeau, line 165: **👏 praise:** ‏ nice!

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -136,6 +160,53 @@ describe('trackLoadingTime', () => {
...
+
+    expect(loadingTimeCallback).not.toHaveBeenCalled()
+  })
+
+  it('should NOT discard loading time if page is hidden before a new view start happened', () => {
+    setPageVisibility('hidden')
+    startLoadingTimeTracking()
+
     expect(loadingTimeCallback).not.toHaveBeenCalled()
   })
```
thomas-lebeau, line 211: **❓ question:** ‏ I don't understand what this test does? the name suggest the loading time callback should be called (i.e. not discarded) but the test expect it to not have been called.

thomas-lebeau (approved):

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
mormubis: I thought this test was coming from before but it seems a conflict when testing yesterday.

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
mormubis: Removing it. I kept the support based on other tests, but after testing it, it can be removed.

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
```diff
@@ -136,6 +156,48 @@ describe('trackLoadingTime', () => {
...
+
+    performanceBufferMock.addPerformanceEntry({
+      entryType: 'visibility-state',
+      name: 'hidden',
+      startTime: clock.relative(RANDOM_VIEW_START + 10) as number,
+    } as PerformanceEntry)
+
+    startLoadingTimeTracking(ViewLoadingType.ROUTE_CHANGE, clocksNow())
+
+    emulatePageActivityDuringViewLoading()
```
BenoitZugmeyer, line 199: **suggestion:** here too, changing the order would make things easier to follow

```suggestion
    clock.tick(RANDOM_VIEW_START)


    startLoadingTimeTracking(ViewLoadingType.ROUTE_CHANGE, clocksNow())
    
    clock.tick(10)
    
    performanceBufferMock.addPerformanceEntry({
      entryType: 'visibility-state',
      name: 'hidden',
      startTime: performance.now(),
    } as PerformanceEntry)
    
    emulatePageActivityDuringViewLoading()
```

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
```diff
@@ -19,12 +19,12 @@ export function trackLoadingTime(
...
   let isWaitingForActivityLoadingTime = true
   const loadingTimeCandidates: Duration[] = []
-  const firstHidden = trackFirstHidden(configuration)
+  const firstHidden = trackFirstHidden(configuration, window, { viewStart })
 
   function invokeCallbackIfAllCandidatesAreReceived() {
     if (!isWaitingForActivityLoadingTime && !isWaitingForLoadEvent && loadingTimeCandidates.length > 0) {
       const loadingTime = Math.max(...loadingTimeCandidates)
-      if (loadingTime < firstHidden.timeStamp) {
+      if (loadingTime < firstHidden.timeStamp - viewStart.relative) {
```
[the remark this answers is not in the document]
mormubis: `firstHidden` does not represent a `TimeStamp` but a `RelativeTime`. Additionally, it is relative to the time origin and not to the `viewStart` time.

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.spec.ts
mormubis: As per our conversation yesterday, I adapted the unit tests following the convo example.

packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
```diff
@@ -19,7 +20,8 @@ export interface InitialViewMetrics {
 export function trackInitialViewMetrics(
   configuration: RumConfiguration,
   setLoadEvent: (loadEnd: Duration) => void,
-  scheduleViewUpdate: () => void
+  scheduleViewUpdate: () => void,
+  viewStart: ClocksState = clocksOrigin()
```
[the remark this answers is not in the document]
mormubis: Done!

packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
```diff
@@ -29,7 +30,7 @@ export function trackInitialViewMetrics(
     scheduleViewUpdate()
   })
 
-  const firstHidden = trackFirstHidden(configuration)
+  const firstHidden = trackFirstHidden(configuration, viewStart, window)
```
BenoitZugmeyer, line 33: **nitpick:** 
```suggestion
  const firstHidden = trackFirstHidden(configuration, viewStart)
```

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
```diff
@@ -25,12 +25,13 @@ export function trackLoadingTime(
   let isWaitingForLoadEvent = loadType === ViewLoadingType.INITIAL_LOAD
   let isWaitingForActivityLoadingTime = true
   const loadingTimeCandidates: Duration[] = []
-  const firstHidden = trackFirstHidden(configuration)
+  const firstHidden = trackFirstHidden(configuration, viewStart, window)
```
BenoitZugmeyer, line 28: **nitpick:** 
```suggestion
  const firstHidden = trackFirstHidden(configuration, viewStart)
```

packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.ts
```diff
@@ -1,19 +1,29 @@
...
   if (document.visibilityState === 'hidden') {
     return { timeStamp: 0 as RelativeTime, stop: noop }
   }
 
   if (supportPerformanceTimingEvent(RumPerformanceEntryType.VISIBILITY_STATE)) {
     const firstHiddenEntry = performance
       .getEntriesByType(RumPerformanceEntryType.VISIBILITY_STATE)
-      .find((entry) => entry.name === 'hidden')
+      .filter((entry) => entry.name === 'hidden')
+      .find((entry) => entry.startTime > viewStart.relative)
```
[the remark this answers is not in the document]
BenoitZugmeyer: Could you add a `trackFirstHidden` test with a visibility state entry with `startTime:0`? Like:

```ts
    it('should return 0 when the page was loaded hidden', () => {
      setPageVisibility('visible')

      performanceBufferMock.addPerformanceEntry({
        entryType: 'visibility-state',
        name: 'hidden',
        startTime: 0,
      } as PerformanceEntry)

      firstHidden = trackFirstHiddenWithDefaults({ configuration })
      expect(firstHidden.timeStamp).toBe(0 as RelativeTime)
    })

```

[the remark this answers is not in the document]
mormubis: Added

packages/rum-core/src/domain/view/viewMetrics/trackInitialViewMetrics.ts
mormubis: Added

packages/rum-core/src/domain/view/viewMetrics/trackLoadingTime.ts
mormubis: Added

BenoitZugmeyer (approved):

PR #3577, opened 2025-05-27
🔥 🐛 fix Next.js 12 + swcMinify compatibility

BenoitZugmeyer:
## Motivation

Next.js 12 introduced the `swcMinify` to opt-in using SWC to minify JS. The SWC version used at the time was very early and suffered from major issues. In particular, it was observed that the `computeSessionState` function in `rumSessionManager.ts` was minified incorrectly in such a way that it returned `undefined` every time.

## Changes

This commit fixes this issue by adjusting the faulty function implementation.

Instead of returning a "session state", it only returns the "session tracking type" and consider that all sessions with tracking type '0' are not tracked.

This reduces the bundle size a tiny bit, and prevents the concept of "session state" to leak outside of `sessionStore`.

## Test instructions

* Create a Next.js v12 app 
* Enable `swcMinify` using the following `next.config.js`:
```
module.exports = {
  swcMinify: true,
};
```
* Add the browser SDK (using local builds generated by `npm pack` for example) and configure it
* Build the app with `next build`
* Serve the app with `next start`
* Open the app in a browser

-> RUM data should be sent to datadog

## Checklist

- [x] Tested locally
- [x] Tested on staging


sethfowler-datadog (approved):
LGTM!

PR #3554, opened 2025-05-22
feat: add openfeature dependency and datadog provider (FFL-24)

leoromanovsky:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging
- [ ] Added unit tests for this change.
- [ ] Added e2e/integration tests for this change.

PR #3580, opened 2025-05-29
✨ FFL-16 Precomputed flags evaluation

leoromanovsky:
## Motivation

Precomputed flags evaluation + small fixes on top of #3554.

🐛 **Open Bug**: unit tests were failing in chrome 63 (https://gitlab.ddbuild.io/DataDog/browser-sdk/-/jobs/959532643) and trying to fix with changes to polyfill.

## Changes

- Precomputed flags evaluation
- Fetching remote configuration
- Option to set initial configuration from configuration wire

See individual commits for more details.

## Checklist

- [X] Tested locally
- [ ] Tested on staging


sandbox/react-app/main.tsx
```diff
@@ -1,18 +1,44 @@
...
-datadogFlagging.init()
+const subject = {
+  targetingKey: 'subject-key-1',
+}
+
+async function initializeOpenFeature() {
+  const datadogFlaggingProvider = new DatadogProvider({
+    applicationId: 'xxx',
+    clientToken: 'xxx',
+    baseUrl: 'http://localhost:8000',
```
BenoitZugmeyer: **question:** What's this URL? Do we need it to launch the sandbox now?

packages/flagging/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@datadog/browser-flagging",
+  "name": "@datadog/openfeature-provider",
```
BenoitZugmeyer, line 2: **suggestion:** (as discussed) it could be nice to rename the package folder as well

sandbox/react-app/main.tsx
leoromanovsky: It was a local rapid server, good observation - I reverted changes to sandbox back to main, will do testing in the static application we are building in `web-ui`

packages/flagging/package.json
leoromanovsky: My notes from the conversation was we left it without consensus but I am ok revisiting it as the SDK scope becomes more clear; there might be more than just the openfeature provided being exported; perhaps where we will land is a core flagging package (here or in a shared javascript repository with react native) and openfeature provider as a wrapper on it.

BenoitZugmeyer: Sounds good to me.

BenoitZugmeyer (approved):

PR #3592, opened 2025-06-03
✨ [RUM-10144] apply context defined just after init to events generated during init

BenoitZugmeyer:
## Motivation

Usually, context (global/user/account/feature flags etc.) is defined just after calling `RUM.init()`, which is an issue because events that are generated during `init()` will not have this context.

For now, this is only problematic for View events: RUM.init() generates a first View event that is missing the context defined right after. It is usually fine, as we can expect a RUM View Update to be generated at some point later. But sometimes no View Update happen for some reason (connectivity error, browser exiting abruptly...). In those cases, it is unexpected that the View isn't including the context.

This issue will be more pronounced when we'll start to collect errors that happen before init (in a future PR). In this case, we'll want the context to apply to those events (for example, to clearly identify which user was affected by such early error).

## Changes

This PR fixes this issue by slightly delaying the events assembly. It uses a new data structure, `BufferedObservable`, that will be used to progressively replace the `BoundedBuffer` and what we call "pre start strategies". This `BufferedObservable` is asynchrone, so we leverage this fact to delay the assembly.

## Test instructions

In the sandbox, define a user just after `init()`:

```
      DD_RUM.init({
        clientToken: 'xxx',
        applicationId: 'xxx',
        ...
      })
      DD_RUM.setUser({ id: 'foo' })

```

Using the devtools extension, look at individual View events by unchecking  `Show only the latest View event` in the "Events" tab.

When loading the sandbox URL, all View events should contain the `@usr.id`, even the very first View.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -56,7 +56,7 @@ export {
 } from './domain/telemetry'
 export { monitored, monitor, callMonitored, setDebugMode, monitorError } from './tools/monitor'
 export type { Subscription } from './tools/observable'
-export { Observable } from './tools/observable'
+export { Observable, BufferedObservable } from './tools/observable'
 export type { SessionManager } from './domain/session/sessionManager'
 export { startSessionManager, stopSessionManager } from './domain/session/sessionManager'
 export {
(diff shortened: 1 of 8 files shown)


BenoitZugmeyer:
Superseded by https://github.com/DataDog/browser-sdk/pull/3597

PR #3594, opened 2025-06-04
♻️ [RUM-10256] Use hooks for Logs SDK to decouple global context

amortemousque:
## Motivation

Use hooks for Logs SDK to decouple global context

## Test instructions

- Create a `globalContext.ts` for Logs (could not make a shared one in core because in Logs the context attributes are attached to the root of the logs not in the 'context' attribute)
- Buffer global context calls using `defineContextMethod` in `logsPublicApi.ts`
- Remove global context in logs `assembly.ts`
- Add tests for contexts precedence in `startLogs.spec` 

## Checklist

- [x] Tested locally
- [x] Tested on staging


packages/logs/src/domain/contexts/globalContext.ts
BenoitZugmeyer, line 1: **question:** why can't we do the same as for account and use the same code between logs and rum?

packages/logs/src/domain/contexts/globalContext.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { LogsConfiguration } from '../configuration'
+
+export function startGlobalContext(hooks: Hooks, configuration: LogsConfiguration) {
+  const globalContextManager = buildGlobalContextManager()
+
+  if (configuration.storeContextsAcrossPages) {
+    storeContextManager(configuration, globalContextManager, 'logs', CustomerDataType.GlobalContext)
+  }
+
+  hooks.register(HookNames.Assemble, () => globalContextManager.getContext())
```
BenoitZugmeyer: **suggestion:** factorize with RUM, but make the location configurable?

```suggestion
  hooks.register(HookNames.Assemble, () => {
    const context = globalContextManager.getContext()
    return useContextNamespace ? { context } : context 
  })
```

amortemousque: Okay, let's give it a try :)

amortemousque: Done! Better and few bytes saved ;) thanks

BenoitZugmeyer (approved):

packages/core/src/domain/contexts/globalContext.ts
BeltranBulbarellaDD, line 1: Nice!

BeltranBulbarellaDD (approved):

PR #3595, opened 2025-06-04
👷 move testing extensions folder

BeltranBulbarellaDD:
## Motivation

Fix cmd_version script

## Changes

Moved testing extensions 1 file up to be under /test/apps

## Test instructions

CI should pass.

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/apps/vanilla/.gitignore
+++ b/test/apps/vanilla/.gitignore
@@ -1,7 +1,5 @@
 dist
 node_modules
-.yarn/*
-
 
 # https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
 .pnp.*


test/apps/.gitignore
```diff
@@ -0,0 +1,8 @@
+*
+!.gitignore
+!base-extension/
+!base-extension/**
+!react/
+!react/**
+!vanilla/
+!vanilla/**
```
bcaudan: **💬 suggestion:** ‏any reason to ignore everything by default? listing the generated extension instead could be less surprising when adding something new in `test/apps`

bcaudan:
**💬 suggestion:** PR title‏
```diff
- moved testing extensions folder
+ 👷 move testing extensions folder
```
- Emoji are useful to generate automatically the changelog and sort changes in the right section (Public VS internal)
- about commit message in present tense, not sure if it is something that we ever discussed but consistency sake, it could be nice

maycmlee (approved):

test/apps/vanilla/.gitignore
```diff
@@ -1,4 +1,7 @@
 dist
+node_modules
+.yarn/*
+
```
thomas-lebeau: **❓ question:** ‏ Why did you need this?

BeltranBulbarellaDD: I added it because when I tried to commit after modifying the files and building everything a few times while testing, I had those 2 folders to commit and I thought we don't want to push them so I added it just to be safe.

test/apps/vanilla/.gitignore
```diff
@@ -1,4 +1,7 @@
 dist
+node_modules
+.yarn/*
```
bcaudan: **💬 suggestion:** already declared below‏
```suggestion
```

test/apps/vanilla/.gitignore
bcaudan: `node_modules` was probably a miss, it is excluded in other test apps.
`.yarn/*` is already there though.

thomas-lebeau (approved):

PR #3597, opened 2025-06-04
✨ [RUM-10144] apply context defined just after init to View event generated during init

### LINKED ISSUE #3935: 🐛 Initial version of RUM view event is missing global context
**Describe the bug**
We recently upgraded our DataDog RUM client from v5 to v6.19.0.  We funnel all RUM log events through a proxy that locally captures the data for later internal analytics before sending it on to DataDog.  We noticed that there has been a change to the v6 client where an initial version 1 of the `view` event is sent up without any of the global context attached.

FWIW, after trying different versions, I have found that this bug entered the DD RUM browser sdk client exactly in version 6.6.3.  v6.6.2 does not exhibit this bug and 6.6.3 does.

The relevant code to initialize the RUM client is something like this:

```
for (each of several global context properties) {
  datadogRum.setGlobalContextProperty(key, value);
}
datadogRum.init({ ... config ... });
datadogRum.startView();
```

For completeness, I will include the initial view event's JSON data, with slight redaction to `xxx` for a few fields:

```
{
  "type": "view",
  "_dd": {
    "format_version": 2,
    "drift": 1,
    "configuration": {
      "session_sample_rate": 0,
      "session_replay_sample_rate": 100,
      "profiling_sample_rate": 0,
      "start_session_replay_recording_manually": true
    },
    "sdk_name": "rum",
    "page_states": [
      {
        "state": "active",
        "start": 0
      }
    ],
    "document_version": 1
  },
  "application": {
    "id": "xxx"
  },
  "date": 1761171363445,
  "source": "browser",
  "view": {
    "url": "xxx",
    "referrer": "",
    "id": "1f51f66d-b4aa-4079-94ac-303c893b1320",
    "action": {
      "count": 0
    },
    "frustration": {
      "count": 0
    },
    "cumulative_layout_shift": 0,
    "error": {
      "count": 0
    },
    "is_active": true,
    "loading_type": "initial_load",
    "long_task": {
      "count": 0
    },
    "performance": {
      "cls": {
        "score": 0
      }
    },
    "resource": {
      "count": 0
    },
    "time_spent": 817000000
  },
  "session": {
    "id": "f568556f-a7c9-47a3-b843-8a8731501722",
    "type": "user",
    "sampled_for_replay": true
  },
  "connectivity": {
    "status": "connected",
    "effective_type": "4g"
  },
  "context": {},
  "usr": {
    "anonymous_id": "xxx"
  },
  "service": "xxx",
  "version": "1.0.0-746bed91",
  "privacy": {
    "replay_level": "mask"
  },
  "device": {
    "locale": "en-US",
    "locales": [
      "en-US",
      "en"
    ],
    "time_zone": "America/Los_Angeles"
  },
  "ddtags": "sdk_version:6.19.0,env:local,service:xxx,version:1.0.0-746bed91"
}
```

The above view event is followed by later versions of the event for the same view ID that -do- include the context, but this initial version does not, which is breaking some of our view logging that depends on that context being there for all versions of a view.  We believe this is a bug and that the context should be there in v1 of the view since it was set prior to the client init and the view starting.

I am not certain, but I believe this issue may affect resource events that are fired off early in the RUM client's lifecycle, too.

If we downgrade to v6.6.2 or earlier, the view v1 event has similar shape as above, but the `context` member is filled with all of the properties we set prior to initialization.

**To Reproduce**
See above

**Expected behavior**
The `context` property of the version 1 view event should contain all global context properties set prior to the start of the view.

BenoitZugmeyer:
## Motivation

Usually, context (global/user/account/feature flags etc.) is defined just after calling `RUM.init()`, which is an issue because events that are generated during `init()` will not have this context.

This is only problematic for View events: RUM.init() generates a first View event that is missing the context defined right after. It is usually fine, as we can expect a RUM View Update to be generated at some point later. But sometimes no View Update happen for some reason (connectivity error, browser exiting abruptly...). In those cases, it is unexpected that the View isn't including the context.

Fixes #3935

## Changes

This PR fixes this issue by slightly delaying the first View update. It uses `Promise.resolve()` to schedule a view update in the next micro task. It should not impact reliability as using `Promise.resolve().then()` to schedule a micro task is not impacted by timers throttling.

## Test instructions

In the sandbox, define a user just after `init()`:

```
      DD_RUM.init({
        clientToken: 'xxx',
        applicationId: 'xxx',
        ...
      })
      DD_RUM.setUser({ id: 'foo' })

```

Using the devtools extension, look at individual View events by unchecking  `Show only the latest View event` in the "Events" tab.

When loading the sandbox URL, all View events should contain the `@usr.id`, even the very first View.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


BenoitZugmeyer:
@codex review

packages/browser-rum-core/src/domain/view/trackViews.ts
```diff
@@ -286,7 +287,11 @@ function newView(
   })
 
   // Initial view update
-  triggerViewUpdate()
+  mockable(queueMicrotask)(() => {
+    if (documentVersion === 0) {
+      triggerViewUpdate()
+    }
```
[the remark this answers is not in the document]
BenoitZugmeyer: It's probably fine, `setViewName` usage is a bit niche, and fixing makes things a bit more complex: it means introducing yet another lifecycle event to signal that a view is about to be updated, so we can the new view name without waiting for the deferred promise.

packages/browser-rum-core/src/domain/view/trackViews.ts
```diff
@@ -286,7 +287,11 @@ function newView(
   })
 
   // Initial view update
-  triggerViewUpdate()
+  mockable(queueMicrotask)(() => {
```
gogusarov, line 290: `question`

I thought zone.js patches queueMicrotask as well, shouldn't we use it via `getZoneJsOriginalValue` here?

packages/browser-core/test/emulate/mockClock.ts
```diff
@@ -26,7 +33,10 @@ export function mockClock() {
      * invokation (the start of the test).
      */
     timeStamp: (duration: number) => (timeStampStart + duration) as TimeStamp,
-    tick: (ms: number) => jasmine.clock().tick(ms),
+    tick: (ms: number) => {
```
gogusarov, line 42: `question`

if we want to emulate somewhat `honest` tick here shouldn't we flush microtasks twice? the second one to clear microtasks set in timer's callbacks

gogusarov (approved):

rgaignault (approved):

packages/browser-rum-core/src/domain/view/trackViews.ts
BenoitZugmeyer: That function actually comes from https://github.com/DataDog/browser-sdk/blob/main/packages/browser-core/src/tools/queueMicrotask.ts. This function is currently used without `getZoneJsOriginalValue`, and AFAIK it's fine, so I'd rather keep it like this (and apply `getZoneJsOriginalValue` later if it causes an issue)

packages/browser-core/test/emulate/mockClock.ts
BenoitZugmeyer: Something like this?

```ts
pendingMicroTasks.splice(0).forEach((task) => task())
jasmine.clock().tick(ms)
pendingMicroTasks.splice(0).forEach((task) => task())
```

gogusarov (approved):

gogusarov (approved):

PR #3598, opened 2025-06-05
adlrb/rum 10223 locale ootb

mormubis:
- **feat: add locale and timezone information to the view collection**
- **fix: add device info to the right place**
- **fix: types for raw rum event**
- **test: update submodule**
- **refactor: update types**

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewCollection.ts
+++ b/packages/rum-core/src/domain/view/viewCollection.ts
@@ -95,7 +95,6 @@ function processViewUpdate(
       cumulative_layout_shift: view.commonViewMetrics.cumulativeLayoutShift?.value,
       cumulative_layout_shift_time: toServerDuration(view.commonViewMetrics.cumulativeLayoutShift?.time),
       cumulative_layout_shift_target_selector: view.commonViewMetrics.cumulativeLayoutShift?.targetSelector,
-      current_locale: navigator.language,
       first_byte: toServerDuration(view.initialViewMetrics.navigationTimings?.firstByte),
       dom_complete: toServerDuration(view.initialViewMetrics.navigationTimings?.domComplete),
       dom_content_loaded: toServerDuration(view.initialViewMetrics.navigationTimings?.domContentLoaded),
@@ -117,7 +116,6 @@ function processViewUpdate(
       load_event: toServerDuration(view.initialViewMetrics.navigationTimings?.loadEvent),
       loading_time: discardNegativeDuration(toServerDuration(view.commonViewMetrics.loadingTime)),
       loading_type: view.loadingType,
-      locales: navigator?.languages ?? [],
       long_task: {
         count: view.eventCounts.longTaskCount,
       },
@@ -126,7 +124,6 @@ function processViewUpdate(
         count: view.eventCounts.resourceCount,
       },
       time_spent: toServerDuration(view.duration),
-      timezone: new Intl.DateTimeFormat().resolvedOptions().timeZone,
     },
     display: view.commonViewMetrics.scroll
       ? {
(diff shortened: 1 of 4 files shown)


packages/core/src/tools/utils/polyfills.ts
```diff
@@ -21,3 +21,17 @@ export function objectValues<T = unknown>(object: { [key: string]: T }) {
 export function objectEntries<T = unknown>(object: { [key: string]: T }): Array<[string, T]> {
   return Object.entries(object)
 }
+
+export function getTimeZone() {
+  // Use Intl.DateTimeFormat().resolvedOptions().timeZone to get the time zone
+  // This is supported in all modern browsers and Node.js
+  // Making sure to handle cases where the time zone cannot be determined
```
thomas-lebeau: **💬 suggestion:** ‏ I think comments are useful when they are adding information not available in the code (or when the code is really hard to read and cannot be simplified)

In here, I think the code is pretty obvious what the code do, the added information would be `This is supported in all modern browsers and Node.js`, but:
- we don't support Node environment, so it's not necessary to precise
- "modern browser" is quite vague. It seems it is supported on all browser we officially support, in that case I would not mention anything specific about browser support.

packages/core/src/tools/utils/polyfills.ts
```diff
@@ -21,3 +21,17 @@ export function objectValues<T = unknown>(object: { [key: string]: T }) {
...
+  try {
+    const intl = new Intl.DateTimeFormat()
+
+    return intl.resolvedOptions().timeZone
+  } catch {
+    // Fallback to a default value if the time zone cannot be determined
```
thomas-lebeau: **💬 suggestion:** ‏ same here, this comment does not really add value, no?
```suggestion
```

packages/core/src/tools/utils/polyfills.ts
thomas-lebeau, line 1: **💭 thought:** ‏ This does not sounds like a polyfill, is it? Also, It looks like it's only being used in rum (i.e. not logs) so I think we could mo it to `rum-core` (maybe `rum-core/src/browser/timezone.ts`), even though it looks like tree shaking is working properly and no bytes are added to the logs bundle, so I would be fine keeping it in the `core` package

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -138,6 +138,11 @@ function processViewUpdate(
     privacy: {
       replay_level: configuration.defaultPrivacyLevel,
     },
+    device: {
+      locale: navigator?.language,
+      locales: navigator?.languages ?? [],
```
BenoitZugmeyer: **suggestion:** you can assume `navigator` is defined
Also, no need to fallback on an empty array. Undefined is fine, the field will just not be sent.

```suggestion
      locale: navigator.language,
      locales: navigator.languages,
```

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -137,6 +137,11 @@ export interface RawRumViewEvent {
       start_session_replay_recording_manually: boolean
     }
   }
+  device?: {
+    locale?: string
+    locales?: string[] | readonly string[]
```
BenoitZugmeyer: **suggestion:** mentioning both types seems superfluous
```suggestion
    locales?: string[]
```

mormubis: `navigator.languages` is just `readonly string[]` and cannot be assigned to mutable array.

At the same time, if we want to test it out, it's fine to have a regular array `['a', 'b']` vs `['a', 'b'] as const`.

What do you think?

packages/core/src/tools/utils/polyfills.ts
mormubis: Moved!

packages/core/src/tools/utils/polyfills.ts
mormubis: Removed!

packages/rum-core/src/rawRumEvent.types.ts
BenoitZugmeyer: I see! Yeah, I think we could just keep `readonly string[]` then

rgaignault (approved):
LGTM 👍

PR #3600, opened 2025-06-05
👷release: do not try to update lockfiles of ignore app

bcaudan:
## Motivation

Following #3595, still an issue during the release command:
```
The following paths are ignored by one of your .gitignore files:
test/apps/allowed-tracking-origin
hint: Use -f if you really want to add them.
hint: Disable this message with "git config set advice.addIgnoredFile false"
```

## Changes

Skip lockfile update for app in `.gitignore` file 

## Test instructions

Tested locally by adding

```sh
cmd_foo() {
  # keep test apps lockfiles up to date
  for app_package_path in $(git ls-files test/apps/*/package.json); do
    app_dir=$(dirname "$app_package_path");
    cd "$app_dir"
    yarn up
    git add yarn.lock
    cd - > /dev/null
  done
}
```
and running
```
yarn build:testing-extensions
./scripts/cli foo
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/cli
+++ b/scripts/cli
@@ -77,10 +77,8 @@ cmd_version () {
   node ./scripts/release/update-peer-dependency-versions.js
   lerna run pack --stream
   # keep test apps lockfiles up to date
-  cd test/apps
-  for app_dir in *; do
-    # unless the app is ignored
-    cat .gitignore | grep -q "$app_dir" && continue
+  for app_package_path in $(git ls-files test/apps/*/package.json); do
+    app_dir=$(dirname "$app_package_path");
     cd "$app_dir"
     yarn up
     git add yarn.lock


scripts/cli
```diff
@@ -77,7 +77,10 @@ cmd_version () {
   node ./scripts/release/update-peer-dependency-versions.js
   lerna run pack --stream
   # keep test apps lockfiles up to date
-  for app_dir in test/apps/*; do
+  cd test/apps
+  for app_dir in *; do
+    # unless the app is ignored
+    cat .gitignore | grep -q "$app_dir" && continue
```
BenoitZugmeyer: **suggestion:** useless use of cat spotted
```suggestion
    # unless the app is ignored
    grep -q "$app_dir" .gitignore && continue
```

Although it would be better to leverage git to list tracked apps, instead of trying to parse the .gitignore ourself. Something like:

```bash
for app_package_path in `git ls-files test/apps/*/package.json`; do
  app_dir=$(dirname $app_package_path);
  ...
done
```

bcaudan: nicer, indeed!

PR #3602, opened 2025-06-05
🔈 [RUM-10284] collect telemetry before starting the session manager

BenoitZugmeyer:
## Motivation

We observed that in some condition, the SessionManager crashes. While the exception is caught and buffered in-memory for telemetry, the telemetry event itself is not sent to Datadog.

The goal of this PR is to be able to cover more ground in our telemetry monitoring and caught exceptions that are coming from the Session Manager.

## Changes

Changes are kind of significative, so bare with me. As always, please review commit by commit.

The overall strategy is first to refactor Telemetry code so both Logs and RUM behave the same. This implies using a separate batch for telemetry, which was not the case for RUM.

With telemetry logic refactored, we can do some modifications that allows to send telemetry earlier, without waiting for startLogs or startRum to finish.

## Test instructions

* Open the sandbox in Chrome
* Using chrome devtools, edit the session cookie to add `&lock=1` at the end. This makes the Logs and RUM SDKs crash (specific changes for this crash will come in later PRs).
* Refresh the page

* Check that telemetry errors are listed in the devtools extension.
* Also, check that those telemetry errors are sent to Datadog intake in the Network tab.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/core/src/domain/telemetry/telemetry.spec.ts
+++ b/packages/core/src/domain/telemetry/telemetry.spec.ts
@@ -175,6 +175,26 @@ describe('telemetry', () => {
       })
       expect(notifySpy.calls.mostRecent().args[0].foo).not.toBeDefined()
     })
+
+    it('allows adding context progressively', () => {
+      const { telemetry, notifySpy } = startAndSpyTelemetry()
+      telemetry.setContextProvider('application.id', () => 'bar')
+      callMonitored(() => {
+        throw new Error('foo')
+      })
+      telemetry.setContextProvider('session.id', () => '123')
+      callMonitored(() => {
+        throw new Error('bar')
+      })
+
+      expect(notifySpy.calls.argsFor(0)[0]['application.id']).toEqual('bar')
+      expect(notifySpy.calls.argsFor(1)[0]).toEqual(
+        jasmine.objectContaining({
+          'application.id': 'bar',
+          'session.id': '123',
+        })
+      )
+    })
   })
 
   describe('sampling', () => {


packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -45,8 +50,8 @@ export const enum TelemetryService {
 }
 
 export interface Telemetry {
-  setContextProvider: (provider: () => Context) => void
-  observable: Observable<TelemetryEvent & Context>
+  setContextProvider: (key: string, contextProvider: () => ContextValue | undefined) => void
```
amortemousque, line 53: **💭 thought:** ‏Using a new hook like `assembleTelemetry` you could get rid of this `setContextProvider`(No need to do it right now but just to keep in mind)

packages/rum-core/src/domain/startCustomerDataTelemetry.spec.ts
```diff
@@ -46,17 +46,15 @@ describe('customerDataTelemetry', () => {
     })
   }
 
-  function setupCustomerTlemertyCollection(partialConfig: Partial<RumConfiguration> = config) {
+  function setupCustomerTelemertyCollection(partialConfig: Partial<RumConfiguration> = config) {
```
amortemousque: Typo
```suggestion
  function setupCustomerTelemetryCollection(partialConfig: Partial<RumConfiguration> = config) {
```

packages/core/src/domain/telemetry/telemetry.spec.ts
```diff
@@ -154,15 +154,14 @@ describe('telemetry', () => {
     it('should be added to telemetry events', () => {
       const { telemetry, notifySpy } = startAndSpyTelemetry()
 
-      telemetry.setContextProvider(() => ({
-        foo: 'bar',
-      }))
+      telemetry.setContextProvider('foo', () => 'bar')
```
bcaudan, line 165: **💬 suggestion:** what about adding a test to illustrate "the progressively set context" case?‏

bcaudan (approved):
Very nice!

PR #3616, opened 2025-06-10
Thomas.lebeau/ddtags

thomas-lebeau:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


developer-extension/src/panel/components/tabs/eventsTab/copyEvent.ts
```diff
@@ -90,6 +90,5 @@ export function getIntakeUrlForEvent(sdkInfos: SdkInfos, event: SdkEvent) {
 
   return builder
     .build('manual', { data: 'a', bytesCount: 1 })
-    .replace(/sdk_version%3A[^%&]+/g, `sdk_version%3A${encodeURIComponent(version)}`)
```
thomas-lebeau, line 93: **🔨 warning:** Unfortunately I don't think there is a good way to replace this feature 😞

packages/core/src/domain/configuration/intakeSites.ts
```diff
@@ -13,4 +13,4 @@ export const INTAKE_SITE_EU1: Site = 'datadoghq.eu'
 export const INTAKE_SITE_US1_FED: Site = 'ddog-gov.com'
 
 export const PCI_INTAKE_HOST_US1 = 'pci.browser-intake-datadoghq.com'
-export const INTAKE_URL_PARAMETERS = ['ddsource', 'ddtags']
+export const INTAKE_URL_PARAMETERS = ['ddsource', 'dd-api-key', 'dd-request-id']
```
thomas-lebeau: **💭 thought:** ‏ `ddtags` is not a url parameter anymore, however, this `INTAKE_URL_PARAMETERS` is used to know if a request is an intake url request or not, so I needed to find a replacement. It seems obvious to use `dd-api-key`, `dd-request-id` instead.

packages/core/src/domain/contexts/tagContext.ts
```diff
@@ -0,0 +1,25 @@
+import type { Configuration } from '../configuration'
+import type { AbstractHooks } from '../../tools/abstractHooks'
+import { HookNames } from '../../tools/abstractHooks'
+
+export function startTagContext(hooks: AbstractHooks, configuration: Configuration) {
```
thomas-lebeau: **💭 thought:** ‏ I find it a bit weird to call this a context but it's consistent with [connectivityContext](https://github.com/DataDog/browser-sdk/blob/4935a3880a631a17e12a82751e4278cddd9ef25b/packages/rum-core/src/domain/contexts/connectivityContext.ts#L4)

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -113,6 +108,12 @@ function buildEndpointParameters(
 
   if (trackType === 'rum') {
     parameters.push(`batch_time=${timeStampNow()}`)
+    parameters.push(`_dd.api=${api}`)
+
+    if (retry) {
+      parameters.push(`_dd.retry_count=${retry.count}`)
+      parameters.push(`_dd.retry_after=${retry.lastFailureStatus}`)
+    }
```
BenoitZugmeyer: **nitpick:** you might be able to shave a few bytes using `parameters.push(param1, param2)`

packages/core/src/domain/configuration/intakeSites.ts
BenoitZugmeyer: There is just a slight drawback with this. `ddsource` and `ddtags` have been used since v1 of the SDK, this is why we chose them. `dd-api-key` is only used since v4.0.0. So, *if* a SDK anterior to v4.0.0 is used alongside a newer SDK, the newer SDK will not discard requests from the old SDK anymore.

Using different versions of the SDK at the same time is unsupported but happens. The risk is limited though, and if a customer ever notices this as being an issue, we can always suggest to upgrade their old version.

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -113,6 +100,12 @@ function buildEndpointParameters(
 
   if (trackType === 'rum') {
     parameters.push(`batch_time=${timeStampNow()}`)
+    parameters.push(`_dd.api=${api}`)
+
+    if (retry) {
+      parameters.push(`_dd.retry_count=${retry.count}`)
```
amortemousque: **💬 suggestion:** ‏We should probably add these fields in the rum-event-format

developer-extension/src/panel/components/tabs/eventsTab/copyEvent.ts
BenoitZugmeyer: This was meant to apply the actual sdk version on events, but since it's now inside events, it shouldn't be needed anymore!

packages/core/src/domain/contexts/tagContext.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  if (b?.ddtags) {
+    if (ddtags && ddtags.length > 0) {
+      ddtags += `,${b.ddtags}`
+    } else {
+      ddtags = b.ddtags
+    }
+  }
+
+  return ddtags
```
BenoitZugmeyer: **suggestion:** 
```ts
const tagsA = a?.ddtags
const tagsB = b?.ddtags
return tagsA && tagsB ? `${tagsA},${tagsB}` : tagsA || tagsB
```

test/e2e/scenario/logs.scenario.ts
```diff
@@ -182,6 +182,46 @@ test.describe('logs', () => {
...
+  createTest('add tags from message context')
+    .withLogs()
+    .run(async ({ intakeRegistry, flushEvents, page }) => {
+      await page.evaluate(() => {
+        window.DD_LOGS!.logger.log('hello world!', { ddtags: 'planet:earth' })
+      })
+
+      await flushEvents()
+      expect(intakeRegistry.logsEvents).toHaveLength(1)
+      expect(intakeRegistry.logsEvents[0].ddtags).toMatch(/sdk_version:(.*),planet:earth$/)
```
BenoitZugmeyer: **suggestion:** make sure we also have env, service and version

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -314,6 +316,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
      */
     batchMessagesLimit: 50,
     messageBytesLimit: 256 * ONE_KIBI_BYTE,
+    tags: buildTags(initConfiguration),
```
amortemousque: **💬 suggestion:**
Not directly related to this PR, but it might be a good opportunity to improve consistency.

Currently, `service` and `version` are set both as configuration attributes and as tags ([service](https://github.com/DataDog/browser-sdk/blob/e7f06b4f08d66671c699053e9da45e4418afc8a0/packages/core/src/domain/configuration/configuration.ts#L293), [version](https://github.com/DataDog/browser-sdk/blob/e7f06b4f08d66671c699053e9da45e4418afc8a0/packages/rum-core/src/domain/configuration/configuration.ts#L237)). Because of this, they bypass the tag sanitization process, which could lead to discrepancies between the attribute and the tag values.

Instead of building the tag list here, we could simplify the logic by applying sanitization directly:
```ts
service: sanitizeTag(initConfiguration.service),
env: sanitizeTag(initConfiguration.env),
version: sanitizeTag(initConfiguration.version),
datacenter: sanitizeTag(initConfiguration.datacenter),
```
Then, handle the concatenation in the `tagContext`. Wdyt?

packages/logs/src/domain/assembly.ts
```diff
@@ -51,6 +52,7 @@ export function startLogsAssembly(
         startTime,
       }) as DefaultLogsEventAttributes
 
+      const ddtags = combineTags(defaultLogsEventAttributes, messageContext)
```
BenoitZugmeyer: **thought:** it's a bit unfortunate that we need to do `combineTags` outside of the hook. I understand that it's because the hook is generic over RUM and Logs events. Maybe the hook could apply to any object with an optional `ddtags` property? Or maybe we could have separate hooks?

amortemousque: Agreed!
Proposal: Use an object for ddtags:
```ts
ddtags: {
  service: ...,
  version: ...,
}
```
This way, context merging works out of the box. We can stringify it later during assembly.
Wdyt?

packages/core/src/domain/configuration/intakeSites.ts
thomas-lebeau: Thanks for the context! I guess we could easily support both old/new url parameters later if we need.

packages/logs/src/domain/assembly.ts
thomas-lebeau: > Maybe the hook could apply to any object with an optional ddtags property.

Yeah, but the hook is not aware of the event / log, so it does not have access to the ddtag in the `messageContext`. We could pass it to the hook, but I'm not sure this is the intended usage.

> Use an object for ddtags

the issue with that is we lose the ability to set the same tag key twice with different value (e.g. `region:us,region:eu`) that is supported if we let customer use a string notation.

we can discuss that though, maybe we don't want this feature?

packages/core/src/domain/configuration/endpointBuilder.ts
thomas-lebeau: yes, the PR is here: https://github.com/DataDog/rum-events-format/pull/279

packages/logs/src/boot/startLogs.ts
```diff
@@ -66,6 +67,7 @@ export function startLogs(
   const accountContext = startAccountContext(hooks, configuration, 'logs')
   const globalContext = startGlobalContext(hooks, configuration, 'logs', false)
   const { stop, getRUMInternalContext } = startRUMInternalContext(hooks)
+  startTagContext(hooks, configuration)
```
amortemousque: **🔨 warning:** To allow customers to override `ddtags` with global context, we should set it before calling `startGlobalContext`
We have a test to ensure contexts precedence: https://github.com/DataDog/browser-sdk/blob/b25a3ccd3319d7f6a40d1d8e676dac9ee51c66cd/packages/logs/src/boot/startLogs.spec.ts#L286-L287‏

packages/core/src/domain/configuration/tags.ts
```diff
@@ -23,23 +26,21 @@ export function buildTags(configuration: InitConfiguration): string[] {
...
+      `Tag ${tag} doesn't meet tag requirements and will be sanitized. ${MORE_DETAILS} ${DOCS_ORIGIN}/getting_started/tagging/#defining-tags`
     )
   }
 
   // Let the backend do most of the sanitization, but still make sure multiple tags can't be crafted
   // by forging a value containing commas.
-  const sanitizedValue = rawValue.replace(/,/g, '_')
-
-  return `${key}:${sanitizedValue}`
+  return tag.replace(/,/g, '_')
```
amortemousque, line 47: **❓ question:** ‏Now that we also sanitize the tag key, is it a breaking change?

packages/core/src/domain/configuration/tags.ts
```diff
@@ -23,23 +26,21 @@ export function buildTags(configuration: InitConfiguration): string[] {
...
-export function buildTag(key: string, rawValue: string) {
+export function buildTag(key: string, rawValue?: string) {
   // See https://docs.datadoghq.com/getting_started/tagging/#defining-tags for tags syntax. Note
   // that the backend may not follow the exact same rules, so we only want to display an informal
   // warning.
-  const valueSizeLimit = TAG_SIZE_LIMIT - key.length - 1
+  const tag = rawValue ? `${key}:${rawValue}` : key
 
-  if (rawValue.length > valueSizeLimit || hasForbiddenCharacters(rawValue)) {
+  if (tag.length > TAG_SIZE_LIMIT || hasForbiddenCharacters(tag)) {
```
amortemousque, line 35: **👏 praise:** ‏Nice! Better to check all the tag indeed :)

packages/logs/src/domain/logger.ts
```diff
@@ -121,6 +124,18 @@ export class Logger {
     this.contextManager.clearContext()
   }
 
+  addTag(key: string, value?: string) {
+    this.tags.push(buildTag(key, value))
+  }
+
+  removeTagsWithKey(key: string) {
+    this.tags = this.tags.filter((tag) => tag !== key && !tag.startsWith(`${key}:`))
```
amortemousque: **💬 suggestion:** ‏Since buildTag sanitizes the key when addTag is called, we should apply the same sanitization to the key parameter here.

packages/logs/src/domain/assembly.ts
```diff
@@ -30,7 +31,7 @@ export function startLogsAssembly(
 
   lifeCycle.subscribe(
     LifeCycleEventType.RAW_LOG_COLLECTED,
-    ({ rawLogsEvent, messageContext = undefined, savedCommonContext = undefined, domainContext }) => {
+    ({ rawLogsEvent, messageContext = undefined, savedCommonContext = undefined, domainContext, ddtags }) => {
```
amortemousque: **🥜 nitpick:** ‏
```suggestion
    ({ rawLogsEvent, messageContext = undefined, savedCommonContext = undefined, domainContext, ddtags = [] }) => {
```
this will simplify the code bellow
```ts
{
    ddtags: defaultDdtags.concat(ddtags).join(','),
}
```

packages/logs/src/domain/logger.ts
```diff
@@ -34,6 +35,7 @@ export const STATUSES = Object.keys(StatusType) as StatusType[]
 // eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging, no-restricted-syntax
 export class Logger {
   private contextManager: ContextManager
+  private tags: string[]
```
amortemousque, line 41: **💬 suggestion:** ‏To explicit the tag format, it could be nice to have a Tag type: `type Tag = `${string}:${string}`

packages/core/src/domain/configuration/tags.ts
thomas-lebeau: it's not because customer did not have the possibility to add tags, and our own tags keys are already good.

packages/logs/src/domain/logger.ts
thomas-lebeau: the issues is that is should be:
```
type Tag = `${string}:${string}` | string
```

but `string` always overrides `${string}:${string}`

Screenshot 2025-07-08 at 15 13 20

However I can add a comment to clarify the tag format

mormubis: Can you try `${string}:${string} | (string & {})`?

packages/logs/src/domain/assembly.spec.ts
```diff
@@ -232,6 +232,21 @@ describe('startLogsAssembly', () => {
     })
   })
 
+  describe('ddtags', () => {
```
BenoitZugmeyer, line 235: **suggestion:** add a test with `ddtags` in the `rawLogsEvent` to document precedence:

```ts
it('ddtags attribute from the logs event is ignored', () => {
  lifeCycle.notify(LifeCycleEventType.RAW_LOG_COLLECTED, {
    rawLogsEvent: { ...DEFAULT_MESSAGE, ddtags: 'foo:bar' },
  })
  expect(serverLogs[0].ddtags).toEqual('sdk_version:test,env:test,service:service,version:1.0.0')
})
```

EDIT: I see the logs event cannot have ddtags, so feel free to dismiss my comment.

packages/logs/src/domain/logger.ts
```diff
@@ -121,6 +127,20 @@ export class Logger {
...
+  }
+
+  removeTagsWithKey(key: string) {
+    const sanitizedKey = sanitizeTag(key)
+
+    this.tags = this.tags.filter((tag) => tag !== sanitizedKey && !tag.startsWith(`${sanitizedKey}:`))
+  }
+
+  getTags() {
+    return this.tags
```
BenoitZugmeyer: **nitpick:** make sure users don't update tags in place:
```suggestion
    return this.tags.slice()
```

packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.spec.ts
```diff
@@ -18,6 +19,10 @@ describe('trackFirstHidden', () => {
     return trackFirstHidden(configuration, viewStart, eventTarget)
   }
 
+  beforeEach(() => {
+    configuration = mockRumConfiguration()
+  })
```
BenoitZugmeyer: **question:** sorry if I missed something, but why those changes with `mockRumConfiguration` ?

BenoitZugmeyer (approved):

packages/rum-core/src/domain/view/viewMetrics/trackFirstHidden.spec.ts
thomas-lebeau: It's because `mockRumConfiguration` uses `validateAndBuildConfiguration` and at one point, my PR was building tags during `validateAndBuildConfiguration` and that the `sdk_version` tag depend on the build time constant `__BUILD_ENV__SDK_VERSION__`.

In unit tests, it is defined in a [top level beforeEach](https://github.com/DataDog/browser-sdk/blob/5dc393a0b4886858a7fe334fceadc620dfc6c6a5/packages/core/test/forEach.spec.ts#L5-L5) so calling `mockRumConfiguration` outside of a `beforeEach` or a `it` was breaking a few things because `__BUILD_ENV__SDK_VERSION__` was not defined yet.

You're right that this is probably not needed anymore as the PR evolves to only pass some config property from `InitConfiguration` to `Configuration` and build the tag later during assembly. I'm reverting thoses changes

packages/logs/src/domain/logger.ts
thomas-lebeau: > Can you try ${string}:${string} | (string & {})?

That works but it's not compatible with older typescript version we're still supporting (v3.8)

PR #3617, opened 2025-06-10
✨ [RUM-246] Support prerendered pages for web vitals

BeltranBulbarellaDD:
## Motivation

Chrome released prerendering support for Chrome 108+ and Edge 108+. This means that all load-related metrics should be relative to when the page was ["activated"](https://github.com/WICG/nav-speculation/blob/main/ua-initiated-prerendering.md#how-it-works) rather than when it started loading (for non-prerendered pages these values are the same).

## Changes

Created interface DocumentWithPrerendering.
Created function getActivationStart which uses the new activationStart.
Adjusted LCP, FCP, TTFB and firstHidden for this new calculation.

## Test instructions

All previous tests should pass, and when we mock activation start or have 2 html files where 1 prerenders the other we should see something similar to the image below.

Screenshot 2025-06-11 at 10 24 55

https://github.com/user-attachments/assets/20d9154f-559a-4672-9a53-9b7741963140

## Checklist

- [X ] Tested locally
- [ ] Tested on staging
- [X ] Added unit tests for this change.
- [X ] Added e2e/integration tests for this change.


developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
BeltranBulbarellaDD, line 1: I'm fairly certain CI won't pass until I add this in RUM-Events-Format.

PR #3620, opened 2025-06-10
♻️ [RUM-10311] Use hooks for Logs SDK to decouple user

amortemousque:
## Motivation

Use hooks for Logs SDK to decouple user context

## Changes

- Move `User` and `Account` types to `userContext` and `accountContext` respectively
- Buffer user context calls via `defineContextMethod` in `logsPublicApi.ts`
- Remove user context usage in `logs/assembly.ts`
- Add tests for context precedence in `startLogs.spec.ts`

## Checklist

- [x] Tested locally
- [x] Tested on staging


packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -176,331 +175,84 @@ describe('logs entry', () => {
...
-        logsPublicApi.setUserProperty('email', { foo: 'bar' })
-        const userClone = logsPublicApi.getUser()
-
-        expect(userClone.id).toEqual('123')
-        expect(userClone.name).toEqual('Adam,Smith')
-        expect(userClone.email).toEqual('[object Object]')
+      it('should call setContext', () => {
+        spyOn(userContext, 'setContext')
+        logsPublicApi.setUser(2 as any)
+        expect(userContext.setContext).toHaveBeenCalledTimes(1)
```
bcaudan, line 195: **💬 suggestion:** ‏it seems that we have traded a bunch of integration tests for a couple of implementation tests. I assume that the different features are unit tested elsewhere but it could be valuable to keep a couple of high level tests just to ensure that everything is (and will stay) wired properly.

amortemousque: I added e2e test https://github.com/DataDog/browser-sdk/pull/3620/commits/7b1cb29f0f92451cd8f55db471f169460eb7a643

bcaudan (approved):

PR #3622, opened 2025-06-11
✨ [RUM-10145] collect errors on module evaluation

BenoitZugmeyer:
## Motivation

Currently, the RUM and Logs Browser SDK are starting to collect errors when the following conditions are met:

* the SDK is loaded  
* `init()` was called  
* tracking consent was granted (either via the `trackingConsent` init parameter or when `setTrackingConsent(‘granted’)` is called)  
* a RUM View was created (RUM only, in case `trackViewsManually` is used)  
* the Remote Configuration is fetched

To improve observability coverage, we want to collect errors earlier, and the easiest way to do so is to start collecting errors on module evaluation, buffer them, and send them when RUM is ready.

This is a first step. We chose to start with runtime errors first as it was deemed to be the most valuable without much drawback, but our approach needs to be easily extendable to collect other kind of values.

Also, in the future this mechanism could be separated from the main SDK to allow customer inline this code in their HTML to cover even more ground, mitigating the performance impact of inlining the whole SDK entirely.

## Changes

We start by introducing a `BufferedObserver`, which is intended to be used for other purposes and eventually replace `BoundedBuffer`.

After a bit of refactoring in tests, we implement early runtime error collection in RUM and Logs by tracking runtime errors and putting them in a `BufferedObserver` instance.

## Test instructions

In the sandbox (which uses a sync CDN setup), throw an exception before calling init, ex:

```js
throw new Error('test')
setTimeout(() => {
  DD_RUM.init(...)
  DD_LOGS.init(...)
})
```

The error should be collected.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/core/src/domain/bufferedData.ts
+++ b/packages/core/src/domain/bufferedData.ts
@@ -8,7 +8,7 @@ export const enum BufferedDataType {
   RUNTIME_ERROR,
 }
 
-export type BufferedData = {
+export interface BufferedData {
   type: BufferedDataType.RUNTIME_ERROR
   error: RawError
 }


packages/core/src/tools/observable.ts
```diff
@@ -37,3 +47,62 @@ export function mergeObservables<T>(...observables: Array<Observable<T>>) {
...
+        this.addObserver(observer)
+      }
+    })
+
+    return subscription
+  }
+
+  /**
+   * Drop buffered data and don't buffer future data. This is to avoid leaking memory when it's not
+   * needed anymore. This is not be required in most cases, but still useful to clarify our intent
```
amortemousque: Why would it be not required in most cases?

packages/core/src/tools/observable.ts
```diff
@@ -37,3 +47,62 @@ export function mergeObservables<T>(...observables: Array<Observable<T>>) {
     return () => subscriptions.forEach((subscription) => subscription.unsubscribe())
   })
 }
+
+// eslint-disable-next-line no-restricted-syntax
+export class BufferedObservable<T> extends Observable<T> {
+  private buffer: T[] = []
+
+  constructor(private maxBufferSize: number) {
+    // no onFirstSubscribe as it makes less sense with buffered data
```
amortemousque: I don't think we need this comment.

packages/core/src/domain/error/trackRuntimeError.ts
```diff
@@ -1,39 +1,40 @@
...
-      startClocks: clocksNow(),
-      nonErrorPrefix: NonErrorPrefix.UNCAUGHT,
-      source: ErrorSource.SOURCE,
-      handling: ErrorHandling.UNHANDLED,
-    })
-    errorObservable.notify(rawError)
-  }
-  const { stop: stopInstrumentingOnError } = instrumentOnError(handleRuntimeError)
-  const { stop: stopInstrumentingOnUnhandledRejection } = instrumentUnhandledRejection(handleRuntimeError)
+export function trackRuntimeError() {
```
amortemousque, line 13: **💬 suggestion:** ‏Why not using BufferedObservable directly in here.

packages/core/src/domain/earlyDataCollection.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function startEarlyDataCollection(trackRuntimeErrorImpl = trackRuntimeError) {
+  const observable = new BufferedObservable<EarlyData>(BUFFER_LIMIT)
+
+  const runtimeErrorSubscription = trackRuntimeErrorImpl().subscribe((error) => {
+    observable.notify({
+      type: EarlyDataType.RUNTIME_ERROR,
+      error,
+    })
+  })
```
amortemousque: **💬 suggestion:** 
Following up on my other comment, I don’t think we need an EarlyData concept. trackRuntimeErrorImpl can directly return an object with a type.

Later, when we have multiple observables, we can just use mergeObservables to combine them :)
Wdyt?

BenoitZugmeyer: I'd be fine with it, but how would you name the "object with a type"?

packages/core/src/domain/error/trackRuntimeError.ts
BenoitZugmeyer: I tried, but:
* It makes a circular dependency between `trackRuntimeError` and `bufferedData`
* The "Buffered Data" concept leaks everywhere in `trackRuntimeError`. It makes it more complex as it's not scoped to `rawError` anymore.

packages/core/src/tools/observable.ts
BenoitZugmeyer: I clarified the comment.

packages/core/src/tools/observable.ts
```diff
@@ -37,3 +47,58 @@ export function mergeObservables<T>(...observables: Array<Observable<T>>) {
...
+      }
+    })
+
+    return subscription
+  }
+
+  /**
+   * Drop buffered data and don't buffer future data. This is to avoid leaking memory when it's not
+   * needed anymore. This can be seen as a performance optimization, and things will work probably
+   * work even if this method isn't called, but still useful to clarify our intent and lowering our
```
rgaignault: nitpick :
```suggestion
   * needed anymore. This can be seen as a performance optimization, and things will work probably
   * even if this method isn't called, but still useful to clarify our intent and lowering our
```

packages/core/src/domain/bufferedData.ts
```diff
@@ -0,0 +1,32 @@
+import { BufferedObservable } from '../tools/observable'
+import type { RawError } from './error/error.types'
+import { trackRuntimeError } from './error/trackRuntimeError'
+
+const BUFFER_LIMIT = 500
```
rgaignault, line 5: 💭 suggestion : Could we put a single const for the buffer limit or is it expected since the other is deprecated?
(see boundedBuffer.ts)

BenoitZugmeyer: I'd like to keep `BufferedObservable` as generic as possible to encourage using it in more places. But I could add a default value in the `BufferedObservable` constructor, to keep it configurable while not having to define a limit every single time when it doesn't matter much. WDYT?

packages/core/src/domain/error/trackRuntimeError.ts
amortemousque: Hum, fine by me! Let’s move forward and maybe refine the design later as this pattern is used more widely :)

packages/core/src/tools/observable.ts
```diff
@@ -37,3 +47,58 @@ export function mergeObservables<T>(...observables: Array<Observable<T>>) {
...
+   * memory impact.
+   */
+  unbuffer() {
```
bcaudan, line 99: **❓ question:** ‏could the `unbuffer` be done without an explicit call after the first `subscribe`?

bcaudan (approved):
Great work!

packages/core/src/tools/observable.ts
BenoitZugmeyer: It could for now, as it is subscribed only once. But in the future I expect this observable to be used in multiple places where the buffer is leveraged (ex: we could imagine using the observable to generate resources, actions... not just errors).

PR #3629, opened 2025-06-12
[IR-39238] Add session diagnostic telemetry and improve configuration timing

BeltranBulbarellaDD:
## Motivation

To diagnose and monitor session-related issues from incident-39238, we need better telemetry and debugging information.

## Changes

**Session Diagnostic Telemetry:**
- Added `addTelemetryDebug` for "Unexpected session state" in `sessionManager.ts` with detailed session information:
  - Raw session data and synthetics test detection
  - Time calculations (timeFromCreated, timeToExpire, differenceCreatedToExpire)
- Added `addTelemetryDebug` for "Aborted session operation after max lock retries" in `sessionStoreOperations.ts`
- Added `usr.anonymous_id` to telemetry context provider in `startRum.ts` for better user tracking

**Workaround issues:**
- Wrapped `addTelemetryConfiguration` calls in `setTimeout` in both `preStartLogs.ts` and `preStartRum.ts` to avoid missing application.id / session.id on telemetry configuration
- Avoid crashing the SDK on buildSessionContext to avoid duplicate telemetry logs

**Code Organization:**
- Moved `intakeSites` file up one directory level to resolve circular dependency issues
- Updated all related imports


packages/core/src/tools/intakeConstants.ts
```diff
@@ -1,3 +1,6 @@
+// Intake site constants used by telemetry
```
cy-moi: 🥜 nit: I don't think they are used by telemetry only

packages/core/src/tools/intakeConstants.ts
```diff
@@ -1,3 +1,6 @@
...
+// These are defined here to avoid circular dependencies between telemetry and configuration modules
+
```
bcaudan: **💬 suggestion:** I don't think that this comment add much value:

- we should probably not list the usage of a file, the constants are currently not used only by telemetry and it may change over time 
- we don't really need to document the circular dependency issue that we faced, that could have been just a comment on your PR or a mention in the PR description, to explain why you needed to move the file

```suggestion
```

cy-moi (approved):

packages/core/src/tools/intakeConstants.ts
bcaudan, line 1: **🥜 nitpick:** ‏IMO, intake sites are still domain related so I'd keep them in `domain` rather than moving them in `tools`

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -88,11 +89,17 @@ export function startSessionManager<TrackingType extends string>(
   trackResume(configuration, () => sessionStore.restartSession())
 
   function buildSessionContext() {
+    const session = sessionStore.getSession()
+    if (!session) {
+      addTelemetryDebug('Unexpected session state', {
+        sessionStore: sessionStore.getSession(),
```
cy-moi: 💬 Suggestion: not sure about this new change. I don't think we can get the session when there is no session.

BeltranBulbarellaDD: Should I move it right below the session? Line 92-93?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -88,11 +90,17 @@ export function startSessionManager<TrackingType extends string>(
   trackResume(configuration, () => sessionStore.restartSession())
 
   function buildSessionContext() {
+    const session = sessionStore.getSession()
+
+    addTelemetryDebug('Session Cookie', {
```
bcaudan: **💬 suggestion:** ‏we should send this log only when there is the issue though, no?

bcaudan: **💬 suggestion:** ‏we could use a more informative message, something like "Session context built with undefined session"

BeltranBulbarellaDD: But don't we get back to the issue that if it's undefined we don't have anything to send?

bcaudan: FMU, with the lock issue, `sessionStore.getSession()` is undefined because we did not manage to get our own lock but `retrieveSessionCookie` would get the cookie value.
So we could send a debug log only when the issue is about to happen and have cookie info in the log.

packages/core/src/domain/synthetics/syntheticsWorkerValues.ts
```diff
@@ -26,3 +26,7 @@ export function getSyntheticsResultId(): string | undefined {
     (window as BrowserWindow)._DATADOG_SYNTHETICS_RESULT_ID || getInitCookie(SYNTHETICS_RESULT_ID_COOKIE_NAME)
   return typeof value === 'string' ? value : undefined
 }
+
+export function isSyntheticsTest(): boolean {
+  return Boolean(getSyntheticsTestId() && getSyntheticsResultId())
+}
```
bcaudan, line 32: **💬 suggestion:** nice! what about using it in synteticsContext as well?‏

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -68,7 +69,9 @@ export function createPreStartStrategy(
 
       // Expose the initial configuration regardless of initialization success.
       cachedInitConfiguration = initConfiguration
-      addTelemetryConfiguration(serializeLogsConfiguration(initConfiguration))
+      setTimeout(() => {
```
bcaudan, line 73: **💬 suggestion:** 
‏
```suggestion
      // FIXME temporary hack to avoid sending configuration without all the context data
      setTimeout(() => {
```

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -100,7 +101,9 @@ export function createPreStartStrategy(
 
     // Update the exposed initConfiguration to reflect the bridge and remote configuration overrides
     cachedInitConfiguration = initConfiguration
-    addTelemetryConfiguration(serializeRumConfiguration(initConfiguration))
+    setTimeout(() => {
```
bcaudan, line 105: **💬 suggestion:** 
‏
```suggestion
    // FIXME temporary hack to avoid sending configuration without all the context data
    setTimeout(() => {
```

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -88,11 +91,33 @@ export function startSessionManager<TrackingType extends string>(
...
+    if (!session) {
+      const rawSession = retrieveSessionCookie()
+      const debugData: any = {
+        session: rawSession,
+        isSyntheticsTest: isSyntheticsTest(),
+        createdTimestamp: rawSession?.created,
+        expireTimestamp: rawSession?.expire,
+      }
+
+      debugData.differenceCreatedToExpire = Number(rawSession?.expire) - Number(rawSession?.created)
```
bcaudan: **💬 suggestion:** ‏I think we could remove this one too, just to be safe, we could use a processor too for this one

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -186,7 +186,7 @@ export function startSessionStore<TrackingType extends string>(
   }
 
   function hasSessionInCache() {
-    return sessionCache[productKey] !== undefined
+    return sessionCache && sessionCache[productKey] !== undefined
```
thomas-lebeau: **🥜 nitpick:** ‏ 
```suggestion
    return sessionCache?.[productKey] !== undefined
```

thomas-lebeau (approved):

PR #3630, opened 2025-06-13
Beltran.bulbarella/fix stale cookie

BeltranBulbarellaDD:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/core/src/domain/session/sessionStoreOperations.ts
```diff
@@ -20,15 +23,37 @@ export function processSessionStoreOperations(
...
   const { isLockEnabled, persistSession, expireSession } = sessionStoreStrategy
-  const persistWithLock = (session: SessionState) => persistSession({ ...session, lock: currentLock })
+  const persistWithLock = (session: SessionState) =>
+    persistSession({ ...session, lock: currentLock, lockTimestamp: String(dateNow()) })
   const retrieveStore = () => {
     const session = sessionStoreStrategy.retrieveSession()
-    const lock = session.lock
+    let lock = session.lock
+
+    if (lock && session.lockTimestamp) {
```
johnHackworth, line 32: I think we should also delete the lock if `(lock && ! session.lockTimestamp)` so we clear the cookies that are already in a locked state before this change is deployed

PR #3637, opened 2025-06-16
🐛 [RUM-10366] mitigate cookie dead lock

BenoitZugmeyer:
## Motivation

In Incident 39238, we observed that the cookie lock is sometimes not removed, and it makes the SDK crash on init until the cookie expires, which could be one year.

## Changes

This PR mitigates this by adding a creation date in the lock so we can ignore it if it was created too long ago.

## Test instructions

* Open the sandbox
* Edit the `_dd_s` cookie to add `&lock=1`
* Refresh the page
* SDKs should successfully send data

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/core/src/domain/session/sessionStoreOperations.ts
```diff
@@ -118,3 +120,12 @@ function next(sessionStore: SessionStoreStrategy) {
     processSessionStoreOperations(nextOperations, sessionStore)
   }
 }
+
+export function createLock(): string {
+  return generateUUID() + LOCK_SEPARATOR + timeStampNow()
```
mormubis, line 129: Why do we use a UUID for the lock? Does the UUID any meaning?

packages/core/src/domain/session/sessionStoreOperations.ts
```diff
@@ -11,6 +13,12 @@ type Operations = {
 
 export const LOCK_RETRY_DELAY = 10
 export const LOCK_MAX_TRIES = 100
+
+// Locks should be hold for a few milliseconds top, just the time it takes to read and write a
+// cookie. Using one second should be enough in most situations.
+export const LOCK_EXPIRATION_DELAY = ONE_SECOND
```
thomas-lebeau, line 20: **💭 thought:** ‏ It should be enough, but is it harmful to use a longer delay? 
Should we send telemetry when we discard the expired lock, with the expired time so we can monitor it?

thomas-lebeau (approved):

packages/core/src/domain/session/sessionStoreOperations.ts
BenoitZugmeyer: If the lock was a constant value (ex: `lock=true`), when two instances of the SDK tried to lock at the same time, both would consider the lock to be successful and proceed to modify the cookie.

With a random lock id, each SDK instance write the lock then read the cookie, and consider it to be a success only if it's still the same value.

packages/core/src/domain/session/sessionStoreOperations.ts
BenoitZugmeyer: > It should be enough, but is it harmful to use a longer delay?

The longer the delay, the less effective the mitigation would be.

> Should we send telemetry when we discard the expired lock, with the expired time so we can monitor it?

What would be the purpose? Let's say we observe that we frequently discard the lock when it's 2 seconds old. Should we increase the lock expiration delay, and make the SDK stop in this case? Let's talk about it.

packages/core/src/domain/session/sessionStoreOperations.ts
```diff
@@ -118,3 +120,12 @@ function next(sessionStore: SessionStoreStrategy) {
...
   }
 }
+
+export function createLock(): string {
+  return generateUUID() + LOCK_SEPARATOR + timeStampNow()
+}
+
+function isLockExpired(lock: string) {
+  const [, timeStamp] = lock.split(LOCK_SEPARATOR)
+  return !timeStamp || elapsed(Number(timeStamp) as TimeStamp, timeStampNow()) > LOCK_EXPIRATION_DELAY
```
thomas-lebeau, line 130: **❓ question:** ‏ shouldn't we check the uuid too? otherwise we may consider a lock expired when it's a valid lock from another tab, no?

packages/core/src/domain/session/sessionStoreOperations.ts
thomas-lebeau: hmm, you're right, I was wondering if this can give us clue about what is going wrong. but maybe not.

PR #3640, opened 2025-06-17
🐛 [RUM-5645] Add a ff to avoid using "fetch keepalive"

rgaignault:
## Motivation

Currently, the Browser SDK uses [fetch keepalive](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#keepalive) to send data to Datadog when it is supported (see [code](https://github.com/DataDog/browser-sdk/blob/73681aa8ba8de42d4b91dbd2d6a81679b1ef22b5/packages/core/src/transport/httpRequest.ts#L104)).

Because browsers enforce a limit on the amount of data that can be sent by “fetch keepalive” requests at the same time, it can impact some websites when it also use “fetch keepalive” 
 

## Changes

We could avoid using fetch keepalive to send data to Datadog when the page is sending request. For the moment we want to test it with a a feature flag

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/transport/httpRequest.spec.ts
+++ b/packages/core/src/transport/httpRequest.spec.ts
@@ -372,7 +372,7 @@ describe('httpRequest with AVOID_FETCH_KEEPALIVE feature flag', () => {
 
   it('should use regular fetch (without keepalive) when feature flag is enabled', async () => {
     addExperimentalFeatures([ExperimentalFeature.AVOID_FETCH_KEEPALIVE])
-    request = createHttpRequest(endpointBuilder, BATCH_BYTES_LIMIT, noop)
+    request = createHttpRequest([endpointBuilder], BATCH_BYTES_LIMIT, noop)
 
     request.send({ data: '{"foo":"bar"}', bytesCount: 10 })
     await interceptor.waitForAllFetchCalls()
@@ -387,7 +387,7 @@ describe('httpRequest with AVOID_FETCH_KEEPALIVE feature flag', () => {
       pending('no fetch keepalive support')
     }
 
-    request = createHttpRequest(endpointBuilder, BATCH_BYTES_LIMIT, noop)
+    request = createHttpRequest([endpointBuilder], BATCH_BYTES_LIMIT, noop)
 
     request.send({ data: '{"foo":"bar"}', bytesCount: 10 })
     await interceptor.waitForAllFetchCalls()


packages/core/src/transport/httpRequest.ts
```diff
@@ -38,8 +39,13 @@ export function createHttpRequest(
...
   const retryState = newRetryState()
-  const sendStrategyForRetry = (payload: Payload, onResponse: (r: HttpResponse) => void) =>
-    fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload, onResponse)
+  const sendStrategyForRetry = (payload: Payload, onResponse: (r: HttpResponse) => void) => {
+    if (isExperimentalFeatureEnabled(ExperimentalFeature.AVOID_FETCH_KEEPALIVE)) {
+      fetchStrategy(endpointBuilder, payload, onResponse)
+    } else {
+      fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload, onResponse)
+    }
+  }
```
thomas-lebeau: **❓ question:** ‏ If I understand correctly the description, we want to use `fetch` when the first try using fetch-keepalive failed, correct?

Then I think unlike `sendStrategyForRetry` name suggest, it's also used for the first try too. Instead you should do something like this:
```suggestion
  const sendStrategyForRetry = (payload: Payload, onResponse: (r: HttpResponse) => void) => {
    if (isExperimentalFeatureEnabled(ExperimentalFeature.AVOID_FETCH_KEEPALIVE) && payload.retry?.count) {
      fetchStrategy(endpointBuilder, payload, onResponse)
    } else {
      fetchKeepAliveStrategy(endpointBuilder, bytesLimit, payload, onResponse)
    }
  }
```

packages/core/src/transport/httpRequest.spec.ts
```diff
@@ -263,3 +264,48 @@ describe('httpRequest intake parameters', () => {
...
+    requests = interceptor.requests
+    endpointBuilder = mockEndpointBuilder(ENDPOINT_URL)
+  })
+
+  afterEach(() => {
+    resetExperimentalFeatures()
+  })
+
+  it('should use regular fetch (without keepalive) when feature flag is enabled', async () => {
+    addExperimentalFeatures([ExperimentalFeature.AVOID_FETCH_KEEPALIVE])
```
BenoitZugmeyer, line 374: **suggestion:** just use `mockExperimentalFeatures`, so you don't need to reset them

BenoitZugmeyer (approved):

packages/core/src/transport/httpRequest.ts
rgaignault: Thanks for your comment ! 
Why would we want to not use the normal fetchStrategy for the first try ? The original issue was related to the buffer limit between customer fetchKeepAlive and our. The goal is to avoid any request error even for the first try. 
But I agree the description of my pr could be a bit misleading let me change that 👌 (I took this name from the function name, we could also probably just rename it sendStrategy instead of sendStrategyForRetry)

Let me know what you think

thomas-lebeau: 👍 ok, if the goal is to unconditionally use fetch instead of fetch-keepalive, then your PR is all good!

thomas-lebeau (approved):

PR #3642, opened 2025-06-17
🐛 [RUM-146] Modify label behavior in dead click logic

rgaignault:
## Motivation

We discard dead clicks happening on input elements. But in some cases the click happens on the label 

## Changes

We should probably find a way to discard it as well.

## Test instructions

How to reproduce 

```
<input type=”file” id=”foo” />
<label for=”foo”>click me</label>
```
in html file and yarn dev

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/action/computeFrustration.ts
+++ b/packages/rum-core/src/domain/action/computeFrustration.ts
@@ -67,5 +67,20 @@ export function isDead(click: Click) {
   if (click.hasPageActivity || click.getUserActivity().input || click.getUserActivity().scroll) {
     return false
   }
-  return !click.event.target.matches(DEAD_CLICK_EXCLUDE_SELECTOR)
+
+  const target = click.event.target
+
+  if (target.matches(DEAD_CLICK_EXCLUDE_SELECTOR)) {
+    return false
+  }
+
+  if (target.tagName === 'LABEL' && target.hasAttribute('for')) {
+    const forId = target.getAttribute('for')
+    const referencedElement = document.getElementById(forId!)
+    if (referencedElement && referencedElement.matches(DEAD_CLICK_EXCLUDE_SELECTOR)) {
+      return false
+    }
+  }
+
+  return true
 }
--- a/packages/rum-core/src/domain/action/computeFrustration.spec.ts
+++ b/packages/rum-core/src/domain/action/computeFrustration.spec.ts
@@ -172,4 +172,34 @@ describe('isDead', () => {
       ).toBe(expected)
     })
   }
+
+  describe('label elements', () => {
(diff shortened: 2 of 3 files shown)


packages/rum-core/src/domain/action/computeFrustration.ts
```diff
@@ -67,5 +67,20 @@ export function isDead(click: Click) {
...
+
+  if (target.tagName === 'LABEL' && target.hasAttribute('for')) {
+    const forId = target.getAttribute('for')
+    const referencedElement = document.getElementById(forId!)
+    if (referencedElement && referencedElement.matches(DEAD_CLICK_EXCLUDE_SELECTOR)) {
+      return false
+    }
+  }
+
+  return true
```
amortemousque: **💬 suggestion:** ‏We can simplify: 
```ts

  let target: Element | null = click.event.target

  if (target.tagName === 'LABEL' && target.hasAttribute('for')) {
    target = document.getElementById(target.getAttribute('for')!)
  }

  return target && !target.matches(DEAD_CLICK_EXCLUDE_SELECTOR)
```

packages/rum-core/src/domain/action/computeFrustration.ts
```diff
@@ -67,5 +67,12 @@ export function isDead(click: Click) {
...
   }
-  return !click.event.target.matches(DEAD_CLICK_EXCLUDE_SELECTOR)
+
+  let target: Element | null = click.event.target
+
+  if (target.tagName === 'LABEL' && target.hasAttribute('for')) {
+    target = document.getElementById(target.getAttribute('for')!)
+  }
+
+  return target && !target.matches(DEAD_CLICK_EXCLUDE_SELECTOR)
```
mormubis: **🥜 nitpick:** ‏ previous signature was `boolean`, the new implementation is `null | boolean`

Are we good with this?

rgaignault: Yeah I think we are good with this since a null would return a falsy.  And that's expected since the only way a label could be considered not a deadclick is with a correct linked attribute such as button/file...

packages/rum-core/src/domain/action/computeFrustration.ts
rgaignault: With your code if target is not defined we return false since null is a falsy. That's not what we want right ? Am I missing something ?

packages/rum-core/src/domain/action/computeFrustration.ts
rgaignault: good catch 👌 I answered to Aymeric comment, indeed now we discard deadclick if the target is not found. I am not sure this would be expected

PR #3643, opened 2025-06-18
♻️ [RUM-10407] Use hooks for Logs SDK to decouple session

amortemousque:
## Motivation

Use hooks for Logs SDK to decouple session

## Changes

- create startSessionContext module which uses hooks
- remove session related logic from assembly
- Add tests for context precedence in `startLogs.spec.ts`

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/logs/src/domain/contexts/internalContext.spec.ts
amortemousque, line 1: Not related to this PR, but I simplified `internalContext` tests by using the newly created `createLogsSessionManagerMock`

cy-moi (approved):

PR #3648, opened 2025-06-19
✨ [RUM-10415] Add support for privacy plugin extracted data for masking

cy-moi:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/action/actionCollection.ts
+++ b/packages/rum-core/src/domain/action/actionCollection.ts
@@ -11,6 +11,13 @@ import type { DefaultRumEventAttributes, Hooks } from '../hooks'
 import type { RumMutationRecord } from '../../browser/domMutationObservable'
 import type { ActionContexts, ClickAction } from './trackClickActions'
 import { trackClickActions } from './trackClickActions'
+import {
+  addAllowlistObserver,
+  createActionAllowList,
+  maskAutoActionName,
+} from './privacy/allowedDictionary'
+import type { AllowedDictionary } from './privacy/allowedDictionary'
+import { ActionNameSource } from './getActionNameFromElement'
 
 export type { ActionContexts }
 
@@ -31,9 +38,12 @@ export function startActionCollection(
   windowOpenObservable: Observable<void>,
   configuration: RumConfiguration
 ) {
-  lifeCycle.subscribe(LifeCycleEventType.AUTO_ACTION_COMPLETED, (action) =>
-    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action))
-  )
+  const actionNameDictionary = createActionAllowList()
+  addAllowlistObserver(actionNameDictionary)
+
+  lifeCycle.subscribe(LifeCycleEventType.AUTO_ACTION_COMPLETED, (action) => {
+    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action, actionNameDictionary))
+  })
 
   hooks.register(HookNames.Assemble, ({ startTime, eventType }): DefaultRumEventAttributes | SKIPPED => {
     if (
@@ -69,14 +79,19 @@ export function startActionCollection(
 
   return {
     addAction: (action: CustomAction) => {
-      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action))
+      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action, actionNameDictionary))
     },
     actionContexts,
     stop,
   }
 }
 
-function processAction(action: AutoAction | CustomAction): RawRumEventCollectedData<RawRumActionEvent> {
+function processAction(
+  action: AutoAction | CustomAction,
+  actionNameDictionary: AllowedDictionary
+): RawRumEventCollectedData<RawRumActionEvent> {
+  const { name: updatedName, masked } = maskAutoActionName(action.name, actionNameDictionary.allowlist)
+
   const autoActionProperties = isAutoAction(action)
     ? {
         action: {
@@ -99,14 +114,22 @@ function processAction(action: AutoAction | CustomAction): RawRumEventCollectedD
           action: {
             target: action.target,
             position: action.position,
-            name_source: action.nameSource,
+            name_source: masked ? ActionNameSource.MASK_DISALLOWED : action.nameSource,
           },
         },
       }
-    : undefined
+    : masked
+      ? {
+          _dd: {
+            action: {
+              name_source: ActionNameSource.MASK_DISALLOWED,
+            },
+          },
+        }
+      : undefined
   const actionEvent: RawRumActionEvent = combine(
     {
-      action: { id: generateUUID(), target: { name: action.name }, type: action.type },
+      action: { id: generateUUID(), target: { name: updatedName }, type: action.type },
       date: action.startClocks.timeStamp,
       type: RumEventType.ACTION as const,
     },
--- a/packages/rum-core/src/domain/action/getActionNameFromElement.ts
+++ b/packages/rum-core/src/domain/action/getActionNameFromElement.ts
@@ -14,8 +14,9 @@ export const enum ActionNameSource {
   TEXT_CONTENT = 'text_content',
   STANDARD_ATTRIBUTE = 'standard_attribute',
   BLANK = 'blank',
+  MASK_DISALLOWED = 'mask_disallowed',
 }
-type ActionName = {
+export type ActionName = {
   name: string
   nameSource: ActionNameSource
 }
--- a/packages/rum-core/src/domain/action/privacy/allowedDictionary.spec.ts
+++ b/packages/rum-core/src/domain/action/privacy/allowedDictionary.spec.ts
@@ -1,0 +1,114 @@
+import {
+  addAllowlistObserver,
+  createActionAllowList,
+  maskAutoActionName,
+  SPLIT_REGEX,
+  updateDictionaryOnDDAllowChange,
+} from './allowedDictionary'
+
+const TEST_STRINGS = {
+  EMOJI: '💥',
+  EMOJI_WITH_NUMBERS: '💥123',
+  SPECIAL_CHARS: '$$$',
+  SPECIAL_CHARS_WITH_NUMBERS: '$$$123',
+  HYPHENATED_SPECIAL_CHARS: '$$$-123',
+  COMPLEX_MIXED: 'test-<$>-123 hello>=42@world?',
+}
+
+describe('allowedDictionary', () => {
+  let originalWindowAllow: any
+  let originalWindowAllowObservers: any
+
+  beforeEach(() => {
+    originalWindowAllow = window.$DD_ALLOW
+    originalWindowAllowObservers = window.$DD_ALLOW_OBSERVERS
+    window.$DD_ALLOW = new Set([
+      TEST_STRINGS.EMOJI,
+      TEST_STRINGS.EMOJI_WITH_NUMBERS,
+      TEST_STRINGS.SPECIAL_CHARS,
+      TEST_STRINGS.SPECIAL_CHARS_WITH_NUMBERS,
+      TEST_STRINGS.HYPHENATED_SPECIAL_CHARS,
+      TEST_STRINGS.COMPLEX_MIXED,
+    ])
+    window.$DD_ALLOW_OBSERVERS = new Set<() => void>()
+  })
+
+  afterEach(() => {
+    window.$DD_ALLOW = originalWindowAllow
+    window.$DD_ALLOW_OBSERVERS = originalWindowAllowObservers
+  })
+
+  it('MATCH_REGEX matches words and symbols in TEST_STRINGS', () => {
+    expect(TEST_STRINGS.EMOJI.match(SPLIT_REGEX)).toBeNull()
+    expect(TEST_STRINGS.EMOJI_WITH_NUMBERS.match(SPLIT_REGEX)).toEqual(['123'])
+    expect(TEST_STRINGS.SPECIAL_CHARS.match(SPLIT_REGEX)).toBeNull()
+    expect(TEST_STRINGS.SPECIAL_CHARS_WITH_NUMBERS.match(SPLIT_REGEX)).toEqual(['123'])
+    expect(TEST_STRINGS.HYPHENATED_SPECIAL_CHARS.match(SPLIT_REGEX)).toEqual(['123'])
+    expect(TEST_STRINGS.COMPLEX_MIXED.match(SPLIT_REGEX)).toEqual(['test', '123', 'hello', '=42', 'world'])
+  })
+
+  it('initializes allowlist with normalized words from $DD_ALLOW', () => {
+    const dict = createActionAllowList()
+    // EMOJI and EMOJI_WITH_NUMBERS
+    expect(dict.allowlist.has('123')).toBeTrue()
+    // COMPLEX_MIXED
+    expect(dict.allowlist.has('test')).toBeTrue()
+    expect(dict.allowlist.has('hello')).toBeTrue()
+    expect(dict.allowlist.has('=42')).toBeTrue()
+    expect(dict.allowlist.has('world')).toBeTrue()
+  })
+
+  describe('maskAutoActionName', () => {
+    it('masks words not in allowlist (with dictionary from $DD_ALLOW)', () => {
+      const dict = createActionAllowList()
+      const testString1 = maskAutoActionName('test-💥-$>=123-pii', dict.allowlist)
+      expect(testString1.masked).toBeTrue()
+      expect(testString1.name).toBe('test-💥-$>MASKED-MASKED')
+      const testString2 = maskAutoActionName('test-💥+123*hello world', dict.allowlist)
+      expect(testString2.masked).toBeTrue()
+      expect(testString2.name).toBe('test-💥MASKED*hello world')
+    })
+
+    it('handles empty string', () => {
+      const dict = createActionAllowList()
+      const result = maskAutoActionName('', dict.allowlist)
+      expect(result.masked).toBeFalse()
+      expect(result.name).toBe('')
+    })
+  })
+
+  it('updates dictionary when $DD_ALLOW changes', () => {
+    const dict = createActionAllowList()
+    expect(dict.allowlist.size).toBe(5)
+
+    // Simulate a change in $DD_ALLOW
+    window.$DD_ALLOW?.add('new-Word')
+    window.$DD_ALLOW?.add('another-Word')
+    // Trigger the observer manually
+    updateDictionaryOnDDAllowChange(dict)
+
+    // Verify dictionary is updated with new words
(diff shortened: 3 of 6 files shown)


packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+export function getMatchRegex(): RegExp {
+  try {
+    new RegExp('\\p{Letter}', 'u')
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (_) {
+    // Fallback to support european letters and apostrophes
+    return /(?:(?![×Þß÷þø])[a-zÀ-ÿ’])+|(?:(?!(?:(?![×Þß÷þø])[a-zÀ-ÿ’]))[^\s])+/gi
+  }
+  return /[\p{Letter}]+|[\p{Symbol}\p{Number}]+/gu
+}
```
sethfowler-datadog: You can simplify this a bit; just try to create the regular expression you actually want to use. Also, I probably would not use a regular expression literal for the regular expression with the `u` flag, since it could conceivably cause some ancient parsers to barf. (If they exist, such parsers would probably only be used by browsers too old to work with the browser SDK anyway, but it's trivial to avoid, so may as well.)

Beyond that, please memoize the result for performance; throwing an exception is expensive, so it's better not to do it for every string and action we process.

Suggestion:
```suggestion
let matchRegex: RegExp | undefined;
export function getMatchRegex(): RegExp {
  if (!matchRegex) {
    try {
      matchRegex = new RegExp('\\p{Letter}+|[\\p{Symbol}\\p{Number}]+', 'gu')
    } catch {
      // Fallback to support european letters and apostrophes
      matchRegex = /(?:(?![×Þß÷þø])[a-zÀ-ÿ’])+|(?:(?!(?:(?![×Þß÷þø])[a-zÀ-ÿ’]))[^\s])+/gi
    }
  }
  return matchRegex
}
```

packages/rum-core/src/domain/action/privacy/allowedDictionary.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+} from './allowedDictionary'
+
+const TEST_STRINGS = {
+  EMOJI: '💥',
+  EMOJI_WITH_NUMBERS: '💥123',
+  SPECIAL_CHARS: '$$$',
+  SPECIAL_CHARS_WITH_NUMBERS: '$$$123',
+  HYPHENATED_SPECIAL_CHARS: '$$$-123',
+  COMPLEX_MIXED: 'test-<$>-123 hello>=42@world?',
+}
```
sethfowler-datadog: It'd be great to have some test strings that contain more realistic action names, including sentences with full punctuation in a few languages.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+  initializeAllowlist: () => void
+  lastRawString: SetIterator<string> | undefined
+}
+
+export function createActionAllowList(): AllowedDictionary {
+  const actionNameDictionary: AllowedDictionary = {
+    updatedCounter: 0,
+    allowlist: new Set<string>(),
+    initializeAllowlist: () => {
+      if (!actionNameDictionary.allowlist || actionNameDictionary.allowlist.size === 0) {
```
sethfowler-datadog: How can `actionNameDictionary.allowlist` be falsy here?

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+        processRawAllowList(window.$DD_ALLOW, actionNameDictionary)
+      }
+    },
+    lastRawString: window.$DD_ALLOW?.values(),
```
sethfowler-datadog: Better to call this something like `rawStringIterator`. `lastRawString` sounds like it should be of type `string`.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function processRawAllowList(rawAllowlist: Set<string> | undefined, dictionary: AllowedDictionary) {
+  if (!rawAllowlist) {
+    return
+  }
+  if (!dictionary.lastRawString) {
+    dictionary.lastRawString = rawAllowlist.values()
+  }
+  const size = rawAllowlist.size
+  let nextItem = dictionary.lastRawString.next()
```
sethfowler-datadog: I feel like it's probably not a good idea to call `next()` unconditionally here; if this is called by accident when no strings have been added, calling `next()` will kill the iterator, right? 

The `if (dictionary.updatedCounter !== size - 1)` check in the loop also feels iffy to me.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+  while (dictionary.updatedCounter < size && nextItem.value) {
+    processRawString(nextItem.value, dictionary)
+    if (dictionary.updatedCounter !== size - 1) {
+      nextItem = dictionary.lastRawString.next()
+    }
+    dictionary.updatedCounter++
+  }
```
sethfowler-datadog: Would this alternative work?
```suggestion
  while (dictionary.updatedCounter < size) {
    const nextItem = dictionary.lastRawString.next();
    dictionary.updatedCounter++
    if (nextItem.value) {
      processRawString(nextItem.value, dictionary)
    }
  }
```

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+export type AllowedDictionary = {
+  updatedCounter: number
+  allowlist: Set<string>
+  initializeAllowlist: () => void
+  lastRawString: SetIterator<string> | undefined
+}
+
+export function createActionAllowList(): AllowedDictionary {
+  const actionNameDictionary: AllowedDictionary = {
+    updatedCounter: 0,
```
sethfowler-datadog: This might be clearer if you called it something like `rawStringCounter`; otherwise, it's not clear what kind of updated is being tracked.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function processRawString(str: string, dictionary: AllowedDictionary) {
+  const words: string[] | null = str.match(getMatchRegex())
+  if (words) {
+    for (const word of words) {
+      const normalizeWord = word.toLocaleLowerCase()
+      if (!dictionary.allowlist.has(normalizeWord)) {
+        dictionary.allowlist.add(normalizeWord)
+      }
```
sethfowler-datadog: No need to check if the allowlist already has the word; it's actually cheaper to just call `add()`, since the membership check has to be performed in `add()` anyway!
```suggestion
      dictionary.allowlist.add(normalizeWord)
```

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+    return {
+      name,
+      masked: false,
+    }
+  }
+
+  let masked = false
+  return {
+    name: name.replace(getMatchRegex(), (word: string) => {
+      if (!processedAllowlist.has(word.toLowerCase())) {
```
sethfowler-datadog: You use `toLocaleLowerCase()` when constructing the allowlist, but `toLowerCase()` here; they'll need to both be the same or things won't match up.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+    updatedCounter: 0,
+    allowlist: new Set<string>(),
+    initializeAllowlist: () => {
+      if (!actionNameDictionary.allowlist || actionNameDictionary.allowlist.size === 0) {
+        processRawAllowList(window.$DD_ALLOW, actionNameDictionary)
+      }
+    },
+    lastRawString: window.$DD_ALLOW?.values(),
+  }
+  actionNameDictionary.initializeAllowlist()
```
sethfowler-datadog: Since you call `initializeAllowlist()` immediately, does it need to be a property on the object?

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -31,9 +34,12 @@ export function startActionCollection(
   windowOpenObservable: Observable<void>,
   configuration: RumConfiguration
 ) {
-  lifeCycle.subscribe(LifeCycleEventType.AUTO_ACTION_COMPLETED, (action) =>
-    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action))
-  )
+  const actionNameDictionary = createActionAllowList()
+  addAllowlistObserver(actionNameDictionary)
```
sethfowler-datadog: Could `createActionAllowList()` register the observer? It feels like this code shouldn't need to know about that. (Maybe `createActionAllowList()` is really `startActionAllowListObserver()` or something?)

Going a bit further: right now, there is no code that removes the allowlist observer when `startActionCollection()`'s `stop()` is called. So we create a new dictionary and add another observer whenever `startActionCollection()` is called, but nothing ever removes them, and they keep building up. I don't think we want that.

We should pick one of two approaches:
* Use a single global allowlist observer and dictionary, and never replace them or clear them. (Reasonable, since `$DD_ALLOW` is also global.)
* Register the allowlist observer and set up the dictionary when action collection starts; unregister the allowlist observer and clear the underlying dictionary when action collection stops.

The first avoids reprocessing the raw strings when recording restarts, so it has a performance benefit, but naturally you'll have to add some additional affordances for testing with that approach.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
cy-moi: My bad, the thought here was that we may have to initialize again somewhere after. But this should not be necessary anymore.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (!window.$DD_ALLOW) {
+    return {
+      name,
+      masked: false,
+    }
+  }
+
+  let masked = false
+  return {
+    name: name.replace(getMatchRegex(), (word: string) => {
```
amortemousque: **❓ question:** IIUC, when multiple tokens are masked, we might end up with a name like "MASKED MASKED MASKED", which could be confusing from a UI perspective.  Especially, given that the other masking strategy displays "Masked Element", shouldn’t we aim for consistency? Do we have product input on this?

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -14,8 +14,9 @@ export const enum ActionNameSource {
   TEXT_CONTENT = 'text_content',
   STANDARD_ATTRIBUTE = 'standard_attribute',
   BLANK = 'blank',
+  MASK_DISALLOWED = 'mask_disallowed',
```
amortemousque: **💬 suggestion:** ‏I find this source name confusing. As mentioned in my other [comment](https://github.com/DataDog/browser-sdk/pull/3648/files#r2163930651), I think we should ensure consistent behavior across our masking strategies. That way, we can keep using MASK_PLACEHOLDER as the source. If we need to identify which masking strategy was applied, we could include it as a separate event attribute. Wdyt?

packages/rum-core/src/domain/action/actionCollection.ts
cy-moi: I see. 
~With the first approach, do we need end-to-end tests or just keeping the contexts in unit tests would be enough?~ ~I'm ok to proceed with the second approach for now but clearing the dictionary and re-process could get expensive. We might need some field data on this.~
Fixed with the 2nd approach.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: Ideally, yes we would like to distinguish the two strategies. We did add `mask_disallowed` as a name source in [`rum-event-format`](https://github.com/DataDog/rum-events-format/commit/8beec7c74ef59e5363da86cc30877e65094055d2) as a separate attribute. But I'm open to change it to another name.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
cy-moi: Indeed, although `Masked Element` is not idea for tokenized the names either, maybe we should seek another way, ie XXX XXX in session replay. Will ping product on this.

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -72,7 +80,11 @@ export function startActionCollection(
       lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action))
     },
     actionContexts,
-    stop,
+    stop: () => {
+      clearActionNameDictionary()
```
sethfowler-datadog: Nit: since `actionNameDictionary` is a constant, this is the same as calling `actionNameDictionary.clear()` directly, so you could get rid of `clearActionNameDictionary`.

packages/rum-core/src/domain/action/privacy/allowedDictionary.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+        'test',
+        'paragraph',
+        'with',
+        'various',
+        'symbols',
+        'and',
+        'more',
+      ]
+      expectedParagraphMixed.forEach((expected) => {
+        expect(paragraphMixedTokens).toContain(expected)
```
sethfowler-datadog: Why not assert against the exact value returned by `tokenize()`, instead of checking that certain words are present? That way we will know if something unexpected has changed, or if something is present that shouldn't be.

packages/rum-core/src/domain/action/privacy/allowedDictionary.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+      })
+      const frenchTokens = tokenize(LANGUAGES_TEST_STRINGS.FRENCH_MIXED_SENTENCE)
+      const expectedFrench = [
+        'C',
+        'est',
```
sethfowler-datadog: These two should be one word, ideally. Seems like a bug in the regular expression.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,160 @@
...
+export function createActionAllowList(): AllowedDictionary {
+  const dictionary: AllowedDictionary = {
+    rawStringCounter: 0,
+    allowlist: new Set<string>(),
+    rawStringIterator: window.$DD_ALLOW?.values(),
+    clear: () => {
+      dictionary.allowlist.clear()
+      dictionary.rawStringCounter = 0
+      dictionary.rawStringIterator = undefined
+      supportsUnicodeRegex = undefined
```
sethfowler-datadog: Nit: I don't think it's necessary to reset `supportsUnicodeRegex`; that won't change just because action collection starts or stops.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,160 @@
...
+      window.$DD_ALLOW_OBSERVERS?.delete(observer)
+    },
+  }
+
+  const observer = () => processRawAllowList(window.$DD_ALLOW, dictionary)
+
+  // Initialize allowlist if needed
+  if (dictionary.allowlist.size === 0) {
```
sethfowler-datadog: Nit: I think you can get rid of the "if needed" comment and the check; you are creating a new `AllowedDictionary` every time this function is called, right?

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,160 @@
...
+  allowlist: Set<string>
+  rawStringIterator: IterableIterator<string> | undefined
+  clear: () => void
+}
+
+export function createActionAllowList(): AllowedDictionary {
+  const dictionary: AllowedDictionary = {
+    rawStringCounter: 0,
+    allowlist: new Set<string>(),
+    rawStringIterator: window.$DD_ALLOW?.values(),
```
sethfowler-datadog: Nit: you may as well just let `processRawAllowList()` handle initializing this, and set it to `undefined` here. That way, it always gets initialized using the same mechanism.

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+export function maskActionName(
+  actionName: ClickActionBase,
+  nodeSelfPrivacy: NodePrivacyLevel,
+  processedAllowlist: Set<string>
+): ClickActionBase {
+  if (!window.$DD_ALLOW) {
+    return actionName
+  }
```
sethfowler-datadog: I don't love this behavior, since it seems like there are edge cases where it could make it nondeterministic whether actions get masked or not. (For example, consider a customer who loads the browser SDK via CDN, and who combines server-generated HTML with JS that's built with our build plugin. Clicks on the server-generated HTML would produce unmask actions if they happen early, before the customer's JS bundle loads.)

Is there any way we could require the customer to opt in via their configuration, and if they're opted in and `$DD_ALLOW` is not present, we fail closed, just like we do if `initializeUnicodeSupport()` fails? Or did we already consider and reject this idea?

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  if (!initializeUnicodeSupport()) {
+    return {
+      ...actionName,
+      name: name ? ACTION_NAME_PLACEHOLDER : '',
+      nameSource: name ? ActionNameSource.MASK_DISALLOWED : nameSource,
+    }
+  }
+
+  const { splitRegex } = cachedRegexes!
```
sethfowler-datadog: Nit: perhaps, to avoid the need for a non-null assertion here, you could replace `initializeUnicodeSupport()` with a function that returns the regular expressions if Unicode regular expressions are supported, and returns `undefined` otherwise? So you'd have e.g.:
```ts
const regexes = getOrInitRegexes();
if (!regexes) {
  // No Unicode regular expression support; fall back.
  return
}

// Unicode regular expressions were supported; use them!
const maskedName = name.replace(regexes.split, // ...
```

packages/rum-core/src/domain/action/privacy/allowedDictionary.ts
cy-moi: Per discussion with team, we should fail close with a new `defaultPrivacyLevel`.

cy-moi:
Per discussion internally, we changed to the exact match with DD_ALLOW approach.

PR #3652, opened 2025-06-20
✨ don't write lock to cookie for read only operation

### LINKED ISSUE #3385: 🐛 _dd_s cookie being written multiple times per second
This relates to #2907 . We noticed that _dd_s cookie is being written multiple times per second. Considering the main purpose of this cookie is tag the session id, it should change when the session id changes.

We have found that out because for some reason chrome is having trouble with _dd_s cookie when the front end application is running inside an iframe added to another website. The following message shows up around 8 times per second:

> Chrome is moving towards a new experience that allows users to choose to browse without third-party cookies.

Image

Notice that *only show cookie with an issue* is checked:

Image

thomas-lebeau:
## Motivation

Don't write lock to cookie for read only operations.
fixes #3385

## Changes

Prevent the `watchSession()` method of the session store to write the lock in the cookie if it does not need to be written to.

refactor `sessionStore` unit test to rely on `createFakeSessionStore` instead of the real cookie session store

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/core/src/domain/session/sessionStore.spec.ts
```diff
@@ -185,14 +192,22 @@ describe('session store', () => {
         fail('Unable to initialize cookie storage')
         return
       }
+
+      const sessionStoreStrategy = getSessionStoreStrategy(sessionStoreStrategyType, DEFAULT_CONFIGURATION)
+      persistSessionSpy = spyOn(sessionStoreStrategy, 'persistSession').and.callThrough()
+      isLockEnabled = sessionStoreStrategy.isLockEnabled
```
BenoitZugmeyer: **suggestion:** maybe use [`createFakeSessionStoreStrategy`](https://github.com/DataDog/browser-sdk/blob/a70d7487a9c98c1f2046e33717ff3152902170be/packages/core/src/domain/session/sessionStoreOperations.spec.ts#L237)

packages/core/src/domain/session/sessionStore.spec.ts
```diff
@@ -252,7 +267,7 @@ describe('session store', () => {
         'when session not in cache, session not in store and new session not tracked, ' +
           'should store not tracked session and trigger renew session',
         () => {
-          setupSessionStore(() => FakeTrackingType.NOT_TRACKED)
+          setupSessionStore(undefined, () => FakeTrackingType.NOT_TRACKED)
```
bcaudan: **🥜 nitpick:** ‏using `undefined` to set `{}` seems confusing to me, what about extracting an `const EMPTY_SESSION_STATE = {}` to use there

packages/core/src/domain/session/sessionStore.spec.ts
```diff
@@ -456,70 +461,74 @@ describe('session store', () => {
     })
 
     describe('regular watch', () => {
-      it('when session not in cache and session not in store, should do nothing', () => {
+      it('when session not in cache and session not in store, should expire session in store', () => {
```
bcaudan: **💬 suggestion:** FMU, expire session means `expect(expireSpy).toHaveBeenCalled()`‏
```suggestion
      it('when session not in cache and session not in store, should store the expired session', () => {
```

packages/core/src/domain/session/sessionStore.spec.ts
```diff
@@ -456,70 +461,74 @@ describe('session store', () => {
...
         expect(sessionStoreManager.getSession().expire).toBe(getStoreExpiration())
         expect(expireSpy).not.toHaveBeenCalled()
+        expect(sessionStoreStrategy.persistSession).not.toHaveBeenCalled()
       })
 
-      it('when session id in cache is different than session id in store, should expire session', () => {
-        setSessionInStore(FakeTrackingType.TRACKED, FIRST_ID)
-        setupSessionStore()
-        setSessionInStore(FakeTrackingType.TRACKED, SECOND_ID)
+      it('when session id in cache is different than session id in store, should expire session in cache', () => {
```
bcaudan: **💬 suggestion:** same reasoning than previous comment
```suggestion
      it('when session id in cache is different than session id in store, should expire session and not touch the store', () => {
```
‏

packages/core/src/domain/session/sessionStore.spec.ts
```diff
@@ -456,70 +461,74 @@ describe('session store', () => {
...
-        setupSessionStore()
-        setSessionInStore(FakeTrackingType.TRACKED, FIRST_ID)
+        setupSessionStore(createSessionState(FakeTrackingType.NOT_TRACKED, FIRST_ID))
+        setSessionInStore(createSessionState(FakeTrackingType.TRACKED, FIRST_ID))
 
         clock.tick(STORAGE_POLL_DELAY)
 
         expect(sessionStoreManager.getSession().id).toBeUndefined()
         expect(expireSpy).toHaveBeenCalled()
+        expect(sessionStoreStrategy.persistSession).not.toHaveBeenCalled()
```
bcaudan, line 533: **💬 suggestion:** ‏what about using a similar test name than the previous one?
from previous suggestion, something like `should expire session and not touch the store`

bcaudan (approved):

PR #3653, opened 2025-06-23
♻️ [RUM-10453] Use hooks for Logs SDK to decouple url context 

amortemousque:
## Motivation

Use hooks for Logs SDK to decouple url context 

## Changes

- create urlContext.ts using hooks
- remove url context related logic from assembly
- update tests

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/logs/src/boot/preStartLogs.ts
```diff
@@ -43,6 +43,13 @@ export function createPreStartStrategy(
   const userContext = buildUserContextManager()
   bufferContextCalls(userContext, CustomerContextKey.userContext, bufferApiCalls)
 
+  // TODO next major: remove cacheUrlContext, less precision in pre-init URL context is acceptable, and RUM is already less precise
+  // `location` is undefined during server-side rendering (e.g., in Node)
+  const location = getGlobalObject().location
+  if (location) {
+    cacheUrlContext(location)
```
amortemousque: Caches the url context for pre-start calls to prevent breaking changes

packages/logs/src/domain/contexts/rumInternalContext.spec.ts
```diff
@@ -47,6 +47,18 @@ describe('startRUMInternalContext', () => {
     expect(defaultLogsEventAttributes).toEqual({ foo: 'bar' })
   })
 
+  it('should include rum internal context  related to the start time', () => {
```
amortemousque, line 50: Unrelated to this PR, but moved this test from assembly to rumInternalContext

packages/logs/src/boot/preStartLogs.ts
bcaudan: > less precision in pre-init URL context is acceptable

so it is a breaking change, no? 
FMU, with this approach, we may associate logs with wrong URLs whereas it is not the case for now.

Instead of caching the init value, would it make sense to track a url context history?

packages/logs/src/domain/contexts/urlContexts.ts
```diff
@@ -0,0 +1,34 @@
+import { SESSION_TIME_OUT_DELAY, relativeNow, HookNames } from '@datadog/browser-core'
+import type { Hooks } from '../hooks'
+
+export const URL_CONTEXT_TIME_OUT_DELAY = SESSION_TIME_OUT_DELAY
+
```
bcaudan, line 14: **💬 suggestion:** ‏unused
```suggestion
```

packages/logs/src/boot/preStartLogs.ts
amortemousque: Good point! Done: https://github.com/DataDog/browser-sdk/pull/3653/commits/5d285025568463ba2b32275ba9ab5f0de74af0fb

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -43,6 +43,13 @@ export function createPreStartStrategy(
   const userContext = buildUserContextManager()
   bufferContextCalls(userContext, CustomerContextKey.userContext, bufferApiCalls)
 
+  // TODO next major: remove startUrlContextHistory from preStartStrategy, less precision in pre-init URL context is acceptable, and RUM is already less precise
```
bcaudan, line 46: **❓ question:** ‏why not keeping it? I am wondering if it is not something that we should have on RUM as well rather than dropping it on logs.

packages/logs/src/domain/contexts/urlContexts.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export function stopUrlContextHistory() {
+  urlContextHistory?.stop()
+  locationChangeSubscription?.unsubscribe()
+}
+
+export function startUrlContexts(hooks: Hooks) {
+  hooks.register(HookNames.Assemble, ({ startTime }) => {
+    const urlContext = urlContextHistory.find(startTime)
```
bcaudan, line 50: **💬 suggestion:** ‏as per convention, we should pass `urlContextHistory` dependency as a parameter.

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -1,12 +1,12 @@
 import type { Duration, ServerDuration, Observable } from '@datadog/browser-core'
 import { DISCARDED, HookNames, isEmptyObject, mapValues, toServerDuration } from '@datadog/browser-core'
+import type { LocationChange } from '@datadog/browser-core/src/browser/locationChangeObservable'
```
bcaudan, line 3: **💬 suggestion:** ‏that seems odd

packages/logs/src/domain/contexts/urlContexts.ts
```diff
@@ -0,0 +1,60 @@
...
+
+let urlContextHistory: ValueHistory<UrlContext>
+let locationChangeSubscription: Subscription | undefined
+
+export function startUrlContextHistory(
+  location: Location,
+  locationChangeObservable: Observable<LocationChange> = createLocationChangeObservable(
+    // startUrlContextHistory is called from preStartLogs.ts before the full configuration is available,
+    // so a minimal placeholder configuration is used here.
+    { allowUntrustedEvents: false } as LogsConfiguration,
```
bcaudan, line 29: **❓ question:** ‏any reasoning on allowUntrustedEvents: false VS true?

packages/logs/src/boot/preStartLogs.spec.ts
```diff
@@ -150,20 +147,6 @@ describe('preStartLogs', () => {
       expect(getLoggedMessage(0).savedDate).toEqual((Date.now() - ONE_SECOND) as TimeStamp)
     })
 
-    it('saves the URL', () => {
```
bcaudan, line 153: **💬 suggestion:** ‏what about keeping a test for this case?

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -43,6 +43,13 @@ export function createPreStartStrategy(
...
+  // `location` is undefined during server-side rendering (e.g., in Node)
+  const location = getGlobalObject().location
+  if (location) {
+    startUrlContextHistory(location)
+  }
```
bcaudan, line 51: **❓ question:** ‏ ‏IIRC, for sever-side rendering, we just need the evaluation of the SDK to work, so we don't need to handle the case where we don't have location, is that it?

amortemousque: Yes exactly. I'll update the comment to make it more explicit.

packages/logs/src/boot/preStartLogs.ts
amortemousque: If go that route, I think it should be in the scope of early data collection. I don't have a strong opinion yet, besides being consistent across SDKs. @BenoitZugmeyer wdyt?

packages/logs/src/domain/contexts/urlContexts.ts
amortemousque: Do you mean passing it down from preStart -> start -> startUrlContexts?
I originally though about doing it but I wanted the changes to be minimal assuming we would drop it in next major.
In your opinion I should do it regardless of dropping it?

packages/logs/src/domain/contexts/urlContexts.ts
amortemousque: Not really, what's on your mind?

bcaudan: I am pondering:
- with which one should we are less likely to have customers requesting the other behavior, I don't have an answer 😅
- should we adjust the locationObservable to the configured value after start?

packages/logs/src/domain/contexts/urlContexts.ts
bcaudan: I am kind of a stickler on this one 😕
Regardless of what our current plans are, we will live with it for some time and the plans may change. I'd be in favor of doing it the expected way from the start, it can expose a flaw in our current design if we are not happy with it or help us identify a pattern that we may have with early data collection.

packages/logs/src/domain/contexts/urlContexts.ts
amortemousque: Erf, now that I’m using the location observable, it also means we’re instrumenting the `pushState` API at evaluation time. Not sure if you’d call that a breaking change, but I’m pretty sure @BenoitZugmeyer would 😅
What's your opinion?

bcaudan: Indeed, that is not great.
It seems that we do need to attach some context info at pre-start time, which was kind of the role of the common context that is removed by this PR.
I am wondering if we could want to do the same thing with other contexts 🤔
If yes, we could think of a way to adjust the hook design to support that.
If not, maybe we should keep this specific thing for logs for now and revisit it later. 
wdyt?

PR #3657, opened 2025-06-24
👷 [RUM-9218] Adjust E2E setup to use cross-origin URLs for loading CDN bundles

rgaignault:
## Motivation

Currently, our E2E test suite is loading bundles using a relative URL (ex: /datadog-rum.js). Because of this, we missed a faulty configuration change that broke Session Replay

## Changes

Let’s use a cross-origin URL to load our bundles in E2E tests. We could just use the [“crossOrigin” server](https://github.com/DataDog/browser-sdk/blob/42739576f614e5b25582706fd83833389c4a0231/test/e2e/lib/framework/httpServers.ts#L27) that we already use for some tests.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/lib/framework/pageSetups.ts
+++ b/test/e2e/lib/framework/pageSetups.ts
@@ -53,11 +53,13 @@ n=o.getElementsByTagName(u)[0];n.parentNode.insertBefore(d,n)
 })(window,document,'script','${url}','${globalName}')`
   }
 
+  const logsScriptUrl = `${servers.crossOrigin.url}/datadog-logs.js`
+  const rumScriptUrl = `${servers.crossOrigin.url}/${options.useRumSlim ? 'datadog-rum-slim.js' : 'datadog-rum.js'}`
+
   if (options.logs) {
-    const logsUrl = `${servers.crossOrigin.url}/datadog-logs.js`
     body += html`
       <script>
-        ${formatSnippet(logsUrl, 'DD_LOGS')}
+        ${formatSnippet(logsScriptUrl, 'DD_LOGS')}
         DD_LOGS.onReady(function () {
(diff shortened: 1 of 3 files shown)


test/e2e/lib/framework/pageSetups.ts
```diff
@@ -92,7 +94,7 @@ export function bundleSetup(options: SetupOptions, servers: Servers) {
 
   if (options.logs) {
     header += html`
-      <script type="text/javascript" src="./datadog-logs.js"></script>
+      <script type="text/javascript" src="${servers.crossOrigin.url}/datadog-logs.js""></script>
```
mormubis: **❓ question:** ‏ Is there a typo here?


**💬 suggestion:** ‏ Should we add a URL variable just for consistency between tests?

mormubis:
```suggestion
      <script type="text/javascript" src="${servers.crossOrigin.url}/datadog-logs.js"></script>
```

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -90,9 +93,12 @@ export function bundleSetup(options: SetupOptions, servers: Servers) {
     header += setupEventBridge(servers)
   }
 
+  const logsScriptUrl = `${servers.crossOrigin.url}/datadog-logs.js`
+  const rumScriptUrl = `${servers.crossOrigin.url}/${options.useRumSlim ? 'datadog-rum-slim.js' : 'datadog-rum.js'}`
```
bcaudan: **💬 suggestion:** ‏what about extracting a helper to avoid duplicating the logic that builds the urls?

mormubis (approved):

PR #3659, opened 2025-06-25
✨ [RUM-10044] Support for overriding the source

mormubis:
## Motivation

This PR adds support for plugins to override the `source`, `variant` and `sdk_version` with their own information.

## Changes

Add `source`, `variant` and `sdk_version` to the initial configuration that allows third party to provide extra information about their plugins.

## Test instructions

- Go to staging
- Override the configuration and add:
```
"source": "flutter",
"variant": "test-variant",
"sdk_version": "my-version"
```
- _Check that events are sending the new information_

- _Check that events in https://dd.datad0g.com/rum/sessions?query=%40type%3Asession%20%40usr.id%3A33803929&agg_m=count&agg_m_source=base&agg_t=count&from_ts=1752574307484&to_ts=1753179107484&live=true shows the new information_

## Checklist

- [X] Tested locally
- [X] Tested on staging

### DIFF
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -384,7 +384,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     /**
      * The source of the SDK, used for support plugins purposes.
      */
-    source: initConfiguration?.source || 'browser',
+    source: initConfiguration?.source === 'flutter' ? 'flutter' : 'browser',
     variant: initConfiguration?.variant,
     sdkVersion: initConfiguration?.sdkVersion,
 
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -284,9 +284,9 @@ export interface Configuration extends TransportConfiguration {
   messageBytesLimit: number
 
   // internal
-  sdkVersion?: string | undefined
+  sdkVersion: string | undefined
   source: 'browser' | 'flutter'
-  variant?: string | undefined
+  variant: string | undefined
 }
 
 function isString(tag: unknown, tagName: string): tag is string | undefined | null {
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -390,7 +390,6 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     /**
      * The source of the SDK, used for support plugins purposes.
      */
-    source: initConfiguration?.source === 'flutter' ? 'flutter' : 'browser',
     variant: initConfiguration?.variant,
     sdkVersion: initConfiguration?.sdkVersion,
 
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -220,7 +220,7 @@ export interface InitConfiguration {
   /**
    * [Internal option] Additional configuration for the SDK.
    */
-  source?: string | undefined
+  source?: 'browser' | 'flutter' | undefined
 
   /**
    * [Internal option] Additional configuration for the SDK.
@@ -279,7 +279,7 @@ export interface Configuration extends TransportConfiguration {
 
   // internal
   sdkVersion?: string | undefined
-  source: string
+  source: 'browser' | 'flutter'
   variant?: string | undefined
 }
 
@@ -411,7 +411,7 @@ export function serializeConfiguration(initConfiguration: InitConfiguration) {
     allow_untrusted_events: !!initConfiguration.allowUntrustedEvents,
     tracking_consent: initConfiguration.trackingConsent,
     use_allowed_tracking_origins: Array.isArray(initConfiguration.allowedTrackingOrigins),
-    source: initConfiguration.source || 'browser',
+    source: initConfiguration.source,
     sdk_version: initConfiguration.sdkVersion,
     variant: initConfiguration.variant,
   } satisfies RawTelemetryConfiguration
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -216,6 +216,11 @@ export interface InitConfiguration {
    * @defaultValue 5
    */
   telemetryUsageSampleRate?: number
+
+  /**
+   * [Internal option] Additional configuration for the SDK.
+   */
+  additionalConfig?: Record<string, unknown>
 }
 
 // This type is only used to build the core configuration. Logs and RUM SDKs are using a proper type
@@ -261,6 +266,11 @@ export interface Configuration extends TransportConfiguration {
   flushTimeout: Duration
   batchMessagesLimit: number
   messageBytesLimit: number
+
+  // internal
+  source?: string | undefined
+  variant?: string | undefined
+  sdk_version?: string | undefined
 }
 
 function isString(tag: unknown, tagName: string): tag is string | undefined | null {
@@ -360,6 +370,14 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
      */
     batchMessagesLimit: 50,
     messageBytesLimit: 256 * ONE_KIBI_BYTE,
+
+    /**
+     * The source of the SDK, used for support plugins purposes.
+     */
+    source: (initConfiguration.additionalConfig?.source as string) || 'browser',
+    variant: initConfiguration.additionalConfig?.variant as string,
+    sdk_version: initConfiguration.additionalConfig?.sdk_version as string,
+
     ...computeTransportConfiguration(initConfiguration),
   }
 }
@@ -383,5 +401,6 @@ export function serializeConfiguration(initConfiguration: InitConfiguration) {
     allow_untrusted_events: !!initConfiguration.allowUntrustedEvents,
     tracking_consent: initConfiguration.trackingConsent,
     use_allowed_tracking_origins: Array.isArray(initConfiguration.allowedTrackingOrigins),
+    additional_config: initConfiguration.additionalConfig,
   } satisfies RawTelemetryConfiguration
 }
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -390,8 +390,8 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     /**
      * The source of the SDK, used for support plugins purposes.
      */
-    variant: initConfiguration?.variant,
-    sdkVersion: initConfiguration?.sdkVersion,
+    variant: initConfiguration.variant,
+    sdkVersion: initConfiguration.sdkVersion,
 
     ...computeTransportConfiguration(initConfiguration),
   }
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -220,7 +220,17 @@ export interface InitConfiguration {
   /**
    * [Internal option] Additional configuration for the SDK.
    */
-  additionalConfig?: Record<string, unknown>
+  source?: string | undefined
+
+  /**
+   * [Internal option] Additional configuration for the SDK.
+   */
+  sdkVersion?: string | undefined
+
+  /**
+   * [Internal option] Additional configuration for the SDK.
+   */
+  variant?: string | undefined
 }
 
 // This type is only used to build the core configuration. Logs and RUM SDKs are using a proper type
@@ -268,9 +278,9 @@ export interface Configuration extends TransportConfiguration {
   messageBytesLimit: number
 
   // internal
-  source?: string | undefined
+  sdkVersion?: string | undefined
+  source: string
   variant?: string | undefined
-  sdk_version?: string | undefined
 }
 
 function isString(tag: unknown, tagName: string): tag is string | undefined | null {
@@ -374,9 +384,9 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     /**
      * The source of the SDK, used for support plugins purposes.
      */
-    source: (initConfiguration.additionalConfig?.source as string) || 'browser',
-    variant: initConfiguration.additionalConfig?.variant as string,
-    sdk_version: initConfiguration.additionalConfig?.sdk_version as string,
+    source: initConfiguration?.source || 'browser',
+    variant: initConfiguration?.variant,
+    sdkVersion: initConfiguration?.sdkVersion,
 
     ...computeTransportConfiguration(initConfiguration),
   }
@@ -401,5 +411,8 @@ export function serializeConfiguration(initConfiguration: InitConfiguration) {
     allow_untrusted_events: !!initConfiguration.allowUntrustedEvents,
     tracking_consent: initConfiguration.trackingConsent,
     use_allowed_tracking_origins: Array.isArray(initConfiguration.allowedTrackingOrigins),
+    source: initConfiguration.source || 'browser',
+    sdk_version: initConfiguration.sdkVersion,
+    variant: initConfiguration.variant,
   } satisfies RawTelemetryConfiguration
 }
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -219,16 +219,22 @@ export interface InitConfiguration {
 
   /**
    * [Internal option] Additional configuration for the SDK.
+   *
+   * @internal
    */
   source?: 'browser' | 'flutter' | undefined
 
   /**
    * [Internal option] Additional configuration for the SDK.
+   *
+   * @internal
    */
   sdkVersion?: string | undefined
 
   /**
    * [Internal option] Additional configuration for the SDK.
+   *
+   * @internal
    */
   variant?: string | undefined
 }
--- a/packages/core/src/domain/configuration/configuration.ts
+++ b/packages/core/src/domain/configuration/configuration.ts
@@ -401,6 +401,5 @@ export function serializeConfiguration(initConfiguration: InitConfiguration) {
     allow_untrusted_events: !!initConfiguration.allowUntrustedEvents,
     tracking_consent: initConfiguration.trackingConsent,
     use_allowed_tracking_origins: Array.isArray(initConfiguration.allowedTrackingOrigins),
-    additional_config: initConfiguration.additionalConfig,
   } satisfies RawTelemetryConfiguration
 }
--- a/packages/core/src/domain/configuration/endpointBuilder.spec.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.spec.ts
@@ -154,7 +154,7 @@ describe('endpointBuilder', () => {
 
   describe('source and variant configuration', () => {
     it('should use the default source when no configuration is provided', () => {
-      const config = initConfiguration
+      const config = { ...initConfiguration, source: 'browser' as const }
 
       const endpoint = createEndpointBuilder(config, 'rum').build('fetch', DEFAULT_PAYLOAD)
       expect(endpoint).toContain('ddsource=browser')
@@ -163,9 +163,9 @@ describe('endpointBuilder', () => {
     })
 
     it('should use source and variant when provided', () => {
-      const config = { ...initConfiguration, source: 'my-source', variant: 'my-variant' }
+      const config = { ...initConfiguration, source: 'flutter' as const, variant: 'my-variant' }
       const endpoint = createEndpointBuilder(config, 'rum').build('fetch', DEFAULT_PAYLOAD)
-      expect(endpoint).toContain('ddsource=my-source')
+      expect(endpoint).toContain('ddsource=flutter')
       expect(endpoint).toContain('_dd.variant=my-variant')
     })
 
--- a/packages/core/src/domain/configuration/endpointBuilder.spec.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.spec.ts
@@ -162,17 +162,12 @@ describe('endpointBuilder', () => {
       expect(endpoint).not.toContain('_dd.sdk_version=')
     })
 
-    it('should use source and variant when provided', () => {
-      const config = { ...initConfiguration, source: 'flutter' as const, variant: 'my-variant' }
+    it('should use sdk_version, source and variant when provided', () => {
+      const config = { ...initConfiguration, sdkVersion: '1.2.3', source: 'flutter' as const, variant: 'my-variant' }
       const endpoint = createEndpointBuilder(config, 'rum').build('fetch', DEFAULT_PAYLOAD)
       expect(endpoint).toContain('ddsource=flutter')
-      expect(endpoint).toContain('_dd.variant=my-variant')
-    })
-
-    it('should include sdk_version when provided in additionalConfig', () => {
-      const config = { ...initConfiguration, sdkVersion: '1.2.3' }
-      const endpoint = createEndpointBuilder(config, 'rum').build('fetch', DEFAULT_PAYLOAD)
       expect(endpoint).toContain('_dd.sdk_version=1.2.3')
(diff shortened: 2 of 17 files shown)


packages/core/src/domain/configuration/configuration.ts
```diff
@@ -220,6 +230,10 @@ export interface Configuration extends TransportConfiguration {
   flushTimeout: Duration
   batchMessagesLimit: number
   messageBytesLimit: number
+
+  // internal
+  source?: string | undefined
```
thomas-lebeau: **❓ question:** ‏ can we type this from rum-event-format?

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -269,6 +283,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     !isString(initConfiguration.version, 'Version') ||
     !isString(initConfiguration.env, 'Env') ||
     !isString(initConfiguration.service, 'Service') ||
+    !isString(initConfiguration.source, 'Source') ||
```
thomas-lebeau: **💬 suggestion:** ‏ do the same with variant?

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -81,18 +81,20 @@ export function buildEndpointHost(
...
+
   const parameters = [
-    'ddsource=browser',
+    `ddsource=${source}`,
     `dd-api-key=${clientToken}`,
     `dd-evp-origin-version=${encodeURIComponent(__BUILD_ENV__SDK_VERSION__)}`,
     'dd-evp-origin=browser',
     `dd-request-id=${generateUUID()}`,
-  ]
+  ].concat(variant ? [`_dd.variant=${variant}`] : [])
```
thomas-lebeau: **💬 suggestion:** ‏ let's not put variant here, especially if it's not supported

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -81,18 +81,20 @@ export function buildEndpointHost(
...
+  { clientToken, internalAnalyticsSubdomain, source: configSource, variant }: InitConfiguration,
   trackType: TrackType,
   api: ApiType,
   { retry, encoding }: Payload
 ) {
+  const source = configSource || 'browser'
+
   const parameters = [
-    'ddsource=browser',
+    `ddsource=${source}`,
```
thomas-lebeau, line 95: **❓ question:** ‏ do we need to have source both in the url query and in the event itself?

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -10,9 +10,18 @@ export function startDefaultContext(
   configuration: RumConfiguration,
   sdkName: 'rum' | 'rum-slim' | 'rum-synthetics' | undefined
 ) {
-  hooks.register(
-    HookNames.Assemble,
-    ({ eventType }): DefaultRumEventAttributes => ({
+  hooks.register(HookNames.Assemble, ({ eventType }): DefaultRumEventAttributes => {
+    const source = (configuration.source || 'browser') as
```
thomas-lebeau: **💬 suggestion:** what about‏ setting the default in validateAndBuildConfiguration?

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -23,12 +32,13 @@ export function startDefaultContext(
         },
         browser_sdk_version: canUseEventBridge() ? __BUILD_ENV__SDK_VERSION__ : undefined,
```
thomas-lebeau: **💬 suggestion:** ‏ I think here we should also send `__BUILD_ENV__SDK_VERSION__` when `configuration.source` is not 'browser', no?

packages/core/src/domain/configuration/endpointBuilder.ts
mormubis: If I remove this parameter from the URL, logs are not sent.

packages/core/src/domain/configuration/endpointBuilder.ts
mormubis: I thought `_dd` could send anything.

https://github.com/DataDog/browser-sdk/blob/f417957fe4056427b89edc0510b9e7b2fc22083e/packages/core/src/domain/configuration/endpointBuilder.ts#L106

packages/core/src/domain/configuration/endpointBuilder.ts
thomas-lebeau: really? that doesn't sound right. could it be related to https://github.com/DataDog/browser-sdk/blob/65f51eccb703d530ac3006cc6812e7b89698bff9/packages/core/src/domain/intakeSites.ts#L16C7-L16C7

packages/core/src/domain/configuration/endpointBuilder.ts
thomas-lebeau: hmm, yeah right it would work. I think we should really define when to send things in the data url and when in the event and when in both. it's really confusing all of these

packages/rum-core/src/domain/contexts/defaultContext.ts
mormubis: Added!

packages/core/src/domain/configuration/configuration.ts
mormubis: `rumEvent.types` is stored in `rum-core`. I don't think we want to create a cycle between the packages. What are your thoughts?

packages/core/src/domain/configuration/configuration.ts
mormubis: added

packages/core/src/domain/configuration/endpointBuilder.ts
mormubis: I agree. My suggestion would be to move away from the URL parameters, but I'm not sure if there is any reason why the information is there in the first place.

packages/rum-core/src/domain/contexts/defaultContext.ts
mormubis: Done!

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -98,6 +102,14 @@ function buildEndpointParameters(
     parameters.push(`dd-evp-encoding=${encoding}`)
   }
 
+  if (sdkVersion) {
+    parameters.push(`_dd.sdk_version=${sdkVersion}`)
+  }
+
+  if (variant) {
+    parameters.push(`_dd.variant=${variant}`)
+  }
```
thomas-lebeau: **💬 suggestion:** ‏ This method works of setting `_dd.xxx` properties on event only for rum tracks and is ignored in other tracks. You're already setting `variant` in `defaultContext.ts`, so I think it would be better to keep it there together with `browser_sdk_version` and `sdk_version`

packages/core/src/domain/configuration/tags.ts
```diff
@@ -7,8 +7,8 @@ export const TAG_SIZE_LIMIT = 200
 declare const __BUILD_ENV__SDK_VERSION__: string
 
 export function buildTags(configuration: Configuration): string[] {
-  const { env, service, version, datacenter } = configuration
-  const tags = [buildTag('sdk_version', __BUILD_ENV__SDK_VERSION__)]
+  const { env, service, version, datacenter, sdk_version: sdkVersion, variant } = configuration
+  const tags = [buildTag('sdk_version', sdkVersion ?? __BUILD_ENV__SDK_VERSION__)]
```
thomas-lebeau, line 11: **💬 suggestion:** ‏ I think here we should also set `browser_sdk_version` if `sdkVersion` was passed in the config, otherwise we would not have this information for logs.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -178,6 +178,11 @@ export interface InitConfiguration {
    * @default 5
    */
   telemetryUsageSampleRate?: number
+
+  /**
+   * [Internal option] Additional configuration for the SDK.
+   */
+  additionalConfig?: Record<string, unknown>
```
thomas-lebeau: **💬 suggestion:** ‏ 

```suggestion
  additionalConfig?: {
    source: string
    variant?: string
    sdk_version?: string
  }
```

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -319,6 +329,14 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
      */
     batchMessagesLimit: 50,
     messageBytesLimit: 256 * ONE_KIBI_BYTE,
+
+    /**
+     *
+     */
+    source: (initConfiguration.additionalConfig?.source as string) || 'browser',
+    variant: initConfiguration.additionalConfig?.variant as string,
+    sdk_version: initConfiguration.additionalConfig?.sdk_version as string,
```
thomas-lebeau:
```suggestion
    source: initConfiguration.additionalConfig?.source || 'browser',
    variant: initConfiguration.additionalConfig?.variant,
    sdk_version: initConfiguration.additionalConfig?.sdk_version,
```

packages/core/src/domain/configuration/tags.ts
mormubis: `browser_sdk_version` is not a tag, it's an attribute in the event and it's set on the `defaultContext`.

Should we include it as a tag too?

packages/core/src/domain/configuration/configuration.ts
mormubis: I followed the definition of mobile. But I'm happy to restrict it even more. 

What I like from this solution is that there is no documentation anywhere of this field.

What are your thoughts?

packages/core/src/domain/configuration/endpointBuilder.ts
mormubis: This won't work for Logs, isn't it?

thomas-lebeau: indeed, but if we add these to logs, it will be available in the customer data because logs does not do anything special with `_dd`, unlike RUM.

But we're sending them as tag too, and these will works for log (event though they will also be public :( )

packages/core/src/domain/configuration/configuration.ts
amortemousque: Are we sure we want to expose these attributes in initConfiguration? It would allow customers, even without plugins to override values like source, which might not be desirable. Wdyt?

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -22,16 +26,17 @@ export function startDefaultContext(
           session_replay_sample_rate: round(configuration.sessionReplaySampleRate, 3),
           profiling_sample_rate: round(configuration.profilingSampleRate, 3),
         },
-        browser_sdk_version: canUseEventBridge() ? __BUILD_ENV__SDK_VERSION__ : undefined,
+        browser_sdk_version: canUseEventBridge() || isSourceOverridden ? __BUILD_ENV__SDK_VERSION__ : undefined,
```
amortemousque: **💬 suggestion:** To simplify it, ‏I'm wondering if we could just send `browser_sdk_version` everytime, regardless of the bridge or override?

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -10,9 +10,13 @@ export function startDefaultContext(
   configuration: RumConfiguration,
   sdkName: 'rum' | 'rum-slim' | 'rum-synthetics' | undefined
 ) {
-  hooks.register(
-    HookNames.Assemble,
-    ({ eventType }): DefaultRumEventAttributes => ({
+  hooks.register(HookNames.Assemble, ({ eventType }): DefaultRumEventAttributes => {
+    const source = configuration.source
```
amortemousque, line 14: **❓ question:** If the source is overridden, shouldn’t we still send `source:browser` in a separate field like we do with the version? It could help with debugging. We might also consider sending both sources, similar to how we handle WebViews

packages/core/src/domain/configuration/configuration.ts
amortemousque: As discussed, since both RUM and Logs need this capability and Logs doesn’t support plugins. I agree with using initConfig and marking the option as internal. Let’s keep it simple and include this in our architectural OKR.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -383,5 +411,8 @@ export function serializeConfiguration(initConfiguration: InitConfiguration) {
     allow_untrusted_events: !!initConfiguration.allowUntrustedEvents,
     tracking_consent: initConfiguration.trackingConsent,
     use_allowed_tracking_origins: Array.isArray(initConfiguration.allowedTrackingOrigins),
+    source: initConfiguration.source || 'browser',
```
thomas-lebeau: **🔨 warning:** ‏ For telemetry, the convention is to instrument what was set by the user as is and to not apply any defaults or validation.
```suggestion
    source: initConfiguration.source,
```

packages/core/src/domain/configuration/endpointBuilder.spec.ts
```diff
@@ -144,4 +144,27 @@ describe('endpointBuilder', () => {
       )
     })
   })
+
+  describe('source and variant configuration', () => {
+    it('should use the default source when no configuration is provided', () => {
+      const config = { ...initConfiguration, source: 'browser' }
```
thomas-lebeau: **❓ question:** ‏ The test says "when no configuration is provided", so you should not set the source, no?

```suggestion
      const config = initConfiguration
```

packages/core/src/domain/configuration/endpointBuilder.spec.ts
```diff
@@ -144,4 +144,27 @@ describe('endpointBuilder', () => {
...
+    })
+
+    it('should use source and variant when provided', () => {
+      const config = { ...initConfiguration, source: 'my-source', variant: 'my-variant' }
+      const endpoint = createEndpointBuilder(config, 'rum').build('fetch', DEFAULT_PAYLOAD)
+      expect(endpoint).toContain('ddsource=my-source')
+      expect(endpoint).toContain('_dd.variant=my-variant')
+    })
+
+    it('should include sdk_version when provided in additionalConfig', () => {
```
thomas-lebeau: **💬 suggestion:** ‏ why not check for the sdk_version inside the previous test instead of it's own test?

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -261,6 +276,11 @@ export interface Configuration extends TransportConfiguration {
   flushTimeout: Duration
   batchMessagesLimit: number
   messageBytesLimit: number
+
+  // internal
+  sdkVersion?: string | undefined
+  source: string
```
thomas-lebeau: **💬 suggestion:** ‏ I think during one of the meeting we talked about enforcing source to be either `browser` or `flutter` so we avoid misuse.

```suggestion
  source: string
```
```suggestion
  source: 'browser' | 'flutter'
```

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -360,6 +380,14 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
      */
     batchMessagesLimit: 50,
     messageBytesLimit: 256 * ONE_KIBI_BYTE,
+
+    /**
+     * The source of the SDK, used for support plugins purposes.
+     */
+    source: initConfiguration?.source || 'browser',
```
thomas-lebeau:
```suggestion
    source: initConfiguration?.source === 'flutter' ? 'flutter' : 'browser',
```

packages/rum-core/src/domain/contexts/defaultContext.ts
thomas-lebeau: Sounds like a good idea

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -22,16 +26,17 @@ export function startDefaultContext(
           session_replay_sample_rate: round(configuration.sessionReplaySampleRate, 3),
           profiling_sample_rate: round(configuration.profilingSampleRate, 3),
         },
-        browser_sdk_version: canUseEventBridge() ? __BUILD_ENV__SDK_VERSION__ : undefined,
+        browser_sdk_version: canUseEventBridge() || isSourceOverridden ? __BUILD_ENV__SDK_VERSION__ : undefined,
         sdk_name: sdkName,
+        ...(variant || version ? { variant, version } : {}),
```
thomas-lebeau:
```suggestion
        ...(variant ? { variant } : {}),
        ...(version ? { version } : {}),
```

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -22,16 +26,17 @@ export function startDefaultContext(
...
       },
       application: {
         id: configuration.applicationId,
       },
       date: timeStampNow(),
-      source: 'browser',
-    })
-  )
+      source: source as DefaultRumEventAttributes['source'],
```
thomas-lebeau: **💬 suggestion:** ‏ If you set `initConfiguration.source` as `'flutter' | 'browser'` in configuration.ts, then I think you won't need this anymore

packages/core/src/domain/configuration/tags.ts
thomas-lebeau: yes, I think, otherwise with logs we won't know the version of our SDK when it comes from flutter web

packages/core/src/domain/configuration/endpointBuilder.spec.ts
mormubis: I'm not sure why we call it `initConfiguration` in the test but it's the representation of the internal configuration, hence it should contain `browser` or `flutter`.

packages/core/src/domain/configuration/configuration.ts
mormubis: Resolving the conversation as we moved to some other option

packages/rum-core/src/domain/contexts/defaultContext.ts
mormubis: In the last version, we always send `browser_sdk_version`, meaning that we can always know if the event belongs to us and which version sent it.

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -81,13 +81,13 @@ export function buildEndpointHost(
  * request, as they change randomly.
  */
 function buildEndpointParameters(
-  { clientToken, internalAnalyticsSubdomain }: InitConfiguration,
+  { clientToken, internalAnalyticsSubdomain, source = 'browser' }: InitConfiguration,
```
thomas-lebeau, line 88: **🔨 warning:** ‏ I think we should do the same validation of `source` than in `validateAndBuildConfiguration()`, otherwise, we might have different value as a tag and as an attribute.

packages/rum-core/src/domain/contexts/defaultContext.ts
```diff
@@ -22,16 +20,15 @@ export function startDefaultContext(
           session_replay_sample_rate: round(configuration.sessionReplaySampleRate, 3),
           profiling_sample_rate: round(configuration.profilingSampleRate, 3),
         },
-        browser_sdk_version: canUseEventBridge() ? __BUILD_ENV__SDK_VERSION__ : undefined,
```
thomas-lebeau: **🔨 warning:** ‏ let's confirm with the mobile team how they are using this property, if we can remove it safely, then all good, but otherwise I propose to keep it for now and start working together with the mobile team to remove that property from here and. rely on the one in `ddtags`

packages/core/src/domain/configuration/endpointBuilder.ts
mormubis: Done!

packages/core/src/domain/configuration/tags.ts
```diff
@@ -7,8 +7,8 @@ export const TAG_SIZE_LIMIT = 200
 declare const __BUILD_ENV__SDK_VERSION__: string
 
 export function buildTags(configuration: Configuration): string[] {
-  const { env, service, version, datacenter } = configuration
-  const tags = [buildTag('sdk_version', __BUILD_ENV__SDK_VERSION__)]
+  const { env, service, version, datacenter, sdkVersion, variant } = configuration
+  const tags = [buildTag('sdk_version', sdkVersion ?? __BUILD_ENV__SDK_VERSION__)]
```
mormubis, line 11: @amortemousque we couldn't always send `browser_sdk_version` as it is currently used in Web UI to detect if the event is coming from a WebView.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -360,6 +386,14 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
      */
     batchMessagesLimit: 50,
     messageBytesLimit: 256 * ONE_KIBI_BYTE,
+
+    /**
+     * The source of the SDK, used for support plugins purposes.
+     */
+    source: initConfiguration?.source === 'flutter' ? 'flutter' : 'browser',
```
amortemousque: **💬 suggestion:** ‏We’re starting to duplicate validation logic in multiple places: 
- `source` in this file and `endpointBuilder`
- `site` in `transportConfiguration` and `endpointBuilder`. 

We could centralize these checks in `computeTransportConfiguration`.
```ts
  const source = initConfiguration.source === 'flutter' ? 'flutter' : 'browser'

  const endpointBuilders = computeEndpointBuilders({ ...initConfiguration, source, site })
  const replicaConfiguration = computeReplicaConfiguration({ ...initConfiguration, source, site })

  return {
    replica: replicaConfiguration,
    site,
    source,
    ...endpointBuilders,
  }
}
```

mormubis: Done!

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -360,6 +386,13 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
      */
     batchMessagesLimit: 50,
     messageBytesLimit: 256 * ONE_KIBI_BYTE,
+
+    /**
+     * The source of the SDK, used for support plugins purposes.
+     */
+    variant: initConfiguration?.variant,
+    sdkVersion: initConfiguration?.sdkVersion,
```
BenoitZugmeyer: **suggestion:** no need for optional chaining operator

```suggestion
    variant: initConfiguration.variant,
    sdkVersion: initConfiguration.sdkVersion,
```

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -261,6 +282,11 @@ export interface Configuration extends TransportConfiguration {
   flushTimeout: Duration
   batchMessagesLimit: number
   messageBytesLimit: number
+
+  // internal
+  sdkVersion?: string | undefined
+  source: 'browser' | 'flutter'
+  variant?: string | undefined
```
BenoitZugmeyer: **suggestion:** make everything required

```suggestion
  sdkVersion: string | undefined
  source: 'browser' | 'flutter'
  variant: string | undefined
```

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -261,6 +282,11 @@ export interface Configuration extends TransportConfiguration {
   flushTimeout: Duration
   batchMessagesLimit: number
   messageBytesLimit: number
+
+  // internal
+  sdkVersion?: string | undefined
```
BenoitZugmeyer: **suggestion:** use `sdkVersion` over `__BUILD_ENV_SDK_VERSION___` in more places?

mormubis: Do you mean to replace `__BUILD_ENV_SDK_VERSION___` in more places? Do you have any place in mind?

PR #3661, opened 2025-06-30
[RUM Profiler] Sticky sampling based on session

thomasbertet:
## Motivation

- We want the `profilingSampleRate` to be relative to the sampled sessions. Currently it's relative to page loads.
- We want to use consistent logic with the Trace sampling logic.

## Changes

- Move the `sampler` to its own `domain` folder to be re-used for the RUM Profiler.
- Revamp the sampler cache to use the `sampleRate` as cache key.
- Unique `isSampled` to be used by both Trace and Profiling feature.

## Test instructions
- Added unit test to avoid conflicts of feature decisions override.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/sampler/sampler.ts
```diff
@@ -13,8 +27,8 @@ export function isTraceSampled(sessionId: string, sampleRate: number) {
     return false
   }
 
-  if (sampleDecisionCache && sessionId === sampleDecisionCache.sessionId) {
-    return sampleDecisionCache.decision
+  if (sampleDecisionPerFeatureCache[feature] && sessionId === sampleDecisionPerFeatureCache[feature].sessionId) {
+    return sampleDecisionPerFeatureCache[feature].decision
   }
```
BenoitZugmeyer: **suggestion:** use the sampleRate as cache key instead of introducing `feature`?

thomasbertet: That's an interesting idea 🤔 It's a bit less straightforward I'd say.
Also, it's a bit changing the existing behaviour with `isTraceSampled` (unless I've read the code incorrectly)  as, today, if you call twice the method with two different sample rates, you would still get the first cached decision. If we were to use the sample rate as cache key, it would change this behavior.

BenoitZugmeyer: In practice the sample rate cannot change once the SDK is initialized, so unless the SDK is initialized twice with different sample rates (which.... is probably a bad idea), we should be good.

thomasbertet: I did something 😬 
I used the sample rate as cache key, and I then simplified the logic to only have a single `isSampled` method instead of variants for each feature (profiling, trace). 

WDYT ?

BenoitZugmeyer (approved):

packages/rum-core/src/domain/sampler/sampler.ts
BenoitZugmeyer: Thank you, I like it!

PR #3662, opened 2025-06-30
👷‍♀️ [devx] Add cursor rules to improve testing experience

cy-moi:
## Motivation
We are lacking test documentations in the repo for cursor to understand how to collaborate.

## Changes
Added `package-management`, `unit-test` and TDD rules to guide the agent.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/.cursor/rules/unit-test-best-practices.mdc
+++ b/.cursor/rules/unit-test-best-practices.mdc
@@ -24,7 +24,7 @@ yarn test:unit
 yarn test:unit --spec packages/core/src/browser/addEventListener.spec.ts
 
 # Pattern matching (if filename is unique)
-yarn test:unit --spec "**/addEventListener.spec.ts"
+yarn test:unit --spec "packages/**/addEventListener.spec.ts"
 
 # Multiple files with wildcards
 yarn test:unit --spec "packages/core/**/browser/*.spec.ts"


.cursor/rules/package-management.mdc
```diff
@@ -0,0 +1,11 @@
+---
+description: 
+globs: 
+alwaysApply: true
+---
```
BenoitZugmeyer: **question:** is this header duplication intended?

.cursor/rules/test-driven-development.mdc
```diff
@@ -0,0 +1,138 @@
...
+
+## 2. Spec First, Code Second
+
+**Before writing ANY implementation:**
+```bash
+# Check for existing spec file
+ls packages/*/src/path/to/feature.spec.ts
+
+# If missing, create spec file first
+touch packages/core/src/domain/newFeature.spec.ts
```
BenoitZugmeyer: **suggestion:** adjust the paths so they are the same, also the glob `*` might be confusing:

```suggestion
# Check for existing spec file
ls packages/some-package/src/path/to/feature.spec.ts

# If missing, create spec file first
touch packages/some-package/src/path/to/feature.spec.ts
```

.cursor/rules/unit-test-best-practices.mdc
```diff
@@ -0,0 +1,671 @@
...
+  subscription.unsubscribe()  // Might not execute
+  element.remove()           // Memory leak potential
+})
+
+// DON'T: Forget to clean up DOM
+it('should test DOM interaction', () => {
+  const element = document.createElement('div')
+  document.body.appendChild(element)
+  // Missing cleanup - memory leak!
+})
```
BenoitZugmeyer: **suggestion:** it should prefer using the [`appendElement`](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/test/dom.ts#L14) helper instead of creating dom like this

.cursor/rules/test-driven-development.mdc
```diff
@@ -0,0 +1,138 @@
...
+---
+
+# Test-Driven Development (TDD) Rule
+
+## 1. RED-GREEN-REFACTOR Cycle
+
+**ALWAYS follow this sequence:**
+a) 🔴 **RED**: Write failing test first
+b) 🟢 **GREEN**: Write minimal code to make test pass
+c) 🔵 **REFACTOR**: Improve code while keeping tests green
```
thomas-lebeau: **🥜 nitpick:** ‏ used numbers instead of letter so it formats better in markdown.

```suggestion
1. 🔴 **RED**: Write failing test first
2. 🟢 **GREEN**: Write minimal code to make test pass
3. 🔵 **REFACTOR**: Improve code while keeping tests green
```

.cursor/rules/unit-test-best-practices.mdc
```diff
@@ -0,0 +1,671 @@
...
+  beforeEach(() => {
+    // Setup for each test
+    variable = setupVariable()
+    mockObject = createMock()
+  })
+  
+  afterEach(() => {
+    // Cleanup after each test (when needed)
+    cleanupResources()
+  })
```
thomas-lebeau: **🔨 warning:** ‏ This seems to contradict the "Resource Cleanup Management" section that recommends using `registerCleanupTask()`

.cursor/rules/package-management.mdc
cy-moi: Ah definitely not. Cursor turns the header into UI automatically, yet I didn't see them.

.cursor/rules/unit-test-best-practices.mdc
```diff
@@ -0,0 +1,211 @@
...
+```bash
+# Full path
+yarn test:unit --spec packages/core/src/browser/addEventListener.spec.ts
+
+# Pattern matching (if filename is unique)
+yarn test:unit --spec "**/addEventListener.spec.ts"
+
+# Multiple files with wildcards
+yarn test:unit --spec "packages/core/**/browser/*.spec.ts"
+```
```
thomas-lebeau: I think this does not work with Karma, no? (Or at least I can't seems to make that work)

cy-moi: Indeed, it actually always runs the same tests. I removed the wildcards.

.cursor/rules/unit-test-best-practices.mdc
```diff
@@ -0,0 +1,211 @@
...
+yarn test:unit
+```
+
+#### Run Single File
+```bash
+# Full path
+yarn test:unit --spec packages/core/src/browser/addEventListener.spec.ts
+
+# Pattern matching (if filename is unique)
+yarn test:unit --spec "**/addEventListener.spec.ts"
```
thomas-lebeau: **🔨 warning:** ‏ this will also match some files in `test/app/*`

```
packages/core/src/browser/addEventListener.spec.ts
test/apps/react/node_modules/@datadog/browser-core/src/browser/addEventListener.spec.ts
test/apps/vanilla/node_modules/@datadog/browser-core/src/browser/addEventListener.spec.ts
test/apps/base-extension/node_modules/@datadog/browser-core/src/browser/addEventListener.spec.ts
```

it's not very convenient, but the following works:

```suggestion
yarn test:unit --spec "packages/**/addEventListener.spec.ts"
```

PR #3663, opened 2025-06-30
👷[RUM-9918] add a node version check

bcaudan:
## Motivation

When a new node version is available, we need to manually bump the ci-image version and update the version used by volta inside package.json, cf [example](https://github.com/DataDog/browser-sdk/pull/3049/commits).
However, we can easily forget to do those manual changes and a potential issue with the node version bump could unexpectedly appear when the docker image is rebuilt.

## Changes

Add a node version check job that ensure versions are the same between docker image, volta and from the node process

Output:

```
 Check that node version across configurations are matching...

 docker: 24.3.0
 volta: 23.11.1
 process: 23.11.1

 Different node versions detected!

 Ensure to:
 - run `volta pin node@24.3.0`
 - bump `CURRENT_CI_IMAGE` and run `ci-image` gitlab job
```
## Test instructions

Locally:

- running `node scripts/check-node-version.js` should be successful
- running `volta pin node@24.3.0` and `node scripts/check-node-version.js` should fail

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/check-node-version.js
+++ b/scripts/check-node-version.js
@@ -3,7 +3,6 @@ const readline = require('readline')
 const path = require('path')
 const packageJson = require('../package.json')
 const { printLog, printError, runMain } = require('./lib/executionUtils')
-const { command } = require('./lib/command')
 
 runMain(async () => {
   printLog('Check that node version across configurations are matching...\n')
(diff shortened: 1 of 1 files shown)


scripts/check-node-version.js
```diff
@@ -0,0 +1,45 @@
...
+  const rl = readline.createInterface({ input: fileStream })
+  for await (const line of rl) {
+    // node image on first line
+    return extractVersion(line)
+  }
+}
+
+function retrieveCliVersion() {
+  // node cli returns vX.Y.Z
+  return extractVersion(command`node -v`.run())
```
BenoitZugmeyer: **nitpick:** you could use [`process.version`](https://nodejs.org/docs/latest/api/process.html#processversion) or do you expect that `node -v` returns a different version? i.e. the current node process is not the same as the one in the PATH?

thomas-lebeau (approved):

scripts/check-node-version.js
bcaudan: > you could use [process.version](https://nodejs.org/docs/latest/api/process.html#processversion)

sounds good!

> do you expect that node -v returns a different version?

no, when writing scripts, I have a tendency to pick the first thing that come to mind and not challenge it that much 😇
Happy to change to a simpler approach 😉

PR #3669, opened 2025-07-02
[RUM Profiler] add default view names to Profiles

thomasbertet:
## Motivation

- In case the RUM View names are not custom, the Profiler is not able to associate the default RUM View name to the Profile.
- In turn, this means the Profiling data is not linked directly to specific RUM View names and the filters on View names is useless.

Let's have the default RUM View name be determined directly in the SDK (for the Profiler only for now)

## Changes

- Replicate BE "URL" processor to create a default view name based on URL.
- Add unit tests. 
- Use that default view name in case there is not default custom names.

## Test instructions

- Added unit tests to cover our back.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -28,6 +28,7 @@ import { cleanupLongTaskRegistryAfterCollection, getLongTaskId } from './utils/l
 import { mayStoreLongTaskIdForProfilerCorrelation } from './profilingCorrelation'
 import { transport } from './transport/transport'
 import type { ProfilingContextManager } from './profilingContext'
+import { getCustomOrDefaultViewName } from './utils/getCustomOrDefaultViewName'
 
 export const DEFAULT_RUM_PROFILER_CONFIGURATION: RUMProfilerConfiguration = {
   sampleIntervalMs: 10, // Sample stack trace every 10ms
@@ -124,7 +125,7 @@ export function createRumProfiler(
     // Whenever the View is updated, we add a views entry to the profiler instance.
     const viewUpdatedSubscription = lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (view) => {
       // Note: `view.name` is only filled when users use manual view creation via `startView` method.
-      collectViewEntry({ viewId: view.id, viewName: view.name, startClocks: view.startClocks })
+      collectViewEntry({ viewId: view.id, viewName: getCustomOrDefaultViewName(view.name, document.location.pathname), startClocks: view.startClocks })
     })
     cleanupTasks.push(viewUpdatedSubscription.unsubscribe)
 
--- a/packages/rum/src/domain/profiling/utils/getDefaultViewName.ts
+++ b/packages/rum/src/domain/profiling/utils/getDefaultViewName.ts
@@ -1,0 +1,15 @@
+
+// This is the regex used to extract the path from the url (from SimpleUrlGroupingProcessor.java)
+const PATH_MIXED_ALPHANUMERICS = new RegExp('(?<=/)(?![vV]\\d{1,2}/)(?:[^\\/\\d\\?]*[\\d]+[^\\/\\?]*)', 'g');
+
+
+export function getDefaultViewName(viewPathUrl: string):string {
+    if (!viewPathUrl) {
+        return '/';
+    } 
+
+    // Replace all the mixed alphanumerics with a ?
+    return viewPathUrl.replace(PATH_MIXED_ALPHANUMERICS, '?');
+}
+
+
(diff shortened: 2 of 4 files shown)


packages/rum/src/domain/profiling/profiler.spec.ts
```diff
@@ -205,6 +205,61 @@ describe('profiler', () => {
     expect(lastCall.longTasks[0].id).toBeDefined()
     expect(lastCall.longTasks[0].startClocks.relative).toBe(12345 as RelativeTime)
   })
+
+  it('should collect views and set default view name in the Profile', async () => {
```
BenoitZugmeyer, line 241: **nitpick:** those tests are huge and should probably be factorized or split

packages/rum/src/domain/profiling/profiler.ts
```diff
@@ -60,7 +61,11 @@ export function createRumProfiler(
     // Add initial view
     // Note: `viewEntry.name` is only filled when users use manual view creation via `startView` method.
     lastViewEntry = viewEntry
-      ? { startClocks: viewEntry.startClocks, viewId: viewEntry.id, viewName: viewEntry.name }
+      ? {
+          startClocks: viewEntry.startClocks,
+          viewId: viewEntry.id,
+          viewName: getCustomOrDefaultViewName(viewEntry.name, document.location.pathname),
```
BenoitZugmeyer, line 67: **suggestion:** use the `urlHistory` similarly to `viewHistory` to get the URL. It's closer to what we do to assign the URL to View events, and you won't need to change and restore the real document location in your tests.

packages/rum/src/domain/profiling/utils/getDefaultViewName.ts
```diff
@@ -0,0 +1,11 @@
+// This is the regex used to extract the path from the url (from SimpleUrlGroupingProcessor.java)
+const PATH_MIXED_ALPHANUMERICS = new RegExp('(?<=/)(?![vV]\\d{1,2}/)(?:[^\\/\\d\\?]*[\\d]+[^\\/\\?]*)', 'g')
```
BenoitZugmeyer: **suggestion:** use the regex syntax. It's easier to read, ESLint will give some advices, and it's shorter:

```suggestion
const PATH_MIXED_ALPHANUMERICS = /(?<=\/)(?![vV]\d{1,2}\/)(?:[^/\d?]*[\d]+[^/?]*)/g
```

packages/rum/src/domain/profiling/utils/getDefaultViewName.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function getDefaultViewName(viewPathUrl: string): string {
+  if (!viewPathUrl) {
+    return '/'
+  }
```
BenoitZugmeyer, line 7: **suggestion:** I don't think this is useful, as view paths should not be empty in practice

packages/rum/src/domain/profiling/profiler.spec.ts
```diff
@@ -205,6 +205,61 @@ describe('profiler', () => {
...
+
+    // Wait for stop of collection.
+    await waitForBoolean(() => profiler.isStopped())
+
+    // Go back to the original pathname
+    history.pushState({}, '', originalPathname)
+
+    expect(profilingContextManager.get()?.status).toBe('stopped')
+
+    const lastCall: RumProfilerTrace = sendProfileSpy.calls.mostRecent().args[0] as unknown as RumProfilerTrace
```
BenoitZugmeyer: **suggestion:** Apply proper typing to `sendProfileSpy` so you don't need to cast things.

```ts
  let sendProfileSpy: jasmine.Spy<typeof transport.sendProfile>
```

packages/rum/src/domain/profiling/profiler.spec.ts
thomasbertet: I like it extra explicit for now if that's OK, but I get the idea !

packages/rum/src/domain/profiling/utils/getDefaultViewName.ts
thomasbertet: I think it's safer to keep it as it's mimic-ing what we have in the BE ([implementation](https://github.com/DataDog/logs-backend/blob/prod/domains/event-platform/libs/processing/processing-common/src/main/java/com/dd/logs/processing/processors/SimpleUrlGroupingProcessor.java#L25), [test](https://github.com/DataDog/logs-backend/blob/prod/domains/event-platform/libs/processing/processing-common/src/test/java/com/dd/logs/processing/processors/SimpleUrlGroupingProcessorTest.java))

packages/rum/src/domain/profiling/profiler.ts
thomasbertet: Discussed together in Slack, since there is two points of collection for views (initially in `start()` and then using the view's lifecycle) let's keep using the `document.location` object for now.

BenoitZugmeyer (approved):

PR #3671, opened 2025-07-03
✨ [RUM-8780] Make developer extension capable of injecting the browser SDK

BeltranBulbarellaDD:
## Motivation

The idea is to be able to inject the SDK on a page that does not have it. Injecting the local version and CDN version.

## Changes
Added a Trial tab that allows to configure the SDK to be injected.
Screenshot 2025-07-17 at 10 40 02

If CDN injection is enabled, the setting's tab gets set to:
`Event collection strategy: Requests` and
`Auto Flush` as it is needed for CDN.
Screenshot 2025-07-17 at 10 32 57

I can note an issue which is that when I `Disable event intake (do not send data to Datadog)` for which I do: 
`{
        beforeSend: () => false,
      }`
For which the initial view gets created but no other information gets sent.

## Test instructions

Injecting using local version:
https://github.com/user-attachments/assets/b63b6fc2-5901-4703-88d3-798d9ea4be49

Injecting CDN version:
https://github.com/user-attachments/assets/874df7c6-b921-40eb-b01d-84fbba7a2423

Testing Session Replay tab on CDN injection.
https://github.com/user-attachments/assets/edab2e73-f596-428f-b89b-29a8f6d21d23

## Checklist

- [X ] Tested locally
- [ ] Tested on staging


developer-extension/README.md
```diff
@@ -36,6 +36,7 @@ Then, in Google Chrome:
 - Load the SDK development bundles instead of production ones
 - Switch between `rum` and `rum-slim` bundles
 - Retrieve Logs/RUM configuration
+- **Trial Mode**: Enable SDK injection capabilities to automatically inject and configure the Datadog Browser SDK into pages that don't have it
```
sethfowler-datadog, line 39: Nit: maybe reformulate this so it matches the other items in the list a bit better. As an example:
```suggestion
- Trial the SDK locally without deploying the snippet
```

Doesn't have to be that; I'm just trying to match the "voice" of the other items with a brief description. Then, the longer description you wrote below can explain the details of what's available.

developer-extension/README.md
```diff
@@ -83,6 +84,8 @@ Info tab contains information about Session and RUM SDK configurations
 
 - **Request Interception**: override the current SDK bundle with local build, or ​​switch between `rum` and `rum-slim` bundles on any site that is using RUM SDK. (note: if the SDK is installed from NPM, this override might not work, as it is still in an experimental stage.)
 
+- **Trial Mode**: Access SDK injection capabilities through the Trial tab. When enabled, you can configure and inject the Datadog Browser SDK into pages that don't have it. Configure your Application ID, Client Token, site, and other parameters to test SDK features on any website.
```
sethfowler-datadog, line 87: No action requested in this PR, but: "Request Interception" and "Trial Mode" feel like they have a bit of overlap, since they both involve injection. I don't think it's necessary to change things in this PR, because handling it cleanly would probably require changes to the some of the wording in the extension UI, but it'd be nice to rename things a bit to clarify when you would use each feature. It feels like "Request Interception" is really "SDK Version Swapping", or something like that.

Edit: actually, looking at the settings panel changes, I'm starting to think that maybe we should address this issue in this PR. We may want to outright remove existing features that are duplicated by Trial Mode, so that there's a single clear place to accomplish each task.

developer-extension/src/background/domain/messageRelay.ts
```diff
@@ -51,10 +51,17 @@ onDevtoolsLastDisconnection.subscribe(() => {
...
 // devtools panel connection.
-chrome.runtime.onMessage.addListener((message, sender) => {
-  if (sender.tab && sender.tab.id) {
+chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
+  if (sender.tab && sender.tab.id && message && typeof message === 'object') {
+    if ('type' in message && message.type === 'get_settings_for_content_script') {
+      chrome.storage.local.get(null, (allSettings) => {
+        sendResponse({ settings: allSettings })
+      })
+      return true
```
sethfowler-datadog, line 60: Nit: IMO it'd be nicer to return a Promise, which AFAIK is also supported. I had to look up what `return true` mean to understand what was happening here.

developer-extension/src/common/extension.types.ts
```diff
@@ -57,4 +57,30 @@ export interface Settings {
...
   debugMode: boolean
   datadogMode: boolean
+  trialMode: boolean
+  sdkInjection: SdkInjectionConfig
+  bundleSource: 'dev' | 'cdn'
+}
+
+export interface SdkInjectionConfig {
+  enabled: boolean
+  sdkTypes: Array<'rum' | 'logs'>
```
sethfowler-datadog, line 67: Nit: maybe `Record<'rum' | 'logs', boolean>` would be a bit nicer.

developer-extension/src/common/packagesUrlConstants.ts
```diff
@@ -3,6 +3,11 @@ export const DEV_LOGS_URL = `${DEV_SERVER_ORIGIN}/datadog-logs.js`
 export const DEV_RUM_SLIM_URL = `${DEV_SERVER_ORIGIN}/datadog-rum-slim.js`
 export const DEV_RUM_URL = `${DEV_SERVER_ORIGIN}/datadog-rum.js`
 
+export const CDN_BASE_URL = 'https://www.datadoghq-browser-agent.com'
+export const CDN_RUM_URL = `${CDN_BASE_URL}/v6/datadog-rum.js`
+export const CDN_RUM_SLIM_URL = `${CDN_BASE_URL}/v6/datadog-rum-slim.js`
+export const CDN_LOGS_URL = `${CDN_BASE_URL}/v6/datadog-logs.js`
```
sethfowler-datadog: If there's a way to avoid hard-coding `v6`, that'd be great, but I realize that might not be easy. If we need to hard-code it and manually bump the version, please split it out into a separate constant like `PROD_REPLAY_SANDBOX_VERSION`, so when we update it in the future the PR will be a one-liner.

developer-extension/src/content-scripts/isolated.ts
```diff
@@ -5,6 +5,18 @@ import { createLogger } from '../common/logger'
 
 const logger = createLogger('content-script-isolated')
 
+chrome.runtime.onMessage.addListener((message) => {
+  if (message && typeof message === 'object' && 'type' in message && message.type === 'settings_updated') {
```
sethfowler-datadog: Nit: Perhaps this version would work?
```suggestion
  if (message?.type === 'settings_updated') {
```

developer-extension/src/content-scripts/isolated.ts
```diff
@@ -37,3 +51,21 @@ function browserSdkMessageListener(event: unknown) {
     }
   }
 }
+
+function getSettingsListener() {
+  try {
+    chrome.runtime.sendMessage({ type: 'get_settings_for_content_script' }, (response) => {
+      if (response && response.settings) {
```
sethfowler-datadog: Likewise:
```suggestion
      if (response?.settings) {
```

developer-extension/src/content-scripts/main.ts
```diff
@@ -136,15 +178,24 @@ function loadSdkScriptFromURL(url: string) {
...
+    // Webpack v6+ changed the wording of the runtime guard that throws when `scriptUrl` cannot be
+    // determined.  Patch *any* occurrence of that guard so `scriptUrl` is initialised with the
+    // actual bundle URL we just downloaded, ensuring chunks/workers resolve correctly.
     sdkCode = sdkCode.replace(
-      'if (!scriptUrl) throw new Error("Automatic publicPath is not supported in this browser");',
+      /if\s*\(!scriptUrl\)\s*throw new Error\([^)]*Automatic publicPath[^)]*\);/,
       `if (!scriptUrl) scriptUrl = ${JSON.stringify(url)};`
     )
+    // Fallback – if the above pattern did not match (future wording change), explicitly set the
+    // webpack public path at the very top of the bundle so chunk loading keeps working.
```
sethfowler-datadog, line 192: If this is reliable, heck, we could just get rid of the old method!

developer-extension/src/panel/components/tabs/trialTab.tsx
```diff
@@ -0,0 +1,220 @@
...
+  return (
+    <TabBase>
+      <Columns>
+        <Columns.Column title="Trial Mode">
+          <Text>After each change please reload the page.</Text>
+          {trialMode && (
+            <>
+              <Space h="lg" />
+              <Text size="lg" fw={600}>
+                SDK Injection Configuration
```
sethfowler-datadog, line 21: It feels a bit confusing that we're using the term "injection" both here and as one of the override strategies on the other page, and that one prevents you from using the other, even though they're very nearly the same thing. I think this ties back to the concern I had that the two features are confusingly similar in the docs, but I think the issue is even more visible when looking at how the different settings panels interact.

Perhaps we should just get rid of the override stuff that's duplicated by Trial Mode?

developer-extension/src/panel/hooks/useEvents/eventCollection.ts
```diff
@@ -58,8 +58,10 @@ function listenEventsFromRequests(callback: (events: SdkEvent[]) => void) {
     if (!INTAKE_DOMAINS.find((rootDomain) => url.hostname.endsWith(rootDomain))) {
       return
     }
-    // intake request path is /api/vX/track
-    if (!['rum', 'logs'].includes(url.pathname.split('/')[3])) {
+    // Accept various intake path formats: older /api/vX/track, newer /v1/input, or anything containing /rum or /logs.
```
sethfowler-datadog, line 61: It seems like you are low-key fixing a bunch of existing bugs here, in addition to implementing Trial Mode. 😄

developer-extension/src/panel/hooks/useSettings.ts
```diff
@@ -46,12 +70,39 @@ async function loadSettingsFromStorage() {
 
 function setSetting<Name extends keyof Settings>(name: Name, value: Settings[Name]) {
   settings![name] = value
+
+  // Special case: keep bundleSource and sdkInjection.bundleSource in sync
```
sethfowler-datadog, line 74: Another indication that we might want to consolidate things.

developer-extension/src/common/packagesUrlConstants.ts
BeltranBulbarellaDD: Oh nice!

developer-extension/src/background/domain/messageRelay.ts
BeltranBulbarellaDD: `Promise returned in function argument where a void return was expected  @typescript-eslint/no-misused-promises`

So I am having issues returning a Promise. I could eslint-ignore it.

PR #3677, opened 2025-07-07
📝 [RUM-10899] Add API reference using typedoc

amortemousque:
## Motivation

Automatically generate an API reference from code comments. This will let us remove some snippets and the full configuration set from the public documentation

The GitHub page publication will be implemented in a following PR.

## Changes

1. Configure typedoc
2. Export the missing type from the packages
3. Add and update the JSdoc from the exposed types
4. Enforce jsdoc comment with eslint
5. Simplify the TypeDoc theme with custom styling

## Test instructions

```
yarn build
yarn build:docs:html
npx http-server ./docs
```
## Checklist

- [x] Tested locally
- [ ] Tested on staging


typedoc.json
```diff
@@ -0,0 +1,23 @@
+{
+  "entryPoints": ["packages/*"],
+  "entryPointStrategy": "packages",
+  "includeVersion": false,
```
rgaignault: ❓ Question : Could it be interesting to indicate version ?

rgaignault (approved):

docs.json
thomas-lebeau, line 1: **❓ question:** ‏ Do we need to commit this file?

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -16,13 +16,23 @@ import type { Site } from '../intakeSites'
 import type { TransportConfiguration } from './transportConfiguration'
 import { computeTransportConfiguration } from './transportConfiguration'
 
+/**
+ * Default privacy level for the browser SDK.
+ *
+ * See [Replay Privacy Options](https://docs.datadoghq.com/real_user_monitoring/session_replay/browser/privacy_options) for further information.
```
thomas-lebeau: **❓ question:** ‏ maybe we could use the [`@see` tag](https://typedoc.org/documents/Tags._see.html)? 

```suggestion
 * @see [Replay Privacy Options](https://docs.datadoghq.com/real_user_monitoring/session_replay/browser/privacy_options) for further information.
```

docs.json
bcaudan, line 1: **❓ question:** ‏Do we need to version this file?

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -6,8 +6,8 @@ export function containsExtensionUrl(str: string): boolean {
 
 /**
  * Utility function to detect if the SDK is being initialized in an unsupported browser extension environment.
- * @param windowLocation The current window location to check
- * @param stack The error stack to check for extension URLs
+ * @param windowLocation - The current window location to check
+ * @param stack - The error stack to check for extension URLs
```
bcaudan, line 11: **❓ question:** ‏could we add some lint rules for all the conventions you have mentioned in the PR description?

packages/core/src/tools/serialisation/context.ts
```diff
@@ -4,5 +4,8 @@ export interface Context {
 
 export type ContextValue = string | number | boolean | Context | ContextArray | undefined | null
 
+/**
+ * @hidden
```
thomas-lebeau, line 8: **❓ question:** ‏ what's the difference between `@hidden` and `@internal`?

packages/logs/src/entries/main.ts
```diff
@@ -1,16 +1,54 @@
...
+  Site,
+  ConsoleApiName,
+  RawReportType,
+  ErrorSource,
+} from '@datadog/browser-core'
 
+/**
+ * The global Logs instance. Use this to call Logs methods.
+ *
+ * @see {@link https://docs.datadoghq.com/logs/log_collection/javascript/ | Browser Log Collection}
```
thomas-lebeau: **❓ question:** ‏ should we use `[Browser Log Collection](https://docs.datadoghq.com/logs/log_collection/javascript/)` for consistency?

dd-typedoc.css
thomas-lebeau, line 1: **💬 suggestion:** ‏ rename `typedoc.css` so it shows up next to typedoc.json in the explorer.

thomas-lebeau (commented):
**🔨 warning:** ‏ Some types are duplicated
Screenshot 2025-07-09 at 15 15 43

thomas-lebeau:
> Prefer interface over type
> ```diff
> - export type User = { id?: string | undefined, ... }
> + export interface User { id?: string | undefined, ... }
> ```

Could you elaborate why we should prefer interface?

packages/core/src/tools/serialisation/context.ts
amortemousque: @hidden, means that they are not present in the generated doc while @internal can. In our case I activated the `"excludeInternal": true` so it should be the same. In this case I used hidden because typedoc tried to resolve the native `Array` type leading to warning and a weird generated doc.

docs.json
amortemousque: I hesitated to commit it to keep doc changes easier to review. But I think it needs more thought, so I’ll remove it for now.

amortemousque:
> Could you elaborate why we should prefer interface?

We already use both interface and type for similar use cases, so I think it’s worth being consistent, especially now that it affects the public documentation output. interfaces and types are rendered with different titles and grouped in separate sections in the docs.

Also, api-documenter does not preserve comments on type properties, whereas TypeDoc does. To keep maximum compatibility (in case we switch tools later), I think it's better to prefer interface when we can.

packages/core/src/domain/configuration/configuration.ts
amortemousque: We using @see I found the layout not ideal. It shows the link at the end of the page instead of collocated to the description

- With @see
Screenshot 2025-07-10 at 17 23 22

- Without @see
Screenshot 2025-07-10 at 17 24 15

packages/core/src/domain/extension/extensionUtils.ts
amortemousque: Good point! I added ESLint rules to enforce the comment conventions.
There are two more I’d like to add, but they affect code as well (not just comments) so I’ll include them in a separate PR:
- [@typescript-eslint/method-signature-style: ['error', 'method']](https://typescript-eslint.io/rules/method-signature-style/)
- [@typescript-eslint/consistent-type-definitions: ['error', 'interface']](https://typescript-eslint.io/rules/consistent-type-definitions/)

typedoc.json
```diff
@@ -0,0 +1,24 @@
...
+  "includeVersion": true,
+  "exclude": ["packages/core", "packages/rum-core", "packages/flagging", "packages/worker"],
+  "validation": {
+    "notExported": true,
+    "invalidLink": true,
+    "rewrittenLink": true,
+    "notDocumented": false,
+    "unusedMergeModuleWith": true
+  },
+  "visibilityFilters": {}, // disable all visibility filters
```
bcaudan, line 13: **🥜 nitpick:** ‏no comments in JSON
```suggestion
  "visibilityFilters": {},
```

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -16,23 +16,41 @@ import type { Site } from '../intakeSites'
...
+ * See [Connect RUM and Traces](https://docs.datadoghq.com/real_user_monitoring/platform/connect_rum_and_traces/?tab=browserrum) for further information.
+ *
+ */
 export type TraceContextInjection = (typeof TraceContextInjection)[keyof typeof TraceContextInjection]
 
 export interface InitConfiguration {
   /**
    * The client token for Datadog. Required for authenticating your application with Datadog.
+   *
+   * @category Core
```
bcaudan: **❓ question:** ‏For the category, should we have some explanation somewhere on what we expect to have on each category?
Should we also ensure that everything falls into a category?

typedoc.json
amortemousque: According to their [doc](https://typedoc.org/documents/Options.Configuration.html#json-files) 

> Like tsconfig.json, JSON configuration files are parsed as JSONC, which means that you can safely use trailing commas and comments in your file.

And we already have comment in `tsconfig.json`

packages/core/src/domain/configuration/configuration.ts
amortemousque: I was trying to make the page more explorable by grouping items, but the category names are quite opinionated. I’m still unsure about formalizing them for now. I think we can remove the categories for the configuration. In publicApi, the categories are more straightforward, so I’d keep them there. Wdyt?

docs.json
amortemousque: I removed the file. We don’t have a real need for the JSON version of the doc right now.
What I have in mind for next step (and following PRs) are:
- Apply Datadog branding to the theme
- Publish via GitHub Pages
- Set up versioning

typedoc.json
bcaudan: oh, ok then

packages/core/src/domain/configuration/configuration.ts
bcaudan: My concern is that we will probably forgot to use/maintain those categories, so it looks good now but without constraints, it will probably get messy over time.

bcaudan: Discussed together, let's start like that and see how it will go.

bcaudan (approved):

PR #3680, opened 2025-07-08
🐛 prevent log reporting after consent revocation

mormubis:
## Motivation

A customer reported that Logs continue to send information even after consent has been revoked. We discovered that this issue arises from the PR that allows logs to be sent after the session has expired.

## Changes

This PR introduces consent context and utilizes it to verify whether the event was consented to be sent.

Additionally, the consent state is passed to telemetry and check every time we want to send any information.

> [!IMPORTANT]
> Although this change should be applied to all products (RUM and Logs), there is added complexity with the manual and automatic start views. This PR just focused on resolved the problem of blocking sending logs after the consent is revoked.

## Test instructions

**Case 1**
- _Check the consent was granted_
- Send a log with `DD_LOGS.logger.log('something')`
- _Verify that the log was sent_
- Revoke the consent
- Send another log `DD_LOGS.logger.log('else')`
- _Confirm that this log was not sent_

**Case 2**
- _Check the consent was granted_
- Wait for the session to expire
- Send a log with `DD_LOGS.logger.log('something')`
- _Verify that the log was sent_

**Case 3**
- _Check the consent was granted_
- Wait for telemetry to be sent
- _Check the telemetry was sent correctly_
- Revoke the consent
- _Check there are no more telemetry requests_

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/telemetry/telemetry.ts
+++ b/packages/core/src/domain/telemetry/telemetry.ts
@@ -76,16 +76,15 @@ export function startTelemetry(
 ): Telemetry {
   const observable = new Observable<TelemetryEvent & Context>()
 
-  const { stop } = startTelemetryTransport(
+  const { stop } = startTelemetryTransport(configuration, reportError, pageMayExitObservable, createEncoder, observable)
+
+  const { enabled } = startTelemetryCollection(
+    telemetryService,
     configuration,
-    reportError,
-    pageMayExitObservable,
-    createEncoder,
+    hooks,
     observable,
     trackingConsentState
   )
-
-  const { enabled } = startTelemetryCollection(telemetryService, configuration, hooks, observable)
 
   return {
     stop,
@@ -97,7 +96,8 @@ export function startTelemetryCollection(
   telemetryService: TelemetryService,
   configuration: Configuration,
   hooks: AbstractHooks,
-  observable: Observable<TelemetryEvent & Context>
+  observable: Observable<TelemetryEvent & Context>,
+  trackingConsentState: TrackingConsentState
 ) {
   const alreadySentEvents = new Set<string>()
 
@@ -114,6 +114,7 @@ export function startTelemetryCollection(
   onRawTelemetryEventCollected = (rawEvent: RawTelemetryEvent) => {
     const stringifiedEvent = jsonStringify(rawEvent)!
     if (
+      trackingConsentState.isGranted() &&
       telemetryEnabledPerType[rawEvent.type!] &&
       alreadySentEvents.size < configuration.maxTelemetryEventsPerPage &&
       !alreadySentEvents.has(stringifiedEvent)
@@ -169,8 +170,7 @@ function startTelemetryTransport(
   reportError: (error: RawError) => void,
   pageMayExitObservable: Observable<PageMayExitEvent>,
   createEncoder: (streamId: DeflateEncoderStreamId) => Encoder,
-  telemetryObservable: Observable<TelemetryEvent & Context>,
-  trackingConsentState: TrackingConsentState
+  telemetryObservable: Observable<TelemetryEvent & Context>
 ) {
   const cleanupTasks: Array<() => void> = []
   if (canUseEventBridge()) {
@@ -197,9 +197,7 @@ function startTelemetryTransport(
     )
     cleanupTasks.push(() => telemetryBatch.stop())
     const telemetrySubscription = telemetryObservable.subscribe((event) => {
-      if (trackingConsentState.isGranted()) {
-        telemetryBatch.add(event, isTelemetryReplicationAllowed(configuration))
-      }
+      telemetryBatch.add(event, isTelemetryReplicationAllowed(configuration))
     })
     cleanupTasks.push(() => telemetrySubscription.unsubscribe())
   }
--- a/packages/core/src/domain/telemetry/telemetry.ts
+++ b/packages/core/src/domain/telemetry/telemetry.ts
@@ -22,7 +22,6 @@ import { canUseEventBridge, getEventBridge, startBatchWithReplica } from '../../
 import type { Encoder } from '../../tools/encoder'
 import type { PageMayExitEvent } from '../../browser/pageMayExitObservable'
 import { DeflateEncoderStreamId } from '../deflate'
-import type { TrackingConsentState } from '../trackingConsent'
 import { HookNames, DISCARDED } from '../../tools/abstractHooks'
 import type { AbstractHooks, DISCARDED as DISCARDED_TYPE } from '../../tools/abstractHooks'
 import type { TelemetryEvent } from './telemetryEvent.types'
--- a/packages/core/src/domain/telemetry/telemetry.ts
+++ b/packages/core/src/domain/telemetry/telemetry.ts
@@ -78,13 +78,7 @@ export function startTelemetry(
 
   const { stop } = startTelemetryTransport(configuration, reportError, pageMayExitObservable, createEncoder, observable)
 
-  const { enabled } = startTelemetryCollection(
-    telemetryService,
-    configuration,
-    hooks,
-    observable,
-    trackingConsentState
-  )
+  const { enabled } = startTelemetryCollection(telemetryService, configuration, hooks, observable, trackingConsentState)
 
   return {
     stop,
--- a/packages/core/src/domain/telemetry/telemetry.ts
+++ b/packages/core/src/domain/telemetry/telemetry.ts
@@ -71,14 +71,13 @@ export function startTelemetry(
   hooks: AbstractHooks,
   reportError: (error: RawError) => void,
   pageMayExitObservable: Observable<PageMayExitEvent>,
-  createEncoder: (streamId: DeflateEncoderStreamId) => Encoder,
-  trackingConsentState: TrackingConsentState
+  createEncoder: (streamId: DeflateEncoderStreamId) => Encoder
 ): Telemetry {
   const observable = new Observable<TelemetryEvent & Context>()
 
   const { stop } = startTelemetryTransport(configuration, reportError, pageMayExitObservable, createEncoder, observable)
 
-  const { enabled } = startTelemetryCollection(telemetryService, configuration, hooks, observable, trackingConsentState)
+  const { enabled } = startTelemetryCollection(telemetryService, configuration, hooks, observable)
 
   return {
     stop,
@@ -90,8 +89,7 @@ export function startTelemetryCollection(
   telemetryService: TelemetryService,
   configuration: Configuration,
   hooks: AbstractHooks,
-  observable: Observable<TelemetryEvent & Context>,
-  trackingConsentState: TrackingConsentState
+  observable: Observable<TelemetryEvent & Context>
 ) {
   const alreadySentEvents = new Set<string>()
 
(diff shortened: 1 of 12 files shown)


packages/core/src/domain/trackingConsent.ts
```diff
@@ -9,25 +12,35 @@ export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingCons
...
     update(trackingConsent: TrackingConsent) {
       currentConsent = trackingConsent
+      history.closeActive(relativeNow())
+      history.add(trackingConsent, relativeNow())
       observable.notify()
     },
-    isGranted() {
-      return currentConsent === TrackingConsent.GRANTED
+    isGranted(startTime?: RelativeTime) {
+      const value = startTime === undefined ? currentConsent : (history.find(startTime) ?? currentConsent)
```
mormubis: > [!NOTE]
> Just for the sake of types, `history.find` can return `undefined`, it should never be empty, but fallback to `currentConsent` fixes the types

packages/core/src/domain/trackingConsent.ts
```diff
@@ -9,25 +13,36 @@ export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingCons
...
   update: (trackingConsent: TrackingConsent) => void
-  isGranted: () => boolean
+  isGranted: (startTime?: RelativeTime) => boolean
   observable: Observable<void>
 }
 
 export function createTrackingConsentState(currentConsent?: TrackingConsent): TrackingConsentState {
   const observable = new Observable<void>()
+  // Reuse the same expire delay as the session timeout delay because the tracking consent is expected to be valid for the same duration as the session.
+  const history = createValueHistory<TrackingConsent>({ expireDelay: SESSION_TIME_OUT_DELAY })
```
thomas-lebeau: **💬 suggestion:** ‏ in some other paces, I've see it done like:
```ts
export const CONSENT_TIME_OUT_DELAY = SESSION_TIME_OUT_DELAY
```

and then

```suggestion
  const history = createValueHistory<TrackingConsent>({ expireDelay: CONSENT_TIME_OUT_DELAY })
```

packages/core/src/domain/trackingConsent.ts
amortemousque: why not just?
```ts
isGranted(startTime?: RelativeTime) {
   return history.find(startTime) === TrackingConsent.GRANTED
}
```

packages/core/src/domain/trackingConsent.ts
```diff
@@ -9,25 +13,36 @@ export type TrackingConsent = (typeof TrackingConsent)[keyof typeof TrackingCons
...
 
   return {
     tryToInit(trackingConsent: TrackingConsent) {
       if (!currentConsent) {
         currentConsent = trackingConsent
+        history.add(trackingConsent, relativeNow())
       }
     },
     update(trackingConsent: TrackingConsent) {
       currentConsent = trackingConsent
```
amortemousque: Now that you use the history, you can probably drop the `currentConsent` update and rename the parameter in `createTrackingConsentState` to `initialConsent`. Wdyt?

packages/logs/src/domain/contexts/trackingConsentContext.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function startTrackingConsentContext(hooks: Hooks, trackingConsentState: TrackingConsentState) {
+  hooks.register(HookNames.Assemble, ({ startTime }) => {
+    const wasConsented = trackingConsentState.isGranted(startTime)
+
+    if (!wasConsented) {
+      return DISCARDED
+    }
+
+    return {}
```
amortemousque:
```suggestion
    return SKIPPED
```

packages/core/src/domain/trackingConsent.ts
mormubis: This would change the behavior un RUM.

thomas-lebeau: **💭 thought:** ‏ hmm, `history.find()` CAN return `undefined`, because it expires with the session, however, LOGS can continue to send logs after the session expires.

so you condition is actually correct, but maybe we can make it easier to read

```suggestion
      const value = history.find(startTime) ?? currentConsent
```

packages/core/src/domain/trackingConsent.ts
thomas-lebeau: Also, I don't know if the comment makes sense, because the problem we're trying to solve is exactly how to make logs honor consent withdrawal even though they continue sending logs when the session expires.

So we do need to track consent beyond the duration of a session, although the last value (i.e. `currentConsent`) should be enough

packages/logs/src/domain/contexts/trackingConsentContext.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function startTrackingConsentContext(hooks: Hooks, trackingConsentState: TrackingConsentState) {
+  hooks.register(HookNames.Assemble, ({ startTime }) => {
+    const wasConsented = trackingConsentState.isGranted(startTime)
+
+    if (!wasConsented) {
+      return DISCARDED
+    }
+
+    return SKIPPED
```
rgaignault, line 13: 💬 suggestion : it should probably return an empty object to match the test, or otherwise the test need to match SKIPPED
```suggestion
    return {}
```

packages/core/src/domain/trackingConsent.ts
mormubis: I think it's fine because if you had consent during the session, it means you should have consent after the session?

packages/logs/src/domain/contexts/trackingConsentContext.ts
mormubis: I'll fix the test. I accepted the change per @amortemousque recommendation

packages/core/src/domain/trackingConsent.ts
mormubis: I really want to avoid changing behavior (at the moment) in other parts of the SDK. If we make this change, it will change consent behavior in RUM too.

**After some investigation I found that the history will return the most updated value if `startTime` is not defined**. It should work, making the changes.

Edit: further investigation.

packages/core/src/domain/telemetry/telemetry.ts
thomas-lebeau, line 1: **❓ question:** ‏ I don't know if we should wait for #3685 so we can solve this for log using the same `trackingConsentContext` or keep it as is and refactor once the other PR is merged.

packages/logs/src/boot/startLogs.ts
```diff
@@ -57,7 +58,8 @@ export function startLogs(
     configuration,
     reportError,
     pageMayExitObservable,
-    createIdentityEncoder
+    createIdentityEncoder,
+    trackingConsentState
```
bcaudan: **💬 suggestion:** ‏what about adding some test cases at logs and telemetry level?

mormubis: Added!

packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -185,9 +195,11 @@ function startTelemetryTransport(
       new Observable()
     )
     cleanupTasks.push(() => telemetryBatch.stop())
-    const telemetrySubscription = telemetryObservable.subscribe((event) =>
-      telemetryBatch.add(event, isTelemetryReplicationAllowed(configuration))
-    )
+    const telemetrySubscription = telemetryObservable.subscribe((event) => {
+      if (trackingConsentState.isGranted()) {
+        telemetryBatch.add(event, isTelemetryReplicationAllowed(configuration))
+      }
```
bcaudan: **❓ question:** ‏any reason to do it at transport level compared to collection level?
I would lean toward doing it at collection level to not deal with the event earlier and for example avoid to populate `alreadySentEvents` with events that would not have been sent.

mormubis: That's a good point, I am ok with moving it to the collection level. My thought is to be consistent with next steps. If we plan to block RUM/Logs at the transport level, should we keep it here?

bcaudan: FMU for logs we do it at assembly (so not transport level) and AFAIK for RUM it is unclear for now.
Am I missing something?

mormubis: Following our offline conversation, it currently makes sense to implement the blocking mechanism at the collection level.

packages/logs/src/domain/contexts/trackingConsentContext.ts
thomas-lebeau, line 1: let's move this one to the 'core' package, remove the one in 'rum-core' so we only have one `startTrackingConsentContext` function that register two hooks (`Assemble` and `AssembleTelemetry`)

packages/logs/src/domain/contexts/trackingConsentContext.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  hooks.register(HookNames.AssembleTelemetry, () => {
+    const wasConsented = trackingConsentState.isGranted()
+
+    if (!wasConsented) {
+      return DISCARDED
+    }
+
+    return SKIPPED
+  })
```
thomas-lebeau:
```suggestion
  function handleTrackingConsent() {
      const wasConsented = trackingConsentState.isGranted()

    if (!wasConsented) {
      return DISCARDED
    }

    return SKIPPED
  }

  hooks.register(HookNames.Assemble, handleTrackingConsent)
  hooks.register(HookNames.AssembleTelemetry, handleTrackingConsent)
```

packages/logs/src/domain/contexts/trackingConsentContext.ts
mormubis: But then we could also be discarding RUM events beyond the initial changes that we thought.

What do you think?

mormubis: And the Hooks are not defined outside of `rum-core` or `logs` because the hooks are different for each one. They are compatible, tho

thomas-lebeau: > But then we could also be discarding RUM events beyond the initial changes that we thought.

😞 That's really annoying. Well, keep it like this then

packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -128,12 +130,16 @@ export function startTelemetryCollection(
...
-  ): TelemetryEvent & Context {
+  ): (TelemetryEvent & Context) | DISCARDED_TYPE {
     const clockNow = clocksNow()
     const defaultTelemetryEventAttributes = hooks.triggerHook(HookNames.AssembleTelemetry, {
       startTime: clockNow.relative,
     })
 
+    if (defaultTelemetryEventAttributes === DISCARDED) {
+      return DISCARDED
+    }
```
bcaudan: **💬 suggestion:** ‏what about extracting the hook execution above?

it would allow to control the execution in `onRawTelemetryEventCollected` and only building the event inside `toTelemetryEvent`, and avoid to propagate the `DISCARDED_TYPE`

```ts
    if (
      telemetryEnabledPerType[rawEvent.type!] &&
      alreadySentEvents.size < configuration.maxTelemetryEventsPerPage &&
      !alreadySentEvents.has(stringifiedEvent)
    ) {
      const defaultTelemetryEventAttributes = hooks.triggerHook(HookNames.AssembleTelemetry, {
        startTime: clocksNow().relative,
      })

      if (defaultTelemetryEventAttributes === DISCARDED) {
        return
      }
      const event = toTelemetryEvent(defaultTelemetryEventAttributes, telemetryService, rawEvent, runtimeEnvInfo)
      observable.notify(event)
      sendToExtension('telemetry', event)
      alreadySentEvents.add(stringifiedEvent)
    }
```

bcaudan:
**💬 suggestion:** ‏about the PR title, it would be nice to be more explicit about what is actually fixed

packages/core/src/domain/telemetry/telemetry.ts
mormubis: Updated!

PR #3682, opened 2025-07-09
✨ [RUM-8313] Remove URL polyfill and retrieve native URL constructor

rgaignault:
## Motivation

Remove IE11 fallback now we don't support it anymore and prevent customer polyfill issues by using the native constructor from Iframe

## Changes
 
- Remove IE fallback
- New getNativeURL function that allow to get the URL constructor from the iframe.
- Added unit test for getNative and cache mechanism

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -18,40 +18,43 @@ export function getPathName(url: string) {
...
+ * This avoids polyfill issues by getting the native implementation from a fresh iframe context
+ * Falls back to the original URL constructor if iframe approach fails
+ */
+let cachedNativeURL: typeof URL | undefined
+export function getNativeURL(): typeof URL | undefined {
+  if (cachedNativeURL !== undefined) {
+    return cachedNativeURL
+  }
+  try {
+    const iframe = document.createElement('iframe')
```
sethfowler-datadog: Two nits here:
1. It'd be a good idea to set `iframe.style.display = 'none'`, so that the iframe will not be rendered if anything goes wrong. (Right now, if something goes wrong, the iframe would be quite visible!)
2. It'd be a good idea to remove the iframe from `document.body` in a `finally` block, so that it's guaranteed to happen even if an exception is thrown.

So, to start with, you might want to make these changes:

```suggestion
  }
  let iframe: HTMLIFrameElement
  try {
    iframe = document.createElement('iframe')
    iframe.style.display = 'none'
```

packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -18,40 +18,43 @@ export function getPathName(url: string) {
...
+      const testURL = new iframeURL('http://test.com')
+      if (testURL.href === 'http://test.com/') {
+        cachedNativeURL = iframeURL
+      }
     }
+
+    document.body.removeChild(iframe)
+  } catch {
+    // If iframe approach fails, we'll use the original URL constructor
   }
```
sethfowler-datadog: ...then, down here, finish by doing something like this:
```suggestion
  } catch {
    // If iframe approach fails, we'll use the original URL constructor
  } finally {
    if (iframe) {
        document.body.removeChild(iframe)
    }
  }
```

packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -18,40 +18,43 @@ export function getPathName(url: string) {
...
-      isURLSupported = url.href === 'http://test/path'
-    } catch {
-      isURLSupported = false
+/**
+ * Get native URL constructor from a clean iframe
+ * This avoids polyfill issues by getting the native implementation from a fresh iframe context
+ * Falls back to the original URL constructor if iframe approach fails
+ */
+let cachedNativeURL: typeof URL | undefined
+export function getNativeURL(): typeof URL | undefined {
```
sethfowler-datadog: The comment says "falls back to the original URL constructor if iframe approach fails", but actually, the caller handles that. I think it'd actually be better to have this function handle it, as the comment says, and further more, to cache the fallback value. If one attempt fails, it's unlikely that future attempts will succeed, so we'll always end up using the original URL constructor, but with the current approach, we'll create a new iframe every single time this function is called before ultimately falling back. That's quite expensive!

So, tldr: I'd just set `cachedNativeURL = URL` in the `catch` block in this function, and change the return type to `typeof URL`.

packages/core/src/tools/utils/urlPolyfill.spec.ts
```diff
@@ -79,3 +79,51 @@ describe('buildUrl', () => {
...
+    const originalURL = window.URL
+    ;(window as any).URL = function badURL() {
+      throw new Error('Bad polyfill')
+    }
+    const nativeURL = getNativeURL()
+    if (nativeURL) {
+      expect(typeof nativeURL).toBe('function')
+      const url = new nativeURL('http://example.com')
+      expect(url.href).toBe('http://example.com/')
+    }
```
thomas-lebeau: **❓ question:** ‏‏ If `nativeURL` return undefined, the test will pass too, no?

rgaignault: good catch 👌 I put a bunch of additional defined check. Let me know what you think

packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -18,40 +18,52 @@ export function getPathName(url: string) {
...
+    document.body.appendChild(iframe)
+
+    const iframeWindow = iframe.contentWindow
+    if (iframeWindow && (iframeWindow as any).URL) {
+      const iframeURL = (iframeWindow as any).URL as typeof URL
+      const testURL = new iframeURL('http://test.com')
+      if (testURL.href === 'http://test.com/') {
+        cachedNativeURL = iframeURL
+      }
+    }
```
BenoitZugmeyer: **suggestion:** I think it's safe to assume the URL constructor is correct here, so remove this check.

packages/core/src/tools/utils/urlPolyfill.ts
```diff
@@ -18,40 +18,52 @@ export function getPathName(url: string) {
...
-      isURLSupported = url.href === 'http://test/path'
-    } catch {
-      isURLSupported = false
+/**
+ * Get native URL constructor from a clean iframe
+ * This avoids polyfill issues by getting the native implementation from a fresh iframe context
+ * Falls back to the original URL constructor if iframe approach fails
+ */
+let cachedNativeURL: typeof URL | undefined
+export function getNativeURL(): typeof URL {
```
BenoitZugmeyer: **suggestion:** To encourage other usages, could we have a function like:

```ts
let getPristineWindowCache: Window & typeof globalThis | undefined

function getPristineWindow() {
  if (!getPristineGlobalObjectCache) {
    let iframe: HTMLIFrameElement | undefined
    try {
      iframe = document.createElement('iframe')
      iframe.style.display = 'none'
      document.body.appendChild(iframe)
      getPristineGlobalObjectCache = iframe.contentWindow
    } catch {
      getPristineGlobalObjectCache = window
    }
    iframe?.remove()
  }

  return getPristineGlobalObjectCache
}
```

and then:

```ts
function buildURL() {
  const { URL } = getPristineWindow()
  ...
}

```

thomas-lebeau (approved):

PR #3685, opened 2025-07-10
♻️ [RUM-10900] Introduce assemble telemetry hook

amortemousque:
## Motivation

Introduce assemble telemetry hook and use it to colocate telemetry context within its relevant module.
Used it to add user, action, session, user, application and view view telemetry context.

## Changes

- Remove telemetry context provider and its usage from startRum and startLogs
- Introduce a new assemble telemetry hook 
- Used the hook in RUMInternalContext, defaultContext, sessionContext, userContext, actionCollection, viewCollection

## Checklist

- [x] Tested locally
- [x] Tested on staging


packages/core/src/tools/abstractHooks.ts
```diff
@@ -59,7 +64,8 @@ export function abstractHooks<T extends { [K in HookNames]: (...args: any[]) =>
         results.push(result)
       }
 
-      return combine(...(results as unknown as [object, object])) as E
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
amortemousque: The inferred return type from `triggerHook` looks good, but I couldn’t make it return safe. Open to suggestions :)

packages/logs/src/domain/contexts/rumInternalContext.ts
```diff
@@ -31,6 +31,20 @@ export function startRUMInternalContext(hooks: Hooks) {
...
+    }
+
+    return {
+      application: internalContext.application_id ? { id: internalContext.application_id as string } : undefined,
+      view: internalContext.view ? { id: (internalContext.view as Context).id as string } : undefined,
+      action: internalContext.user_action ? { id: (internalContext.user_action as Context).id as string } : undefined,
+    }
+  })
+
   function getRUMInternalContext(startTime?: RelativeTime) {
```
thomas-lebeau, line 48: **💬 suggestion:** ‏ I think we can remove `getRUMInternalContext` from the returned object Line 73

thomas-lebeau (approved):

packages/logs/src/domain/contexts/rumInternalContext.ts
amortemousque: Oh yes nice :)

packages/core/src/tools/abstractHooks.ts
BenoitZugmeyer: Looks like it's because `Exclude<ReturnType<T[K]>, SKIPPED>` is resolved to `any`. Since it's not actually enforcing anything, we might want to simplify the types within the function (ex: you could use `const results: any[]`)

packages/logs/src/domain/contexts/rumInternalContext.ts
```diff
@@ -31,6 +31,20 @@ export function startRUMInternalContext(hooks: Hooks) {
...
+
+    if (!internalContext) {
+      return SKIPPED
+    }
+
+    return {
+      application: internalContext.application_id ? { id: internalContext.application_id as string } : undefined,
+      view: internalContext.view ? { id: (internalContext.view as Context).id as string } : undefined,
+      action: internalContext.user_action ? { id: (internalContext.user_action as Context).id as string } : undefined,
+    }
```
BenoitZugmeyer: **nitpick:** all those casts are a bit annoying to read. Maybe you could cast `internalContext` above instead, like:

```ts
const internalContext = getRUMInternalContext(startTime) as  { ... } | undefined

```

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -55,6 +55,15 @@ export function startActionCollection(
     }
   })
 
+  hooks.register(HookNames.AssembleTelemetry, ({ startTime }): DefaultTelemetryEventAttributes | SKIPPED => {
+    const actionId = actionContexts.findActionId(startTime)
+    if (!actionId) {
+      return SKIPPED
+    }
+
+    return { action: { id: actionId as string } }
```
BenoitZugmeyer: **suggestion:** `actionId` can be an array of strings... maybe remove the cast?

BenoitZugmeyer (commented):
💭 Thoughts: it's a bit annoying that small one-liners are now 10 lines of code, ex:

```ts
  telemetry.setContextProvider('view.id', () => viewHistory.findView()?.id)
```

```ts
  hooks.register(HookNames.AssembleTelemetry, ({ startTime }): DefaultTelemetryEventAttributes | SKIPPED => {
    const view = viewHistory.findView(startTime)

    if (!view) {
      return SKIPPED
    }

    return {
      view: {
        id: view.id,
      },
    }
  })
```

From a bundle size perspective, it's probably not a huge deal, but we are missing a bit of conciseness.

I guess the hook can be rewritten as:

```ts
  hooks.register(HookNames.AssembleTelemetry, ({ startTime }): DefaultTelemetryEventAttributes => ({
      view: {
        id: viewHistory.findView(startTime)?.id,
      }
  })
```

which would be functionally equivalent but we probably need to adjust the types.

packages/rum-core/src/domain/action/actionCollection.ts
amortemousque: The action.id format from the telemetry is wrong. It only accepts a string. I should fix the rum-event-format

packages/core/src/tools/abstractHooks.ts
amortemousque: Exclude<ReturnType<T[K]>, SKIPPED> is resolved to any in the context of the abstractHook  because of `[K in HookNames]: (...args: any[]) => any }` I changed it to unknown and simplified internal types

packages/logs/src/domain/contexts/rumInternalContext.ts
amortemousque: I put the `RumInternalContext` type in `core` to share it with Logs and have an explicit taype. Let me know what you think.

PR #3686, opened 2025-07-15
♻️ Enforce using interface over type

amortemousque:
## Motivation

To ensure consistent API reference documentation, this PR reinforces our convention of using interface instead of type for object-like definitions.

## Changes

Replace type alias with interface for object definition

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/startCustomerDataTelemetry.ts
+++ b/packages/rum-core/src/domain/startCustomerDataTelemetry.ts
@@ -6,13 +6,13 @@ import { LifeCycleEventType } from './lifeCycle'
 
 export const MEASURES_PERIOD_DURATION = 10 * ONE_SECOND
 
-type Measure = {
+interface Measure {
   min: number
   max: number
   sum: number
 }
 
-type CurrentPeriodMeasures = {
+interface CurrentPeriodMeasures {
   batchCount: number
   batchBytesCount: Measure
   batchMessagesCount: Measure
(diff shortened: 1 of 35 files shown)


packages/rum-core/src/domain/startCustomerDataTelemetry.ts
```diff
@@ -1,18 +1,18 @@
-import type { FlushEvent, Observable, Telemetry } from '@datadog/browser-core'
+import type { Context, FlushEvent, Observable, Telemetry } from '@datadog/browser-core'
 import { performDraw, ONE_SECOND, addTelemetryDebug, setInterval } from '@datadog/browser-core'
 import type { RumConfiguration } from './configuration'
 import type { LifeCycle } from './lifeCycle'
 import { LifeCycleEventType } from './lifeCycle'
 
 export const MEASURES_PERIOD_DURATION = 10 * ONE_SECOND
 
-type Measure = {
```
mormubis, line 9: **❓ question:** ‏ Why do we need to extend from `Context`?

mormubis: I meant here https://github.com/DataDog/browser-sdk/pull/3686/files#diff-d9f26438eb193efa6ca1d2a3c7cf7f794686dc384f26b860322ed642c1ed2dbeR9

amortemousque: `CurrentPeriodMeasures` is assigned to a context, but its type is missing the `[x: string]: ContextValue` index signature, which causes a type-checking error. The same issue applies to the `Measure` attribute inside `CurrentPeriodMeasures`.

mormubis (approved):

PR #3687, opened 2025-07-15
✨[RUM-10825] Add hot-reload to developer-extension

BeltranBulbarellaDD:
## Motivation

Adding hot-reload on Panel so that all UI components (React Components, CSS and JS) changes get built automatically and allow for a better developer experience.

## Changes

- Added named configurations in Webpack to enable targeting specific configs.
- Added conditional dev server setup on port 3001.
- Added new build script: `yarn dev:panel` which builds all extension parts + starts HMR dev server for panel.
- Added panel entry point.
- Created HTML template that conditionally loads scripts based on environment.
- Created development-specific manifest with Content Security Policy allowing `localhost:3001`.

Changes after review:
- I decided to remove the `panel.html.template` and `manifest-dev.json` and inject them in the webpack configuration

## Test instructions

Use the extension running `yarn dev:panel` make a change and it should be seen automatically in the UI.

https://github.com/user-attachments/assets/edaa807c-2ce8-4048-b10a-5da1baed26fa

## Checklist

- [X ] Tested locally
- [ ] Tested on staging


developer-extension/package.json
```diff
@@ -4,18 +4,26 @@
   "private": true,
   "scripts": {
     "build": "rm -rf dist && webpack --mode production",
-    "dev": "webpack --mode development --watch"
+    "dev": "webpack --mode development --watch",
+    "dev:panel": "rm -rf dist && webpack --mode development && webpack serve --mode development --config-name=panel --hot"
```
sethfowler-datadog: Should `dev:panel` just replace `dev`? Seems like it's basically always better; any downsides to it?

developer-extension/webpack.config.js
```diff
@@ -1,51 +1,133 @@
...
+      entry: './src/panel/index.tsx',
+      devServer: isDevelopment
+        ? {
+            hot: true,
+            port: 3001,
+            static: false,
+            headers: {
+              'Access-Control-Allow-Origin': '*',
+            },
+            allowedHosts: 'all',
```
sethfowler-datadog: Nit: is it possible to use something more restrictive, like `auto`?

sethfowler-datadog (approved):
Looks great; this seems like a really nice usability improvement!

Some nits below:

developer-extension/package.json
BeltranBulbarellaDD: I agree! I didn't want to change current behaviour just in case but I think its better to have just 1

developer-extension/webpack.config.js
BeltranBulbarellaDD: Would `localhost` be better? I'm afraid it might be too restrictive for what we want to do.

BeltranBulbarellaDD:
@Cursor review

BeltranBulbarellaDD:
@Cursor review

developer-extension/webpack.config.js
```diff
@@ -1,51 +1,133 @@
...
+            <body>
+              <div id="root"></div>
+              ${
+                isDevelopment
+                  ? "<script src='http://localhost:3001/panel.js'></script>"
+                  : "<script src='panel.js'></script>"
+              }
+            </body>
+            </html>
+          `,
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Your comment is misleading as there are no changes on your code.

PR #3688, opened 2025-07-15
[PANA-3411] ⚗ Add telemetry for session replay reliability

sethfowler-datadog:
## Motivation

We'd like to better understand the factors that affect reliable delivery of session replay segments to the intake. This will allow us to identify the best opportunities for improving the reliability of session replay going forward.

## Changes

This PR adds telemetry for network requests associated with session replay segments. In particular, when telemetry is enabled and we're in a sampled session, we generate new telemetry events for:
- Network requests for session replay segments that fail.
- Network requests for session replay segments that are dropped because the request queue is full.
- Network requests for all full snapshot session replay segments, failing or not.

These events several types of information:
* The result of the request, corresponding to the reasons above. (`failure`, `queue-full`, or `success`)
* Whether the associated segment contains a full snapshot or not.
* Measures of the segment's size (number of records, compressed size, uncompressed size).
* Measures of the amount of `_cssText` data in the segment (count, max, sum of `_cssText` lengths).
* Measures of the amount of requests in flight (byte count, request count).

The hope is that this information will help us clarify to what extent each of these variables is associated with network request failures, and that this will help clarify where we should make improvements to achieve better reliability.

To collect all this information, this PR contains a fair quantity of wiring. The main moving pieces are:
* A new `Observable` exposed by `HttpRequest`, used to collect information about `HttpRequestEvent` events related to the actual network requests we use to send the segments.
* `HttpRequest` is now generic in the type of its payload, allowing us to attach additional metadata to each payload. The payload is exposed through the `HttpRequestEvent` events so that this metadata can be associated with each network request.
* The DOM serialization code now passes around a new `SerializationStats` type on the `SerializationContext`; this type is used to record statistics about `_cssText` data.
* When a session replay mutation record is emitted, we now wire its associated `SerializationStats` value along with it, so that it can be included in the `ReplayPayload` for the resulting segment.
* The new `startSegmentTelemetry()` function handles observes `HttpRequestEvents`, enriched with payloads that include `SerializationStats` and other metadata, and generates the telemetry events we ultimately want to produce.

## Checklist

- [x] Tested locally
- [ ] Tested on staging *(I plan to tomorrow.)*
- [x] Added unit tests for this change. *(I plan to add more.)*


packages/rum/src/domain/segmentCollection/startSegmentTelemetry.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    ;({ stop: stopSegmentTelemetry } = startSegmentTelemetry(
+      configuration,
+      { enabled: true } as Telemetry,
+      requestObservable
+    ))
+  }
+
+  afterEach(() => {
+    stopSegmentTelemetry?.()
+  })
```
BenoitZugmeyer: **suggestion:** 
```suggestion
  function setupSegmentTelemetryCollection(partialConfig: Partial<RumConfiguration> = config) {
    const configuration = mockRumConfiguration(partialConfig)
    requestObservable = new Observable()
    telemetryEvents = startFakeTelemetry()
    const { stop: stopSegmentTelemetry } = startSegmentTelemetry(
      configuration,
      { enabled: true } as Telemetry,
      requestObservable
    )
    registerCleanupCallback(stopSegmentTelemetry)
  }
```

sethfowler-datadog: I'll make that change!

packages/rum/src/domain/segmentCollection/startSegmentTelemetry.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const { unsubscribe } = requestObservable.subscribe((event) => {
+    const payload = event.payload
+    if (
+      event.type === 'failure' ||
+      event.type === 'queue-full' ||
+      (event.type === 'success' && payload.isFullSnapshot)
+    ) {
+      const metrics = createSegmentMetrics(event.type, event.bandwidth, payload)
+      addTelemetryDebug(SEGMENT_METRICS_TELEMETRY_NAME, metrics)
```
bcaudan: **💬 suggestion:** it could be handy to regroup those metrics in a single namespace to not have them scattered across the other properties of the telemetry event:
‏
Screenshot 2025-07-29 at 09 27 27

```suggestion
      addTelemetryDebug(SEGMENT_METRICS_TELEMETRY_NAME, { metrics })
```

packages/core/src/transport/sendWithRetryStrategy.ts
```diff
@@ -24,76 +25,85 @@ const enum RetryReason {
...
   if (
     state.transportStatus === TransportStatus.UP &&
     state.queuedPayloads.size() === 0 &&
     state.bandwidthMonitor.canHandle(payload)
   ) {
-    send(payload, state, sendStrategy, {
-      onSuccess: () => retryQueuedPayloads(RetryReason.AFTER_SUCCESS, state, sendStrategy, trackType, reportError),
+    send(payload, state, sendStrategy, observable, {
+      onSuccess: () =>
+        retryQueuedPayloads(RetryReason.AFTER_SUCCESS, state, sendStrategy, trackType, reportError, observable),
```
bcaudan: **🥜 nitpick:** ‏it could be nice to have a more meaningful name for this observable across the file

packages/rum/src/domain/segmentCollection/segmentCollection.ts
```diff
@@ -39,14 +41,19 @@ export let SEGMENT_BYTES_LIMIT = 60_000
 // To help investigate session replays issues, each segment is created with a "creation reason",
 // indicating why the session has been created.
 
+export interface SegmentCollector {
+  addRecord(this: void, record: BrowserRecord, stats?: SerializationStats): void
+  stop(this: void): void
+}
```
bcaudan: **🥜 nitpick:** ‏it does not seem to be used outside of this file
```suggestion
interface SegmentCollector {
  addRecord(this: void, record: BrowserRecord, stats?: SerializationStats): void
  stop(this: void): void
}
```

packages/rum/src/domain/segmentCollection/startSegmentTelemetry.ts
```diff
@@ -0,0 +1,68 @@
...
+  configuration: RumConfiguration,
+  telemetry: Telemetry,
+  requestObservable: Observable<HttpRequestEvent<ReplayPayload>>
+) {
+  const segmentTelemetryEnabled = telemetry.enabled && performDraw(configuration.segmentTelemetrySampleRate)
+  if (!segmentTelemetryEnabled) {
+    return { stop: noop }
+  }
+
+  const { unsubscribe } = requestObservable.subscribe((event) => {
```
bcaudan: **💬 suggestion:** ‏it could be interesting to have a consistent naming between `request`, `event` and `requestEvent`

bcaudan (commented):
**❓ question:** ‏IIRC, you had some concerns on `maxTelemetryEventsPerPage: 15` with this change, where did we land on that?

packages/rum/src/domain/segmentCollection/startSegmentTelemetry.ts
sethfowler-datadog: Done! I also updated this line to use `addTelemetryMetrics()`, added in #3692, since that PR has now landed.

sethfowler-datadog: I went further and grouped some of the internal fields into sub-objects, so hopefully things are generally a bit better namespaced now.

bcaudan (approved):
Nice!

PR #3691, opened 2025-07-17
🐛 Allow service and version to be modifiable for long_task and vital events

thomas-lebeau:
## Motivation

I'm not sure how we forgot these in the first place but there is no reason to not have them modifiable.

## Changes

add `service` and `version` to the list of modifiable fields for vitals and long_task events

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/assembly.spec.ts
+++ b/packages/rum-core/src/domain/assembly.spec.ts
@@ -412,13 +412,6 @@ describe('rum assembly', () => {
         })
         expect((serverRumEvents[0] as RumResourceEvent).service).toBe('bar')
         expect((serverRumEvents[0] as RumResourceEvent).version).toBe('0.2.0')
-
-        notifyRawRumEvent(lifeCycle, {
-          rawRumEvent: createRawRumEvent(RumEventType.VIEW),
-        })
-        expect((serverRumEvents[1] as RumViewEvent).service).toBe('bar')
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -29,7 +29,7 @@ import type { RumSessionManager } from './rumSessionManager'
 import { startSessionContext } from './contexts/sessionContext'
 import { createHooks } from './hooks'
 
-describe('rum assembly', () => {
+fdescribe('rum assembly', () => {
```
bcaudan: **❗ warning:** ‏
```suggestion
describe('rum assembly', () => {
```

packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -408,7 +408,7 @@ describe('rum assembly', () => {
         })
 
         notifyRawRumEvent(lifeCycle, {
-          rawRumEvent: createRawRumEvent(RumEventType.RESOURCE),
+          rawRumEvent: createRawRumEvent(eventType),
```
bcaudan, line 411: **💬 suggestion:** ‏should we remove the assertions on view for this test then?

bcaudan (approved):
Nice!

sethfowler-datadog (approved):
LGTM! Thank you!

PR #3694, opened 2025-07-17
♻️ [Remote config] use types from JSON schema

bcaudan:
## Motivation

Prepare the introduction of new config options

## Changes

- Add remote config `rum-sdk-config.json` schema, duplicated here for now let's see if we feel the need to have better sharing in the future
- Generate remote config types from the schema
- Rename schema commands: `rum-events-format:[sync|generate]` to `json-schemas:[sync|generate]`
- Apply only supported remote config field in the init config

## Test instructions

- Get a `remoteConfigurationId` from an auto-instrumented app
- Use it in an RUM init
- Remote config values should still be used

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -14,21 +26,34 @@ export function fetchAndApplyRemoteConfiguration(
...
+  // intents:
+  // - explicitly set each supported field to limit risk in case an attacker can create configurations
+  // - check the existence in the remote config to avoid clearing a provided init field
+  const appliedConfiguration = { ...initConfiguration } as RumInitConfiguration & { [key: string]: unknown }
+  STATIC_OPTIONS.forEach((option: string) => {
+    if (option in rumRemoteConfiguration) {
+      appliedConfiguration[option] = rumRemoteConfiguration[option]
+    }
+  })
+  return appliedConfiguration
```
bcaudan, line 40: 📝 **note**: to avoid to have one `if` by option, I needed to have index signatures for the config types. 
I choose to add extra casts only here but let me know if you have other ideas.

BenoitZugmeyer: Looks good!

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -14,21 +26,34 @@ export function fetchAndApplyRemoteConfiguration(
...
   addEventListener(configuration, xhr, 'load', function () {
     if (xhr.status === 200) {
-      const remoteConfiguration = JSON.parse(xhr.responseText)
-      callback(remoteConfiguration.rum)
+      const remoteConfiguration = JSON.parse(xhr.responseText) as RumSdkConfig
+      if (remoteConfiguration.rum) {
+        callback(remoteConfiguration.rum)
+      } else {
+        display.error('No remote configuration for RUM.')
+      }
```
BenoitZugmeyer, line 56: **thought:** we could probably use `fetch()... json()` instead in the future.

remote-config/rum-sdk-config.json
BenoitZugmeyer, line 1: **question:** since the schema isn't used at runtime (ex: via `ajv`) or shared anywhere else, what is the advantage of using a JSON schema instead of a handcrafted TS type?

amortemousque: I guess, the value is that the original schema is in JSON, so copying the entire JSON schema is faster and less error-prone than manually converting it to TypeScript. 
That said, I agree it’s not ideal. We should at least mention that the original RC schema lives elsewhere else to avoid confusion.

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
bcaudan: Yep, I want to do that in a future PR, probably at the same time as adding an e2e test. 
It seemed to have enough changes for one PR 😉

amortemousque (approved):

remote-config/rum-sdk-config.json
bcaudan: added a note with https://github.com/DataDog/browser-sdk/pull/3694/commits/c8ac3ad37837f032207065c4648ddbf00035b536

BenoitZugmeyer (approved):

PR #3721, opened 2025-07-23
✅ [Remote config] add a remote config e2e scenario

bcaudan:
## Motivation

Allow to add e2e tests for future complex cases

## Changes

- introduce internal `remoteConfigurationProxy` configuration parameter
- update e2e test framework to provide remote configuration

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -78,11 +78,21 @@ export interface RumInitConfiguration extends InitConfiguration {
...
    * @internal
    */
   remoteConfigurationId?: string | undefined
 
+  /**
+   * [Internal option] set a proxy URL for the remote configuration
+   *
+   * @internal
+   */
+  remoteConfigurationProxy?: string | undefined
```
bcaudan, line 94: **💭 thought:** ‏About adding this new config parameter, I've flagged it as internal for now but I think we should:

- promote it to public and add it to the [proxy guide](https://docs.datadoghq.com/real_user_monitoring/guide/proxy-rum-data/) with the GA of auto injection, since the remote config endpoint will likely be targeted by privacy blockers
- not allow to set it through the remote config to avoid CSRF

test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -2,13 +2,14 @@ import type { ServerResponse } from 'http'
...
 import express from 'express'
+import type { RumSdkConfig } from '@datadog/browser-rum-core'
 import { getSdkBundlePath, getTestAppBundlePath } from '../sdkBuilds'
 import type { MockServerApp, Servers } from '../httpServers'
 import { DEV_SERVER_BASE_URL } from '../../helpers/playwright'
 
 export const LARGE_RESPONSE_MIN_BYTE_SIZE = 100_000
 
-export function createMockServerApp(servers: Servers, setup: string): MockServerApp {
+export function createMockServerApp(servers: Servers, setup: string, remoteConfig?: RumSdkConfig): MockServerApp {
```
bcaudan: **💭 thought:** ‏I initially went with adding a new server for the remote config but it felt overkill for just this simple endpoint, so I added it to the mock server.

amortemousque (approved):

packages/rum-core/src/domain/configuration/configuration.ts
BenoitZugmeyer: I think this new option should behave like the `proxy` option (with `ddforward` and allow passing a function), to avoid customers to implement different kind of proxies.

test/e2e/lib/framework/createTest.ts
```diff
@@ -69,6 +69,7 @@ class TestBuilder {
   private rumConfiguration: RumInitConfiguration | undefined = undefined
   private alsoRunWithRumSlim = false
   private logsConfiguration: LogsInitConfiguration | undefined = undefined
+  private remoteConfig?: RumSdkConfig = undefined
```
BenoitZugmeyer: **💭 thought:** ‏seeing `LogsInitConfiguration` and `RumSdkConfig` side by side feels like we missed an opportunity to use a more aligned and explicit naming.
* `Config` should be `Configuration`, as other configuration types
* `RumSdk` should be `Rum`. We never call the RUM Browser SDK the "RUM SDK".
* It should probably include `Remote` somewhere

packages/rum-core/src/domain/configuration/configuration.ts
bcaudan: Yeah, I think so too, my plan was to handle it properly later down the road but it could be done now.

test/e2e/lib/framework/createTest.ts
bcaudan: Yep, I used the generated type names but I can remap that

packages/rum-core/src/domain/configuration/configuration.ts
BenoitZugmeyer: Mmh maybe that's fine, assuming people don't use it until RC is publicly available

bcaudan: I'll add a task to not forget it then

test/e2e/lib/framework/createTest.ts
bcaudan: added [♻️consistent remote configuration naming](https://github.com/DataDog/browser-sdk/pull/3721/commits/b9cd40c6b28e285b475c44cfc3726675a0fe8956)

PR #3723, opened 2025-07-24
♻️ use fetch to retrieve remote configuration

bcaudan:
## Motivation

With IE drop in last major, we can now use fetch instead of xhr

## Changes

- use fetch to retrieve remote configuration
- switch to promise instead of callback

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/configuration/remoteConfiguration.ts
+++ b/packages/rum-core/src/domain/configuration/remoteConfiguration.ts
@@ -1,4 +1,4 @@
-import { display, addEventListener, buildEndpointHost } from '@datadog/browser-core'
+import { display, buildEndpointHost } from '@datadog/browser-core'
 import type { RumInitConfiguration } from './configuration'
 import type { RumSdkConfig } from './remoteConfiguration.types'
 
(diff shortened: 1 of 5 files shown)


BeltranBulbarellaDD:
Hey, using fetch instead of XHR is something that would help for my Service Workers implementation. Are we planning on doing this in only some functions of the SDK?

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -41,31 +43,22 @@ export function applyRemoteConfiguration(
...
+  const FETCHING_ERROR_MESSAGE = 'Error fetching the remote configuration.'
+  let response: Response
+  try {
+    response = await fetch(buildEndpoint(configuration))
+  } catch {
+    throw new Error(FETCHING_ERROR_MESSAGE)
+  }
+  if (!response.ok) {
+    throw new Error(FETCHING_ERROR_MESSAGE)
+  }
```
BenoitZugmeyer: nitpick: might be a bit shorter like this:
```ts
  let response: Response | undefined
  try {
    response = await fetch(buildEndpoint(configuration))
  } catch {}
  if (!response || !response.ok) {
    throw new Error('Error fetching the remote configuration.')
  }
```

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -41,31 +43,22 @@ export function applyRemoteConfiguration(
...
+  const remoteConfiguration: RemoteConfiguration = await response.json()
+  if (remoteConfiguration.rum) {
+    return remoteConfiguration.rum
+  }
+  throw new Error('No remote configuration for RUM.')
```
BenoitZugmeyer: **Suggestion:** We probably should avoid using exceptions for control flow because TypeScript hides them, ex: by looking at `fetchRemoteConfiguration` signature I would assume that I'd be able to use it without worrying about exceptions (outside of unexpected exceptions to be collected by `monitor`).

So, I feel like using `display.error` + returning undefined to be better, or if we really want to make things explicit, maybe something like:

```ts
type RemoteConfigurationResult = {
  ok: true,
  value: RemoteConfiguration['rum'],
} | {
  ok: false,
  error: string
}
function fetchRemoteConfiguration(): Promise<RemoteConfigurationResult> { ... }
```

PR #3732, opened 2025-07-25
👷 🏷️ `scripts/` TS migration

BenoitZugmeyer:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


scripts/deploy/publish-developer-extension.ts
```diff
@@ -1,14 +1,13 @@
 // chrome-webstore-upload only support ESM syntax.
```
thomas-lebeau: **🥜 nitpick:** ‏ I think we can remove this comment now

thomas-lebeau (approved):
**🥜 nitpick:** ‏ There is a bunch of instance where instead of exporting the functions directly (i.e. `export function foo() { ...}`, we're using `export { foo }`. This could be more consistent.

eslint.config.mjs
BeltranBulbarellaDD, line 1: I think line 5 is not needed anymore
`// @ts-expect-error no types available`

PR #3738, opened 2025-07-28
✨ [FFL-450] Add support for exposures track

dd-oleksii:
## Motivation

Adding support for new EVP exposures track, so that feature flagging SDK can send exposure events.

## Changes

Adds new `exposures` track type to the core.

~The one difference from the rest of the tracks is that staging endpoint is `https://event-platform-intake.datad0g.com/api/v2/exposures` (production stays on `browser-intake-datadoghq.com`).~ Decided to use the same intake for now.

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -2,13 +2,13 @@ import type { Payload } from '../../transport'
 import { timeStampNow } from '../../tools/utils/timeUtils'
 import { normalizeUrl } from '../../tools/utils/urlPolyfill'
 import { generateUUID } from '../../tools/utils/stringUtils'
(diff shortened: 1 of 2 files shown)


packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -63,6 +63,10 @@ export function buildEndpointHost(
     return PCI_INTAKE_HOST_US1
   }
 
+  if (trackType === 'exposures' && site === INTAKE_SITE_STAGING) {
+    return `event-platform-intake.${site}`
+  }
```
dd-oleksii: cc @petzel to review this logic. i.e. staging on `event-platform-intake.datad0g.com` and production is on `browser-intake-{user-specified site}`

BenoitZugmeyer (approved):

PR #3740, opened 2025-07-28
✨⚗️ [RUM-10146] implement early request collection

BenoitZugmeyer:
## Motivation

Collect fetch and XHR requests that happened before the RUM SDK is initialized.

## Changes

Functional changes from this PR are behind the `early_request_collection` feature flag.

This PR starts by doing some tests changes in preparation to early request collection.

To be able to implement early request collection, we factorise the `processRequest` and `processPerformanceEntry` into a single function that takes both `request` and `performanceEntry` as argument. Then, we call this function with the expected arguments following the feature flag.

This PR also contains a bit of telemetry debug to validate the approach, in particular to see if we need a "bounded buffer" for the request registry, or a simple set is enough.

## Test instructions

* Enable the feature flag
* Make sure there is no significant drop in collected resources
* Make sure we collect early requests

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,13 +17,13 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@df559355d593797519d70b90fc8edd5db049e7a2 # v3.29.9
         with:
           languages: javascript
           config-file: .github/codeql-config.yml
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@df559355d593797519d70b90fc8edd5db049e7a2 # v3.29.9
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -25,12 +25,12 @@ jobs:
 
     steps:
       - name: Check out code at tag
-        uses: actions/checkout@v4
+        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
         with:
           fetch-depth: 0
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: '23'
           cache: 'yarn'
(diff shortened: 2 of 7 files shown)


packages/rum-core/src/domain/resource/resourceCollection.spec.ts
```diff
@@ -433,19 +483,37 @@ describe('resourceCollection', () => {
...
+    if (options.notifyPerformanceEntry !== false) {
+      notifyPerformanceEntries([
+        createPerformanceEntry(RumPerformanceEntryType.RESOURCE, {
+          initiatorType: requestCompleteEvent.type === RequestType.FETCH ? 'fetch' : 'xmlhttprequest',
+        }),
+      ])
+    }
+
+    runTasks()
   }
```
thomas-lebeau: **🥜 nitpick:** ‏ 
```suggestion
  function notifyRequest(options: { request?: Partial<RequestCompleteEvent>; notifyPerformanceEntry: boolean = true } = {}) {
    const requestCompleteEvent = {
      duration: 100 as Duration,
      method: 'GET',
      startClocks: { relative: 200 as RelativeTime, timeStamp: 123456789 as TimeStamp },
      status: 200,
      type: RequestType.XHR,
      url: 'https://resource.com/valid',
      handlingStack:
        'Error: \n  at <anonymous> @ http://localhost/foo.js:1:2\n    at <anonymous> @ http://localhost/vendor.js:1:2',
      ...options.request,
    } satisfies Partial<RequestCompleteEvent> as RequestCompleteEvent

    lifeCycle.notify(LifeCycleEventType.REQUEST_COMPLETED, requestCompleteEvent)

    if (options.notifyPerformanceEntry) {
      notifyPerformanceEntries([
        createPerformanceEntry(RumPerformanceEntryType.RESOURCE, {
          initiatorType: requestCompleteEvent.type === RequestType.FETCH ? 'fetch' : 'xmlhttprequest',
        }),
      ])
    }

    runTasks()
  }
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -175,9 +170,37 @@ function processResourceEntry(
...
+    return {
+      requestInput: request.input as RequestInfo,
+      requestInit: request.init,
+      response: request.response,
+      error: request.error,
+      ...baseDomainContext,
+    }
+  }
+  return {
+    performanceEntry: entry!,
```
thomas-lebeau, line 200: **💭 thought:** ‏ shouldn't `RumOtherResourceEventDomainContext` be defined with undefined instead?
```ts
export interface RumOtherResourceEventDomainContext {
  performanceEntry: PerformanceEntry | undefined
}
```

thomas-lebeau (approved):

packages/rum-core/src/domain/resource/resourceCollection.ts
BenoitZugmeyer: No, we know `entry` is always defined here (I added a comment to clarify). If it was potentially undefined, that would be a breaking change.

PR #3743, opened 2025-07-29
✨[RUM-10962][Remote config] support dynamic options

bcaudan:
## Motivation

Support remote configuration of new fields retrieving value from execution context:

- version
- user context
- global context

## Changes

- update remote config json + types (not yet validated on the backend side)
- support retrieving value from cookie or DOM element (javascript variable support in next PR)
- support setting user and global context from the remote configuration

## Test instructions

No backend to test the new options yet 

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -43,6 +61,14 @@ export function applyRemoteConfiguration(
       appliedConfiguration[option] = resolveConfigurationProperty(rumRemoteConfiguration[option])
     }
   })
+  SUPPORTED_CONTEXTS.forEach((context) => {
+    if (context in rumRemoteConfiguration && rumRemoteConfiguration[context] !== undefined) {
```
BenoitZugmeyer: *Nit*: this should be sufficient:

```suggestion
    if (rumRemoteConfiguration[context]) {
```

test/e2e/lib/framework/createTest.ts
```diff
@@ -194,7 +194,7 @@ function declareTestsForSetups(
         declareTest(name!, setupOptions, factory, runner)
       }
     })
-  } else {
+  } else if (setups.length === 1) {
```
BenoitZugmeyer, line 197: **question:** All tests should have at least one setup no? What's expected to happen if a test don't have any setup? Should we add a warning for this case?

packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
```diff
@@ -138,30 +159,124 @@ describe('remoteConfiguration', () => {
...
+      })
+
+      it('should resolve to undefined if the cookie is missing', () => {
+        deleteCookie(COOKIE_NAME)
+        expectAppliedRemoteConfigurationToBe(
+          { version: { rcSerializedType: 'dynamic', strategy: 'cookie', name: COOKIE_NAME } },
+          { version: undefined }
+        )
+      })
+    })
```
BenoitZugmeyer: **suggestion:** 

```ts
    describe('cookie strategy', () => {
      const COOKIE_NAME = 'unit_rc'

      it('should resolve a configuration value from a cookie', () => {
        setCookie(COOKIE_NAME, 'my-version', ONE_MINUTE)
        registerCleanupTask(() => deleteCookie(COOKIE_NAME))
        expectAppliedRemoteConfigurationToBe(
          { version: { rcSerializedType: 'dynamic', strategy: 'cookie', name: COOKIE_NAME } },
          { version: 'my-version' }
        )
      })

      it('should resolve to undefined if the cookie is missing', () => {
        expectAppliedRemoteConfigurationToBe(
          { version: { rcSerializedType: 'dynamic', strategy: 'cookie', name: COOKIE_NAME } },
          { version: undefined }
        )
      })
    })
```

packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
```diff
@@ -219,6 +219,88 @@ describe('remoteConfiguration', () => {
...
+
+        const span2 = document.createElement('span')
+        span2.id = 'version2'
+        span2.classList.add('version')
+        span2.setAttribute('data-version', 'version-456')
+        div.appendChild(span2)
+
+        registerCleanupTask(() => {
+          document.body.removeChild(div)
+        })
```
BenoitZugmeyer: **suggestion:** use the `appendElement` helper, it could be easier to read:

```ts
appendElement(`
<div>
  <span id="version1" class="version">version-123</span>
  <span id="version2" class="version" data-version="version-456"></span>
</div>
`)
```

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
BenoitZugmeyer: Also, why do we need `SUPPORTED_CONTEXTS`? Can we just iterate over the provided `supportedContextManagers` keys?

test/e2e/lib/framework/createTest.ts
bcaudan: To have the issue, I manually change the default setup from bundle to npm in pageSetup.ts
The tests declared for the rum-slim setup explicitly filter out npm setup, so for this case we end up with an empty setups array and crashing on `setups[0].factory`.
Allowing to locally tweak the setup without crashing seemed nice, any other concerns?

BenoitZugmeyer: I see. Well, I would still add a `console.warn` if there is no setups, that could help debugging issues in the future. But other than that it looks fine.

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -26,13 +26,13 @@ const SUPPORTED_FIELDS: Array<keyof RumInitConfiguration> = [
...
+  context = 'context',
+}
 
-interface SupportedContextManagers {
-  user: ReturnType<typeof createContextManager>
-  context: ReturnType<typeof createContextManager>
-  [key: string]: ReturnType<typeof createContextManager>
+type SupportedContextManagers = {
+  [key in SupportedContexts]: ReturnType<typeof createContextManager>
 }
```
BenoitZugmeyer: **suggestion:** Not sure why you need the enum here. Just use:
```ts
interface SupportedContextManagers {
  user: ReturnType<typeof createContextManager>
  context: ReturnType<typeof createContextManager>
}
```

and 

```ts
  contextConfiguration: RumRemoteConfiguration[keyof SupportedContextManagers] & { [key: string]: unknown }
```
no?

bcaudan: indeed!

BenoitZugmeyer (approved):

PR #3745, opened 2025-07-30
🔊 [RUM-11144] Apply telemetry context to early telemetry message

BenoitZugmeyer:
## Motivation

When sending a telemetry message before RUM or Logs are started, the telemetry event lacks the context (ex: session.id, application.id etc.) This seems to be a regression from https://github.com/DataDog/browser-sdk/pull/3602 , and blocks investigations like https://github.com/DataDog/browser-sdk/pull/3717  

## Changes

The main change is to make sure we wait a tiny bit before sending telemetry messages, so the context is available when assembling them. To achieve this, we use a `BufferedObserver`.

Other than that, many tests were adjusted to take this change into account.

## Test instructions

Using the sandbox, add a telemetry debug log (using `addTelemetryDebug`) before starting RUM and Logs (ex: for example in `preStartRum`). Make sure that message contains the session.id using the devtools extension.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


rgaignault (approved):

packages/core/test/emulate/mockTelemetry.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function startMockTelemetry() {
+  resetTelemetry()
+  const events: RawTelemetryEvent[] = []
+
+  const telemetryObservable = getTelemetryObservable()
+  telemetryObservable.subscribe(({ rawEvent }) => {
+    events.push(rawEvent)
+  })
```
sethfowler-datadog: Right now there's no mechanism to unsubscribe from this observable, so we leak these subscriptions and the event buffers they're filling. It might be nice to either return a `stop` function or take some sort of steps internally to ensure cleanup.

BenoitZugmeyer: Right, good call

PR #3751, opened 2025-08-01
Congyao/rum 10415 add privacy allowlist support treewalker

cy-moi:
## Motivation

Mask action name with allowlists generated from rum privacy build plugin(WIP). This approach is purely client side and allowlist-based. We aim to mask all action names (custom & auto) OOTB with build time configuration using a [build plugin.](https://github.com/DataDog/build-plugins)

This PR is a diverge from https://github.com/DataDog/browser-sdk/pull/3648 and replaces the `innerText` with dom tree walker in order to mask granular action names and fix potential leaking risks ([link](https://jsfiddle.net/t18wexrv/)) when using `innerText`.
## Changes
- Add allowlist processing helpers 
- Mask all action names with the allowlist when privacy build plugin is opt-in
- Stop using `innerText`

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,26 +179,33 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
   if ((element as HTMLElement).isContentEditable) {
     return
   }
 
+  const {
+    enablePrivacyForActionName,
+    actionNameAttribute: userProgrammaticAttribute,
+    defaultPrivacyLevel,
+  } = rumConfiguration
+  const enableAllowlistMask = isAllowlistMaskEnabled(defaultPrivacyLevel, getNodeSelfPrivacyLevel(element))
```
BenoitZugmeyer: **suggestion:** this could be done in the `getTextualContentWithTreeWalker` function

packages/rum-core/src/domain/action/privacy/maskWithAllowlist.ts
```diff
@@ -0,0 +1,31 @@
...
+export function maskDisallowedTextContent(text: string, fixedMask?: string): string {
+  if (!text.trim()) {
+    return text
+  }
+  // We are using toLocaleLowerCase when adding to the allowlist to avoid case sensitivity
+  // so we need to do the same here
+  if (window.$DD_ALLOW && window.$DD_ALLOW.has(text.toLocaleLowerCase())) {
+    return text
+  }
+  return fixedMask || text.replace(/\S/g, TEXT_MASKING_CHAR)
```
BenoitZugmeyer: **suggestion:** use the [`censorText`](https://github.com/DataDog/browser-sdk/blob/0e07e9a786beff30d85928d9244ca52e7cce4421/packages/rum-core/src/domain/privacy.ts#L190) function

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -141,9 +144,21 @@ function processPointerDown(
...
-    : NodePrivacyLevel.ALLOW
+  const nodeSelfPrivacy = getNodePrivacyLevel(pointerDownEvent.target, configuration.defaultPrivacyLevel)
+  // If the node privacy level is MASK_UNLESS_ALLOWLISTED, we use the default privacy level
+  // If the enablePrivacyForActionName is true, we use the node privacy level
+  // Otherwise, we use the allow level
+  // TODO: we should make enablePrivacyForActionName true by default for the next major version
+
+  // this won't work because there would always be a case when we set the nodeSelfPrivacy to ALLOW
+  // if we use configuration.defaultPrivacyLevel === NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED as a check
+  // then when we set defaultPrivacyLevel to ALLOW, and override the node privacy level to MASK_UNLESS_ALLOWLISTED
```
BenoitZugmeyer: **question:** I don't get this comment

> If the node privacy level is MASK_UNLESS_ALLOWLISTED, we use the default privacy level
> If the enablePrivacyForActionName is true, we use the node privacy level
> Otherwise, we use the allow level

That's exactly what's the code is doing. Could you explain why we do this instead?

> TODO: we should make enablePrivacyForActionName true by default for the next major version

Same here, why?

> this won't work because there would always be a case when we set the nodeSelfPrivacy to ALLOW
> if we use configuration.defaultPrivacyLevel === NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED as a check
> then when we set defaultPrivacyLevel to ALLOW, and override the node privacy level to MASK_UNLESS_ALLOWLISTED

I guess something is missing here?

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -141,9 +144,21 @@ function processPointerDown(
...
+
+  const nodePrivacyLevel = isAllowlistMaskEnabled(configuration.defaultPrivacyLevel, nodeSelfPrivacy)
+    ? nodeSelfPrivacy
+    : configuration.enablePrivacyForActionName
+      ? nodeSelfPrivacy
+      : NodePrivacyLevel.ALLOW
```
BenoitZugmeyer: **suggestion:** What about something like this:

```ts
const nodePrivacyLevel =
  nodeSelfPrivacy === PrivacyLevel.MASK_UNLESS_ALLOW_LISTED || configuration.enablePrivacyForActionName
    ? nodeSelfPrivacy
    : NodePrivacyLevel.ALLOW
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
BenoitZugmeyer: **suggestion:** use [`getNodePrivacyLevel`](https://github.com/DataDog/browser-sdk/blob/0e07e9a786beff30d85928d9244ca52e7cce4421/packages/rum-core/src/domain/privacy.ts#L48) instead, so that privacy level follows the privacy inheritance rules.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -940,6 +1020,19 @@ describe('serializeDocumentNode handles', function testAllowDomTree() {
     })
   })
 
+  describe('for privacy tag `mask-unless-allowlisted`, a DOM tree', function testMaskUnlessAllowlistedDomTree() {
+    it('is serialized correctly when no allowlist is provided', () => {
+      // Without allowlist, should behave like mask mode
+      window.$DD_ALLOW = undefined
+      registerCleanupTask(() => {
+        window.$DD_ALLOW = undefined
+      })
```
BenoitZugmeyer: **suggestion:** This should be unnecessary. `window.$DD_ALLOW` should be undefined at the beginning of the test (if it's not, then a cleanup is missing in another test), and since the test does not change it, it should stay undefined.

packages/rum-core/src/domain/action/privacy/maskWithAllowlist.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    type: ActionType.CLICK,
+    name: '',
+    nameSource: ActionNameSource.TEXT_CONTENT,
+    target: {
+      selector: 'button',
+      width: 100,
+      height: 100,
+    },
+    position: { x: 0, y: 0 },
+  }
```
BenoitZugmeyer: **suggestion:** it looks like you only use the `clickActionBase.name`. Maybe keep `ClickActionBase` private and just use a string variable instead of defining the whole structure?

packages/rum-core/src/domain/action/trackClickActions.ts
cy-moi: > TODO: we should make enablePrivacyForActionName true by default for the next major version

If we set this by default when the `configuration.defaultPrivacyLevel: MASK`, it would simplify things greatly and align with Session Replay.

The last paragraph was badly formed, I was explaining why we have to check for `isAllowlistMaskEnabled(configuration.defaultPrivacyLevel, nodeSelfPrivacy)`. When we set defaultPrivacyLevel to ALLOW, and then set the node class name to MASK_UNLESS_ALLOWLISTED, without checking the `nodeSelfPrivacy` level we will not be able to apply the allowlist, because we are ignoring the `nodePrivacyLevel` and setting it to ALLOW.

We need all these checks because we override the node privacy level to allow here. By making `MASK` level actually masks action names, we can get rid of the override and the checks for it. 

I updated the comments to reflect this. I hope it explains better this time.

packages/rum-core/src/domain/privacy.ts
```diff
@@ -306,6 +278,37 @@ export function shouldIgnoreElement(element: Element): boolean {
   return false
 }
 
-export function getPrivacySelector(privacyLevel: string) {
-  return `[${PRIVACY_ATTR_NAME}="${privacyLevel}"], .${PRIVACY_CLASS_PREFIX}${privacyLevel}`
+declare global {
+  interface Window {
+    $DD_ALLOW?: Set<string>
+  }
+}
```
BenoitZugmeyer: **suggestion:** don't use `declare global` as it will declare a global variable everywhere, including in our customers source code (i.e. `$DD_ALLOW` will be visible in their code editor autocompletion).

Instead, use something like [this](https://github.com/DataDog/browser-sdk/blob/main/packages/logs/src/domain/contexts/rumInternalContext.ts#L16-L23):

```ts
interface BrowserWindow {
    $DD_ALLOW?: Set<string>
}
// then
(window as BrowserWindow).$DD_ALLOW
```

(Not pretty, but we have some thoughts to improve this in the future

packages/rum-core/src/domain/privacyConstants.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export function getPrivacySelector(privacyLevel: string) {
+  return `[${PRIVACY_ATTR_NAME}="${privacyLevel}"], .${PRIVACY_CLASS_PREFIX}${privacyLevel}`
+}
+
+/**
+ * Text censoring non-destructively maintains whitespace characters in order to preserve text shape
+ * during replay.
+ */
+export const censorText = (text: string) => text.replace(/\S/g, TEXT_MASKING_CHAR)
```
BenoitZugmeyer: **nitpick:** maybe move those functions back to `privacy.ts` as they are not constants

BenoitZugmeyer (approved):

PR #3759, opened 2025-08-04
👷‍♀️ [RUM-11139] Improve textual content extraction for action names

cy-moi:
## Motivation
Improve text content extraction in`getActionNameFromElement`.

## Changes
- User treewalker to parse node by node instead of using innerText

## Test instructions
should pass ci

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
+++ b/packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
@@ -96,6 +96,14 @@ describe('getActionNameFromElement', () => {
     expect(nameSource).toBe('text_content')
   })
 
+  it('should correctly compute the name for a complex case', () => {
+    const element = appendElement(
+      '<div> <span> hello</span> \n <span>world</span> <div><span>!</span></div><br><span>another</span>-<span>one</span> '
+    )
+    const { name } = getActionNameFromElement(element, defaultConfiguration)
+    expect(name).toBe('hello world ! another-one')
+  })
+
   it('ignores the inline script textual content', () => {
     const { name, nameSource } = getActionNameFromElement(
       appendElement("<div><script>console.log('toto')</script>b</div>"),
@@ -418,13 +426,22 @@ describe('getActionNameFromElement', () => {
       expect(nameSource).toBe('custom_attribute')
     })
 
-    it('remove children with programmatic action name in textual content', () => {
+    it('removes children with programmatic action name in textual content', () => {
       const { name, nameSource } = getActionNameFromElement(
         appendElement('<div>Foo <div data-dd-action-name="custom action">bar<div></div>'),
         defaultConfiguration
       )
 
       expect(name).toBe('Foo')
+      expect(nameSource).toBe('text_content')
+    })
+
+    it('removes only the child with programmatic action name in textual content', () => {
+      const { name, nameSource } = getActionNameFromElement(
+        appendElement('<div>Foobar <div data-dd-action-name="custom action">bar<div></div>'),
+        defaultConfiguration
+      )
+      expect(name).toBe('Foobar')
       expect(nameSource).toBe('text_content')
     })
 
--- a/packages/rum-core/src/domain/action/getActionNameFromElement.ts
+++ b/packages/rum-core/src/domain/action/getActionNameFromElement.ts
@@ -204,8 +204,25 @@ function getActionNameFromTextualContent(
   }
 }
 
+function shouldExcludeNode(
+  node: Node,
+  userProgrammaticAttribute: string | undefined,
+  privacyEnabledActionName?: boolean
+) {
+  const exclusionSelectors = [`[${DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE}]`]
+  if (userProgrammaticAttribute) {
+    exclusionSelectors.push(`[${userProgrammaticAttribute}]`)
+  }
+  if (privacyEnabledActionName) {
+    exclusionSelectors.push(
+      `${getPrivacySelector(NodePrivacyLevel.HIDDEN)}, ${getPrivacySelector(NodePrivacyLevel.MASK)}`
+    )
+  }
+  return (node as HTMLElement).closest(exclusionSelectors.join(','))
+}
+
 function getTextualContent(
-  element: Element | HTMLElement,
+  element: Element,
   userProgrammaticAttribute: string | undefined,
   privacyEnabledActionName?: boolean
 ) {
@@ -213,38 +230,65 @@ function getTextualContent(
     return
   }
 
-  if ('innerText' in element) {
-    let text = element.innerText
-
-    const removeTextFromElements = (query: string) => {
-      const list = element.querySelectorAll<Element | HTMLElement>(query)
-      for (let index = 0; index < list.length; index += 1) {
-        const element = list[index]
-        if ('innerText' in element) {
-          const textToReplace = element.innerText
-          if (textToReplace && textToReplace.trim().length > 0) {
-            text = text.replace(textToReplace, '')
-          }
-        }
-      }
+  const rejectInvisibleOrMaskedElementsFilter = (node: Node) => {
+    if (node.nodeType === Node.TEXT_NODE) {
+      return NodeFilter.FILTER_ACCEPT
     }
-
-    // remove the text of elements with programmatic attribute value
-    removeTextFromElements(`[${DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE}]`)
-
-    if (userProgrammaticAttribute) {
-      removeTextFromElements(`[${userProgrammaticAttribute}]`)
+    if (shouldExcludeNode(node, userProgrammaticAttribute, privacyEnabledActionName)) {
+      return NodeFilter.FILTER_REJECT
     }
-
-    if (privacyEnabledActionName) {
-      // remove the text of elements with privacy override
-      removeTextFromElements(
-        `${getPrivacySelector(NodePrivacyLevel.HIDDEN)}, ${getPrivacySelector(NodePrivacyLevel.MASK)}`
-      )
+    const style = getComputedStyle(node as Element)
+    if (style.visibility !== 'visible' || style.display === 'none') {
+      return NodeFilter.FILTER_REJECT
     }
-
-    return text
+    return NodeFilter.FILTER_ACCEPT
   }
 
-  return element.textContent
+  const walker = document.createTreeWalker(
+    element,
+    // eslint-disable-next-line no-bitwise
+    NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_TEXT,
+    rejectInvisibleOrMaskedElementsFilter
+  )
+
+  let text = ''
+  let lastSubstringEndedWithWhiteSpace = false
+
+  while (walker.nextNode()) {
+    const node = walker.currentNode
+    if (node.nodeType !== Node.TEXT_NODE) {
+      if (node.nodeType === Node.ELEMENT_NODE && (node as Element).nodeName === 'BR') {
+        // innerText will preserve <br> tags, but it makes more sense to add a space for action names.
+        if (!lastSubstringEndedWithWhiteSpace) {
+          text += ' '
+        }
+        lastSubstringEndedWithWhiteSpace = true
+      }
+      continue
+    }
+
+    const nodeText = node.textContent
+    if (!nodeText) {
+      continue
+    }
+    const startsWithWhiteSpace = /^\s/.test(nodeText)
+    const endsWithWhiteSpace = /\s$/.test(nodeText)
+
+    const trimmedNodeText = nodeText.trim()
+    if (trimmedNodeText.length === 0) {
+      if (startsWithWhiteSpace || endsWithWhiteSpace) {
+        lastSubstringEndedWithWhiteSpace = true
+      }
+      continue
+    }
+
+    if ((startsWithWhiteSpace || lastSubstringEndedWithWhiteSpace) && text[text.length - 1] !== ' ') {
+      text += ' '
+    }
+
+    text += trimmedNodeText
+    lastSubstringEndedWithWhiteSpace = endsWithWhiteSpace
+  }
+
+  return text
 }


packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -204,47 +204,91 @@ function getActionNameFromTextualContent(
...
-          }
+  const rejectInvisibleOrMaskedElementsFilter = (node: Node) => {
+    if (node.nodeType === Node.TEXT_NODE) {
+      return NodeFilter.FILTER_ACCEPT
+    }
+    if (shouldExcludeNode(node, userProgrammaticAttribute, privacyEnabledActionName)) {
+      return NodeFilter.FILTER_REJECT
+    }
+    const style = getComputedStyle(node as Element)
+    if (style.visibility !== 'visible' || style.display === 'none') {
```
sethfowler-datadog: We should also reject if `style.contentVisibility !== 'visible'`.
```suggestion
    if (style.visibility !== 'visible' || style.display === 'none' || style.contentVisibility !== 'visible') {
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -204,47 +204,91 @@ function getActionNameFromTextualContent(
...
+
+  while (walker.nextNode()) {
+    const node = walker.currentNode
+    if (node.nodeType !== Node.TEXT_NODE) {
+      if (node.nodeType === Node.ELEMENT_NODE && (node as Element).nodeName === 'BR') {
+        // innerText will preserve <br> tags, but it makes more sense to add a space for action names.
+        if (!lastSubstringEndedWithWhiteSpace) {
+          text += ' '
         }
+        lastSubstringEndedWithWhiteSpace = true
```
sethfowler-datadog: Instead of inserting the space here eagerly, wait to insert it until we have some non-whitespace content. In other words, you don't need to do anything here except this:
```suggestion
        lastSubstringEndedWithWhiteSpace = true
```
This will simplify the algorithm; more on this below.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -204,47 +204,91 @@ function getActionNameFromTextualContent(
...
+      }
+      continue
     }
 
-    if (privacyEnabledActionName) {
-      // remove the text of elements with privacy override
-      removeTextFromElements(
-        `${getPrivacySelector(NodePrivacyLevel.HIDDEN)}, ${getPrivacySelector(NodePrivacyLevel.MASK)}`
-      )
+    if ((startsWithWhiteSpace || lastSubstringEndedWithWhiteSpace) && text[text.length - 1] !== ' ') {
```
sethfowler-datadog: I don't think we should need the `text[text.length - 1] !== ' '` condition here. My best guess is that you needed to add it because you are adding spaces in a second place (in the code for handling `<br>`); am I right? If so, I think you can remove it if you make the change to the `<br>` handling code that I recommend above:
```suggestion
    if (startsWithWhiteSpace || lastSubstringEndedWithWhiteSpace) {
```
This simplifies things, and it should also produce a slightly better result, because we will only add the space for a `<br>` element if there's some actual text after the `<br>` element; we won't end up with a dangling space at the end of the action name if the `<br>` element is the last thing.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -204,47 +204,91 @@ function getActionNameFromTextualContent(
...
+  const exclusionSelectors = [`[${DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE}]`]
+  if (userProgrammaticAttribute) {
+    exclusionSelectors.push(`[${userProgrammaticAttribute}]`)
+  }
+  if (privacyEnabledActionName) {
+    exclusionSelectors.push(
+      `${getPrivacySelector(NodePrivacyLevel.HIDDEN)}, ${getPrivacySelector(NodePrivacyLevel.MASK)}`
+    )
+  }
+  return (node as HTMLElement).closest(exclusionSelectors.join(','))
```
sethfowler-datadog: With `TreeWalker`, we are recursively walking over the tree, and our filter is applied to each node in preorder. If the filter returns `FILTER_REJECT`, it doesn't just mean "skip this node"; it means "skip this entire subtree". We can take advantage of this to make this code more performant.

Let's consider in a bit more detail what we actually need to do:

For programmatic action names, `getActionNameFromElement()` should've already handled programmatic action names on the `getTextualContent()` target element or one of its ancestors. We are thus only concerned with rejecting descendant elements which have a programmatic action name attribute. We don't need to worry about ancestors, because we'll always visit and reject ancestors before we ever get to their descendants.

For `HIDDEN`, and for `MASK` if we are to match the current implementation (which does not allow overriding by `ALLOW` in descendant elements), again we can rely on the fact that `getActionNameForElement()` has already checked the target element, so we are only concerned with rejecting descendant elements which are directly marked `HIDDEN` or `MASKED`. Again, we don't need to worry about ancestors, because we'll always visit them first.

We can also improve the correctness of the code and avoid reinventing the wheel by reusing the `getNodePrivacyLevel()` and `getNodeSelfPrivacyLevel()` functions from `privacy.ts`. Using them will ensure consistency with session replay and reduce the risk of privacy bugs.

With these considerations in mind, here's what I'd recommend:
1. In `rejectInvisibleOrMaskedElementsFilter`, check for the programmatic action name attributes using `Element#hasAttribute()`, and return `FILTER_REJECT` if you find them.
2. In `rejectInvisibleOrMaskedElementsFilter`, call `shouldMaskNode(node, getNodeSelfPrivacyLevel(node))`, and return `FILTER_REJECT` if it returns true.
3. Remove the call to `shouldExcludeNode()` in `rejectInvisibleOrMaskedElementsFilter`, and remove `shouldExcludeNode()` itself, since the changes above replace it.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
   }
 }
 
 function getTextualContent(
-  element: Element | HTMLElement,
+  element: Element,
   userProgrammaticAttribute: string | undefined,
-  privacyEnabledActionName?: boolean
+  privacyEnabledActionName?: boolean,
+  nodePrivacyLevel?: NodePrivacyLevel
```
sethfowler-datadog: Nit: make this non-optional and propagate it through the callers. Only tests call `getActionNameFromElement()` without specifying a privacy level, so I'd either update the tests or give the parameter a default value in `getActionNameFromElement()`.
```suggestion
  nodePrivacyLevel: NodePrivacyLevel
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
 ) {
   if ((element as HTMLElement).isContentEditable) {
     return
   }
+  if (nodePrivacyLevel && privacyEnabledActionName && shouldMaskNode(element, nodePrivacyLevel)) {
```
sethfowler-datadog:
```suggestion
  if (privacyEnabledActionName && shouldMaskNode(element, nodePrivacyLevel)) {
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
-          const textToReplace = element.innerText
-          if (textToReplace && textToReplace.trim().length > 0) {
-            text = text.replace(textToReplace, '')
-          }
-        }
+  const rejectInvisibleOrMaskedElementsFilter = (node: Node) => {
+    if (node.nodeType === Node.TEXT_NODE) {
+      return NodeFilter.FILTER_ACCEPT
+    }
+    if (node instanceof HTMLElement && node.hasAttribute(DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE)) {
```
sethfowler-datadog: `instanceof` actually isn't the ideal way to do this check, for two reasons:
* `instanceof` is cheap but does have a small performance cost.
*  There are edge cases where `instanceof` doesn't behave like you'd expect. Moving an element from one document to another can break a check like this, because you're _really_ checking if `node instanceof window.HTMLElement`, and `window.HTMLElement` will be different in e.g. another iframe. People unfortunately do things like that.

In addition, checking for `HTMLElement` specifically isn't ideal; we don't want to exclude `SVGElement` nodes!

We don't need this check anyway, since we know that `node` is either a `TEXT_NODE` or an `ELEMENT_NODE`, and we already eliminated the `TEXT_NODE` case above. Just a type assertion is enough.

Summary:
```suggestion
    if ((node as Element).hasAttribute(DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE)) {
```

However, there's actually an even cleaner way to do it! Just restructure the entire body of `rejectInvisibleOrMaskedElementsFilter` so that it has this form:
```ts
const rejectInvisibleOrMaskedElementsFilter = (node: Node) => {
  if (isElementNode(node) {
    // Do element node stuff.
  } else {
    // We're dealing with a text node.
    return NodeFilter.FILTER_ACCEPT
  }
}
```

The `isElementNode()` helper, defined in `htmlDomUtils.ts`, is a type assertion function. TypeScript will understand that `node` has type `Element` when that function returns true. So, inside the body of the `if`, you can just write:
```suggestion
    if (node.hasAttribute(DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE)) {
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+      return NodeFilter.FILTER_REJECT
+    }
+    if (userProgrammaticAttribute && node instanceof HTMLElement && node.hasAttribute(userProgrammaticAttribute)) {
```
sethfowler-datadog: Assuming you restructure to use `isElementNode()`:
```suggestion
    if (userProgrammaticAttribute && node.hasAttribute(userProgrammaticAttribute)) {
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+      return NodeFilter.FILTER_REJECT
+    }
+    if (
+      nodePrivacyLevel &&
```
sethfowler-datadog:
```suggestion
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+  )
+
+  let text = ''
+  let lastSubstringEndedWithWhiteSpace = false
+
+  while (walker.nextNode()) {
+    const node = walker.currentNode
+    if (node.nodeType !== Node.TEXT_NODE) {
+      if (
+        node.nodeType === Node.ELEMENT_NODE &&
```
sethfowler-datadog: Here again, `TEXT_NODE` and `ELEMENT_NODE` are the only options, so you can simplify:
```suggestion
    if (isElementNode(node)) {
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+    if (node instanceof HTMLElement && node.hasAttribute(DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE)) {
+      return NodeFilter.FILTER_REJECT
+    }
+    if (userProgrammaticAttribute && node instanceof HTMLElement && node.hasAttribute(userProgrammaticAttribute)) {
+      return NodeFilter.FILTER_REJECT
+    }
+    if (
+      nodePrivacyLevel &&
+      privacyEnabledActionName &&
+      shouldMaskNode(node as Element, getNodeSelfPrivacyLevel(node as Element) || nodePrivacyLevel)
```
sethfowler-datadog: Assuming you restructure to use `isElementNode()`:
```suggestion
      shouldMaskNode(node, getNodeSelfPrivacyLevel(node) || nodePrivacyLevel)
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+    ) {
+      return NodeFilter.FILTER_REJECT
+    }
+    const style = getComputedStyle(node as Element)
```
sethfowler-datadog: Assuming you restructure to use isElementNode():
```suggestion
    const style = getComputedStyle(node)
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+
+  let text = ''
+  let lastSubstringEndedWithWhiteSpace = false
+
+  while (walker.nextNode()) {
+    const node = walker.currentNode
+    if (node.nodeType !== Node.TEXT_NODE) {
+      if (
+        node.nodeType === Node.ELEMENT_NODE &&
+        (node.nodeName === 'BR' ||
```
sethfowler-datadog: You should do this for `P` too:
```suggestion
        (node.nodeName === 'BR' || node.nodeName === 'P' ||
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -196,55 +217,101 @@ function getActionNameFromStandardAttribute(element: Element | HTMLElement, attr
...
+          ['block', 'flex', 'grid', 'list-item', 'table', 'table-caption'].includes(
+            getComputedStyle(node as Element).display
```
sethfowler-datadog: You can remove the type assertion if you switch to `isElementNode()`:
```suggestion
            getComputedStyle(node).display
```

packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
```diff
@@ -96,6 +96,24 @@ describe('getActionNameFromElement', () => {
...
+  })
+
+  it('should correctly compute with display elements block, grid, flex, list-item, table, table-caption', () => {
+    const element = appendElement(
+      '<div><div style="display: block;">hello</div><div style="display: grid;">world</div><div style="display: flex;">!<div style="display: table-caption;">!</div></div></div>'
+    )
+    const { name, nameSource } = getActionNameFromElement(element, defaultConfiguration)
+    expect(name).toBe('hello world ! !')
+    expect(nameSource).toBe('text_content')
+  })
```
sethfowler-datadog: A table-driven test might be easier to maintain:
```suggestion
  it('should introduce whitespace for block-level display values', () => {
    const testCases = { display: string, expect: 'space' | 'no-space' | 'nothing' }[] = [
      { display: 'block', expected: 'space' },
      // ...
    ]
    testCases.forEach(({ display, expected }) => {
      const element = appendElement(`<div><div style="display: ${display}">a</div><div>b</div></div>`);
      const { name, nameSource } = getAcitonNameFromElement(element, defaultConfiguration)
      expect(nameSource).toBe('text_content')
      switch (expected) {
        case 'space':
          expect(name).toBe('a b')
        // ...
      }
    }
 ```

File: packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts [hunk 16]
```diff
@@ -428,6 +446,15 @@ describe('getActionNameFromElement', () => {
       expect(nameSource).toBe('text_content')
     })
 
+    it('removes only the child with programmatic action name in textual content', () => {
```
sethfowler-datadog, line 484: Thanks for adding these tests! Thankfully, there are lots of tests already, so we have a good amount of coverage. I think a couple more might help round out the test suite and confirm that everything is working right:
* A test for `<br>` and `<p>`.
* A test to confirm that `<div>unmasked <span>secret</span> followed by masked <span dd-privacy-level"mask">secret</span></div>` masks the second `secret` and not the first, confirming that the issue with the old implementation is fixed.

packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
```diff
@@ -428,6 +478,15 @@ describe('getActionNameFromElement', () => {
...
+      const { name, nameSource } = getActionNameFromElement(
+        appendElement('<div>Foobar <div data-dd-action-name="custom action">bar<div></div>'),
+        defaultConfiguration
+      )
+      expect(name).toBe('Foobar')
```
sethfowler-datadog: The difference between the buggy implementation and the working one comes down to a single space here, which might be a bit confusing when debugging this in the future. I think it'd be worth adding a bit more text to make the difference more clear:
```suggestion
      const { name, nameSource } = getActionNameFromElement(
        appendElement('<div>Foobar Baz<div data-dd-action-name="custom action">bar<div></div>'),
        defaultConfiguration
      )
      expect(name).toBe('Foobar Baz')
```
With this version, the buggy implementation would yield `Foo Bazbar`, which makes it clearer what the problem is.

packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
```diff
@@ -96,6 +96,56 @@ describe('getActionNameFromElement', () => {
     expect(nameSource).toBe('text_content')
   })
 
+  it('should correctly compute whitespace for <br> and <p> elements', () => {
```
sethfowler-datadog, line 101: Thank you for adding these new test cases!

sethfowler-datadog (approved):
Looks great! Thanks for all your hard work on this; I know this one required a lot of discussion and experimentation!

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -248,3 +281,78 @@ function getTextualContent(
...
+  while (walker.nextNode()) {
+    const node = walker.currentNode
+    if (isElementNode(node)) {
+      if (
+        node.nodeName === 'BR' ||
+        node.nodeName === 'P' ||
+        ['block', 'flex', 'grid', 'list-item', 'table', 'table-caption'].includes(getComputedStyle(node).display)
+      ) {
+        lastSubstringEndedWithWhiteSpace = true
+      }
```
BenoitZugmeyer: **suggestion:** I am assuming this comes from the `innerText` spec? Do you have a link to an external documentation that you could add to provide some details?

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -248,3 +281,78 @@ function getTextualContent(
...
   return element.textContent
 }
+
+function getTextualContentWithTreeWalker(
+  element: Element,
+  userProgrammaticAttribute: string | undefined,
+  privacyEnabledActionName: boolean,
+  nodePrivacyLevel: NodePrivacyLevel
+) {
+  const rejectInvisibleOrMaskedElementsFilter = (node: Node) => {
```
BenoitZugmeyer: **nitpick:** to help with readability, I would use a function declaration, and move it at the end of the function. This is quite subjective, feel free to ignore:

```ts
function getTextualContentWithTreeWalker(
  element: Element,
  userProgrammaticAttribute: string | undefined,
  privacyEnabledActionName: boolean,
  nodePrivacyLevel: NodePrivacyLevel
) {

  // ...

  return text


  function rejectInvisibleOrMaskedElementsFilter(node: Node) {
    ...
  }
}
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -248,3 +281,78 @@ function getTextualContent(
...
+    if (isElementNode(node)) {
+      if (
+        node.hasAttribute(DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE) ||
+        (userProgrammaticAttribute && node.hasAttribute(userProgrammaticAttribute)) ||
+        (privacyEnabledActionName && shouldMaskNode(node, getNodeSelfPrivacyLevel(node) || nodePrivacyLevel))
```
BenoitZugmeyer: **question:** I'm not sure to understand the intent with the `nodePrivacyLevel` fallback in `getNodeSelfPrivacyLevel(node) || nodePrivacyLevel)`. It looks fishy:
* if we want to emulate some kind of inheritance, we would probably need to use the `reducePrivacyLevel()` function to respect ancestors privacy levels.
* If we want to REJECT the click target itself, I think it's already done [at a higher level](https://github.com/DataDog/browser-sdk/blob/fe1bead1e13b730729cf0f7ec5f558349636db82/packages/rum-core/src/domain/action/getActionNameFromElement.ts#L42)
* or maybe it is to prepare for the new privacy level for mask allow list?

In any case, tests are passing when I remove the fallback. At least add a test to illustrate the intent.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -248,3 +281,78 @@ function getTextualContent(
...
+    if (trimmedNodeText.length === 0) {
+      if (startsWithWhiteSpace || endsWithWhiteSpace) {
+        lastSubstringEndedWithWhiteSpace = true
+      }
+      continue // skip empty nodes
+    }
+
+    if (startsWithWhiteSpace || lastSubstringEndedWithWhiteSpace) {
+      text += ' '
+    }
```
BenoitZugmeyer: **suggestion:** I wonder if it wouldn't be easier to concatenate text without worrying about white space too much, and then normalizing only at the end:

* For elements that introduce a space (cf test above with `nodeName === 'BR'` etc), just do `text += ' '`.
* Don't trim, don't check if it starts with white space or not.
* At the end, return `text.replace(/\s+/g, ' ').trim()`

Bonus: it would normalize text *within* text nodes, like `"foo      bar"` would be `"foo bar"`, like `innerText` does.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: Indeed, the rejection is already done at a higher level. We will need to pass privacy decisions in for the mask allowlist, but not necessarily `nodePrivacyLevel`. I removed it for now.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: I've added a comment to link the spec.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
cy-moi: Thanks, indeed!

PR #3765, opened 2025-08-07
🐛[RUM-11247] Pass down error stack to isAllowedTrackingOrigins

BeltranBulbarellaDD:
## Motivation

Currently there are 9 function calls before `isAllowedTrackingOrigins`. Chrome passes the first 9, so if we were to add 1 more function call before, we would never get the extension prefix in the stack

## Changes

Created the error stack in `makeRumPublicApi` and `makeLogsPublicApi` and passed it down to `isAllowedTrackingOrigins` function. 

(First error is the new stack, second error is the previous stack)
Screenshot 2025-09-02 at 14 58 32

## Test instructions

Use the dev extension and inject it in localHost:8080 while running the local SDK. If you print the error stack that `isAllowedTrackingOrigins` receives now and print a new error stack below it should be the same as the image above. We should see the error stack in a different location, closer to the top.

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/build/build-test-apps.ts
+++ b/scripts/build/build-test-apps.ts
@@ -8,10 +8,15 @@ import { modifyFile } from '../lib/filesUtils.ts'
 interface ExtensionConfig {
   name: string
   initParameter: string
+  world?: 'MAIN' | 'ISOLATED'
 }
 
 const EXTRA_EXTENSIONS: ExtensionConfig[] = [
-  { name: 'allowed-tracking-origin', initParameter: 'allowedTrackingOrigins: [/^chrome-extension:\\/\\//],' },
+  {
+    name: 'allowed-tracking-origin',
+    initParameter: 'allowedTrackingOrigins: [/^chrome-extension:\\/\\//],',
+    world: 'ISOLATED',
+  },
   { name: 'invalid-tracking-origin', initParameter: "allowedTrackingOrigins: ['https://app.example.com']," },
 ]
 
@@ -67,7 +72,7 @@ async function buildExtensions(): Promise<void> {
 
   buildApp(baseExtDir)
 
-  for (const { name, initParameter } of EXTRA_EXTENSIONS) {
+  for (const { name, initParameter, world } of EXTRA_EXTENSIONS) {
     const targetDir = path.join('test/apps', name)
 
     fs.rmSync(targetDir, { recursive: true, force: true })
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -505,6 +505,7 @@ export function makeRumPublicApi(
   const trackingConsentState = createTrackingConsentState()
   const customVitalsState = createCustomVitalsState()
   const bufferedDataObservable = startBufferingData().observable
+  const errorStack = new Error().stack
```
thomas-lebeau: **🔨 warning:** ‏ I think this will not work when injecting the SDK via CDN, you need to create the error stack in the `init()` methods, as they are the only one that will (almost*) always be ran from the extension's JS

I think adding a e2e test that inject the SDK from the extension using the CDN (i.e. the local bundle using `createCrossOriginScriptUrls()`) will help catch that issue 

(*) Almost, because I'm pretty sure it's possible to work around that by injecting and initializing the SDK with an injected `<script>` tag.

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -6,6 +6,7 @@ export function containsExtensionUrl(str: string): boolean {
 
 /**
  * Utility function to detect if the SDK is being initialized in an unsupported browser extension environment.
+ * Note: Because we check error stack, this will not work if the error stack is too long as Chrome truncates it.
```
thomas-lebeau: **💬 suggestion:** all browser limit the size of the stack trace, most of them to 10 frames.

```suggestion
 * Note: Because we check error stack, this will not work if the error stack is too long as Browsers truncates it.
```

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -336,7 +339,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     !isString(initConfiguration.version, 'Version') ||
     !isString(initConfiguration.env, 'Env') ||
     !isString(initConfiguration.service, 'Service') ||
-    !isAllowedTrackingOrigins(initConfiguration)
+    !isAllowedTrackingOrigins(initConfiguration, undefined, errorStack)
```
thomas-lebeau: **💬 suggestion:** ‏ maybe we can take the opportunity to improve the `isAllowedTrackingOrigins` by 
- removing the default value for `errorStack` as it should always be passed explicitely
- switch the order of arguments

```diff
export function isAllowedTrackingOrigins(
  configuration: InitConfiguration,
-  windowOrigin = typeof location !== 'undefined' ? location.origin : '',
+  errorStack: string,
-  errorStack = new Error().stack
+  windowOrigin = typeof location !== 'undefined' ? location.origin : ''
): boolean {
```

I think this will makes it less confusing

PR #3766, opened 2025-08-07
✨[RUM-10962][Remote config] support js strategy

bcaudan:
## Motivation

Support resolving a remote configuration option to a js variable value by using a simple JSON path.

## Changes

- implements a simple JSON path parser supporting:
  - dot notation
  - bracket notation
  - array exact index
- support retrieving value from a js variable 

## Test instructions

No backend to test the new options yet

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/configuration/jsonPathParser.spec.ts
+++ b/packages/rum-core/src/domain/configuration/jsonPathParser.spec.ts
@@ -24,11 +24,23 @@ describe('parseJsonPath', () => {
     expect(parseJsonPath(String.raw`['foo'][12]`)).toEqual(['foo', '12'])
   })
 
+  it('should extract name selectors replacing escaped sequence by equivalent character', () => {
+    expect(parseJsonPath(String.raw`['foo\n']`)).toEqual(['foo\n'])
+    expect(parseJsonPath(String.raw`['foo\b']`)).toEqual(['foo\b'])
+    expect(parseJsonPath(String.raw`['foo\t']`)).toEqual(['foo\t'])
+    expect(parseJsonPath(String.raw`['foo\f']`)).toEqual(['foo\f'])
+    expect(parseJsonPath(String.raw`['foo\r']`)).toEqual(['foo\r'])
+    expect(parseJsonPath(String.raw`["foo\u03A9"]`)).toEqual(['fooΩ'])
+    expect(parseJsonPath(String.raw`["\u03A9A"]`)).toEqual(['ΩA'])
+    expect(parseJsonPath(String.raw`["\t\u03A9\n"]`)).toEqual(['\tΩ\n'])
+    expect(parseJsonPath(String.raw`['foo\'']`)).toEqual([String.raw`foo'`])
+    expect(parseJsonPath(String.raw`["foo\""]`)).toEqual([String.raw`foo"`])
+    expect(parseJsonPath(String.raw`["foo\/"]`)).toEqual([String.raw`foo/`])
+  })
+
   it('should extract name selectors containing characters not supported in name shorthands', () => {
-    expect(parseJsonPath(String.raw`['foo\n']`)).toEqual([String.raw`foo\n`])
-    expect(parseJsonPath(String.raw`['foo\'']`)).toEqual([String.raw`foo\'`])
-    expect(parseJsonPath(String.raw`["foo\""]`)).toEqual([String.raw`foo\"`])
     expect(parseJsonPath(String.raw`['foo[]']`)).toEqual([String.raw`foo[]`])
+    expect(parseJsonPath(String.raw`['foo.']`)).toEqual([String.raw`foo.`])
   })
 
   it('should return an empty array for an invalid path', () => {
@@ -44,6 +56,8 @@ describe('parseJsonPath', () => {
     expect(parseJsonPath(String.raw`[foo']`)).toEqual([])
     expect(parseJsonPath(String.raw`['foo''bar']`)).toEqual([])
     expect(parseJsonPath(String.raw`['foo\o']`)).toEqual([])
+    expect(parseJsonPath(String.raw`["\u03Z9"]`)).toEqual([])
+    expect(parseJsonPath(String.raw`['foo\u12']`)).toEqual([])
     expect(parseJsonPath(String.raw`['foo']a`)).toEqual([])
     expect(parseJsonPath(String.raw`["foo']`)).toEqual([])
   })
--- a/packages/rum-core/src/domain/configuration/jsonPathParser.ts
+++ b/packages/rum-core/src/domain/configuration/jsonPathParser.ts
@@ -11,6 +11,11 @@
  * - https://github.com/jsonpath-standard
  */
 
+interface ParsingContext {
+  quote: string | undefined
+  escapeSequence: string | undefined
+}
+
 /**
  * Extract selectors from a simple JSON path expression, return [] for an invalid path
  *
@@ -30,13 +35,20 @@ export function parseJsonPath(path: string): string[] {
   const selectors: string[] = []
   let previousToken = Token.START
   let currentToken: Token | undefined
-  let quoteContext: string | undefined
+  const parsingContext: ParsingContext = { quote: undefined, escapeSequence: undefined }
   let currentSelector = ''
   for (const char of path) {
     // find which kind of token is this char
-    currentToken = ALLOWED_NEXT_TOKENS[previousToken].find((token) => TOKEN_PREDICATE[token](char, quoteContext))
+    currentToken = ALLOWED_NEXT_TOKENS[previousToken].find((token) => TOKEN_PREDICATE[token](char, parsingContext))
     if (!currentToken) {
       return []
+    }
+    if (parsingContext.escapeSequence !== undefined && currentToken !== Token.ESCAPE_SEQUENCE_CHAR) {
+      if (!isValidEscapeSequence(parsingContext.escapeSequence)) {
+        return []
+      }
+      currentSelector += resolveEscapeSequence(parsingContext.escapeSequence)
+      parsingContext.escapeSequence = undefined
     }
     if (ALLOWED_SELECTOR_TOKENS.includes(currentToken)) {
       // buffer the char if it belongs to the selector
@@ -49,10 +61,12 @@ export function parseJsonPath(path: string): string[] {
       //        ^   ^     ^
       selectors.push(currentSelector)
       currentSelector = ''
+    } else if (currentToken === Token.ESCAPE_SEQUENCE_CHAR) {
+      parsingContext.escapeSequence = parsingContext.escapeSequence ? `${parsingContext.escapeSequence}${char}` : char
     } else if (currentToken === Token.QUOTE_START) {
-      quoteContext = char
+      parsingContext.quote = char
     } else if (currentToken === Token.QUOTE_END) {
-      quoteContext = undefined
+      parsingContext.quote = undefined
     }
     previousToken = currentToken
   }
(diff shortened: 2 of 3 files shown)


packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+  [Token.BRACKET_START]: (char: string) => char === '[',
+  [Token.BRACKET_END]: (char: string) => char === ']',
+  [Token.NUMBER_LETTER]: (char: string) => NUMBER_CHAR.test(char),
+
+  [Token.QUOTE_START]: (char: string) => QUOTE_CHAR.includes(char),
+  [Token.QUOTE_END]: (char: string, quoteContext?: string) => char === quoteContext,
+  [Token.QUOTE_PROPERTY_LETTER]: (_: string) => true, // any char can be used in property
+  [Token.QUOTE_ESCAPE]: (char: string) => char === '\\',
+  [Token.QUOTE_ESCAPABLE_LETTER]: (char: string, quoteContext?: string) =>
+    `${quoteContext}${QUOTE_ESCAPABLE_LETTERS}`.includes(char),
```
bcaudan: **💭 thought:** ‏To avoid `quoteContext:? string` when we actually expect it to be defined, I explored some strategies with specific state / tokens / predicates for when we are parsing quotes or not but it added quite some extra complexity... 
Let me know if you have some other ideas.

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+ * parseJsonPath("['foo")
+ * => []
+ *
+ *
+ * Useful references:
+ * - https://goessner.net/articles/JsonPath/
+ * - https://jsonpath.com/
+ * - https://github.com/jsonpath-standard
+ */
+export function parseJsonPath(path: string): string[] {
```
amortemousque, line 34: **💭 thought:** ‏I think, this validation logic should also be shared with Web UI so customers can see upfront whether the path is valid.

amortemousque, line 34: **💭 thought:** I think this function should be exported in `rum/src/entries/internal.ts` so that the Web UI can show customers upfront whether the path is valid.

bcaudan: Yes! It is mentioned in the related RFC, I am waiting for the UI to be a bit more advanced, so probably in a following PR.

bcaudan: that seems to be an implementation proposal of the previous comment, so same answer 🙂

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -152,16 +152,36 @@ function resolveDomValue({
...
+    }
+    try {
+      current = current[pathPart] as { [key: string]: unknown }
+    } catch (e) {
+      display.error(`Error accessing: '${path}'`, e)
+      return
+    }
+  }
+
+  return current as unknown
```
amortemousque: **🥜 nitpick:** ‏ Type assertion, seems unnecessary
```suggestion
  return current
```

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+  BRACKET_START,
+  BRACKET_END,
+  NUMBER_LETTER,
+
+  QUOTE_START,
+  QUOTE_END,
+  QUOTE_PROPERTY_LETTER,
+  QUOTE_ESCAPE,
+  QUOTE_ESCAPABLE_LETTER,
+}
```
amortemousque: **🥜 nitpick:** ‏
An example for each could ease the understanding 
```ts

/**
 * List of all tokens in the path
 *
 * @example                        foo.bar['qu\'x'][0]
 *                                |   |   |        |  |
 * Token sequence:                |   |   |        |  |
 * 1. START (before first char) <-+   |   |        |  |
 * 2. VARIABLE_FIRST_LETTER: f        |   |        |  |
 * 3. VARIABLE_LETTER: oo             |   |        |  |
 * 4. DOT: . <------------------------+   |        |  |
 * 5. VARIABLE_FIRST_LETTER: b            |        |  |
 * 6. VARIABLE_LETTER: ar                 |        |  |
 * 7. BRACKET_START: [ <------------------+        |  |
 * 8. QUOTE_START: '                               |  |
 * 9. QUOTE_PROPERTY_LETTER: qu                    |  |
 * 10. QUOTE_ESCAPE: \                             |  |
 * 11. QUOTE_ESCAPABLE_LETTER: '                   |  |
 * 12. QUOTE_PROPERTY_LETTER: x                    |  |
 * 13. QUOTE_END: '                                |  |
 * 14. BRACKET_END: ]                              |  |
 * 15. BRACKET_START: [ <--------------------------+  |
 * 16. NUMBER_LETTER: 0                               |
 * 17. BRACKET_END: ]                                 |
 * 18. END (after last char) <------------------------+
 */
```

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+
+const VARIABLE_FIRST_LETTER = /[a-zA-Z_$]/
+const VARIABLE_LETTER = /[a-zA-Z0-9_$]/
+const NUMBER_CHAR = /[0-9]/
+const QUOTE_ESCAPABLE_LETTERS = '/\\bfnrtu'
```
amortemousque: **🥜 nitpick:** ‏ Maybe add a comment: 
```ts
// /: allow escaping, \: backslash, \b: backspace, \f: form feed, \n: newline, \r: carriage return, \t: tab, \u: unicode

```

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+  [Token.BRACKET_START]: new Set([Token.QUOTE_START, Token.NUMBER_LETTER]),
+  [Token.BRACKET_END]: new Set([Token.DOT, Token.BRACKET_START, Token.END]),
+  [Token.NUMBER_LETTER]: new Set([Token.NUMBER_LETTER, Token.BRACKET_END]),
+
+  [Token.QUOTE_START]: new Set([Token.QUOTE_ESCAPE, Token.QUOTE_END, Token.QUOTE_PROPERTY_LETTER]),
+  [Token.QUOTE_END]: new Set([Token.BRACKET_END]),
+  [Token.QUOTE_PROPERTY_LETTER]: new Set([Token.QUOTE_ESCAPE, Token.QUOTE_END, Token.QUOTE_PROPERTY_LETTER]),
+  [Token.QUOTE_ESCAPE]: new Set([Token.QUOTE_ESCAPABLE_LETTER]),
+  [Token.QUOTE_ESCAPABLE_LETTER]: new Set([Token.QUOTE_ESCAPE, Token.QUOTE_END, Token.QUOTE_PROPERTY_LETTER]),
+}
```
BenoitZugmeyer: **suggestion:** It would be simpler and smaller to have arrays instead of sets:

* No need for `findInSet`, you can just use `[].find`
* Sets are really small, `[].includes` should be as efficient as `set.has`

packages/rum-core/src/domain/configuration/jsonPathParser.ts
BenoitZugmeyer: **question:** I don't understand the escape logic. It doesn't seem to escape anything appart from ignoring the end quote?

* I would expect that `parseJsonPath("['foo\\'bar']")` to be equivalent to the js `["foo'bar"]` but instead the backslash character isn't removed.
* I would expect that "escapable characters" to have a special meaning, like `\n` is a newline etc.

packages/rum-core/src/domain/configuration/jsonPathParser.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  })
+
+  it('should parse array indexes', () => {
+    expect(parseJsonPath('[0]')).toEqual(['0'])
+    expect(parseJsonPath('foo[12]')).toEqual(['foo', '12'])
+    expect(parseJsonPath("['foo'][12]")).toEqual(['foo', '12'])
+  })
+
+  it('should parse property names with unsupported variable name characters', () => {
+    expect(parseJsonPath("['foo\\n']")).toEqual(['foo\\n'])
```
BenoitZugmeyer: **suggestion:** Use `String.raw` to avoid confusing escapes
```suggestion
    expect(parseJsonPath(String.raw`['foo\n']`)).toEqual([String.raw`foo\n`])
```

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+  NUMBER_LETTER,
+
+  QUOTE_START,
+  QUOTE_END,
+  QUOTE_PROPERTY_LETTER,
+  QUOTE_ESCAPE,
+  QUOTE_ESCAPABLE_LETTER,
+}
+
+const VARIABLE_FIRST_LETTER = /[a-zA-Z_$]/
```
BenoitZugmeyer: **nitpick:** `VARIABLE` could be named `KEYWORD` to stick with JavaScript AST types.

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+const VARIABLE_LETTER = /[a-zA-Z0-9_$]/
+const NUMBER_CHAR = /[0-9]/
```
BenoitZugmeyer: **nitpick:** `DIGIT` might be clearer

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  QUOTE_START,
+  QUOTE_END,
+  QUOTE_PROPERTY_LETTER,
+  QUOTE_ESCAPE,
+  QUOTE_ESCAPABLE_LETTER,
+}
+
+const VARIABLE_FIRST_LETTER = /[a-zA-Z_$]/
+const VARIABLE_LETTER = /[a-zA-Z0-9_$]/
```
BenoitZugmeyer: **nitpick:** Rename `LETTER` to `CHAR`, as `_` and `$` are not letters.

I think we could avoid `LETTER` altogether and just une `CHAR` everywhere in this file.

packages/rum-core/src/domain/configuration/jsonPathParser.ts
```diff
@@ -0,0 +1,148 @@
...
+const QUOTE_CHAR = '\'"'
+
+const TOKEN_PREDICATE: { [token in Token]: (char: string, quoteContext?: string) => boolean } = {
+  // no char should match to START or END
+  [Token.START]: (_: string) => false,
+  [Token.END]: (_: string) => false,
+
+  [Token.VARIABLE_FIRST_LETTER]: (char: string) => VARIABLE_FIRST_LETTER.test(char),
+  [Token.VARIABLE_LETTER]: (char: string) => VARIABLE_LETTER.test(char),
+  [Token.DOT]: (char: string) => char === '.',
```
BenoitZugmeyer: **nitpick:** you don't need to specify types for arguments (or unused arguments) in all those functions

```suggestion
  [Token.START]: () => false,
  [Token.END]: () => false,

  [Token.VARIABLE_FIRST_LETTER]: (char) => VARIABLE_FIRST_LETTER.test(char),
  [Token.VARIABLE_LETTER]: (char) => VARIABLE_LETTER.test(char),
  [Token.DOT]: (char) => char === '.',
```

packages/rum-core/src/domain/configuration/jsonPathParser.ts
bcaudan: The idea is that in jsonPath, only specific characters in the property name can be escaped. So when encountering a backslash, we should expect to have one of the [escapable characters](https://www.rfc-editor.org/rfc/rfc9535.html#name-semantics-3) to ensure that the jsonPath is valid.

bcaudan: I'll link to the spec instead https://www.rfc-editor.org/rfc/rfc9535.html#name-semantics-3

packages/rum-core/src/domain/configuration/jsonPathParser.spec.ts
bcaudan: Nice!

packages/rum-core/src/domain/configuration/jsonPathParser.ts
bcaudan: hum, maybe it could be better to stick with JSON path spec naming, I'll try something

packages/rum-core/src/domain/configuration/jsonPathParser.ts
BenoitZugmeyer: I'm fine with having only specific escapable characters, but the output is wrong in your implementation. The spec says:

> A name-selector string MUST be converted to a member name M by removing the surrounding quotes and replacing each escape sequence with its equivalent Unicode character, 

So the escape sequence must be replaced by its corresponding character, so for example `\n` must be replaced by an actual line feed, `\'` must be replaced by `'` etc.

bcaudan: oh I see, I indeed missed that

PR #3769, opened 2025-08-08
[RUM-9181] ✨ Service Worker support for Logs

### LINKED ISSUE #432: Support logging from web workers
Converting #387 to follow this topic.

This behavior is not requested by many customers so it is not in our top priorities right now.
It is definitely something that we would want to integrate but it could take some time before we make progress on that.

mormubis:
## Motivation

Fixes #432. It allows for Logs to run on a Service Worker environment. 

> [!WARNING]
> Some features might not be available, such as: session. 

## Changes

The PR creates a cached global object that either could be a worker service container or window.
Then, it applies the global variable where `window` was used directly.

Special attention to:
- https://github.com/DataDog/browser-sdk/pull/3769/files#diff-b0c8f1a5970d830e7a7ba31f3745c2beaf868d8d035bbd7dcbc0f8b90dcac804R40 XHR is not available in Service Workers
- https://github.com/DataDog/browser-sdk/pull/3769/files#diff-f45e6ec71cb5a31e0ef9036a3c4659589fd66a095e9cac2ffd566cd5945ada76R362 It changes the batch amount of events to 1. This is critical as the Service Worker can be stopped in any moment, and due to privacy and security concerns, Service Workers do not have a way to detect `unload` or `idle` states.
- https://github.com/DataDog/browser-sdk/pull/3769/files#diff-e84bd5db623fd055ca6a8444de22a4f2c2e8632713aff8402dbf79469af41a85R57 Visibility events are not available, so we ignore them with an empty Observable.

## Test instructions

Tests passed

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/browser/cookie.ts
+++ b/packages/core/src/browser/cookie.ts
@@ -1,7 +1,6 @@
 import { display } from '../tools/display'
 import { ONE_MINUTE, ONE_SECOND } from '../tools/utils/timeUtils'
 import { findCommaSeparatedValue, findCommaSeparatedValues, generateUUID } from '../tools/utils/stringUtils'
-import { globalObject } from '../tools/globalObject'
 
 export interface CookieOptions {
   secure?: boolean
@@ -47,10 +46,9 @@ export function deleteCookie(name: string, options?: CookieOptions) {
 }
 
 export function areCookiesAuthorized(options: CookieOptions): boolean {
-  if (globalObject.document?.cookie === undefined || globalObject.document?.cookie === null) {
+  if (document.cookie === undefined || document.cookie === null) {
     return false
   }
-
   try {
     // Use a unique cookie name to avoid issues when the SDK is initialized multiple times during
     // the test cookie lifetime


test/e2e/lib/framework/pageSetups.ts
```diff
@@ -22,6 +23,10 @@ export interface SetupOptions {
   testFixture: typeof test
 }
 
+export interface WorkerOptions {
+  importScripts?: boolean
+}
```
mormubis: This allows to compile in environments without `lib`: `WebWorker` available.

packages/core/src/transport/eventBridge.ts
```diff
@@ -47,11 +47,15 @@ export function bridgeSupports(capability: BridgeCapability): boolean {
...
   const bridge = getEventBridge()
+
   return (
     !!bridge &&
     bridge
       .getAllowedWebViewHosts()
-      .some((allowedHost) => currentHost === allowedHost || currentHost.endsWith(`.${allowedHost}`))
+      .some(
+        (allowedHost) =>
+          currentHost === allowedHost || currentHost.endsWith(`.${allowedHost}`) || currentHost === 'localhost'
```
mormubis: It seems harmless to allow development mode here. But I want to hear your thoughts on this.

test/e2e/lib/framework/httpServers.ts
```diff
@@ -92,7 +93,7 @@ function instantiateServerOnPort(port: number): Promise<http.Server> {
   return new Promise((resolve, reject) => {
     const server = http.createServer()
     server.on('error', reject)
-    server.listen(port, getIp(), () => {
+    server.listen(port, () => {
```
mormubis, line 96: This allows me to use `localhost`.

test/e2e/scenario/logs.scenario.ts
```diff
@@ -5,7 +5,66 @@ import { APPLICATION_ID } from '../lib/helpers/configuration'
...
 const UNREACHABLE_URL = 'http://localhost:9999/unreachable'
 
+declare global {
+  interface Window {
+    myServiceWorker: ServiceWorkerRegistration
+  }
+}
+
 test.describe('logs', () => {
+  createTest('service worker with worker logs - esm').run(
```
bcaudan: **💬 suggestion:** it could be valuable to cover:
- the features that we expect to work in this environment, like forward console / error / reports 
- the expected transport behavior with maybe something around worker termination

packages/core/src/tools/globalObject.ts
```diff
@@ -0,0 +1,22 @@
+import { getGlobalObject } from './getGlobalObject'
```
bcaudan: **💬 suggestion:** ‏those two files seems pretty related, any reason to not merge them?

packages/core/src/tools/globalObject.ts
```diff
@@ -0,0 +1,22 @@
...
+type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U
+
+export type GlobalObject = XOR<Window, WorkerGlobalScope>
+
+/**
+ * Cached reference to the global object so it can be imported and re-used without
+ * re-evaluating the heavyweight fallback logic in `getGlobalObject()`.
+ */
+// eslint-disable-next-line local-rules/disallow-side-effects
+export const globalObject = getGlobalObject<GlobalObject>()
```
bcaudan, line 49: **❓ question:** ‏I am curious about the heavyweight part, do you have more details on that?
also, should we favor using this cached reference instead where getGlobalObject is currently used then?

packages/core/src/tools/globalObject.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const isSW = !('document' in globalObject)
```
bcaudan: **💬 suggestion:** ‏avoid using acronyms to improve readability

**❓ question:** ‏About "service worker" mentions, is there anything that prevent to use it from web workers for example? if not, maybe we could just mention "worker" instead
```suggestion
export const isInsideWorker = !('document' in globalObject)
```

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -184,6 +189,22 @@ export function reactSetup(options: SetupOptions, servers: Servers, appName: str
...
+      DD_LOGS.init(${formatConfiguration(DEFAULT_LOGS_CONFIGURATION, servers)})
+
+      // Handle messages from main thread
+      self.addEventListener('message', (event) => {
+        const message = event.data;
+
+        DD_LOGS.logger.log(message);
+      });
+    `
+}
```
bcaudan: **💬 suggestion:** ‏if we want to build more service worker tests, having a helper that allow to build the worker code from the test would be helpful

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -358,7 +359,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     /**
      * Logs intake limit
      */
-    batchMessagesLimit: 50,
+    batchMessagesLimit: isSW ? 1 : 50,
```
bcaudan: **❓ question:** ‏that seems to open the door to workers spamming us with requests, did we consider alternatives for that?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -53,7 +54,7 @@ export function selectSessionStoreStrategyType(
       return selectLocalStorageStrategy()
 
     case undefined: {
-      let sessionStoreStrategyType = selectCookieStrategy(initConfiguration)
+      let sessionStoreStrategyType = isSW ? selectLocalStorageStrategy() : selectCookieStrategy(initConfiguration)
```
bcaudan: **❓ question:** ‏what is our intent with session management in worker? would a stub session manager be enough for this use case?

packages/core/src/domain/synthetics/syntheticsWorkerValues.ts
```diff
@@ -12,7 +13,8 @@ export interface BrowserWindow extends Window {
 
 export function willSyntheticsInjectRum(): boolean {
   return Boolean(
-    (window as BrowserWindow)._DATADOG_SYNTHETICS_INJECTS_RUM || getInitCookie(SYNTHETICS_INJECTS_RUM_COOKIE_NAME)
```
bcaudan, line 15: **💬 suggestion:** We should probably just return false if we are inside a worker

packages/core/src/tools/sendToExtension.ts
```diff
@@ -1,11 +1,13 @@
+import { globalObject } from './globalObject'
+
 interface BrowserWindow {
   __ddBrowserSdkExtensionCallback?: (message: unknown) => void
 }
 
 type ExtensionMessageType = 'logs' | 'record' | 'rum' | 'telemetry'
 
 export function sendToExtension(type: ExtensionMessageType, payload: unknown) {
-  const callback = (window as BrowserWindow).__ddBrowserSdkExtensionCallback
```
bcaudan, line 8: **❓ question:** ‏Does the extension allow us to see the logs generated by a worker? if not, it could be a nice to see if we can do something about it

packages/logs/src/boot/startLogs.ts
```diff
@@ -52,7 +54,7 @@ export function startLogs(
   lifeCycle.subscribe(LifeCycleEventType.LOG_COLLECTED, (log) => sendToExtension('logs', log))
 
   const reportError = startReportError(lifeCycle)
-  const pageMayExitObservable = createPageMayExitObservable(configuration)
+  const pageMayExitObservable = isSW ? new Observable<PageMayExitEvent>() : createPageMayExitObservable(configuration)
```
bcaudan: **❓ question:** ‏for xhrObservable, the worker check + dummy observable fallback is done inside the component. I guess we should probably be consistent on how we do it, any pros/cons on doing it inside or outside?

packages/logs/src/domain/contexts/commonContext.ts
```diff
@@ -1,10 +1,11 @@
...
 
 export function buildCommonContext(): CommonContext {
   return {
     view: {
-      referrer: document.referrer,
-      url: window.location.href,
+      referrer: globalObject.document?.referrer ?? '',
+      url: globalObject.location.href,
     },
   }
```
bcaudan: **❓ question:** ‏does that make sense in a worker context?

packages/core/src/domain/session/sessionStore.ts
mormubis: I'm unsure what would be the best choice for session management in a worker environment. I think a stub session is fine, but I'll defer to you or @BenoitZugmeyer

packages/logs/src/domain/contexts/commonContext.ts
mormubis: It's a bit annoying because `CommonContext` is expected in every event. Although, I don't see the intent behind this, it keeps all the events in compliance with the format.

bcaudan: For logs, unlike RUM events, there is not a really a format to comply to. The types are more for static checking and documentation
If some required attributes don't make sense for workers, I think it should be ok to put them as optional.

packages/core/src/tools/sendToExtension.ts
mormubis: I don't think we can inject anything into the Service Worker (or Web Worker) to be called. However, I confirmed that we can debug the events if we set the extension to request mode instead of using the SDK interceptor.

packages/logs/src/boot/startLogs.ts
mormubis: I would suggest that makes more sense to be on the outside so the component is agnostic. Doing the changes.

packages/core/src/domain/synthetics/syntheticsWorkerValues.ts
mormubis: Done!

packages/logs/src/boot/startLogs.ts
mormubis: Done!

packages/logs/src/domain/contexts/commonContext.ts
mormubis: Returning empty object while running inside a Service Worker

test/e2e/lib/framework/pageSetups.ts
mormubis: Created a few tools

packages/core/src/tools/globalObject.ts
mormubis: Done

test/e2e/scenario/logs.scenario.ts
mormubis: Added another test for console forwarding

packages/core/src/tools/globalObject.ts
mormubis: Combined both

packages/core/src/browser/cookie.ts
```diff
@@ -46,9 +47,10 @@ export function deleteCookie(name: string, options?: CookieOptions) {
 }
 
 export function areCookiesAuthorized(options: CookieOptions): boolean {
-  if (document.cookie === undefined || document.cookie === null) {
+  if (isWorkerEnvironment || globalObject.document?.cookie === undefined || globalObject.document?.cookie === null) {
```
BenoitZugmeyer: **question:** why do we need the `isWorkerEnvironment` condition here?

packages/core/src/domain/synthetics/syntheticsWorkerValues.ts
BenoitZugmeyer: > **💬 suggestion:** We should probably just return false if we are inside a worker

Why? I don't think it would hurt to check those global variables/cookies event in the worker. I would keep checking for `isWorkerEnvironment` only when we really need to.

packages/core/src/tools/utils/timeUtils.ts
```diff
@@ -109,7 +110,8 @@ let navigationStart: TimeStamp | undefined
  */
 function getNavigationStart() {
   if (navigationStart === undefined) {
-    navigationStart = performance.timing.navigationStart as TimeStamp
+    navigationStart = (isWorkerEnvironment ? performance.timeOrigin : performance.timing.navigationStart) as TimeStamp
```
BenoitZugmeyer: **suggestion:** maybe we could have a fallback like:

```ts
performance.timing?.navigationStart ?? performance.timeOrigin
```

packages/core/src/transport/eventBridge.ts
BenoitZugmeyer: I'm not sure about that... If you are going to create a bridge object for testing purpose, why not returning `localhost` from `getAllowedWebViewHosts()`?

packages/logs/src/boot/startLogs.ts
```diff
@@ -52,7 +54,9 @@ export function startLogs(
   lifeCycle.subscribe(LifeCycleEventType.LOG_COLLECTED, (log) => sendToExtension('logs', log))
 
   const reportError = startReportError(lifeCycle)
-  const pageMayExitObservable = createPageMayExitObservable(configuration)
+  const pageMayExitObservable = isWorkerEnvironment
+    ? new Observable<PageMayExitEvent>()
+    : createPageMayExitObservable(configuration)
```
BenoitZugmeyer, line 60: **suggestion:** move the `isWorkerEnvironment` check in the `createPageMayExitObservable` function, so we can directly see that we return early because events are not available in workers

packages/logs/src/domain/networkError/networkErrorCollection.ts
```diff
@@ -24,7 +26,10 @@ export function startNetworkErrorCollection(configuration: LogsConfiguration, li
     return { stop: noop }
   }
 
-  const xhrSubscription = initXhrObservable(configuration).subscribe((context) => {
+  // XHR is not available in web workers
+  const xhrSubscription = (
+    isWorkerEnvironment ? new Observable<XhrCompleteContext>() : initXhrObservable(configuration)
+  ).subscribe((context) => {
```
BenoitZugmeyer: **suggestion:** do the `isWorkerEnvironment` in `initXhrObservable`, so we don't need to do it multiple times in case we use that function in other places.

test/e2e/lib/framework/createTest.ts
```diff
@@ -236,10 +220,12 @@ function createTestContext(
...
   browserLogsManager: BrowserLogsManager,
   browserName: TestContext['browserName'],
-  { basePath }: SetupOptions
+  { basePath, logsWorker }: SetupOptions
 ): TestContext {
+  const url = servers.base.url
+
   return {
-    baseUrl: servers.base.url + basePath,
+    baseUrl: (logsWorker ? url.replace(/http:\/\/[^:]+:/, 'http://localhost:') : url) + basePath,
```
BenoitZugmeyer: **question:** why are we doing this again? Why not running things always on localhost ?

test/e2e/lib/framework/createTest.ts
```diff
@@ -290,6 +283,28 @@ async function setUpTest(browserLogsManager: BrowserLogsManager, { baseUrl, page
...
+    }
+
+    await page.evaluate(`
+      if ('serviceWorker' in navigator) {
+        navigator.serviceWorker.register('/sw.js${params.length > 0 ? `?${params.join('&')}` : ''}', ${isModule ? '{ type: "module" }' : '{}'})
+          .then(registration => {
+            window.myServiceWorker = registration;
+          });
+      }
+    `)
```
BenoitZugmeyer: **suggestion:** this looks like something we could put in `pageSetup.ts` in a `script` tag, so we don't need to use `evaluate` and trim a bit of logic from `createTest`

test/e2e/scenario/logs.scenario.ts
```diff
@@ -5,7 +5,53 @@ import { APPLICATION_ID } from '../lib/helpers/configuration'
...
+  interface Window {
+    myServiceWorker: ServiceWorkerRegistration
+  }
+}
+
 test.describe('logs', () => {
+  createTest('service worker with worker logs - esm')
+    .withWorker()
+    .run(async ({ flushEvents, intakeRegistry, browserName, withWorker }) => {
+      console.log('>>>', browserName)
```
BenoitZugmeyer: **suggestion:** 
```suggestion
```

test/e2e/scenario/logs.scenario.ts
```diff
@@ -5,7 +5,53 @@ import { APPLICATION_ID } from '../lib/helpers/configuration'
...
+      test.skip(browserName !== 'chromium', 'Non-Chromium browsers do not support ES modules in Service Workers')
+
+      await withWorker((worker) => {
+        worker.postMessage('Some message')
+      })
+
+      await flushEvents()
+
+      expect(intakeRegistry.logsRequests).toHaveLength(1)
```
BenoitZugmeyer, line 26: **suggestion:** check the log message, just in case (it might be an error or something)

Same for tests below

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -384,7 +385,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     /**
      * Logs intake limit
      */
-    batchMessagesLimit: 50,
+    batchMessagesLimit: isWorkerEnvironment ? 1 : 50,
```
BenoitZugmeyer, line 390: **suggestion:**  It could be nice to add a comment why we are doing this.

packages/core/src/tools/globalObject.ts
```diff
@@ -2,6 +2,19 @@
  * inspired by https://mathiasbynens.be/notes/globalthis
  */
 
+// Extend/Create the WorkerGlobalScope interface to avoid issues when used in a non-browser tsconfig environment
+declare global {
+  interface WorkerGlobalScope {
+    empty: never
+  }
+}
```
BenoitZugmeyer: **suggestion:** never use `declare global` or it will end up in customer global variables.

packages/core/src/tools/globalObject.ts
```diff
@@ -2,6 +2,19 @@
...
+
+// Utility type to enforce that exactly one of the two types is used
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never }
+type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U
+
+export type GlobalObject = XOR<Window, WorkerGlobalScope>
```
BenoitZugmeyer, line 14: **thought:** this is a clever trick but it could be nice, at some point, to have a more declarative approach of defining the global object so we don't need to cast it all the time. Example:


```ts
export interface GlobalObject {
  // Some properties are not available in workers, make them optional
  location?: Window['location'],
  
  // We use types that are not yet available in the TypeScript version we support, declare them here:
  cookieStore: ...,

  // We use custom types because we know better than TS:
  PerformanceObserver: RumPerformanceObserver,

   // We use some custom global variables from time to time, put them here too
  _DATADOG_SYNTHETICS_PUBLIC_ID: ...
  $DD_ALLOWED: ...
}
```

(But please don't do it in this PR)

packages/core/src/tools/globalObject.ts
mormubis: If the customer environment does not add `Worker` to the `lib` field in the `tsconfig.json` file, it breaks the compilation. This is a safety precaution. Do you think there is other way to do it?

packages/core/src/tools/globalObject.ts
mormubis: Most of the time, we don't need to cast `globalObject`, but it's needed for accessing to the document, cookies or the active service worker.

packages/core/src/domain/configuration/configuration.ts
mormubis: I am writing a small document to discuss this further with the team. @bcaudan suggested we need everybody's eyes as it's very critical. After everybody reviews the document, I'll update it with a comment.

test/e2e/lib/framework/createTest.ts
mormubis: There were several issues with the bridge and Safari. BrowserStack replaces localhost with `bs-local.com` and blocks some tests.

Additionally, localhost has higher privileges, and I don't want to skip any checks, such as CSP or similar, just for this test.

packages/core/src/domain/synthetics/syntheticsWorkerValues.ts
bcaudan: My first thought was that if we need to tweak something there to make it work, maybe we can just exit early, as I would not expect RUM SDK to be injected in worker environment.
I guess the only issue is with cookies, where if we want to be able to call the methods in a worker environment, we need to use `globalObject.document?.cookie`. 
It could be nice to either use that everywhere we access cookies or prevent to use any methods accessing cookies when we are in a worker, in order to not have a different implementation depending on where a function accessing cookies is used.

test/e2e/scenario/logs.scenario.ts
mormubis: Fixed

packages/core/src/browser/cookie.ts
mormubis: Removed!

packages/core/src/tools/utils/timeUtils.ts
mormubis: Applied

test/e2e/scenario/logs.scenario.ts
mormubis: Added

packages/logs/src/boot/startLogs.ts
mormubis: In my opinion, keeping this kind of checks outside of the observable itself, make the observables simpler. What are your thoughts?

packages/core/src/transport/eventBridge.ts
mormubis: Removed it!

packages/core/src/tools/globalObject.ts
BeltranBulbarellaDD: IIUC 'heavyweight' means that we don't keep reevaluating getGlobalObject on every usage. I believe it comes from mutating the [Object.prototype](https://stackoverflow.com/questions/23807805/why-is-mutating-the-prototype-of-an-object-bad-for-performance)

BeltranBulbarellaDD: "should we favor using this cached...", I would think yes!

bcaudan: > I believe it comes from mutating the [Object.prototype](https://stackoverflow.com/questions/23807805/why-is-mutating-the-prototype-of-an-object-bad-for-performance)

Interesting

> "should we favor using this cached...", I would think yes!

It could have been a nice preliminary refactor but it could be tackled in a following PR.

packages/core/src/browser/cookie.ts
```diff
@@ -46,9 +47,10 @@ export function deleteCookie(name: string, options?: CookieOptions) {
 }
 
 export function areCookiesAuthorized(options: CookieOptions): boolean {
-  if (document.cookie === undefined || document.cookie === null) {
+  if (globalObject.document?.cookie === undefined || globalObject.document?.cookie === null) {
```
bcaudan: **❓ question:** ‏can this function still be called from a worker environment?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -53,7 +54,9 @@ export function selectSessionStoreStrategyType(
       return selectLocalStorageStrategy()
 
     case undefined: {
-      let sessionStoreStrategyType = selectCookieStrategy(initConfiguration)
+      let sessionStoreStrategyType = isWorkerEnvironment
+        ? selectLocalStorageStrategy()
+        : selectCookieStrategy(initConfiguration)
```
bcaudan: **💬 suggestion:** should not we avoid completely to call `selectSessionStoreStrategyType` in a worker environment?

packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -229,7 +230,7 @@ function startTelemetryTransport(
 
 function getRuntimeEnvInfo(): RuntimeEnvInfo {
   return {
-    is_local_file: window.location.protocol === 'file:',
+    is_local_file: globalObject.location?.protocol === 'file:',
     is_worker: 'WorkerGlobalScope' in self,
```
bcaudan: **💬 suggestion:** ‏what about using `isWorkerEnvironment` here as well?

packages/core/src/tools/utils/timeUtils.ts
```diff
@@ -109,7 +109,8 @@ let navigationStart: TimeStamp | undefined
  */
 function getNavigationStart() {
   if (navigationStart === undefined) {
-    navigationStart = performance.timing.navigationStart as TimeStamp
+    navigationStart = (performance.timing?.navigationStart ?? performance.timeOrigin) as TimeStamp
```
bcaudan, line 113: **💬 suggestion:** the fallback to `timeOrigin` does not really make the intent explicit, it could be interesting to either test `isWorkerEnvironment` to select the right metric or add a comment‏

test/e2e/scenario/logs.scenario.ts
```diff
@@ -5,7 +5,65 @@ import { APPLICATION_ID } from '../lib/helpers/configuration'
...
+      expect(intakeRegistry.logsEvents[0].message).toBe('Other message')
+    })
+
+  createTest('service worker console forwarding')
+    .withWorker({ importScript: true, nativeLog: true })
```
bcaudan, line 49: **💬 suggestion:** ‏instead of having a single worker implementation with parameters to enable/disable different behaviors, what about passing the worker implementation for the test to the `withWorker` method?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -5,7 +5,65 @@ import { APPLICATION_ID } from '../lib/helpers/configuration'
...
+    .run(async ({ flushEvents, intakeRegistry, withWorker, browserName }) => {
+      test.skip(
+        browserName === 'webkit',
+        'BrowserStack overrides the localhost URL with bs-local.com and cannot be used to install a Service Worker'
+      )
+
+      await withWorker((worker) => {
```
bcaudan: **💬 suggestion:** ‏having two `withWorker` methods is a bit confusing, what about renaming this one to something like `interactWithWorker`?

test/e2e/scenario/logs.scenario.ts
bcaudan: what about the other features?

test/e2e/scenario/logs.scenario.ts
mormubis: Is that possible? Mocks run with the servers for testing, but they execute before any of the tests. I don't believe it's feasible to do this.

packages/core/src/tools/utils/timeUtils.ts
mormubis: That was the previous version of this code and @BenoitZugmeyer suggested the fallback. Let's agree on one. What are your thoughts?

packages/core/src/domain/session/sessionStore.ts
mormubis: I don't think it's possible. It's called when assembling the configuration. Do you have any suggestions on how to do it?

packages/core/src/browser/cookie.ts
mormubis: I kept this just in case the configuration is trying to select the cookie persistence.

test/e2e/scenario/logs.scenario.ts
mormubis: Updated it!

packages/core/src/domain/synthetics/syntheticsWorkerValues.ts
BenoitZugmeyer: > we need to use globalObject.document?.cookie. It could be nice to either use that everywhere we access cookies

This sounds good to me. If it's annoying we could have a helper like `function getCookieString() { return globalObject.document?.cookie }`

packages/core/src/tools/globalObject.ts
BenoitZugmeyer: Customers shouldn't have to add `Worker` to the `lib` field. This is a breaking change. Let's see if we can find something else.

packages/core/src/tools/utils/timeUtils.ts
BenoitZugmeyer: A comment is good.

packages/logs/src/boot/startLogs.ts
BenoitZugmeyer: Let's document it then. From looking at those three lines, it's really not obvious why we don't call the `createPageMayExitObservable` function

test/e2e/lib/framework/createTest.ts
```diff
@@ -238,10 +255,12 @@ function createTestContext(
...
   browserLogsManager: BrowserLogsManager,
   browserName: TestContext['browserName'],
-  { basePath }: SetupOptions
+  { basePath, useServiceWorker }: SetupOptions
 ): TestContext {
+  const url = servers.base.url
+
   return {
-    baseUrl: servers.base.url + basePath,
+    baseUrl: (useServiceWorker ? url.replace(/http:\/\/[^:]+:/, 'http://localhost:') : url) + basePath,
```
BenoitZugmeyer: **suggestion:** this isn't trivial either. Please document it.

test/e2e/lib/framework/httpServers.ts
```diff
@@ -47,7 +47,8 @@ export async function waitForServersIdle() {
 
 async function createServer<App extends ServerApp>(): Promise<Server<App>> {
   const server = await instantiateServer()
-  const { address, port } = server.address() as AddressInfo
+  const address = getIp()
+  const { port } = server.address() as AddressInfo
```
BenoitZugmeyer, line 51: **thought:** It's a bit confusing to me that we don't use the same address than the one we listen to. But maybe that's fine..

test/e2e/lib/framework/createTest.ts
mormubis: Added doc

test/e2e/lib/framework/httpServers.ts
mormubis: It's because when you listen to all ports, the `address` given by `server.address()` is `undefined`. So I needed to use the IP.

packages/core/src/tools/utils/timeUtils.ts
mormubis: Added

packages/core/src/tools/globalObject.ts
mormubis: I removed it

packages/logs/src/boot/startLogs.ts
mormubis: Added a comment

test/e2e/lib/framework/httpServers.ts
```diff
@@ -93,7 +93,7 @@ function instantiateServerOnPort(port: number): Promise<http.Server> {
   return new Promise((resolve, reject) => {
     const server = http.createServer()
     server.on('error', reject)
-    server.listen(port, () => {
+    server.listen(server.address(), port, () => {
```
BenoitZugmeyer: **suggestion:** this is incorrect, `server.address()` will return `null` before calling `listen()`, and `server.address(null, port)` won't work as expected. If you want to be explicit, use `0.0.0.0` as an address to listen to any traffic:

```
server.listen(port, "0.0.0.0", ...)
```

This is similar to the default host (only difference is that by default it can bind as IPv6, providing a IPv4 address will force to bind as IPv4)

Or just don't provide a host like you didi before.

test/e2e/scenario/logs.scenario.ts
bcaudan: I think we could do something similar to what is done for remoteConfiguration:
- add an helper to add the worker implementation to the test builder
- pass the worker implementation to setup options from the run function
- pass the worker implementation to the createMockServerApp from the declare test function
- use the worker implementation in the `/sw.js` route

packages/core/src/domain/session/sessionStore.ts
bcaudan: what about not calling it when assembling the configuration:
```ts
sessionStoreStrategyType: !isWorkerEnvironment 
    ? selectSessionStoreStrategyType(initConfiguration)
    : undefined
```

it would prevent any access to cookie/local storage and spare us the `!isWorkerEnvironment` check in start logs:
```ts
const session =
    configuration.sessionStoreStrategyType && !canUseEventBridge() && !willSyntheticsInjectRum() && !isWorkerEnvironment
      ? startLogsSessionManager(configuration, trackingConsentState)
      : startLogsSessionManagerStub(configuration)
```

packages/core/src/browser/cookie.ts
bcaudan: if we manage to avoid to select cookie persistence we could spare us that, cf https://github.com/DataDog/browser-sdk/pull/3769#discussion_r2325127426

test/e2e/scenario/logs.scenario.ts
```diff
@@ -5,7 +5,65 @@ import { APPLICATION_ID } from '../lib/helpers/configuration'
...
 
+declare global {
+  interface Window {
+    myServiceWorker: ServiceWorkerRegistration
+  }
+}
+
 test.describe('logs', () => {
+  createTest('service worker with worker logs - esm')
+    .withWorker()
```
bcaudan, line 16: **💬 suggestion:** ‏what about extracting worker tests to a dedicated file?

mormubis: Can I do that in a follow-up PR?

packages/core/src/domain/session/sessionStore.ts
mormubis: Should we move all the conditions to the assembling configuration step?

packages/core/src/browser/cookie.ts
bcaudan: ping

packages/logs/src/boot/startLogs.ts
```diff
@@ -65,7 +70,7 @@ export function startLogs(
   cleanupTasks.push(telemetry.stop)
 
   const session =
-    configuration.sessionStoreStrategyType && !canUseEventBridge() && !willSyntheticsInjectRum()
+    configuration.sessionStoreStrategyType && !canUseEventBridge() && !willSyntheticsInjectRum() && !isWorkerEnvironment
```
bcaudan: **💬 suggestion:** ‏worker environment check should not be needed anymore as `sessionStoreStrategyType` should be `undefined`

bcaudan (approved):
Nice! LGTM with suggested changes on tests addressed in a follow up PR

BenoitZugmeyer (approved):

PR #3776, opened 2025-08-11
[RUM Profiler] Fix discrepancy on view-id correlation

thomasbertet:
## Motivation

- We've identified a flaw in the view-id correlation for Profiles collected. This PR fixes it.
- The flaw can be summarized as : 
   - The view id associated with Profiles is always the initial view id defined when the Profiler initially started.
   - Any new view id created along the way only lives in the Profile that was active at the time, but not in the following profiles.

## Changes

- Whenever a "view" is created, we update the `lastViewEntry` that will be used for the next Profiler instance.

## Test instructions

- Added a unit test to cover regressions.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -128,12 +128,17 @@ export function createRumProfiler(
 
     // Whenever the View is updated, we add a views entry to the profiler instance.
     const viewUpdatedSubscription = lifeCycle.subscribe(LifeCycleEventType.VIEW_CREATED, (view) => {
-      // Note: `view.name` is only filled when users use manual view creation via `startView` method.
-      collectViewEntry({
+      const viewEntry = {
         viewId: view.id,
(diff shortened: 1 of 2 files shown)


packages/rum/src/domain/profiling/profiler.ts
```diff
@@ -128,12 +128,17 @@ export function createRumProfiler(
...
+        // Note: `viewName` is only filled when users use manual view creation via `startView` method.
         viewName: getCustomOrDefaultViewName(view.name, document.location.pathname),
         startClocks: view.startClocks,
-      })
+      }
+
+      collectViewEntry(viewEntry)
+
+      // Update last view entry
+      lastViewEntry = viewEntry
```
thomasbertet, line 141: This is the key part of the PR. Keeping track of the latest view-id created.

BenoitZugmeyer (approved):

PR #3779, opened 2025-08-13
🔊 [RUM-11233] collect remote configuration metrics

bcaudan:
## Motivation

Monitor success/failure of:

- remote configuration fetching
- resolution of dynamic options

## Changes

- simplify telemetry metric management and use a single draw for all metrics
- collect remote configuration metrics

## Test instructions

UI not yet available, you can test locally by:
- setting a remoteConfigurationId and a local override
- updating `remoteConfigurationTelemetrySampleRate` from 1 to 100

Then metrics should be collected corresponding to the fetching/resolution of the remote configuration:

Screenshot 2025-09-15 at 14 52 28

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -109,10 +123,18 @@ export function startTelemetryCollection(
     [TelemetryType.USAGE]: telemetryEnabled && performDraw(configuration.telemetryUsageSampleRate),
   }
 
+  const telemetryEnabledPerMetrics: { [metricName: string]: boolean } = {}
+  Object.keys(sampleRateByMetric).forEach((metricName) => {
+    telemetryEnabledPerMetrics[metricName] = telemetryEnabled && performDraw(sampleRateByMetric[metricName])
+  })
```
amortemousque: **💭 thought:** ‏I’m wondering if we could, since every telemetry metric ends with `TelemetrySampleRate`. We could:
```ts
export interface TelemetrySampleRate {
  customerDataTelemetrySampleRate: number
  initialViewMetricsTelemetrySampleRate: number
  replayTelemetrySampleRate: number
  remoteConfigurationTelemetrySampleRate: number
}

export type TelemetryMetric = keyof TelemetrySampleRate extends `${infer Name}TelemetrySampleRate` ? Name : never

export interface RumConfiguration extends Configuration, TelemetrySampleRate { ... }

function startTelemetry(...) {
  const telemetryEnabledPerMetrics: Partial<Record<TelemetryMetric, boolean>> = {}

  return {
    isTelemetryMetricEnabled(telemetryMetric: TelemetryMetric) {
      if (!telemetryEnabledPerMetrics[telemetryMetric]) {
        telemetryEnabledPerMetrics[telemetryMetric] = telemetryEnabled && performDraw(configuration[`${telemetryMetric}TelemetrySampleRate`)
      } 
      return telemetryEnabledPerMetrics[telemetryMetric]
    }
  }
}
```

Wdyt?

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -149,24 +205,40 @@ export function applyRemoteConfiguration(
...
-      }
-      try {
-        current = current[pathPart] as { [key: string]: unknown }
-      } catch (e) {
-        display.error(`Error accessing: '${path}'`, e)
-        return
+    } else {
+      for (const pathPart of pathParts) {
+        if (!(pathPart in current)) {
+          missing = true
```
BenoitZugmeyer: **suggestion:** it looks like it would be simpler to keep the early returns and call `incrementMetrics('js', 'failure')` directly. No need for the observable either?

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -322,6 +323,7 @@ export function validateAndBuildRumConfiguration(
     customerDataTelemetrySampleRate: 1,
     initialViewMetricsTelemetrySampleRate: 1,
     replayTelemetrySampleRate: 1,
+    remoteConfigurationTelemetrySampleRate: 1,
```
BenoitZugmeyer: **suggestion:** it looks like managing different sample rates for different type of metrics brings a lot of complexity for something that is constant in practice. Could we:
* Use a single metric sample rate for all metrics
* No need to store it in the configuration, we can just use a constant in `telemetry.ts`
* Find a different solution to mock that constant for tests? ([example](https://github.com/DataDog/browser-sdk/blob/main/packages/rum/src/domain/segmentCollection/segmentCollection.ts#L191-L193))

cc  @sethfowler-datadog

sethfowler-datadog: Yeah, I think that'd be a great improvement. It'd definitely reduce some of the plumbing that's currently required to add new telemetry.

I'm wondering if we should just perform a _single_ draw for all metrics up front in `startTelemetry()`, and add a boolean `isSampledForMetrics` property to the `Telemetry` interface. Tests could then set the value of `isSampledForMetrics` directly, just as they do for `enabled`. I find this appealing both because it makes mocking simple and because having a consistent sampling decision for all metrics makes reasoning about the interaction between different metrics a bit easier. But there are lots of ways to approach the problem!

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -32,23 +40,52 @@ interface SupportedContextManagers {
   context: ReturnType<typeof createContextManager>
 }
 
+export interface RemoteConfigurationMetrics {
```
sethfowler-datadog: If you make this extend `Context`, you can get rid of the `as unknown as Context` type assertions elsewhere:
```suggestion
export interface RemoteConfigurationMetrics extends Context {
```

packages/core/src/domain/telemetry/telemetry.ts
bcaudan: I went with the other comment suggestion and mutualized the sample rate for all metrics, so we should not need to do that anymore.

packages/rum-core/src/domain/configuration/configuration.ts
bcaudan: Indeed, much simpler with a single metric sample rate and removing that from the config.
I sticked with one draw by metric to minimize the risk of reaching the number of telemetry messages threshold when multiple metrics are enabled at the same time.

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -164,7 +155,7 @@ export function applyRemoteConfiguration(
 
   function resolveCookieValue({ name }: { name: string }) {
     const value = getCookie(name)
-    resolveEventObservable.notify(['cookie', value !== undefined ? 'success' : 'missing'])
+    incrementMetrics(['cookie', value !== undefined ? 'success' : 'missing'])
```
BenoitZugmeyer: **suggestion:** just use two arguments?

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -32,22 +39,50 @@ interface SupportedContextManagers {
...
+
 export async function fetchAndApplyRemoteConfiguration(
   initConfiguration: RumInitConfiguration,
   supportedContextManagers: SupportedContextManagers
 ) {
+  let rumInitConfiguration: RumInitConfiguration | undefined
+  const metrics: RemoteConfigurationMetrics = { fetch: {} }
   const fetchResult = await fetchRemoteConfiguration(initConfiguration)
   if (!fetchResult.ok) {
+    metrics.fetch.failure = 1
```
BenoitZugmeyer: **suggestion:** Keep things uniform? `incrementMetrics('fetch', 'failure')`

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -64,30 +99,144 @@ export function applyRemoteConfiguration(
...
+      const domValue = attribute !== undefined ? element.getAttribute(attribute) : element.textContent
+      if (domValue === null) {
+        missing = true
+      } else {
+        value = domValue
+      }
+    }
+    incrementMetrics(['dom', failure ? 'failure' : missing ? 'missing' : 'success'])
+    return value
+  }
```
BenoitZugmeyer, line 191: **suggestion:** I still think using early returns would make this kind of code way more readable

```ts
  function resolveDomValue({ selector, attribute }: { selector: string; attribute?: string }) {
    let element: Element | null
    try {
      element = document.querySelector(selector)
    } catch {
      incrementMetrics('dom', 'failure')
      display.error(`Invalid selector in the remote configuration: '${selector}'`)
      return
    }
 
    if (!element) {
      incrementMetrics('dom', 'missing')
      return
    }
 
    if (isForbidden(element, attribute)) {
      incrementMetrics('dom', 'failure')
      display.error(`Forbidden element selected by the remote configuration: '${selector}'`)
      return
    }

    const domValue = attribute !== undefined ? element.getAttribute(attribute) : element.textContent
    if (domValue === null) {
      incrementMetrics('dom', 'missing')
      return
    }

    incrementMetrics('dom', 'success')
    return domValue
  }
```

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -64,30 +99,144 @@ export function applyRemoteConfiguration(
...
+          failure = true
+          display.error(`Error accessing: '${path}'`, e)
+          break
+        }
       }
     }
-    return mapValues(property, resolveConfigurationProperty)
+    incrementMetrics(['js', failure ? 'failure' : 'success'])
+    return !failure ? current : undefined
+  }
```
BenoitZugmeyer, line 220: **suggestion:** same here

```ts
  function resolveJsValue({ path }: { path: string }): unknown {
    const pathParts = parseJsonPath(path)
    if (pathParts.length === 0) {
      display.error(`Invalid JSON path in the remote configuration: '${path}'`)
      incrementMetrics('js', 'failure')
      return
    }

    let current = window as unknown as { [key: string]: unknown }
    for (const pathPart of pathParts) {
      if (!(pathPart in current)) {
        incrementMetrics('js', 'missing')
        return
      }
      try {
        current = current[pathPart] as { [key: string]: unknown }
      } catch (e) {
        incrementMetrics('js', 'failure')
        display.error(`Error accessing: '${path}'`, e)
        return
      }
    }
    incrementMetrics('js',  'success'])
    return current
  }
```

packages/core/src/domain/telemetry/telemetry.ts
```diff
@@ -109,10 +123,18 @@ export function startTelemetryCollection(
     [TelemetryType.USAGE]: telemetryEnabled && performDraw(configuration.telemetryUsageSampleRate),
   }
 
+  const telemetryEnabledPerMetrics: { [metricName: string]: boolean } = {}
+  Object.values(TelemetryMetrics).forEach((metricName) => {
+    telemetryEnabledPerMetrics[metricName] = telemetryEnabled && performDraw(metricSampleRate)
+  })
```
BenoitZugmeyer: **suggestion:** perform a single draw:

```
const enabledForMetrics = telemetryEnabled && performDraw(metricSampleRate)

...

return { enabled, enabledForMetrics }j
```

Also, maybe we can introduce a proper telemetry event type `metric`, so we can use `telemetryEnabledPerType[rawEvent.type!]` for it?

amortemousque (approved):

packages/core/src/domain/telemetry/telemetry.ts
bcaudan: as mentioned [here](https://github.com/DataDog/browser-sdk/pull/3779#discussion_r2358447621), I'm a bit concerned about easily reaching the telemetry threshold (for log or metric if we go with a dedicated type) when enabling all the metrics at the same time.
wdyt?

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
bcaudan: my concern was to limit the repetition of the constants to limit the size increase but the extra ~100 bytes should be fine

packages/rum-core/src/domain/configuration/configuration.ts
BenoitZugmeyer: Aah I see. How many metrics do we expect to have? What is a metric anyway? Should it be a temporary log like 'debug'? Or are we expecting to keep it for a longer time?

bcaudan: Looking at our current usage:

> How many metrics do we expect to have?

We have added quite a few lately, so I'd say probably a growing number

> What is a metric anyway? 

Telemetry data that we aggregate locally before sending 

> Should it be a temporary log like 'debug'? Or are we expecting to keep it for a longer time?

It can be one or the other depending on the need

bcaudan: As discussed together, the thresholds are already separated by metric so it should be fine to have a single draw.

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -71,7 +69,10 @@ describe('makeRecorderApi', () => {
         sessionManager ?? createRumSessionManagerMock().setId('1234'),
         mockViewHistory(),
         worker,
-        { enabled: true } as Telemetry
+        {
+          enabled: true,
+          metricsEnabled: true,
+        } as unknown as Telemetry
```
sethfowler-datadog: Nit: you probably don't need `as unknown` here.

packages/rum-core/src/domain/view/viewMetrics/startInitialViewMetricsTelemetry.spec.ts
```diff
@@ -43,26 +41,15 @@ describe('startInitialViewMetricsTelemetry', () => {
...
-    const { stop: stopInitialViewMetricsTelemetryCollection } = startInitialViewMetricsTelemetry(
-      configuration,
-      lifeCycle,
-      {
-        enabled: true,
-      } as Telemetry
-    )
+    const { stop: stopInitialViewMetricsTelemetryCollection } = startInitialViewMetricsTelemetry(lifeCycle, {
+      metricsEnabled,
+    } as unknown as Telemetry)
```
sethfowler-datadog: Nit: you may be able to remove this `as unknown`.

packages/rum/src/domain/segmentCollection/startSegmentTelemetry.spec.ts
```diff
@@ -50,13 +42,11 @@ describe('segmentTelemetry', () => {
...
 
-  function setupSegmentTelemetryCollection(partialConfig: Partial<RumConfiguration> = config) {
-    const configuration = mockRumConfiguration(partialConfig)
+  function setupSegmentTelemetryCollection(metricsEnabled: boolean = true) {
     requestObservable = new Observable()
     telemetry = startMockTelemetry()
     ;({ stop: stopSegmentTelemetry } = startSegmentTelemetry(
-      configuration,
-      { enabled: true } as Telemetry,
+      { metricsEnabled } as unknown as Telemetry,
```
sethfowler-datadog: Nit: same deal here.

If you have to use `{ enabled: true, metricsEnabled }` to avoid `as unknown` in this and similar cases, IMO that's probably worth it, since `as unknown` is completely losing type safety.

packages/rum/src/domain/startRecorderInitTelemetry.spec.ts
```diff
@@ -2,28 +2,18 @@ import type { Telemetry, RawTelemetryEvent } from '@datadog/browser-core'
...
-
-  function startRecorderInitTelemetryCollection(partialConfig: Partial<RumConfiguration> = config) {
-    const configuration = mockRumConfiguration(partialConfig)
+  function startRecorderInitTelemetryCollection(metricsEnabled: boolean = true) {
     observable = new Observable<RecorderInitEvent>()
     telemetry = startMockTelemetry()
     const { stop: stopRecorderInitTelemetry } = startRecorderInitTelemetry(
-      configuration,
-      { enabled: true } as Telemetry,
+      { metricsEnabled } as unknown as Telemetry,
```
sethfowler-datadog: Nit: here too.

BenoitZugmeyer (approved):

PR #3780, opened 2025-08-14
👷 add `consistent-type-specifier-style: 'prefer-top-level'` eslint rule

bcaudan:
## Motivation

When an inline type import `import { type foo } from bar` is exported by a package, it breaks the compatibility with older ts versions.

## Changes

Introduce an eslint rule to disallow the usage of inline type import in source files.

## Test instructions

Use an inline type import, it should be raised by the linter.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


eslint-local-rules/disallowInlineTypeImport.js
mormubis, line 1: **💭 thought:** ‏Is there any chance of an `autofix` for this rule?

mormubis (approved):
🙌

eslint-local-rules/disallowInlineTypeImport.js
bcaudan: by switching to eslint-plugin, yes 🙂

PR #3788, opened 2025-08-18
✨[PANA-3817] Add telemetry for initial view metrics

sethfowler-datadog:
## Motivation

We're currently investigating changes to the set of criteria that determines when session replay recording starts. We want to be sure that these changes do not cause any unintended performance impact, so we'd like to gather some timing metrics around page load before and after the change.

## Changes

This PR adds telemetry for the `NavigationTiming` and LCP components of the initial view metrics, sampled at 1%.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/startInitialViewMetricsTelemetry.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/startInitialViewMetricsTelemetry.ts
@@ -1,6 +1,7 @@
 import type { Context, Telemetry } from '@datadog/browser-core'
 import { performDraw, addTelemetryMetrics, noop } from '@datadog/browser-core'
-import { LifeCycleEventType, type LifeCycle } from '../../lifeCycle'
+import { LifeCycleEventType } from '../../lifeCycle'
+import type { LifeCycle } from '../../lifeCycle'
 import type { RumConfiguration } from '../../configuration'
 import type { LargestContentfulPaint } from './trackLargestContentfulPaint'
 import type { NavigationTimings } from './trackNavigationTimings'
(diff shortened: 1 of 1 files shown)


packages/rum-core/src/domain/view/viewMetrics/startInitialViewMetricsTelemetry.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const { unsubscribe } = lifeCycle.subscribe(LifeCycleEventType.VIEW_UPDATED, ({ initialViewMetrics }) => {
+    if (!initialViewMetrics.largestContentfulPaint || !initialViewMetrics.navigationTimings) {
+      return
+    }
+
+    addTelemetryMetrics(INITIAL_VIEW_METRICS_TELEMETRY_NAME, {
+      metrics: createCoreInitialViewMetrics(
+        initialViewMetrics.largestContentfulPaint,
```
amortemousque, line 50: **❓ question:** The `largestContentfulPaint` metric can be reported multiple times for the same view. With this implementation, you’d only capture the first entry, while the actual final value may come later. Would that create any issues for your telemetry?

sethfowler-datadog: For these metrics, the focus is on monitoring for regressions that might affect performance during early page load. The current implementation snapshots the LCP value as soon as the navigation timings are available (so, shortly after the load event) and at least one LCP value has been generated. So, the LCP value is not necessarily the first, but it's the value we'd see if we stopped observing LCP values shortly after the load event fires. Since we're focused on early page load, I think this is a good fit, but it's true that it might not be the best fit for every use case!

I'll add a comment to document this, so we're aware that this LCP value may not be the final one.

PR #3793, opened 2025-08-20
✨[PANA-3819] Add telemetry for recorder initialization

sethfowler-datadog:
## Motivation

We're investigating whether some changes to the initialization process for the session replay recorder could improve reliability and/or fidelity during early page load. To decide how best to approach the problem, and to evaluate any change we do make, it would be helpful to have some metrics about the recorder initialization process.

## Changes

This PR adds metrics for recorder initialization that are relevant to near-term arcs of work around session replay reliability and fidelity. The metrics (sampled at 1%) measure:
* How long it took to lazy-load the recorder module.
* How long it took to wait for the document to be ready. (Under the hood, this boils down to "how long did we wait for DOMContentLoaded?")
* How long recorder initialization took overall.
* Whether recorder initialization succeeded or failed, and if it failed, which failure code path was taken.
* Whether recorder startup was forced or automatic.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/configuration/configuration.ts
+++ b/packages/rum-core/src/domain/configuration/configuration.ts
@@ -248,6 +248,7 @@ export interface RumConfiguration extends Configuration {
   subdomain?: string
   customerDataTelemetrySampleRate: number
   initialViewMetricsTelemetrySampleRate: number
+  recorderInitTelemetrySampleRate: number
   segmentTelemetrySampleRate: number
   traceContextInjection: TraceContextInjection
   plugins: RumPlugin[]
@@ -321,6 +322,7 @@ export function validateAndBuildRumConfiguration(
     enablePrivacyForActionName: !!initConfiguration.enablePrivacyForActionName,
     customerDataTelemetrySampleRate: 1,
     initialViewMetricsTelemetrySampleRate: 1,
+    recorderInitTelemetrySampleRate: 1,
     segmentTelemetrySampleRate: 1,
     traceContextInjection: objectHasValue(TraceContextInjection, initConfiguration.traceContextInjection)
       ? initConfiguration.traceContextInjection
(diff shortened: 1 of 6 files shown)


packages/rum/src/boot/lazyLoadRecorder.spec.ts
```diff
@@ -43,22 +56,27 @@ describe('lazyLoadRecorder', () => {
     stopRecordingSpy = jasmine.createSpy('stopRecording')
     startRecordingSpy = jasmine.createSpy('startRecording')
 
-    // Workaround because using resolveTo(startRecordingSpy) was not working
-    loadRecorderSpy = jasmine.createSpy('loadRecorder').and.resolveTo((...args: any) => {
+    loadRecorderSpy = jasmine.createSpy('loadRecorder').and.callFake((...args) => {
```
sethfowler-datadog, line 59: The previous implementation creates a weird nested closure/promise thing that causes the promise rejection to happen after `recorderApi.start()` returns. This means that this code was not actually testing the failure path of `postStartStrategy` before. However, for the new assertions I've added, we do need to really make `postStartStrategy`'s attempt to start recording fail, so I had to fix this.

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -321,6 +322,7 @@ export function validateAndBuildRumConfiguration(
     enablePrivacyForActionName: !!initConfiguration.enablePrivacyForActionName,
     customerDataTelemetrySampleRate: 1,
     initialViewMetricsTelemetrySampleRate: 1,
+    recorderInitTelemetrySampleRate: 1,
```
amortemousque, line 325: Nit: Maybe a single sample rate could cover all telemetry required for Replay investigations?

amortemousque (approved):

packages/rum-core/src/domain/configuration/configuration.ts
sethfowler-datadog: Yeah, agreed; we're really starting to have a lot of these sample rate options! It makes sense to consolidate them. I'll open a followup PR to do that.

PR #3797, opened 2025-08-22
🔧 👷  convert webpack configuration files to TS, improve overall TS config

BenoitZugmeyer:
## Motivation

* More typescript coverage
* Improved typescript configuration, now the default `tsconfig.json` references other configs so we can typecheck the whole project with a signel `tsc -b . --noEmit` command

## Changes

* Convert wepack.config files to typescript
* Adjust configuration to support this change
* Adjust dev-server config imports

## Test instructions

CI should be enough, but you can run `yarn typecheck` to typecheck the project, and run `yarn tsc -b . --noEmit --listFiles` to list files that are included by typescript.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


performances/src/profilers/startNetworkProfiling.ts
```diff
@@ -27,7 +27,7 @@ export async function startNetworkProfiling(options: ProfilingOptions, client: C
   client.on('Network.loadingFinished', loadingFinishedListener)
   return () => {
     client.off('Network.requestWillBeSent', requestListener)
-    client.off('Network.loadingFinishedListener', loadingFinishedListener)
+    client.off('Network.loadingFinished', loadingFinishedListener)
```
BenoitZugmeyer, line 30: **note:** this was an actual bug caught by TS

developer-extension/webpack.config.ts
```diff
@@ -183,18 +192,16 @@ module.exports = (_env, argv) => {
...
       },
-
       plugins,
     }
   }
 }
 
 function getVersion() {
-  const version = require('./package.json').version
+  const version: string = JSON.parse(fs.readFileSync('./package.json', 'utf8')).version
```
mormubis: **❓ question:** ‏Can we import the json file at the top?

tsconfig.default.json
```diff
@@ -0,0 +1,33 @@
+// This tsconfig used as a default for typechecking files that are not included in any other
+// tsconfig.
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "types": ["jasmine"],
```
mormubis, line 6: **❓ question:** ‏Do we need `jasmine` types by default? Should we just include them in tests?

developer-extension/webpack.config.ts
BenoitZugmeyer: Done

tsconfig.default.json
BenoitZugmeyer: We could! we would need to create one more tsconfig file and tweak globs (to exclude spec files from default and include them in that new tsconfig). I think it's fine to keep it like this for now, as it's pretty minor.

PR #3801, opened 2025-08-26
🔊Add telemetry on consecutive short session renewals

amortemousque:
## Motivation

Collect insights on consecutive short session renewals

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/startSessionRenewTelemetry.ts
+++ b/packages/core/src/domain/session/startSessionRenewTelemetry.ts
@@ -1,0 +1,35 @@
+import { getCookie } from '../../browser/cookie'
+import type { TimeStamp } from '../../tools/utils/timeUtils'
+import { elapsed, ONE_SECOND, timeStampNow } from '../../tools/utils/timeUtils'
+import { addTelemetryDebug } from '../telemetry'
+import type { SessionState } from './sessionState'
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/session/startSessionRenewTelemetry.ts
```diff
@@ -0,0 +1,35 @@
...
+export function startSessionRenewTelemetry() {
+  let lastRenewTime = timeStampNow()
+  let fastRenewCount = 0
+
+  return {
+    onRenew(sessionCache: SessionState, sessionState: SessionState) {
+      const now = timeStampNow()
+      const interval = elapsed(lastRenewTime, now)
+      if (interval < FAST_RENEW_INTERVAL) {
+        fastRenewCount += 1
```
mormubis, line 19: Should we not reset the `fastRenewCount` if the renewal was correct?

```
|--|--------|--|
   ^           ^
 detected    ignored
```

amortemousque: I think being notified once per page load is enough. Wdyt?

mormubis: Sorry, I missed the comment. I'm ok with that, but for some customers like web apps, might hide the issue a bit.

PR #3803, opened 2025-08-26
[RUM-10415] [alt] add privacy allowlist support treewalker

cy-moi:
## Motivation
Mask action name with allowlists generated from rum privacy build plugin(WIP). This approach is purely client side and allowlist-based. We aim to mask all action names (custom & auto) OOTB with build time configuration using a [build plugin.](https://github.com/DataDog/build-plugins)

This PR is a diverge from https://github.com/DataDog/browser-sdk/pull/3751 to add remove automatically applying on action names.

## Changes
- Add allowlist processing helpers
- Mask all action names with the allowlist when privacy build plugin and configurations are opt-in

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -847,6 +867,66 @@ describe('serializeNodeWithId', () => {
...
+      })
+
+      afterEach(() => {
+        ;(window as BrowserWindow).$DD_ALLOW = undefined
+      })
+
+      it('obfuscates text content not in allowlist', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        // Most text should be masked with 'x' characters
+        expect(JSON.stringify(serializedDoc)).toContain('xxxxx')
```
sethfowler-datadog: It'd be nice to make the test more precise. Pseudocode:
```suggestion
        const textContents = getAllTextContents(serializedDoc)
        for (const textContent of textContents) {
          if (textContent.trim() && !allowlist.contains(textContent)) {
            expect(textContent).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
          }
        }
```

You could combine this with the "preserves text content in allowlist" test, as well, by adding a second branch to the `if`.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -847,6 +867,66 @@ describe('serializeNodeWithId', () => {
...
+      it('obfuscates input fields not in allowlist', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        expect(JSON.stringify(serializedDoc)).toContain('***')
+      })
+
+      it('obfuscates attributes and non-allowlisted text content', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        // Check that non-allowlisted private content is masked
+        expect(JSON.stringify(serializedDoc)).toContain('xxxx') // Some content should be masked
+      })
```
sethfowler-datadog: It'd be nice to make these more precise too, if we can. (I'm not sure if we have helpers similar to `getAllTextContents()` that would make it easy to inspect these things.)

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -847,6 +867,66 @@ describe('serializeNodeWithId', () => {
...
+
+      it('fails close when allowlist is empty', () => {
```
sethfowler-datadog: Nit:
```suggestion
      it('fails closed when allowlist is empty', () => {
```

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -847,6 +867,66 @@ describe('serializeNodeWithId', () => {
...
+        // All text content should be masked
+        const textContents = getAllTextContents(serializedDoc)
+        for (const textContent of textContents) {
+          if (textContent.trim()) {
+            expect(textContent).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
+          }
+        }
+      })
+
+      it('fails close when allowlist is undefined', () => {
```
sethfowler-datadog: Nit:
```suggestion
      it('fails closed when allowlist is undefined', () => {
```

packages/rum-core/src/domain/privacy.spec.ts
```diff
@@ -388,6 +427,45 @@ describe('shouldMaskNode', () => {
...
+
+  beforeEach(() => {
+    ;(window as BrowserWindow).$DD_ALLOW = new Set([TEST_STRINGS.PARAGRAPH_MIXED])
+  })
+
+  afterEach(() => {
+    ;(window as BrowserWindow).$DD_ALLOW = undefined
+  })
+
+  it('should fail close if $DD_ALLOW is not defined', () => {
```
sethfowler-datadog: Nit:
```suggestion
  it('should fail closed if $DD_ALLOW is not defined', () => {
```

packages/rum-core/src/domain/privacy.spec.ts
```diff
@@ -388,6 +427,45 @@ describe('shouldMaskNode', () => {
...
+    ;(window as BrowserWindow).$DD_ALLOW = undefined
+    const testString = maskDisallowedTextContent('mask-feature-on', ACTION_NAME_MASK)
+    expect(testString).toBe(ACTION_NAME_MASK)
+  })
+
+  it('masks words not in allowlist (with dictionary from $DD_ALLOW)', () => {
+    const testString1 = maskDisallowedTextContent('This is an action name in allowlist', ACTION_NAME_MASK)
```
sethfowler-datadog: I am a bit confused by this test case. Did you intend this to match the allowlist?

BenoitZugmeyer (approved):

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -257,14 +245,16 @@ function getTextualContent(
...
+  privacyEnabledActionName: boolean,
+  nodePrivacyLevel: NodePrivacyLevel
 ) {
   const walker = document.createTreeWalker(
     element,
     // eslint-disable-next-line no-bitwise
     NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_TEXT,
     rejectInvisibleOrMaskedElementsFilter
   )
+  const allowlistMaskEnabled = nodePrivacyLevel === NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED
```
sethfowler-datadog: This isn't quite right, unfortunately. We are using the privacy level of the target node here for all the descendant nodes, but that isn't how it should work; we should instead be using the privacy level of each individual node.

Ideally we'd handle it in exactly the same way that normal masking is handled: as part of `rejectInvisibleOrMaskedElementsFilter`. Indeed, you _are_ doing that, because you updated `shouldMaskNode()` to return `true` for `NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED`, and `rejectInvisibleOrMaskedElementsFilter` uses `shouldMaskNode()` to do its work.

That means that you _should_, in an ideal world, be able to revert all changes to `getTextualContentWithTreeWalker()`. You shouldn't need `allowlistMaskEnabled` since `rejectInvisibleOrMaskedElementsFilter` will skip `MASK_UNLESS_ALLOWLISTED` elements. (You'd also need to update the `} else if (nodePrivacyLevel === NodePrivacyLevel.MASK) {` check in `getActionNameFromElement()` to check for `MASK_UNLESS_ALLOWLISTED` as well.)

*BUT*: maintaining consistency here is kinda a bummer, because the current approach we use for `NodePrivacyLevel.MASK` is not great. By skipping all descendants of nodes with privacy level `MASK`, we make it impossible to apply `MASK` at a higher level and selectively override it. We'd be transferring that limitation to `MASK_UNLESS_ALLOWLISTED` if we maintain consistency. Not ideal considering we want people to adopt this feature, in most cases, by setting `defaultPrivacyLevel`, so we expect almost every node in the document to have privacy level `MASK_UNLESS_ALLOWLISTED`.

We're caught between a rock and a hard place here. Adopting the existing behavior of `MASK` is not good for the usability of this feature (and frankly not good for the usability of `MASK`), but it will be _really_ confusing if `MASK` and `MASK_UNLESS_ALLOWLISTED` have different inheritance rules.

The best solution may be to allow overriding for _both_ `MASK` and `MASK_UNLESS_ALLOWLISTED` when the tree walker is enabled, and gate the tree walker with a new value for `enablePrivacyForActionName`. We can then try to migrate everybody over to the new model in a future major. But we should discuss this; it's a bigger issue than what can be resolved in a PR review.

sethfowler-datadog: So, after some additional discussion, we concluded that we'll ship with a version that maintains consistency with the existing treatment of `NodePrivacyLevel.MASK`, and then solve the other issues as a separate task.

After some thought, I think that means that the handling of `NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED` _must_ be different than `NodePrivacyLevel.MASK`. That's because `MASK_UNLESS_ALLOWLISTED` just will not work right unless it's applied at the text node level.

Say you have this setup:
```
<div id="target" dd-privacy-level="mask-unless-allowlisted">
  <span id="a">a</span>
  <span id="b">b</span>
</div>
```

Say that both `a` and `b` are allowlisted strings.

What should `shouldMaskNode()` return for `#target`?

If we say "`shouldMaskNode()` should return `true`, since `#target` is `MASK_UNLESS_ALLOWLISTED`", then in the `enablePrivacyForActionName: true` model we will stop recursion there, and neither `a` or `b` will appear in the action name, even though they're both allowlisted.

If we say "`shouldMaskNode()` should return `true` if its text content is allowlisted", then the situation doesn't seem any better. `#target` not a text node, so it doesn't have its own text content. If we use its `innerText`, then we get `ab`, and `ab` is not allowlisted, so again we will stop recursion there and neither `a` nor `b` will appear in the action name.

The only reasonable answer is that `shouldMaskNode()` should return false for `#target`. This implies that we can't use the same strategy we use for today's version of `MASK`; there's no easy way for us to stop traversal when we see `MASK_UNLESS_ALLOWLISTED`, because we don't know whether it _actually_ means "mask" or "don't mask" until we go deeper in the tree. (And the answer may be different for different descendant elements.) `shouldMaskNode()` can only reasonably return `true` based on `MASK_UNLESS_ALLOWLISTED` for text nodes.

Handling this would be easier if you switched to a recursive implementation, as @BenoitZugmeyer suggested, but we should be able to make it work with the tree walker as well. (Though with less ideal performance.) I'll add some comments.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -282,7 +272,9 @@ function getTextualContentWithTreeWalker(
       continue // skip element nodes
     }
 
-    text += node.textContent || ''
+    text += allowlistMaskEnabled
+      ? maskDisallowedTextContent(node.textContent || '', ACTION_NAME_MASK)
+      : node.textContent || ''
```
sethfowler-datadog: You can revert this change, since we'll rely on `rejectInvisibleOrMaskedElementsFilter()` to exclude `MASK_UNLESS_ALLOWLISTED` elements that need to be masked.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
sethfowler-datadog: In `rejectInvisibleOrMaskedElementsfilter()`, you should switch from `getNodeSelfPrivacyLevel()` to `getNodePrivacyLevel()`. This is part that's not ideal from a performance perspective. You can reduce the impact by providing a `NodePrivacyLevelCache` argument, so that at least we won't compute the answer over and over for the same nodes.

You will also need to update `rejectInvisibleOrMaskedElementsFilter()` to call `getNodePrivacyLevel()` even for text nodes. (Right now text nodes are always accepted, but now we need to reject them sometimes.)

packages/rum-core/src/domain/privacy.ts
```diff
@@ -158,6 +133,7 @@ export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
     case NodePrivacyLevel.MASK:
     case NodePrivacyLevel.HIDDEN:
     case NodePrivacyLevel.IGNORE:
+    case NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED:
```
sethfowler-datadog: Instead of unconditionally returning true here, you should use an implementation like this:
```ts
case NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED:
  return isTextNode() ? !isAllowlisted(node.textContent) : false
```

cy-moi: I don't think this works with Session Replay `serializeAttributes` which censors attribtutes depending on the `element` privacy level. So I moved back this logic inside `rejectInvisibleOrMaskedElementsfilter()` again.

cy-moi: Per discussion on slack, this is fixed by handling form elements like mask-user-input does (cr. @sethfowler-datadog 's solution)

packages/rum-core/src/domain/privacy.ts
```diff
@@ -133,6 +104,10 @@ export function getNodeSelfPrivacyLevel(node: Node): NodePrivacyLevel | undefine
     return NodePrivacyLevel.MASK_USER_INPUT
   }
 
+  if (node.matches(getPrivacySelector(NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED))) {
+    return NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED
+  }
+
```
sethfowler-datadog, line 106: `MASK_UNLESS_ALLOWLISTED` is more restrictive than `MASK_USER_INPUT`, so you should check for `MASK_UNLESS_ALLOWLISTED` first.

packages/rum-core/src/domain/privacy.ts
```diff
@@ -306,6 +292,22 @@ export function shouldIgnoreElement(element: Element): boolean {
...
+export function isAllowlisted(text: string): boolean {
+  return (window as BrowserWindow).$DD_ALLOW?.has(text.toLocaleLowerCase()) || false
+}
+
+export function maskDisallowedTextContent(text: string, fixedMask?: string): string {
+  if (!text.trim()) {
+    return text
+  }
+  // We are using toLocaleLowerCase when adding to the allowlist to avoid case sensitivity
+  // so we need to do the same here
```
sethfowler-datadog: Nit: this comment probably belongs in `isAllowlisted()`.

packages/rum-core/src/domain/privacy.spec.ts
```diff
@@ -383,11 +422,46 @@ describe('shouldMaskNode', () => {
       expect(shouldMaskNode(element, NodePrivacyLevel.MASK_USER_INPUT)).toBeFalse()
     })
 
-    it('returns true if the privacy level is not ALLOW nor MASK_USER_INPUT', () => {
+    it('returns true if the privacy level is not ALLOW nor MASK_USER_INPUT or MASK_UNLESS_ALLOWLISTED', () => {
```
sethfowler-datadog: This test no longer checks `MASK_UNLESS_ALLOWLISTED` (which is correct, since `shouldMaskNode()` doesn't necessarily return `true` for `MASK_UNLESS_ALLOWLISTED` now!) but the test name still includes `MASK_UNLESS_ALLOWLISTED`. So, you'll want to revert this change to the test name.

In addition, you should add some new tests in this `describe()` block that cover the different cases for `MASK_UNLESS_ALLOWLISTED`. Off the top of my head, you should check:
1. Text nodes which have allowlisted `textContent`.
2. Text nodes with whitespace-only `textContent`.
3. Text nodes which have non-allowlisted, non-whitespace `textContent`.
4. Non-text nodes (e.g. normal HTML elements).

packages/rum-core/src/domain/privacy.spec.ts
```diff
@@ -383,11 +422,46 @@ describe('shouldMaskNode', () => {
...
+    ;(window as BrowserWindow).$DD_ALLOW = undefined
+  })
+
+  it('should fail closed if $DD_ALLOW is not defined', () => {
+    ;(window as BrowserWindow).$DD_ALLOW = undefined
+    const testString = maskDisallowedTextContent('mask-feature-on', ACTION_NAME_MASK)
+    expect(testString).toBe(ACTION_NAME_MASK)
+  })
+
+  it('masks words not in allowlist (with dictionary from $DD_ALLOW)', () => {
```
sethfowler-datadog: Seems like it'd be good to include a test for text content that _is_ in the allowlist. (I might avoid the term "words", since we no longer split in the input into words, as well.)

packages/rum-core/src/domain/privacy.spec.ts
```diff
@@ -383,11 +422,46 @@ describe('shouldMaskNode', () => {
...
+    const testString = maskDisallowedTextContent('any unallowed string', ACTION_NAME_MASK)
+    expect(testString).toBe(ACTION_NAME_MASK)
+  })
+
+  it('handles empty string', () => {
```
sethfowler-datadog, line 526: Probably worth checking a non-empty whitespace-only string too. (e.g. `'   '`)

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,74 @@ describe('serializeNodeWithId', () => {
...
+        ;(window as BrowserWindow).$DD_ALLOW = undefined
+      })
+
+      it('obfuscates text content not in allowlist', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        const textContents = getAllTextContents(serializedDoc)
+        for (const textContent of textContents) {
+          if (textContent.trim() && !isAllowlisted(textContent)) {
+            expect(textContent).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
+          }
```
sethfowler-datadog: Two small nits:
1. You can simplify this slightly, since `isAllowlisted()` checks `textContent.trim()` internally already; no need to check it here.
2. You should have an else branch that checks that the text content is _not_ masked when it's allowlisted.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,74 @@ describe('serializeNodeWithId', () => {
...
+      it('obfuscates input fields not in allowlist', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        expect(JSON.stringify(serializedDoc)).toContain('***')
+      })
+
+      it('obfuscates attributes and non-allowlisted text content', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        const textContents = getAllTextContents(serializedDoc)
+        for (const textContent of textContents) {
+          if (textContent.trim() && !isAllowlisted(textContent)) {
```
sethfowler-datadog: Both of the above nits apply here too.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -288,12 +279,15 @@ function getTextualContentWithTreeWalker(
   return text.replace(/\s+/g, ' ').trim()
 
   function rejectInvisibleOrMaskedElementsFilter(node: Node) {
+    const nodeSelfPrivacyLevel = getNodePrivacyLevel(node, nodePrivacyLevel, nodePrivacyLevelCache)
```
sethfowler-datadog: Instead of passing in `nodePrivacyLevel` for the second argument of `getNodePrivacyLevel()`, you should be passing in the `defaultPrivacyLevel`. You can just replace the `nodePrivacyLevel` parameter to `getTextualContentWithTreeWalker()` with a `defaultPrivacyLevel` parameter, and then do:
```suggestion
    const nodeSelfPrivacyLevel = getNodePrivacyLevel(node, defaultPrivacyLevel, nodePrivacyLevelCache)
```
When calling `getTextualContentWithTreeWalker()`, you can get the default privacy level from the `RumConfiguration` via `configuration.defaultPrivacyLevel`.

(It'd also be fine to just pass the `RumConfiguration` directly to `getTextualContentWithTreeWalker()`, if you prefer.)

I think this means that you can also remove the `nodePrivacyLevel` parameter from a bunch of calling functions, but double check me on that.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -288,12 +279,15 @@ function getTextualContentWithTreeWalker(
...
+    const shouldRejectMaskedNode =
+      privacyEnabledActionName && nodeSelfPrivacyLevel && shouldMaskNode(node, nodeSelfPrivacyLevel)
```
sethfowler-datadog: It seems like it would be less complicated to just early return here:
```suggestion
    if (privacyEnabledActionName && nodeSelfPrivacyLevel && shouldMaskNode(node, nodeSelfPrivacyLevel)) {
      return NodeFilter.FILTER_REJECT
    }
```

packages/rum-core/src/domain/action/trackClickActions.spec.ts
```diff
@@ -261,6 +264,10 @@ describe('trackClickActions', () => {
   })
 
   describe('with enablePrivacyForActionName false', () => {
+    beforeAll(() => {
+      ;(window as BrowserWindow).$DD_ALLOW = undefined
```
sethfowler-datadog: Does this affect this block of tests?

packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
```diff
@@ -504,6 +505,170 @@ describe('getActionNameFromElement', () => {
...
+          expectedName: 'foo bar baz',
+          expectedNameSource: 'text_content',
+        },
+      ]
+      testCases.forEach(
+        ({ html, defaultPrivacyLevel, nodePrivacyLevel, allowlist, expectedName, expectedNameSource }) => {
+          mockExperimentalFeatures([ExperimentalFeature.USE_TREE_WALKER_FOR_ACTION_NAME])
+          ;(window as BrowserWindow).$DD_ALLOW = new Set(allowlist)
+          const { name, nameSource } = getActionNameFromElement(
+            appendElement(html),
```
sethfowler-datadog: These are excellent tests; thank you for adding them!

One tweak would improve them a bit: instead of specifying `nodePrivacyLevel` as part of the test definition _and_ within the HTML, please remove the `nodePrivacyLevel` option from the test configuration and determine the right value by looking at the result of `appendElement()`. You can probably just call `getNodeSelfPrivacyLevel()` on it. That will ensure that, as these tests as maintained over time, we don't end up with a situation where the `nodePrivacyLevel` option gets out of sync with the HTML, possibly resulting in an incorrect test that could hide a bug.

packages/rum-core/src/domain/privacy.spec.ts
cy-moi: The test name is `return true if ... is not ...`, so I assumed we should add `MASK_UNLESS_ALLOWLISTED` there?

sethfowler-datadog: Ah yes, you're absolutely right; the `is not` threw me off! Current test name LGTM in that case.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,78 @@ describe('serializeNodeWithId', () => {
...
+
+      afterEach(() => {
+        ;(window as BrowserWindow).$DD_ALLOW = undefined
+      })
+
+      it('obfuscates text content not in allowlist', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        const textContents = getAllTextContents(serializedDoc)
+        for (const textContent of textContents) {
+          if (textContent && isAllowlisted(textContent)) {
```
sethfowler-datadog: Does this fail if you don't include the `textContent &&` part here? If so, don't we have a bug? `isAllowlisted()` should be treating all whitespace-only strings as allowlisted. So, I think this is all you need:
```suggestion
          if (isAllowlisted(textContent)) {
```
If the test fails without the `textContent &&` part, it'd be good to understand why.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,78 @@ describe('serializeNodeWithId', () => {
...
+      })
+
+      it('obfuscates attributes and non-allowlisted text content', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        const textContents = getAllTextContents(serializedDoc)
+        for (const textContent of textContents) {
+          if (textContent.trim() && !isAllowlisted(textContent)) {
+            expect(textContent).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
+          }
+        }
```
sethfowler-datadog: You marked this one as resolved but it seems like you didn't change the test... 😃 

I'm realizing now that this test doesn't test the thing it says it's testing anyway. The test name suggests this test is a test about _both_ attributes and text content, but the test expectations involve only text content. It seems to me that you have text content pretty well covered by the other tests, so maybe remove text content from the test name; let's make this test focus on attributes and ensure that we have test coverage for them.

So, first, we need the body of the test to be like this:
```suggestion
        const attributeValues = getAllAttributeValues(serializedDoc)
        for (const attributeValue of attributeValues) {
          if (isAllowlisted(attributeValue)) {
            expect(attributeValue).not.toEqual(jasmine.stringMatching(/^[x\s*]*$/))
          } else {
            expect(attributeValue).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
          }
        }
```

Then you just need to implement `getAllAttributeValues()`, following the pattern established by `getAllTextContents()`.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,78 @@ describe('serializeNodeWithId', () => {
...
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+
+        // All text content should be masked
+        const textContents = getAllTextContents(serializedDoc)
+        for (const textContent of textContents) {
+          if (textContent && isAllowlisted(textContent)) {
+            expect(textContent).not.toEqual(jasmine.stringMatching(/^[x*]*$/))
+          } else {
+            expect(textContent).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
+          }
```
sethfowler-datadog: Nit: I think here you can actually just assert that the text content is masked; nothing should be allowlisted in this test case, right?
```suggestion
          expect(textContent).toEqual(jasmine.stringMatching(/^[x\s*]*$/))
```

packages/rum-core/src/domain/privacy.spec.ts
```diff
@@ -383,11 +422,114 @@ describe('shouldMaskNode', () => {
...
+      })
+
+      describe('for text nodes', () => {
+        let textNode: Text
+
+        beforeEach(() => {
+          textNode = document.createTextNode('')
+        })
+
+        it('returns false for allowlisted text content', () => {
```
sethfowler-datadog, line 445: Nice, having these thorough tests will give us a lot of peace of mind down the road. Thank you!

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -73,15 +58,15 @@ function getActionNameFromElementProgrammatically(targetElement: Element, progra
 
 type NameStrategy = (
   element: Element | HTMLElement | HTMLInputElement | HTMLSelectElement,
-  userProgrammaticAttribute: string | undefined,
-  privacyEnabledActionName: boolean
+  rumConfiguration: RumConfiguration,
+  nodePrivacyLevel: NodePrivacyLevel
```
sethfowler-datadog: Nit: It's a bit tricky to tell from the diff, but it _seems_ like none of the strategies use the `nodePrivacyLevel` parameter defined in this type anymore, so you can probably just delete this line. (But only if that's true!)
```suggestion
```
You'll also need to remove the `nodePrivacyLevel` parameter from `getActionNameFromElementForStrategies()`.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,76 @@ describe('serializeNodeWithId', () => {
...
+    describe('for privacy tag `mask-unless-allowlisted`, a DOM tree', () => {
+      beforeEach(() => {
+        ;(window as BrowserWindow).$DD_ALLOW = new Set(['private title', 'hello private world'])
+      })
+
+      afterEach(() => {
+        ;(window as BrowserWindow).$DD_ALLOW = undefined
+      })
+
+      it('obfuscates text content not in allowlist', () => {
```
sethfowler-datadog, line 887: So now this test checks attributes, but the test name says `obfuscates text content not in allowlist`...

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -854,6 +875,76 @@ describe('serializeNodeWithId', () => {
...
+        expect(JSON.stringify(serializedDoc)).toContain('private title')
+        expect(JSON.stringify(serializedDoc)).toContain('hello private world')
+      })
+
+      it('obfuscates input fields not in allowlist', () => {
+        const serializedDoc = generateLeanSerializedDoc(HTML, 'mask-unless-allowlisted')
+        expect(JSON.stringify(serializedDoc)).toContain('***')
+      })
+
+      it('obfuscates attributes and non-allowlisted text content', () => {
```
sethfowler-datadog, line 911: ...while this one checks text contents, but the test name says `obfuscates attributes and non-allowlisted text content`.

So the names are backwards. But the actual tests look good! Please swap the bodies of these two tests, and then I think we're good to go.

PR #3804, opened 2025-08-27
✨ [RUM-11398] Add feature operation step vital APIs

cy-moi:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -1,42 +1,60 @@
...
   context?: any
   description?: string
 }
-
-export interface AddDurationVitalOptions {
+export type DurationVitalOptions = VitalOptions
+export interface FeatureOperationOptions extends VitalOptions {
+  operationKey?: string
+}
+export interface FullFeatureOperationOptions extends FeatureOperationOptions {
```
BenoitZugmeyer: **suggestion:** Rename `FullFeatureOperationOptions` to `FailFeatureOperationOptions`

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -771,6 +801,21 @@ export function makeRumPublicApi(
...
+
+    succeedFeatureOperation: monitor((name, options) => {
+      addTelemetryUsage({ feature: 'add-operation-step-vital', action_type: 'succeed' })
+      strategy.addOperationStepVital(name, 'end', options)
+    }),
+
+    failFeatureOperation: monitor((name, options) => {
+      addTelemetryUsage({ feature: 'add-operation-step-vital', action_type: 'fail' })
+      strategy.addOperationStepVital(name, 'end', options)
+    }),
```
BenoitZugmeyer: **question:** how do you distinguish between `succeed` and `fail` if the user is not providing a `failureReason`?

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -148,6 +149,16 @@ export function createPreStartStrategy(
     bufferApiCalls.add((startRumResult) => startRumResult.addDurationVital(vital))
   }
 
+  const addOperationStepVital = (
+    name: string,
+    stepType: 'start' | 'end',
+    options?: Partial<FullFeatureOperationOptions>
```
BenoitZugmeyer: **suggestion:** 
```suggestion
    options?: FeatureOperationOptions | FailFeatureOperationOptions
```

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -65,7 +83,32 @@ export function startVitalCollection(
     }
   }
 
+  function addOperationStepVital(
+    name: string,
+    stepType: 'start' | 'end',
+    options?: Partial<FullFeatureOperationOptions>
```
BenoitZugmeyer: **suggestion:** 
```suggestion
    options?: FeatureOperationOptions | FailFeatureOperationOptions
```

packages/rum-core/src/boot/rumPublicApi.ts
cy-moi: Indeed, `failureReason` should be required.

cy-moi: I made the type changes (mentioned in comments below) based on this as well.

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -424,9 +426,36 @@ export interface RumPublicApi extends PublicApi {
...
+   * @param options - Options for the operation step (operationKey, context, description)
+   */
+  succeedFeatureOperation: (name: string, options?: FeatureOperationOptions) => void
+
+  /**
+   * [Experimental] fail a feature operation
+   *
+   * @category Vital
+   * @param name - Name of the operation step
+   * @param options - Options for the operation step (operationKey, failureReason, context, description)
```
BenoitZugmeyer: **nitpick:** update the doc

```suggestion
   * @param failureReason
   * @param options - Options for the operation step (operationKey, context, description)
```

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -135,32 +177,35 @@ function buildDurationVital(
...
   return {
-    rawRumEvent,
-    startTime: vital.startClocks.relative,
-    duration: vital.duration,
+    rawRumEvent: {
+      date: startClocks.timeStamp,
+      vital: vitalData,
+      type: RumEventType.VITAL,
+      context,
+      ...(valueComputedBySdk && { _dd: { vital: { computed_value: true } } }),
```
BenoitZugmeyer: **nitpick:** I'm not sure what `valueComputedBySdk` is, but it looks like it's always `true`, so maybe we can simplify?

cy-moi: Ah wait, this one should be false when customers use `addDurationVital` and compute their own values right? I think we can either remove this or we change how `start` or `stop` calling `addDurationVital` under the hood to pass in `valueComputedBySdk`.

cy-moi: Per discussion on slack, we are removing this value because it was only relevant to the pre-GA vital and is no longer in use.

BenoitZugmeyer (approved):

PR #3805, opened 2025-08-28
✨ [RUM-11434] GraphQL support

rgaignault:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/configuration/configuration.ts
+++ b/packages/rum-core/src/domain/configuration/configuration.ts
@@ -220,11 +220,23 @@ export interface RumInitConfiguration extends InitConfiguration {
    * @defaultValue 0
    */
   profilingSampleRate?: number | undefined
+
+  /**
+   * A list of GraphQL endpoint URLs to track and enrich with GraphQL-specific metadata.
+   *
+   * @category Data Collection
+   */
+  allowedGraphQlUrls?: Array<MatchOption | GraphQlUrlOption> | undefined
 }
 
 export type HybridInitConfiguration = Omit<RumInitConfiguration, 'applicationId' | 'clientToken'>
 
 export type FeatureFlagsForEvents = 'vital' | 'action' | 'long_task' | 'resource'
+
+export interface GraphQlUrlOption {
+  match: MatchOption
+  trackPayload?: boolean
+}
 
 export interface RumConfiguration extends Configuration {
   // Built from init configuration
@@ -254,6 +266,7 @@ export interface RumConfiguration extends Configuration {
   trackFeatureFlagsForEvents: FeatureFlagsForEvents[]
   profilingSampleRate: number
   propagateTraceBaggage: boolean
+  allowedGraphQlUrls: GraphQlUrlOption[]
 }
 
 export function validateAndBuildRumConfiguration(
@@ -287,6 +300,8 @@ export function validateAndBuildRumConfiguration(
   if (!allowedTracingUrls) {
     return
   }
+
+  const allowedGraphQlUrls = validateAndBuildGraphQlOptions(initConfiguration)
 
   const baseConfiguration = validateAndBuildConfiguration(initConfiguration)
   if (!baseConfiguration) {
@@ -329,6 +344,7 @@ export function validateAndBuildRumConfiguration(
     trackFeatureFlagsForEvents: initConfiguration.trackFeatureFlagsForEvents || [],
     profilingSampleRate: initConfiguration.profilingSampleRate ?? 0,
     propagateTraceBaggage: !!initConfiguration.propagateTraceBaggage,
+    allowedGraphQlUrls,
     ...baseConfiguration,
   }
 }
@@ -384,6 +400,36 @@ function getSelectedTracingPropagators(configuration: RumInitConfiguration): Pro
   }
 
   return Array.from(usedTracingPropagators)
+}
+
+/**
+ * Build GraphQL options from configuration
+ * More lenient than tracing URLs: no service dependency, always returns array
+ */
+function validateAndBuildGraphQlOptions(initConfiguration: RumInitConfiguration): GraphQlUrlOption[] {
+  if (!initConfiguration.allowedGraphQlUrls) {
+    return []
+  }
+
+  if (!Array.isArray(initConfiguration.allowedGraphQlUrls)) {
+    display.warn('allowedGraphQlUrls should be an array')
+    return []
+  }
+
+  const graphQlOptions: GraphQlUrlOption[] = []
+
+  initConfiguration.allowedGraphQlUrls.forEach((option) => {
+    if (isMatchOption(option)) {
+      graphQlOptions.push({ match: option, trackPayload: false })
+    } else if (option && typeof option === 'object' && 'match' in option && isMatchOption(option.match)) {
+      graphQlOptions.push({
+        match: option.match,
+        trackPayload: option.trackPayload ?? false,
+      })
+    }
+  })
+
+  return graphQlOptions
 }
 
 export function serializeRumConfiguration(configuration: RumInitConfiguration) {
--- a/packages/rum-core/src/domain/resource/graphql.ts
+++ b/packages/rum-core/src/domain/resource/graphql.ts
@@ -1,0 +1,74 @@
+import { matchList, ONE_KIBI_BYTE } from '@datadog/browser-core'
+import type { RumConfiguration, GraphQlUrlOption } from '../configuration'
+
+const GRAPHQL_PAYLOAD_LIMIT = 32 * ONE_KIBI_BYTE
+
+export interface GraphQlMetadata {
+  operationType: 'query' | 'mutation' | 'subscription'
+  operationName?: string
+  variables?: string
+  payload?: string
+}
+
+export function isGraphQlRequest(url: string, configuration: RumConfiguration): GraphQlUrlOption | false {
+  for (const graphQlOption of configuration.allowedGraphQlUrls) {
+    if (matchList([graphQlOption.match], url)) {
+      return graphQlOption
+    }
+  }
+  return false
+}
+
+export function extractGraphQlMetadata(requestBody: string | undefined): GraphQlMetadata | undefined {
+  if (!requestBody || typeof requestBody !== 'string') {
+    return undefined
+  }
+
+  let graphqlBody: { query?: string; operationName?: string; variables?: string } | undefined
+
+  try {
+    graphqlBody = JSON.parse(requestBody)
+  } catch {
+    // Not valid JSON
+    return undefined
+  }
+
+  if (!graphqlBody || !graphqlBody.query) {
+    return undefined
+  }
+
+  const query = graphqlBody.query.trim()
+  const operationType = getOperationType(query)
+  const operationName = graphqlBody.operationName
+
+  let variables: string | undefined
+  if (graphqlBody.variables !== null && graphqlBody.variables !== undefined) {
+    variables = JSON.stringify(graphqlBody.variables)
+  }
+
+  return {
+    operationType,
+    operationName,
+    variables,
+    payload: truncatePayload(query, GRAPHQL_PAYLOAD_LIMIT),
+  }
+}
+
+function getOperationType(query: string): 'query' | 'mutation' | 'subscription' {
+  const trimmedQuery = query.trim()
+
+  if (trimmedQuery.startsWith('mutation')) {
+    return 'mutation'
+  } else if (trimmedQuery.startsWith('subscription')) {
+    return 'subscription'
+  }
+
+  return 'query'
+}
+
+function truncatePayload(payload: string, limit: number): string {
+  if (payload.length > limit) {
+    return `${payload.substring(0, limit)}...`
+  }
+  return payload
+}
--- a/packages/rum-core/src/domain/resource/resourceCollection.ts
+++ b/packages/rum-core/src/domain/resource/resourceCollection.ts
@@ -40,6 +40,7 @@ import {
 import { retrieveInitialDocumentResourceTiming } from './retrieveInitialDocumentResourceTiming'
 import type { RequestRegistry } from './requestRegistry'
 import { createRequestRegistry } from './requestRegistry'
+import { isGraphQlRequest, extractGraphQlMetadata } from './graphql'
 
 export function startResourceCollection(
   lifeCycle: LifeCycle,
@@ -133,6 +134,31 @@ function assembleResource(
     ? computeResourceEntryDuration(entry)
     : computeRequestDuration(pageStateHistory, startClocks, request!.duration)
 
+  // Check if this is a GraphQL request and extract metadata
+  const graphqlData =
+    request && request.type === RequestType.FETCH
+      ? (() => {
+          const graphQlConfig = isGraphQlRequest(request.url, configuration)
+          if (graphQlConfig) {
+            const requestBody = request.init?.body
+            if (requestBody && typeof requestBody === 'string') {
+              const graphqlMetadata = extractGraphQlMetadata(requestBody)
+              if (graphqlMetadata) {
+                return {
+                  graphql: {
+                    operationType: graphqlMetadata.operationType,
+                    operationName: graphqlMetadata.operationName,
+                    variables: graphqlMetadata.variables,
+                    payload: graphQlConfig.trackPayload === true ? graphqlMetadata.payload : undefined,
+                  },
+                }
+              }
+            }
+          }
+          return {}
+        })()
+      : {}
+
   const resourceEvent = combine(
     {
       date: startClocks.timeStamp,
@@ -150,6 +176,7 @@ function assembleResource(
         url: request ? sanitizeIfLongDataUrl(request.url) : entry!.name,
         protocol: entry && computeResourceEntryProtocol(entry),
         delivery_type: entry && computeResourceEntryDeliveryType(entry),
+        ...graphqlData,
       },
       type: RumEventType.RESOURCE,
       _dd: {
--- a/packages/rum-core/src/domain/configuration/configuration.spec.ts
+++ b/packages/rum-core/src/domain/configuration/configuration.spec.ts
@@ -511,6 +511,60 @@ describe('validateAndBuildRumConfiguration', () => {
       expect(displayWarnSpy).toHaveBeenCalledOnceWith('trackFeatureFlagsForEvents should be an array')
     })
   })
+
+  describe('allowedGraphQlUrls', () => {
+    it('defaults to an empty array', () => {
+      const configuration = validateAndBuildRumConfiguration(DEFAULT_INIT_CONFIGURATION)!
+      expect(configuration.allowedGraphQlUrls).toEqual([])
+    })
+
+    it('should accept string URLs', () => {
+      const configuration = validateAndBuildRumConfiguration({
+        ...DEFAULT_INIT_CONFIGURATION,
+        allowedGraphQlUrls: ['https://api.example.com/graphql', '/graphql'],
+      })!
+      expect(configuration.allowedGraphQlUrls).toEqual([
+        { match: 'https://api.example.com/graphql', trackPayload: false },
+        { match: '/graphql', trackPayload: false },
+      ])
+    })
+
+    it('should accept MatchOption objects', () => {
+      const configuration = validateAndBuildRumConfiguration({
+        ...DEFAULT_INIT_CONFIGURATION,
+        allowedGraphQlUrls: [{ match: /\/graphql$/i }, { match: 'https://api.example.com/graphql' }],
+      })!
+      expect(configuration.allowedGraphQlUrls).toEqual([
+        { match: /\/graphql$/i, trackPayload: false },
+        { match: 'https://api.example.com/graphql', trackPayload: false },
+      ])
+    })
+
+    it('should accept function matchers', () => {
+      const customMatcher = (url: string) => url.includes('graphql')
+      const configuration = validateAndBuildRumConfiguration({
+        ...DEFAULT_INIT_CONFIGURATION,
+        allowedGraphQlUrls: [{ match: customMatcher }],
+      })!
+      expect(configuration.allowedGraphQlUrls).toEqual([{ match: customMatcher, trackPayload: false }])
+    })
+
+    it('should accept GraphQL options with trackPayload', () => {
+      const configuration = validateAndBuildRumConfiguration({
+        ...DEFAULT_INIT_CONFIGURATION,
+        allowedGraphQlUrls: [{ match: '/graphql', trackPayload: true }],
+      })!
+      expect(configuration.allowedGraphQlUrls).toEqual([{ match: '/graphql', trackPayload: true }])
+    })
+
+    it('should reject invalid values', () => {
+      validateAndBuildRumConfiguration({
+        ...DEFAULT_INIT_CONFIGURATION,
+        allowedGraphQlUrls: 'not-an-array' as any,
+      })
+      expect(displayWarnSpy).toHaveBeenCalledOnceWith('allowedGraphQlUrls should be an array')
+    })
+  })
 })
 
 describe('serializeRumConfiguration', () => {
--- a/packages/rum-core/src/domain/resource/graphql.spec.ts
+++ b/packages/rum-core/src/domain/resource/graphql.spec.ts
@@ -1,0 +1,127 @@
+import { mockRumConfiguration } from '../../../test'
+import { extractGraphQlMetadata, isGraphQlRequest } from './graphql'
+
+describe('GraphQL detection and metadata extraction', () => {
+  describe('isGraphQlRequest', () => {
+    it('should detect GraphQL requests matching string URLs', () => {
+      const configuration = mockRumConfiguration({
+        allowedGraphQlUrls: [
+          { match: 'http://localhost/graphql', trackPayload: false },
+          { match: 'https://api.example.com/graphql', trackPayload: false },
+        ],
+      })
+
+      expect(isGraphQlRequest('https://api.example.com/graphql', configuration)).toBeTruthy()
+      expect(isGraphQlRequest('http://localhost/api', configuration)).toBe(false)
+    })
+
+    it('should detect GraphQL requests matching regex patterns', () => {
+      const configuration = mockRumConfiguration({
+        allowedGraphQlUrls: [{ match: /\/graphql$/i, trackPayload: false }],
+      })
+
+      expect(isGraphQlRequest('/api/graphql', configuration)).toBeTruthy()
+      expect(isGraphQlRequest('/graphql/admin', configuration)).toBe(false)
+    })
+
+    it('should detect GraphQL requests matching function matchers', () => {
+      const configuration = mockRumConfiguration({
+        allowedGraphQlUrls: [{ match: (url: string) => url.includes('gql'), trackPayload: false }],
+      })
+
+      expect(isGraphQlRequest('/api/gql', configuration)).toBeTruthy()
+      expect(isGraphQlRequest('/gql-endpoint', configuration)).toBeTruthy()
+      expect(isGraphQlRequest('/api/rest', configuration)).toBe(false)
+    })
+  })
+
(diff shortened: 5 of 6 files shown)


packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -386,6 +402,35 @@ function getSelectedTracingPropagators(configuration: RumInitConfiguration): Pro
...
+
+  const graphQlOptions: GraphQlUrlOption[] = []
+
+  initConfiguration.allowedGraphQlUrls.forEach((option) => {
+    if (isMatchOption(option)) {
+      graphQlOptions.push({ match: option, trackPayload: false })
+    } else if (option && typeof option === 'object' && 'match' in option && isMatchOption(option.match)) {
+      graphQlOptions.push({
+        match: option.match,
+        trackPayload: option.trackPayload ?? false,
```
bcaudan: **💬 suggestion:** ‏it could be nice to ensure that we end up with a boolean
```suggestion
        trackPayload: !!option.trackPayload,
```

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,74 @@
+import { matchList, ONE_KIBI_BYTE } from '@datadog/browser-core'
+import type { RumConfiguration, GraphQlUrlOption } from '../configuration'
+
+const GRAPHQL_PAYLOAD_LIMIT = 32 * ONE_KIBI_BYTE
```
bcaudan, line 7: **❓ question:** ‏Could you add a comment with the reasoning for this value?

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -133,6 +134,31 @@ function assembleResource(
...
+                    variables: graphqlMetadata.variables,
+                    payload: graphQlConfig.trackPayload === true ? graphqlMetadata.payload : undefined,
+                  },
+                }
+              }
+            }
+          }
+          return {}
+        })()
+      : {}
```
bcaudan: **💬 suggestion:** ‏what about extracting a `computeGraphQlData` for all the graphql logic?
using iife here also feels quite weird

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export interface GraphQlMetadata {
+  operationType: 'query' | 'mutation' | 'subscription'
+  operationName?: string
+  variables?: string
+  payload?: string
+}
+
+export function isGraphQlRequest(url: string, configuration: RumConfiguration): GraphQlUrlOption | false {
```
bcaudan: **💬 suggestion:** having a function returning multiple types (`GraphQlUrlOption | false`) is a smell. A function named `isXXX`‏ should probably returning only a boolean. Here, it looks like a `findXXX()` would be more suitable.

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -133,6 +134,31 @@ function assembleResource(
...
 
+  // Check if this is a GraphQL request and extract metadata
+  const graphqlData =
+    request && request.type === RequestType.FETCH
+      ? (() => {
+          const graphQlConfig = isGraphQlRequest(request.url, configuration)
+          if (graphQlConfig) {
+            const requestBody = request.init?.body
+            if (requestBody && typeof requestBody === 'string') {
+              const graphqlMetadata = extractGraphQlMetadata(requestBody)
```
bcaudan: **💬 suggestion:** ‏those checks seems to also be performed inside extractGraphQlMetadata, what about keeping all the extraction logic inside the function?

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -133,6 +134,31 @@ function assembleResource(
...
+              if (graphqlMetadata) {
+                return {
+                  graphql: {
+                    operationType: graphqlMetadata.operationType,
+                    operationName: graphqlMetadata.operationName,
+                    variables: graphqlMetadata.variables,
+                    payload: graphQlConfig.trackPayload === true ? graphqlMetadata.payload : undefined,
+                  },
```
bcaudan: **💬 suggestion:** ‏what about handling the addition of the payload into the metada inside the extraction method?
```suggestion
                return {
                  graphql: graphqlMetada
                },
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -150,6 +176,7 @@ function assembleResource(
         url: request ? sanitizeIfLongDataUrl(request.url) : entry!.name,
         protocol: entry && computeResourceEntryProtocol(entry),
         delivery_type: entry && computeResourceEntryDeliveryType(entry),
+        ...graphqlData,
```
bcaudan: **🥜 nitpick:** ‏we could be more explicit/consistent with something like
```suggestion
        graphql,
```

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,74 @@
...
+      return graphQlOption
+    }
+  }
+  return false
+}
+
+export function extractGraphQlMetadata(requestBody: string | undefined): GraphQlMetadata | undefined {
+  if (!requestBody || typeof requestBody !== 'string') {
+    return undefined
+  }
```
bcaudan: **❓ question:** ‏what is the plan for [other body types](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#body)?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,74 @@
...
+  let variables: string | undefined
+  if (graphqlBody.variables !== null && graphqlBody.variables !== undefined) {
+    variables = JSON.stringify(graphqlBody.variables)
+  }
+
+  return {
+    operationType,
+    operationName,
+    variables,
+    payload: truncatePayload(query, GRAPHQL_PAYLOAD_LIMIT),
```
bcaudan: **💬 suggestion:** ‏let's not do the truncation if we don't need the payload

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const query = graphqlBody.query.trim()
+  const operationType = getOperationType(query)
+  const operationName = graphqlBody.operationName
+
+  let variables: string | undefined
+  if (graphqlBody.variables !== null && graphqlBody.variables !== undefined) {
+    variables = JSON.stringify(graphqlBody.variables)
+  }
```
bcaudan: **❓ question:** ‏do we have some evidences that we need to trim, handle null cases?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  return 'query'
+}
+
+function truncatePayload(payload: string, limit: number): string {
+  if (payload.length > limit) {
+    return `${payload.substring(0, limit)}...`
+  }
+  return payload
+}
```
bcaudan: **💬 suggestion:** ‏don't we already have some utility function to do that? I'm pretty sure that we are doing the same thing elsewhere

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: Hey ! It was an arbitrary value, I took inspiration from an other truncation in the logs. A generated payload with the maximum size would look like [this](https://app.datad0g.com/rum/sessions?query=%40type%3Aresource%20%40application.id%3Abc3c5c74-8897-4368-9933-49767f641c41%20%40resource.graphql.operationName%3A%2A&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZjvtk-JvJQH6gAAABhBWmp2dGstVEFBQUo5WC1sX2lKUFZRQUEAAAAkZDE5OGVmYjYtNGY4OS00MTA4LWE1NmQtNDI4YjAwZTI4Y2E5AAAAAQ&fromUser=false&viz=stream&from_ts=1756225157107&to_ts=1756829957107&live=true)

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: Yes exactly, I added this part in case of some null cases, also in variables.

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: good catch we have one, I imported it in the file 👍

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: Graphql request are sent as a json, otherwise it's not standard so I don't think we should support no ?

packages/rum-core/src/domain/resource/resourceCollection.ts
rgaignault: Seems clearer indeed

packages/rum-core/src/domain/resource/graphql.ts
bcaudan: saying that in a comment could still be helpful for someone looking into it in the future

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -168,6 +173,29 @@ function assembleResource(
   }
 }
 
+function computeGraphQlData(
+  request: RequestCompleteEvent | undefined,
+  configuration: RumConfiguration
+): { graphql?: GraphQlMetadata } {
```
bcaudan: **💬 suggestion:** ‏what about returning the metadata directly?
```suggestion
): GraphQlMetadata | undefined {
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -168,6 +173,29 @@ function assembleResource(
...
+  if (!request || request.type !== RequestType.FETCH) {
```
bcaudan: **❓ question:** ‏does anything prevent customers to use XHR for graphql queries?

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -168,6 +173,29 @@ function assembleResource(
...
+    return {}
+  }
+
+  const graphQlConfig = findGraphQlConfiguration(request.url, configuration)
+  if (!graphQlConfig) {
+    return {}
+  }
+
+  const graphqlMetadata = extractGraphQlMetadata(request.init?.body, graphQlConfig.trackPayload)
```
bcaudan: **❓ question:** ‏does anything prevent customers to use [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) for their calls? cf `input` field of `RequestCompleteEvent`

packages/rum-core/src/domain/resource/graphql.ts
bcaudan: From my understanding, some other types could resolve to a json, like a stream.
It should be pretty rare so not supporting until someone is requesting it should be fine.

packages/rum-core/src/domain/resource/graphql.ts
bcaudan: Did you see some queries with those edge cases? or was it implemented this way on the mobile side?

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -397,6 +442,16 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
     action_name_attribute: configuration.actionNameAttribute,
     use_allowed_tracing_urls:
       Array.isArray(configuration.allowedTracingUrls) && configuration.allowedTracingUrls.length > 0,
+    use_allowed_graph_ql_urls:
+      Array.isArray(configuration.allowedGraphQlUrls) && configuration.allowedGraphQlUrls.length > 0,
+    use_track_graph_ql_payload:
+      Array.isArray(configuration.allowedGraphQlUrls) &&
+      configuration.allowedGraphQlUrls.some((option) => {
+        if (typeof option === 'object' && 'trackPayload' in option) {
+          return option.trackPayload === true
```
bcaudan: **💬 suggestion:** ‏to be consistent with how it is resolved for the configuration
```suggestion
          return !!option.trackPayload
```

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: I mean the variable field is nullable, so if I don't do this check the event would have **`variables : null`** in the event in some cases. For the trim part I guess this maybe not necessary as this probably trimmed in a later process before displaying in the ui, I will remove it 👌

packages/rum-core/src/domain/resource/resourceCollection.ts
rgaignault: Hmm I guess no but it's a readable stream so we can not access it at this time, what are you thoughts about handling this case ?

packages/rum-core/src/domain/resource/resourceCollection.ts
rgaignault: In fact no, customers can send as they want. We didn't collect body as we do for fetch in the XHR requests, since it's mainly fetch also nowadays for graphql I was not sure if we would like to collect it or not. Looking at mobile implementation they do have a xhr proxy for graphql. I added support for xhr to match it 👌

packages/rum-core/src/domain/resource/graphql.ts
bcaudan: ok for null.
about the trim, I am more concerned about the parsing of the query on the SDK side. Is it possible for customers to send query with additional spaces? 
if not, let's not trim, you still have one left in getOperationType. 
if yes, let's handle all the possible cases and add some tests.

packages/rum-core/src/domain/resource/resourceCollection.ts
bcaudan: From my understanding, you can use Request with a string body, so I'd support this usage like we already do in fetchObservable.

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (!graphqlBody || !graphqlBody.query) {
+    return undefined
+  }
+
+  const query = graphqlBody.query.trim()
+  const operationType = getOperationType(query)
+  const operationName = graphqlBody.operationName
+
+  let variables: string | undefined
+  if (graphqlBody.variables !== null && graphqlBody.variables !== undefined) {
```
BenoitZugmeyer: **suggestion:** 
```suggestion
  if (graphqlBody.variables) {
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -168,6 +173,35 @@ function assembleResource(
...
+  } else if (request.type === RequestType.XHR) {
+    requestBody = request.body
+  }
+
+  const graphqlMetadata = extractGraphQlMetadata(requestBody, graphQlConfig.trackPayload)
+  if (!graphqlMetadata) {
+    return undefined
+  }
+
+  return graphqlMetadata
```
BenoitZugmeyer: **suggestion:** 
```suggestion
  return extractGraphQlMetadata(requestBody, graphQlConfig.trackPayload)
```

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -397,6 +442,16 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
     action_name_attribute: configuration.actionNameAttribute,
     use_allowed_tracing_urls:
       Array.isArray(configuration.allowedTracingUrls) && configuration.allowedTracingUrls.length > 0,
+    use_allowed_graph_ql_urls:
+      Array.isArray(configuration.allowedGraphQlUrls) && configuration.allowedGraphQlUrls.length > 0,
```
BenoitZugmeyer: **suggestion:** to reduce a bit the boilerplate, you could introduce a `isNonEmptyArray(a)` function and use it for all those `Array.isArray(a) && a.length > 0`

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export function findGraphQlConfiguration(url: string, configuration: RumConfiguration): GraphQlUrlOption | undefined {
+  for (const graphQlOption of configuration.allowedGraphQlUrls) {
+    if (matchList([graphQlOption.match], url)) {
+      return graphQlOption
+    }
+  }
+  return undefined
+}
```
BenoitZugmeyer: **suggestion:** 
```suggestion
export function findGraphQlConfiguration(url: string, configuration: RumConfiguration): GraphQlUrlOption | undefined {
  return configuration.allowedGraphQlUrls.find(graphQlOption => matchList([graphQlOption.match], url))
}
```

Or you could introduce a helper that factorizes this with `tracer.ts` usage

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,71 @@
...
+
+function getOperationType(query: string): 'query' | 'mutation' | 'subscription' {
+  if (query.startsWith('mutation')) {
+    return 'mutation'
+  } else if (query.startsWith('subscription')) {
+    return 'subscription'
+  }
+
+  return 'query'
```
BenoitZugmeyer: **suggestion:** shorter, and you don't need to `trim()`
```ts
function getOperationType(query: string): 'query' | 'mutation' | 'subscription' | undefined {
  return query.match(/^\s*(query|mutation|subscription)\b/i)?.[1]
}
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -168,6 +173,35 @@ function assembleResource(
   }
 }
 
+function computeGraphQlData(
```
BenoitZugmeyer: **suggestion:** in some place you call it `GraphQlData`, in other places `GraphQlMetadata`. Maybe unify the names?

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: I can not directly return since match return a string, I had to cast. Let me know what you think

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,64 @@
...
+ * arbitrary value, byte precision not needed
+ */
+const GRAPHQL_PAYLOAD_LIMIT = 32 * ONE_KIBI_BYTE
+
+export interface GraphQlMetadata {
+  operationType: 'query' | 'mutation' | 'subscription'
+  operationName?: string
+  variables?: string
+  payload?: string
+}
```
samcolmanetti, line 14: Hey, i'm just a GraphQL fan checking in. Feel free to disregard. 

Just curious: do we also plan to supporting persisted queries or anything in the extensions field? Persisted queries aren't officially in the GraphQL spec but supported by [Apollo](https://www.apollographql.com/docs/react/data/persisted-queries#protocol) and [Relay](https://relay.dev/docs/guides/persisted-queries/). When the persisted query id/hash is provided, the query document itself won't be provided. This would allow customers to figure out what was executed by the client. It might be worth looking into saving the entire extensions field. There shouldn't be anything privacy sensitive in there if they're following the [spec](https://spec.graphql.org/September2025/#sel-CANFHLAAACJAlH) and only use it for "implementation-specific additional information."


Apollo seems to include it in the extensions field: 
```
{
  operationName: 'MyQuery',
  variables: null,
  extensions: {
    persistedQuery: {
      version: 1,
      sha256Hash: hashOfQuery
    }
  }
}
```

packages/core/src/browser/xhrObservable.ts
```diff
@@ -72,7 +73,7 @@ function openXhr({ target: xhr, parameters: [method, url] }: InstrumentedMethodC
 }
 
 function sendXhr(
-  { target: xhr, handlingStack }: InstrumentedMethodCall<XMLHttpRequest, 'send'>,
+  { target: xhr, parameters, handlingStack }: InstrumentedMethodCall<XMLHttpRequest, 'send'>,
```
bcaudan: **💬 suggestion:** similarly to the open method instrumentation, we can use destructuring to be more explicit:

```diff
-  { target: xhr, parameters, handlingStack }: InstrumentedMethodCall<XMLHttpRequest, 'send'>,
+  { target: xhr, parameters: [body], handlingStack }: InstrumentedMethodCall<XMLHttpRequest, 'send'>,
...
- startContext.body = parameters[0]
+ startContext.body = body
```

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -396,13 +442,20 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
...
+    use_allowed_tracing_urls: isNonEmptyArray(configuration.allowedTracingUrls),
+    use_allowed_graph_ql_urls: isNonEmptyArray(configuration.allowedGraphQlUrls),
+    use_track_graph_ql_payload:
+      isNonEmptyArray(configuration.allowedGraphQlUrls) &&
+      configuration.allowedGraphQlUrls.some((option) => {
+        if (typeof option === 'object' && 'trackPayload' in option) {
+          return !!option.trackPayload
+        }
+        return false
+      }),
```
bcaudan: **🥜 nitpick:** ‏we could extract a function to limit the logic inside `serializeRumConfiguration`

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -169,6 +174,34 @@ function assembleResource(
...
+    return undefined
+  }
+
+  // Get request body based on request type
+  let requestBody: unknown
+  if (request.type === RequestType.FETCH) {
+    requestBody = request.init?.body
+  } else if (request.type === RequestType.XHR) {
+    requestBody = request.body
+  }
```
bcaudan: **💬 suggestion:** ‏what about mapping request.init.body to request.body? as both xhr and fetch emit RequestCompleteEvent, it would avoid to check the type of request here

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -169,6 +174,34 @@ function assembleResource(
   }
 }
 
+function computeGraphQlMetaData(
+  request: RequestCompleteEvent | undefined,
+  configuration: RumConfiguration
+): GraphQlMetadata | undefined {
+  if (!request) {
+    return undefined
+  }
```
bcaudan: **💬 suggestion:** ‏to be consistent, you could move this check up like the other computations and do:

```
graphql: request && computeGraphQlMetaData(request, configuration)
```

test/e2e/scenario/rum/graphql.scenario.ts
```diff
@@ -0,0 +1,122 @@
+import { test, expect } from '@playwright/test'
+import { createTest } from '../../lib/framework'
+
+test.describe('GraphQL tracking', () => {
+  createTest('track GraphQL query via XHR and include payload')
+    .withRum({
+      allowedGraphQlUrls: [{ match: (url: string) => url.includes('graphql'), trackPayload: true }],
+      enableExperimentalFeatures: ['graphql_tracking'],
+    })
```
bcaudan: **🥜 nitpick:** ‏we could extract a function to build the configuration for those tests:

```suggestion
    .withRum(buildGraphQlConfig({ trackPayload: true }))
```

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,70 @@
...
+  trackPayload: boolean = false,
+  request?: RequestCompleteEvent
+): GraphQlMetadata | undefined {
+  if (!requestBody || typeof requestBody !== 'string') {
+    // Check if we have a Request object with a body that we couldn't access (stream)
+    if (request?.input && typeof request.input === 'object' && 'body' in request.input) {
+      display.warn('GraphQL tracking does not support Request objects with body streams.')
+    }
+    return undefined
+  }
```
bcaudan: **💬 suggestion:** ‏what about warning for all the cases where:
- a body is present but we don't support the type (ex: fetch with a stream body)
- a body may be present but we don't support the API used by the customer (ex: usage of Request API)

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const query = graphqlBody.query.trim()
+  const operationType = getOperationType(query)
+  const operationName = graphqlBody.operationName
+
+  if (!operationType) {
+    return undefined
+  }
```
bcaudan: **💬 suggestion:** ‏we could add a unit test to cover missing operation type

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -0,0 +1,72 @@
...
+): GraphQlMetadata | undefined {
+  if (!requestBody || typeof requestBody !== 'string') {
+    if (request?.input && typeof request.input === 'object' && 'body' in request.input) {
+      display.warn('GraphQL tracking does not support Request objects with body streams.')
+    } else if (requestBody !== null && requestBody !== undefined) {
+      const bodyType = requestBody.constructor.name || typeof requestBody
+      display.warn(`GraphQL tracking does not support body type: ${bodyType}. Only string bodies are supported.`)
+    }
+    return undefined
+  }
```
bcaudan: **💬 suggestion:** ‏it seems that a lot is going on there, I think we could:
- make things clearer by separating the cases where the body is missing from the cases where it is not a supported type
- check directly if it the input is a Request instance, unless there was a reason to check if the body is also in the input
- Not sure to get why we should include "body streams" in the warning as we won't support a string body either, no?
- We could hint in the warning messages that we may support it in the future

```suggestion
if (!requestBody) {
    if (request?.input instanceof Request) {
      display.warn('GraphQL tracking does not yet support Request objects.')
    }
    return undefined
  }

  if (typeof requestBody !== 'string') {
    const bodyType = requestBody.constructor.name || typeof requestBody
    display.warn(`GraphQL tracking does not support body type: ${bodyType}. Only string bodies are currently supported.`)
    return undefined
  }
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -169,6 +174,24 @@ function assembleResource(
   }
 }
 
+function computeGraphQlMetaData(
+  request: RequestCompleteEvent,
+  configuration: RumConfiguration
+): GraphQlMetadata | undefined {
+  if (!isExperimentalFeatureEnabled(ExperimentalFeature.GRAPHQL_TRACKING)) {
+    return undefined
```
BenoitZugmeyer: **nitpick:** I see `return undefined` everywhere, and while it's fine since the `undefined` will be removed during the build, it adds unnecessary noise in the source code. What about just removing `undefined` everywhere `return undefined` is used?

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: The body become a stream before we intercept it either it was originally a string or not, so I would think this does not change the behaviour to detect this or the stream directly. I agree with the hint about the warning messages

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -169,6 +174,24 @@ function assembleResource(
...
+  }
+
+  const graphQlConfig = findGraphQlConfiguration(request.url, configuration)
+  if (!graphQlConfig) {
+    return
+  }
+
+  const requestBody = request.body
+
+  return extractGraphQlMetadata(requestBody, graphQlConfig.trackPayload)
```
bcaudan: **💬 suggestion:** let's inline requestBody
```suggestion
  return extractGraphQlMetadata(request.body, graphQlConfig.trackPayload)
```
‏

PR #3808, opened 2025-08-28
✨ [PANA-3969] Try to collect some initial view metrics even when the page unloads early

sethfowler-datadog:
## Motivation

After looking at the initial view metrics telemetry we've collected so far, I'm a bit concerned that the numbers may be distorted by not including any data about situations where a navigation occurs before the load event has fired. I'd like to try to get at least some data about what's happening in these situations.

## Changes

This PR adds a second variant to the initial view metrics telemetry. Now, if `PAGE_MAY_EXIT` fires with reason `UNLOADING` before we've send the telemetry, we'll try to send a minimal telemetry message that includes:
* The timestamp at which the page is unloading.
* The timestamp at which DOMContentLoaded fired, if it has fired already.

## Test instructions

Probably easiest to test this on staging; I'll push it there.

## Checklist

- [x] Tested locally
- [ ] Tested on staging *(In progress.)*


thomas-lebeau:
@sethfowler-datadog is this still relevant?
I'm moving that to draft as this seems stale. Please fill free to ether reopen or close as needed.

PR #3810, opened 2025-09-02
[🐛] Fix playwright version in BrowserStack to match package.json

BeltranBulbarellaDD:
## Motivation

Currently we have an error in Edge BrowserStack job: `Error: page.waitForTimeout: waitTimeout: expected number, got undefined`. We believe it comes from a version mismatch between what is from the playwright [helper file](https://github.com/DataDog/browser-sdk/blob/main/test/e2e/lib/helpers/playwright.ts#L36) and the playwright version in [package.json](https://github.com/DataDog/browser-sdk/blob/main/package.json#L44C5-L44C34). 

## Changes

This PR makes `browserstack.playwrightVersion` and `client.playwrightVersion` have the same version than the one in package.json

## Test instructions

CI should pass. 
Screenshot 2025-09-02 at 16 57 17

## Checklist

- [X] Tested locally
- [ ] Tested on staging


test/e2e/lib/helpers/playwright.ts
```diff
@@ -33,8 +33,8 @@ function getCapabilities(configuration: BrowserConfiguration) {
     build: getBuildInfos(),
     name: configuration.sessionName,
     'browserstack.local': true,
-    'browserstack.playwrightVersion': '1.latest',
-    'client.playwrightVersion': '1.latest',
+    'browserstack.playwrightVersion': '1.53.0',
+    'client.playwrightVersion': '1.53.0',
```
BenoitZugmeyer: **suggestion:** can we read the playwright version from `package.json`?

BeltranBulbarellaDD: Great! Here I see 2 options, creating a function `getLocalPlaywrightVersion` in test/e2e/lib/helpers/playwright.ts or adding a [prebuild script](https://stackoverflow.com/a/67701490).
What do you think?

BenoitZugmeyer: Creating a function would be simpler!

BeltranBulbarellaDD: Perfect! I am thinking that the function's test would only be a console log. But maybe it's worth for debugging purposes?
Screenshot 2025-09-02 at 17 14 32

test/e2e/lib/helpers/playwright.ts
```diff
@@ -33,11 +36,25 @@ function getCapabilities(configuration: BrowserConfiguration) {
...
+    const pkgPath = path.resolve(process.cwd(), 'package.json')
+    const content = fs.readFileSync(pkgPath, 'utf-8')
+    const pkg = JSON.parse(content) as {
+      devDependencies?: Record<string, string>
+      dependencies?: Record<string, string>
+    }
+    return pkg.devDependencies?.['@playwright/test'] || pkg.dependencies?.['@playwright/test'] || '1.latest'
+  } catch {
+    return '1.latest'
+  }
```
BenoitZugmeyer: **suggestion:** Don't fail silently. Feel free to throw an exception to abort the process when we fail to find a correct version.

test/e2e/lib/helpers/playwright.ts
```diff
@@ -33,11 +36,25 @@ function getCapabilities(configuration: BrowserConfiguration) {
...
     'browserstack.networkLogs': false,
     'browserstack.interactiveDebugging': false,
   }
 }
+
+function getLocalPlaywrightVersion(): string {
+  try {
+    const pkgPath = path.resolve(process.cwd(), 'package.json')
+    const content = fs.readFileSync(pkgPath, 'utf-8')
+    const pkg = JSON.parse(content) as {
```
BenoitZugmeyer: **suggestion:** you can do:

```
import packageJson from '../../../../package.json' with { type: 'json' }
```

at the top of the module

BeltranBulbarellaDD: The only issue I run is with:
`with { type: 'json' }`
As I got:
`import attributes are only supported when the '--module' option is set to 'esnext', 'node18', 'nodenext', or 'preserve'.`
I did a workaround

mormubis (approved):

PR #3813, opened 2025-09-03
✨ [RUM-10292] Add noRestoredReasons support

rgaignault:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewCollection.ts
+++ b/packages/rum-core/src/domain/view/viewCollection.ts
@@ -129,6 +129,7 @@ function processViewUpdate(
         count: view.eventCounts.longTaskCount,
       },
       performance: computeViewPerformanceData(view.commonViewMetrics, view.initialViewMetrics),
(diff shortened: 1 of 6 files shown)


thomas-lebeau:
Are we able to see these new data on staging?

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -171,6 +173,26 @@ function processViewUpdate(
   }
 }
 
+function mapNotRestoredReasons(
```
amortemousque: **❓ question:** ‏Why do we need this mapping if the source and target model are the same? Couldn’t we just call notRestoredReasons.toJson() directly?

rgaignault: True ! I made some testing and thought this would be necessary with the childrens, but it doesn't seem in fact. I switched to json

PR #3815, opened 2025-09-03
🔊 add telemetry for unexpected session id changes

BenoitZugmeyer:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/core/src/domain/session/sessionManager.ts
```diff
@@ -184,3 +194,39 @@ async function reportUnexpectedSessionState() {
...
+    }
+
+    const sessionAge = dateNow() - sessionCreatedTime
+    if (sessionAge > 14 * ONE_MINUTE) {
+      // The session might have expired just because it's too old or lack activity
+      stop()
+    } else {
+      const newSessionState = toSessionState(changed.value)
+      if (newSessionState.id !== initialSessionState.id) {
+        addTelemetryDebug('Session cookie changed', {
```
thomas-lebeau: **💬 suggestion:** ‏ I think there is still some amount of short session that are due to cookie set with different domain or config resulting to having multiple `_dd_s` cookies. 

What about adding all the `_dd_s` cookies from `cookieStore.getAll('_dd_s')` as these will include the domain and other options the cookie were set with and it could be useful to understand why the session was shorter than expected

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -184,3 +194,39 @@ async function reportUnexpectedSessionState() {
...
+  stopCallbacks.push(stop)
+
+  function listener(event: CookieChangeEvent) {
+    const changed = findLast(event.changed, (change): change is CookieListItem => change.name === SESSION_STORE_KEY)
+    if (!changed) {
+      return
+    }
+
+    const sessionAge = dateNow() - sessionCreatedTime
+    if (sessionAge > 14 * ONE_MINUTE) {
```
bcaudan, line 205: **❓ question:** ‏why not using the session expiration delay?

BenoitZugmeyer: I want to prevent a few false positives at the cost of losing a bit of logs. I'm fearing that if we are too close to the actual limit, we could face clock issues. I'm not sure that `Date.now()` returns the exact same value across tabs, so the cookie could be expired in some tab and still be considered within the 15 minutes limit in another. With a 1 minute margin of error, we get rid of this potential issue.

Do you think that could have a negative impact on the data?

bcaudan: I just wanted to understand the intent.

IMO, that should be fine to get telemetry data, if we want to later include that in a potential fix, it could be interesting to explore a bit more this hypothesis around different `Date.now()` values across tabs.

thomas-lebeau (approved):

PR #3822, opened 2025-09-05
✨ Video Quality Monitoring adaptation

mormubis:
## Motivation

Video Quality Monitoring initiative plans to reuse view events to send stream information. This PR adds those foundational changes. Additionally, there will be an additional PR in the playground where the VQM plugin will be published.

## Changes

This PR adds the `raw` types for `stream` that gets transformed in the assembly step to a `view` event.

## Test instructions

This PR is difficult to manually test. Changes are planned for a follow-up PR in the playground where the new VQM plugin uses `stream` events.

## Checklist

- [X] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/assembly.spec.ts
```diff
@@ -575,6 +575,67 @@ describe('rum assembly', () => {
...
+      }
+
+      notifyRawRumEvent(lifeCycle, {
+        rawRumEvent: createRawRumEvent(RumEventType.STREAM, {
+          stream: streamData,
+          view: { id: 'original-view-id', url: '/test-page' },
+        }),
+      })
+
+      expect(serverRumEvents.length).toBe(1)
```
eloytoro, line 618: should there be an assertion for the event having a `stream` attribute at this stage?

mormubis: Added!

eloytoro (approved):

packages/rum-core/src/rumEvent.types.ts
```diff
@@ -826,7 +826,8 @@ export type RumResourceEvent = CommonProperties &
  * Schema of all properties of a View event
  */
 export type RumViewEvent = CommonProperties &
-  ViewContainerSchema & {
+  ViewContainerSchema &
+  StreamSchema & {
```
BenoitZugmeyer: **question:** Is `Stream` supposed to be its own event or part of the RUM View? I'm confused

packages/rum-core/src/domain/trackEventCounts.ts
```diff
@@ -30,7 +30,7 @@ export function trackEventCounts({
   }
 
   const subscription = lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event): void => {
-    if (event.type === 'view' || event.type === 'vital' || !isChildEvent(event)) {
+    if (event.type === 'view' || event.type === 'vital' || !isChildEvent(event) || ['stream'].includes(event.type)) {
```
BenoitZugmeyer: **suggestion:** let's keep things uniform, either use `event.type === 'stream'` or use a list of event types that are never counted containing `view`, `vital` and `stream`

packages/rum-core/src/domain/assembly.ts
```diff
@@ -126,7 +131,39 @@ export function startRumAssembly(
         if (isEmptyObject(serverRumEvent.context!)) {
           delete serverRumEvent.context
         }
-        lifeCycle.notify(LifeCycleEventType.RUM_EVENT_COLLECTED, serverRumEvent)
+
+        if (rawRumEvent.type === 'stream') {
+          const streamEvent = {
```
BenoitZugmeyer, line 142: **thought:** this doesn't look like production ready code..

packages/rum-core/src/rumEvent.types.ts
BeltranBulbarellaDD: The idea is it to be part of the RUM View

packages/rum-core/src/domain/assembly.ts
BeltranBulbarellaDD: My knowledge is that the idea is to make it work, and have it tested next week with customers and if no one want's it, to be dropped. That's why I think maybe the integration is not production ready. 

But I will move this to a helper function, use the types and unit test it to make it prod ready.

PR #3823, opened 2025-09-05
♻️ improve browser extension e2e tests

thomas-lebeau:
## Motivation

Testing browser extensions in e2e was not working the same way as other tests, this PR refactor this to simplify and harmonize
It also fixes the e2e test for the browser extension that were not working properly

## Changes

- Remove unused `popup` and `background-script` from the test browser extension as we don't do any test on them
- `withExtension()` will now get an `Extension` object that can contain a separate rum configuration
- Add `createExtension(path)` to use with `withExtension()`. It's rum configuration is configured separately from the base page rum config, allowing to test our different use-cases.
  ```
  createTest()
    .withRum()
    .withExtension(createExtension().withRum({ allowTrackingOrigin: ['example.com'] }))
    .run()
  ```
- The extension follows the same `DEFAULT_RUM_CONFIGURATION` as other tests, including the intake proxy config so we can now do assertion on the request send to the intake and the data will be send to datadog with the proper context and api keys
- Exposes `withSetup()` in order to run some test with a unique setup. 
- Simplify the browser extension tests to test what actually matter
- Add a test (failing from now) for the false positive we encountered

## Checklist

- [x] Tested locally
- [ ] Tested on staging


test/apps/base-extension/src/cdnContentScript.ts
```diff
@@ -0,0 +1,46 @@
...
+      setGlobalContext: (ctx: Record<string, unknown>) => void
+    }
+    DD_LOGS?: {
+      init: (config: Record<string, unknown>) => void
+      setGlobalContext: (ctx: Record<string, unknown>) => void
+    }
+  }
+}
+
+const scriptUrl = window.RUM_BUNDLE_URL || window.LOGS_BUNDLE_URL
```
thomas-lebeau: **🔨 warning:** ‏ you won't be able to use both rum and log SDK this way, I would do something like:

```ts
function load(sdk, url, initConfig, globalContext) {
  const script = document.createElement('script')
  script.src = url
  script.onload = () => {
    window[sdk].init(initConfig)

    if (globalContext) {
      window[sdk].setGlobalContext(globalContext)
    }
  }
  document.head.appendChild(script)
}

load('DD_RUM', window.RUM_BUNDLE_URL, window.RUM_CONFIGURATION, window.RUM_CONTEXT)
load('DD_LOGS', window.LOGS_BUNDLE_URL, window.LOGS_CONFIGURATION, window.LOGS_CONTEXT)

```

test/apps/base-extension/src/cdnContentScript.ts
```diff
@@ -0,0 +1,46 @@
...
+if (scriptUrl) {
+  const script = document.createElement('script')
+  script.src = scriptUrl
+  script.async = true
```
thomas-lebeau: **💬 suggestion:** ‏ I think the sync setup will be enough to test what we need and it will be simpler.

test/e2e/scenario/browser-extensions/browserExtensions.scenario.ts
```diff
@@ -1,133 +1,61 @@
...
+const DISABLE_SESSION_REPLAY_CONFIGURATION = { sessionReplaySampleRate: 0 }
+
+interface Ext {
+  name: string
+  path: string
+}
+
+const EXTENSIONS: Ext[] = [
+  { name: 'bundle', path: pathToBundleExtension },
+  { name: 'cdn', path: pathToCdnExtension },
```
thomas-lebeau:
```suggestion
  { 
    name: 'bundle',
    path: path.join(__dirname, '../../../../test/apps/base-extension')
  },
  {
    name: 'cdn',
    path: path.join(__dirname, '../../../../test/apps/cdn-extension')
  },
```

test/e2e/scenario/browser-extensions/browserExtensions.scenario.ts
```diff
@@ -1,133 +1,61 @@
...
+]
+
+function declareExtensionTests(ext: Ext) {
+  const warningMessage =
+    'Datadog Browser SDK: Running the Browser SDK in a Web extension content script is discouraged and will be forbidden in a future major release unless the `allowedTrackingOrigins` option is provided.'
+  const errorMessage = 'Datadog Browser SDK: SDK initialized on a non-allowed domain.'
+
+  createTest(`[${ext.name}] SDK is initialized in an unsupported environment and warns`)
```
thomas-lebeau: **💬 suggestion:** ‏ you're already have the extension name in the describe line 59

```suggestion
  createTest(`SDK is initialized in an unsupported environment and warns`)
```

test/e2e/scenario/browser-extensions/browserExtensions.scenario.ts
```diff
@@ -1,133 +1,92 @@
...
+          <script src="${rumScriptUrl}"></script>
+          <script>
+            const script = document.createElement('script')
+            script.innerHTML = 'window.DD_RUM.init(${formatConfiguration(options.rum!, servers)})'
+            document.head.appendChild(script)
+          </script>
+        `
+    })
+    .run(async ({ withBrowserLogs, flushEvents, intakeRegistry }) => {
+      test.fail() // TODO: remove this once the issue is fixed
```
thomas-lebeau, line 110: This ensure the test is failing but mark the test as passed. When we fix the implementation of `allowTrackingOrigin` we will be able to remove this.

PR #3824, opened 2025-09-05
🐛[RUM-11247] Pass down error stack to isAllowedTrackingOrigins

BeltranBulbarellaDD:
## Motivation

Currently there are 9 function calls before `isAllowedTrackingOrigins`. Chrome passes the first 9, so if we were to add 1 more function call before, we would never get the extension prefix in the stack

## Changes

Created the error stack in `makeRumPublicApi` and `makeLogsPublicApi` and passed it down to `isAllowedTrackingOrigins` function. 

(First error is the new stack, second error is the previous stack)
Screenshot 2025-09-10 at 18 03 56

## Test instructions

Use the dev extension and inject it in localHost:8080 while running the local SDK. If you print the error stack that `isAllowedTrackingOrigins` receives now and print a new error stack below it should be the same as the image above. We should see the error stack in a different location, closer to the top.

Also, to test the new detection logic. Go to a page that uses RUM, and install a browser extension, use the Local SDK on the page. Without this PR you should see the `WARN_DOES_NOT_HAVE_ALLOWED_TRACKING_ORIGIN` now you should see `SDK is loaded more than once. This is unsupported and might have unexpected behavior.`

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/extension/extensionUtils.ts
+++ b/packages/core/src/domain/extension/extensionUtils.ts
@@ -18,11 +18,10 @@ export function isUnsupportedExtensionEnvironment(windowLocation: string, stack:
   }
 
   // Since we generate the error on the init, we check the 2nd frame line.
-  const frameLines = stack
-    .split('\n')
-    .map((l) => l.trim())
-    .filter((l) => l.length > 0)
-    .filter((l) => /^at\s+/.test(l) || /@/.test(l))
+  const frameLines = stack.split('\n').filter((line) => {
+    const trimmedLine = line.trim()
+    return trimmedLine.length && /^at\s+|@/.test(trimmedLine)
+  })
   const target = frameLines[1] || ''
 
   return containsExtensionUrl(target)
--- a/test/apps/base-extension/yarn.lock
+++ b/test/apps/base-extension/yarn.lock
@@ -7,14 +7,14 @@ __metadata:
 
 "@datadog/browser-core@file:../../../packages/core/package.tgz::locator=rum-testing-extension%40workspace%3A.":
   version: 6.21.1
-  resolution: "@datadog/browser-core@file:../../../packages/core/package.tgz#../../../packages/core/package.tgz::hash=3291ba&locator=rum-testing-extension%40workspace%3A."
-  checksum: 10c0/2953b2ed3ac94bdc6540ad8b3e61de3697bdd28916755eebae51e18af4325067411078a3775017f872924dcf09d08dacaea09471775b4df41e526dc8435bdf67
+  resolution: "@datadog/browser-core@file:../../../packages/core/package.tgz#../../../packages/core/package.tgz::hash=7a9f3f&locator=rum-testing-extension%40workspace%3A."
+  checksum: 10c0/4a42aca1f7fe4af3048e7483052efa47aa4cc901c056c5e7171aa64526bb6f1b2f48b73655b85c9a55510871d5940d7900e3562d63afcd3aa431e938e91d59be
   languageName: node
   linkType: hard
 
 "@datadog/browser-logs@file:../../../packages/logs/package.tgz::locator=rum-testing-extension%40workspace%3A.":
   version: 6.21.1
-  resolution: "@datadog/browser-logs@file:../../../packages/logs/package.tgz#../../../packages/logs/package.tgz::hash=f7cefe&locator=rum-testing-extension%40workspace%3A."
+  resolution: "@datadog/browser-logs@file:../../../packages/logs/package.tgz#../../../packages/logs/package.tgz::hash=e43f7a&locator=rum-testing-extension%40workspace%3A."
   dependencies:
     "@datadog/browser-core": "npm:6.21.1"
   peerDependencies:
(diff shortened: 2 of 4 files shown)


packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -293,7 +293,10 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs): LogsPublicApi {
   return makePublicApi<LogsPublicApi>({
     logger: mainLogger,
 
-    init: monitor((initConfiguration) => strategy.init(initConfiguration)),
+    init: monitor((initConfiguration) => {
+      const errorStack = new Error().stack
+      strategy.init(initConfiguration, errorStack)
+    }),
```
thomas-lebeau: **💬 suggestion:** ‏ Here the stack will have a few unnecessary frames
- anonymous function
- callMonitored
- monitor
- init

It would be easier to do something similar to `addAction()`
```suggestion
    init: (initConfiguration) => {
      const errorStack = new Error().stack

      callMonitored(() => strategy.init(initConfiguration, errorStack))
    },
```

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -1,20 +1,51 @@
...
+    at moz-extension://abcdef/dist/contentScript.js:13304:14
+    at moz-extension://abcdef/dist/contentScript.js:13315:3`
+
+// Edge case, the extension patches a function from the page.
+export const STACK_WITH_INIT_IN_PAGE_PATCHED = `Error
+    at Object.init (http://localhost:8080/datadog-rum.js:3919:16)
+    at <anonymous>:2:23
+    at Object.apply (chrome-extension://hgijklmn/WXkq0oBd.js:10:4624)
+    at http://localhost:8080/:16:21`
+
```
BeltranBulbarellaDD: These ones I didn't know where to have them, because they are used by multiple files.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -337,7 +340,7 @@ export function validateAndBuildConfiguration(initConfiguration: InitConfigurati
     !isString(initConfiguration.version, 'Version') ||
     !isString(initConfiguration.env, 'Env') ||
     !isString(initConfiguration.service, 'Service') ||
-    !isAllowedTrackingOrigins(initConfiguration)
+    !isAllowedTrackingOrigins(initConfiguration, errorStack!, undefined)
```
thomas-lebeau: **❓ question:** ‏ why not simply type `errorStack` as not optional in the function  signature?

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -1,20 +1,51 @@
...
+    at chrome-extension://abcdef/dist/contentScript.js:13304:14
+    at chrome-extension://abcdef/dist/contentScript.js:13315:3`
+
+export const STACK_WITH_INIT_IN_EXTENSION_FIREFOX = `Error
+    at Object.init (moz-extension://abcdef/dist/contentScript.js:254:14)
+    at moz-extension://abcdef/dist/contentScript.js:13304:14
+    at moz-extension://abcdef/dist/contentScript.js:13315:3`
+
+// Edge case, the extension patches a function from the page.
+export const STACK_WITH_INIT_IN_PAGE_PATCHED = `Error
```
thomas-lebeau: **🔨 warning:** ‏ this one does not seems used

packages/core/src/domain/extension/extensionUtils.ts
thomas-lebeau: As it's used only for tests, it would be better to move it in `/packages/core/test/browserExtension.ts`

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -1,20 +1,51 @@
...
-  // If we're on a regular web page but the error stack shows extension URLs,
-  // then an extension is injecting RUM.
-  return !containsExtensionUrl(windowLocation) && containsExtensionUrl(stack)
+  // If the page itself is an extension page.
+  if (containsExtensionUrl(windowLocation)) {
+    return false
+  }
+
+  // Consider only the 1st frame line, which is the init caller.
+  const frameLines = stack.split('\n').filter((l) => /^\s*at\s+/.test(l))
```
thomas-lebeau: **🔨 warning:** ‏ in firefox I think stacktraces don't contain `at`

packages/core/src/domain/configuration/configuration.ts
rgaignault: We would need to fill all test in all configuration files with stackTrace. Do you see any problem with optional errorStack ?

thomas-lebeau: right, what about the following then?
```suggestion
    !isAllowedTrackingOrigins(initConfiguration, errorStack ?? '')
```

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -1,20 +1,51 @@
...
+// Base case for extension, the extension has the SDK in the init and the error stack is in the extension.
+export const STACK_WITH_INIT_IN_EXTENSION = `Error
+    at Object.init (chrome-extension://abcdef/dist/contentScript.js:254:14)
+    at chrome-extension://abcdef/dist/contentScript.js:13304:14
+    at chrome-extension://abcdef/dist/contentScript.js:13315:3`
+
+export const STACK_WITH_INIT_IN_EXTENSION_FIREFOX = `Error
+    at Object.init (moz-extension://abcdef/dist/contentScript.js:254:14)
+    at moz-extension://abcdef/dist/contentScript.js:13304:14
+    at moz-extension://abcdef/dist/contentScript.js:13315:3`
```
thomas-lebeau: **❓ question:** where did you find this stack trace example? it does not looks like a Firefox trace

packages/core/src/domain/extension/extensionUtils.ts
```diff
@@ -1,20 +1,51 @@
...
+    at Object.apply (chrome-extension://hgijklmn/WXkq0oBd.js:10:4624)
+    at http://localhost:8080/:16:21`
+
 export function containsExtensionUrl(str: string): boolean {
   return EXTENSION_PREFIXES.some((prefix) => str.includes(prefix))
 }
 
 /**
  * Utility function to detect if the SDK is being initialized in an unsupported browser extension environment.
+ * Note: Because we check error stack, this will not work if the error stack is too long as Browsers truncate it.
```
thomas-lebeau: **❓ question:** Is it still a problem?

packages/rum-core/src/domain/configuration/configuration.spec.ts
```diff
@@ -610,9 +610,9 @@ describe('serializeRumConfiguration', () => {
             | 'allowedGraphQlUrls'
         ? `use_${CamelToSnakeCase<Key>}`
         : Key extends 'trackLongTasks'
-          ? 'track_long_task' // oops
+          ? 'track_long_task'
           : // The following options are not reported as telemetry. Please avoid adding more of them.
-            Key extends 'applicationId' | 'subdomain'
+            Key extends 'applicationId' | 'subdomain' | 'trackEarlyRequests'
```
thomas-lebeau: sounds like unrelated change, no?

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -472,7 +474,6 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
     track_resources: configuration.trackResources,
     track_long_task: configuration.trackLongTasks,
     track_bfcache_views: configuration.trackBfcacheViews,
-    track_early_requests: configuration.trackEarlyRequests,
```
thomas-lebeau, line 499: same, this looks like an unrelated change

packages/rum-core/src/domain/configuration/configuration.spec.ts
BeltranBulbarellaDD: Yeah, got it while merging with events format. If not I would get typecheck error in CI

thomas-lebeau (approved):

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -597,9 +597,10 @@ export function makeRumPublicApi(
   })
 
   const rumPublicApi: RumPublicApi = makePublicApi<RumPublicApi>({
-    init: monitor((initConfiguration) => {
-      strategy.init(initConfiguration, rumPublicApi)
-    }),
+    init: (initConfiguration) => {
+      const errorStack = new Error().stack
```
mormubis, line 601: **💬 suggestion:** ‏Should we assert and warn the user here? Then, should we simply pass the flag for telemetry purposes?

mormubis (approved):

packages/rum-core/src/boot/rumPublicApi.ts
BeltranBulbarellaDD: I understand but there we would have to do it twice as that is `rum-core` and currently we do it in `core`. Does that make sense?

PR #3838, opened 2025-09-12
✨ [RUM-11701] Add unity as a source override

fuzzybinary:
## Motivation

Source overrides only currently support Flutter, and should support Unity as a source override as well.

## Changes

Add `unity` both to the allowed TypeScript definitions and to the check that validates proper sources. Add a unit tests that checks that the `unity` source is forwarded properly.

## Test instructions

To to staging

Override the configuration and add:

```js
"source": "unity",
```

Check that events are sending the new information

Check that events in https://dd.datad0g.com/rum/sessions?query=%40type%3Asession%20%40usr.id%3A33803929&agg_m=count&agg_m_source=base&agg_t=count&from_ts=1752574307484&to_ts=1753179107484&live=true shows the new information

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/configuration/transportConfiguration.ts
+++ b/packages/core/src/domain/configuration/transportConfiguration.ts
@@ -13,7 +13,7 @@ export interface TransportConfiguration {
   datacenter?: string | undefined
   replica?: ReplicaConfiguration
   site: Site
-  source: 'browser' | 'flutter'
+  source: 'browser' | 'flutter' | 'unity'
 }
 
 export interface ReplicaConfiguration {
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -36,6 +36,13 @@ export function computeTransportConfiguration(initConfiguration: InitConfigurati
   }
 }
 
+function valiidateSource(source: string | undefined) {
```
BenoitZugmeyer: **suggestion:** 
```suggestion
function validateSource(source: string | undefined) {
```

packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -23,7 +23,7 @@ export interface ReplicaConfiguration {
 
 export function computeTransportConfiguration(initConfiguration: InitConfiguration): TransportConfiguration {
   const site = initConfiguration.site || INTAKE_SITE_US1
-  const source = initConfiguration.source === 'flutter' ? 'flutter' : 'browser'
+  const source = valiidateSource(initConfiguration.source)
```
mormubis:
```suggestion
  const source = validateSource(initConfiguration.source)
```

mormubis (approved):

PR #3851, opened 2025-09-17
[RUM Profiler] send view names as event attributes

thomasbertet:
## Motivation

We want to allow filtering by view name in the Profiling explorer (ie. `@view.name:/my-view`). 
In order to make the view name query-able, we need to pass it as the event's attributes (just as we do for view ids)

Screenshot 2025-09-17 at 14 39 06

## Changes

- new `view.name` attribute in the Profile's `event.json` sent to intake.

## Test instructions

- Browse a website with this version of the SDK, with the RUM Profiler active.
- After a while (tipically a minute of navigation), assert you can see a call the the profile intake.
- Assert you can see the `event.json` sent includes the new `view.name` attribute.
- Go over the Profiling's Explorer [page](https://app.datadoghq.com/profiling/explorer) 
- Filter using the search bar to `@view.name:<the-view-you-were-browsing>` and you should see a Flame Graph.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/transport/transport.ts
+++ b/packages/rum/src/domain/profiling/transport/transport.ts
@@ -1,14 +1,10 @@
 import type { Payload } from '@datadog/browser-core'
 import { addTelemetryDebug, buildTags, currentDrift } from '@datadog/browser-core'
 import type { RumConfiguration } from '@datadog/browser-rum-core'
-import type { RumProfilerTrace, RumViewEntry } from '../types'
+import type { RumProfilerTrace } from '../types'
+import { buildProfileEventAttributes } from './buildProfileEventAttributes'
+import type { ProfileEventAttributes } from './buildProfileEventAttributes'
 
-interface ProfileEventAttributes {
-  application: { id: string }
-  session?: { id: string }
-  view?: { id: string[]; name: string[] }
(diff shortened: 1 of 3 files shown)


packages/rum/src/domain/profiling/transport/transport.ts
```diff
@@ -108,43 +104,6 @@ function buildProfilingPayload(profilerTrace: RumProfilerTrace, profileEvent: Pr
...
-    application: {
-      id: applicationId,
-    },
-  }
-  if (sessionId) {
-    attributes.session = {
-      id: sessionId,
-    }
-  }
-  const viewIds = Array.from(new Set(profilerTrace.views.map((viewEntry) => viewEntry.viewId)))
```
BenoitZugmeyer, line 134: **question:** view ids were de-duplicated. It's not needed anymore?

BenoitZugmeyer (approved):

packages/rum/src/domain/profiling/transport/transport.ts
thomasbertet: I believe it's not needed yes. I don't see how we could generate duplicated ids 🤔

PR #3860, opened 2025-09-19
🐛 add support for arbitrary causes

mormubis:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/error/error.spec.ts
+++ b/packages/core/src/domain/error/error.spec.ts
@@ -8,7 +8,7 @@ import {
   isError,
   NO_ERROR_STACK_PRESENT_MESSAGE,
 } from './error'
-import type { RawErrorCause, ErrorWithCause } from './error.types'
+import type { ErrorWithCause, RawFlatErrorCause } from './error.types'
 import { ErrorHandling, ErrorSource, NonErrorPrefix } from './error.types'
 
 describe('computeRawError', () => {
@@ -179,7 +179,7 @@ describe('computeRawError', () => {
     expect(formatted.causes!.length).toBe(2)
     expect(formatted.stack).toContain('Error: foo: bar')
 
-    const causes = formatted.causes as RawErrorCause[]
+    const causes = formatted.causes as RawFlatErrorCause[]
 
     expect(causes[0].message).toContain(nestedError.message)
     expect(causes[0].source).toContain(ErrorSource.SOURCE)
@@ -250,7 +250,7 @@ describe('flattenErrorCauses', () => {
     error.cause = new Error('bar')
 
     const errorCauses = flattenErrorCauses(error, ErrorSource.LOGGER)
-    expect(errorCauses?.[0].type).toEqual('Error')
+    expect((errorCauses?.[0] as RawFlatErrorCause).type).toEqual('Error')
   })
 
   it('should only return the first 10 errors if nested chain is longer', () => {
--- a/packages/core/src/domain/error/error.ts
+++ b/packages/core/src/domain/error/error.ts
@@ -5,7 +5,15 @@ import { jsonStringify } from '../../tools/serialisation/jsonStringify'
 import type { StackTrace } from '../../tools/stackTrace/computeStackTrace'
 import { computeStackTrace } from '../../tools/stackTrace/computeStackTrace'
 import { toStackTraceString } from '../../tools/stackTrace/handlingStack'
-import type { ErrorSource, ErrorHandling, RawError, RawErrorCause, ErrorWithCause, NonErrorPrefix } from './error.types'
+import type {
+  ErrorSource,
+  ErrorHandling,
+  RawError,
+  RawErrorCause,
+  ErrorWithCause,
+  NonErrorPrefix,
+  RawFlatErrorCause,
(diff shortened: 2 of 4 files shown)


packages/core/src/domain/error/error.ts
```diff
@@ -87,17 +87,31 @@ export function isError(error: unknown): error is Error {
...
+        type: stackTrace?.name,
+        stack: stackTrace && toStackTraceString(stackTrace),
+      })
+
+      currentCause = (currentCause as ErrorWithCause).cause
+    } else {
+      causes.push({
+        message: jsonStringify(sanitize(currentCause))!,
+        source: parentSource,
+        type: typeof currentCause,
```
BenoitZugmeyer: **suggestion:** `type` should be undefined

packages/core/src/domain/error/error.ts
```diff
@@ -87,17 +87,31 @@ export function isError(error: unknown): error is Error {
...
+        stack: undefined,
+      })
+
+      currentCause = undefined
+    }
```
BenoitZugmeyer: **suggestion:** Can we refactor some logic with `computeRawError`? Ex:

```ts
function computeErrorBase({
  originalError,
  stackTrace,
  source,
  useFallbackStack = true,
  nonErrorPrefix,
}: Pick<RawErrorParams, 'stackTrace' | 'originalError' | 'source' | 'useFallbackStack' | 'nonErrorPrefix'>) {
  const isErrorInstance = isError(originalError)
  if (!stackTrace && isErrorInstance) {
    stackTrace = computeStackTrace(originalError)
  }

  return {
    source,
    type: stackTrace ? stackTrace.name : undefined,
    message: computeMessage(stackTrace, isErrorInstance, nonErrorPrefix, originalError),
    stack: stackTrace ? toStackTraceString(stackTrace) : useFallbackStack ? NO_ERROR_STACK_PRESENT_MESSAGE : undefined,
  }
}

```

packages/core/src/domain/error/error.ts
bcaudan: **💬 suggestion:** ‏FMU, the error type contains values like TypeError, ReferenceError, ... that we get from the stacktrace and without stacktrace we currently set this type to undefined.
Unless I am missing something, I think we could do the same here.

packages/core/src/domain/error/error.spec.ts
```diff
@@ -259,6 +249,96 @@ describe('flattenErrorCauses', () => {
...
+
+  describe('with non-Error values', () => {
+    it('should handle string cause with consistent structure', () => {
+      const error = new Error('main') as ErrorWithCause
+      error.cause = 'string cause'
+
+      const causes = flattenErrorCauses(error, ErrorSource.CUSTOM)
+      expect(causes?.length).toBe(1)
+      expect(causes?.[0]).toEqual({
+        message: '"string cause"', // JSON stringified
```
bcaudan, line 261: **💬 suggestion:** ‏For reporting thrown string we also stringify the message since we add a prefix the quotes act like a separator, ex: `Uncaught "string cause"`.
Not sure the quotes add much value in this case.

packages/core/src/domain/error/error.spec.ts
```diff
@@ -259,6 +249,96 @@ describe('flattenErrorCauses', () => {
...
+      // Second cause: Object with normalized structure
+      expect(causes?.[1]).toEqual({
+        message: '{"code":"ERR_ROOT"}',
+        source: ErrorSource.CUSTOM,
+        type: 'object',
+        stack: undefined,
+      })
+    })
+
+    it('should stop chain after non-Error cause', () => {
```
bcaudan, line 319: **❓ question:** ‏that sounds ok to me but I wonder if it is a choice you made or if it is based on spec.

mormubis: There are no specifications regarding how to handle complaint objects that are not errors. I made this decision to prevent a plain object with a `cause` field from causing any issues.

packages/core/src/domain/error/error.ts
mormubis: Updated it!

mormubis: Changed

packages/core/src/domain/error/error.ts
mormubis: Refactored!

packages/core/src/domain/error/error.spec.ts
mormubis: I didn't want to affect how they look in the UI. Also, it improves the readability when they are objects.

Thoughts?

BenoitZugmeyer (approved):

packages/core/src/domain/error/error.spec.ts
bcaudan: No strong opinion on this one, just pointing it out. I'll let you decide how to move forward.

packages/core/src/domain/error/error.spec.ts
bcaudan: LGTM

bcaudan (approved):

PR #3861, opened 2025-09-19
✨ enable rum profiler compression when possible

thomas-lebeau:
## Motivation

This PR uses core SDK platform transport features for the profiler endpoint. This enable:
- compression of profiler attachments when using `compressIntakeRequest: true`
- retry and queue mechanism 
- using fetch-keepalive when possible

## Changes

- Creates a reusable `createFormDataTransport` method that can be reused for other track.
behind the hood, this uses `httpRequest` and `deflateEncoder`
- refactor profiler unit tests to use the request interceptor to validate the format send to the intake

## Test instructions

In staging, verify that the request to the profiler intake contains compressed attachments.

Screenshot 2025-09-29 at 15 33 07

You can also verify in the view waterfall that the profiles are available.

Screenshot 2025-09-30 at 09 08 24

## Checklist

- [x] Tested locally
- [x] Tested on staging


packages/rum-core/src/transport/formDataTransport.ts
mormubis, line 1: **❓ question:** ‏Asking for a friend, why this transport lives within rum, instead of core?

packages/rum/src/domain/profiling/profiler.ts
```diff
@@ -42,8 +54,10 @@ export function createRumProfiler(
   lifeCycle: LifeCycle,
   session: RumSessionManager,
   profilingContextManager: ProfilingContextManager,
+  createEncoder: (streamId: DeflateEncoderStreamId) => Encoder,
   profilerConfiguration: RUMProfilerConfiguration = DEFAULT_RUM_PROFILER_CONFIGURATION
 ): RUMProfiler {
+  const transport = createFormDataTransport(configuration, lifeCycle, createEncoder, DeflateEncoderStreamId.PROFILING)
```
mormubis, line 60: **💭 thought:** ‏I feel dependency injection is becoming a bigger problem. We should think about a long-term solution.

packages/core/test/readFormData.ts
```diff
@@ -0,0 +1,51 @@
...
+function getEntries(payload: FormData) {
+  const entries = {} as Record<string, FormDataEntryValue>
+  payload.forEach((data, key) => {
+    entries[key] = data
+  })
+  return entries
+}
+
+function isZlibCompressed(buffer: ArrayBuffer) {
+  return new Uint8Array(buffer).subarray(0, 2).toString() === Z_LIB_MAGIC_BYTES.toString()
```
BenoitZugmeyer: **nitpick:** This is really nitpicking but it's too bad to use `toString()` when working with bytes :D 

Maybe `new DataView(buffer).getUint16(0, true) === 0x789c`

packages/core/test/readFormData.ts
```diff
@@ -0,0 +1,51 @@
+import { inflate } from 'pako'
+
+// Zlib streams using a default compression are starting with bytes 120 156 (0x78 0x9c)
+// https://stackoverflow.com/a/9050274
+const Z_LIB_MAGIC_BYTES = [120, 156]
+
+export async function readFormData<T>(formData: FormData): Promise<T> {
+  const entries = getEntries(formData)
+  const data = {} as Record<string, unknown>
```
BenoitZugmeyer: **nitpick:** 
```suggestion
  const data: Record<string, unknown> = {}
```

packages/core/test/readFormData.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  for (const [key, value] of Object.entries(entries)) {
+    if (value instanceof Blob) {
+      data[key] = await readJsonBlob(value as Blob)
```
BenoitZugmeyer: **nitpick:** 
```suggestion
      data[key] = await readJsonBlob(value)
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -570,7 +575,9 @@ export function makeRumPublicApi(
         startRumResult.hooks,
         configuration,
         startRumResult.session,
-        startRumResult.viewHistory
+        startRumResult.viewHistory,
+        createEncoder
+        // createIdentityEncoder // TODO: use `createEncoder` when evp supports zlib-wrapped deflate
```
BenoitZugmeyer: **suggestion:** remove the comment?

BenoitZugmeyer (approved):

packages/rum-core/src/transport/formDataTransport.ts
thomas-lebeau: yeah, I would like to have it moved to core, but the issue is that the reportError is only for rum events. LOGS SDK does not support reporting SDK errors to customers.

I think for now it's fine, all events except the logs lives in the rum SDK. depending on the new architecture we will be able to simplify

PR #3862, opened 2025-09-19
👷 [mob-session] make `test-performances` more resilient

BenoitZugmeyer:
## Motivation

The "bundle sizes" PR comment was proven very valuable, however the `test-performances` job is sometimes failing.

## Changes

Make the `test-performances` script more resilient by clearly separating the three tasks (bundle sizes, memory and CPU). Each task can update its own part of the PR comment, so if a task fails it doesn't prevent the other tasks to succeed.

## Test instructions

See [this PR comment](https://github.com/apps/cit-pr-commenter)

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/performance/lib/memoryPerformance.ts
+++ b/scripts/performance/lib/memoryPerformance.ts
@@ -1,67 +1,52 @@
 import type { Browser, CDPSession, Page, Protocol } from 'puppeteer'
 import puppeteer from 'puppeteer'
 import { fetchPR, LOCAL_BRANCH } from '../../lib/gitUtils.ts'
+import { formatSize, printLog } from '../../lib/executionUtils.ts'
+import { markdownArray, type PrComment } from './reportAsAPrComment.ts'
+import type { Test } from './constants.ts'
+import { TESTS } from './constants.ts'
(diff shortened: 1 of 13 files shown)


scripts/performance/lib/memoryPerformance.ts
```diff
@@ -175,3 +159,33 @@ function median(values: number[]): number {
...
+
+function formatMemoryPerformance({
+  baseMemoryPerformances,
+  localMemoryPerformances,
+}: {
+  baseMemoryPerformances: PerformanceMetric[]
+  localMemoryPerformances: PerformanceMetric[]
+}) {
+  return markdownArray({
+    headers: ['Action Name', 'Base Consumption Memory (bytes)', 'Local Consumption Memory (bytes)', '𝚫 (bytes)'],
```
thomas-lebeau: **🥜 nitpick:** ‏ 
- values are formatted and includes the unit, so `(bytes)` is unecessary
- The phrasing is a bit weird (maybe even incorrect), `Base memory consumption` would be better

 ```suggestion
    headers: ['Action Name', 'Base Memory Consumption', 'Local Memory Consumption', '𝚫'],
```

PR #3877, opened 2025-09-26
✨ [RUM-8780] Make developer extension capable of injecting the browser SDK

BeltranBulbarellaDD:
## Motivation

The idea is to be able to inject the SDK on a page that does not have it. Injecting the local version and CDN version.

## Changes

Load the extension dist folder, in chrome://extensions/. And on the SDK side go to developer-extension and `yarn dev`.
Disable the deployed Developer Extension to not have issues. 
Added under Datadog Employee mode, the possibility to inject using CDN the SDK on pages that do not have it.

## Test instructions

https://github.com/user-attachments/assets/dcc6799c-f09a-4612-9f7b-8061f84650f4

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/developer-extension/src/panel/hooks/useSettings.ts
+++ b/developer-extension/src/panel/hooks/useSettings.ts
@@ -48,17 +48,17 @@ async function loadSettingsFromStorage() {
 
 function setSetting<Name extends keyof Settings>(name: Name, value: Settings[Name]) {
   settings![name] = value
-  
+
   const settingsToStore: Partial<Settings> = { [name]: value }
-  
+
   // Reset injectionVariant to default when Datadog mode is disabled
   if (name === 'datadogMode' && value === false) {
     settings!.injectionVariant = DEFAULT_SETTINGS.injectionVariant
     settingsToStore.injectionVariant = DEFAULT_SETTINGS.injectionVariant
   }
-  
+
   onSettingsChange.notify()
-  
+
   chrome.storage.local
     .set(settingsToStore)
     .catch((error) => logger.error('Error while storing setting to the storage', error))


developer-extension/src/content-scripts/main.ts
```diff
@@ -180,3 +183,109 @@ function instrumentGlobal(global: 'DD_RUM' | 'DD_LOGS') {
...
+  }
+  if (site?.startsWith('us5.')) {
+    return 'us5'
+  }
+  if (site?.endsWith('datad0g.com')) {
+    return 'us3'
+  }
+
+  return 'us1'
+}
```
BeltranBulbarellaDD, line 261: I can probably remove this, the thing is that I was getting 403 so this was the fallback. 
Gonna look for a better way.
(Open to suggestions)

developer-extension/src/content-scripts/main.ts
```diff
@@ -184,3 +187,109 @@ function instrumentGlobal(global: 'DD_RUM' | 'DD_LOGS') {
...
+      const rumConfig =
+        settings.rumConfigurationOverride ||
+        (settings.datadogMode
+          ? {
+              applicationId: 'bd3472ea-efc2-45e1-8dff-be4cea9429b3',
+              clientToken: 'pub7216f8a2d1091e263c95c1205882474e',
+              site: 'datad0g.com',
+              allowedTrackingOrigins: [() => true],
+              sessionReplaySampleRate: 100,
+            }
```
BeltranBulbarellaDD, line 205: Because the SDK serializes the config and allowedTrackingOrigins is a function. It appears as Null or Empty Object, even though it works. Same happens with `allowedTracingUrls`

Screenshot 2025-10-07 at 10 26 26

Screenshot 2025-10-07 at 10 32 35

thomas-lebeau: it doesn't have to be a function though, you can do: `allowedTrackingOrigins: [location.origin]`

BeltranBulbarellaDD: Right you are! I was hesitant because if there was an issue injecting to the page I didn't want the SDK to stop. (But it does not make sense)

developer-extension/src/content-scripts/main.ts
```diff
@@ -184,3 +187,109 @@ function instrumentGlobal(global: 'DD_RUM' | 'DD_LOGS') {
...
+  const injectWhenReady = () => {
+    if (settings.sdkInjectionType === 'RUM' || settings.sdkInjectionType === 'BOTH') {
+      const rumSite = (settings.rumConfigurationOverride as any)?.site as string | undefined
+      const rumUrl = getRumBundleUrl(settings.useRumSlim ? 'rum-slim' : 'rum', rumSite)
+      const rumConfig =
+        settings.rumConfigurationOverride ||
+        (settings.datadogMode
+          ? {
+              applicationId: 'bd3472ea-efc2-45e1-8dff-be4cea9429b3',
+              clientToken: 'pub7216f8a2d1091e263c95c1205882474e',
```
mormubis: **💬 suggestion:** ‏I don't think we should publish any token here, even when employee mode is enabled. Can we just remove it and see the events in the dev tools?

BeltranBulbarellaDD: Totally right!

BeltranBulbarellaDD: Yeah, we need to leave site only to get events. Thats weird to be honest

developer-extension/src/content-scripts/main.ts
```diff
@@ -184,3 +187,109 @@ function instrumentGlobal(global: 'DD_RUM' | 'DD_LOGS') {
...
+  }
+  if (site?.startsWith('us3.')) {
+    return 'us3'
+  }
+  if (site?.startsWith('us5.')) {
+    return 'us5'
+  }
+  if (site?.endsWith('datad0g.com')) {
+    return 'us3'
+  }
```
BenoitZugmeyer, line 258: **🔨 warning:** ‏this seems incorrect

developer-extension/src/common/extension.types.ts
```diff
@@ -57,4 +60,6 @@ export interface Settings {
   logsConfigurationOverride: object | null
   debugMode: boolean
   datadogMode: boolean
+  injectionVariant: InjectionVariant
+  sdkInjectionType: SdkInjectionType
```
BenoitZugmeyer, line 64: **🔨 warning:** ‏The `settings` should be redesigned to make it more explicit and ideally avoid forbidden states. Things that should be clarified at the settings level:

* `useDevBundles: npm` won't work if `injectionVariant` is not `cdn`.
* How can I disable injection?
* Why injection variant CDN only available in `datadogMode`?
* What's the difference between `injectionVariant` and `sdkInjectionType` just by looking at property names?

My suggestion: keep things simple. Always inject both RUM and Logs CDN bundles with a default config. If dev bundle override is enabled, those bundles will be overridden (including rum-slim override). If config override is enabled, the config will be overridden. No need to do anything at the injection level.

So in the end you have a single option, `inject: boolean`. No need to enforce `datadogMode` usage, people can use it it's fine.

developer-extension/src/common/packagesUrlConstants.ts
```diff
@@ -3,6 +3,13 @@ export const DEV_LOGS_URL = `${DEV_SERVER_ORIGIN}/datadog-logs.js`
 export const DEV_RUM_SLIM_URL = `${DEV_SERVER_ORIGIN}/datadog-rum-slim.js`
 export const DEV_RUM_URL = `${DEV_SERVER_ORIGIN}/datadog-rum.js`
 
+export const CDN_BASE_URL = 'https://www.datadoghq-browser-agent.com'
+// This version corresponds to the major version of the Browser SDK and needs to be manually updated when bumping major versions
+export const CDN_VERSION = 'v6'
+export const CDN_RUM_URL = `${CDN_BASE_URL}/${CDN_VERSION}/datadog-rum.js`
+export const CDN_RUM_SLIM_URL = `${CDN_BASE_URL}/${CDN_VERSION}/datadog-rum-slim.js`
+export const CDN_LOGS_URL = `${CDN_BASE_URL}/${CDN_VERSION}/datadog-logs.js`
```
BenoitZugmeyer, line 11: Looks like those constants aren't used

developer-extension/src/content-scripts/main.ts
```diff
@@ -184,3 +187,109 @@ function instrumentGlobal(global: 'DD_RUM' | 'DD_LOGS') {
...
+
+function getRumBundleUrl(bundle: 'rum' | 'rum-slim', site?: string): string {
+  const region = getCdnRegion(site)
+  return `${CDN_BASE_URL}/${region}/${CDN_VERSION}/datadog-${bundle}.js`
+}
+
+function getLogsBundleUrl(site?: string) {
+  const region = getCdnRegion(site)
+  return `${CDN_BASE_URL}/${region}/${CDN_VERSION}/datadog-logs.js`
+}
```
BenoitZugmeyer, line 241: **💬 suggestion:** ‏Just include us1 prod rum and logs. No need to worry about regions, staging, etc.

BeltranBulbarellaDD, line 241: As per Benoit's good suggestion. This will be removed and use US1 prod.

developer-extension/src/panel/hooks/useSettings.ts
```diff
@@ -46,9 +48,19 @@ async function loadSettingsFromStorage() {
 
 function setSetting<Name extends keyof Settings>(name: Name, value: Settings[Name]) {
   settings![name] = value
+
+  const settingsToStore: Partial<Settings> = { [name]: value }
+
+  // Reset injectionVariant to default when Datadog mode is disabled
+  if (name === 'datadogMode' && value === false) {
```
BeltranBulbarellaDD, line 55: Not applicable anymore.

PR #3883, opened 2025-09-30
🐛 [RUM-11850] Remove anonymous Id when consent is not granted

cy-moi:
## Motivation
We want to remove the anonymous id from the cookie when consent is not granted.

## Changes
- Create the anonymous Id if it is not present only on `startSession`
- Pass in `hasConsent` boolean when the consent is `not-granted` and remove the anonymous Id accordingly
-  ( cr. to @BenoitZugmeyer )

## Test instructions
- Start local SDK 
- Retract consent in the console
- `anonymousId` should not present in cookie anymore

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
+++ b/packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
@@ -46,6 +46,7 @@ describe('session in cookie strategy', () => {
   })
 
   it('should not generate an anonymousId if not present', () => {
+    const cookieStorageStrategy = setupCookieStrategy()
     cookieStorageStrategy.persistSession(sessionState)
     const session = cookieStorageStrategy.retrieveSession()
     expect(session).toEqual({ id: '123', created: '0' })


packages/core/src/domain/session/sessionState.ts
```diff
@@ -18,16 +17,18 @@ export interface SessionState {
...
+  hasConsent: boolean = true
 ): SessionState {
   const expiredSessionState: SessionState = {
     isExpired: EXPIRED,
   }
+  if (!hasConsent) {
+    return expiredSessionState
+  }
   if (configuration.trackAnonymousUser) {
     if (previousSessionState?.anonymousId) {
```
BenoitZugmeyer: Suggestion: use a single condition:

```ts
 if (hasConsent && configuration.trackAnonymousUser && previousSessionState?.anonymousId) {
   expiredSessionState.anonymousId = previousSessionState.anonymousId
 }
```

packages/logs/src/domain/logsSessionManager.ts
```diff
@@ -35,7 +35,7 @@ export function startLogsSessionManager(
       return session && session.trackingType === LoggerTrackingType.TRACKED
         ? {
             id: session.id,
-            anonymousId: session.anonymousId,
+            anonymousId: trackingConsentState.isGranted() ? session.anonymousId : undefined,
```
BenoitZugmeyer: Question: why do we need this? also why we don't need this in RUM?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -231,10 +234,11 @@ export function startSessionStore<TrackingType extends string>(
...
     sessionStateUpdateObservable,
     restartSession: startSession,
-    expire: () => {
+    expire: (hasConsent?: boolean) => {
       cancelExpandOrRenewSession()
-      sessionStoreStrategy.expireSession(sessionCache)
-      synchronizeSession(getExpiredSessionState(sessionCache, configuration))
+      const expiredSessionState = getExpiredSessionState(sessionCache, configuration, hasConsent)
+      sessionStoreStrategy.expireSession(expiredSessionState)
+      synchronizeSession(expiredSessionState)
```
BenoitZugmeyer: Suggestion: instead of passing `hasConsent` to `getExpiredSessionState`, maybe just delete the anonymous id here? Also reduce the diff:
```suggestion
      if (!hasConsent && sessionCache) delete sessionCache.anonymousId
      sessionStoreStrategy.expireSession(sessionCache)
      synchronizeSession(getExpiredSessionState(sessionCache, configuration))
```

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -203,8 +206,8 @@ export function startSessionStore<TrackingType extends string>(
     return sessionCache.id !== sessionState.id || sessionCache[productKey] !== sessionState[productKey]
   }
 
-  function expireSessionInCache() {
-    sessionCache = getExpiredSessionState(sessionCache, configuration)
+  function expireSessionInCache(sessionState: SessionState) {
+    sessionCache = getExpiredSessionState(sessionState, configuration)
```
BenoitZugmeyer: Question: is this change relevant?

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -170,6 +170,9 @@ export function startSessionStore<TrackingType extends string>(
       {
         process: (sessionState) => {
           if (isSessionInNotStartedState(sessionState)) {
+            if (!sessionState.anonymousId) {
+              sessionState.anonymousId = generateUUID()
+            }
```
BenoitZugmeyer: Suggestion: `isSessionInNotStartedState` is equivalent to `isEmptyObject`, so the condition will always be true.. maybe remove it?

packages/core/src/domain/session/sessionStore.ts
cy-moi: It was mainly to update the cache. So we can use without this after deleting the anonymousId from `sessionCache`

packages/logs/src/domain/logsSessionManager.ts
cy-moi: Good catch! We don't need this. It was actually added by another tab when I mis-categorized to this.

packages/core/src/domain/session/sessionState.ts
```diff
@@ -18,17 +17,14 @@ export interface SessionState {
...
     isExpired: EXPIRED,
   }
-  if (configuration.trackAnonymousUser) {
-    if (previousSessionState?.anonymousId) {
-      expiredSessionState.anonymousId = previousSessionState?.anonymousId
-    } else {
-      expiredSessionState.anonymousId = generateUUID()
-    }
+  if (hasConsent && configuration.trackAnonymousUser && previousSessionState?.anonymousId) {
+    expiredSessionState.anonymousId = previousSessionState?.anonymousId
```
BenoitZugmeyer: Suggestion: remove the `hasConsent` parameter as it's not used anymore

thomas-lebeau (approved):

PR #3887, opened 2025-10-01
🎨 [PANA-4372] Explicitly scope serialization state

sethfowler-datadog:
## Motivation

I'm preparing to implement a prototype of a new DOM serialization algorithm behind a feature flag. To cleanly integrate this new algorithm into the existing code, we need to add a small amount of abstraction.

This PR focuses on adding an abstraction layer around node ids. Currently, the association between DOM nodes and their ids is tracked using global variables which persist across views; however, with the new algorithm, each view needs its own node id namespace. To abstract over this difference cleanly, it's helpful to get rid of the global variables and create an object to provide an explicit lifetime to this state.

## Changes

This PR adds a new `SerializationScope` type which tracks state associated with a specific stream of serialized events. It also removes the global variables we use right now to track node ids and moves them onto a new `NodeIds` type which is exposed via `SerializationScope`. The new `SerializationScope` type is plumbed to the various places in the code which need it.

There is not yet a mechanism for resetting the state associated with the `SerializationScope` when the view changes; in this PR, the focus is on maintaining existing behavior as much as possible, and in any case today's algorithm would not use this functionality. This mechanism will be added in a future PR.

Future PRs will also add more state to `SerializationScope`. Some of this will be associated with the new DOM serialization algorithm, but there is also existing state with similar lifetime requirements (e.g. `ElementScrollPositions`) that would probably make sense to fold into `SerializationScope`.

Note that while this PR is intended to maintain existing behavior, it _does_ actually change things slightly, because the lifetime of node ids is now tied to a particular invocation of `record()`. This means that stopping and starting recording will reset node ids. I think this is fine, and indeed desirable, but it's something to be aware of.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/record.ts
+++ b/packages/rum/src/domain/record/record.ts
@@ -23,6 +23,8 @@ import { initShadowRootsController } from './shadowRootsController'
 import { startFullSnapshots } from './startFullSnapshots'
 import { initRecordIds } from './recordIds'
 import type { SerializationStats } from './serialization'
+import { createSerializationScope } from './serialization'
+import { createNodeIds } from './nodeIds'
 
 export interface RecordOptions {
   emit?: (record: BrowserRecord, stats?: SerializationStats) => void
@@ -52,14 +54,20 @@ export function record(options: RecordOptions): RecordAPI {
   }
 
   const elementsScrollPositions = createElementsScrollPositions()
-
-  const shadowRootsController = initShadowRootsController(configuration, emitAndComputeStats, elementsScrollPositions)
+  const scope = createSerializationScope(createNodeIds())
+  const shadowRootsController = initShadowRootsController(
+    configuration,
+    scope,
+    emitAndComputeStats,
+    elementsScrollPositions
+  )
 
   const { stop: stopFullSnapshots } = startFullSnapshots(
     elementsScrollPositions,
     shadowRootsController,
     lifeCycle,
     configuration,
+    scope,
     flushMutations,
     emitAndComputeStats
   )
@@ -70,16 +78,16 @@ export function record(options: RecordOptions): RecordAPI {
   }
 
   const recordIds = initRecordIds()
-  const mutationTracker = trackMutation(emitAndComputeStats, configuration, shadowRootsController, document)
+  const mutationTracker = trackMutation(emitAndComputeStats, configuration, scope, shadowRootsController, document)
   const trackers: Tracker[] = [
     mutationTracker,
-    trackMove(configuration, emitAndComputeStats),
-    trackMouseInteraction(configuration, emitAndComputeStats, recordIds),
-    trackScroll(configuration, emitAndComputeStats, elementsScrollPositions, document),
+    trackMove(configuration, scope, emitAndComputeStats),
+    trackMouseInteraction(configuration, scope, emitAndComputeStats, recordIds),
+    trackScroll(configuration, scope, emitAndComputeStats, elementsScrollPositions, document),
     trackViewportResize(configuration, emitAndComputeStats),
-    trackInput(configuration, emitAndComputeStats),
-    trackMediaInteraction(configuration, emitAndComputeStats),
-    trackStyleSheet(emitAndComputeStats),
+    trackInput(configuration, scope, emitAndComputeStats),
+    trackMediaInteraction(configuration, scope, emitAndComputeStats),
+    trackStyleSheet(scope, emitAndComputeStats),
     trackFocus(configuration, emitAndComputeStats),
     trackVisualViewportResize(configuration, emitAndComputeStats),
     trackFrustration(lifeCycle, emitAndComputeStats, recordIds),
--- a/packages/rum/src/domain/record/serialization/serialization.types.ts
+++ b/packages/rum/src/domain/record/serialization/serialization.types.ts
@@ -1,6 +1,7 @@
 import type { RumConfiguration, NodePrivacyLevel } from '@datadog/browser-rum-core'
 import type { ElementsScrollPositions } from '../elementsScrollPositions'
 import type { ShadowRootsController } from '../shadowRootsController'
+import type { SerializationScope } from './serializationScope'
 import type { SerializationStats } from './serializationStats'
 
 // Those values are the only one that can be used when inheriting privacy levels from parent to
@@ -42,6 +43,5 @@ export interface SerializeOptions {
   parentNodePrivacyLevel: ParentNodePrivacyLevel
   serializationContext: SerializationContext
   configuration: RumConfiguration
+  scope: SerializationScope
 }
-
-export type NodeWithSerializedNode = Node & { s: 'Node with serialized node' }
--- a/packages/rum/src/domain/record/serialization/serializationScope.ts
+++ b/packages/rum/src/domain/record/serialization/serializationScope.ts
@@ -1,0 +1,9 @@
+import type { NodeIds } from '../nodeIds'
+
+export interface SerializationScope {
+  nodeIds: NodeIds
+}
+
+export function createSerializationScope(nodeIds: NodeIds): SerializationScope {
+  return { nodeIds }
+}
--- a/packages/rum/src/domain/record/serialization/serializeNode.ts
+++ b/packages/rum/src/domain/record/serialization/serializeNode.ts
@@ -20,7 +20,7 @@ import type {
   TextNode,
 } from '../../../types'
 import { NodeType } from '../../../types'
-import { getSerializedNodeId, getValidTagName, setSerializedNodeId } from './serializationUtils'
+import { getValidTagName } from './serializationUtils'
 import type { SerializeOptions } from './serialization.types'
 import { serializeStyleSheets } from './serializeStyleSheets'
 import { serializeAttributes } from './serializeAttributes'
@@ -31,20 +31,13 @@ export function serializeNodeWithId(node: Node, options: SerializeOptions): Seri
     return null
   }
 
-  // Try to reuse the previous id
-  const id = getSerializedNodeId(node) || generateNextId()
+  const id = options.scope.nodeIds.assign(node)
   const serializedNodeWithId = serializedNode as SerializedNodeWithId
   serializedNodeWithId.id = id
-  setSerializedNodeId(node, id)
   if (options.serializedNodeIds) {
     options.serializedNodeIds.add(id)
   }
   return serializedNodeWithId
-}
-
-let _nextId = 1
-export function generateNextId(): number {
-  return _nextId++
 }
 
 export function serializeChildNodes(node: Node, options: SerializeOptions): SerializedNodeWithId[] {
--- a/packages/rum/src/domain/record/trackers/trackMediaInteraction.ts
+++ b/packages/rum/src/domain/record/trackers/trackMediaInteraction.ts
@@ -4,14 +4,15 @@ import { NodePrivacyLevel, getNodePrivacyLevel } from '@datadog/browser-rum-core
 import type { BrowserIncrementalSnapshotRecord, MediaInteractionData } from '../../../types'
 import { IncrementalSource, MediaInteractionType } from '../../../types'
 import { getEventTarget } from '../eventsUtils'
-import { getSerializedNodeId, hasSerializedNode } from '../serialization'
 import { assembleIncrementalSnapshot } from '../assembly'
+import type { SerializationScope } from '../serialization'
 import type { Tracker } from './tracker.types'
 
 export type MediaInteractionCallback = (incrementalSnapshotRecord: BrowserIncrementalSnapshotRecord) => void
 
 export function trackMediaInteraction(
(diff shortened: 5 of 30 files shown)


packages/rum/src/domain/record/serialization/serializationScope.ts
```diff
@@ -0,0 +1,44 @@
+import { getParentNode, isNodeShadowRoot } from '@datadog/browser-rum-core'
+
+export type NodeWithSerializedNode = Node & { __brand: 'NodeWithSerializedNode' }
+export type NodeId = number & { __brand: 'NodeId' }
+
+export interface SerializationScope {
+  assignSerializedNodeId(node: Node): NodeId
+  getSerializedNodeId(node: Node): NodeId | undefined
```
sethfowler-datadog: There is no longer an overload of `getSerializedNodeId()` for `NodeWithSerializedNode`, and `hasSerializedNode()` has been removed. I find it cleaner to just call `getSerializedNodeId()` and check the result.

packages/rum/src/domain/record/serialization/serializationScope.ts
```diff
@@ -0,0 +1,44 @@
+import { getParentNode, isNodeShadowRoot } from '@datadog/browser-rum-core'
+
+export type NodeWithSerializedNode = Node & { __brand: 'NodeWithSerializedNode' }
+export type NodeId = number & { __brand: 'NodeId' }
+
+export interface SerializationScope {
+  assignSerializedNodeId(node: Node): NodeId
```
sethfowler-datadog: `SerializationScope` now encapsulates the logic for assigning ids to nodes, so `setSerializedNodeId()` has been removed from the interface.

packages/rum/src/domain/record/serialization/serializationScope.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function createSerializationScope(): SerializationScope {
+  const nodeIds = new WeakMap<Node, NodeId>()
+  let nextNodeId = NodeIds.FIRST
+
+  return {
+    assignSerializedNodeId: (node: Node): NodeId => {
+      // Try to reuse any existing id.
+      let nodeId = nodeIds.get(node)
```
yannickadam: :nit: we could use getSerializedNodeId() here.

packages/rum/src/domain/record/serialization/serialization.types.ts
```diff
@@ -43,6 +44,5 @@ export interface SerializeOptions {
   parentNodePrivacyLevel: ParentNodePrivacyLevel
   serializationContext: SerializationContext
   configuration: RumConfiguration
+  scope: SerializationScope
```
yannickadam, line 46: :nit: `scope` feels more fundamental than an option. Maybe it could be passed as a first level param?  I would also argue that mandatory options are not options anymore :)

packages/rum/src/domain/record/serialization/serializeNode.ts
```diff
@@ -31,22 +31,15 @@ export function serializeNodeWithId(node: Node, options: SerializeOptions): Seri
     return null
   }
 
-  // Try to reuse the previous id
-  const id = getSerializedNodeId(node) || generateNextId()
+  const id = options.scope.assignSerializedNodeId(node)
   const serializedNodeWithId = serializedNode as SerializedNodeWithId
   serializedNodeWithId.id = id
```
yannickadam, line 36: We should change the type of `SerializedNodeWithId.id` from `number` to `NodeID`? Else the branded type loses some value.

packages/rum/src/domain/record/trackers/trackMediaInteraction.ts
```diff
@@ -20,16 +21,19 @@ export function trackMediaInteraction(
...
-        !hasSerializedNode(target)
+        id === undefined ||
+        getNodePrivacyLevel(target, configuration.defaultPrivacyLevel) === NodePrivacyLevel.HIDDEN
       ) {
         return
       }
       mediaInteractionCb(
         assembleIncrementalSnapshot<MediaInteractionData>(IncrementalSource.MediaInteraction, {
-          id: getSerializedNodeId(target),
+          id,
```
yannickadam, line 36: Can `assembleIncrementalSnapshot` accept only an `id` of type `NodeID`, or does our link with rum events format prevents this?

packages/rum/src/domain/record/trackers/trackStyleSheet.ts
```diff
@@ -1,19 +1,24 @@
...
-export function trackStyleSheet(styleSheetCb: StyleSheetCallback): Tracker {
+export function trackStyleSheet(scope: SerializationScope, styleSheetCb: StyleSheetCallback): Tracker {
   function checkStyleSheetAndCallback(styleSheet: CSSStyleSheet | null, callback: (id: number) => void): void {
-    if (styleSheet && hasSerializedNode(styleSheet.ownerNode!)) {
-      callback(getSerializedNodeId(styleSheet.ownerNode))
+    if (!styleSheet || !styleSheet.ownerNode) {
+      return
     }
+    const id = scope.getSerializedNodeId(styleSheet.ownerNode)
+    if (id === undefined) {
```
yannickadam, line 18: Looks like we had a risk of executing the callback with `undefined` before. I've seen a couple places where the new code looks more defensive. I wonder => Are we just being extra cautious or do we believe these cases could happen?

yannickadam (approved):

packages/rum/src/domain/record/serialization/serializeNode.ts
sethfowler-datadog: I'd love to! It's a multistep process, though, since `SerializedNodeWithId` is a type generated from JSON Schema which lives in `rum-events-format`. We can probably use the `tsType` feature of `json-schema-to-typescript` to apply the branding there. I'll give it a try in a future PR in `rum-events-format`, and if everything looks good there, I'll update the code here to take advantage of it.

packages/rum/src/domain/record/serialization/serialization.types.ts
sethfowler-datadog: I feel the same way! In the version of this code for the new data format, I've totally eliminated this options object and just converted everything to actual function arguments. However, I did that in part by adding one more abstraction which isn't yet present in this PR. I can make the same changes for the old serialization algorithm, but if it's OK with you, let's defer that until the PR that adds that other abstraction. (It's an object representing a mutation "transaction", which is what a lot of these "options" are really associated with.)

packages/rum/src/domain/record/trackers/trackMediaInteraction.ts
sethfowler-datadog: As you say, the difficulty is that types from `rum-events-format` are used here. It should be possible using `tsType`, but it'll require some work in the context of that repo.

packages/rum/src/domain/record/trackers/trackStyleSheet.ts
sethfowler-datadog: You're right; it's not super likely for us to encounter an `undefined` node id in these situations today because node ids have an unbounded lifetime. As we transition to a world where node ids are a bit more scoped, and may get periodically reset, a slight increase in caution is probably warranted.

packages/rum/src/domain/record/serialization/serializationScope.ts
sethfowler-datadog: I modified both this function and `nodeAndAncestorsHaveSerializedNode()` to use `getSerializedNodeId()`!

packages/rum/src/domain/record/serialization/serialization.types.ts
sethfowler-datadog: Apropos of this discussion, this PR removes one of these options totally: https://github.com/DataDog/browser-sdk/pull/3889

BenoitZugmeyer: 2 cents: Options is used to designate named parameters here. I agree it might not be the best naming.

packages/rum/src/domain/record/record.ts
```diff
@@ -52,14 +53,20 @@ export function record(options: RecordOptions): RecordAPI {
   }
 
   const elementsScrollPositions = createElementsScrollPositions()
-
-  const shadowRootsController = initShadowRootsController(configuration, emitAndComputeStats, elementsScrollPositions)
+  const scope = createSerializationScope()
```
BenoitZugmeyer: Question: are you going to reset the scope on each full snapshot?

packages/rum/src/domain/record/serialization/serialization.types.ts
```diff
@@ -42,6 +43,5 @@ export interface SerializeOptions {
...
```
BenoitZugmeyer, line 46: Suggestion: One might wonder what's the distinction between the serialization context and scope. Maybe rename scope to something more explicit like `nodeRegistry`? Or rename context to "state"? Or maybe put the scope in the context?

packages/rum/src/domain/record/serialization/serializationScope.ts
```diff
@@ -0,0 +1,9 @@
+import type { NodeIds } from '../nodeIds'
+
+export interface SerializationScope {
+  nodeIds: NodeIds
+}
+
```
BenoitZugmeyer, line 6: Question: just for my understanding, what are you planning to add in the `scope`?

packages/rum/src/domain/record/record.ts
sethfowler-datadog: Yes, that's right. Today, and in the near term future, that also means that we reset the scope on each view. (Longer term, I'm hoping that we end up being able to separate these two concepts.)

packages/rum/src/domain/record/serialization/serialization.types.ts
sethfowler-datadog: I want to do this kind of simplification, but I'm hoping to introduce one more concept first: an object representing a serialization "transaction". The general structure I'm thinking of this:
* `SerializationScope` contains stuff which has a lifetime tied to a particular stream of serialized events. For today's serialization algorithm, that's node ids and statistics. The new algorithm also has the string table at this level.
* `SerializationScope` has a method with a signature like `captureMutations((transaction: SerializationTransaction) => BrowserRecord)`. We capture full snapshots, incremental mutations, and the like inside the callback to this method. It handles some generic things, like collecting timing statistics for telemetry, initializing the state of the `SerializationTransaction`, and passing the resulting record to a callback so it can be included in the segment. The new algorithm will do a bit of post-processing at this level, too.
* `SerializationTransaction` contains stuff which has a lifetime tied to a particular invocation of the serialization code. Think of things like `SerializeOptions#serializedNodeIds`, `SerializationContext#status`, etc. It will also have a reference to the `SerializationScope` that generated it, so we don't have to thread both objects through the call tree everywhere. The new algorithm will have a version of `SerializationTransaction` with builder-like methods that add new content to the transaction (e.g. new serialized nodes, new serialized attribute mutations, etc etc), so that we will have a single place where we handle the "smart" aspects of the encoding, like converting literal strings into references to the string table.

I'm planning to introduce the skeleton of this structure in the next PR in the series, which I'll push today. Once we have this in place, we can start moving things from `SerializationContext` and `SerializationOptions` to new homes based on their lifetime: some will end up on `SerializationScope`, some on `SerializationTransaction`. `SerializeOptions#parentNodePrivacyLevel` will get converted to a normal function parameter, since it's an exceptional case that doesn't really belong on either of these objects; its natural scope is much more local.

I'm hoping this structure will simplify things a little bit overall, while also providing a foundation that both the old and new algorithms can sit on top of cleanly, despite their differences.

packages/rum/src/domain/record/serialization/serializationScope.ts
sethfowler-datadog: See above!

packages/rum/src/domain/record/serialization/serialization.types.ts
BenoitZugmeyer: That sounds great, and I'm looking forward to that next PR. Clarifying the different concepts with some comments could be nice!

BenoitZugmeyer (approved):

packages/rum/src/domain/record/serialization/serialization.types.ts
sethfowler-datadog: Yeah, that's totally fair; naming is always the hard part. 😄 

I just pushed #3893 (based on this one, to avoid further rebasing) that pushes further forward on restructuring these options by converting `parentNodePrivacyLevel` into a normal function parameter. I'll open more PRs to try to further improve the organization of these options.

packages/rum/src/domain/record/serialization/serialization.types.ts
sethfowler-datadog: Yeah, totally agreed! I'll make sure that PR includes some solid comments.

PR #3893, opened 2025-10-03
🎨 [PANA-4398] Convert SerializeOptions#parentNodePrivacyLevel into a normal function argument

sethfowler-datadog:
## Motivation

`SerializeOptions#parentNodePrivacyLevel` doesn't really make sense as member of `SerializeOptions`; as the name implies, it's a value that's only applicable for a particular node's children. This scope mismatch has led to the implementation of some optimizations in the code that try to avoid copying `SerializeOptions` constantly during serialization, but the more natural fix is simply to pass this value around as a normal function parameter.

## Changes

This PR converts `SerializeOptions#parentNodePrivacyLevel` into a normal function parameter, updates all call sites, and removes the now-unnecessary code for the optimization described above.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/serialization/serialization.types.ts
+++ b/packages/rum/src/domain/record/serialization/serialization.types.ts
@@ -7,7 +7,7 @@ import type { SerializationStats } from './serializationStats'
 // Those values are the only one that can be used when inheriting privacy levels from parent to
 // children during serialization, since HIDDEN and IGNORE shouldn't serialize their children. This
 // ensures that no children are serialized when they shouldn't.
-export type ParentPrivacyLevel =
+export type ParentNodePrivacyLevel =
   | typeof NodePrivacyLevel.ALLOW
   | typeof NodePrivacyLevel.MASK
   | typeof NodePrivacyLevel.MASK_USER_INPUT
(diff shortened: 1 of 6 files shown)


packages/rum/src/domain/record/serialization/serialization.types.ts
```diff
@@ -7,7 +7,7 @@ import type { SerializationStats } from './serializationStats'
 // Those values are the only one that can be used when inheriting privacy levels from parent to
 // children during serialization, since HIDDEN and IGNORE shouldn't serialize their children. This
 // ensures that no children are serialized when they shouldn't.
-type ParentNodePrivacyLevel =
+export type ParentPrivacyLevel =
```
yannickadam: I think I prefered the old name? Any reason why you removed `node`?

yannickadam (approved):

packages/rum/src/domain/record/serialization/serialization.types.ts
sethfowler-datadog: The "Node" part felt clear from context to me, but I don't feel strongly; happy to change it back. I'll push a new commit.

packages/rum/src/domain/record/serialization/serializeNode.ts
```diff
@@ -156,19 +177,7 @@ function serializeElementNode(element: Element, options: SerializeOptions): Elem
...
-    // the same object as much as possible, and clone it only if we need to.
-    let childNodesSerializationOptions
-    if (options.parentNodePrivacyLevel === nodePrivacyLevel) {
-      childNodesSerializationOptions = options
-    } else {
-      childNodesSerializationOptions = {
-        ...options,
-        parentNodePrivacyLevel: nodePrivacyLevel,
-      }
-    }
```
BenoitZugmeyer, line 170: Beautiful.

PR #3901, opened 2025-10-06
👷 change e2e implementation for workers

mormubis:
## Motivation

The initial implementation of Worker E2E tests was somewhat tied to our specific use case and will not scale if we need to test different scenarios within the Worker environment.

## Changes

It introduces a version of `withWorker` that automatically sets up the logs inside the worker. We can also pass an implementation, similar to what we do with `withLogs` and `withBody`.

## Test instructions

All e2e still passed.

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/lib/framework/createTest.ts
+++ b/test/e2e/lib/framework/createTest.ts
@@ -13,8 +13,8 @@ import { IntakeRegistry } from './intakeRegistry'
 import { flushEvents } from './flushEvents'
 import type { Servers } from './httpServers'
 import { getTestServers, waitForServersIdle } from './httpServers'
-import type { SetupFactory, SetupOptions, WorkerImplementationFactory } from './pageSetups'
-import { workerSetup, html, DEFAULT_SETUPS, npmSetup, reactSetup } from './pageSetups'
+import type { SetupFactory, SetupOptions, WorkerOptions } from './pageSetups'
+import { buildWorkerScript, html, DEFAULT_SETUPS, npmSetup, reactSetup } from './pageSetups'
 import { createIntakeServerApp } from './serverApps/intake'
 import { createMockServerApp } from './serverApps/mock'
 import type { Extension } from './createExtension'
@@ -59,7 +59,7 @@ class TestBuilder {
     logsConfiguration?: LogsInitConfiguration
   } = {}
   private hostName?: string
-  private workerImplementationFactory?: WorkerImplementationFactory
+  private workerOptions?: WorkerOptions
 
   constructor(private title: string) {}
 
@@ -131,15 +131,15 @@ class TestBuilder {
     return this
   }
 
-  withWorker(implementation: WorkerImplementationFactory['implementation'], options: RegistrationOptions = {}) {
-    this.workerImplementationFactory = { ...options, implementation }
+  withWorker(testCase: WorkerOptions['testCase'], registrationOptions: RegistrationOptions = {}) {
+    this.workerOptions = { ...registrationOptions, testCase }
 
     // Service workers require HTTPS or localhost due to browser security restrictions
     this.withHostName('localhost')
     this.withBody(html`
       <script>
         if (!window.myServiceWorker && 'serviceWorker' in navigator) {
-          navigator.serviceWorker.register('/sw.js', ${JSON.stringify(options)}).then((registration) => {
+          navigator.serviceWorker.register('/sw.js', ${JSON.stringify(registrationOptions)}).then((registration) => {
             window.myServiceWorker = registration
           })
         }
@@ -173,7 +173,7 @@ class TestBuilder {
       testFixture: this.testFixture,
       extension: this.extension,
       hostName: this.hostName,
-      workerImplementationFactory: this.workerImplementationFactory,
+      workerOptions: this.workerOptions,
     }
 
     if (this.alsoRunWithRumSlim) {
@@ -253,7 +253,7 @@ function declareTest(title: string, setupOptions: SetupOptions, factory: SetupFa
     servers.base.bindServerApp(
       createMockServerApp(servers, setup, {
         remoteConfiguration: setupOptions.remoteConfiguration,
-        workerImplementation: setupOptions.workerImplementationFactory && workerSetup(setupOptions, servers),
+        workerScript: buildWorkerScript(setupOptions, servers),
       })
     )
     servers.crossOrigin.bindServerApp(createMockServerApp(servers, setup))
--- a/test/e2e/lib/framework/pageSetups.ts
+++ b/test/e2e/lib/framework/pageSetups.ts
@@ -6,8 +6,8 @@ import { DEFAULT_LOGS_CONFIGURATION } from '../helpers/configuration'
 import { isBrowserStack, isContinuousIntegration } from './environment'
 import type { Servers } from './httpServers'
 
-export interface WorkerImplementationFactory extends RegistrationOptions {
-  implementation: (self: WorkerGlobalScope & { DD_LOGS?: DatadogLogs }) => void
+export interface WorkerOptions extends RegistrationOptions {
+  testCase: (self: WorkerGlobalScope & { DD_LOGS?: DatadogLogs }) => void
 }
 export interface SetupOptions {
   rum?: RumInitConfiguration
@@ -30,7 +30,7 @@ export interface SetupOptions {
     logsConfiguration?: LogsInitConfiguration
   }
   hostName?: string
-  workerImplementationFactory?: WorkerImplementationFactory
+  workerOptions?: WorkerOptions
 }
 
 export type SetupFactory = (options: SetupOptions, servers: Servers) => string
@@ -208,23 +208,23 @@ export function reactSetup(options: SetupOptions, servers: Servers, appName: str
   })
 }
 
-export function workerSetup(options: SetupOptions, servers: Servers) {
+export function buildWorkerScript(options: SetupOptions, servers: Servers) {
   let script = ''
 
-  if (!options.workerImplementationFactory) {
+  if (!options.workerOptions) {
     return script
   }
 
   if (options.logs) {
     script += js`
-      ${options.workerImplementationFactory.type === 'module' ? js`import '/datadog-logs.js';` : js`importScripts('/datadog-logs.js');`}
+      ${options.workerOptions.type === 'module' ? js`import '/datadog-logs.js';` : js`importScripts('/datadog-logs.js');`}
       
       // Initialize DD_LOGS in service worker
       DD_LOGS.init(${formatConfiguration({ ...DEFAULT_LOGS_CONFIGURATION, ...options.logs }, servers)})
     `
   }
 
-  script += `;(${options.workerImplementationFactory.implementation.toString()})(self);`
+  script += `;(${options.workerOptions.testCase.toString()})(self);`
 
   return script
 }
--- a/test/e2e/lib/framework/serverApps/mock.ts
+++ b/test/e2e/lib/framework/serverApps/mock.ts
@@ -11,12 +11,12 @@ export const LARGE_RESPONSE_MIN_BYTE_SIZE = 100_000
 
 interface MockServerOptions {
   remoteConfiguration?: RemoteConfiguration
-  workerImplementation?: string
+  workerScript?: string
 }
 export function createMockServerApp(
   servers: Servers,
   setup: string,
-  { remoteConfiguration, workerImplementation }: MockServerOptions = {}
+  { remoteConfiguration, workerScript }: MockServerOptions = {}
 ): MockServerApp {
   const app = express()
   let largeResponseBytesWritten = 0
@@ -49,7 +49,7 @@ export function createMockServerApp(
   })
 
   app.get('/sw.js', (_req, res) => {
-    res.contentType('application/javascript').send(workerImplementation)
+    res.contentType('application/javascript').send(workerScript)
   })
 
   function generateLargeResponse(res: ServerResponse, chunkText: string) {


test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -6,14 +6,19 @@ import type { RemoteConfiguration } from '@datadog/browser-rum-core'
...
 import { DEV_SERVER_BASE_URL } from '../../helpers/playwright'
-import { workerSetup } from '../pageSetups'
 
 export const LARGE_RESPONSE_MIN_BYTE_SIZE = 100_000
 
 export function createMockServerApp(
   servers: Servers,
   setup: string,
-  remoteConfiguration?: RemoteConfiguration
+  {
```
BeltranBulbarellaDD: Wouldn't an interface be more readable?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -13,7 +13,21 @@ declare global {
 
 test.describe('logs', () => {
   createTest('service worker with worker logs - esm')
-    .withWorker()
+    .withLogs()
+    .withWorker(
```
BeltranBulbarellaDD, line 17: Would it be possible to move this to the createTest? So that we do something like:
`.withWorker({ isModule: true, handleMessageWith: 'logger' })`
Similar to what we do in: https://github.com/DataDog/browser-sdk/blob/35fd80db64450d58483af7c8a8530be377e51a71/test/e2e/scenario/browser-extensions/browserExtensions.scenario.ts#L22

mormubis: I think the idea is to avoid that scenario and be more generic, or that's what I understood from @bcaudan feedback.

test/e2e/lib/framework/serverApps/mock.ts
mormubis: I could definitely transform it to a type/interface.

test/e2e/scenario/logs.scenario.ts
bcaudan: I think we could make the API more convenient by:

- providing a function to `withWorker`, to work directly with code instead of a block of text
- have the worker implementation as complete as possible:
  - let it import the script rather than handle that in workerSetup, it would be easier to follow
  - provide it the consolidated configuration

to be able to do something like:

```js
.withWorker((config) => {
  import '/datadog-logs.js'
      
  // Initialize DD_LOGS in service worker
  DD_LOGS.init(config)

  // Handle messages from main thread
  self.addEventListener('message', (event) => {
    const message = event.data
    DD_LOGS.logger.log(message)
  })
})
```

test/e2e/scenario/logs.scenario.ts
```diff
@@ -13,7 +13,21 @@ declare global {
...
+      (setup) => js`
+        // Initialize DD_LOGS in service worker
+        ${setup}
+
+        // Handle messages from main thread
+        self.addEventListener('message', (event) => {
+          const message = event.data;
+
+          DD_LOGS.logger.log(message);
+        });
```
bcaudan: **❓ question:** ‏Do we need to use post message for every test? Could we instead have the log needed for the test be triggered when the worker is started from the worker implementation?
It would reduce the complexity of those tests

mormubis: @bcaudan I thought about it, and it's a bit tricky, because it means the test it auto-executes it. Are you ok with that?

bcaudan: From my understanding the test would instantiate the worker and this instantiation could generate the log use case that we want to test, avoiding the extra message exchanges with the main thread.
Was there any issue that this message exchange was preventing or any other issue that you see with the "auto-execution"?
if not, I'd say it is worth a try.

mormubis: Done!

test/e2e/scenario/logs.scenario.ts
mormubis: @bcaudan I had this as an earlier version, but then I found that the current implementation is closer to what we do with Logs/RUM (`withLogs` and `withRUM`).

test/e2e/scenario/logs.scenario.ts
bcaudan: Could we do it with other tests as well?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -13,7 +13,17 @@ declare global {
...
+    .withWorker(
+      function (self) {
+        self.addEventListener('message', (event) => {
+          const message = (event as MessageEvent<string>).data
+
+          self.DD_LOGS!.logger.log(message)
+        })
+      },
+      { type: 'module' }
+    )
```
bcaudan, line 22: **💬 suggestion:** ‏I think we could factorize the message communication used in every tests, that would remove some boilerplate and make the tests more focus on their use cases:

something like:
```js
.withWorker({ type: 'module' })
.withLogs()
.run(async ({ flushEvents, intakeRegistry, browserName, interactWithWorker }) => {
  await evaluateInWorker(() => {
    self.DD_LOGS!.logger.log('Some message')
  })
   
  await flushEvents()
  
  expect(intakeRegistry.logsRequests).toHaveLength(1)
  expect(intakeRegistry.logsEvents[0].message).toBe('Some message')
})

Comment (mormubis) [reply test/e2e/scenario/logs.scenario.ts:22]: Updated!

File: test/e2e/tsconfig.json [hunk 5]
```diff
@@ -10,7 +10,9 @@
     "target": "ES2022",
     "module": "ES2020",
     "types": ["node", "ajv"],
+    "lib": ["ESNext", "DOM", "WebWorker"],
     "allowJs": true,
     "noEmit": true
-  }
+  },
+  "include": ["."]
```
bcaudan: **💬 suggestion:** ‏that seems superfluous

```suggestion
  }
```

test/e2e/lib/types/global.ts
```diff
@@ -1,9 +1,9 @@
...
+import type { DatadogRum } from '@datadog/browser-rum'
 
 declare global {
   interface Window {
-    DD_LOGS?: LogsGlobal
-    DD_RUM?: RumGlobal
+    DD_LOGS?: DatadogLogs
+    DD_RUM?: DatadogRum
   }
 }
```
bcaudan, line 10: **👏 praise:** ‏Nice!

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -27,6 +32,7 @@ export interface SetupOptions {
     logsConfiguration?: LogsInitConfiguration
   }
   hostName?: string
+  workerImplementation?: WorkerImplementationFactory
 }
 
 export interface WorkerOptions {
```
bcaudan: **💬 suggestion:** ‏`WorkerOptions` does not seem to be used anymore

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -1,11 +1,16 @@
...
+import type { LogsInitConfiguration, DatadogLogs } from '@datadog/browser-logs'
 import type { RumInitConfiguration, RemoteConfiguration } from '@datadog/browser-rum-core'
+import type { DatadogRum } from '@datadog/browser-rum'
 import type test from '@playwright/test'
 import { DEFAULT_LOGS_CONFIGURATION } from '../helpers/configuration'
 import { isBrowserStack, isContinuousIntegration } from './environment'
 import type { Servers } from './httpServers'
 
+export interface WorkerImplementationFactory {
+  (self: WorkerGlobalScope & { DD_LOGS?: DatadogLogs; DD_RUM?: DatadogRum }): void
```
bcaudan: **💬 suggestion:** ‏as long as we don't support rum in the worker, we should not need DD_RUM there 

```suggestion
  (self: WorkerGlobalScope & { DD_LOGS?: DatadogLogs }): void
```

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -244,7 +255,7 @@ export function html(parts: readonly string[], ...vars: string[]) {
   return parts.reduce((full, part, index) => full + vars[index - 1] + part)
 }
 
-function js(parts: readonly string[], ...vars: string[]) {
+export function js(parts: readonly string[], ...vars: string[]) {
```
bcaudan: **💬 suggestion:** ‏the implementation is actually a duplicate of html, what about aliasing it instead

```js
export const js = html
```

test/e2e/lib/framework/createTest.ts
```diff
@@ -135,37 +132,21 @@ class TestBuilder {
...
-            navigator.serviceWorker.register('${url}', ${options}).then((registration) => {
-              window.myServiceWorker = registration
-            })
-          }
-        </script>
-      `)
-    }
+  withWorker(implementation: WorkerImplementationFactory, options: RegistrationOptions = {}) {
+    implementation.isModule = options.type === 'module'
+    this.workerImplementationFactory = implementation
```
bcaudan: **💬 suggestion:** ‏the data structure with implementation, factory, RegistrationOptions and the option later attached to the implementation seems more complex that it needs to be.

What about having a single `WorkerOptions` type used everywhere that could either:
- extend `RegistrationOptions` with the `implementation`
- contain the `implementation` and the `isModule` boolean

test/e2e/lib/framework/createTest.ts
```diff
@@ -270,7 +252,12 @@ function declareTest(title: string, setupOptions: SetupOptions, factory: SetupFa
     servers.intake.bindServerApp(createIntakeServerApp(testContext.intakeRegistry))
 
     const setup = factory(setupOptions, servers)
-    servers.base.bindServerApp(createMockServerApp(servers, setup, setupOptions.remoteConfiguration))
+    servers.base.bindServerApp(
+      createMockServerApp(servers, setup, {
+        remoteConfiguration: setupOptions.remoteConfiguration,
+        workerImplementation: setupOptions.workerImplementation && workerSetup(setupOptions, servers),
```
bcaudan: 🥜 nitpick: ‏it could be less confusing to have two different namings for the implementation function and the implementation content. Then, we could rename workerSetup to align with the new naming.

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -209,20 +215,25 @@ export function reactSetup(options: SetupOptions, servers: Servers, appName: str
   })
 }
 
-export function workerSetup(options: WorkerOptions, servers: Servers) {
-  return js`
-      ${options.importScripts ? js`importScripts('/datadog-logs.js');` : js`import '/datadog-logs.js';`}
+export function workerSetup(options: SetupOptions, servers: Servers) {
```
bcaudan: **💭 thought:** ‏as this helper is not really related to the page setup, maybe we could move it to createTest to colocate it with other worker setup code and avoid the back and forth between those files.

test/e2e/lib/framework/createTest.ts
```diff
@@ -135,37 +132,21 @@ class TestBuilder {
...
+    this.workerImplementationFactory = implementation
+
+    // Service workers require HTTPS or localhost due to browser security restrictions
+    this.withHostName('localhost')
+    this.withBody(html`
+      <script>
+        if (!window.myServiceWorker && 'serviceWorker' in navigator) {
+          navigator.serviceWorker.register('/sw.js', ${JSON.stringify(options)}).then((registration) => {
+            window.myServiceWorker = registration
+          })
```
bcaudan: **💬 suggestion:** ‏if we don't plan to use `interactWithWorker` in the coming tests, we could remove it  and the `window.myServiceWorker` reference.

mormubis: Removed, I kept the `window.myServiceWorker` to avoid double registering.

test/e2e/lib/framework/createTest.ts
mormubis: Renamed it

test/e2e/lib/framework/createTest.ts
mormubis: Updated

test/e2e/lib/framework/pageSetups.ts
mormubis: Done!

test/e2e/lib/framework/pageSetups.ts
mormubis: Updated!

test/e2e/lib/framework/pageSetups.ts
mormubis: Remvoed

test/e2e/tsconfig.json
mormubis: Removed

test/e2e/lib/framework/createTest.ts
bcaudan: How can this happen?

test/e2e/lib/framework/createTest.ts
bcaudan: We still have the same naming (`workerImplementation`) for two differents things though 😕

Counter-proposal on namings:

- `workerSetup()` ->  `buildWorkerScript()`, script naming is used internally and seems clear enough to me
- `WorkerImplementationFactory` -> `WorkerOptions`, as this object is not a factory
- `WorkerOptions.implementation` -> `WorkerOptions.testCase`, as it is only the code relative to the current test

The updates I played with: https://github.com/DataDog/browser-sdk/commit/a99bdbe5991bb2d7c687aa0b2251c7f8cf12f5e8

wdyt?

test/e2e/lib/framework/createTest.ts
```diff
@@ -270,7 +250,12 @@ function declareTest(title: string, setupOptions: SetupOptions, factory: SetupFa
     servers.intake.bindServerApp(createIntakeServerApp(testContext.intakeRegistry))
 
     const setup = factory(setupOptions, servers)
-    servers.base.bindServerApp(createMockServerApp(servers, setup, setupOptions.remoteConfiguration))
+    servers.base.bindServerApp(
+      createMockServerApp(servers, setup, {
+        remoteConfiguration: setupOptions.remoteConfiguration,
+        workerImplementation: setupOptions.workerImplementationFactory && workerSetup(setupOptions, servers),
```
bcaudan: **💬 suggestion:** ‏the check is also done inside the function, so we could probably remove it from here


```suggestion
        workerImplementation: workerSetup(setupOptions, servers),
```

test/e2e/lib/framework/createTest.ts
mormubis: Someone executing `withWorker` multiple times.

PR #3904, opened 2025-10-07
fix: block logs in node

mormubis:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


test/apps/vanilla/app.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
mormubis: Line 27 is reporting `@typescript-eslint/no-unsafe-call` even when I disable the line 😕

test/apps/vanilla/app.ts
```diff
@@ -17,8 +19,16 @@ if (typeof window !== 'undefined') {
     window.RUM_INIT()
   }
 } else {
+  // Document is always generated by the SSR environment
+  // @ts-ignore If document is mocked, the SDK executes further engouh to throw an error
+  globalThis.document = {}
+
+  // @ts-ignore Force throw errors
+  datadogLogs._setDebug(true)
```
thomas-lebeau: **❓ question:** ‏How is this forcing to throw errors? it is supposed to log the errors caught by `monitor()` in the console

packages/core/src/tools/globalObject.ts
```diff
@@ -48,4 +48,6 @@ export function getGlobalObject<T = typeof globalThis>(): T {
 // eslint-disable-next-line local-rules/disallow-side-effects
 export const globalObject = getGlobalObject<GlobalObject>()
 
+export const isBrowserEnvironment = 'document' in globalObject
```
cy-moi, line 51: ❓ Question: we do not prevent the cases when `document` is mocked?

test/apps/vanilla/app.ts
mormubis: I've changed the command to throw when printing anything on the screen.

packages/core/src/tools/globalObject.ts
mormubis: I updated the node check! Now we are checking for `typeof process`. That should give us the clue we need, instead of relying on `isBrowserEnvironment`.

test/apps/vanilla/app.ts
thomas-lebeau: hmm, that seems brittle as it rely on the implementation to throw anything at one point. If for any reason the SDK start successfully without any issues, then the test will pass, no?

can't we do something like?:
```
if (datadogLogs.getInternalContext()) {
  throw new Error('SDK should not start on SSR environments')
}
```

thomas-lebeau (approved):

PR #3905, opened 2025-10-07
♻️ encode cookie options in cookie value

thomas-lebeau:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -53,16 +56,54 @@ function storeSessionCookie(
...
+  /* eslint-disable no-bitwise */
+  let byte = 0
+  byte |= SESSION_COOKIE_VERSION << 6 // Store version in upper 2 bits
+  byte |= domainCount << 2 // Store domain count in next 4 bits
+  byte |= configuration.usePartitionedCrossSiteSessionCookie ? 1 : 0 << 1 // Store useCrossSiteScripting in next bit
+  // there is one bit left for future use
+  /* eslint-enable no-bitwise */
+
+  return byte.toString(16) // Convert to hex string
+}
```
thomas-lebeau: This is a starting point, but this is not set in stone in any way. Open to suggestion.

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -83,3 +106,41 @@ export function buildCookieOptions(initConfiguration: InitConfiguration): Cookie
...
 }
+
+function encodeCookieOptions(configuration: Configuration, cookieOptions: CookieOptions): string {
+  const domainCount = cookieOptions.domain ? cookieOptions.domain.split('.').length - 1 : 0
+
+  /* eslint-disable no-bitwise */
+  let byte = 0
+  byte |= SESSION_COOKIE_VERSION << 6 // Store version in upper 2 bits
+  byte |= domainCount << 2 // Store domain count in next 4 bits
+  byte |= configuration.usePartitionedCrossSiteSessionCookie ? 1 : 0 << 1 // Store useCrossSiteScripting in next bit
```
BenoitZugmeyer: Nit:

```suggestion
  byte |= configuration.usePartitionedCrossSiteSessionCookie ? 1 : 0 // Store useCrossSiteScripting in next bit
```

Suggestion: FMU we are using 7 bits on the 8 available bits. We might want to give the version one more bit, so we can have up to 7 versions?

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
BenoitZugmeyer: I like it!

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -83,3 +106,41 @@ export function buildCookieOptions(initConfiguration: InitConfiguration): Cookie
...
+  for (const cookie of cookies.reverse()) {
+    sessionState = toSessionState(cookie)
+
+    if (sessionState.c === opts) {
+      break
+    }
+  }
+
+  // remove the cookie options from the session state
+  delete sessionState?.c
```
BenoitZugmeyer, line 144: Suggestion: this might not be necessary?

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -83,3 +106,41 @@ export function buildCookieOptions(initConfiguration: InitConfiguration): Cookie
...
+
+  return byte.toString(16) // Convert to hex string
+}
+
+function retrieveSessionCookieFromEncodedCookie(
+  configuration: Configuration,
+  cookieOptions: CookieOptions
+): SessionState {
+  const cookies = getCookies(SESSION_STORE_KEY) ?? []
+  const opts = encodeCookieOptions(configuration, cookieOptions)
```
BenoitZugmeyer: Nit: this could be a oneliner (really nitpicking here, do as you prefer)

```ts
return findLast(
  getCookies(SESSION_STORE_KEY) ?? []).map(toSessionState),
  (sessionState) => sessionState.c === opts
) || {}
```

packages/core/src/browser/cookie.ts
```diff
@@ -21,15 +26,28 @@ export function setCookie(name: string, value: string, expireDelay: number = 0,
...
 export function getCookie(name: string) {
   return findCommaSeparatedValue(document.cookie, name)
 }
 
+/**
+ * Returns all the values of the cookies with the given name
+ */
+export function getCookies(name: string): string[] | undefined {
+  return findAllCommaSeparatedValues(document.cookie).get(name)
+}
```
BenoitZugmeyer: Nit: to simplify a bit, always return an array (similar to `cookieStore.getAll()`)


```suggestion
export function getCookies(name: string): string[] {
  return findAllCommaSeparatedValues(document.cookie).get(name) || []
}
```

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -172,17 +173,35 @@ function trackResume(configuration: Configuration, cb: () => void) {
...
+  } else if (sessionStoreStrategyType.type === SessionPersistence.LOCAL_STORAGE) {
+    const rawSession = retrieveSessionFromLocalStorage()
+    // monitor-until: forever, could be handy to troubleshoot issues until session manager rework
+    addTelemetryDebug('Unexpected session state', {
+      sessionStoreStrategyType: sessionStoreStrategyType.type,
+      session: rawSession,
+      isSyntheticsTest: isSyntheticsTest(),
+      createdTimestamp: rawSession?.created,
+      expireTimestamp: rawSession?.expire,
+    })
```
BenoitZugmeyer: Suggestion: factorize a bit

```ts
let rawSession
let cookieContext
if (sessionStoreStrategyType.type === SessionPersistence.COOKIE) {
  rawSession = retrieveSessionCookie(configuration, sessionStoreStrategyType.cookieOptions)
  cookieContext = {
    cookie: await getSessionCookies(),
    currentDomain: `${window.location.protocol}//${window.location.hostname}`,
  }
} else {
  rawSession = retrieveSessionFromLocalStorage()
}


  // monitor-until: forever, could be handy to troubleshoot issues until session manager rework
  addTelemetryDebug('Unexpected session state', {
    sessionStoreStrategyType: sessionStoreStrategyType.type,
    session: rawSession,
    isSyntheticsTest: isSyntheticsTest(),
    createdTimestamp: rawSession?.created,
    expireTimestamp: rawSession?.expire,
    ...cookieContext
  })
```

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -172,17 +173,35 @@ function trackResume(configuration: Configuration, cb: () => void) {
...
+  }
```
mormubis, line 194:
```suggestion
  const rawSession = sessionStoreStrategyType.type === SessionPersistence.COOKIE ? retrieveSessionCookie(configuration, sessionStoreStrategyType.cookieOptions) : retrieveSessionFromLocalStorage()
  
  const payload = {
    sessionStoreStrategyType: sessionStoreStrategyType.type,
    session: rawSession,
    isSyntheticsTest: isSyntheticsTest(),
    createdTimestamp: rawSession?.created,
    expireTimestamp: rawSession?.expire,
  }
  
  if (sessionStoreStrategyType.type === SessionPersistence.COOKIE) {
    payload.cookie = await getSessionCookies()
    payload.origin = globalObject.location.origin
  }
  
  addTelemetryDebug('Unexpected session state', payload)
```

packages/core/test/cookie.ts
```diff
@@ -8,7 +8,9 @@ export function expireCookie() {
 }
 
 export function getSessionState(sessionStoreKey: string) {
-  return toSessionState(getCookie(sessionStoreKey))
+  const sessionState = toSessionState(getCookie(sessionStoreKey))
+  delete sessionState.c
```
mormubis, line 14: **💬 suggestion:** ‏Can we leave a comment here? I'm not sure why we delete `c`.

test/e2e/scenario/sessionStore.scenario.ts
```diff
@@ -107,6 +107,60 @@ test.describe('Session Stores', () => {
...
+        .withRum({ trackSessionAcrossSubdomains: true, enableExperimentalFeatures })
+        .withLogs({ trackSessionAcrossSubdomains: false, enableExperimentalFeatures })
+        .withHostName(FULL_HOSTNAME)
+        .run(async ({ page }) => {
+          await page.waitForTimeout(1000)
+
+          if (!encodeCookieOptions) {
+            // ensure the test is failing when the Feature Flag is disabled
+            test.fail()
+          }
```
BenoitZugmeyer, line 127: Thought: after we remove the feature flag, it could be nice to remove the `waitForTimeout` and use something like [`expect.poll`](https://playwright.dev/docs/test-assertions#expectpoll)

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
thomas-lebeau: `findLast` return `undefined` if there is no match, but here I need to return the first cookie so that when we release this feature, we can keep the existing session if any.

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
thomas-lebeau: Even if it's not necessary per se, I'd rather not leak properties unrelated to the session in the `sessionState`. I think it's cleaner if this property only live in the cookie.

Also, our session manager being complex, it's hard to know with certainty that this won't break anything. I think it's safer to remove it.

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
BenoitZugmeyer: Ah! I missed that. Alright :)

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -401,6 +402,7 @@ export function validateAndBuildConfiguration(
     beforeSend:
       initConfiguration.beforeSend && catchUserErrors(initConfiguration.beforeSend, 'beforeSend threw an error:'),
     sessionStoreStrategyType: isWorkerEnvironment ? undefined : selectSessionStoreStrategyType(initConfiguration),
+    usePartitionedCrossSiteSessionCookie: initConfiguration.usePartitionedCrossSiteSessionCookie ?? false,
```
BenoitZugmeyer: **💬 suggestion:** use `cookieOption.partitioned` or `crossSite` instead of exposing this property to the config‏

PR #3906, opened 2025-10-08
👷 fix MODULE_TYPELESS_PACKAGE_JSON warning

thomas-lebeau:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


eslint-local-rules/disallowSideEffects.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
   },
 }
 
-const packagesRoot = path.resolve(__dirname, '..', 'packages')
+const packagesRoot = path.resolve(import.meta.dirname, '..', 'packages')
```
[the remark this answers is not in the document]
thomas-lebeau: we use node 24

eslint-local-rules/disallowProtectedDirectoryImport.js
```diff
@@ -1,10 +1,15 @@
...
+import { minimatch } from 'minimatch'
 
-module.exports = {
+import resolvePackage from 'eslint-module-utils/resolve.js'
+import moduleVisitorPackage from 'eslint-module-utils/moduleVisitor.js'
+import importTypePackage from 'eslint-plugin-import/lib/core/importType.js'
+
+const moduleVisitor = moduleVisitorPackage.default
+const importType = importTypePackage.default
+const resolve = resolvePackage.default
```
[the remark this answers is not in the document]
thomas-lebeau: these modules are weird, we need the `default` key from the default export 🙄 

```
[Module: null prototype] {
  CASE_SENSITIVE_FS: false,
  __esModule: true,
  default: {
    __esModule: true,
    CASE_SENSITIVE_FS: false,
    fileExistsWithCaseSync: [Function: fileExistsWithCaseSync],
    relative: [Function: relative],
    default: [Function: resolve] { relative: [Function: relative] }
  },
  fileExistsWithCaseSync: [Function: fileExistsWithCaseSync],
  'module.exports': {
    __esModule: true,
    CASE_SENSITIVE_FS: false,
    fileExistsWithCaseSync: [Function: fileExistsWithCaseSync],
    relative: [Function: relative],
    default: [Function: resolve] { relative: [Function: relative] }
  },
  relative: [Function: relative]
}
```

mormubis, line 10: **❓ question:** ‏Not from this PR, but are we using internals here?

test/e2e/package.json
mormubis, line 1: **❓ question:** ‏I remember seeing an explanation for this in Slack. How can we keep the explanation in the codebase so we don't forget why we need this as private?

eslint-local-rules/disallowGenericUtils.js
```diff
@@ -1,6 +1,6 @@
-const path = require('path')
+import path from 'path'
```
mormubis, line 1:
```suggestion
import path from 'node:path'
```

eslint-local-rules/disallowSideEffects.js
```diff
@@ -1,6 +1,6 @@
-const path = require('path')
+import path from 'path'
```
mormubis, line 1:
```suggestion
import path from 'node:path'
```

test/envUtils.ts
```diff
@@ -1,12 +1,12 @@
-const os = require('os')
+import os from 'os'
```
mormubis, line 1:
```suggestion
import os from 'node:os'
```

mormubis (approved):

eslint-local-rules/disallowProtectedDirectoryImport.js
thomas-lebeau: it seems so

test/e2e/package.json
thomas-lebeau: `test/e2e` is not `type: module` because in some scenario we import some build cjs files ([example](https://github.com/DataDog/browser-sdk/blob/bab2fea409ae2b9b4c1df691755ae803b4bcd93a/test/e2e/scenario/logs.scenario.ts#L1C109-L1C109))

Indeed this is a bit of a hack, but this PR already improve things and we can always review the e2e test to not rely on these cjs file later.

I can't add comment to a `package.json file`, I guess blaming this file will make this PR (and this comment) visible.

eslint-local-rules/disallowSideEffects.js
thomas-lebeau: Would that work for you if I fix these in a separate PR using [import/enforce-node-protocol-usage](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/enforce-node-protocol-usage.md)? there is a lot more than these few instances.

thomas-lebeau: wip: https://github.com/DataDog/browser-sdk/pull/3943

eslint-local-rules/disallowProtectedDirectoryImport.js
BenoitZugmeyer: It's not internal, it's actual public exports: https://github.com/import-js/eslint-plugin-import/blob/01c9eb04331d2efa8d63f2d7f4bfec3bc44c94f3/utils/package.json#L9-L38

BenoitZugmeyer: Well, maybe `eslint-plugin-import/lib/core/importType.js` is internal :D

PR #3907, opened 2025-10-10
🐛 [RUM-12503] Add mask-unless-allowlisted privacy level support for standard attr

cy-moi:
## Motivation
Following up on the new `mask-unless-allowlisted` behavior, in which case we mask with allowlists when `shouldMaskNode` is true and `enablePrivacyForActionName` turned on. However, it seems like we did not cover standard attributes.

## Changes
Make masking for standard attributes the same as the textual content.

## Test instructions
Should pass CI
Note: This PR has a dependency on https://github.com/DataDog/web-ui/pull/237816 when testing on staging.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/action/getActionNameFromElement.ts
+++ b/packages/rum-core/src/domain/action/getActionNameFromElement.ts
@@ -189,7 +189,7 @@ function getActionNameFromStandardAttribute(
   let attributeValue = element.getAttribute(attribute)
   if (attributeValue && enablePrivacyForActionName) {
     const nodePrivacyLevel = getNodePrivacyLevel(element, defaultPrivacyLevel, nodePrivacyLevelCache)
-    if (shouldMaskAttribute(element.tagName, attribute, nodePrivacyLevel, rumConfiguration, attributeValue)) {
+    if (shouldMaskAttribute(element.tagName, attribute, attributeValue, nodePrivacyLevel, rumConfiguration)) {
       attributeValue = maskDisallowedTextContent(attributeValue, ACTION_NAME_PLACEHOLDER)
     }
   } else if (!attributeValue) {
--- a/packages/rum-core/src/domain/privacy.ts
+++ b/packages/rum-core/src/domain/privacy.ts
@@ -156,35 +156,38 @@ export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
 export function shouldMaskAttribute(
   tagName: string,
   attributeName: string,
+  attributeValue: string | null,
   nodePrivacyLevel: NodePrivacyLevel,
-  configuration: RumConfiguration,
-  attributeValue?: string | null
+  configuration: RumConfiguration
 ) {
-  if (nodePrivacyLevel === NodePrivacyLevel.MASK || nodePrivacyLevel === NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED) {
-    if (
-      attributeName !== PRIVACY_ATTR_NAME &&
-      !STABLE_ATTRIBUTES.includes(attributeName) &&
-      attributeName !== configuration.actionNameAttribute
-    ) {
-      switch (attributeName) {
-        case 'title':
-        case 'alt':
-        case 'placeholder':
-          return true
-      }
-      if (tagName === 'A' && attributeName === 'href') {
-        return true
-      }
-      if (tagName === 'IFRAME' && attributeName === 'srcdoc') {
-        return true
-      }
-      if (attributeValue && attributeName.startsWith('data-')) {
-        return true
-      }
-      if ((tagName === 'IMG' || tagName === 'SOURCE') && (attributeName === 'src' || attributeName === 'srcset')) {
-        return true
-      }
-    }
+  if (nodePrivacyLevel !== NodePrivacyLevel.MASK && nodePrivacyLevel !== NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED) {
+    return false
+  }
+  if (
+    attributeName === PRIVACY_ATTR_NAME ||
+    STABLE_ATTRIBUTES.includes(attributeName) ||
+    attributeName === configuration.actionNameAttribute
+  ) {
+    return false
+  }
+
+  switch (attributeName) {
+    case 'title':
+    case 'alt':
+    case 'placeholder':
+      return true
+  }
+  if (tagName === 'A' && attributeName === 'href') {
+    return true
+  }
+  if (tagName === 'IFRAME' && attributeName === 'srcdoc') {
+    return true
+  }
+  if (attributeValue && attributeName.startsWith('data-')) {
+    return true
+  }
+  if ((tagName === 'IMG' || tagName === 'SOURCE') && (attributeName === 'src' || attributeName === 'srcset')) {
+    return true
   }
 
   return false
--- a/packages/rum/src/domain/record/serialization/serializeAttribute.ts
+++ b/packages/rum/src/domain/record/serialization/serializeAttribute.ts
@@ -24,7 +24,7 @@ export function serializeAttribute(
   }
   const attributeValue = element.getAttribute(attributeName)
   const tagName = element.tagName
-  if (shouldMaskAttribute(tagName, attributeName, nodePrivacyLevel, configuration, attributeValue)) {
+  if (shouldMaskAttribute(tagName, attributeName, attributeValue, nodePrivacyLevel, configuration)) {
     // mask image URLs
     if (tagName === 'IMG') {
       // generate image with similar dimension than the original to have the same rendering behaviour


packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -169,9 +169,19 @@ function getElementById(refElement: Element, id: string) {
...
 }
 
-function getActionNameFromStandardAttribute(element: Element | HTMLElement, attribute: string): ActionName {
+function getActionNameFromStandardAttribute(
+  element: Element | HTMLElement,
+  attribute: string,
+  rumConfiguration: RumConfiguration
+): ActionName {
+  const { enablePrivacyForActionName, defaultPrivacyLevel } = rumConfiguration
+  const nodeSelfPrivacyLevel = getNodePrivacyLevel(element, defaultPrivacyLevel)
```
sethfowler-datadog: Nit: I would avoid calling this `nodeSelfPrivacyLevel`, because that phrase has a particular meaning in the code (it's the privacy level of a node _ignoring_ its ancestors), and this variable has a different meaning (it contains the privacy level of `element` _considering_ its ancestors).
```suggestion
  const nodePrivacyLevel = getNodePrivacyLevel(element, defaultPrivacyLevel)
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -169,9 +169,19 @@ function getElementById(refElement: Element, id: string) {
...
+  const attributeValue = element.getAttribute(attribute)
   return {
-    name: element.getAttribute(attribute) || '',
+    name:
+      enablePrivacyForActionName && nodeSelfPrivacyLevel && shouldMaskNode(element, nodeSelfPrivacyLevel, false)
```
sethfowler-datadog: I don't think it makes sense to check if `nodeSelfPrivacyLevel` is truthy, because `getNodePrivacyLevel` always returns a valid privacy level:
```suggestion
      enablePrivacyForActionName && shouldMaskNode(element, nodeSelfPrivacyLevel, false)
```

Looking at the bigger picture, it's unfortunate that we do all of the work of calling `getNodePrivacyLevel()` (which can be a bit expensive) and then ignore the result if `enablePrivacyForActionName` is false. It would be better to restructure this function so that you check `enablePrivacyForActionName` first, and then only call `getNodePrivacyLevel()` if `enablePrivacyForActionName` is true.

packages/rum-core/src/domain/privacy.ts
```diff
@@ -128,13 +128,16 @@ export function getNodeSelfPrivacyLevel(node: Node): NodePrivacyLevel | undefine
  * Other `shouldMaskNode` cases are edge cases that should not matter too much (ex: should we mask a
  * node if it is ignored or hidden? it doesn't matter since it won't be serialized).
  */
-export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
+export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel, isTextContent: boolean = true) {
```
sethfowler-datadog: Instead of adding this extra argument to `shouldMaskNode()`, I would suggest creating a new `shouldMaskAttribute()` function in this file and migrating as much of the logic as you can from `serializeAttribute()` in `serializeAttribute.ts` into `shouldMaskAttribute()`. This would give us a central place to control privacy for attributes, similar to what we already have for nodes, instead of spreading the logic between multiple places in the code.

I think that you should be able to restructure `serializeAttribute()` so that it more or less looks like this:
```ts
export function serializeAttribute() {
  if (/* HIDDEN */) {
    // Handle HIDDEN case.
    return ...
  }
  if (shouldMaskAttribute(...)) {
    if (/* is one of the image cases */) {
      // Handle image cases.
      return ...
    }
    // We're in a non-image case.
    return CENSORED_STRING_MARK 
  }
  // Handle non-HIDDEN, non-masked cases...
  return
}
```

With this setup, we'll ensure that action names and session replay handle attribute masking in a consistent way.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
BenoitZugmeyer: **🔨 warning:** ‏It's very inefficient to call `getNodePrivacyLevel` over and over again for each attribute / element ancestors. You could use the `cache` argument to make it faster.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -169,9 +169,19 @@ function getElementById(refElement: Element, id: string) {
...
+        ? maskDisallowedTextContent(attributeValue || '', ACTION_NAME_PLACEHOLDER)
+        : attributeValue || '',
```
BenoitZugmeyer: **💬 suggestion:** ‏maybe don't do the expensive operations if the attribute is missing

```ts
let attributeValue = element.getAttribute(attribute)
if (attributeValue) {
  const nodePrivacyLevel = ...
  etc.
} else {
  attributeValue = ''
}
return { name: attributeValue, nameSource: ActionNameSource.STANDARD_ATTRIBUTE }
```

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -169,24 +179,49 @@ function getElementById(refElement: Element, id: string) {
...
+  element: Element | HTMLElement,
+  attribute: string,
+  rumConfiguration: RumConfiguration,
+  nodePrivacyLevelCache: NodePrivacyLevelCache
+): ActionName {
+  const { enablePrivacyForActionName, defaultPrivacyLevel } = rumConfiguration
+  let attributeValue = element.getAttribute(attribute)
+  let nodeSelfPrivacyLevel
+  if (attributeValue) {
+    nodeSelfPrivacyLevel = getNodePrivacyLevel(element, defaultPrivacyLevel, nodePrivacyLevelCache)
```
BenoitZugmeyer: Suggestion:
* It should be `nodePrivacyLevel`, not `nodeSelfPrivacyLevel`
* You can use `const` since you only define it once
* You can shortcut the whole block with ` && enablePrivacyForActionName`‏
```ts
  if (attributeValue && enablePrivacyForActionName) {
    const nodePrivacyLevel = getNodePrivacyLevel(element, defaultPrivacyLevel, nodePrivacyLevelCache)
```

packages/rum-core/src/domain/privacy.ts
```diff
@@ -150,6 +153,45 @@ export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
...
+      }
+      if (tagName === 'IFRAME' && attributeName === 'srcdoc') {
+        return true
+      }
+      if (attributeValue && attributeName.startsWith('data-')) {
+        return true
+      }
+      if (tagName === 'IMG' || tagName === 'SOURCE') {
+        return true
+      }
```
BenoitZugmeyer: **🔨 warning:** ‏you changed the logic while moving the code: it makes it hard to catch the exact changes here. Don't hesitate to use multiple commits.

The logic changed and now:
* `name` and `aria-label` attributes are masked. Could you add it to the "change" section of the PR? And explain the motivation.
* all attributes of `img` and `source` elements are masked. This sounds quite problematic, because attributes like `class` or `id` will be masked and it will break the replay.

sethfowler-datadog: +1. @cy-moi , it might make sense to make the `name` and `aria-label` changes in a separate PR, since this is a policy change and not just a refactor. I do think it's a good idea! But making it in a separate PR could have several advantages:
* Easier to review.
* Easier to back out the change if we discover there's a reason to do so.
* It'd be clearly announced in the changelog.

Re: the `img` and `source` changes, I think that's probably just a bug, right? The original code in `serializeAttribute.ts` masks the `src` and `srcset` on `img` and `source` elements, but it doesn't mask all attributes on those elements.

cy-moi: I removed the `name` and `aria-label` for a later PR. And yes, the `img` and `source` changes were a bug.

packages/rum/src/domain/record/serialization/serializeAttribute.ts
```diff
@@ -55,26 +40,7 @@ export function serializeAttribute(
       return CENSORED_IMG_MARK
     }
 
-    // mask source URLs
-    if (tagName === 'SOURCE' && (attributeName === 'src' || attributeName === 'srcset')) {
-      return CENSORED_IMG_MARK
-    }
```
BenoitZugmeyer, line 61: **nitpick:** the `<SOURCE>` element is not censored with an IMG_MARK anymore, which might be fine as it's rarely used. But IDK.

cy-moi: Ah right. I will fix this and add a test case.

cy-moi: I'm confused though why we didn't apply width or height like we do for img here?

BenoitZugmeyer: I don't think we have access to the natural width and height for the source element. But maybe we can use the same codepath, just `naturalWidth` and `naturalHeight` will be undefined.

cy-moi: I see. It can have [`width` and `height`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSourceElement) though, do we want to consider them as fallback for `naturalWidth` and `naturalHeight`? But either way `getBoundingClientRect` could work as well.

packages/rum/src/domain/record/serialization/htmlAst.specHelper.ts
```diff
@@ -854,7 +854,7 @@ export const AST_MASK_UNLESS_ALLOWLISTED = {
               type: 2,
               tagName: 'a',
               attributes: {
-                href: 'https://private.com/path/nested?query=param#hash',
+                href: '***',
```
sethfowler-datadog, line 980: Definitely a better result!

packages/rum/src/domain/record/serialization/serializeAttribute.ts
sethfowler-datadog: TBH `getBoundingClientRect()` is probably a better solution overall, since there are some complexities like the CSS `image-orientation` property that mean that an image may render with different dimensions in different contexts, and there are oddities like SVGs with no intrinsic width or height where `naturalWidth` and `naturalHeight` will return `0`. (With all that said, it's probably best to just maintain the current behavior in the context of this PR.)

BenoitZugmeyer: No, it's pointless to use `width` and `height` as they reflect the `width` and `height` attributes which will be serialized anyway if present. I don't think we should go with `getBoundingClientRect` as well, as it can lead to incorrect results. I think it's fine to leave it undefined.

cy-moi: Hmm but if we share the code path, we would get values from `const { width, height } = element.getBoundingClientRect()` right?

cy-moi: I tested a bit on `getBoundingClientRect()` and we would get incorrect `width`/`height` but not necessarily 0. So in the end I didn't share the code path.

BenoitZugmeyer: Ah great.

packages/rum-core/src/domain/privacy.ts
```diff
@@ -150,6 +153,43 @@ export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
   }
 }
 
+export function shouldMaskAttribute(
+  tagName: string,
+  attributeName: string,
+  nodePrivacyLevel: NodePrivacyLevel,
+  configuration: RumConfiguration,
+  attributeValue?: string | null
```
sethfowler-datadog: It's hard to understand when it would be appropriate to not supply a value for `attributeValue`, and I think the implementation is actively unsafe if you choose to do that, because _less_ data is masked if you don't provide `attributeValue`. Please make `attributeValue` non-optional. I think all callers provide it anyway.

Also, a style nit: It's a bit awkward that `attributeName` and `attributeValue` aren't together. I'd just move `attributeValue` to be right after `attributeName`.

packages/rum-core/src/domain/privacy.ts
```diff
@@ -150,6 +153,43 @@ export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
...
+) {
+  if (nodePrivacyLevel === NodePrivacyLevel.MASK || nodePrivacyLevel === NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED) {
```
sethfowler-datadog: Style nit: please invert this condition and return early, so that you can reduce the indentation level of the rest of the function.

```suggestion
  if (nodePrivacyLevel !== NodePrivacyLevel.MASK && nodePrivacyLevel !== NodePrivacyLevel.MASK_UNLESS_ALLOWLISTED) {
    return false
  }
```

packages/rum-core/src/domain/privacy.ts
```diff
@@ -150,6 +153,43 @@ export function shouldMaskNode(node: Node, privacyLevel: NodePrivacyLevel) {
...
+    if (
+      attributeName !== PRIVACY_ATTR_NAME &&
+      !STABLE_ATTRIBUTES.includes(attributeName) &&
+      attributeName !== configuration.actionNameAttribute
+    ) {
```
sethfowler-datadog: Same style nit here; please invert this condition and return early:

```suggestion
    if (
      attributeName === PRIVACY_ATTR_NAME ||
      STABLE_ATTRIBUTES.includes(attributeName) ||
      attributeName === configuration.actionNameAttribute
    ) {
      return false
    }
```

sethfowler-datadog (commented):
Looks great, @cy-moi ; very close to being ready! Happy to approve with the comments below addressed.

PR #3914, opened 2025-10-16
🐛 [RUM-11848] Block Logs initialization on Node environments

mormubis:
## Motivation

The latest changes in Logs enable the SDK to function in Node environments. This side effect was unintended.

## Changes

The PR introduces a condition that blocks SDK initialization in Node environments.

## Test instructions

Compatibility test passed.

## Checklist

- [X] Tested locally
- [ ] Tested on staging


packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -100,6 +101,10 @@ export function createPreStartStrategy(
   }
 
   function doInit(initConfiguration: RumInitConfiguration, errorStack?: string) {
+    if (isNodeEnvironment) {
+      return
+    }
```
bcaudan, line 106: **💬 suggestion:** ‏what about doing this check earlier in the init (like for logs)? a lot can happen before `doInit` is called

PR #3921, opened 2025-10-21
✨ [RUM-12133] Track GraphQl Response

rgaignault:
## Motivation

This PR extends the existing GraphQL tracking to capture GraphQL response errors, allowing developers to monitor server-side errors in their GraphQL queries
See spec for more informations : [Graphql Spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md)

## Changes

- New trackResponseErrors option in GraphQL configuration
- Extraction of GraphQL errors from HTTP responses (fetch and XHR)
- Addition of telemetry for usage

- Simplifies the logic to save bytes since we are reading response body from logs and rum. 
- Dropped support for reading a limited number of bytes from the response body. The only drawback is reading full responses even for network errors, but these are typically short.

- Removed the “Unable to retrieve response” log message when failing to read a response body. Now we simply fallback to "Failed to load".

- E2E tests

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/browser/fetchObservable.spec.ts
+++ b/packages/core/src/browser/fetchObservable.spec.ts
@@ -30,6 +30,7 @@ describe('fetch proxy', () => {
     registerCleanupTask(() => {
       requestsTrackingSubscription.unsubscribe()
       contextEditionSubscription?.unsubscribe()
+      resetFetchObservable()
     })
   })
 
--- a/packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
+++ b/packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
@@ -1,4 +1,4 @@
-import { ErrorSource } from '@datadog/browser-core'
+import { ErrorSource, resetFetchObservable } from '@datadog/browser-core'
 import type { MockFetch, MockFetchManager } from '@datadog/browser-core/test'
 import { SPEC_ENDPOINTS, mockFetch, registerCleanupTask } from '@datadog/browser-core/test'
 import type { RawNetworkLogsEvent } from '../../rawLogsEvent.types'
@@ -32,7 +32,10 @@ describe('network error collection', () => {
   function startCollection(forwardErrorsToLogs = true) {
     mockFetchManager = mockFetch()
     const { stop } = startNetworkErrorCollection({ ...CONFIGURATION, forwardErrorsToLogs }, lifeCycle)
-    registerCleanupTask(stop)
+    registerCleanupTask(() => {
+      stop()
+      resetFetchObservable()
+    })
     fetch = window.fetch as MockFetch
   }
 
--- a/packages/rum-core/src/domain/requestCollection.spec.ts
+++ b/packages/rum-core/src/domain/requestCollection.spec.ts
@@ -41,6 +41,7 @@ describe('collect fetch', () => {
 
     registerCleanupTask(() => {
       stopFetchTracking()
+      resetFetchObservable()
     })
   })
 
@@ -354,7 +355,10 @@ describe('GraphQL response text collection', () => {
     })
     const tracerStub: Partial<Tracer> = { clearTracingIfNeeded, traceFetch: jasmine.createSpy() }
     const { stop } = trackFetch(lifeCycle, configuration, tracerStub as Tracer)
-    registerCleanupTask(stop)
+    registerCleanupTask(() => {
+      stop()
+      resetFetchObservable()
+    })
 
     return { mockFetchManager, completeSpy, fetch: window.fetch as MockFetch }
   }


test/e2e/scenario/rum/graphql.scenario.ts
```diff
@@ -78,4 +84,82 @@ test.describe('GraphQL tracking', () => {
...
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({
+            query: 'query GetUser { user { unknownField } }',
+            operationName: 'GetUser',
+          }),
+        })
+      )
+
+      await flushEvents()
+      const resourceEvent = intakeRegistry.rumResourceEvents.find((r) => r.resource.url.includes('/graphql'))!
```
bcaudan: **🥜 nitpick:** let's avoid one letter variable for readability sake.

‏
```suggestion
      const resourceEvent = intakeRegistry.rumResourceEvents.find((event) => event.resource.url.includes('/graphql'))!
```

test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -97,9 +97,34 @@ export function createMockServerApp(
     setTimeout(() => res.send('ok'), timeoutDuration)
   })
 
-  app.post('/graphql', (_req, res) => {
+  app.post('/graphql', (req, res) => {
     res.header('Content-Type', 'application/json')
-    res.json({ data: { result: 'success' } })
+
+    const scenario = req.query.scenario as string | undefined
```
bcaudan, line 103: **👏 praise:** ‏Nice!

packages/rum-core/src/domain/configuration/configuration.spec.ts
```diff
@@ -546,15 +546,29 @@ describe('validateAndBuildRumConfiguration', () => {
...
+      ])
+    })
+
+    it('should accept GraphQL options with trackResponseErrors', () => {
+      const configuration = validateAndBuildRumConfiguration({
+        ...DEFAULT_INIT_CONFIGURATION,
+        allowedGraphQlUrls: [{ match: '/graphql', trackResponseErrors: true }],
+      })!
+      expect(configuration.allowedGraphQlUrls).toEqual([
+        { match: '/graphql', trackResponseErrors: true, trackPayload: false },
```
bcaudan: **🥜 nitpick:** ‏use consistent order to ease readability


```suggestion
        { match: '/graphql', trackPayload: false, trackResponseErrors: true },
```

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -170,21 +180,37 @@ function getNextRequestIndex() {
...
-        collectStreamBody: false,
-      }
-    )
+    return
   }
+
+  const graphQlConfig = findGraphQlConfiguration(context.url, configuration)
+  const shouldExtractGraphQlErrors = graphQlConfig?.trackResponseErrors
+
+  readBytesFromStream(
```
bcaudan: **💬 suggestion:** ‏that indeed seems to be the right spot to add the graphql response error logic, however now this function responsibility is not solely to wait for the response to complete, to compute a more accurate duration. 
So it could be interesting to rename this function accordingly and see if the different responsibilities of this function could better be reflected in the implementation.

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -136,28 +140,34 @@ export function trackFetch(lifeCycle: LifeCycle, tracer: Tracer) {
...
-            input: context.input,
-            isAborted: context.isAborted,
-            handlingStack: context.handlingStack,
-            body: context.init?.body,
-          })
-        })
+        waitForResponseToComplete(
+          context,
+          configuration,
+          (duration: Duration, graphqlErrorsCount?: number, graphqlErrors?: GraphQlError[]) => {
```
bcaudan: **💬 suggestion:** ‏what about adding `duration: Duration, graphqlErrorsCount?: number, graphqlErrors?: GraphQlError[]` to `RumFetchResolveContext` and let the function mutate the context directly?
Similarly to what `tracer.traceFetch` is doing

packages/rum-core/src/domain/requestCollection.spec.ts
```diff
@@ -335,3 +335,70 @@ describe('collect xhr', () => {
...
+    const completeSpy = jasmine.createSpy('requestComplete')
+    const lifeCycle = new LifeCycle()
+    lifeCycle.subscribe(LifeCycleEventType.REQUEST_COMPLETED, completeSpy)
+
+    const configuration = mockRumConfiguration({
+      allowedGraphQlUrls: [{ match: /\/graphql$/, trackResponseErrors: true }],
+    })
+    const tracerStub: Partial<Tracer> = { clearTracingIfNeeded, traceFetch: jasmine.createSpy() }
+    const { stop } = trackFetch(lifeCycle, configuration, tracerStub as Tracer)
+    registerCleanupTask(stop)
```
bcaudan: **💬 suggestion:** ‏since compared to the next test, there is only a change in a boolean value, what about extracting a helper function for this setup?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,109 @@
...
+}
+
+export function extractGraphQlMetadata(
+  request: RequestCompleteEvent,
+  graphQlConfig: GraphQlUrlOption
+): GraphQlMetadata | undefined {
+  const metadata = extractGraphQlRequestMetadata(request.body, graphQlConfig.trackPayload)
+  if (!metadata) {
+    return
+  }
```
bcaudan: **❓ question:** ‏Does it mean that we can't track error response if we don't track the payload?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,109 @@
...
+      }
+    })
+  } else if (request.type === RequestType.FETCH) {
+    if (request.graphqlErrorsCount !== undefined) {
+      metadata.errors_count = request.graphqlErrorsCount
+    }
+    if (request.graphqlErrors !== undefined) {
+      metadata.errors = request.graphqlErrors
+    }
+  }
```
bcaudan: **💬 suggestion:** ‏At this stage, we should only rely on what has been set on the RequestCompleteEvent.
So we should handle everything xhr or fetch specific where we handle xhr and fetch.

rgaignault: Sounds good ! I moved it to requestCollection to not split the computation

packages/rum-core/src/domain/requestCollection.ts
rgaignault: Agree I renamed but I am not sure of the name, let me know what you think. Related to the other comment, graphql properties are now part of the context

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: Yes we can, payload is just undefined in this case but the metadata object is still here with operationType, operationName and variables 👍

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+
+  return metadata
+}
+
+export function parseGraphQlResponse(
+  responseText: string,
+  callback: (errorsCount?: number, errors?: GraphQlError[]) => void
+) {
+  try {
```
BenoitZugmeyer, line 48: **💬 suggestion:** ‏remove this try catch. Do proper error handling.

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+    const response = JSON.parse(responseText)
+
+    if (!response || !Array.isArray(response.errors) || response.errors.length === 0) {
```
BenoitZugmeyer: **💬 suggestion:** ‏use `isNonEmptyArray`?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+      callback()
+      return
+    }
+
+    const errors = (response.errors as unknown[]).map((error: unknown) => {
+      const errorObj = error as Record<string, unknown>
```
BenoitZugmeyer: **💬 suggestion:** ‏There is too many casts here. If you want errors to be object, just cast it directly as object


```suggestion
    const errors = (response.errors as Record<string, unknown>[]).map((error) => {
```

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+      const graphqlError: GraphQlError = {
+        message: typeof errorObj.message === 'string' ? errorObj.message : 'Unknown GraphQL error',
+      }
```
BenoitZugmeyer: **💬 suggestion:** ‏If there is no message, why not just leave this empty?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+
+      if (Array.isArray(errorObj.locations)) {
+        graphqlError.locations = (errorObj.locations as unknown[]).map((loc: unknown) => {
+          const locObj = loc as Record<string, unknown>
+          return {
+            line: typeof locObj.line === 'number' ? locObj.line : 0,
+            column: typeof locObj.column === 'number' ? locObj.column : 0,
+          }
+        })
```
BenoitZugmeyer: **💬 suggestion:** ‏can we simplify this as


```suggestion
        graphqlError.locations = errorObj.locations
```

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+) {
+  try {
+    const response = JSON.parse(responseText)
+
+    if (!response || !Array.isArray(response.errors) || response.errors.length === 0) {
+      callback()
+      return
+    }
+
+    const errors = (response.errors as unknown[]).map((error: unknown) => {
```
BenoitZugmeyer: **💬 suggestion:** ‏Do we need to go through all those validation? Can't we just have `const errors = response.errors`?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+    metadata.errors = request.graphqlErrors
+  }
+
+  return metadata
+}
+
+export function parseGraphQlResponse(
+  responseText: string,
+  callback: (errorsCount?: number, errors?: GraphQlError[]) => void
+) {
```
BenoitZugmeyer: **💬 suggestion:** ‏this function is fully synchronous, just return the result instead of using a callback

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,98 @@
...
+  if (!metadata) {
+    return
+  }
+
+  if (request.graphqlErrorsCount !== undefined) {
+    metadata.errors_count = request.graphqlErrorsCount
+  }
+  if (request.graphqlErrors !== undefined) {
+    metadata.errors = request.graphqlErrors
+  }
```
BenoitZugmeyer: **💬 suggestion:** ‏`parseGraphQlResponse` should be done here instead of forwarding `graphqlErrorsCount` and `graphqlErrors` in different layers.

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -170,21 +186,52 @@ function getNextRequestIndex() {
...
+}
+
+function waitForFetchResponseAndExtractGraphQlErrors(
+  context: RumFetchResolveContext,
+  configuration: RumConfiguration,
+  onComplete: () => void
+) {
   const clonedResponse = context.response && tryToClone(context.response)
   if (!clonedResponse || !clonedResponse.body) {
     // do not try to wait for the response if the clone failed, fetch error or null body
```
BenoitZugmeyer: **💬 suggestion:** ‏We already read the response in Logs. Can you factorize the two? We could have a function like this:

```ts
function readResponseBody(
  request: XhrCompleteContext | FetchResolveContext,
  callback: (result: { body: unknown, limitExceeded: boolean }) => void,
  options: { bytesLimit?: number }
) {
}
```

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: well, don’t we need to have  a validation for the expected type ? Furthermore the extensions code need to be set at top level

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: it’s more about a malformed error type, since we’re expecting a string in the response format. So I instead of not reporting I thought it would be clearer to inform with an unknown error message

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -170,21 +180,50 @@ function getNextRequestIndex() {
...
-        callback(elapsed(context.startClocks.timeStamp, timeStampNow()))
-      },
-      {
-        bytesLimit: Number.POSITIVE_INFINITY,
-        collectStreamBody: false,
-      }
-    )
+    context.duration = elapsed(context.startClocks.timeStamp, timeStampNow())
+    onComplete()
+    return
```
bcaudan: **💬 suggestion:** ‏The clone response check seems to also be done in the `readResponseBody`, do we still need to do it here?

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -170,21 +180,50 @@ function getNextRequestIndex() {
...
   }
+
+  const graphQlConfig = findGraphQlConfiguration(context.url, configuration)
+  const shouldCollectResponseText = graphQlConfig?.trackResponseErrors
+
+  readResponseBody(
```
bcaudan: **❓ question:** ‏it seems that now we are always decoding the response regardless of if we need it or not, would it worth avoiding to decode the response in `readResponseBody` if we don't need it?

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -170,21 +180,50 @@ function getNextRequestIndex() {
...
+  const graphQlConfig = findGraphQlConfiguration(context.url, configuration)
+  if (!graphQlConfig?.trackResponseErrors) {
+    return
+  }
+
+  readResponseBody(
+    context,
+    (result) => {
+      if (typeof result.body === 'string') {
+        context.responseText = result.body
```
bcaudan: **💬 suggestion:** ‏instead of forcing every user of `readResponseBody` to ensure that the body is a string, what about only returning a body if it is a string?

packages/rum-core/src/domain/requestCollection.ts
rgaignault: I added a collectBody parameter 👌 also for the future

packages/rum-core/src/domain/requestCollection.ts
rgaignault: seems not, good catch !

rgaignault:
@codex please review

packages/core/src/browser/fetchObservable.ts
```diff
@@ -25,16 +27,31 @@ export interface FetchResolveContext extends FetchContextBase {
...
 }
 
 export type FetchContext = FetchStartContext | FetchResolveContext
 
+type ResponseBodyActionGetter = (context: FetchResolveContext) => ResponseBodyAction
+
+export const enum ResponseBodyAction {
+  IGNORE = 0,
+  // TODO(next-major): Remove the "WAIT" action when `trackEarlyRequests` is removed, as the
+  // duratiorn of fetch requests will always come from PerformanceResourceTiming
```
bcaudan: **💬 suggestion:** ‏typo

```suggestion
  // duration of fetch requests will always come from PerformanceResourceTiming
```

packages/core/src/browser/fetchObservable.ts
```diff
@@ -25,16 +27,31 @@ export interface FetchResolveContext extends FetchContextBase {
...
   responseType?: string
   isAborted: boolean
   error?: Error
 }
 
 export type FetchContext = FetchStartContext | FetchResolveContext
 
+type ResponseBodyActionGetter = (context: FetchResolveContext) => ResponseBodyAction
+
+export const enum ResponseBodyAction {
```
bcaudan, line 44: **💬 suggestion:** ‏it could be interesting to add a comment with the intent with this concept

packages/core/src/browser/fetchObservable.ts
```diff
@@ -90,38 +108,51 @@ function beforeSend(
...
-        status: 0,
-        isAborted:
-          context.init?.signal?.aborted || (error instanceof DOMException && error.code === DOMException.ABORT_ERR),
-        error,
-      })
-    })
-  )
+      context.responseBody = bytes && new TextDecoder().decode(bytes)
+    }
+  }
```
bcaudan, line 159: **💬 suggestion:** ‏what about adding some unit tests to cover this logic?

packages/core/src/tools/readBytesFromStream.ts
```diff
@@ -1,77 +1,40 @@
...
-        })
-      }
-      bytes = completeBuffer.slice(0, options.bytesLimit)
-      limitExceeded = completeBuffer.length > options.bytesLimit
-    }
+  const bytes = options.collectStreamBody ? concatBuffers(chunks) : undefined
 
-    callback(undefined, bytes, limitExceeded)
-  }
+  return bytes
```
bcaudan: **🥜 nitpick:** ‏bytes variable could be inlined

```suggestion
  return options.collectStreamBody ? concatBuffers(chunks) : undefined
```

packages/core/src/tools/readBytesFromStream.ts
```diff
@@ -1,77 +1,40 @@
...
  * Read bytes from a ReadableStream until at least `limit` bytes have been read (or until the end of
  * the stream). The callback is invoked with the at most `limit` bytes, and indicates that the limit
  * has been exceeded if more bytes were available.
  */
-export function readBytesFromStream(
-  stream: ReadableStream<Uint8Array>,
-  callback: (error?: Error, bytes?: Uint8Array, limitExceeded?: boolean) => void,
-  options: Options
-) {
+export async function readBytesFromStream(stream: ReadableStream<Uint8ArrayBuffer>, options: Options) {
```
bcaudan, line 15: **💬 suggestion:** ‏outdated comment as limit has been removed from the options

packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
```diff
@@ -149,116 +144,60 @@ describe('network error collection', () => {
...
+    it('should truncate responseBody according to limit', (done) => {
+      const longResponse = 'a'.repeat(100)
 
-  it('should return undefined if body is disturbed', (done) => {
-    const response = new MockResponse({ bodyDisturbed: true })
-    computeFetchResponseText(response, CONFIGURATION, (responseText) => {
-      expect(responseText).toBeUndefined()
-      done()
-    })
-  })
```
bcaudan, line 222: **❓ question:** ‏Should not we still have tests for those cases?

packages/rum-core/src/domain/requestCollection.spec.ts
```diff
@@ -335,3 +335,46 @@ describe('collect xhr', () => {
...
+
+    return { mockFetchManager, completeSpy, fetch: window.fetch as MockFetch }
+  }
+
+  it('should collect responseBody when trackResponseErrors is enabled', (done) => {
+    const { mockFetchManager, completeSpy, fetch } = setupGraphQlFetchTest(true)
```
bcaudan, line 367: **💬 suggestion:** ‏it seems that we don't test the `false` case anymore

packages/logs/src/domain/networkError/networkErrorCollection.spec.ts
rgaignault: Good catch 👌 I added back

packages/rum-core/src/domain/requestCollection.spec.ts
rgaignault: Added a test with a reset on fetchObservable otherwise it could be flaky with logs using the response body.

packages/core/src/browser/fetchObservable.spec.ts
```diff
@@ -265,3 +265,87 @@ describe('fetch proxy', () => {
...
+
+    requests = []
+    requestsTrackingSubscription = initFetchObservable({
+      responseBodyAction: () => ResponseBodyAction.COLLECT,
+    }).subscribe((context) => {
+      if (context.state === 'resolve') {
+        requests.push(context)
+      }
+    })
+    fetch = window.fetch as MockFetch
```
bcaudan: **🥜 nitpick:** ‏this setup is pretty similar to what we have in above tests, we could maybe factor some things

packages/rum-core/src/domain/requestCollection.spec.ts
```diff
@@ -335,3 +335,73 @@ describe('collect xhr', () => {
     })
   })
 })
+
+describe('GraphQL response text collection', () => {
+  const FAKE_GRAPHQL_URL = 'http://fake-url/graphql'
+
+  beforeEach(() => {
+    resetFetchObservable()
```
bcaudan, line 344: **🥜 nitpick:** ‏it could be interesting to identify the test(s) that don't do a proper reset to avoid the same issue in future tests

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -1,23 +1,94 @@
...
+      code: (error.extensions as Record<string, unknown> | undefined)?.code as string,
+    }
+
+    return graphqlError
+  })
+
+  return {
+    errorsCount: errors.length,
+    errors,
+  }
```
BenoitZugmeyer: **🥜 nitpick:** ‏just return `errors` here


```suggestion
  return errors
```

Do the `errors.length` in `extractGraphQlMetadata`. This saves a bit of bytes and removes one data model.

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -32,7 +32,9 @@ export interface CustomContext {
   traceSampled?: boolean
 }
 export interface RumFetchStartContext extends FetchStartContext, CustomContext {}
-export interface RumFetchResolveContext extends FetchResolveContext, CustomContext {}
+export interface RumFetchResolveContext extends FetchResolveContext, CustomContext {
+  duration?: Duration
```
BenoitZugmeyer: **💬 suggestion:** ‏I think this `duration` property is not needed anymore

PR #3926, opened 2025-10-23
⚗️[MFE] Source code context event enrichment

bcaudan:
## Motivation

In the context of micro frontend support, we add the ability to enrich events with source code context (service and version) from window.DD_SOURCE_CODE_CONTEXT.
DD_SOURCE_CODE_CONTEXT is injected at build time by the [build plugin](https://github.com/DataDog/build-plugins/pull/255)￼.

## Changes

- Add `sourceCodeContext.ts`.
- Add `rawRumEvent` and `domainContext` to the assemble hook parameters to allow extracting the stack for sourceCodeContext.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/contexts/pageStateHistory.spec.ts
+++ b/packages/rum-core/src/domain/contexts/pageStateHistory.spec.ts
@@ -207,7 +207,7 @@ describe('pageStateHistory', () => {
           eventType,
           startTime: clock.relative(0),
           duration: 0 as Duration,
-        } as any)
+        } as AssembleHookParams)
 
         expect(defaultRumEventAttributes).toEqual({
           type: eventType,
--- a/packages/rum-core/src/domain/hooks.ts
+++ b/packages/rum-core/src/domain/hooks.ts
@@ -21,6 +21,8 @@ type DeepReadonly<T> = {
   readonly [K in keyof T]: DeepReadonly<T[K]>
 }
 
+// Use readonly and DeepReadonly to prevents assemble hook callbacks from mutating the inputs.
+// DeepReadonly is only applied to objects rather than the entire AssembleHookParams to avoid casts for primitives.
 export interface AssembleHookParams {
   readonly eventType: RumEvent['type']
   rawRumEvent: DeepReadonly<RawRumEvent>
--- a/test/e2e/scenario/microfrontend.scenario.ts
+++ b/test/e2e/scenario/microfrontend.scenario.ts
@@ -31,7 +31,7 @@ const LOGS_CONFIG: Partial<LogsInitConfiguration> = {
   },
 }
 
-// because the evaluation the script in a different context than the page, resulting in unexpected stack traces
+// Use createBody because `page.evaluate()` runs in a different context and breaks stack traces.
 function createBody(eventGenerator: string) {
   return html`
     <button>click me</button>
--- a/test/e2e/scenario/rum/errors.scenario.ts
+++ b/test/e2e/scenario/rum/errors.scenario.ts
@@ -3,7 +3,7 @@ import { test, expect } from '@playwright/test'
 import { createTest, html } from '../../lib/framework'
 
 // Note: using `browser.execute` to throw exceptions may result in "Script error." being reported,
-// because WDIO is evaluating the script in a different context than the page.
+// Use createBody because `page.evaluate()` runs in a different context and breaks stack traces.
 function createBody(errorGenerator: string) {
   return html`
     <button>click me</button>


packages/rum-core/src/domain/contexts/pageStateHistory.spec.ts
```diff
@@ -207,7 +207,7 @@ describe('pageStateHistory', () => {
           eventType,
           startTime: clock.relative(0),
           duration: 0 as Duration,
-        })
+        } as any)
```
bcaudan: **💬 suggestion:** ‏
```suggestion
        } as AssembleHookParams)
```

packages/rum-core/src/domain/contexts/sourceCodeContext.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+        eventType: 'error',
+        startTime: 0 as RelativeTime,
+        domainContext: {},
+        rawRumEvent: {
+          type: 'error',
+          error: {
+            stack: MATCHING_TEST_STACK,
+          },
+        } as any,
+      })
```
bcaudan: **💬 suggestion:** ‏it seems that we can remove the any casts from this file with something like

```suggestion
      const result = hooks.triggerHook(HookNames.Assemble, {
        eventType: 'error',
        startTime: 0 as RelativeTime,
        domainContext: {},
        rawRumEvent: {
          type: 'error',
          error: {
            stack: MATCHING_TEST_STACK,
          },
        },
      } as AssembleHookParams)
```

packages/rum-core/src/domain/hooks.ts
```diff
@@ -9,18 +9,28 @@ import type {
...
 export type DefaultTelemetryEventAttributes = RecursivePartial<TelemetryEvent>
 
+type DeepReadonly<T> = {
+  readonly [K in keyof T]: DeepReadonly<T[K]>
+}
+
+export interface AssembleHookParams {
+  readonly eventType: RumEvent['type']
+  rawRumEvent: DeepReadonly<RawRumEvent>
+  domainContext: DeepReadonly<RumEventDomainContext<RawRumEvent['type']>>
```
bcaudan, line 29: **❓ question:** ‏What is the intent with the DeepReadonly?

packages/rum-core/src/domain/contexts/sourceCodeContext.spec.ts
```diff
@@ -0,0 +1,183 @@
+import { ExperimentalFeature, HookNames } from '@datadog/browser-core'
+import type { RelativeTime } from '@datadog/browser-core'
+import type { Hooks } from '../hooks'
+import { createHooks } from '../hooks'
+import { mockExperimentalFeatures, registerCleanupTask } from '../../../../core/test'
+import type { BrowserWindow } from './sourceCodeContext'
+import { startSourceCodeContext } from './sourceCodeContext'
+
+describe('sourceCodeContext', () => {
```
bcaudan, line 10: **💬 suggestion:** ‏It could be interesting to add e2e tests for this feature

packages/rum-core/src/domain/hooks.ts
amortemousque: I don’t want registered assemble hook callbacks to mutate the input, especially `rawRumEvent`. I avoided `DeepReadonly<AssembleHookParams>` because it forces casts for primitives (eventType, startTime, duration) throughout the code.

Example: 
image

packages/rum-core/src/domain/contexts/sourceCodeContext.spec.ts
amortemousque: Sure, added :) https://github.com/DataDog/browser-sdk/pull/3926/files#diff-88fb28376084e0b79b43bf21317d313440ea697f26ceeb4b8c093452be55d4e5

packages/rum-core/src/domain/hooks.ts
bcaudan: oh ok, what about adding a comment with the code to make the intent more explicit?

test/e2e/scenario/microfrontend.scenario.ts
```diff
@@ -28,6 +31,31 @@ const LOGS_CONFIG: Partial<LogsInitConfiguration> = {
   },
 }
 
+// because the evaluation the script in a different context than the page, resulting in unexpected stack traces
```
bcaudan: **💬 suggestion:** ‏unclear to me what information we want to convey here and how it relates to create body 😕

bcaudan (commented):
LGTM

test/e2e/scenario/microfrontend.scenario.ts
amortemousque: Yeah, it’s a copy-paste from [here](https://github.com/DataDog/browser-sdk/blob/8c27e4b8495ddd1a528bc1a5125c50c9c8af46b8/test/e2e/scenario/rum/errors.scenario.ts#L6)￼. It means we use `createBody` instead of  `page.evaluate()` because  `page.evaluate()` messes up the stack trace and drops the URL. I’ll improve the comments.

packages/rum-core/src/domain/contexts/pageStateHistory.spec.ts
BeltranBulbarellaDD: I might be wrong but this is still `as any`

PR #3938, opened 2025-10-31
👷 reduce build boilerplate

BenoitZugmeyer:
## Motivation

I'm about to introduce a bunch of new packages. I want to reduce the amount of boilerplate we need to maintain within each package.

## Changes

Introduce a new "build-package" script to unify how we build all packages. It supports two options, `--modules` and `--bundle` to either build modules or bundles (or both), following package needs.

The `npm-run-all` dependency has been removed. Packages are not built in parallel anymore, but that doesn't seem to impact the overall build time (at least on my machine):

Before:
```
yarn build  21.685s
yarn build:bundle  10.414s
```

After:
```
yarn build  21.123s
yarn build:bundle 10.409s
```

## Test instructions

Run `yarn build`, check if everything is correctly built.
Check if `yarn dev` is still working.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/build/build-package.ts
+++ b/scripts/build/build-package.ts
@@ -1,7 +1,8 @@
 import fs from 'node:fs/promises'
 import { parseArgs } from 'node:util'
+import path from 'node:path'
+import { globSync } from 'node:fs'
 import ts from 'typescript'
-import { globSync } from 'glob'
 import webpack from 'webpack'
 import webpackBase from '../../webpack.base.ts'
 
@@ -103,10 +104,11 @@ async function buildModules({ outDir, module, verbose }: { outDir: string; modul
 }
 
 async function replaceBuildEnvInDirectory(dir: string, { verbose }: { verbose: boolean }) {
-  for (const path of globSync('**/*.js', { cwd: dir, absolute: true })) {
-    if (await modifyFile(path, (content: string) => replaceBuildEnv(content))) {
+  for (const relativePath of globSync('**/*.js', { cwd: dir })) {
+    const absolutePath = path.resolve(dir, relativePath)
+    if (await modifyFile(absolutePath, (content: string) => replaceBuildEnv(content))) {
       if (verbose) {
-        printLog(`Replaced BuildEnv in ${path}`)
+        printLog(`Replaced BuildEnv in ${absolutePath}`)
       }
     }
   }


scripts/check-packages.ts
```diff
@@ -48,7 +48,7 @@ function getPackageFiles(packagePath: string): string[] {
   //
   // [1]: Quoting Lerna doc: "Lerna always uses npm to publish packages."
   // https://lerna.js.org/docs/features/version-and-publish#from-package
-  const output = command`npm pack --dry-run --json`.withCurrentWorkingDirectory(packagePath).run()
+  const output = command`npm pack --ignore-scripts --dry-run --json`.withCurrentWorkingDirectory(packagePath).run()
```
BenoitZugmeyer, line 51: Note: this is because `npm pack` runs the [`prepack` npm script in the react package](https://github.com/DataDog/browser-sdk/blob/a2998402f6ca1e953cca2fe228ac9683bb8e3040/packages/rum-react/package.json#L11). Previously this script didn't output anything, but now it prints "Building modules...", breaking the JSON output.

Ignoring scripts is fine here, since the `check-packages` command [is run after the packages are freshly built](https://github.com/DataDog/browser-sdk/blob/a2998402f6ca1e953cca2fe228ac9683bb8e3040/.gitlab-ci.yml#L168-L170), so no need to build the react plugin again.

scripts/build/build-package.ts
```diff
@@ -0,0 +1,139 @@
...
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      getCurrentDirectory: ts.sys.getCurrentDirectory,
+      getNewLine: () => ts.sys.newLine,
+    }
+    console.error(ts.formatDiagnosticsWithColorAndContext(diagnostics, formatHost))
+  }
+
+  if (emitResult.emitSkipped) {
+    throw new Error('Failed to build package')
+  }
```
BenoitZugmeyer: I don't *love* this TypeScript code, but at the same time I didn't find a better way to invoke TypeScript with a dynamic configuration (maybe with a temporary file that we put somewhere and feed to `tsc`... but I'm not sure it's better).

As I'm mentioning in the comment above, I hope we can move on from TypeScript for building packages at some point.

scripts/dev-server.ts
```diff
@@ -25,8 +22,19 @@ function createStaticSandboxApp(): express.Application {
...
-    app.use(middleware(webpack(config(null, { mode: 'development' }))!))
+  for (const packageName of PACKAGES_WITH_BUNDLE) {
+    const packagePath = `./packages/${packageName}`
+    app.use(
+      middleware(
+        webpack(
+          webpackBase({
+            mode: 'development',
+            entry: `${packagePath}/src/entries/main.ts`,
+            filename: packageName === 'worker' ? 'worker.js' : `datadog-${packageName}.js`,
```
BenoitZugmeyer, line 33: * There is a little bit of duplication between the `build-package.ts` and `dev-server.ts`. I believe this is acceptable as long as it's only a couple of lines.

* I don't love this `filename` exception for `worker.js`. Maybe we can unify filenames here, and always generate `${packageName}.js` (ex: `rum.js`, `logs.js`) or always `datadog-${packageName}.js` (ex: `datadog-worker.js`). Those paths are currently used in e2e tests and the devtools extension, so that's not as straightforward as it might seem. I think this is fine for now.

scripts/build/build-package.ts
amortemousque: To me it's fine but maybe you could clarify what it does with comments or split it into smaller functions with clear names?

scripts/dev-server.ts
thomas-lebeau: yeah, having a more consistent naming would help having more generic scripts without having to hardcode names. let's keep that in mind, maybe for a mob session?

scripts/build/build-package.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  for (const path of globSync('**/*.js', { cwd: outDir, absolute: true })) {
```
thomas-lebeau: **🥜 nitpick:** globSync is supported natively in node. see #3950‏


```suggestion
  for (const path of fs.globSync('**/*.js', { cwd: outDir, absolute: true })) {
```

BenoitZugmeyer: Done https://github.com/DataDog/browser-sdk/pull/3938/commits/872295479fca1b0bc31b0cde75807817c3a40ac6

scripts/build/build-package.ts
BenoitZugmeyer: Done https://github.com/DataDog/browser-sdk/pull/3938/commits/3b108a1bff4ea20ea6c6cfffbf31d50c612a93b2

thomas-lebeau (approved):

PR #3957, opened 2025-11-12
👷  Reduce packages boilerplate a bit more

BenoitZugmeyer:
## Motivation

My goal is to reduce packages boilerplate a bit more by removing the need to declare a `pack` script in each package.json.

## Changes

This PR is using `yarn workspaces foreach exec` to run `yarn pack` in each package instead of using a package script.

I took the opportunity of revisiting how we execute stuff in workspaces to use `yarn workspaces` instead of `lerna` in more places. This was not really needed for my goals, but:

* The output is smaller, more readable than lerna
* It has less overhead than lerna
* It doesn't take `peerDependencies` into account when executing commands in topological order, so it doesn't complain about recursive dependencies and parallelize more things.
* We progressively reduce our dependency on lerna, and we might want to remove it totally some day.

## Test instructions

If CI is green, it should be alright.

Run `yarn build`, `yarn build:bundle` and check if things are built correctly
Run `yarn run pack` and verify that a `package.tgz` archive has been generated in each package folder (`ls packages/*/package.tgz`)

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/check-typescript-compatibility.ts
+++ b/scripts/check-typescript-compatibility.ts
@@ -13,8 +13,8 @@ interface TypeScriptCheck {
 
 runMain(() => {
   printLog('Building project...')
-  command`yarn build`.run()
-  command`lerna run pack --stream`.run()
+  command`yarn run build`.run()
+  command`yarn run pack`.run()
 
   printLog('Setting up test environment...')
   command`yarn install --no-immutable`.withCurrentWorkingDirectory(TEST_APP_DIR).run()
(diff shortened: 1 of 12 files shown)


scripts/check-typescript-compatibility.ts
```diff
@@ -13,8 +13,8 @@ interface TypeScriptCheck {
 
 runMain(() => {
   printLog('Building project...')
-  command`yarn build`.run()
-  command`lerna run pack --stream`.run()
+  command`yarn run build`.run()
+  command`yarn run pack`.run()
```
BenoitZugmeyer, line 17: **note:** Using `run` is required for running the `pack` script because `yarn pack` is doing something else. Using `run` in `yarn run build` is not required, but I added it just for uniformity.

thomas-lebeau (approved):

PR #3960, opened 2025-11-13
♻️  remove constants from the `Configuration` object

BenoitZugmeyer:
## Motivation

While working on the new architecture, I am introducing a new way to build and sanitize the configuration. To simplify things a bit, let's remove constants from the configuration object, and use actual constants instead.

Also this should reduce the bundle size a bit.

## Changes

Replace those configuration properties with collocated constants:

* eventRateLimiterThreshold
* maxTelemetryEventsPerPage
* batchBytesLimit
* flushTimeout
* batchMessagesLimit
* messageBytesLimit

## Test instructions

Those constants were mostly used around batching. See if events are correctly flushed.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/core/src/transport/batch.ts
+++ b/packages/core/src/transport/batch.ts
@@ -60,7 +60,7 @@ export function createBatch({
 
     if (estimatedMessageBytesCount >= MESSAGE_BYTES_LIMIT) {
       display.warn(
-        `Discarded a message whose size was bigger than the maximum allowed size ${MESSAGE_BYTES_LIMIT}KB. ${MORE_DETAILS} ${DOCS_TROUBLESHOOTING}/#technical-limitations`
+        `Discarded a message whose size was bigger than the maximum allowed size ${MESSAGE_BYTES_LIMIT / ONE_KIBI_BYTE}KiB. ${MORE_DETAILS} ${DOCS_TROUBLESHOOTING}/#technical-limitations`
       )
       return
     }


packages/core/src/transport/batch.ts
```diff
@@ -58,9 +58,9 @@ export function createBatch({
 
     const estimatedMessageBytesCount = encoder.estimateEncodedBytesCount(serializedMessage)
 
-    if (estimatedMessageBytesCount >= messageBytesLimit) {
+    if (estimatedMessageBytesCount >= MESSAGE_BYTES_LIMIT) {
       display.warn(
-        `Discarded a message whose size was bigger than the maximum allowed size ${messageBytesLimit}KB. ${MORE_DETAILS} ${DOCS_TROUBLESHOOTING}/#technical-limitations`
+        `Discarded a message whose size was bigger than the maximum allowed size ${MESSAGE_BYTES_LIMIT}KB. ${MORE_DETAILS} ${DOCS_TROUBLESHOOTING}/#technical-limitations`
```
rgaignault: 💬 suggestion: The KB in the message seems wrong here, as it is 256 * 1024. 
```suggestion
        `Discarded a message whose size was bigger than the maximum allowed size ${MESSAGE_BYTES_LIMIT / 1000)KB. ${MORE_DETAILS} ${DOCS_TROUBLESHOOTING}/#technical-limitations`
```

sethfowler-datadog: You could also divide by 1024 and change it to KiB; that'd give you a round number result.

sethfowler-datadog (approved):
Looks great!

PR #3964, opened 2025-11-18
♻️ refactor new datacenter workflow

thomas-lebeau:
## Motivation

As part of PR0, we're introduction a new DC `pr00test.staging.dog`. 
This PR simplify a bit the runbook and setup the new DC.

It is recommended to review commit by commit.

## Changes

- Collocate `monitorIdsByDatacenter` with other site config
- Remove the hardcoded list of _minor datacenter_. This way new datacenters are automatically deployed as with the other minor DCs.
- Add a new manual deploy job that can be configured in the gitlab UI when ran. This allow to deploy the latest release of the SDK to a new DC without having to create a new release.
- Add the new DC config

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


scripts/deploy/deploy-prod-dc.ts
```diff
@@ -10,9 +12,25 @@ const ONE_MINUTE_IN_SECOND = 60
...
+  positionals,
+} = parseArgs({
+  allowPositionals: true,
+  options: {
+    withMonitorChecks: {
+      type: 'boolean',
+      default: true,
+    },
+  },
+})
```
BenoitZugmeyer, line 26: **issue:** The AI is kind of right [here](https://github.com/DataDog/browser-sdk/pull/3964/files#r2538247244), what you are trying to do won't work.

The simplest solution is to remove the `default: true` here, and call `deploy-prod-dc` script with `--withMonitorChecks` only when you want to check monitors (in the auto pipeline):

```
node ./scripts/deploy/deploy-prod-dc.ts v${VERSION%%.*} $UPLOAD_PATH --withMonitorCheck
```

If you want to keep the `default: true`, you could enable `allowNegative` in parseArgs and invoke the script with `--no-withMonitorCheck` in the manual pipeline:


```
node ./scripts/deploy/deploy-prod-dc.ts v${VERSION%%.*} $UPLOAD_PATH --no-withMonitorCheck
```

Also, nitpicking a bit, but it could be nicer to format the option name as `--monitor-check` or something.

scripts/lib/datacenter.ts
```diff
@@ -1,12 +1,24 @@
...
+  pr00test: 'pr00test.staging.dog',
+}
+
+/**
+ * Each datacenter has 3 monitor IDs:
+ * - Telemetry errors
+ * - Telemetry errors on specific org
+ * - Telemetry errors on specific message
+ */
+export const monitorIdsByDatacenter: Record<string, [number, number, number]> = {
```
BenoitZugmeyer, line 17: **nitpick:** nitpicking and for the record: I think it would be nice to have a `datacenterConfiguration` object like:

```ts
export const datacenterConfigurations = {
  us1: { site: 'xxx', monitorIds: [...] },
  ...
}
```

scripts/lib/datacenter.ts
BenoitZugmeyer, line 1: **nitpick:** this file contains more than sites now

scripts/lib/datacenter.ts
thomas-lebeau: yeah, I think we actually need to discover these monitors in the middle/long term

PR #3973, opened 2025-11-19
👷 do not rely on monitors for deployment gate

thomas-lebeau:
## Motivation

We want to avoid hardcoding the list of datacenter in the release process. To prepare for this, the first step is to get rid of `monitorIdsByDatacenter`. For this we're querying the logs API directly. Newly created DC might not have api or application key setup so we're skipping the monitors for thoses.
Another benefit from this new strategy for monitoring DC deployments is that we're able to only account for the errors related to the version of the SDK currently being deployed. This makes the release more resilient to unrelated monitor alerts.

## Changes

- remove `monitorIdsByDatacenter` in favor of a logs API query

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/deploy/deploy-prod-dc.ts
+++ b/scripts/deploy/deploy-prod-dc.ts
@@ -60,12 +60,12 @@ export async function main(...args: string[]): Promise<void> {
   command`node ./scripts/deploy/upload-source-maps.ts ${version} ${uploadPath}`.withLogs().run()
 
   if (shouldCheckTelemetryErrors && uploadPath !== 'root') {
-    await gateMonitors(uploadPath)
+    await gateTelemetryErrors(uploadPath)
   }
 }
 
-async function gateMonitors(uploadPath: string): Promise<void> {
-  printLog(`Check monitors for ${uploadPath} during ${GATE_DURATION / ONE_MINUTE_IN_SECOND} minutes`)
+async function gateTelemetryErrors(uploadPath: string): Promise<void> {
+  printLog(`Check telemetry errors for ${uploadPath} during ${GATE_DURATION / ONE_MINUTE_IN_SECOND} minutes`)
   for (let i = 0; i < GATE_DURATION; i += GATE_INTERVAL) {
     await checkTelemetryErrors(uploadPath.split(','), browserSdkVersion)
     process.stdout.write('.') // progress indicator
--- a/scripts/deploy/lib/checkTelemetryErrors.ts
+++ b/scripts/deploy/lib/checkTelemetryErrors.ts
@@ -1,5 +1,5 @@
 /**
- * Check monitors status
+ * Check telemetry errors
  */
 import { printLog, fetchHandlingError } from '../../lib/executionUtils.ts'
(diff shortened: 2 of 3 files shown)


mormubis (approved):

scripts/deploy/check-monitors.ts
```diff
@@ -5,50 +5,129 @@
...
+        to: 'now',
+        query: query.query,
+      },
+    }),
   })
-  return response.json() as Promise<MonitorStatus>
+
+  const data = (await response.json()) as QueryResult
+
+  return data.data.buckets
```
BenoitZugmeyer, line 122: **suggestion:** add a bit of runtime check to ensure the returned JSON is the expected one, and avoid failing silently if it doesn't match the typescript type.

```ts
if (!data || !data.data || !Array.isArray(data.data.buckets) || !data.data.buckets.every(...)) {
  throw new Error("Unexpected response from the API: ${JSON.stringify(data)}")
}
```

scripts/deploy/deploy-prod-dc.ts
```diff
@@ -21,27 +21,33 @@ function getAllMinorDcs(): string[] {
...
-  throw new Error('UPLOAD_PATH argument is required')
-}
+  if (!uploadPath) {
+    throw new Error('UPLOAD_PATH argument is required')
+  }
 
-runMain(async () => {
   if (checkMonitors) {
     command`node ./scripts/deploy/check-monitors.ts ${uploadPath}`.withLogs().run()
   }
```
BenoitZugmeyer, line 57: **thought:** checking monitors before deploy is pointless since `check-monitors` only care about the new version being deployed. Maybe we could remove this, or make the query broader

scripts/deploy/lib/checkTelemetryErrors.ts
BenoitZugmeyer, line 1: **suggestion:** check-monitors does not check monitors anymore. Maybe rename the script, or since it's only used in `deploy-prod-dc`, you could just change it to a lib function, not a script.

scripts/deploy/check-monitors.ts
```diff
@@ -5,50 +5,129 @@
...
+const QUERIES: Query[] = [
+  {
+    name: 'Telemetry errors',
+    query: BASE_QUERY,
+    threshold: 300,
+  },
+  {
+    name: 'Telemetry errors on specific org',
+    query: BASE_QUERY,
+    facet: '@org_id',
```
bcaudan: **💬 suggestion:** ‏using something like `groupBy` in the config instead of `facet` could clarify the purpose.

scripts/deploy/lib/checkTelemetryErrors.ts
```diff
@@ -0,0 +1,171 @@
+/**
+ * Check monitors status
```
bcaudan: **💬 suggestion:** ‏There are various "check monitors" / "check telemetry monitors" / "gateMonitors" references left, a search and replace could leave things a cleaner state

PR #3974, opened 2025-11-19
👷 fix temporal dead zone

thomas-lebeau:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


BenoitZugmeyer:
Actually, you wouldn't have run into this issue if you moved the execution in the `runMain` callback :)

PR #3978, opened 2025-11-21
🎨 [PANA-4976] Separate emission of replay records and serialization stats

sethfowler-datadog:
## Motivation

Currently, when we emit a replay record, we can optionally emit a `SerializationStats` object as well; this information is aggregated and will eventually be reported to the server if telemetry is enabled. The function signature for the operation looks like this:
```ts
type EmitRecordCallback = (record: BrowserRecord, stats?: SerializationStats) => void;
```

It turns out that this `EmitRecordCallback` interface generalizes a bit awkwardly to code which needs to aggregate multiple records and `SerializationStats` objects together. If we maintain the mapping between the records we're aggregating and the `SerializationStats` objects they were originally associated with, we end up doing bookkeeping work that just isn't necessary. If we decide that, since the mapping is not necessary, we won't track it, then we need to choose an arbitrary record to emit our `SerializationStats` with, which creates a misleading impression that the record and stats are related and could lead to bugs as the code changes over time.

We can support this kind of aggregation better if we explicitly separate the act of emitting a record from the act of emitting `SerializationStats`. By using two separate callbacks, we make the API better reflect the actual dependencies between the data, and eliminate the tradeoff that currently forces us to pick between unnecessary bookkeeping and long-term confusion.

## Changes

This PR replaces the interface described above with two callbacks, with the following shape:
```ts
export type EmitRecordCallback = (record: BrowserRecord) => void
export type EmitStatsCallback = (stats: SerializationStats) => void
```

These callbacks are used everywhere; they replace the individual callback types defined on each of the replay tracker types. In tests, having a more specific type than `BrowserRecord` for the argument of the `EmitRecordCallback` is often useful; to enable this, the real code uses the following generic signature:
```ts
export type EmitRecordCallback<Record extends BrowserRecord = BrowserRecord> = (record: Record) => void
```

The trackers accept an `EmitRecordCallback<Record>` callback, where `Record` is whatever record type they emit, and we take advantage of this where relevant in tests. Other code generally just uses `EmitRecordCallback` with no explicit generic parameter, since we generally don't care about the specific record type.

The existing unit tests have been updated to exercise the new callbacks.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum/src/domain/record/record.ts
```diff
@@ -40,14 +41,14 @@ export interface RecordAPI {
...
 
 export function record(options: RecordOptions): RecordAPI {
-  const { emit, configuration, lifeCycle } = options
+  const { emitRecord, emitStats, configuration, lifeCycle } = options
   // runtime checks for user options
-  if (!emit) {
-    throw new Error('emit function is required')
+  if (!emitRecord || !emitStats) {
+    throw new Error('emit functions are required')
   }
```
gonzalezreal, line 48: 🤏🏽 non-blocking
```suggestion
  if (!emitRecord) throw new Error('emitRecord function is required')
  if (!emitStats) throw new Error('emitStats function is required')
```

gonzalezreal (approved):
:rocket: LGTM!
Just a minor, non-blocking suggestion.

packages/rum/src/domain/record/record.ts
sethfowler-datadog: Thanks; that's a good suggestion! In `web-ui` I'd definitely have preferred that approach. In the browser SDK it's a bit more nuanced, because adding a second message here will increase the size of the bundle, which is something we're always trying to minimize. (In fact, there's normally a bot in this repo that tells you how your PR affected the bundle size; I'm not sure why it didn't post in this PR.) For that reason, my weak preference here is to stick with a single message, even though it would provide more clarity.

PR #3985, opened 2025-11-26
♻️ separate RUM event collection

BenoitZugmeyer:
## Motivation

In the new architecture, things like the session manager, telemetry and batch will be made by the "core" package. The "rum" package will be focused on producing RUM events.

To be able to reuse most of "start rum", this PR clearly separates RUM events collection from the rest, so the new architecture can simply import it.

## Changes

Please review commit by commit. The main change was to move more stuff into the existing `startRumEventsCollection` function, and export it.

## Test instructions

This PR touches the main part of the RUM SDK. It should be well covered by e2e tests. To test manually, just run the dev sandbox and verify that RUM data is correctly generated

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum-core/src/boot/startRum.spec.ts
+++ b/packages/rum-core/src/boot/startRum.spec.ts
@@ -1,4 +1,4 @@
-import type { RawError, Duration, RelativeTime, BufferedData } from '@datadog/browser-core'
+import type { RawError, Duration, BufferedData } from '@datadog/browser-core'
 import {
   Observable,
   stopSessionManager,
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/boot/startRum.spec.ts
```diff
@@ -63,46 +44,26 @@ function startRumStub(
...
     noopRecorderApi,
-    viewHistory
+    undefined,
+    createCustomVitalsState(),
+    new Observable(),
+    undefined,
+    reportError
   )
 
-  startLongAnimationFrameCollection(lifeCycle, configuration)
```
BenoitZugmeyer, line 99: **note:** one nice thing with this change is that in the spec, we don't need to manually launch as many collection as before.

packages/rum-core/src/boot/startRum.spec.ts
```diff
@@ -219,103 +154,6 @@ describe('rum session keep alive', () => {
...
-    changeLocation('/bar')
-
-    serverRumEvents.length = 0
-
-    triggerOnLoad()
-    clock.tick(THROTTLE_VIEW_UPDATE_PERIOD)
-
-    expect(serverRumEvents.length).toEqual(1)
-    expect(serverRumEvents[0].view.url).toEqual('http://foo.com/')
-  })
```
BenoitZugmeyer, line 316: **note:** those tests were moved to `trackViews.spec` because they were mostly related to views (the assertions related to the long task event should be covered by `urlContext` specs), and this way `startRumEventsCollection` don't need to support `location` and `locationChangeObservable` parameters (it already has a lot of parameters).

PR #3990, opened 2025-12-01
👷 bump monitor

mormubis:
## Motivation

The telemetry remains relevant, and the ticket is still open.

## Changes

Bump the comment date

## Test instructions

N/A

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/sessionManager.ts
+++ b/packages/core/src/domain/session/sessionManager.ts
@@ -231,7 +231,7 @@ function detectSessionIdChange(configuration: Configuration, initialSessionState
         const time = dateNow() - sdkInitTime
         getSessionCookies()
           .then((cookie) => {
-            // monitor-until: 2026-01-01, after RUM-10845 investigation done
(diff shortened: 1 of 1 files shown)


packages/core/src/domain/session/sessionManager.ts
```diff
@@ -231,7 +231,7 @@ function detectSessionIdChange(configuration: Configuration, initialSessionState
         const time = dateNow() - sdkInitTime
         getSessionCookies()
           .then((cookie) => {
-            // monitor-until: 2025-12-01, after RUM-10845 investigation done
+            // monitor-until: 2026-01-01, after RUM-10845 investigation done
```
bcaudan: **💬 suggestion:** ‏Unless we know we'll get rid of it this month, we could give ourself a bit more time.
```suggestion
            // monitor-until: 2026-04-01, after RUM-10845 investigation done
```

PR #3992, opened 2025-12-03
Add support for persisted GraphQL Queries

rgaignault:
## Motivation

Our GraphQL server makes use of persisted queries for operations which means the query and operationType are not known when making calls to GQL which prevents us from being able to use the built in GQL tracking for enriching our resource events

example:
```

{
    "graphqlBody": {
        "operationName": "Navigation",
        "variables": {},
        "extensions": {
            "persistedQuery": {
                "version": 1,
                "sha256Hash": "0e669423cef683226cb8eb295664619c8e0f95945734e0a458095f51ee89efb3"
            }
        }
    }
}
```

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/resource/graphql.spec.ts
+++ b/packages/rum-core/src/domain/resource/graphql.spec.ts
@@ -158,13 +158,8 @@ describe('GraphQL detection and metadata extraction', () => {
       })
     })
 
-    it('should extract query operation name and variables for persisted queries without query', () => {
+    it('should extract operation name and variables when query is absent', () => {
       const requestBody = JSON.stringify({
-        extensions: {
-          persistedQuery: {
-            sha256Hash: 'somehashvalue',
-          },
-        },
         operationName: 'GetUser',
         variables: { id: '123' },
       })
@@ -176,28 +171,6 @@ describe('GraphQL detection and metadata extraction', () => {
         operationName: 'GetUser',
         variables: '{"id":"123"}',
         payload: undefined,
-      })
-    })
-
-    it('should extract full metadata when both query and persistedQuery are present', () => {
-      const requestBody = JSON.stringify({
-        query: 'query GetUser { user { id name } }',
-        extensions: {
-          persistedQuery: {
-            sha256Hash: 'somehashvalue',
-          },
-        },
-        operationName: 'GetUser',
-        variables: { id: '123' },
-      })
-
-      const result = extractGraphQlRequestMetadata(requestBody, true)
-
-      expect(result).toEqual({
-        operationType: 'query',
-        operationName: 'GetUser',
-        variables: '{"id":"123"}',
-        payload: 'query GetUser { user { id name } }',
       })
     })
   })


packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -97,6 +102,13 @@ export function extractGraphQlRequestMetadata(
     return
   }
 
+  if (graphqlBody && graphqlBody.extensions?.persistedQuery && !graphqlBody.query) {
+    return {
+      operationName: graphqlBody.operationName,
+      variables: graphqlBody.variables ? JSON.stringify(graphqlBody.variables) : undefined,
+    }
+  }
+
```
BenoitZugmeyer: **suggestion:** you don't need this. Just treat `query`/`operationType` as optional, and follow the same execution flow.

```ts
 if (!graphqlBody) {
    return
  }

  let operationType: string | undefined
  if (graphqlBody.query) {
    operationType = getOperationType(graphqlBody.query.trim())
  }

  const operationName = graphqlBody.operationName

  let variables: string | undefined
  if (graphqlBody.variables) {
    variables = JSON.stringify(graphqlBody.variables)
  }

  return {
    operationType,
    operationName,
    variables,
    payload: trackPayload ? safeTruncate(query, GRAPHQL_PAYLOAD_LIMIT, '...') : undefined,
  }
```

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -88,7 +88,12 @@ export function extractGraphQlRequestMetadata(
     return
   }
 
-  let graphqlBody: { query?: string; operationName?: string; variables?: unknown }
+  let graphqlBody: {
+    query?: string
+    operationName?: string
+    variables?: unknown
+    extensions?: { persistedQuery?: unknown }
```
BenoitZugmeyer: **suggestion:** you don't need to know about `extensions.persistedQuery`. Just treating query as optional should be enough.

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: question : Is it a cleaner way to handle things ? It seems we can now report an empty object (even if it isn’t displayed), and if there’s no persisted query, I’m not sure we should report a GraphQL request that only contains an operationName, it feels like a malformed request.

BenoitZugmeyer: It's simpler and DRY. The empty object is not too problematic by itself, but could be avoided by:

```
if (operationType || operationName || variables || payload) {
  return {
    operationType,
    operationName,
    variables,
    payload,
  }
}
```

> report a GraphQL request that only contains an operationName, it feels like a malformed request

We report what we can. All fields are optional after all. Reporting partial data is often better than not reporting anything.

BenoitZugmeyer (approved):

packages/rum-core/src/domain/resource/graphql.spec.ts
```diff
@@ -140,7 +150,55 @@ describe('GraphQL detection and metadata extraction', () => {
...
+      })
+
+      const result = extractGraphQlRequestMetadata(requestBody, true)
+
+      expect(result).toEqual({
+        operationType: 'query',
+        operationName: 'GetUser',
+        variables: '{"id":"123"}',
+        payload: 'query GetUser { user { id name } }',
+      })
```
bcaudan: **❓ question:** ‏removing `extensions: { persistedQuery: ... }` from test cases does not change the outcome, so do we even need to use / mention it in the tests?

rgaignault: It was related to the use case, to do a more realistic scenario but that's not necessarily useful, I deleted it

PR #3994, opened 2025-12-03
Add more complete tests for DOM attribute serialization

sethfowler-datadog:
## Motivation

I'd like to make some small changes to DOM attribute serialization, both to improve consistency and to resolve some bugs. I also want to separate out the code that generates "virtual attributes" like `_cssText` and `rr_scrollTop` from the code that captures real DOM attributes, so that I can handle the two cases differently in the new serialization algorithm. Before doing any of that, though, it's important that we clearly understand what the existing code is doing, so that we can make changes to it safely. That would be easier if we had a more thorough set of tests for this code.

## Changes

This PR adds two new sets of tests for `serializeAttributes()`, one for DOM attributes and one for virtual attributes. These test suites are quite thorough, covering a wide variety of situations and explicitly testing all privacy levels.

To make it easier validate that upcoming PRs don't change the behavior of this code, the tests for virtual attributes use a helper function which returns only virtual attributes. I plan to split `serializeAttributes()` into two functions in a future PR; we should be able to remove the helper function and keep the exact same tests. The hope is that this will give us greater confidence that the refactoring has been performed correctly.

I've added numerous TODOs to the test suites to indicates opportunities to improve consistency and fix bugs; some surprising edge cases are exposed by these tests. If there are no objections to the changes, I'll fix these issues as well in forthcoming PRs.

## Test instructions

There are no behavior changes in this PR, so you just need to run the tests!

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/serialization/serializeAttributes.spec.ts
+++ b/packages/rum/src/domain/record/serialization/serializeAttributes.spec.ts
@@ -228,49 +228,6 @@ describe('serializeAttributes for DOM attributes', () => {
       { html: '<input type="url" value>', expectedBehavior: 'maskable-form', ignoreBehavior: 'masked' },
       { html: '<input type="week" value="2017-W01">', expectedBehavior: 'maskable', ignoreBehavior: 'masked' },
       { html: '<input type="week" value>', expectedBehavior: 'maskable-form', ignoreBehavior: 'masked' },
-      /*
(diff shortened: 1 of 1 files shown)


sethfowler-datadog:
Note that this is using the APIs from #3982, so it can't be merged until that one gets merged.

PR #3998, opened 2025-12-04
🎨 [PANA-5053] Separate DOM and virtual attribute serialization

sethfowler-datadog:
## Motivation

The current DOM serialization algorithm produces two kinds of attributes on elements:
1. Real DOM attributes, like `foo="bar"` in `<div foo="bar">`.
2. "Virtual" attributes representing metadata about the element. For example, `rr_scrollLeft` and `rr_scrollTop` are used to capture information about the initial scroll position of an element.

Many of these virtual attributes are primarily of interest in full snapshots; later, when the values change incrementally, we represent them using dedicated records. For example, later changes to `rr_scrollLeft` and `rr_scrollTop` are represented as `BrowserIncrementalSnapshotRecord` records with `IncrementalSource.Scroll`. (Speculating on the reasoning for the different representation in full snapshots, I'd guess this was done because there's no easy way to bundle multiple changes into a single, atomic full snapshot record except to include them in the DOM snapshot itself. Incremental snapshots can bundle together many changes at once, so they don't have the same issue.)

The new DOM serialization algorithm doesn't have a special record type for full snapshots; everything uses a representation comparable to our current incremental snapshots. For that reason, the new algorithm doesn't need these virtual attributes; it makes more sense to encode these values in the same way we do today for incremental snapshots.

To make it easier to share code between the current DOM serialization algorithm and the new one, let's separate the code that serializes real DOM attributes from the code that serializes virtual attributes.

## Changes

This PR:
* Renames `safeAttrs` to `attrs` in `serializeAttributes()` as a minor opportunistic refactor. I'm not a big fan of the name `safeAttrs` since it implies that everything in `safeAttrs` satisfies our privacy rules and is safe to record, but in fact that's not true until the end of the function.
* Splits `serializeAttributes()` into two functions, `serializeDOMAttributes()` and `serializeVirtualAttributes()`. (`serializeAttributes()` survives as a trivial wrapper around these two functions.)
* Tweaks the types in a few places to more accurately reflect the kinds of values that `serializeDOMAttributes()` can now return.
* Removes the call to `getValidTagName()` from both functions since it contains a regular expression which it would be wasteful to invoke in two places. `getValidTagName()` doesn't really provide any benefit since we are only comparing the tag name with exact known strings in this code; we can just use `Element#tagName` directly.
* Updates the tests for `serializeAttributes()` to call `serializeDOMAttributes()` or `serializeVirtualAttributes()`, as appropriate. (I wrote the tests with this in mind, so making the switch is easy.)

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/serialization/serializationUtils.ts
+++ b/packages/rum/src/domain/record/serialization/serializationUtils.ts
@@ -90,6 +90,14 @@ export function getValidTagName(tagName: string): string {
   return processedTagName
 }
 
+/**
+ * Returns the tag name of the given element, normalized to ensure a consistent lowercase
+ * representation regardless of whether the element is HTML, XHTML, or SVG.
+ */
+export function normalizedTagName(element: Element): string {
+  return element.tagName.toLowerCase()
+}
+
 export function censoredImageForSize(width: number, height: number) {
   return `data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='${width}' height='${height}' style='background-color:silver'%3E%3C/svg%3E`
 }
--- a/packages/rum/src/domain/record/serialization/serializeAttributes.spec.ts
+++ b/packages/rum/src/domain/record/serialization/serializeAttributes.spec.ts
@@ -335,6 +335,19 @@ describe('serializeDOMAttributes', () => {
       }
     }
   })
+
+  it('normalizes tag names', () => {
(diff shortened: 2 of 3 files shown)


yannickadam (commented):
What worries me is the removal of `getValidTagName`. I know it's probably a stretch nowadays, but if one of our customer still serves their application using `application/xhtml+xml`, wouldn't the tag names be lower case?

PR #4002, opened 2025-12-05
♻️ [PANA-5123] Assign node ids in preorder when serializing

sethfowler-datadog:
## Motivation

The current DOM serialization algorithm serializes a node's children before assigning it an id. This results in a post-order assignment, like this:
```html
#document  // id: 5
  <html>  // id: 4
    <body>  // id: 3
      <div></div>  // id:1
      <p></p>  // id: 2
```

This is a fine solution for the current data format; node ids are recorded explicitly, so we can assign them using whatever scheme we wish. However, the new format assigns node ids implicitly in the order that the nodes appear, which means that parent nodes will naturally be assigned a node id before their descendants. The new format uses a pre-order assignment, which looks like this:
```html
#document  // id: 0
  <html>  // id: 1
    <body>  // id: 2
      <div></div>  // id:3
      <p></p>  // id: 4
```

This difference in node id numbering is the biggest source of complexity when comparing the output of the current DOM serialization algorithm with the output of the new one. It'd be ideal if both algorithms used the same ids; this would make translation between their representations straightforward.

The good news is that, because the current data format records node ids explicitly, there's no backwards compatibility issue with simply changing how node ids are assigned. So, we can switch the existing algorithm to use a pre-order assignment, and eliminate this impedance mismatch between the two algorithms.

## Changes

This PR:
1. Moves the code for assigning an id to a node from `serializeNodeWithId()` to `SerializationTransaction#assignId()`.
2. Updates each of the DOM node serialization functions to assign ids when each serialized node is constructed, using `SerializationTransaction#assignId()` to do the work.
3. Removes `serializeNodeWithId()`, which now does nothing, and replaces calls to it with direct calls to `serializeNode()`.
4. Changes the first node id we'll assign from `1` to `0`, for better alignment between the old and new algorithms.
5. Updates the small number of tests that needed to be adjusted.

## Test instructions

You can test the changes using the "Live Replay" tab of the browser SDK extension.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum/src/domain/record/trackers/trackMutation.spec.ts
```diff
@@ -473,16 +473,14 @@ describe('trackMutation', () => {
         expect(emitRecordCallback).toHaveBeenCalledTimes(1)
         const { validate, expectNewNode, expectInitialNode } = createMutationPayloadValidator(serializedDocument)
 
+        const expectedHost = expectNewNode({ type: NodeType.Element, tagName: 'div' })
+        const shadowRootNode = expectNewNode({ type: NodeType.DocumentFragment, isShadowRoot: true })
         const child = expectNewNode({ type: NodeType.Element, tagName: 'span' })
```
sethfowler-datadog, line 478: The helper function `expectNewNode()` assigns ids to each node in order, so we have to reverse the order of these calls to get ids that match the new approach.

packages/rum/src/domain/record/serialization/serializeNode.spec.ts
```diff
@@ -987,6 +987,7 @@ describe('serializeDocumentNode handles', function testAllowDomTree() {
 
       expect(serializeDocumentNode(document, NodePrivacyLevel.ALLOW, transaction)).toEqual({
         type: NodeType.Document,
+        id: 0,
```
sethfowler-datadog, line 990: It was necessary to add this because `serializeDocumentNode()` used to return a node with no id assigned, but now it returns a node with an id. Conveniently, the id of the document node is always zero with pre-order assignment!

packages/rum/src/entries/internal.ts
```diff
@@ -15,4 +15,4 @@ export {
 
 export * from '../types'
 
-export { serializeNodeWithId } from '../domain/record'
+export { serializeNode, serializeNode as serializeNodeWithId } from '../domain/record'
```
sethfowler-datadog, line 18: There are some known external callers of this function in Datadog-internal code. I've exported the function under both the old and new names for compatibility, but I plan to update those callers once this change ships, and then we can remove the old name.

packages/rum/src/domain/record/nodeIds.ts
```diff
@@ -4,7 +4,7 @@ export type NodeWithSerializedNode = Node & { __brand: 'NodeWithSerializedNode'
 export type NodeId = number & { __brand: 'NodeId' }
 
 export const enum NodeIdConstants {
-  FIRST_ID = 1,
+  FIRST_ID = 0,
```
gonzalezreal, line 7: Why it needs to start at 0, serialization requirement?
Is this change backwards compatible?
Are we positive we are not using `0` elsewhere as a sentinel value?

Sorry if these don't make sense, this change seems risky for someone without all the context (like me 😄)

gonzalezreal (approved):

packages/rum/src/domain/record/nodeIds.ts
sethfowler-datadog: > Why it needs to start at 0, serialization requirement?

It's because in the new data format, the ids can also be interpreted as array indices; it'd be awkward if they didn't start at 0.

> Is this change backwards compatible? Are we positive we are not using 0 elsewhere as a sentinel value?

For the data format, yes. For code which might confuse `0` with other values like `undefined` (since `0` is falsy), there is definitely a potential risk. I've updated the code in both this repo and in Datadog-internal repos to handle this case correctly. (Almost everything already did, fortunately, but there were a few small issues that have been hammered out.)

PR #4004, opened 2025-12-05
Innovation Week POC: Live Debugger browser support

watson:


ahfuckit:
> _ No description provided. _
edited from
> It should probably live somewhere else, but for now I've just hardcoded it in the sandbox app as a starting point

Why? Somewhere less obvious? Don't want to get caught?

PR #4011, opened 2025-12-08
[RUM Browser Profiler] stop profiler when session expires

thomasbertet:
## Motivation

RUM Browser SDK should stop the Profiler once the Session is expired. 

This PR also adds a "reason" for being stopped to the Profiler instance, in order to handle cases where user manually stopped the Profiler and it should not be re-activated after a session renew.

## Changes

- Make the Profiler listen to the SESSION_EXPIRED event, and stop the profiler once triggered.
- Add a `stateReason` to stopped instances. 
- Check that `stateReason` to avoid restarting the Profiler once it has been stopped by user.
- Keep `.start()` and `.stop()` API for the "user-controlled" state (may be an official API later)
- Have an internal `stopProfiler` function that handles all the required clean-up and data collection on stop, called whatever the reason for stop. 
- Differentiate between `pause` and `stop` via `pauseProfilerInstance` and `stopProfilerInstance` to allow providing a reason for one and not the other, keeping all functions focused on their use-cases.

## Test instructions

- Added a few unit tests with start/stop of the profiler.
- In an app, you can run `DD_RUM.setTrackingConsent('not-granted')` and assert the Profiler is stopped, no more profiles are sent after that point.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/boot/profilerApi.ts
+++ b/packages/rum/src/boot/profilerApi.ts
@@ -61,8 +61,16 @@ export function makeProfilerApi(): ProfilerApi {
           return
         }
 
-        profiler = createRumProfiler(configuration, lifeCycle, sessionManager, profilingContextManager, createEncoder)
-        profiler.start(viewHistory.findView())
+        profiler = createRumProfiler(
+          configuration,
+          lifeCycle,
+          sessionManager,
+          profilingContextManager,
+          createEncoder,
+          viewHistory,
+          undefined
+        )
+        profiler.start()
       })
       .catch(monitorError)
   }
--- a/packages/rum/src/domain/profiling/profiler.ts
+++ b/packages/rum/src/domain/profiling/profiler.ts
@@ -19,7 +19,7 @@ import type {
   RumConfiguration,
   RumSessionManager,
   TransportPayload,
-  ViewHistoryEntry,
+  ViewHistory,
 } from '@datadog/browser-rum-core'
 import {
   createFormDataTransport,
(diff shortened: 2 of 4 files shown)


packages/rum/src/boot/profilerApi.ts
```diff
@@ -61,8 +61,16 @@ export function makeProfilerApi(): ProfilerApi {
...
+        profiler = createRumProfiler(
+          configuration,
+          lifeCycle,
+          sessionManager,
+          profilingContextManager,
+          createEncoder,
+          viewHistory,
+          undefined
+        )
+        profiler.start()
```
thomasbertet, line 73: note : no more `viewHistory.findView()` in the "public" API. It's now done inside the profiler, as we pass in the `viewHistory` as dependency of `createRumProfiler`.

packages/rum/src/domain/profiling/profiler.ts
```diff
@@ -65,13 +67,28 @@ export function createRumProfiler(
   // Global clean-up tasks for listeners that are not specific to a profiler instance (eg. visibility change, before unload)
   const globalCleanupTasks: Array<() => void> = []
 
-  let instance: RumProfilerInstance = { state: 'stopped' }
+  let instance: RumProfilerInstance = { state: 'stopped', stateReason: 'initializing' }
```
amortemousque, line 70: **💬 suggestion:** Instead of using stateReason, can we simply add a new state?

thomasbertet: There is indeed multiple ways we could do this. 
Are you thinking of this suggestion from a stylistic POV ?

thomasbertet: I mostly used a `reason` to allow having a clear "stop" state from outside (ie. the `isStopped` function) and avoids diverging from the `profilingContextManager` where we have a few other status as well that mostly relates 1-1 to the internal "profiler" state.

amortemousque (approved):

PR #4013, opened 2025-12-09
♻️ Replace longTaskRegistry by longTaskContexts

amortemousque:
## Motivation

Remove longTaskRegistry and use a new longTaskContext exposed by longTaskCollection. This keeps the codebase consistent and preserves the bundle size.

## Changes

- Move `longAnimationFrameCollection` into `longTaskCollection`, since long tasks act as the fallback for LOAF.
- Expose `longTaskContexts` from `longTaskCollection`.
- Replace `longTaskRegistry` with `longTaskContexts` in profiling.
- Delete `longTaskRegistry` and `profilingCorrelation` modules.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/longTask/longTaskCollection.spec.ts
+++ b/packages/rum-core/src/domain/longTask/longTaskCollection.spec.ts
@@ -1,4 +1,3 @@
-import { relativeToClocks } from '@datadog/browser-core'
 import type { RelativeTime, ServerDuration } from '@datadog/browser-core'
 import { registerCleanupTask } from '@datadog/browser-core/test'
 import {
--- a/packages/rum/src/domain/profiling/profiler.spec.ts
+++ b/packages/rum/src/domain/profiling/profiler.spec.ts
@@ -1,4 +1,4 @@
-import type { ViewHistoryEntry} from '@datadog/browser-rum-core';
+import type { ViewHistoryEntry } from '@datadog/browser-rum-core'
 import { LifeCycle, LifeCycleEventType, RumPerformanceEntryType, createHooks } from '@datadog/browser-rum-core'
 import type { Duration, RelativeTime, TimeStamp } from '@datadog/browser-core'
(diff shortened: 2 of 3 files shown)


packages/rum/src/domain/profiling/profilingContext.ts
```diff
@@ -7,13 +7,11 @@ export interface ProfilingContextManager {
   get: () => ProfilingInternalContextSchema | undefined
 }
 
-export const startProfilingContext = (hooks: Hooks): ProfilingContextManager => {
-  // Default status is `starting`.
+export function startProfilingContext(hooks: Hooks): ProfilingContextManager {
```
amortemousque, line 10: Unrelated but use classic function declaration for consistency

packages/rum-core/src/domain/longTask/longTaskCollection.ts
```diff
@@ -1,47 +1,133 @@
...
+export interface LongTaskContext {
+  id: string
+  startClocks: ClocksState
+  duration: Duration
+  entryType: RumPerformanceEntryType.LONG_ANIMATION_FRAME | RumPerformanceEntryType.LONG_TASK
+}
+
+export interface LongTaskContexts {
+  findLongTasks: (startTime: RelativeTime) => LongTaskContext[]
+}
```
thomasbertet, line 37: Is this `s` plural to represent the API object a convention ? Otherwise I can suggest : 

```tsx
export interface LongTaskEntry {
  id: string
  startClocks: ClocksState
  duration: Duration
  entryType: RumPerformanceEntryType.LONG_ANIMATION_FRAME | RumPerformanceEntryType.LONG_TASK
}
```
 and 
```tsx
export interface LongTaskContext {
  findLongTasks: (startTime: RelativeTime) => LongTaskEntry[]
}
```

I'd find this a bit more clear - but it may not be following existing convention so overall I trust what you think is best !

thomasbertet (approved):
LGTM overall ! 

Added a nit on naming, but otherwise I believe that's a great refactoring!

PR #4025, opened 2025-12-10
🎨 Add flagEvaluationEndpointBuilder to TransportConfiguration interface.

lisaah:
## Motivation

Adds the flagEvaluationEndpointBuilder that was previously added to the configuration interface for accurate typing.

## Changes

This is a type only change.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/configuration/transportConfiguration.ts
+++ b/packages/core/src/domain/configuration/transportConfiguration.ts
@@ -10,6 +10,7 @@ export interface TransportConfiguration {
   sessionReplayEndpointBuilder: EndpointBuilder
   profilingEndpointBuilder: EndpointBuilder
   exposuresEndpointBuilder: EndpointBuilder
+  flagEvaluationEndpointBuilder: EndpointBuilder
   datacenter?: string | undefined
   replica?: ReplicaConfiguration
   site: Site


packages/core/src/domain/configuration/transportConfiguration.ts
```diff
@@ -10,6 +10,7 @@ export interface TransportConfiguration {
   sessionReplayEndpointBuilder: EndpointBuilder
   profilingEndpointBuilder: EndpointBuilder
   exposuresEndpointBuilder: EndpointBuilder
+  flagEvaluationEndpointBuilder: EndpointBuilder
```
bcaudan, line 13: **💭 thought:** ‏I wonder if it is something that should live in core as exposures and flagEvaluation are not even used by the browser-sdk.
cc @BenoitZugmeyer

BenoitZugmeyer: Yes, absolutely! I [even gave it a try](https://github.com/DataDog/browser-sdk/pull/3988), and it's almost working! Not sure exactly why I stopped there, I was focusing on other tasks. In the meantime I accepted @lisaah PR to avoid blocking her, as I think this is benign and can be improved later.

cc @thomas-lebeau as you were also talking about this.

PR #4028, opened 2025-12-11
👷 Other strategy to handle test app lockfiles with renovate

bcaudan:
## Motivation

Previous strategy #4019, #4021, #4023 did not work

## Changes

- Rollback previous strategy
- Switch packages dependencies as optional peer dependencies
- Install peer packages when building test apps

## Test instructions

- CI should pass
- lock files maintenance should pass

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/build/build-test-apps.ts
+++ b/scripts/build/build-test-apps.ts
@@ -29,6 +29,8 @@ function buildApp(appPath: string) {
   command`yarn install --no-immutable`.withCurrentWorkingDirectory(appPath).run()
 
   // install peer dependencies if any
+  // intent: renovate does not allow to generate local packages before install
+  // so local packages are marked as optional peer dependencies and only installed when we build the test apps
   const packageJson = JSON.parse(fs.readFileSync(path.join(appPath, 'package.json'), 'utf-8'))
   if (packageJson.peerDependencies) {
     // For each peer dependency, install it


test/apps/base-extension/package.json
```diff
@@ -4,12 +4,22 @@
...
+    "@datadog/browser-logs": "*",
+    "@datadog/browser-rum": "*"
+  },
+  "peerDependenciesMeta": {
+    "@datadog/browser-logs": {
+      "optional": true
+    },
+    "@datadog/browser-rum": {
+      "optional": true
+    }
```
thomas-lebeau, line 17: **🥜 nitpick:** If we're gonna install them anyway, then they are no optional (and it seems that npm would actually install them automatically, unfortunately this is not the case for yarn)

You might need to add the peer dependencies before doing `yarn install` in the build script though. I don't know if it's worth it.

scripts/build/build-test-apps.ts
```diff
@@ -27,6 +27,21 @@ runMain(async () => {
 function buildApp(appPath: string) {
   printLog(`Building app at ${appPath}...`)
   command`yarn install --no-immutable`.withCurrentWorkingDirectory(appPath).run()
+
+  // install peer dependencies if any
```
thomas-lebeau, line 31: **💬 suggestion:** ‏Maybe add a comment explaining why we are using peed dependencies instead of dependencies. I think I got it but it's not obvious. (ideally that comment would be in the package.json, but 🤷)

.gitlab-ci.yml
```diff
@@ -153,6 +153,7 @@ typecheck:
   interruptible: true
   script:
     - yarn
+    - yarn build
```
thomas-lebeau, line 156: **💭 thought:** ‏I guess this is because otherwise the type definitions from the peer dependencies are not found? That's a bit annoying to have to build just to typecheck 😞 (but I don't see an alternative)

test/apps/base-extension/package.json
bcaudan: The issue is that renovate don't allow to build those dependencies before running the install to re-generate the lock file.

So the idea is to have a way to run the install without having those dependencies built, otherwise the install fails as the dependencies are missing.
But we still need to build them before effectively be able to use the test apps.

Do you have a simpler alternative in mind?

.gitlab-ci.yml
bcaudan: FYI, here I'm just reverting to the behavior that we add before https://github.com/DataDog/browser-sdk/pull/4019, since we don't nee to run yarn build at install anymore

test/apps/base-extension/package.json
thomas-lebeau: all good then! thanks

thomas-lebeau (approved):

PR #4035, opened 2025-12-16
✨[PANA-5288] Update rum-events-format to pull in BrowserChangeRecord

sethfowler-datadog:
## Motivation

As part of the new session replay data format arc of work, we need to pull in the new record type definitions that were recently merged into `rum-events-format`.

## Changes

This PR updates the `rum-events-format` submodule. I've run `yarn json-schemas:sync` to automatically synchronize the type definitions in this repo. (It looks like some other changes from `rum-events-format` had not been pulled in yet, so this makes some additional changes besides adding `BrowserChangeRecord`.) I've also manually defined a `ChangeType` enumeration for the different `Change` variants, just as we've already done for other similar types like `NodeType` and `IncrementalSource`.

I had to tweak `scripts/check-schemas.ts` to run `yarn format -w` after generating code from the JSON schemas, because Prettier disagrees with the code generator about how multiline array literals should be formatted.

There are no functional changes in this PR.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/types/sessionReplay.ts
+++ b/packages/rum/src/types/sessionReplay.ts
@@ -356,7 +356,10 @@ export type Change =
   | [4, ...TextChange[]]
   | [5, ...SizeChange[]]
   | [6, ...ScrollPositionChange[]]
-  | [7, ...AdoptedStyleSheetsChange[]]
+  | [7, ...AddStyleSheetChange[]]
+  | [8, ...AttachedStyleSheetsChange[]]
+  | [9, ...MediaPlaybackStateChange[]]
+  | [10, ...VisualViewportChange[]]
 /**
  * Browser-specific. Schema representing the addition of a string to the string table.
  */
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/rumEvent.types.ts
```diff
@@ -482,7 +479,7 @@ export type RumErrorEvent = CommonProperties &
     /**
      * View properties
      */
-    readonly view: {
+    readonly view?: {
```
mormubis, line 482: Is this right? Is now the view optional?

sethfowler-datadog: It's intended -- it comes from Téo's PR [here](https://github.com/DataDog/rum-events-format/pull/322). I don't have a full understanding of the context, though. It's not part of my changes; it just came along because Téo's changes hadn't been pulled into this repo yet.

PR #4037, opened 2025-12-17
👷 do not rely on hardcoded list of DCs

thomas-lebeau:
## Motivation

We want to avoid hardcoding the list of datacenter in the release process. This will remove the need for creating a PR for new DCs.

For this we discover the list of DC and their corresponding site using ~`ddtool`~ Runtime Metadata Service.

PR stacked on top of https://github.com/DataDog/browser-sdk/pull/3973

## Changes

- Refactor a bit to normalize usage of `datacenter` vs `uploadPath` and try to make things more clear
- Added unit tests for deploy-prod-dc script.
- removed `siteByDatacenter` in favor of a ~ddtool command~ a call to Runtime Metadata Service

## Todos
- [x] ~install ddtool on the CI (or find another way)~ used Runtime Metadata Service
- [x] fix upload-source-maps test failing. for some reason the mock seems to initialize too late 

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/deploy/lib/checkTelemetryErrors.spec.ts
+++ b/scripts/deploy/lib/checkTelemetryErrors.spec.ts
@@ -65,6 +65,7 @@ describe('check-telemetry-errors', () => {
 
     await mockModule(path.resolve(import.meta.dirname, '../../lib/executionUtils.ts'), {
       fetchHandlingError: fetchHandlingErrorMock,
+      timeout: mock.fn(() => Promise.resolve()),
     })
 
     await mockModule(path.resolve(import.meta.dirname, '../../lib/datacenter.ts'), {


scripts/lib/datacenter.ts
```diff
@@ -1,9 +1,86 @@
...
+interface DatacentersResponse {
+  datacenters: Datacenter[]
+}
+
+let cachedDatacenters: Record<string, Datacenter> | undefined
+
+async function getAllDatacentersMetadata(): Promise<Record<string, Datacenter>> {
+  if (cachedDatacenters) {
+    return cachedDatacenters
+  }
```
BenoitZugmeyer: **nitpick:** calling `getAllDatacentersMetadata` multiple times synchronously won't leverage the cache.

```suggestion
let cachedDatacenters: Promise<Record<string, Datacenter>> | undefined

async function getAllDatacentersMetadata(): Promise<Record<string, Datacenter>> {
  if (cachedDatacenters) {
    return cachedDatacenters
  }
  cachedDatacenters = fetchDatacentersFromRuntimeMetadataService().then(...)
```

scripts/deploy/lib/testHelpers.ts
```diff
@@ -81,6 +83,77 @@ export function mockCommandImplementation(mockFn: Mock<(...args: any[]) => void>
...
+      return Promise.resolve({
+        json: () => Promise.resolve({ datacenters: MOCK_DATACENTERS }),
+      } as unknown as Response)
+    }
+
+    // Default response
+    return Promise.resolve({
+      ok: true,
+      json: () => Promise.resolve({}),
+    } as unknown as Response)
```
BenoitZugmeyer: **nitpick:** it could be nice to have a `createMockResponse` function with defaults, like this:

```ts
function createMockResopnse({ status = 200, json }: { status?: number, json?: any }) {
    return {
        ok: status < 300,
        status,
        json: () => Promise.resolve(json),
        text: () => Promise.resolve(JSON.stringify(json))
    }
}
```

scripts/lib/datacenter.ts
```diff
@@ -1,9 +1,86 @@
...
+  return (await getAllDatacentersMetadata())[datacenter].site
+}
+
+export async function getAllDatacenters(): Promise<string[]> {
+  return Object.keys(await getAllDatacentersMetadata())
+}
+
+export async function getAllMinorDcs(): Promise<string[]> {
+  return (await getAllDatacenters()).filter((dc) => !MAJOR_DCS.includes(dc) && !dc.startsWith('pr'))
+}
```
BenoitZugmeyer: **suggestion:** maybe introduce something like:

```
enum DatacenterType {
  // ... what is a minor dc ...
  Minor,
  // ... what is a major dc ..
  Major,
  // ... what is a private region ...
  PrivateRegion
}

return getDatacenterType(name: string) {
  // ...
}
```

scripts/lib/datacenter.ts
```diff
@@ -1,9 +1,86 @@
...
+
+export async function getAllPrivateDcs(): Promise<string[]> {
+  return (await getAllDatacenters()).filter((dc) => dc.startsWith('pr'))
+}
+
+interface Datacenter {
+  name: string
+  site: string
+}
```
BenoitZugmeyer: **suggestion:** why not exposing this type more, so we don't need call `getSite` as much and things would be a bit less `async`

`name` could be the `shortName`, as we never use the full name anywhere?

Also (referring to my previous comment) we could add the type:

```suggestion
interface Datacenter {
  name: string
  site: string
  type: DatacenterType
}
```

scripts/deploy/deploy-prod-dc.ts
```diff
@@ -44,32 +35,60 @@ export async function main(...args: string[]): Promise<void> {
...
-    await checkTelemetryErrors(uploadPath.split(','), `${currentBrowserSdkVersionMajor}.*`)
+    await checkTelemetryErrors(datacenters, `${currentBrowserSdkVersionMajor}.*`)
   }
 
-  command`node ./scripts/deploy/deploy.ts prod ${version} ${uploadPath}`.withLogs().run()
-  command`node ./scripts/deploy/upload-source-maps.ts ${version} ${uploadPath}`.withLogs().run()
+  const uploadPathTypes = toDatacenterUploadPathType(datacenters).join(',')
+
+  command`node ./scripts/deploy/deploy.ts prod ${version} ${uploadPathTypes}`.withLogs().run()
+  command`node ./scripts/deploy/upload-source-maps.ts ${version} ${uploadPathTypes}`.withLogs().run()
```
BenoitZugmeyer, line 56: **thought:** there is a bit of back-and-forth between datacenters and "upload path types": in `deploy-prod-dc`, we convert datacenters to path types, and in `upload-source-maps` we convert path types back to datacenters. Maybe we can improve this in the future.

Also `deploy-source-maps.ts` is only used here, so we could imagine changing this script to a function. (same for `deploy.ts`)

thomas-lebeau: these two scripts are also used directly in `.gitlab.yml` for deploying to staging, canary and next major version.

I agree though this is a bit of a mess, I have attempted to clean it a bit to be a bit less confusing (differentiating `datacenter` from `uploadPath`, even though if sometimes they are the same, it's not always the case). There is certainly room for more improvement but I'd rather leave that for later

BenoitZugmeyer: > these two scripts are also used directly in .gitlab.yml for deploying to staging, canary and next major version.

Oooh my editor grep is skipping dotfiles, damn :D

PR #4038, opened 2025-12-17
⚗ [RUM-11393]  Add Start/StopAction API

BeltranBulbarellaDD:
## Motivation

Discussed in this [RFC](https://docs.google.com/document/d/1QnH_ZQoG_K7T-mzsF8-iJfM_TFX0E41WOWgJNm7-nBA/edit?tab=t.0#heading=h.rnd972k0hiye)

Tldr: `addEvent` does not cover all the cases needed for complete action collection in frameworks like Flutter. 
This PR adds Start and Stop action.

## Changes

**New Experimental Feature**
- Added `START_STOP_ACTION` feature flag

**New Public APIs**
- Added `startAction(name, options?)` and `stopAction(name, options?)` to RumPublicApi

**Expanded ActionType**

- Added new action types: `TAP`, `SCROLL`, `SWIPE`, `APPLICATION_START`, `BACK`

**Architecture Refactoring**

- Extracted `ActionTracker `(trackAction.ts), shared infrastructure for tracking actions
- Extracted `trackCustomActions.ts`, manages start/stop custom action lifecycle
- Refactored `trackClickActions.ts` to use the shared `ActionTracker`
- Both click actions and custom actions now share the same action context, enabling proper child event attribution

## Test instructions

I used this [HTML](https://gist.github.com/BeltranBulbarellaDD/53619f8357f5e864922a3309eed15665) file to test locally.
Should be able to start an action, generate events and stop the action. When in RUM explorer Attributes for each Event should have the Action.ID.
Example [action](https://app.datadoghq.com/rum/sessions?query=%40type%3Asession%20%40application.id%3A1397744d-34f4-4a6a-a735-801e31c18221&agg_m=count&agg_m_source=base&agg_t=count&event=AwAAAZuNgNdC2tp9JQAAABhBWnVOZ1RWSUFBQW8yNHV5bThMaGNBQUcAAAAkMTE5YjhkOGYtMzAxNi00YmRmLTkzODgtNDFhODYwNmUxMTJkAAABOA&fromUser=false&from_ts=1767533943782&to_ts=1767620343782&live=true) with error, long_task and resource.

[Example](https://app.datadoghq.com/rum/sessions?query=%40type%3Asession%20%40application.id%3A1397744d-34f4-4a6a-a735-801e31c18221&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZuyZFyYZLj10wAAABhBWnV5WlhEa0FBQjVKRmZSbHBGb2RnQUEAAAAkMDE5YmIyNjYtMjJjOS00NzI4LWFmZDUtZTI0MDA5ZGNkNmYyAAAAKw&fromUser=false&sort_by=timestamp&sort_order=desc&viz=stream&from_ts=1768222239123&to_ts=1768225839123&live=true) in Flutter web.

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/scenario/rum/actions.scenario.ts
+++ b/test/e2e/scenario/rum/actions.scenario.ts
@@ -603,40 +603,12 @@ test.describe('custom actions with startAction/stopAction', () => {
 
       expect(actionEvents).toHaveLength(1)
       expect(actionEvents[0].action.error?.count).toBe(1)
+      expect(actionEvents[0].action.frustration?.type).toContain('error_click')
       expect(errorEvents.length).toBeGreaterThanOrEqual(1)
 
       const actionId = actionEvents[0].action.id
       const relatedError = errorEvents.find((e) => hasActionId(e, actionId!))
       expect(relatedError).toBeDefined()
-    })
-
-  createTest('include error_click frustration when action has errors')
-    .withRum({ enableExperimentalFeatures: ['start_stop_action'] })
-    .run(async ({ intakeRegistry, flushEvents, page }) => {
-      await page.evaluate(() => {
-        window.DD_RUM!.startAction('error-action')
-        window.DD_RUM!.addError(new Error('Something went wrong'))
-        window.DD_RUM!.stopAction('error-action')
-      })
-      await flushEvents()
-
-      const actionEvents = intakeRegistry.rumActionEvents
-      expect(actionEvents).toHaveLength(1)
-      expect(actionEvents[0].action.frustration?.type).toContain('error_click')
-    })
-
-  createTest('have empty frustration when action has no errors')
-    .withRum({ enableExperimentalFeatures: ['start_stop_action'] })
-    .run(async ({ intakeRegistry, flushEvents, page }) => {
-      await page.evaluate(() => {
-        window.DD_RUM!.startAction('success-action')
-        window.DD_RUM!.stopAction('success-action')
-      })
-      await flushEvents()
-
-      const actionEvents = intakeRegistry.rumActionEvents
-      expect(actionEvents).toHaveLength(1)
-      expect(actionEvents[0].action.frustration?.type).toEqual([])
     })
 
   createTest('associate a resource to a custom action')
@@ -657,6 +629,7 @@ test.describe('custom actions with startAction/stopAction', () => {
 
       expect(actionEvents).toHaveLength(1)
       expect(actionEvents[0].action.resource?.count).toBe(1)
+      expect(actionEvents[0].action.frustration?.type).toEqual([])
 
       const actionId = actionEvents[0].action.id
       const relatedResource = resourceEvents.find((e) => hasActionId(e, actionId!))


BeltranBulbarellaDD:
@codex

BeltranBulbarellaDD:
@codex pls review

BeltranBulbarellaDD:
@codex review pls

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -32,13 +82,49 @@ export function startActionCollection(
   windowOpenObservable: Observable<void>,
   configuration: RumConfiguration
 ) {
+  const customActionHistory = createValueHistory<string>({ expireDelay: ACTION_CONTEXT_TIME_OUT_DELAY })
+  const activeCustomActions = new Map<string, ActiveCustomAction>()
```
BeltranBulbarellaDD: I wonder if this should be a weakMap like in VitalCollection so it gets garbage collected

BeltranBulbarellaDD:
@codex pls review

packages/rum-core/src/domain/action/actionCollection.ts
BenoitZugmeyer: You can't have a weakmap with primitive types like `string` as key. So no, that won't be an option. This is why FeatureOperation uses an object as "key" instead of a string.

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,35 +165,104 @@ export function startActionCollection(
...
+      return
+    }
 
-  if (configuration.trackUserInteractions) {
-    ;({ actionContexts, stop } = trackClickActions(
+    const lookupKey = getActionLookupKey(name, options.actionKey)
+
+    const existingAction = activeCustomActions.get(lookupKey)
+    if (existingAction) {
+      existingAction.historyEntry.close(clocksNow().relative)
```
BenoitZugmeyer: **suggestion:** 
```suggestion
      existingAction.historyEntry.close(relativeNow())
```

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -32,13 +82,53 @@ export function startActionCollection(
...
+      const customIds = customActionHistory.findAll(startTime)
+
+      const allIds = (clickIds ? (Array.isArray(clickIds) ? clickIds : [clickIds]) : ([] as string[]))
+        .concat(customIds)
+        .filter(Boolean)
+
+      if (allIds.length === 0) {
+        return undefined
+      }
+      return allIds.length === 1 ? allIds[0] : allIds
```
BenoitZugmeyer: **suggestion:** it's simpler and less error prone to always return an array

```suggestion
      return allIds
```

eslint.config.mjs
BeltranBulbarellaDD, line 1: This is because the command `yarn lint` was linting also playwright files.

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -647,12 +668,31 @@ export function makeRumPublicApi(
...
           type: ActionType.CUSTOM,
+          duration: 0 as Duration,
           handlingStack,
         })
         addTelemetryUsage({ feature: 'add-action' })
       })
     },
 
+    startAction: monitor((name, options) => {
+      // addTelemetryUsage({ feature: 'start-action' })
```
BeltranBulbarellaDD, line 684: To be removed when https://github.com/DataDog/rum-events-format/pull/335

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -647,12 +668,31 @@ export function makeRumPublicApi(
...
+      strategy.startAction(sanitize(name)!, {
+        type: sanitize(options && options.type) as ActionType | undefined,
+        context: sanitize(options && options.context) as Context,
+        actionKey: sanitize(options && options.actionKey) as string | undefined,
```
BenoitZugmeyer: **thought:** Since `actionKey` won't be sent to the intake, I'm not sure if we really need to sanitize it

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,35 +142,89 @@ export function startActionCollection(
     })
   )
 
-  let actionContexts: ActionContexts = { findActionId: noop as () => undefined }
-  let stop: () => void = noop
+  function startCustomActionInternal(name: string, options: ActionOptions = {}) {
```
BenoitZugmeyer: **suggestion:** move custom action related code to `trackCustomActions.ts`

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,35 +142,89 @@ export function startActionCollection(
...
+    const { errorCount, resourceCount, longTaskCount } = activeAction.trackedAction.eventCounts
+
+    const customAction: CustomAction = {
+      id: activeAction.trackedAction.id,
+      name: activeAction.name,
+      type: (options.type ?? activeAction.type) || ActionType.CUSTOM,
+      startClocks: activeAction.trackedAction.startClocks,
+      duration,
+      context: combine(activeAction.context, options.context),
+      counts: { errorCount, resourceCount, longTaskCount },
```
BenoitZugmeyer: **suggestion:** 
```suggestion
      counts: activeAction.trackedAction.eventCounts,
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -168,6 +169,24 @@ export interface RumPublicApi extends PublicApi {
...
+  startAction: (name: string, options?: ActionOptions) => void
+
+  /**
+   * Stop a custom action, stored in `@action`
+   *
+   * @category Data Collection
+   * @param name - Name of the action
+   * @param options - Options of the action
+   */
+  stopAction: (name: string, options?: ActionOptions) => void
```
BenoitZugmeyer, line 190: **suggestion:** When released, those functions will appear in the types / in the API reference doc, but they won't work because of the feature flag, which will be confusing. 

You could either:
* define those functions only if the flag is provided ([ex](https://github.com/DataDog/browser-sdk/blob/d3065fd800dde0c48db1b138edf3840f9ac735ff/packages/rum-core/src/boot/rumPublicApi.ts#L323))
* document them as `experimental` ([ex](https://github.com/DataDog/browser-sdk/blob/7f704e04e525437a3f6ecc40346a415ae5a6f1e2/packages/rum-core/src/boot/rumPublicApi.ts#L433-L440))

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -339,19 +306,19 @@ function newClick(
...
 
-      const { resourceCount, errorCount, longTaskCount } = eventCountsSubscription.eventCounts
+      const { resourceCount, errorCount, longTaskCount } = trackedAction.eventCounts
       const clickAction: ClickAction = {
         duration: activityEndTime && elapsed(startClocks.timeStamp, activityEndTime),
         startClocks,
-        id,
+        id: trackedAction.id,
         frustrationTypes,
         counts: {
```
BenoitZugmeyer: **suggestion:** don't destructure counts, you could just pass the object directly

```
counts: trackedAction.eventCounts
```

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -339,19 +306,19 @@ function newClick(
...
       stop()
       if (status !== ClickStatus.STOPPED) {
         return
       }
 
-      const { resourceCount, errorCount, longTaskCount } = eventCountsSubscription.eventCounts
+      const { resourceCount, errorCount, longTaskCount } = trackedAction.eventCounts
       const clickAction: ClickAction = {
         duration: activityEndTime && elapsed(startClocks.timeStamp, activityEndTime),
         startClocks,
```
BenoitZugmeyer: **suggestion:** see if you could compute `duration` in `trackedAction` and use both `duration` and `startClocks` from `trackedAction`. It could be interesting to have something like this:

```ts
const clickAction: ClickAction = {
  startClocks: trackedAction.startClocks,
  duration: trackedAction.duration,
  events: trackedAction.events,
  id: trackedAction.id,
  ...
}
```

(we could imagine having `type` and `name` in `trackedAction` too)

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,35 +151,91 @@ export function startActionCollection(
...
+      activeCustomActions.forEach((activeAction) => {
+        activeAction.trackedAction.discard()
+      })
+      activeCustomActions.clear()
+      actionTracker.stop()
     },
   }
 }
 
 function processAction(action: AutoAction | CustomAction): RawRumEventCollectedData<RawRumActionEvent> {
```
BenoitZugmeyer: **suggestion:** revamp this function. I believe there is a lot of redundant code now, since `CustomAction` and `AutoAction` are much closer (they both have id, duration and event counts)

Maybe it could look like this:
```ts
function processAction(action: AutoAction | CustomAction): RawRumEventCollectedData<RawRumActionEvent> {
  const actionEvent: RawRumActionEvent = {
    type: RumEventType.ACTION,
    date: action.startClocks.timeStamp,
    action: {
      id: action.id, // make `CustomAction['id']` required
      loading_time: discardNegativeDuration(toServerDuration(action.duration)),
      target: { name: action.name },
      type: action.type,
      ...(action.counts && {
        error: { count: action.counts.errorCount },
        long_task: { count: action.counts.longTaskCount },
        resource: { count: action.counts.resourceCount },
      }),
      frustration: isAutoAction(action) ? { type: action.frustrationTypes } : undefined,
    },
    context: isAutoAction(action) ? undefined : action.context,
    _dd: isAutoAction(action)
      ? {
          action: {
            target: action.target,
            position: action.position,
            name_source: action.nameSource,
          },
        }
      : undefined,
  }

  const domainContext: RumActionEventDomainContext = isAutoAction(action)
    ? { events: action.events }
    : { handlingStack: action.handlingStack }

  return {
    rawRumEvent: actionEvent,
    duration: action.duration,
    startTime: action.startClocks.relative,
    domainContext,
  }
}
```

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -9,36 +21,100 @@ import type { RumConfiguration } from '../configuration'
...
+
+  /**
+   * @internal - used to preserve timing for pre-init calls
+   */
+  startClocks?: ClocksState
+
+  /**
+   * @internal - used to preserve timing for pre-init calls
+   */
+  stopClocks?: ClocksState
```
BenoitZugmeyer: **suggestion:** don't expose internal types if you don't have to. Instead, you could change the start/stop action functions exposed by `actionCollection` to be like:

```ts
startCustomActionInternal(name: string, options: ActionOptions = {}, startClocks = clocksNow())
stopCustomActionInternal(name: string, options: ActionOptions = {}, stopClocks = clocksNow())
```

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,35 +151,91 @@ export function startActionCollection(
...
+    if (!activeAction) {
+      return
+    }
+
+    const stopClocks = options.stopClocks ?? clocksNow()
+    const duration = elapsed(activeAction.trackedAction.startClocks.timeStamp, stopClocks.timeStamp)
+
+    activeAction.trackedAction.stop(stopClocks.relative)
+
+    const { errorCount, resourceCount, longTaskCount } = activeAction.trackedAction.eventCounts
```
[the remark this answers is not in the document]
BenoitZugmeyer: Technically correct, but for this to work we would also need to delay assembling/sending the action to wait for potential sub-events incrementing the counter. So this is a whole different project.

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -0,0 +1,107 @@
...
+  longTaskCount: number
+  resourceCount: number
+}
+
+export interface TrackedAction {
+  id: string
+  startClocks: ClocksState
+  eventCounts: EventCounts
+  stop: (endTime?: RelativeTime) => void
+  discard: () => void
```
BenoitZugmeyer: **suggestion:** make `endTime` required so it's not redundant with `discard`
```suggestion
  stop: (endTime: RelativeTime) => void
  discard: () => void
```

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -0,0 +1,107 @@
...
+    return {
+      id,
+      startClocks,
+      get eventCounts() {
+        return eventCountsSubscription.eventCounts
+      },
+      stop: stopTracking,
+      discard: () => {
+        stopTracking()
+      },
```
BenoitZugmeyer: **suggestion:** rename `stockTracking` to `stopOrDiscard` and return it like this:

```suggestion
      stop: stopOrDiscard,
      discard: stopOrDiscard
```

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -9,36 +21,100 @@ import type { RumConfiguration } from '../configuration'
...
+  /**
+   * @internal - used to preserve timing for pre-init calls
+   */
+  stopClocks?: ClocksState
+}
+
+interface ActiveCustomAction extends ActionOptions {
+  name: string
+  trackedAction: TrackedAction
+}
```
BenoitZugmeyer: **suggestion:** to avoid multiplying the different concepts, maybe it could be simpler to have `name` `type` `context` and `actionKey` directly in `trackedAction`

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -9,36 +21,100 @@ import type { RumConfiguration } from '../configuration'
...
+      lifeCycle,
+      domMutationObservable,
+      windowOpenObservable,
+      configuration,
+      actionTracker
+    ))
+  }
+
+  const actionContexts: ActionContexts = {
+    findActionId: (startTime?: RelativeTime) => actionTracker.findActionId(startTime),
```
BenoitZugmeyer: **nitpick:** 
```suggestion
    findActionId: actionTracker.findActionId
```

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -152,5 +298,9 @@ function processAction(action: AutoAction | CustomAction): RawRumEventCollectedD
 }
 
 function isAutoAction(action: AutoAction | CustomAction): action is AutoAction {
-  return action.type !== ActionType.CUSTOM
+  return action.type === ActionType.CLICK && 'events' in action
+}
+
+function getActionLookupKey(name: string, actionKey?: string): string {
+  return JSON.stringify({ name, actionKey })
```
BenoitZugmeyer: **thought:** I'm not sure we need to care if the same action key is used with different name... do we?

packages/rum-core/src/domain/action/actionCollection.ts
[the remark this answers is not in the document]
BeltranBulbarellaDD: Agreed. I saw it but decided to ignore. "Know limitation"

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    return {
+      id,
+      startClocks,
+      get duration() {
+        return duration
+      },
+      get counts() {
+        const { errorCount, longTaskCount, resourceCount } = eventCountsSubscription.eventCounts
```
BeltranBulbarellaDD: Going to ask an agent how to try to reduce bundle size. For example here.

packages/rum-core/src/domain/action/actionCollection.ts
BeltranBulbarellaDD: Not really, I added it just to be safe. But agreed to remove it.

BeltranBulbarellaDD:
@codex can you give a review and suggest where we could save on bundle size? Is there room for improvement?

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -43,9 +40,9 @@ export function startActionTracker(lifeCycle: LifeCycle): ActionTracker {
   const history = createValueHistory<string>({ expireDelay: ACTION_CONTEXT_TIME_OUT_DELAY })
   const activeEventCountSubscriptions = new Set<ReturnType<typeof trackEventCounts>>()
 
-  const { unsubscribe: unsubscribeSessionRenewal } = lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
+  const sessionRenewalSubscription = lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, () => {
     history.reset()
-    activeEventCountSubscriptions.forEach((subscription) => subscription.stop())
+    activeEventCountSubscriptions.forEach((s) => s.stop())
```
BenoitZugmeyer: **nitpick:** this has no impact on bundle size, as the name will be mangled anyway.

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,82 +114,74 @@ export function startActionCollection(
...
+        type: action.type,
+      },
+      context: action.context,
+    },
+    duration: action.duration,
     startTime: action.startClocks.relative,
-    domainContext,
+    domainContext: { handlingStack: action.handlingStack },
   }
 }
```
BenoitZugmeyer: **suggestion:** I preferred your original version where we treat all actions in the same place. This makes code shorter, because we don't have to repeat common fields.

Suggestion: remove `InstantCustomAction` and adjust `CustomAction` to make `duration` optional, like you had before.

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -38,43 +38,30 @@ export interface ActionTracker {
...
   })
 
   function createTrackedAction(startClocks: ClocksState, metadata?: TrackedActionMetadata): TrackedAction {
     const id = generateUUID()
     const historyEntry: ValueHistoryEntry<string> = history.add(id, startClocks.relative)
-    let stopped = false
     let duration: Duration | undefined
 
-    const eventCountsSubscription = trackEventCounts({
+    const sub = trackEventCounts({
```
BenoitZugmeyer: **suggestion:** here too. Renaming variables have no impact on the bundle size, you should probably go back to the more explicit names

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -0,0 +1,103 @@
...
+export const ACTION_CONTEXT_TIME_OUT_DELAY = 5 * ONE_MINUTE // arbitrary
+
+export type ActionCounts = EventCounts
+
+export interface TrackedActionMetadata {
+  name?: string
+  type?: ActionType
+  context?: Context
+  actionKey?: string
+}
```
BenoitZugmeyer: **thought:** I suggested to store name/type and other properties in `TrackedAction` to remove a data model and factorize a bit of code with `trackClickActions`. It looks like it didn't too well, as click actions has other requirements (needs to be `clone()`). So maybe revert this change, if you feel like it.

packages/rum-core/src/domain/action/actionCollection.ts
BenoitZugmeyer: This is what I have in mind: https://gist.github.com/BenoitZugmeyer/d1a786fda32f1bdbb0eb747b998081b4

packages/rum-core/src/domain/action/actionCollection.ts
```diff
@@ -75,82 +105,60 @@ export function startActionCollection(
...
-  }
-
   return {
-    addAction: (action: CustomAction) => {
-      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction(action))
+    addAction: (action: Omit<CustomAction, 'id' | 'duration' | 'counts'>) => {
+      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processAction({ id: generateUUID(), ...action }))
     },
+    startAction: customActions?.startAction ?? noop,
+    stopAction: customActions?.stopAction ?? noop,
```
amortemousque: **💬 suggestion:** ‏For experimental public API we tend to handle it in `rumPublicAPI.ts` like in [here](https://github.com/DataDog/browser-sdk/pull/3031/files#diff-f16464c555cdc03f7a8b03f1b312bb09ea41cd1acf4fa773629e880a5a1ea72b)

packages/rum-core/src/domain/action/trackCustomActions.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+interface ActiveCustomAction {
+  name: string
+  type?: ActionType
+  context?: Context
+  trackedAction: TrackedAction
+}
+
+export function trackCustomActions(
```
amortemousque: **💬 suggestion:** ‏‏
- What we used to call a custom action had the custom type, but in this module it can be any action type. We could rename it to trackManualActions to separate it from automatic click tracking.
- It would also help to collocate the addAction API in this module.
- FMU, we don’t handle frustration signals yet. I’m thinking in particular about error actions mentioned in the RFC.

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -168,6 +169,24 @@ export interface RumPublicApi extends PublicApi {
    */
   addAction: (name: string, context?: object) => void
 
+  /**
+   * [Experimental] start a custom action, stored in `@action`
```
amortemousque: Nit: The comment is misleading since we can create any type of action with this API
```suggestion
   * [Experimental] start an action, stored in `@action`
```

packages/rum-core/src/domain/action/trackCustomActions.ts
BeltranBulbarellaDD: > FMU, we don’t handle frustration signals yet. I’m thinking in particular about error actions mentioned in the RFC.

Yeah, what do you think about leaving it for a next PR?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -653,6 +676,31 @@ export function makeRumPublicApi(
       })
     },
 
+    startAction: monitor((name, options) => {
+      // Check feature flag only after init; pre-init calls should be buffered
```
BeltranBulbarellaDD, line 680: Important for the e2e test

packages/rum-core/src/domain/action/trackCustomActions.ts
amortemousque: Yeah, but it might require changing your current approach. It would be good to anticipate, at least a bit, how you’d handle it.

packages/rum-core/src/domain/action/trackManualActions.ts
```diff
@@ -0,0 +1,113 @@
...
+  type: ActionType
+  name: string
+  startClocks: ClocksState
+  duration?: Duration
+  context?: Context
+  handlingStack?: string
+  counts?: ActionCounts
+}
+
+interface ActiveManualAction {
```
amortemousque: **💬 suggestion:** ‏Let's use the same naming convention as vitals.
```suggestion
interface ManualActionStart {
```

packages/rum-core/src/domain/action/trackCustomActions.ts
BeltranBulbarellaDD: Adding `ERROR_CLICK` would not be a challenge and I could add it now. But the others could wait as they are much harder and mobile does not send them. WDYT?

packages/rum-core/src/domain/action/trackAction.ts
```diff
@@ -0,0 +1,94 @@
...
+  findActionId: (startTime?: RelativeTime) => string | string[] | undefined
+}
+
+export interface ActionTracker {
+  createTrackedAction: (startClocks: ClocksState) => TrackedAction
+  findActionId: (startTime?: RelativeTime) => string | string[] | undefined
+  stop: () => void
+}
+
+export function startActionTracker(lifeCycle: LifeCycle): ActionTracker {
```
amortemousque, line 31: **💭 thought:** ‏With this implementation, custom actions can have a duration and be linked to other events. I think it’s a good direction, but we should check with Product to get their view. We should also align with Frontend to make sure they can handle these new action types coming from the browser.

packages/rum-core/src/domain/action/trackAction.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+  let lifeCycle: LifeCycle
+  let actionTracker: ActionTracker
+
+  beforeEach(() => {
+    lifeCycle = new LifeCycle()
+    actionTracker = startActionTracker(lifeCycle)
+    registerCleanupTask(() => actionTracker.stop())
+  })
+
+  describe('createTrackedAction', () => {
```
amortemousque, line 18: **🥜 nitpick:** ‏We could simplify the tests by having a single test that checks the initial state of `createTrackedAction`: "should have an ID, start clocks and event counts to zero"

packages/rum-core/src/domain/action/trackAction.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+      const startClocks = { relative: 100 as RelativeTime, timeStamp: 1000 as TimeStamp }
+      const trackedAction = actionTracker.createTrackedAction(startClocks)
+
+      expect(trackedAction.counts.errorCount).toBe(0)
+      expect(trackedAction.counts.resourceCount).toBe(0)
+      expect(trackedAction.counts.longTaskCount).toBe(0)
+    })
+  })
+
+  describe('event counting', () => {
```
amortemousque, line 39: **💬 suggestion:** I think we’re testing the underlying trackEventCounts behavior. We could simplify the test to focus only on what’s specific to this module, for example:
"should count child event associated with the action"
"should not count child events unrelated to the action"

packages/rum-core/src/domain/action/trackAction.spec.ts
BeltranBulbarellaDD: Thought about it! I though it would be easier to debug which test fails explicitly? But interested in your thoughts for next time too.

packages/rum-core/src/domain/action/trackCustomActions.ts
amortemousque: I was only thinking of covering ERROR_CLICK. So nice ;)

PR #4042, opened 2025-12-17
[Extension] Handle function values in configuration

yannickadam:
## Motivation

When inspecting the RUM/Logs configuration using the Browser SDK extension, functions are shown as `{empty object}`.
image
This can mislead users to think that it's undefined / not present. 

## Changes

- Pass function metadata when retrieving the SDKs configuration and display a snippet of the source code:
image
- Fix an issue (already present) when overriding the configuration. Nested functions (such as the ones in `plugins` were lost)
- Add a `Reveal in console` feature. This logs the function in the console, allowing developers to click on it to see the code in Chrome's Source tab.

## Test instructions

- Build the extension locally. 
- Visit sites which use beforeSend and/or other configurations using functions.
- Check that you can see the code, reveal in the explorer.
- Check that you can override the (rest of the) configuration without breaking these functions. They cannot be overridden. 

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/developer-extension/src/panel/components/json.tsx
+++ b/developer-extension/src/panel/components/json.tsx
@@ -4,21 +4,17 @@ import { useColorScheme } from '@mantine/hooks'
 import { IconCopy, IconSearch } from '@tabler/icons-react'
 import type { ForwardedRef, ReactNode } from 'react'
 import React, { forwardRef, useContext, createContext, useState } from 'react'
-import { createLogger } from '../../common/logger'
 import { copy } from '../copy'
-import { evalInWindow } from '../evalInWindow'
 import { formatNumber } from '../formatNumber'
 
 import * as classes from './json.module.css'
-
-const logger = createLogger('json')
 
 interface JsonProps {
   value: unknown
   defaultCollapseLevel?: number
   getMenuItemsForPath?: GetMenuItemsForPath
   formatValue?: FormatValue
-  sdkType?: 'rum' | 'logs'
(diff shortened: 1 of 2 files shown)


developer-extension/src/panel/components/json.tsx
```diff
@@ -185,7 +263,7 @@ function JsonValue({ descriptor }: { descriptor: JsonValueDescriptor }) {
               parentType: 'array',
               parentValue: descriptor.value as unknown[],
               value: child,
-              path: descriptor.path,
+              path: descriptor.path ? `${descriptor.path}.${i}` : String(i),
```
BenoitZugmeyer: Issue: The omission of the index was intentional. It follows the events platform query, so you can copy a path and paste it in the Datadog ui.

developer-extension/src/panel/components/json.tsx
```diff
@@ -44,6 +49,7 @@ const JsonContext = createContext<{
   defaultCollapseLevel: number
   getMenuItemsForPath?: GetMenuItemsForPath
   formatValue: FormatValue
+  sdkType?: 'rum' | 'logs'
```
BenoitZugmeyer: Suggestion: it could be nice to keep the json component more or less generic, because it's used in multiple places, not only for displaying configuration. Maybe add a `onRevealFunctionLocation` prop and externalize that logic?

BeltranBulbarellaDD (approved):
LGTM!

PR #4044, opened 2025-12-18
⚗️ [RUM-13259]Track action names in shadow dom

rgaignault:
## Motivation

Clicks inside Shadow DOM components currently result in imprecise action names because event.target is retargeted to the shadow host. 

## Changes

This PR adds experimental support for extracting accurate action names from elements inside Shadow DOM

 - betaTrackActionsInShadowDom Enables action name extraction from Shadow DOM elements
 - Uses event.composedPath() to get the actual clicked element
 
This is the first phase of Shadow DOM support, focusing on action names only. Selector tracking inside Shadow DOM will be added in a follow-up PR. This feature is experimental and gated behind the betaTrackActionsInShadowDom opt-in parameter.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
+++ b/packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
@@ -988,7 +988,7 @@ describe('getActionNameFromElement', () => {
     })
 
     it('falls back to document when aria-labelledby references an element outside shadow DOM', () => {
-      const label = appendElement('<span id="light-dom-label">External Label</span>')
+      appendElement('<span id="light-dom-label">External Label</span>')
 
       const host = appendElement('<div></div>')
       const shadowRoot = host.attachShadow({ mode: 'open' })


packages/rum-core/src/domain/action/getActionNameFromElement.spec.ts
```diff
@@ -930,4 +930,61 @@ describe('getActionNameFromElement', () => {
       })
     })
   })
+
+  describe('shadow DOM support', () => {
```
amortemousque, line 934: **💬 suggestion:** ‏Could be nice to also test privacy settings with shadow dom?

rgaignault: Good idea ! I added one 👍

packages/rum-core/src/domain/action/getActionNameFromElement.ts
```diff
@@ -173,7 +184,15 @@ function truncate(s: string) {
...
 }
 
-function getElementById(refElement: Element, id: string) {
+function getElementById(refElement: Element, id: string): HTMLElement | null {
+  const rootNode = refElement.getRootNode()
+  if (rootNode instanceof ShadowRoot) {
+    const shadowElement = rootNode.getElementById(id)
+    if (shadowElement) {
+      return shadowElement
+    }
```
amortemousque, line 193: **❓ question:** ‏ Why not simply
```suggestion
   return rootNode.getElementById(id)
```

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -203,6 +203,15 @@ export interface RumInitConfiguration extends InitConfiguration {
...
   actionNameAttribute?: string | undefined
 
+  /**
+   * Enables accurate action names for clicks inside Shadow DOM elements.
+   * ⚠️ This is a beta feature and will be updated in the future with selector tracking.
+   *
+   * @category Beta
+   * @defaultValue false
+   */
+  betaTrackActionsInShadowDom?: boolean | undefined
```
amortemousque, line 213: **💭 thought:** I don’t think we need the `beta` prefix. IMO, the implementation is not too complex to justify it.
I recommend putting the feature behind a feature flag so we can easily enable or disable it during dogfooding.

packages/rum-core/src/domain/action/getActionNameFromElement.ts
rgaignault: We need this to fallback on the original document  because if an element inside a shadow DOM has an aria-labelledby pointing to an ID in the main document, we would return null instead of looking it up in the document.
I added a test for this 👍

PR #4045, opened 2025-12-18
🐛 Fix profiling <-> long task association regression

amortemousque:
## Motivation

This PR fixes a regression introduced in https://github.com/DataDog/browser-sdk/pull/4013, where the profiler only linked long tasks that were active at startTime. It now links every long task that occurs during the profiling session.

## Changes

- Call `findLongTasks` with the duration to get all long tasks that occur within that period.
- Update unit tests to cover this case

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/longTask/longTaskCollection.spec.ts
```diff
@@ -104,7 +104,7 @@ describe('longTaskCollection', () => {
       notifyPerformanceEntries([entry])
 
       const longTask = (rawRumEvents[0].rawRumEvent as RawRumLongTaskEvent).long_task
-      const longTasks = longTaskCollection.longTaskContexts.findLongTasks(1234 as RelativeTime)
```
BenoitZugmeyer, line 107: **suggestion:** could you adjust or add a test to illustrate the issue? In this case, if you were using `findLongTasks` with a time older than the long task start time, the long task wasn't returned.

Edit: I see you added tests below, but maybe add a test closer to the actual issue could be nice

PR #4049, opened 2025-12-19
🎨 [PANA-5260] Consolidate recorder object id tracking code

sethfowler-datadog:
## Motivation

To correlate references to the same object in different records, the recording code needs to assign ids to these objects. Today, we track ids for two kinds of objects:
* DOM nodes. (`Node` objects.)
* Events. (`Event` objects.)

The objects that manage these two kinds of ids are defined in an almost identical way; the only real difference is the naming scheme of the methods.

The new data format will add to the duplication by tracking ids for two new kinds of objects:
* Strings. (`string` values.)
* Stylesheets. (`CSSStyleSheet` objects.)

The new data format also requires the capability to clear these mappings, so the implementation of the tracking code will get slightly more complicated.

Instead of following the existing pattern and duplicating nearly identical code in four places, let's build a single implementation that can handle id tracking for any kind of object or value.

## Changes

This PR replaces the existing `EventIds` and `NodeIds` implementations with simple wrappers around a shared generic type, `ItemIds`. It also implements the new id tracking facilities that the new data format will need by adding `StringIds` and `StyleSheetIds` wrappers.

As the number of id collections grows, `createRecordingScope()` is starting to require a large number of function arguments. Since these are simple data structures with very little behavior, there isn't much benefit in using dependency injection for them. So, `createRecordingScope()` now just creates the id collections internally, simplifying its interface.

The `NodeIds` collection was a special case in that it had a unique method, `NodeIds#areAssignedForNodeAndAncestors()`. This is only called in one place, `trackMutations.ts`, so I've moved the implementation there as a free function, and moved the tests to `trackMutations.spec.ts`.

## Test instructions

This is pure refactoring; there are no functional changes. So, testing that recording continues to work with the browser SDK extension is enough.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/itemIds.ts
+++ b/packages/rum/src/domain/record/itemIds.ts
@@ -38,6 +38,7 @@ export interface ItemIds<ItemType, ItemId extends number> {
   clear(this: void): void
   get(this: void, item: ItemType): ItemId | undefined
   getOrInsert(this: void, item: ItemType): ItemId
+  get size(): number
 }
 
 function createIdMap<ItemType, ItemId extends number>(firstId: ItemId): ItemIds<ItemType, ItemId> {
@@ -76,6 +77,9 @@ function createItemIds<ItemType, ItemId extends number>(
         map.set(object, id)
       }
       return id
+    },
+    get size(): number {
+      return nextId - firstId
     },
   }
 }
--- a/packages/rum/src/domain/record/itemIds.spec.ts
+++ b/packages/rum/src/domain/record/itemIds.spec.ts
@@ -81,6 +81,16 @@ describe('ItemIds', () => {
           expect(itemIds.get(item)).toBe(itemId)
         })
       })
+
+      describe('size', () => {
+        it('increments when an id is assigned', () => {
+          expect(itemIds.size).toBe(0)
+          itemIds.getOrInsert(createItem())
+          expect(itemIds.size).toBe(1)
+          itemIds.getOrInsert(createItem())
+          expect(itemIds.size).toBe(2)
+        })
+      })
     })
   }
 


packages/rum/src/domain/record/itemIds.ts
```diff
@@ -0,0 +1,81 @@
...
+  FIRST_ID = 0,
+}
+export function createStyleSheetIds(): StyleSheetIds {
+  return createWeakIdMap(StyleSheetIdConstants.FIRST_ID as StyleSheetId)
+}
+
+export interface ItemIds<ItemType, ItemId extends number> {
+  clear(): void
+  get(this: void, item: ItemType): ItemId | undefined
+  getOrInsert(this: void, item: ItemType): ItemId
```
sethfowler-datadog, line 40: This method was called `assign()` for the old `NodeIds` and `getIdForEvent()` for the old `EventIds`, but it had the same behavior in both cases. I've renamed it to use the same name as [the standard Map method that does essentially the same thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/getOrInsert). I think this name makes the behavior a bit clearer than either of the previous names.

packages/rum/src/domain/record/itemIds.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function createIdMap<ItemType, ItemId extends number>(firstId: ItemId): ItemIds<ItemType, ItemId> {
+  return createItemIds(() => new Map<ItemType, ItemId>(), firstId)
+}
+
+function createWeakIdMap<ItemType extends object, ItemId extends number>(firstId: ItemId): ItemIds<ItemType, ItemId> {
+  return createItemIds(() => new WeakMap<ItemType, ItemId>(), firstId)
+}
```
sethfowler-datadog, line 50: We need two variations here because `string`s cannot be stored in a `WeakMap`, since they're value types.

packages/rum/src/domain/record/itemIds.ts
```diff
@@ -0,0 +1,81 @@
...
+export function createNodeIds(): NodeIds {
+  return createWeakIdMap(NodeIdConstants.FIRST_ID as NodeId)
+}
+
+export type StringId = number & { __brand: 'StringId' }
+export type StringIds = ItemIds<string, StringId>
+export const enum StringIdConstants {
+  FIRST_ID = 0,
+}
+export function createStringIds(): StringIds {
```
gonzalezreal, line 24: As far as I understand string ids can grow unbounded until clear is called. When / how is clear called?

sethfowler-datadog: The `StringIds` table isn't actually getting populated in this PR (since only the new data format uses it), but in the next PR that adds the serialization code for the new data format, `clear()` will get called when we take a new full snapshot -- in other words, when we transition between views.

If the table gets too big before we transition to a new view, what I had planned to do was to just stop adding new strings to the table, rather than clear it. (The data format allows us to just send a literal string wherever we'd send a string id, so we never _need_ to add a string to the string table.) My guess is that frequently repeated strings will already be in the table at that point, so we'll still get a lot of value out of the table even if no new strings get added. Once the new data format is used more widely, we should be able to gather some telemetry to determine whether that's true in practice, or whether another approach would make more sense -- we could even decide to just clear all the tables and take a new full snapshot when that happens!

sethfowler-datadog: Regarding the other tables: for the new serialization algorithm, they'll also get cleared when we take a new full snapshot. For the current serialization algorithm, the existing behavior is that they never get cleared, but since they use `WeakMap`s, as you say, they more or less auto-GC with time.

gonzalezreal: Thanks for the clarification 🙂

PR #4060, opened 2026-01-05
✨[PANA-3971] Add a more compact experimental DOM mutation encoding

sethfowler-datadog:
## Motivation

We have observed that there are a number of opportunities for reducing the size of session replay data. This would be beneficial for a number of reasons:
* We'd use less client-side resources, decreasing the performance impact of the SDK.
* Along the same lines, we'd have less impact on end user mobile data and battery life.
* Fewer bytes take less time to transmit, which reduces the likelihood that the data is lost because the user navigates away from the page or loses network connectivity.

Less overhead, more reliability; seems like a clear win!

This PR takes a significant step in that direction by adding an experimental implementation of a new encoding for DOM mutations. In experiments, this new encoding reduced the size of session replay data by 77.3% before compression, and by 34.7% after compression is taken into account. (Both of these numbers exclude stylesheet data, which requires a somewhat different approach; this is a separate opportunity which isn't addressed in this PR.)

The new encoding is a drop-in replacement for the old encoding; it can be converted to the old representation losslessly, producing a byte-for-byte identical result. The tests included in this PR verify that this is true for the session replay unit tests. For additional verification in a real-world scenario, I plan to push a commit to staging tomorrow that will perform the same verification on all recorded pages within our staging environment.

## Changes

A new experimental feature, `USE_CHANGE_RECORDS`, has been added. Enabling this feature has the following effects:
1. The full snapshot is captured using a new serialization function, `serializeNodeAsChange`, which produces a `Change` record instead of a `FullSnapshot` record. `serializeNodeAsChange` and the functions it recursively calls are the heart of the new encoding implementation.
2. The `RecordingScope` data structure resets between full snapshots; all ids are cleared. The new encoding assigns ids implicitly in the order that nodes and other objects are encountered; this means that there is no way to carry over ids from a different full snapshot, so resetting the `RecordingScope` is necessary.

`serializeNodeAsChange` is supported by several new types that help implement its functionality:
1. The new `ChangeSerializationTransaction` type exposes builder methods which are used to construct the `Change` record. Once we're confident in this code, we can move this functionality onto the existing `SerializationTransaction` type; I've separated things out in this PR only to keep the new code as isolated as possible.
2. `InsertionCursor` tracks the current position in the DOM during the serialization process. It's responsible for allocating node ids and generating an `InsertionPosition` for each node; these `InsertionPosition`s are used in the new encoding to tell the player where in the DOM tree to insert each node.
3. `ChangeEncoder` automatically optimizes each `Change` record as it's being constructed. It splits out strings into a separate string table for deduplication purposes, and it groups changes of the same type together to minimize the number of bytes that must be devoted to identifying the type of each change.

It's probably also worth calling out `changeConversions.ts`, which is unfortunately by far the largest file in this PR despite not actually being code that runs as part of the browser SDK. This file contains `convertChangeToFullSnapshot()`, which is used in tests to verify that `serializeNodeAsChange()` produces byte-for-byte identical results to `serializeNode()`. It would ordinarily be called `changeConversions.specHelper.ts`, but I need it to be temporarily importable in non-test code so I can use it to perform validation on staging. I'll give it the `.specHelper.ts` extension before merging. (EDIT: I've completed this validation and renamed the file to `changeConversions.specHelper.ts`.)

There are some changes which are deliberately not included in this PR:
* Support for incremental mutations will be added in a separate PR. This one is huge as it is, and I felt it would be better to review those changes in a more focused way, since incremental mutations are more complex. I've tried to remove all vestiges of them in this PR. (This is why, for example, a new `InsertionCursor` can only be positioned at the root of the DOM tree right now.)
* I want to add some new telemetry associated with this change, but I need to make some changes to the implementation of the old encoding as well, so I'll handle telemetry in a followup PR. Within this PR, the goal is to generate the same telemetry that we generate right now, and nothing further. (Edit: This is #4077.)
* I plan to add support for rendering `Change` records to the session replay sandbox immediately, but that code lives in a separate repo, so it can't be included in this PR. I'll bump the session replay sandbox version used in the browser extension in a followup. (Edit: This was #4072.)

## Test instructions

This PR has been validated by pushing a commit to staging that generates serializations in both formats for every full snapshot, converts the `BrowserChangeRecord` to a `BrowserFullSnapshotRecord`. The current version of the PR produces `BrowserChangeRecord`s which, when converted to `BrowserFullSnapshotRecord`s, are byte-for-byte identical in all circumstances encountered on staging.

Manually testing the changes is currently a bit tricky, since the browser SDK extension does not support rendering `BrowserChangeRecord`s. I've merged a PR for the replay code that adds this functionality, so it should be possible to test things visually in an easier way soon, but that PR has not been deployed yet. I'm happy to share the details on Slack, if you want to test things visually.

## Checklist

- [x] Tested locally
- [x] Tested on staging 

### DIFF
--- a/packages/rum/src/domain/record/serialization/insertionCursor.ts
+++ b/packages/rum/src/domain/record/serialization/insertionCursor.ts
@@ -50,14 +50,12 @@ export interface InsertionCursor {
 /** Returns an InsertionCursor which starts positioned at the root of the document. */
 export function createRootInsertionCursor(nodeIds: NodeIds): InsertionCursor {
   interface ChildListCursor {
-    parentCursor: ChildListCursor | undefined
-    parentId: NodeId | undefined
+    parent: ChildListCursor | undefined
     previousSiblingId: NodeId | undefined
   }
 
   let cursor: ChildListCursor = {
-    parentCursor: undefined,
-    parentId: undefined,
+    parent: undefined,
     previousSiblingId: undefined,
   }
 
@@ -66,11 +64,14 @@ export function createRootInsertionCursor(nodeIds: NodeIds): InsertionCursor {
       // Use an AppendAfterPreviousInsertionPoint. (i.e., 0)
       return 0
     }
-    if (cursor.parentId !== undefined) {
+
+    const parentId = cursor.parent?.previousSiblingId
+    if (parentId !== undefined) {
       // Use an AppendChildInsertionPoint. We identify the parent node using a positive
       // integer indicating the difference between the new node's id and its parent's id.
-      return nodeId - cursor.parentId
+      return nodeId - parentId
     }
+
     // There's no parent. Use a RootInsertionPoint. (i.e., null)
     return null
   }
@@ -83,15 +84,14 @@ export function createRootInsertionCursor(nodeIds: NodeIds): InsertionCursor {
       return { nodeId, insertionPoint }
     },
     ascend(): void {
-      if (cursor.parentCursor) {
-        cursor = cursor.parentCursor
+      if (cursor.parent) {
+        cursor = cursor.parent
       }
     },
     descend(): void {
       if (cursor.previousSiblingId !== undefined) {
         cursor = {
-          parentCursor: cursor,
-          parentId: cursor.previousSiblingId,
+          parent: cursor,
           previousSiblingId: undefined,
         }
       }


packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  if (privacyLevel === NodePrivacyLevel.HIDDEN) {
+    serializeHiddenNodePlaceholder(cursor, node, transaction)
+    return
+  }
+
+  // Totally ignore risky or unwanted elements. (e.g. <script>, some <link> and <meta> elements)
+  if (privacyLevel === NodePrivacyLevel.IGNORE) {
+    return
+  }
```
yannickadam, line 43: Any reason to execute this on every node instead of just on element nodes as before?

packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
```diff
@@ -0,0 +1,233 @@
...
+  cursor.ascend()
+}
+
+function serializeDocumentNodeAsChange(
+  cursor: InsertionCursor,
+  document: Document,
+  transaction: ChangeSerializationTransaction
+): void {
+  const { nodeId, insertionPoint } = cursor.advance(document)
+  transaction.addNode(insertionPoint, '#document')
```
yannickadam, line 95: We used to have `NodeType[type]` (which was misleading because it did not match https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType ) Still, shouldn't we keep using numbers instead of strings and use this opportunity to match the spec?

Fhou (approved):

packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
sethfowler-datadog: Regarding the first part, where we call `getNodeSelfPrivacyLevel()`, it seems more different than it really is. `getNodeSelfPrivacyLevel()` immediately returns `undefined` for non-elements, and in response we immediately set `privacyLevel` to `parentPrivacyLevel` and skip the invocation of `reducePrivacyLevel()`.

Regarding the second part, where we make checks like `if privacyLevel === NodePrivacyLevel.HIDDEN`, this part needs to be there for correctness, because external callers can pass arbitrary nodes into this function, and so we might be handed a non-element node without passing through an element ancestor first. The old implementation would've been better off having these checks, too; the current state of affairs requires callers to [defensively implement the checks themselves](https://github.com/DataDog/browser-sdk/blob/dd89e2316859f43d8567aa52b79bbcbbd286d8f6/packages/rum/src/domain/record/trackers/trackMutation.ts#L219-L221), with the potential for a privacy bug if they forget.

sethfowler-datadog: In general, I was also hoping to keep as much of the privacy logic in a single place as possible, to make it easier to understand.

packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
sethfowler-datadog: FWIW we still use numbers, in the end, but the numbers are indexes into the string table. 🙂

We are actually mostly aligned with the spec here, but it's the spec for [nodeName](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName) that we're aligned with, not `nodeType`. (We don't perfectly match it because there are a few cases where we diverge to do something a bit nicer -- e.g. we handle `DocumentType` nodes differently, since spec'd `nodeName` would yield `html` for `<!DOCTYPE html>`, which is not what we want.)

Why `nodeName` and not `nodeType`? `nodeName` is nice because it eliminates the need to store separate fields for the node type and the tag name of element nodes. Non-element nodes have `nodeName`s that start with `#` (excluding the oddballs like `DocumentType`), and because [element tag names cannot start with `#` in HTML](https://dom.spec.whatwg.org/#valid-element-local-name), this is never ambiguous. By eliminating the need for separate node type and tag name fields, each serialized node is a bit smaller.

packages/rum/src/domain/record/serialization/changeEncoder.ts
```diff
@@ -0,0 +1,93 @@
...
+          array[itemIndex] = stringIds.getOrInsert(item)
+          if (stringIds.size > previousSize) {
+            add(ChangeType.AddString, item)
+          }
+        } else if (Array.isArray(item)) {
+          arrays.push(item)
+        }
+      }
+    }
+  }
```
BenoitZugmeyer: **nitpick:** this function could be simpler if it was recursive... are you concerned about exhausting the stack or the performance impact?

packages/rum/src/domain/record/serialization/insertionCursor.ts
```diff
@@ -0,0 +1,94 @@
...
+
+/** Returns an InsertionCursor which starts positioned at the root of the document. */
+export function createRootInsertionCursor(nodeIds: NodeIds): InsertionCursor {
+  interface ChildListCursor {
+    parentId: NodeId | undefined
+    previousSiblingId: NodeId | undefined
+  }
+
+  let childList: ChildListCursor = { parentId: undefined, previousSiblingId: undefined }
+  const stack: ChildListCursor[] = [childList]
```
BenoitZugmeyer: **nitpick:** it could be slightly simpler to have a chained list:

```ts
  interface ChildListCursor {
    parent: ChildListCursor | undefined
    previousSiblingId: NodeId | undefined
  }

  let childList: ChildListCursor = { parent: undefined, previousSiblingId: undefined }
```

packages/rum/src/domain/record/serialization/serializationTransaction.ts
```diff
@@ -92,3 +110,131 @@ export function serializeInTransaction(
...
+    addMetric(metric: keyof SerializationStats, value: number): void {
+      updateSerializationStats(stats, metric, value)
+    },
+    addNode(pos, nodeName, ...params): void {
+      const change: AddNodeChange = [pos, nodeName]
+      for (const param of params) {
+        change.push(param as AddNodeParams<AddNodeChange>[number])
+      }
+      encoder.add(ChangeType.AddNode, change)
+    },
```
BenoitZugmeyer: **suggestion:** this should be equivalent, no?
```suggestion
    addNode(...change): void {
      encoder.add(ChangeType.AddNode, change)
    },
```

packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
```diff
@@ -0,0 +1,233 @@
...
+  // We only generate placeholders for element nodes; other hidden nodes are simply not
+  // serialized.
+  if (!isElementNode(node)) {
+    return
+  }
+
+  const { nodeId, insertionPoint } = cursor.advance(node)
+  transaction.addNode(insertionPoint, encodedElementName(node), [PRIVACY_ATTR_NAME, PRIVACY_ATTR_VALUE_HIDDEN])
+  const { width, height } = node.getBoundingClientRect()
+  transaction.setSize(nodeId, width, height)
```
BenoitZugmeyer, line 198: **question:** will `setSize` be used for something else? Else, why not just use style attributes like we did before?

packages/rum/src/domain/record/startFullSnapshots.spec.ts
```diff
@@ -107,4 +98,36 @@ describe('startFullSnapshots', () => {
...
+    })
+  })
+
+  describe('when generating BrowserChangeRecord', () => {
+    beforeAll(() => {
+      addExperimentalFeatures([ExperimentalFeature.USE_CHANGE_RECORDS])
+    })
+
+    afterAll(() => {
+      resetExperimentalFeatures()
```
BenoitZugmeyer: **suggestion:** you could use `mockExperimentalFeatures` so you don't have to reset them

packages/rum/src/domain/record/serialization/changeEncoder.ts
sethfowler-datadog: More so the latter, but TBH, I think it could be made recursive without much performance impact. The only thing I really wanted to avoid was unnecessarily copying all the arrays, and avoiding that just requires mutating the arrays in place; there's no reason that a recursive approach can't be used. I'll switch over to a recursive approach; it does indeed make things a bit simpler.

packages/rum/src/domain/record/serialization/insertionCursor.ts
sethfowler-datadog: Yeah, that could be simpler indeed! I'll give it a shot.

packages/rum/src/domain/record/serialization/serializationTransaction.ts
sethfowler-datadog: Indeed, that's much better! (Though it requires a type assertion to make TypeScript happy, but the current version does too.) I'll switch to that approach.

packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
sethfowler-datadog: > will setSize be used for something else?

`SizeChange` will eventually be used for recording the size of censored images as well.

> why not just use style attributes like we did before?

1. It's nice to clearly distinguish between content that was present in the real DOM at recording time, and content that we generate. (This is, admittedly, more of a philosophical stance than anything else.)
2. It's more compact to represent these sizes as a simple array with two integers than to represent them using CSS syntax.
3. It's possible for hidden nodes and censored images to have `style` attributes already, and we may find in the future that it's necessary to preserve style information on these nodes to produce the correct layout. (I've seen the need for that in the past.) If we find ourselves needing to support that, it'll be nice to not have to worry about merging our generated styles with customer styles.

packages/rum/src/domain/record/startFullSnapshots.spec.ts
sethfowler-datadog: Ah, true! I'll make that change.

packages/rum/src/domain/record/serialization/insertionCursor.ts
```diff
@@ -50,22 +50,26 @@ export interface InsertionCursor {
...
+    parentCursor: ChildListCursor | undefined
     parentId: NodeId | undefined
     previousSiblingId: NodeId | undefined
   }
 
-  let childList: ChildListCursor = { parentId: undefined, previousSiblingId: undefined }
-  const stack: ChildListCursor[] = [childList]
+  let cursor: ChildListCursor = {
+    parentCursor: undefined,
+    parentId: undefined,
```
BenoitZugmeyer: **nitpick:** `cursor.parentId` could be `cursor.parentCursor.previousSiblingId`

BenoitZugmeyer (approved):

packages/rum/src/domain/record/serialization/insertionCursor.ts
sethfowler-datadog: True! I was being a bit conservative about keeping that property, since for incremental mutations its initial value can be something other than `undefined`, but I think we can achieve the same effect just by adding a sentinel `parentCursor` to hold the value. I'll go ahead and make that change.

PR #4061, opened 2026-01-07
🐛 Fix developer extension crash due to missing build variable

rgaignault:
## Motivation

The developer extension crashes when clicking on the dots to "Copy as curl/fetch" due to a missing build variable after the WXT migration.

## Changes

Define `__BUILD_ENV__SDK_VERSION__` in the WXT/Vite config

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/types/sessionReplay.ts
+++ b/packages/rum/src/types/sessionReplay.ts
@@ -419,7 +419,7 @@ export type AddDocTypeNodeChange = [
   '#doctype' | StringReference,
   StringOrStringReference,
   StringOrStringReference,
-  StringOrStringReference
+  StringOrStringReference,
 ]
 /**
  * Browser-specific. Schema representing a string, either expressed as a literal or as an index into the string table.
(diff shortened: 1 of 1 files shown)


packages/core/src/domain/telemetry/telemetryEvent.types.ts
BenoitZugmeyer, line 1: **question:** Any reason to update the types here?

BeltranBulbarellaDD: ~~I think if not added here, the CI job check-schemas fails because it checks with the one on remote. I had a similar thing on my PR~~
What I said is not correct.

rgaignault: No particular reason, I just sync the schema after the job was failing

BenoitZugmeyer: The job wouldn't have failed if you didn't update the `rum-event-format` submodule. Next time, please keep your PR focused and avoid updating the event formats if it's not necessary.

PR #4066, opened 2026-01-09
🐛 fix reading cookies containing special characters

BenoitZugmeyer:
## Motivation

With remote configuration, users can specify any cookie name. Our `getCookie` function does not handle cookie name with some special characters.

## Changes

Relax the regex to accept any character for the name.

## Test instructions

* Define a remote configuration that extracts the user id from a cookie named `foo.bar`
* Configure the sandbox with this remote config id
* Define the cookie `cookieStore.set('foo.bar', 'toto')`
* Verify that the user id is correctly set to 'toto'

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/tools/utils/stringUtils.ts
+++ b/packages/core/src/tools/utils/stringUtils.ts
@@ -12,7 +12,7 @@ export function generateUUID(placeholder?: string): string {
 // Assuming input string is following the HTTP Cookie format defined in
 // https://www.ietf.org/rfc/rfc2616.txt and https://www.ietf.org/rfc/rfc6265.txt, we don't need to
 // be too strict with this regex.
-const COMMA_SEPARATED_KEY_VALUE = /(\S+?)\s*=\s*(.+?)(;|$)/g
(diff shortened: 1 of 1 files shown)


packages/core/src/tools/utils/stringUtils.ts
```diff
@@ -9,7 +9,10 @@ export function generateUUID(placeholder?: string): string {
     : `${1e7}-${1e3}-${4e3}-${8e3}-${1e11}`.replace(/[018]/g, generateUUID)
 }
 
-const COMMA_SEPARATED_KEY_VALUE = /([\w-]+)\s*=\s*([^;]+)/g
+// Assuming input string is following the HTTP Cookie format defined in
+// https://www.ietf.org/rfc/rfc2616.txt and https://www.ietf.org/rfc/rfc6265.txt, we don't need to
+// be too strict with this regex.
+const COMMA_SEPARATED_KEY_VALUE = /(\S+?)\s*=\s*(.+?)(;|$)/g
```
mormubis:
```suggestion
const COMMA_SEPARATED_KEY_VALUE = /(\S+?)\s*=\s*(.+?)(?:;|$)/g
```

☝️ This way we don't capture unnecessary groups.

PR #4067, opened 2026-01-09
feat: async session manager

mormubis:
## Motivation

The Session Manager Revamp (SMR) will require our session manager to operate asynchronously and postpone the SDK initialization until the session is ready.

## Changes

`onReady` callback in session manager that propagates through all places using the start of the session.

## Test instructions

TBD

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/types/sessionReplay.ts
+++ b/packages/rum/src/types/sessionReplay.ts
@@ -419,7 +419,7 @@ export type AddDocTypeNodeChange = [
   '#doctype' | StringReference,
   StringOrStringReference,
   StringOrStringReference,
-  StringOrStringReference
+  StringOrStringReference,
 ]
 /**
  * Browser-specific. Schema representing a string, either expressed as a literal or as an index into the string table.
@@ -560,7 +560,7 @@ export type VisualViewportChange = [
   VisualViewportPageTop,
   VisualViewportWidth,
   VisualViewportHeight,
-  VisualViewportScale
+  VisualViewportScale,
 ]
 /**
  * The offset of the left edge of the visual viewport from the left edge of the layout viewport in CSS pixels.


packages/core/test/wait.ts
```diff
@@ -7,3 +7,47 @@ export function wait(durationMs: number = 0): Promise<void> {
...
+        const result = callback()
+        if (result && typeof (result as any).then === 'function') {
+          ;(result as Promise<T>).then(handleResult, handleError)
+        } else {
+          handleResult(result as T)
+        }
+      } catch (error) {
+        handleError(error as Error)
+      }
+    }
```
BenoitZugmeyer, line 31: **suggestion:** you can simplify all this as:

```ts
function check() {
  Promise.resolve().then(callback).then(handleResult, handleError)
}
```

And sometimes in the future:

```ts
function check() {
  Promise.try(callback).then(handleResult, handleError)
}
```

packages/logs/src/boot/logsPublicApi.spec.ts
```diff
@@ -68,15 +73,16 @@ describe('logs entry', () => {
   describe('common context', () => {
     let LOGS: LogsPublicApi
 
-    beforeEach(() => {
+    beforeEach(async () => {
       LOGS = makeLogsPublicApi(startLogs)
       LOGS.init(DEFAULT_INIT_CONFIGURATION)
+      await waitFor(() => startLogs.calls.count() > 0)
```
BenoitZugmeyer, line 79: **question:** not sure to understand why you put `waitFor` in many places. When I remove those `waitFor`, tests are still passing?

Also **suggestion:** use `collectAsyncCalls` instead of `waitFor` (it'll be more efficient)

```ts
await collectAsyncCalls(startLogs, 1)
```

packages/core/src/domain/session/sessionStoreOperations.ts
```diff
@@ -58,7 +67,7 @@ export function processSessionStoreOperations(
     }
     // acquire lock
     currentLock = createLock()
-    persistWithLock(currentStore.session)
+    safePersist(() => persistWithLock(currentStore.session))
```
BenoitZugmeyer, line 70: **question:** Why do we need this change?

packages/logs/src/boot/logsPublicApi.spec.ts
mormubis: It was super flaky for me. Most of the time, it varies between 3 or 25 tests failing. So I chose to add `waitFor` everywhere.

packages/core/src/domain/session/sessionStoreOperations.ts
mormubis: I noticed that the session manager inits before the tryInit, so it outside of the monitor call.

PR #4071, opened 2026-01-09
✨ `localStorage` strategy for remote configuration

mormubis:
## Motivation

Remote configuration lets customers tweak their Browser SDK settings (sample rates, privacy levels, tracing URLs, user context, etc.) from the Datadog UI without redeploying their app. At init, the SDK fetches a JSON config and applies it on top of the `init()` options.

Dynamic options can resolve values at runtime from different sources using a strategy-based approach — JavaScript variables (`js`), cookies (`cookie`), or DOM elements (`dom`). But `localStorage` wasn't supported, even though it's a super common way for apps to store metadata like version strings or user IDs.

image

## Changes

Adds `localStorage` as a new strategy for remote configuration dynamic options.

- New `localStorage` strategy in the remote config types and schema
- `resolveLocalStorageValue()` with error handling for when `localStorage` access throws
- Telemetry metrics for `localStorage` resolution (`success`, `missing`, `failure`)
- Supports optional regex `extractor` (same as other strategies)

## Test instructions

No backend ready for it yet.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
+++ b/packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
@@ -245,7 +245,7 @@ describe('remoteConfiguration', () => {
         expect(metrics.get().cookie).toEqual({ success: 1 })
       })
 
-      it('should not be set if the cookie is missing', () => {
+      it('should resolve to undefined if the cookie is missing', () => {
         expectAppliedRemoteConfigurationToBe(
           { version: { rcSerializedType: 'dynamic', strategy: 'cookie', name: COOKIE_NAME } },
           { version: undefined }


packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -91,7 +92,12 @@ export function applyRemoteConfiguration(
   const appliedConfiguration = { ...initConfiguration } as RumInitConfiguration & { [key: string]: unknown }
   SUPPORTED_FIELDS.forEach((option: string) => {
     if (option in rumRemoteConfiguration) {
-      appliedConfiguration[option] = resolveConfigurationProperty(rumRemoteConfiguration[option])
+      const resolvedValue = resolveConfigurationProperty(rumRemoteConfiguration[option])
+      // Skip undefined values to preserve init config fallbacks when any dynamic strategy
+      // (cookie, dom, js, localStorage) fails to resolve.
```
bcaudan: ❓ question: ‏do we have a property for which we want to have this behavior? I may miss something but for now I don't think we can configure a property in init that we can override by a dynamic config.
If we can't really achieve that with current system, I'd be in favor of not adding this fallback and the related tests.

test/e2e/scenario/rum/remoteConfiguration.scenario.ts
```diff
@@ -93,6 +93,93 @@ test.describe('remote configuration', () => {
...
+        applicationId: 'e2e',
+        version: { rcSerializedType: 'dynamic', strategy: 'localStorage', key: 'non_existent_key' },
+      },
+    })
+    .run(async ({ page }) => {
+      const initConfiguration = await page.evaluate(() => window.DD_RUM!.getInitConfiguration()!)
+      expect(initConfiguration.version).toBe('fallback-version')
+    })
+
+  createTest('should handle localStorage access failure gracefully')
```
bcaudan, line 156: **👏 praise:** ‏nice catch 👍

packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
```diff
@@ -248,7 +248,7 @@ describe('remoteConfiguration', () => {
       it('should resolve to undefined if the cookie is missing', () => {
         expectAppliedRemoteConfigurationToBe(
           { version: { rcSerializedType: 'dynamic', strategy: 'cookie', name: COOKIE_NAME } },
-          { version: undefined }
+          {} // version should not be set when resolution fails
```
thomas-lebeau: **💬 suggestion:** ‏Seems that the title could be updated instead of adding a comment

```
it('should not be set if the cookie is missing', () => {
```

packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
```diff
@@ -503,12 +503,69 @@ describe('remoteConfiguration', () => {
...
+          applicationId: 'yyy',
+          version: { rcSerializedType: 'dynamic', strategy: 'localStorage', key: 'non_existent_key' },
+        }
+
+        expect(
+          applyRemoteConfiguration(initConfigWithVersion, rumRemoteConfiguration, supportedContextManagers, metrics)
+        ).toEqual({
+          ...initConfigWithVersion,
+          applicationId: 'yyy',
+          // version should remain 'init-version', not undefined
```
thomas-lebeau: **💬 suggestion:** ‏what about making that explicit then, even though it's already implied by `...initConfigWithVersion,`:

```suggestion
          version: 'init-version'
```

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
mormubis: I reverted this change

packages/rum-core/src/domain/configuration/remoteConfiguration.spec.ts
```diff
@@ -245,7 +245,7 @@ describe('remoteConfiguration', () => {
         expect(metrics.get().cookie).toEqual({ success: 1 })
       })
 
-      it('should resolve to undefined if the cookie is missing', () => {
+      it('should not be set if the cookie is missing', () => {
```
thomas-lebeau: **🥜 nitpick:** ‏I guess if you reverted the undefine things, this test name should be updated too:

```suggestion
      it('should resolve to undefined if the cookie is missing', () => {
```

PR #4077, opened 2026-01-12
📈 [PANA-5371] Add telemetry to help evaluate new DOM mutation encoding

sethfowler-datadog:
## Motivation

#4060 adds a new, experimental DOM mutation encoding that's intended to be more compact, reducing resource usage and improving reliability. To help confirm it behaves as expected, it'd be helpful to update our existing segment telemetry to include information about which encoding is being used. This will allow us to compare the performance of the two encodings.

Additionally, while implementing #4060 I noticed that our existing segment telemetry is not correctly tracking statistics about the size of CSS text; in particular, `adoptedStyleSheets` are not included in the metrics. To ensure that we are comparing apples to apples, I didn't include `adoptedStyleSheets` in the telemetry generated by the new serialization code added in #4060, either. We should fix this for both serialization formats, so that our metrics are correct.

## Changes

This PR:
* Adds a new `encoding` field to our segment telemetry, with subfields indicating the encoding being used for both full snapshots and incremental snapshots. Enabling `USE_CHANGE_RECORDS` currently changes the encoding only for full snapshots, and this is reflected in the telemetry; an upcoming PR will add support for encoding incremental snapshots as `Change` records, and I'll update the telemetry again at that time.
* Updates the code used for serializing `adoptedStyleSheets` in both serialization formats to track statistics correctly.
* Updates relevant tests.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
+++ b/packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
@@ -218,7 +218,7 @@ function serializeStyleSheetAsChange(sheet: CSSStyleSheet, transaction: ChangeSe
   const mediaList = sheet.media.length > 0 ? Array.from(sheet.media) : undefined
   transaction.addMetric(
     'cssText',
-    rules.map((rule) => rule.length).reduce((a, b) => a + b, 0)
+    rules.reduce((totalLength, rule) => totalLength + rule.length, 0)
   )
   transaction.addStyleSheet(rules, mediaList, sheet.disabled)
   return transaction.scope.styleSheetIds.getOrInsert(sheet)
--- a/packages/rum/src/domain/record/serialization/serializeStyleSheets.ts
+++ b/packages/rum/src/domain/record/serialization/serializeStyleSheets.ts
@@ -13,7 +13,7 @@ export function serializeStyleSheets(
     const cssRules = Array.from(rules, (cssRule) => cssRule.cssText)
     transaction.addMetric(
       'cssText',
-      cssRules.map((rule) => rule.length).reduce((a, b) => a + b, 0)
+      cssRules.reduce((totalLength, rule) => totalLength + rule.length, 0)
     )
 
     const styleSheet: StyleSheet = {


packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
```diff
@@ -216,6 +216,10 @@ function serializeStyleSheetsAsChange(
 function serializeStyleSheetAsChange(sheet: CSSStyleSheet, transaction: ChangeSerializationTransaction): StyleSheetId {
   const rules = Array.from(sheet.cssRules || sheet.rules, (rule) => rule.cssText)
   const mediaList = sheet.media.length > 0 ? Array.from(sheet.media) : undefined
+  transaction.addMetric(
+    'cssText',
+    rules.map((rule) => rule.length).reduce((a, b) => a + b, 0)
```
gonzalezreal: Would using a single reduce be more efficient to avoid creating an intermediate array?

```typescript
cssRules.reduce((sum, rule) => sum + rule.length, 0)
```

Or does the JS engine optimize this away?

packages/rum/src/domain/record/serialization/serializeStyleSheets.ts
```diff
@@ -1,12 +1,20 @@
...
+): StyleSheet[] | undefined {
   if (cssStyleSheets === undefined || cssStyleSheets.length === 0) {
     return undefined
   }
   return cssStyleSheets.map((cssStyleSheet) => {
     const rules = cssStyleSheet.cssRules || cssStyleSheet.rules
     const cssRules = Array.from(rules, (cssRule) => cssRule.cssText)
+    transaction.addMetric(
+      'cssText',
+      cssRules.map((rule) => rule.length).reduce((a, b) => a + b, 0)
```
gonzalezreal: Same here:

```typescript
cssRules.reduce((sum, rule) => sum + rule.length, 0)
```

packages/rum/src/domain/record/serialization/serializeNodeAsChange.ts
sethfowler-datadog: It might well be optimized away, but no sense in making life harder for the optimizer. 🙂  That's indeed better; not sure why I did it the other way. I'll update the PR.

PR #4091, opened 2026-01-15
Revert "🐛 [PANA-5375] Treat Change records as full snapshots when appropriate"

ahfuckit:
Reverts DataDog/browser-sdk#4078


sethfowler-datadog:
Thank you for your contribution. I think this change needs to be discussed before we move forward with it, though. Please start by filing an issue describing the problem you're having. For now, I'm going to close this PR.

PR #4097, opened 2026-01-16
⚗️ [RUM-13259] Add shadow DOM support for action selectors

rgaignault:
## Motivation

Second part of shadow DOM support for actions. This PR adds selector generation for elements inside shadow DOM, using a `/shadow/` marker to indicate shadow boundaries.
Follows up on the first PR that added action name support.

## Changes

- Generate selectors that cross shadow DOM boundaries with /shadow/ marker
- Added unit tests for shadow DOM selectors
- Added e2e tests for selector generation

These two pictures are coming from the same click with and without the parameter enabled. The selector is correct and more precise that's why the width and height are smaller. 

Before 

image

After

image

Selector query

image

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/test/e2e/scenario/rum/actions.scenario.ts
+++ b/test/e2e/scenario/rum/actions.scenario.ts
@@ -568,7 +568,7 @@ test.describe('action collection with shadow DOM', () => {
       const actionEvents = intakeRegistry.rumActionEvents
       expect(actionEvents).toHaveLength(1)
       expect(actionEvents[0].action?.target?.name).toBe('Shadow Button')
-      expect(actionEvents[0]._dd.action?.target?.selector).toEqual('#shadow-host::shadow >BUTTON')
+      expect(actionEvents[0]._dd.action?.target?.selector).toEqual('#shadow-host::shadow BUTTON')
     })
 
   createTest('with betaTrackActionsInShadowDom, traverse shadow boundary for data-dd-action-name')
@@ -624,7 +624,7 @@ test.describe('action collection with shadow DOM', () => {
       const actionEvents = intakeRegistry.rumActionEvents
       expect(actionEvents).toHaveLength(1)
       expect(actionEvents[0]._dd.action?.target?.selector).toEqual(
-        '#shadow-host::shadow >BUTTON[data-testid="shadow-btn"]'
+        '#shadow-host::shadow BUTTON[data-testid="shadow-btn"]'
       )
     })
 })


packages/rum-core/src/domain/action/trackClickActions.spec.ts
```diff
@@ -657,6 +657,42 @@ describe('trackClickActions', () => {
...
+        activity: {},
+        eventProperty: {
+          composed: true,
+          composedPath: () => [shadowButton, shadowHost.shadowRoot, shadowHost, document.body, document],
+        },
+      })
+      clock.tick(EXPIRE_DELAY)
+
+      expect(events.length).toBe(1)
+      expect(events[0].target?.selector).toContain('/shadow/')
```
BeltranBulbarellaDD: Would it be better to use `SHADOW_DOM_MARKER` instead of `/shadow/` here? In case it changes in the future?

BeltranBulbarellaDD (approved):
Nice! Works perfectly with the [html test](https://gist.github.com/BeltranBulbarellaDD/bb99823931de43f423f2b11457ba9351) 
Looks good in the [explorer UI](https://dd.datad0g.com/rum/sessions?query=%40type%3Aaction%20%40application.id%3Aa81f40b8-e9bd-4805-9b66-4e4edc529a14&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZvVfeJ1-HfTiwAAABhBWnZWZmtfZkFBQTZSYmxWMTFxS3pRQUUAAAAkMTE5YmQ1ODItN2E1Mi00NjhlLTk5YmYtOTgzMDQwNDQ0MDYyAAFOSQ&fromUser=false&viz=stream&from_ts=1768727602898&to_ts=1768814002898&live=true) too

packages/rum-core/src/domain/getSelectorFromElement.ts
BenoitZugmeyer, line 1: **suggestion:** Implementation is a bit hard to follow. Maybe it could be clearer to have:
* a function that returns a CSS selector to a `target` within a `rootNode` without worrying about shadow dom
* a function that returns a list of all the "root nodes" ancestors from an element
* using the two functions to assemble a selector:

```ts
export function getSelectorFromElement(
  targetElement: Element,
  actionNameAttribute: string | undefined
): string | undefined {
  return getAllRootNodes(targetElement)
    .map((rootNode) => getSelectorFromElementWithinSubtree(rootNode, targetElement, actionNameAttribute))
    .join(" /shadow/ ")
}

function getAllRootNodes(node: Node) {
  const result = []
  while (node) {
    const rootNode = node.getRootNode()
    result.unshift(rootNode)
    if (rootNode instanceof ShadowRoot) {
      node = rootNode.host
    } else {
      break
    }
  }
  return result
}

function getSelectorFromElementWithinSubtree(rootNode: Node, targetElement: Element, actionNameAttribute: string | undefined): string {
  // ... more or less what `getSelectorFromElement` does currently, but scoped to `rootNode`
}
```

packages/rum-core/src/domain/getSelectorFromElement.spec.ts
```diff
@@ -220,3 +220,144 @@ describe('isSelectorUniqueAmongSiblings', () => {
     expect(isSelectorUniqueAmongSiblings(element, 'DIV', 'HR')).toBeTrue()
   })
 })
+
+describe('getSelectorFromElement with shadow DOM', () => {
```
BenoitZugmeyer, line 224: **suggestion:** add a test where two elements, one inside and the other outside the shadow dom, are sharing the same id attribute:

```ts
  it('should handle duplicated ids', () => {
    const container = appendElement('<div target></div><div id="foo"></div>')
    const shadow = container.attachShadow({ mode: 'open' })

    const button = document.createElement('button')
    button.id = 'foo'
    shadow.appendChild(button)

    const selector = getSelectorFromElement(button, undefined)
    expect(selector).toBe('...')
  })
```

This currently yields an unexpected selector.

packages/rum-core/src/domain/getSelectorFromElement.spec.ts
```diff
@@ -220,3 +220,144 @@ describe('isSelectorUniqueAmongSiblings', () => {
...
+    const host = appendElement('<div id="shadow-host"></div>')
+    const shadowRoot = host.attachShadow({ mode: 'open' })
+    const button = document.createElement('button')
+    button.classList.add('shadow-button')
+    shadowRoot.appendChild(button)
+
+    const selector = getSelectorFromElement(button, undefined)
+    expect(selector).toBeDefined()
+    expect(selector).toContain(SHADOW_DOM_MARKER)
+    expect(selector).toContain('BUTTON')
```
BenoitZugmeyer: **suggestion:** it is quite hard to understand what's the expected selector by just using `toContain` assertions. Replace every `toContain` with `toBe` (in all new test cases):

```suggestion
    expect(selector).toBe(`#shadow-host${SHADOW_DOM_MARKER}BUTTON.shadow-button`)
```

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -46,35 +75,67 @@ export function getSelectorFromElement(
...
 
-    currentElement = currentElement.parentElement
+    const elementSelector = uniqueSelectorAmongChildren || getPositionSelector(currentElement)
+    currentSelector = combineSelector(elementSelector, currentSelector)
+
+    // Stop when we've reached a direct child of the rootNode
+    const parent: Element | null = currentElement.parentElement
+    if (!parent || currentElement.parentNode === rootNode) {
+      break
+    }
```
BenoitZugmeyer: **suggestion:** this condition should be merged/adjusted in the `while(...)` statement. Something like:

```ts
while(currentElement && currentElement !== rootNode)
```

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -149,10 +212,32 @@ function getPositionSelector(element: Element): string {
...
+  if (element.parentElement) {
+    return element.parentElement
+  }
+  const parentNode = element.parentNode
+  if (parentNode && isNodeShadowRoot(parentNode)) {
+    return parentNode
+  }
+  return null
+}
+
```
BenoitZugmeyer: **suggestion:** Do we need this? Using `element.parentNode` instead of this functions does not break the tests.

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -37,6 +40,32 @@ const UNIQUE_AMONG_CHILDREN_SELECTOR_GETTERS: SelectorGetter[] = [
...
 
+interface SubtreeTarget {
+  rootNode: Document | ShadowRoot
+  target: Element
+}
+
+/**
+ * Returns all (rootNode, target) pairs from the document down to the element.
+ */
+function getAllSubtreeTargets(element: Element): SubtreeTarget[] {
```
BenoitZugmeyer: **nitpick:** move this function below the `export function` declaration (code is easier to read when public/exported functions are at the top of the file)

packages/rum-core/src/domain/getSelectorFromElement.ts
rgaignault: Oh yeah true we don't need it anymore 👍

packages/rum-core/src/domain/getSelectorFromElement.ts
rgaignault: I am not sure to have fully understand your comment, do you mean it's already handled in the while ? I made some changes

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -1,5 +1,8 @@
+import { isNodeShadowRoot } from '../browser/htmlDomUtils'
 import { DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE } from './action/actionNameConstants'
 
+export const SHADOW_DOM_MARKER = ' ::shadow '
```
sethfowler-datadog: Can we use this instead, to make this match CSS syntax for pseudo-elements better?

```suggestion
export const SHADOW_DOM_MARKER = '::shadow'
```

If you are worried that this might also match some other hypothetical pseudo-element that starts with `::shadow`, like `::shadow-root`, you can split selectors using the regular expression `/::shadow[^a-z-]/`, instead of just the string `::shadow`, but in truth we probably don't need to worry about this for two reasons:
* No standardized pseudo-element currently exists that starts with `::shadow`.
* We are generating these selectors anyway, so even if one were added, we wouldn't generate suddenly start generating selectors that use it; there'd be no risk of collision until and unless we adopted the new pseudo-element.

rgaignault: Thanks your comment, make sense indeed the risk is not important. I don't think we need a regular expression, let me update the code 👍

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -1,5 +1,8 @@
+import { isNodeShadowRoot } from '../browser/htmlDomUtils'
 import { DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE } from './action/actionNameConstants'
 
+export const SHADOW_DOM_MARKER = '::shadow'
```
amortemousque: **💬 suggestion:** ‏Could be nice to add a comment explaining the reasoning behind choosing this marker, like this is not a supported CSS selector etc.

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -46,35 +61,82 @@ export function getSelectorFromElement(
...
+/**
+ * Returns all (rootNode, target) pairs from the document down to the element.
+ */
+function getAllSubtreeTargets(element: Element): SubtreeTarget[] {
+  const result: SubtreeTarget[] = []
+  let currentTarget: Element | undefined = element
+
+  while (currentTarget) {
+    const rootNode = currentTarget.getRootNode() as Document | ShadowRoot
+    result.unshift({ rootNode, target: currentTarget })
```
sethfowler-datadog: Nit: there's an opportunity to improve computational complexity here a bit.

Every time you call `unshift()`, every element that's already in the array needs to be copied to a new position to make space for the element you're unshifting. A single call to `unshift()` is thus O(n), which makes a loop like this O(n^2). A better approach is to use `push()`, which is amortized O(1), resulting in an amortized O(n) loop, and then `reverse()` the array, another O(n) operation. By using `push()` and `reverse()`, the overall task is O(n) instead of O(n^2).

Given that `result` is going to be very short in most cases, I wouldn't expect a substantial performance difference from this change, but it's good to develop the habit of using this approach, since the impact can be surprisingly large in cases where the array is larger!

packages/rum-core/src/domain/getSelectorFromElement.spec.ts
```diff
@@ -217,6 +217,143 @@ describe('isSelectorUniqueAmongSiblings', () => {
...
+
+  it('should handle duplicated IDs between light DOM and shadow DOM', () => {
+    const host = appendElement('<div target></div><div id="foo"></div>')
+    const shadowRoot = host.attachShadow({ mode: 'open' })
+    const button = document.createElement('button')
+    button.id = 'foo'
+    shadowRoot.appendChild(button)
+
+    const selector = getSelectorFromElement(button, undefined)
+    expect(selector).toBe(`BODY>DIV:nth-of-type(1)${SHADOW_DOM_MARKER}#foo`)
```
sethfowler-datadog: Just for clarity, is this the result we intended? The `#foo` inside and outside the shadow subtree are two different `#foo`s, and that's perfectly fine; I'd actually have expected the result here to be this:
```css
BODY>DIV#foo${SHADOW_DOM_MARKER}#foo
```

Using the id at both leaves results in a less fragile selector, so it's a better choice, unless there's some issue that's leading us to avoid it.

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -1,5 +1,15 @@
...
 import { DEFAULT_PROGRAMMATIC_ACTION_NAME_ATTRIBUTE } from './action/actionNameConstants'
 
+/**
+ * Marker used to indicate shadow DOM boundaries in selectors.
+ * Named after the deprecated Shadow DOM v0 '::shadow' pseudo-element which served the same
+ * purpose of marking entry into a shadow root.
+ * Note: This is NOT a valid CSS selector, it's an internal marker that requires custom
+ * parsing logic.
+ */
+export const SHADOW_DOM_MARKER = '::shadow'
```
sethfowler-datadog: Sorry for not realizing this earlier, but I think this should actually be `::shadow>`; otherwise the selectors you're generating would like like e.g. `::shadowBUTTON`, when what we really want is `::shadow>BUTTON`.

packages/rum-core/src/domain/getSelectorFromElement.spec.ts
rgaignault: If I understand correctly, the host should have had id="foo" instead of the sibling? I updated

packages/rum-core/src/domain/getSelectorFromElement.ts
rgaignault: Oh yeah, with the whitespace  removal we need to add this, good catch

test/e2e/scenario/rum/actions.scenario.ts
```diff
@@ -564,5 +564,63 @@ test.describe('action collection with shadow DOM', () => {
...
+    `)
+    .run(async ({ intakeRegistry, flushEvents, page }) => {
+      const button = page.locator('my-button').first().locator('button')
+      await button.click()
+      await flushEvents()
+
+      const actionEvents = intakeRegistry.rumActionEvents
+      expect(actionEvents).toHaveLength(1)
+      expect(actionEvents[0]._dd.action?.target?.selector).toContain('data-testid')
+      expect(actionEvents[0]._dd.action?.target?.selector).toContain('::shadow')
```
BenoitZugmeyer: **suggestion:** use the full selector with `.toEqual`, to make it more explicit

test/e2e/scenario/rum/actions.scenario.ts
```diff
@@ -564,5 +564,63 @@ test.describe('action collection with shadow DOM', () => {
       const actionEvents = intakeRegistry.rumActionEvents
       expect(actionEvents).toHaveLength(1)
       expect(actionEvents[0].action?.target?.name).toBe('Shadow Button')
+      expect(actionEvents[0]._dd.action?.target?.selector).toContain('BUTTON')
+      expect(actionEvents[0]._dd.action?.target?.selector).toContain('::shadow')
```
BenoitZugmeyer: **suggestion:** same here

PR #4104, opened 2026-01-19
⚗️ feat: collect LCP subparts RUM-13818

HugoGarrido:
## Motivation

LCP Subparts are values that compute the final LCP measure the browser SDK. 

This is helpful for troubleshooting LCP since it help to define what type of LCP issue we are facing (server, render etc.) by looking at the biggest contributor ([more details in this doc](https://datadoghq.atlassian.net/wiki/spaces/~7120206dda9a953467432992127e149ac3999c/pages/6089967368/Support+LCP+Subparts+in+RUM))

Here is a breakdown of each part :

Subpart | Meaning
-- | --
TTFB | Already tracked by the Browser SDK as firstByte
Resource load delay | Time between firstByte and the loading start of the resource associated with the LCP. 0 if no resource are attached to the LCP
Resource load duration | Time to takes to load the resource attached to the LCP0 if no resource are attached to the LCP
Element render delay | Time between the LCP resource finishes loading and the LCP element is fully rendered

This PR adds the collection of the LCP subparts when we collect the LCP behind the flag **`lcp_subparts`**.

## Changes

- In `trackLargestContentfulPaint` when we have an LCP entry : 
  - <del>re-measure the TTFB (keeping the name `firstByte`)</del>
  - get the performance entry related to the LCP resource if applicable
  - compute an optional `subParts` object that is added in the final payload
- Extract a helper to safely compute the `firstByte` value
- Extract a helper to safely read performance global
- Update tests, fixtures and types to support the new `subParts` object
- Add code behind `COLLECT_LCP_SUBPARTS`  

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
@@ -46,7 +46,9 @@ describe('trackLargestContentfulPaint', () => {
     ;({ notifyPerformanceEntries } = mockPerformanceObserver())
 
     // This ensures getNavigationEntry() returns controlled values for subParts calculation.
-    notifyPerformanceEntries([createPerformanceEntry(RumPerformanceEntryType.NAVIGATION, { responseStart: firstByte as RelativeTime })])
+    notifyPerformanceEntries([
+      createPerformanceEntry(RumPerformanceEntryType.NAVIGATION, { responseStart: firstByte as RelativeTime }),
+    ])
 
     // Add resource entries to the global performance buffer so findLcpResourceEntry can find them
     // notifyPerformanceEntries adds entries to the global buffer via mockGlobalPerformanceBuffer
@@ -235,9 +237,8 @@ describe('trackLargestContentfulPaint', () => {
         resourceUrl: undefined,
         subParts: undefined,
       })
-    });
-
-    [
+    })
+    ;[
       {
         description: 'should cap lcpResponseEnd at LCP time for resources that complete after LCP',
         firstByte: 200,
@@ -372,11 +373,13 @@ describe('trackLargestContentfulPaint', () => {
 
         const result = lcpCallback.calls.mostRecent().args[0]
 
-        expect(result.subParts).toEqual(expectedSubParts as {
-          loadDelay: RelativeTime
-          loadTime: RelativeTime
-          renderDelay: RelativeTime
-        })
+        expect(result.subParts).toEqual(
+          expectedSubParts as {
+            loadDelay: RelativeTime
+            loadTime: RelativeTime
+            renderDelay: RelativeTime
+          }
+        )
 
         // Validate: firstByte + loadDelay + loadTime + renderDelay = LCP value
         const sum = Object.values(result.subParts!).reduce((acc, curr) => acc + curr, 0)


packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -64,10 +71,33 @@ export function trackLargestContentfulPaint(
...
+
+      const lcpResourceEntry = resourceUrl
+        ? (getResourceEntries()?.find((e) => e.name === resourceUrl) as PerformanceResourceTiming | undefined)
+        : undefined
+
+      let subParts;
+      const firstByte = getSafeFirstByte(getNavigationEntry())
+
+      if (firstByte !== undefined) {
+        const lcpRequestStart = Math.max(firstByte, lcpResourceEntry?.startTime || 0)
```
jcdesousa: We should use requestStart when available, as it’s more accurate and aligns with browser dev tools. I observed a rounding issue and a 1 ms mismatch caused by the ~0.5 ms offset introduced when using start time instead of request time.

```suggestion
        const lcpRequestStart = Math.max(firstByte, lcpResourceEntry?.requestStart || lcpResourceEntry?.startTime || 0 || 0)
```

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -204,6 +204,11 @@ export interface ViewPerformanceData {
     timestamp: ServerDuration
     target_selector?: string
     resource_url?: string
+    sub_parts?: {
+      load_delay: ServerDuration
+      load_time: ServerDuration
+      render_delay: ServerDuration
+    }
```
bcaudan, line 211: **💬 suggestion:** ‏You should run `yarn json-schemas:sync` to also get the new types from rum-events-format

packages/rum-core/src/domain/view/viewCollection.ts
```diff
@@ -198,6 +198,13 @@ function computeViewPerformanceData(
       timestamp: toServerDuration(largestContentfulPaint.value),
       target_selector: largestContentfulPaint.targetSelector,
       resource_url: largestContentfulPaint.resourceUrl,
+      sub_parts: largestContentfulPaint?.subParts
```
bcaudan: **💬 suggestion:** it does not seem needed‏

```suggestion
      sub_parts: largestContentfulPaint.subParts
```

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -429,6 +431,9 @@ describe('view metrics', () => {
   let notifyPerformanceEntries: (entries: RumPerformanceEntry[]) => void
 
   beforeEach(() => {
+    addExperimentalFeatures([ExperimentalFeature.COLLECT_LCP_SUBPARTS])
+    registerCleanupTask(resetExperimentalFeatures)
```
bcaudan: **🥜 nitpick:** ‏we could group cleanup tasks like in other tests

packages/rum-core/src/domain/view/trackViews.spec.ts
```diff
@@ -588,7 +593,16 @@ describe('view metrics', () => {
...
             navigationTimings: jasmine.any(Object),
-            largestContentfulPaint: { value: 789 as Duration, targetSelector: undefined, resourceUrl: undefined },
+            largestContentfulPaint: {
+              value: 789 as Duration,
+              targetSelector: undefined,
+              resourceUrl: undefined,
+              subParts: {
+                loadDelay: 0,
+                loadTime: 0,
+                renderDelay: 666,
```
bcaudan, line 603: Nice value 😅

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -64,10 +77,40 @@ export function trackLargestContentfulPaint(
...
+      const lcpResourceEntry = resourceUrl
+        ? (getResourceEntries()?.find((e) => e.name === resourceUrl) as PerformanceResourceTiming | undefined)
+        : undefined
+
+      const lcpValue = Math.max(0, lcpEntry.startTime) as RelativeTime
+
+      let subParts
+      // TODO: Remove after testing
+      // eslint-disable-next-line no-constant-condition, no-constant-binary-expression
+      if (true || isExperimentalFeatureEnabled(ExperimentalFeature.COLLECT_LCP_SUBPARTS)) {
```
bcaudan: **❗ warning:** ‏ ‏Probably worth removing it and enabling the feature flag on staging

```suggestion
      if (isExperimentalFeatureEnabled(ExperimentalFeature.COLLECT_LCP_SUBPARTS)) {
```

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -64,10 +77,40 @@ export function trackLargestContentfulPaint(
...
+          const lcpRequestStart = getLcpResourceRequestStart(lcpResourceEntry, firstByte)
+
+          // Cap at LCP time to handle resources that continue downloading after LCP (e.g., videos)
+          const lcpResponseEnd = Math.min(lcpValue, Math.max(lcpRequestStart, lcpResourceEntry?.responseEnd || 0))
+
+          subParts = {
+            loadDelay: (lcpRequestStart - firstByte) as RelativeTime,
+            loadTime: (lcpResponseEnd - lcpRequestStart) as RelativeTime,
+            renderDelay: (lcpValue - lcpResponseEnd) as RelativeTime,
+          }
```
bcaudan: **💬 suggestion:** code organisation‏: what about extracting a `computeLcpSubParts` function?

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -64,10 +77,40 @@ export function trackLargestContentfulPaint(
         lcpTargetSelector = getSelectorFromElement(lcpEntry.element, configuration.actionNameAttribute)
       }
 
+      const resourceUrl = computeLcpEntryUrl(lcpEntry)
+
+      const lcpResourceEntry = resourceUrl
+        ? (getResourceEntries()?.find((e) => e.name === resourceUrl) as PerformanceResourceTiming | undefined)
+        : undefined
+
+      const lcpValue = Math.max(0, lcpEntry.startTime) as RelativeTime
```
bcaudan: **❓ question:** ‏scope: Any reason to add this max with this change?

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -64,10 +77,40 @@ export function trackLargestContentfulPaint(
         lcpTargetSelector = getSelectorFromElement(lcpEntry.element, configuration.actionNameAttribute)
       }
 
+      const resourceUrl = computeLcpEntryUrl(lcpEntry)
+
+      const lcpResourceEntry = resourceUrl
+        ? (getResourceEntries()?.find((e) => e.name === resourceUrl) as PerformanceResourceTiming | undefined)
+        : undefined
```
bcaudan: **💬 suggestion:** ‏code organisation: this should be in the subparts computation section

packages/rum-core/src/browser/performanceUtils.ts
```diff
@@ -48,3 +49,19 @@ export function computeTimingsFromDeprecatedPerformanceTiming() {
   }
   return result as TimingsFromDeprecatedPerformanceTiming
 }
+
+export function getSafeFirstByte(entry: RelevantNavigationTiming) {
```
bcaudan: **💬 suggestion:** ‏naming: as it is not a getter, `sanitizeFirstByte` could clarify the intent

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -64,10 +71,27 @@ export function trackLargestContentfulPaint(
...
```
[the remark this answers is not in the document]
bcaudan: codex comment seems pretty relevant, no?

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
bcaudan: **💬 suggestion:** ‏optimisation: no need to do this search if we don't have `firstByte`

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -85,3 +128,19 @@ export function trackLargestContentfulPaint(
...
+function getLcpResourceRequestStart(
+  lcpResourceEntry: PerformanceResourceTiming | undefined,
+  firstByte: RelativeTime
+): RelativeTime {
+  if (!lcpResourceEntry) {
+    return firstByte
+  }
+  const requestStart = Math.max(firstByte, lcpResourceEntry.requestStart || lcpResourceEntry.startTime)
+  return requestStart as RelativeTime
+}
```
bcaudan: **💬 suggestion:** ‏code organisation: after extracting a dedicated function for sub parts, we could inline this function to ease the comparison with the [web-vitals implmentation](https://github.com/GoogleChrome/web-vitals/blob/main/src/attribution/onLCP.ts#L98) (without prerender concerns for now)

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -13,14 +21,28 @@ import type { LargestContentfulPaint } from './trackLargestContentfulPaint'
...
 
+const mockPerformanceEntry = createPerformanceEntry(RumPerformanceEntryType.NAVIGATION, {
+  responseStart: 789 as RelativeTime,
+})
+
+const mockSubParts = {
+  loadDelay: 0 as RelativeTime,
+  loadTime: 0 as RelativeTime,
+  renderDelay: 0 as RelativeTime,
+}
```
bcaudan: **💬 suggestion:** we tend to init the mock for each test to avoid any side effects between tests, doing that inside `startLCPTracking` would seem appropriate

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +194,498 @@ describe('trackLargestContentfulPaint', () => {
       value: 789 as RelativeTime,
       targetSelector: undefined,
       resourceUrl: undefined,
+      subParts: mockSubParts,
+    })
+  })
+
+  it('should not provide subParts if the first byte is not available', () => {
```
bcaudan: **💬 suggestion:** test organization: this one probably belongs to the the LCP subparts group

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +194,498 @@ describe('trackLargestContentfulPaint', () => {
...
+      mockRumConfiguration(),
+      firstHidden,
+      eventTarget,
+      lcpCallback
+    )
+
+    registerCleanupTask(() => {
+      firstHidden.stop()
+      largestContentfulPaint.stop()
+    })
```
bcaudan: **💬 suggestion:** code duplication / readability: ‏can't we tweak `startLCPTraking` to avoid duplicating the setup in the added tests?
it would be nice to factorize some things to reduce the size of each test and easily grasp what are the inputs / expected outputs in each case.

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +194,498 @@ describe('trackLargestContentfulPaint', () => {
...
+
+    expect(lcpCallback).toHaveBeenCalledOnceWith({
+      value: 789 as RelativeTime,
+      targetSelector: undefined,
+      resourceUrl: undefined,
+      subParts: undefined,
+    })
+  })
+
+  describe('LCP subParts capping at LCP time', () => {
```
bcaudan: **💬 suggestion:** code organisation: ‏what about having those groups under LCP subparts group as well?

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
HugoGarrido: Yes definitely I wanted to speed up the test on staging 🙈, but it is removed already on my local branch

packages/rum-core/src/browser/performanceUtils.spec.ts
```diff
@@ -51,3 +53,168 @@ describe('getNavigationEntry', () => {
...
+      // Cleanup is handled by mockGlobalPerformanceBuffer's spies being restored
+    })
+    return mock
+  }
+
+  function skipIfResourceTimingNotSupported() {
+    if (!supportPerformanceTimingEvent(RumPerformanceEntryType.RESOURCE)) {
+      pending('Resource Timing Event is not supported in this browser')
+    }
+  }
```
bcaudan: **💬 suggestion:** ‏let's put that directly in a `beforeEach` to avoid scatter it across the tests

```suggestion
  beforeEach(() => {
    if (!supportPerformanceTimingEvent(RumPerformanceEntryType.RESOURCE)) {
      pending('Resource Timing Event is not supported in this browser')
    }
  })
```

packages/rum-core/src/browser/performanceUtils.spec.ts
```diff
@@ -51,3 +53,168 @@ describe('getNavigationEntry', () => {
...
 })
+
+describe('findLcpResourceEntry', () => {
+  function setupResourceEntries(entries: PerformanceEntry[]) {
+    const mock = mockGlobalPerformanceBuffer(entries)
+    registerCleanupTask(() => {
+      // Cleanup is handled by mockGlobalPerformanceBuffer's spies being restored
+    })
+    return mock
+  }
```
bcaudan: **💬 suggestion:** ‏not sure to get the intent here, what about :
- using `mockGlobalPerformanceBuffer` directly
- not register empty cleanup tasks

packages/rum-core/src/browser/performanceUtils.spec.ts
```diff
@@ -51,3 +53,168 @@ describe('getNavigationEntry', () => {
...
+        name: 'https://example.com/image.jpg',
+        startTime: 1500 as RelativeTime,
+      }),
+    ])
+
+    const result = findLcpResourceEntry('https://example.com/image.jpg', 1000 as RelativeTime)
+
+    expect(result).toBeDefined()
+    expect(result!.startTime).toBe(400 as RelativeTime)
+  })
```
bcaudan: **💬 suggestion:** ‏that seems to be already tested with the two previous tests, unless there is a specific case covered by this one, let's not duplicate test cases

packages/rum-core/src/browser/performanceUtils.spec.ts
```diff
@@ -51,3 +53,168 @@ describe('getNavigationEntry', () => {
...
+        name: 'https://example.com/image.jpg',
+        startTime: 1000 as RelativeTime,
+      }),
+    ])
+
+    const result = findLcpResourceEntry('https://example.com/image.jpg', 1000 as RelativeTime)
+
+    expect(result).toBeDefined()
+    expect(result!.startTime).toBe(1000 as RelativeTime)
+  })
```
bcaudan: **💬 suggestion:** seems to be a duplicate of `'should return the matching resource entry'`‏

packages/rum-core/src/browser/performanceUtils.spec.ts
```diff
@@ -51,3 +53,168 @@ describe('getNavigationEntry', () => {
...
+        startTime: 200 as RelativeTime,
+      }),
+    ])
+
+    const result = findLcpResourceEntry('https://example.com/image.jpg', 1000 as RelativeTime)
+
+    expect(result).toBeDefined()
+    expect(result!.startTime).toBe(200 as RelativeTime)
+    expect(result!.name).toBe('https://example.com/image.jpg')
+  })
```
bcaudan: **💬 suggestion:** ‏seems to be a duplicate of `'should return undefined when no matching URL is found'`

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.ts
```diff
@@ -85,3 +106,33 @@ export function trackLargestContentfulPaint(
...
+/**
+ * Compute the LCP sub-parts breakdown (loadDelay, loadTime, renderDelay).
+ * Returns undefined if navigation timing data or TTFB is unavailable.
+ */
+function computeLcpSubParts(
+  resourceUrl: string | undefined,
+  lcpValue: RelativeTime
+): LargestContentfulPaint['subParts'] {
+  const navigationEntry = getNavigationEntry()
+  const firstByte = sanitizeFirstByte(navigationEntry)
```
bcaudan: **🥜 nitpick:** ‏as we don't use `navigationEntry` elsewhere and it does not provide additional info, we could inline this variable

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -13,14 +21,149 @@ import type { LargestContentfulPaint } from './trackLargestContentfulPaint'
...
+  description: string
+  responseStart: RelativeTime
+  resource: ResourceEntryOptions
+  lcpTime: RelativeTime
+  expectedSubParts: {
+    loadDelay: RelativeTime
+    loadTime: RelativeTime
+    renderDelay: RelativeTime
+  }
+}
```
bcaudan: **💬 suggestion:** ‏`responseStart` seems a bit confusing as it is not obvious which response we are configuring.
What about using something like `firstByte` instead?

```suggestion
interface StartLCPTrackingOptions {
  firstByte?: RelativeTime
  resources?: ResourceEntryOptions[]
}

interface SubPartsTestCase {
  description: string
  firstByte: RelativeTime
  resource: ResourceEntryOptions
  lcpTime: RelativeTime
  expectedSubParts: {
    loadDelay: RelativeTime
    loadTime: RelativeTime
    renderDelay: RelativeTime
  }
}
```

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -13,14 +21,149 @@ import type { LargestContentfulPaint } from './trackLargestContentfulPaint'
...
+const subPartsTestCases: SubPartsTestCase[] = [
+  {
+    description: 'should cap lcpResponseEnd at LCP time for resources that complete after LCP',
+    responseStart: 200 as RelativeTime,
+    resource: {
+      name: 'https://example.com/video.mp4',
+      startTime: 500 as RelativeTime,
+      requestStart: 500 as RelativeTime,
+      responseEnd: 5000 as RelativeTime, // Continues downloading long after LCP
+    },
```
bcaudan: **🥜 nitpick:** ‏we could remove some noise by:

- not configuring the `resource.name` as it does not impact the test
- letting the the test code do the `RelativeTime` casts

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +321,130 @@ describe('trackLargestContentfulPaint', () => {
...
+    expect(lcpCallback).toHaveBeenCalledOnceWith({
+      value: 789 as RelativeTime,
+      targetSelector: undefined,
+      resourceUrl: undefined,
+      subParts: undefined,
+    })
+  })
+
+  describe('LCP subParts', () => {
+    subPartsTestCases.forEach(({ description, responseStart, resource, lcpTime, expectedSubParts }) => {
```
bcaudan: **🥜 nitpick:** ‏What about inlining `subPartsTestCases`? it would avoid some back and forth between the test cases and the implementation

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +321,130 @@ describe('trackLargestContentfulPaint', () => {
...
+      eventTarget,
+      lcpCallback
+    )
+
+    registerCleanupTask(() => {
+      firstHidden.stop()
+      largestContentfulPaint.stop()
+    })
+
+    notifyEntries([createPerformanceEntry(RumPerformanceEntryType.LARGEST_CONTENTFUL_PAINT)])
```
bcaudan: **💬 suggestion:** ‏can't we use `startLCPTracking({ responseStart: -1 })`

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +321,130 @@ describe('trackLargestContentfulPaint', () => {
...
+
+        expect(lcpCallback).toHaveBeenCalledOnceWith({
+          value: 1000 as RelativeTime,
+          targetSelector: undefined,
+          resourceUrl: undefined,
+          subParts: {
+            loadDelay: 0 as RelativeTime, // No resource, so max(firstByte, 0) - firstByte = 0
+            loadTime: 0 as RelativeTime, // No resource, so max(firstByte, 0) - firstByte = 0
+            renderDelay: 800 as RelativeTime, // 1000 - 200 (firstByte is used as lcpResponseEnd when no resource)
+          },
```
bcaudan: **💭 thought:** ‏I am wandering if it is the right data to collect when there is no associated resource.
- loadDelay / loadTime 0, it would skew the aggregation value of those metrics, maybe not collecting anything in this case would be better
- renderDelay: when there is no resource, what does it really represent? probably worth some explanation when displaying that in the UI

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +321,130 @@ describe('trackLargestContentfulPaint', () => {
...
+
+        expect(result.subParts).toEqual(expectedSubParts)
+
+        // Validate: firstByte + loadDelay + loadTime + renderDelay = LCP value
+        const sum = Object.values(result.subParts!).reduce((acc, curr) => acc + curr, 0)
+        expect(sum + responseStart).toBe(result.value)
+      })
+    })
+
+    describe('edge cases', () => {
```
bcaudan: **💬 suggestion:** ‏can't those edge cases be declared as `subPartsTestCases` as well?

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
HugoGarrido: I see your point but I tend to think that this is something that should be explained in the UI by stressing the presence or absence of resource to help understand the breakdown. 

The idea is to measure a set of standard fields, not having all the fields reported might give the impression that it's broken, no?

bcaudan: I don't think we need to answer those concerns right now, we could have a more informed opinion by dogfooding. However, it would be nice to have a strong opinion about it before opening this feature to customer.

packages/rum-core/src/domain/view/viewMetrics/trackLargestContentfulPaint.spec.ts
```diff
@@ -162,6 +217,170 @@ describe('trackLargestContentfulPaint', () => {
...
+      it(description, () => {
+        startLCPTracking({
+          firstByte,
+          resources: resource ? [resource] : [],
+        })
+
+        notifyPerformanceEntries([
+          createPerformanceEntry(RumPerformanceEntryType.LARGEST_CONTENTFUL_PAINT, {
+            startTime: lcpTime as RelativeTime,
+            url: lcpUrl === undefined ? 'https://example.com/image.jpg' : lcpUrl,
```
bcaudan: **💬 suggestion:** ‏as on other tests, we rely on the fact that this URL is the same than the one used for resources, it would be nice to extract a constant for it.

PR #4110, opened 2026-01-22
⚗ [RUM-11393] Add Start/StopResource API

BeltranBulbarellaDD:
## Motivation

Discussed in this [RFC](https://docs.google.com/document/d/1QnH_ZQoG_K7T-mzsF8-iJfM_TFX0E41WOWgJNm7-nBA/edit?tab=t.0#heading=h.rnd972k0hiye)

Following the same pattern as for start/stopAction, this PR introduces a manual resource tracking API.

Since start/stopAction and start/stopResource follow similar patterns, created `eventTracker` the idea is to reuse it for all events which have start/stop logic.

## Changes

**New Experimental Feature**
- Added `START_STOP_RESOURCE ` feature flag

**New Public APIs**
- Added startResource(url, options?) and stopResource(url, options?) to RumPublicApi

**Architecture Refactoring**

- Extracted `eventTracker`: Shared utility for managing start/stop lifecycle (used by both actions and resources)
- Refactored `trackManualActions.ts` to use the shared utility
- Added `trackManualResources.ts` for manual resource lifecycle
- Removed trackAction as it is now replaced by `eventTracker`

## Test instructions

I used this [HTML](https://gist.github.com/BeltranBulbarellaDD/727ac1d62e1db1b47b408567187b6818) file to test locally.

Example [Resource](https://dd.datad0g.com/rum/sessions?query=%40type%3Asession%20%40application.id%3Aa81f40b8-e9bd-4805-9b66-4e4edc529a14&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZwPeuW84EBm1QAAABhBWndQZXVXOEFBRGhDRGVNS0dndjVWTVgAAAAkMDE5YzBmN2YtYjUxZS00OTkzLWI4NGUtZjVjNWNlNjQ2NjM2AAAAWw&fromUser=false&viz=stream&from_ts=1769782876568&to_ts=1769786476568&live=true)

[Example](https://app.datadoghq.com/rum/sessions?query=%40type%3Asession%20%40application.id%3A1397744d-34f4-4a6a-a735-801e31c18221&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZwPjAVMqgJQ9wAAABhBWndQakFWTUFBQWdFZF9OQm4zVkFGVVYAAABdMTE5YzBmOGUtMmQ2Zi00NzgxLWJlZDctZWViYWFkMGMxNjA4LXN5bnRoZXRpYy10aW1lLTE3Njk3ODE2MDAwMDAwMDAwMDBjLTE3Njk3ODc1MDE5MzUwMDAwMDFvAAAABA&fromUser=false&viz=stream&from_ts=1769784030984&to_ts=1769787630984&live=true) in Flutter web.

## Checklist

- [X] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/manualEventLifecycle.ts
BeltranBulbarellaDD, line 1: The idea would be to reuse it for different events, else it adds bloat. 

Without it only for actions and resources:
Without this:  Rum | 169.41 KiB
With it: Rum | 169.52 KiB

After AI refactoring quickly VitalCollection to use it I got:
rum          |   169.45 KiB

packages/rum-core/src/domain/manualEventLifecycle.ts
```diff
@@ -0,0 +1,43 @@
+import type { LifeCycle } from './lifeCycle'
+import { LifeCycleEventType } from './lifeCycle'
+
+export interface ManualEventLifecycle<TStartData> {
```
BenoitZugmeyer: **suggestion:** don't call this "Lifecycle" as it clashes with the already existing lifecycle concept (and especially, at one point you have a `lifeCycle` and a `lifecycle` variable, which is extremely confusing).

Maybe `ManualEventRegistry`

packages/rum-core/src/domain/manualEventLifecycle.ts
```diff
@@ -0,0 +1,43 @@
...
+  start: (key: string, data: TStartData) => void
+  get: (key: string) => TStartData | undefined
+  remove: (key: string) => TStartData | undefined
+  stopAll: () => void
```
BenoitZugmeyer: **suggestion:** 
* rename `start` to `add`: "add" better relates to "remove" than "start".
* rename `stopAll` to `stop`: we always use `stop` as a convention to clean things up
* remove `get` as it's unused outside of tests (adjust tests so it uses `remove` instead of `get`).
* as it's a common pattern to pass additional data to `remove()`, it could make sense to accept a `data: Partial<TSartData>` argument
* this registry could also be in charge of computing the duration

In the end it could look like this:

```ts
export interface ManualEventRegistry<TData> {
  add: (key: string, startClocks: ClocksState, data: TData) => void
  remove: (
    key: string,
    stopClocks: ClocksState,
    data: Partial<TData>
  ) => (TData & { duration: Duration; startClocks: ClocksState }) | undefined
  stop: () => void
}

export function createManualEventLifecycle<TData>(
  lifeCycle: LifeCycle,
  onDiscard?: (data: TData) => void
): ManualEventRegistry<TData> {
  const activeEvents = new Map<string, { clocks: ClocksState; data: TData }>()

  function stopAll() {
    activeEvents.forEach((event) => onDiscard?.(event.data))
    activeEvents.clear()
  }

  lifeCycle.subscribe(LifeCycleEventType.SESSION_RENEWED, stopAll)

  return {
    add: (key, clocks, data) => {
      const existing = activeEvents.get(key)
      if (existing) {
        onDiscard?.(existing.data)
      }
      activeEvents.set(key, { clocks, data })
    },
    remove: (key, stopClocks, stopData) => {
      const startData = activeEvents.get(key)
      if (!startData) {
        return
      }

      activeEvents.delete(key)
      return {
        startClocks: startData.clocks,
        duration: elapsed(startData.clocks.relative, stopClocks.relative),
        ...(combine(startData.data, stopData) as TData),
      }
    },
    stop,
  }
}

```

packages/rum-core/src/domain/manualEventLifecycle.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function createManualEventLifecycle<TStartData>(
```
BenoitZugmeyer: **thought:** I wonder if we shouldn't merge this with `ActionTracker`. The concept is very close: we start and stop things to produce events, events have a startClocks, duration, and id. The main differences are:

* in one side, we need to refer events by key. This could be solved by generating a random uuid as a key in `newClick` for example

* in the other side, we care about counts and history. I think it's fine to have those even if we don't use them, OR add an option to disable them if we worry about the performance impact

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,150 @@
...
+    context: resource.context,
+  } as RawRumResourceEvent
+
+  return {
+    rawRumEvent,
+    startTime: resource.startClocks.relative,
+    duration: resource.duration,
+    domainContext: { isManual: true as const },
+  }
+}
```
BenoitZugmeyer: **suggestion:** The `ManualResource` model is superfluous. Just build the `RawRumResourceEvent` directly, no need to have an intermediary data model.

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,150 @@
...
+      url: resource.url,
+      duration: toServerDuration(resource.duration),
+      method: resource.method,
+      status_code: resource.statusCode,
+    },
+    _dd: {
+      discarded: false,
+    },
+    context: resource.context,
+  } as RawRumResourceEvent
```
BenoitZugmeyer: **suggestion:** instead of casting the value, just add a `context?: Context` to `RawRumResourceEvent`

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,150 @@
...
+      id: resource.id,
+      type: resource.type,
+      url: resource.url,
+      duration: toServerDuration(resource.duration),
+      method: resource.method,
+      status_code: resource.statusCode,
+    },
+    _dd: {
+      discarded: false,
+    },
```
BenoitZugmeyer: **suggestion:** I don't think this property is really required, you could probably make it optional in the type and don't provide it

packages/rum-core/src/domain/manualEventLifecycle.ts
BeltranBulbarellaDD: I see what you mean. I tried doing it and I could not make it smaller than a base bundle size increase of 0.5 MiB. I can push what I have and we could analyze it. 
If we the extend it for start/stopDurationVitals we can centralize everything under eventRegistry so it looks better.
My only worry is that this eventRegistry will do too much as it will handle automatic and manual actions and manual resources. (Since we used ActionTracker for both manual and automatic actions.)

BeltranBulbarellaDD:
@codex pls review

packages/rum-core/src/domain/action/computeFrustration.ts
```diff
@@ -10,7 +10,7 @@ export function computeFrustration(clicks: Click[], rageClick: Click) {
     if (clicks.some(isDead)) {
       rageClick.addFrustration(FrustrationType.DEAD_CLICK)
     }
-    if (rageClick.hasError) {
+    if (rageClick.hasError || clicks.some((click) => click.hasError)) {
```
BenoitZugmeyer: **question:** Why this change? It seems unrelated

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -310,17 +328,21 @@ function newClick(
...
-      return trackedAction.counts.errorCount > 0
+      return counts ? counts.errorCount > 0 : false
     },
     get hasPageActivity() {
-      return activityEndTime !== undefined
+      if (activityEndTime) {
+        return true
+      }
+      const activity = getUserActivity()
+      return !!(activity.selection || activity.input || activity.scroll)
```
BenoitZugmeyer: **question:** This change seems unrelated too no?

packages/rum-core/src/domain/action/trackManualActions.ts
```diff
@@ -37,84 +38,73 @@ export interface ManualAction {
   frustrationTypes: FrustrationType[]
 }
 
-interface ManualActionStart {
-  name: string
+export interface ActionEventData {
+  name?: string
```
BenoitZugmeyer: **question:** why is `name` optional now?

packages/rum-core/src/domain/action/trackManualActions.ts
```diff
@@ -37,84 +38,73 @@ export interface ManualAction {
...
-      activeManualActions.delete(lookupKey)
-    }
+    actionTracker.start(
+      lookupKey,
+      startClocks,
+      {
+        name,
+        type: options.type,
+        context: options.context,
+      },
```
BenoitZugmeyer, line 60: **suggestion:** 
```ts
{ name, ...options }
```

packages/rum-core/src/domain/action/trackManualActions.ts
```diff
@@ -37,84 +38,73 @@ export interface ManualAction {
...
+  function stopManualAction(name: string, options: ActionOptions = {}, stopClocks = clocksNow()) {
+    const lookupKey = options.actionKey ?? name
 
-    activeManualActions.set(lookupKey, {
-      name,
+    const stopped = actionTracker.stop(lookupKey, stopClocks, {
       type: options.type,
       context: options.context,
-      trackedAction,
     })
```
BenoitZugmeyer: **suggestion:** 
```ts
const stopped = actionTracker.stop(lookupKey, stopClocks, options)
```

packages/rum-core/src/domain/action/trackManualActions.ts
```diff
@@ -37,84 +38,73 @@ export interface ManualAction {
...
-      counts: activeAction.trackedAction.counts,
+      id: stopped.id,
+      name: data.name!,
+      type: data.type || ActionTypeEnum.CUSTOM,
+      startClocks: stopped.startClocks,
+      duration: stopped.duration,
+      context: data.context,
+      counts: stopped.counts,
       frustrationTypes,
     }
```
BenoitZugmeyer, line 83: **suggestion:** ideally, you could do:

```ts
const manualActions: ManualAction = {
  ...stopped,
  type: stopped.type || ActionTypeEnum.CUSTOM,
  frustrationTypes,
}
```

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+import { LifeCycleEventType } from './lifeCycle'
+import type { LifeCycle } from './lifeCycle'
+import type { EventCounts } from './trackEventCounts'
+import { trackEventCounts } from './trackEventCounts'
+
+export const EVENT_CONTEXT_TIME_OUT_DELAY = 5 * ONE_MINUTE // arbitrary
+
+interface BaseTrackedEvent<TData = object> {
+  id: string
+  key: string
```
BenoitZugmeyer: **suggestion:** `key` isn't used, you could remove it

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+  startClocks: ClocksState
+  counts?: EventCounts
+  data: TData
```
BenoitZugmeyer: **suggestion:** it could be nice to "merge" data into that object, ex:

```ts
type BaseTrackedEvent<TData> = TData & {
  id: string,
  startClocks: ClocksState,
 ...
}
```

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  function stopAll() {
+    keyedEvents.forEach((event) => {
+      discardEvent(event)
+    })
+    keyedEvents.clear()
+
+    activeEventCountSubscriptions.forEach((sub) => sub.stop())
+    activeEventCountSubscriptions.clear()
```
BenoitZugmeyer: **suggestion:** This shouldn't be necessary, since event count subscriptions are removed in `discardEvent`. In fact, the whole `activeEventCountSubscriptions` variable could be removed.

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+      event.eventCounts.stop()
+      activeEventCountSubscriptions.delete(event.eventCounts)
+    }
+
+    if (event.onDiscard) {
+      event.onDiscard(event.id, event.data, event.startClocks)
+    }
+  }
+
+  function stopAll() {
```
BenoitZugmeyer: **suggestion:** call this function `discardAll` since it's discarding events

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  function start(key: string, startClocks: ClocksState, data: TData, options?: StartOptions<TData>): string {
+    const id = generateUUID()
+
+    const historyEntry = history.add(id, startClocks.relative)
+
+    const existing = keyedEvents.get(key)
+    if (existing) {
+      discardEvent(existing)
+      keyedEvents.delete(key)
```
BenoitZugmeyer: **suggestion:** `keyedEvents.delete(key)` should be done in `discardEvent`

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+      id,
+      key,
+      startClocks,
+      data,
+      historyEntry,
+      eventCounts,
+      onDiscard: options?.onDiscard,
+    })
+
+    return key
```
BenoitZugmeyer: **suggestion:** no need to return the key

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,138 @@
...
+        url,
+        type: options.type,
+        method: options.method,
+        context: options.context,
+      },
+      {
+        onDiscard: (id, data, discardedStartClocks) => {
+          emitResource(id, discardedStartClocks, data)
+        },
+      }
```
BenoitZugmeyer: **question:** why do we need this behavior? It's not the same as other tracked events. We could simplify a bit by removing this.

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    const eventCounts =
+      options?.trackCounts === true
+        ? trackEventCounts({
+            lifeCycle,
+            isChildEvent: (event) =>
+              event.action !== undefined &&
+              (Array.isArray(event.action.id) ? event.action.id.includes(id) : event.action.id === id),
+          })
```
BenoitZugmeyer: **thought:** This `isChildEvent` is only relevant for actions...

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const duration = elapsed(event.startClocks.relative, stopClocks.relative)
+
+    const counts = event.eventCounts?.eventCounts
+
+    keyedEvents.delete(key)
+    if (event.eventCounts) {
+      event.eventCounts.stop()
+      activeEventCountSubscriptions.delete(event.eventCounts)
+    }
```
BenoitZugmeyer: **suggestion:** this could probably be factorized somehow

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,138 @@
...
+    lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, {
+      rawRumEvent,
+      startTime: startClocks.relative,
+      duration,
+      domainContext: { isManual: true as const },
+    })
+  }
+
+  function startManualResource(url: string, options: ResourceOptions = {}, startClocks = clocksNow()) {
+    if (!isAllowedRequestUrl(url)) {
```
BenoitZugmeyer: **suggestion:** this is probably unnecessary

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    resourceTracker.start(
+      lookupKey,
+      startClocks,
+      {
+        url,
+        type: options.type,
+        method: options.method,
+        context: options.context,
+      },
```
BenoitZugmeyer: **suggestion:** `{ url, ...options }`

packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (!stopped) {
+      return
+    }
+
+    emitResource(stopped.id, stopped.startClocks, stopped.data, options.statusCode, stopClocks)
+  }
+
+  function stop() {
+    resourceTracker.stopAll()
+  }
```
BenoitZugmeyer: **suggestion:** as `resourceTracker` is provided as argument, you shouldn't stop it here

packages/rum-core/src/domain/action/trackManualActions.ts
BeltranBulbarellaDD: So that we can use the same interface for both Manual and Click actions. I'll edit so we pass an empty name in click actions so we can keep it required.

packages/rum-core/src/domain/resource/trackManualResources.ts
BeltranBulbarellaDD: I added it because of [this](https://github.com/DataDog/browser-sdk/pull/4110#discussion_r2745300462) Codex Review.

packages/rum-core/src/domain/action/computeFrustration.ts
BeltranBulbarellaDD: Yeah it was a bug introduced in the refactoring. Fixed it with `getCounts`

BeltranBulbarellaDD:
@codex pls review

packages/rum-core/src/domain/eventTracker.ts
```diff
@@ -0,0 +1,171 @@
...
+  id: string
+  key: string
+  startClocks: ClocksState
+  data: TData
+  historyEntry: ValueHistoryEntry<string>
+  eventCounts?: ReturnType<typeof trackEventCounts>
+  onDiscard?: (id: string, data: TData, startClocks: ClocksState) => void
+}
+
+export function startEventTracker<TData>(lifeCycle: LifeCycle): EventTracker<TData> {
```
amortemousque: **🥜 nitpick:** I find the name too abstract. It’s only for manual events, so we could call it `ManualEventTracker`.

packages/rum-core/src/domain/resource/trackManualResources.ts
BenoitZugmeyer: Great, so you won't hurt its feelings by removing this :D 

Codex Review says:

> If a customer wraps fetch/XHR with startResource/stopResource (a likely use case)

But this is not a "likely use case". The Browser SDK does this already.

packages/rum-core/src/domain/action/trackClickActions.ts
```diff
@@ -278,15 +279,30 @@ export type Click = ReturnType<typeof newClick>
...
   startEvent: MouseEventOnElement
 ) {
-  const trackedAction: TrackedAction = actionTracker.createTrackedAction(relativeToClocks(startEvent.timeStamp))
+  const clickKey = generateUUID()
+  const startClocks = relativeToClocks(startEvent.timeStamp)
+
+  actionTracker.start(
+    clickKey,
+    startClocks,
+    { name: '' },
```
BenoitZugmeyer: **thought:** This looks a bit weird. Maybe we could pass `clickActionBase` in there, but I think types will clash too much with manual action tracking. So it might be fine!

BenoitZugmeyer (approved):

PR #4124, opened 2026-01-26
feat: Add in-memory session persistence option

andrewatwood:
## Motivation

A not-insignificant cohort (>5%) of users for the application I work on have their cookies and local storage disabled. Their sessions are short (~15s happy path) and self-contained, and the relevant context for any given session is nearly entirely limited to that individual single-page session, so long term persistence or multi-tab/window support is wholly unnecessary. However, tracking these sessions and any issues that my arise is still a requirement.

In-memory session storage for the RUM SDK allows us to bypass the requirement for enabled cookies or local storage, and have full access to the sessions originating from these users. 

This has been tested extensively, and deployed successfully in production, with over 4M sessions using this persistence strategy in the last two weeks alone.

## Changes

- Adds a new session persistence strategy (`SessionPersistence.IN_MEMORY`) that stores session data in memory
- Updates configuration options to expose this strategy to the end user
- Refactors and rewords existing configuration unit tests related to the `allowFallbackToLocalStorage` flag for clarity

## Test instructions

- Disable all cookies and site storage in browser settings (`On-device site data` in Chrome)
- Launch a RUM session configured to use `sessionPersistence: 'in-memory'`
- Verify that within a single tab, the session is recorded and uploaded successfully

I do likely need some help fleshing out any further e2e test coverage deemed necessary, as some of the test scripts seem to require more setup information than I am privy to.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -136,6 +136,7 @@ export { readBytesFromStream } from './tools/readBytesFromStream'
 export type { SessionState } from './domain/session/sessionState'
 export { STORAGE_POLL_DELAY } from './domain/session/sessionStore'
 export { SESSION_STORE_KEY } from './domain/session/storeStrategies/sessionStoreStrategy'
+export { MEMORY_SESSION_STORE_KEY } from './domain/session/storeStrategies/sessionInMemory'
 export {
   willSyntheticsInjectRum,
   getSyntheticsTestId,
--- a/test/e2e/scenario/sessionStore.scenario.ts
+++ b/test/e2e/scenario/sessionStore.scenario.ts
@@ -1,4 +1,4 @@
-import { SESSION_STORE_KEY } from '@datadog/browser-core'
+import { SESSION_STORE_KEY, MEMORY_SESSION_STORE_KEY } from '@datadog/browser-core'
 import type { BrowserContext, Page } from '@playwright/test'
 import { test, expect } from '@playwright/test'
 import type { RumPublicApi } from '@datadog/browser-rum-core'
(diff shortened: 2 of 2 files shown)


bcaudan:
Hi @andrewatwood, thanks for your contribution!

Having in-memory session storage would be a great addition, two concerns though:

*1. Fallback strategy:*

From my understanding of your issue, you'd want to track users that have both cookies and local storage disabled and you'd want to use in memory session in that case.
During your tests, did you then set the storage in memory for all users or did you detect that cookies and local storage were disabled to configure the sdk accordingly?
We'd prefer to provide and document this session storage strategy with the possibility to configure a fallback to it, in order to avoid any misconfiguration that could have an impact on billing.
As we have deprecated the `allowFallbackToLocalStorage` configuration option, our plan to continue to support the fallback behavior is to update the `sessionPersistence` configuration option to take an array, declaring the allowed storages and fallback order. 
Something like `sessionPersistence: ['cookie', 'local-storage', 'memory']` would try cookie first, then local storage and finally memory.
We plan to work on this update before the next major release which should be next quarter. 

*2. Session sharing between rum and logs:*

To correlate RUM and Logs data, the two SDKs share the same storage entry and we want to have the same behavior with session in memory.
We could probably achieve that by sharing the session state in a global variable, something like `DD_SESSION`.


We don't anticipate the changes needed to address those concerns to be too tricky, so if you want to have a go at it, we could provide additional support.
Otherwise, we'd adjust your PR with those changes when we'll work on the `sessionPersistence` update.

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -40,28 +41,61 @@ export const STORAGE_POLL_DELAY = ONE_SECOND
...
+
+  if (sessionPersistence !== undefined) {
+    return [sessionPersistence]
+  }
+
+  // Legacy default behavior: cookie first, with optional localStorage fallback
+  return initConfiguration.allowFallbackToLocalStorage
+    ? [SessionPersistence.COOKIE, SessionPersistence.LOCAL_STORAGE]
+    : [SessionPersistence.COOKIE]
+}
```
BenoitZugmeyer, line 80: **thought:** 🤔 in the future (v7), we might be able to simplify this using `array.flat()`, but it's not yet available for Chrome 67


```ts
function normalizePersistenceList(
  sessionPersistence: SessionPersistence | SessionPersistence[] | undefined,
  initConfiguration: InitConfiguration
): SessionPersistence[] {
   const defaultSessionPersistence = initConfiguration.allowFallbackToLocalStorage
    ? [SessionPersistence.COOKIE, SessionPersistence.LOCAL_STORAGE]
    : SessionPersistence.COOKIE
  return [sessionPersistence ?? defaultSessionPersistence].flat()
}
```

BenoitZugmeyer (approved):

packages/core/src/domain/session/sessionStore.ts
andrewatwood: Good callout, that'd be nice.

thomas-lebeau (approved):
💯 Thanks a ton for your contribution @andrewatwood !

Issue #4063, opened 2026-01-08
💡 Enhance support of persisted Graphql queries


valtechtmn:
Recently support was added for persisted queries through this PR:
https://github.com/DataDog/browser-sdk/pull/3992

However there is still a gap that affects anyone using useGETForHashedQueries as is recommended by Apollo's doc:
https://www.apollographql.com/docs/apollo-server/performance/apq#step-2-enable-automatic-persisted-queries

In this case any graphql query is a GET instead of a POST and the usual properties (operationName, variables, ...) are passed as query param instead of payload. This means the extractGraphQlRequestMetadata function which is only looking at request body misses the data and extracts nothing.

Would it be possible if the requestBody is empty to check the query params to see if the data is there and then proceed as normal ?

rgaignault:
Hey @valtechtmn, thanks for the feedback ! I created a ticket and I will let you know once it's merged

PR #4125, opened 2026-01-27
✨ [RUM-13680] Add support for GraphQL GET requests with query params (APQ)

rgaignault:
## Motivation

When using Apollo's Automatic Persisted Queries (APQ) with useGETForHashedQueries: true, GraphQL requests are sent as GET instead of POST. In this case, operationName, variables, and query are passed as URL query params instead of request body.

The extractGraphQlRequestMetadata function only inspects the request body, so it misses GraphQL metadata for GET requests in this particular context. As it was requested by a customer we could support it by parsing the url. 

## Changes

We could support it by parsing the url. 

FR:https://github.com/DataDog/browser-sdk/issues/4063

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/resource/graphql.ts
+++ b/packages/rum-core/src/domain/resource/graphql.ts
@@ -124,8 +124,7 @@ function extractFromBody(requestBody: unknown): RawGraphQlMetadata | undefined {
 function extractFromUrlQueryParams(url: string): RawGraphQlMetadata {
   const searchParams = buildUrl(url).searchParams
   const variablesParam = searchParams.get('variables')
-  const variables =
-    variablesParam && tryJsonParse(variablesParam) !== undefined ? variablesParam : undefined
+  const variables = variablesParam && tryJsonParse(variablesParam) !== undefined ? variablesParam : undefined
 
   return {
     query: searchParams.get('query') || undefined,


packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -27,7 +27,7 @@ export function extractGraphQlMetadata(
   request: RequestCompleteEvent,
   graphQlConfig: GraphQlUrlOption
 ): GraphQlMetadata | undefined {
-  const metadata = extractGraphQlRequestMetadata(request.requestBody, graphQlConfig.trackPayload)
+  const metadata = extractGraphQlRequestMetadata(request.requestBody, graphQlConfig.trackPayload, request.url)
```
bcaudan: **💬 suggestion:** ‏what about passing directly the full request to this method? instead of passing every needed request attributes

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -82,12 +82,19 @@ export function findGraphQlConfiguration(url: string, configuration: RumConfigur
...
+  url?: string
 ): GraphQlMetadata | undefined {
-  if (!requestBody || typeof requestBody !== 'string') {
-    return
+  if (requestBody && typeof requestBody === 'string') {
+    return extractFromBody(requestBody, trackPayload)
+  }
+  // Fallback for persisted queries
+  if (url) {
+    return extractFromUrlQueryParams(url, trackPayload)
```
bcaudan: **💬 suggestion:** ‏what about testing for GET or POST to choose the right extraction strategy?

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -105,23 +112,66 @@ export function extractGraphQlRequestMetadata(
...
+}
+
+function extractFromUrlQueryParams(url: string, trackPayload: boolean): GraphQlMetadata | undefined {
+  const queryStringIndex = url.indexOf('?')
+  if (queryStringIndex === -1) {
+    return
+  }
+
+  const searchParams = new URLSearchParams(url.slice(queryStringIndex + 1))
+  const variablesParam = searchParams.get('variables')
```
bcaudan: **💬 suggestion:** ‏what about using `buildUrl(url).searchParams` directly?

packages/rum-core/src/domain/resource/graphql.ts
bcaudan: **💬 suggestion:** ‏Instead of having the extract methods that call the build method internally, we could:
- rename the build method as sanitize since it seems to be the intent
- call the sanitize method from extractGraphQlRequestMetadata , so the specific extract methods would be more focused and the higher level method would manage the execution flow
- move the parseVariableParams to the sanitize function as well

packages/rum-core/src/domain/resource/graphql.ts
rgaignault: Oh yeah I forget about this function, good catch 👍

packages/rum-core/src/domain/resource/graphql.ts
```diff
@@ -105,26 +127,49 @@ export function extractGraphQlRequestMetadata(
...
-  if (graphqlBody.variables) {
-    variables = JSON.stringify(graphqlBody.variables)
+  if (rawMetadata.variables) {
+    try {
+      // Parse to validate it's valid JSON, then keep the string
+      JSON.parse(rawMetadata.variables)
+      variables = rawMetadata.variables
+    } catch {
+      // Invalid JSON in variables, ignore
+    }
```
BenoitZugmeyer: **suggestion:** do we need to validate this? Maybe we can just trust that it is correctly encoded?

If we still want this validation, I suggest to:
* move it to `extractFromUrlQueryParams`, since it's only useful in this case
* maybe introduce a `tryJsonParse` helper, because we already have a few usages in this file. Ex:

```ts

function tryJsonParse<T = unknown>(text: string): T | undefined {
  try { return JSON.parse(text) } catch {}
}
```

Then

```ts
  let graphqlBody: {
    query?: string
    operationName?: string
    variables?: unknown
  }

  try {
    graphqlBody = JSON.parse(requestBody)
  } catch {
    // Not valid JSON
    return
  }

  if (!graphqlBody) {
    return
  }

// becomes

  const graphqlBody = tryJsonParse<{
    query?: string
    operationName?: string
    variables?: unknown
  }>(requestBody)

  if (!graphqlBody) {
    return
  }
```

PR #4129, opened 2026-01-28
✅  Introduce a new way to mock values in unit tests

BenoitZugmeyer:
## Motivation

This is a proposal to allow mocking values (mostly functions) within tests.

The current way to mock values used within a function is to pass them as argument of that function. For example:

```ts
// foo.ts
import { startBar } from 'bar'

function createFoo(startBarImpl = startBar) {
  const bar = startBarImpl()
}


// foo.spec.ts
it('create foo', () => {
  createFoo(mockStartBar)
})
```

However, forces to have unecessary extra arguments in functions. Impls also need to be propagated within subfunctions:

```ts
// foo.ts
function createFoo(startBarImpl = startBar) {
  const baz = trackBaz(startBarImpl)
}

// baz.ts
function trackBaz(startBarImpl = startBar) {
}
```

This PR introduce a new way to mock values that is similar to a lightweight dependency injection system. Values that can be mocked should be wrapped in a `mocked()` call when used, and tests can use `mockValue()` to replace the value or `mockWithSpy()` to replace it with a jasmine spy. The example above would be implemented as:

```ts
// foo.ts
import { startBar } from 'bar'

function createFoo() {
  const bar = mockable(startBar)()
}

// foo.spec.ts
it('create foo', () => {
  replaceMockable(startBar, mockStartBar)
  // or
  replaceMockableWithSpy(startBar).and.callFake(mockStartBar)
  createFoo()
})
```

## Changes

The first 2 commits does a bit of spec cleaning.

Then, we introduce `mockable` capabilities and use it in tests, mostly to replace all `*Impl` arguments and location mocking.

## Test instructions

If tests passes, it should be good.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/resource/resourceCollection.spec.ts
+++ b/packages/rum-core/src/domain/resource/resourceCollection.spec.ts
@@ -39,12 +39,7 @@ describe('resourceCollection', () => {
     const taskQueue = createTaskQueue()
     replaceMockable(createTaskQueue, () => taskQueue)
     taskQueuePushSpy = spyOn(taskQueue, 'push')
-    const startResult = startResourceCollection(
-      lifeCycle,
-      { ...baseConfiguration, ...partialConfig },
-      pageStateHistory,
-      taskQueue
-    )
+    const startResult = startResourceCollection(lifeCycle, { ...baseConfiguration, ...partialConfig }, pageStateHistory)
 
     rawRumEvents = collectAndValidateRawRumEvents(lifeCycle)
 


packages/logs/src/boot/logsPublicApi.ts
```diff
@@ -291,8 +285,7 @@ export function makeLogsPublicApi(startLogsImpl: StartLogs, startTelemetryImpl?:
 
       strategy = createPostStartStrategy(initConfiguration, startLogsResult)
       return startLogsResult
-    },
-    startTelemetryImpl
```
thomas-lebeau, line 295: **👏 praise:** ‏Nice! It was just argument drilling!

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -45,8 +46,7 @@ export function startResourceCollection(
   lifeCycle: LifeCycle,
   configuration: RumConfiguration,
   pageStateHistory: PageStateHistory,
-  taskQueue = createTaskQueue(),
-  retrieveInitialDocumentResourceTimingImpl = retrieveInitialDocumentResourceTiming
+  taskQueue = createTaskQueue()
```
thomas-lebeau: **💬 suggestion:** ‏I think we can use mockable for that too

packages/rum/src/boot/recorderApi.ts
```diff
@@ -91,14 +87,9 @@ export function makeRecorderApi(
...
-          configuration,
-          'Datadog Session Replay',
-          () => {
-            strategy.stop()
-          },
-          createDeflateWorkerImpl
-        )
+        worker ??= startDeflateWorker(configuration, 'Datadog Session Replay', () => {
+          strategy.stop()
+        })
```
thomas-lebeau: **🥜 nitpick:** ‏

```suggestion
        worker ??= startDeflateWorker(configuration, 'Datadog Session Replay', () => strategy.stop())
```

packages/rum/src/domain/deflate/deflateWorker.ts
```diff
@@ -50,12 +50,11 @@ let state: DeflateWorkerState = { status: DeflateWorkerStatus.Nil }
 export function startDeflateWorker(
   configuration: RumConfiguration,
   source: string,
-  onInitializationFailure: () => void,
-  createDeflateWorkerImpl = createDeflateWorker
+  onInitializationFailure: () => void
```
thomas-lebeau, line 53: **❓ question:** ‏This seems to be exposed only to inject a spy on the tests, could we use mockble here too?

AGENTS.md
```diff
@@ -71,11 +71,12 @@ scripts/             # Build, deploy, release automation
...
 
-### Test-Driven Development
+### Unit Tests
 
-- Spec files co-located with implementation: `feature.ts` → `feature.spec.ts`
+- Test framework: Jasmine + Karma. Spec files co-located with implementation: `feature.ts` → `feature.spec.ts`
+- Focus tests with `fit()` / `fdescribe()`, skip with `xit()` / `xdescribe()`
 - Use `registerCleanupTask()` for cleanup, NOT `afterEach()`
-- Test framework: Jasmine + Karma
+- Mock values/functions: wrap with `mockable()` in source, use `replaceMockable()` or `replaceMockableWithSpy()` in tests (auto-cleanup)
```
thomas-lebeau, line 79: **👏 praise:** ‏Yeah! 🤖

packages/rum/src/domain/deflate/deflateWorker.ts
BenoitZugmeyer: We use this callback outside of tests [here](https://github.com/DataDog/browser-sdk/blob/0a1eb6d0e99f902b1237646de470ab13a03eb175/packages/rum/src/boot/recorderApi.ts#L90), no?

thomas-lebeau: oh right, I misread that. all good.

PR #4136, opened 2026-01-30
👷 skip BrowserStack tests for non-code changes

thomas-lebeau:
## Motivation

The `unit-bs` and `e2e-bs` jobs use BrowserStack resources which are limited to 5 seats. These jobs are slow and consume resources even when PRs don't touch code that affects browser behavior (e.g., CI config changes, script updates, documentation).

## Changes

- Convert `ci-bs.sh` to TypeScript (`ci-bs.ts`) for consistency with other scripts
- The script checks if any changed files match patterns that affect browser behavior:
  - `packages/**` (SDK source code)
  - `test/**` (test files and infrastructure)
  - `developer-extension/**` (devtools extension)
  - `scripts/test/**` (test scripts)
  - `package.json`, `yarn.lock` (dependencies)
  - `tsconfig*`, `webpack.base.ts` (build config)

If no relevant files changed, the script exits early without running BrowserStack tests.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/test/ci-bs.ts
+++ b/scripts/test/ci-bs.ts
@@ -3,16 +3,16 @@ import { command } from '../lib/command.ts'
 import { fetchPR, getLastCommonCommit, LOCAL_BRANCH } from '../lib/gitUtils.ts'
 import { getBrowserStackUsername, getBrowserStackAccessKey } from '../lib/secrets.ts'
 
-// Patterns that should trigger BrowserStack tests
+// Patterns that should trigger BrowserStack tests (git pathspecs)
 const RELEVANT_FILE_PATTERNS = [
-  /^packages\//,
-  /^test\//,
-  /^developer-extension\//,
-  /^scripts\/test\//,
-  /^package\.json$/,
-  /^yarn\.lock$/,
-  /^tsconfig/,
-  /^webpack\.base\.ts$/,
+  'packages/*',
+  'test/*',
+  'developer-extension/*',
+  'scripts/test/*',
+  'package.json',
+  'yarn.lock',
+  'tsconfig*',
+  'webpack.base.ts',
 ]
 
 runMain(async () => {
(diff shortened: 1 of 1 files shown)


scripts/test/ci-bs.ts
```diff
@@ -0,0 +1,57 @@
...
+    .withEnvironment({
+      BS_USERNAME: getBrowserStackUsername(),
+      BS_ACCESS_KEY: getBrowserStackAccessKey(),
+    })
+    .withLogs()
+    .run()
+})
+
+function hasRelevantChanges(baseCommit: string): boolean {
+  const changedFiles = command`git diff --name-only ${baseCommit} HEAD`.run()
```
BenoitZugmeyer: **suggestion:** I find glob patterns easier to read than regexes when it comes to file path. Also, git supports it, so why not leveraging it:

```ts
const RELEVANT_FILES_PATTERNS = [
  "packages/*",
  "test/*",
  "developer-extension/*",
  "scripts/test/*",
  "package.json",
  "yarn.lock",
  "tsconfig*",
  "webpack.base.ts"
]

  const changedFiles = command`git diff --name-only ${baseCommit} HEAD ${RELEVANT_FILES_PATTERNS}`.run()
  if (changedFiles.trim()) {
    printLog("Matched files: ${changedFiles)")
    return true
  }
  return false
```

scripts/test/ci-bs.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (!hasRelevantChanges(baseCommit)) {
+    printLog('No code changes affecting browser behavior detected. Skipping BrowserStack tests.')
+    return
+  }
+
+  command`yarn ${testCommand}:bs`
+    .withEnvironment({
+      BS_USERNAME: getBrowserStackUsername(),
+      BS_ACCESS_KEY: getBrowserStackAccessKey(),
```
BenoitZugmeyer, line 36: **thought:** we could probably use those functions directly instead of passing values through environment variables. Although, I use environment variables to run unit/e2e on BS on my local, so... I'm good :D

thomas-lebeau: yeah, me too. I want to keep it working with env for running test on bs from my local env

BenoitZugmeyer (approved):

PR #4148, opened 2026-02-06
✨ Add action and vital metadata to profiles

MandragoreVR:
## Motivation

To enable profiling aggregation on the backend for vitals and actions, we want to add vitals and actions data to the metadata of profiling events.

This PR adds these entries to the profiling events metadata.

## Changes

- Create a `vitalHistory` next to the profiler that catches vital events from the life cycle and stores them in the history
	- Vital events are currently only sent in the lifecycle when they end. However, one vital event might span over multiple profiles, so the profile where this vital event starts must be tagged with this vital event too.
		- This PR defines the id of duration vitals when they start instead of when they end, and adds a new lifecycle event to broadcast vitals when they start
		- Note: this will not cause issues if the vital doesn't end, because this is only used to tag profile events, so in the worst case, the id of this vital will not be used
- Same thing for the actions: creation of an `actionHistory`
	- the id of the action was created on action start, but not returned by the event tracker, so I changed that to make the `start` method of the event tracker return the stored data
- Update the build of a profile's attributes to add actions and vitals
	- update profiler's tests to make sure vitals and actions are collected

## Test instructions
- In the file `sandbox/react-app/main.tsx`, add:
```ts
profilingSampleRate: 100,
  site: 'datadoghq.com',
  service: 'sandbox-react-app',
  env: 'development',
  beforeSend: () => false,
```
to the SDK configuration, and add
```tsx
  useEffect(() => {
    datadogRum.startDurationVital('test vital')

    const timeout = setTimeout(() => {
      datadogRum.stopDurationVital('test vital')
    }, 1000)

    return () => {
      clearTimeout(timeout)
    }
  }, [])
```
before the `return` of the `Layout` function to have a vital that will be displayed in the collected ones.
- Run `yarn dev` and go to `localhost:8080/react-app`
- Wait for a request to `https://browser-intake-datadoghq.com/api/v2/profile` in the network tab (profiles usually last one minute, so this might not come instantly, but you can also edit the constant `collectIntervalMs` in `profiler.ts` to make it faster)
- Look at the payload, that contains two big objects.
	- The first one should contain the fields `"action"` and `"vital"`, both with an object of the shape `{ "id": string[], "label": string[] }`
	- The second one should contain the fields `"action"` and `"vitals"` with the following shape:
image

## Checklist

- [x] Tested locally

### DIFF
--- a/packages/rum-core/src/domain/vital/vitalCollection.ts
+++ b/packages/rum-core/src/domain/vital/vitalCollection.ts
@@ -80,6 +80,7 @@ interface BaseVital extends VitalOptions {
   handlingStack?: string
 }
 export interface DurationVital extends BaseVital {
+  id: string
   type: typeof VitalType.DURATION
   duration: Duration
 }
@@ -111,9 +112,9 @@ export function startVitalCollection(
     return !pageStateHistory.wasInPageStateDuringPeriod(PageState.FROZEN, vital.startClocks.relative, vital.duration)
   }
 
-  function addDurationVital(vital: DurationVital, vitalId?: string) {
+  function addDurationVital(vital: DurationVital) {
     if (isValid(vital)) {
-      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processVital(vital, vitalId))
+      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processVital(vital))
     }
   }
 
@@ -183,7 +184,7 @@ export function startDurationVital(
 }
 
 export function stopDurationVital(
-  stopCallback: (vital: DurationVital, vitalId: string) => void,
+  stopCallback: (vital: DurationVital) => void,
   { vitalsByName, vitalsByReference }: CustomVitalsState,
   nameOrRef: string | DurationVitalReference,
   options: DurationVitalOptions = {}
@@ -194,7 +195,7 @@ export function stopDurationVital(
     return
   }
 
-  stopCallback(buildDurationVital(vitalStart, vitalStart.startClocks, options, clocksNow()), vitalStart.id)
+  stopCallback(buildDurationVital(vitalStart, vitalStart.startClocks, options, clocksNow()))
 
   if (typeof nameOrRef === 'string') {
     vitalsByName.delete(nameOrRef)
@@ -210,6 +211,7 @@ function buildDurationVital(
   stopClocks: ClocksState
 ): DurationVital {
   return {
+    id: vitalStart.id,
     name: vitalStart.name,
     type: VitalType.DURATION,
     startClocks,
@@ -220,11 +222,10 @@ function buildDurationVital(
   }
 }
 
-function processVital(
-  vital: DurationVital | OperationStepVital,
-  vitalId?: string
-): RawRumEventCollectedData<RawRumVitalEvent> {
+function processVital(vital: DurationVital | OperationStepVital): RawRumEventCollectedData<RawRumVitalEvent> {
   const { startClocks, type, name, description, context, handlingStack } = vital
+  const vitalId = vital.type === VitalType.DURATION ? vital.id : undefined
+
   const vitalData = {
     id: vitalId ?? generateUUID(),
     type,
--- a/packages/rum-core/src/boot/preStartRum.spec.ts
+++ b/packages/rum-core/src/boot/preStartRum.spec.ts
@@ -668,7 +668,13 @@ describe('preStartRum', () => {
         addDurationVital: addDurationVitalSpy,
       } as unknown as StartRumResult)
 
-      const vitalAdd = { name: 'timing', type: VitalType.DURATION, startClocks: clocksNow(), duration: 100 as Duration }
+      const vitalAdd = {
+        id: 'id',
+        name: 'timing',
+        type: VitalType.DURATION,
+        startClocks: clocksNow(),
+        duration: 100 as Duration,
+      }
       strategy.addDurationVital(vitalAdd)
       strategy.init(DEFAULT_INIT_CONFIGURATION, PUBLIC_API)
       expect(addDurationVitalSpy).toHaveBeenCalledOnceWith(vitalAdd)
--- a/packages/rum-core/src/boot/rumPublicApi.spec.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.spec.ts
@@ -925,6 +925,7 @@ describe('rum public api', () => {
         description: 'description-value',
       })
       expect(addDurationVitalSpy).toHaveBeenCalledWith({
+        id: jasmine.any(String),
         name: 'foo',
         startClocks: timeStampToClocks(startTime),
         duration: 100,
--- a/packages/rum-core/src/boot/rumPublicApi.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.ts
@@ -36,6 +36,7 @@ import {
   isExperimentalFeatureEnabled,
   ExperimentalFeature,
   mockable,
+  generateUUID,
 } from '@datadog/browser-core'
 
 import type { LifeCycle } from '../domain/lifeCycle'
@@ -890,6 +891,7 @@ export function makeRumPublicApi(
       callMonitored(() => {
         addTelemetryUsage({ feature: 'add-duration-vital' })
         strategy.addDurationVital({
+          id: generateUUID(),
           name: sanitize(name)!,
           type: VitalType.DURATION,
           startClocks: timeStampToClocks(options.startTime as TimeStamp),
--- a/packages/rum-core/src/domain/vital/vitalCollection.spec.ts
+++ b/packages/rum-core/src/domain/vital/vitalCollection.spec.ts
@@ -1,6 +1,6 @@
 import type { Duration } from '@datadog/browser-core'
 import { mockClock, type Clock } from '@datadog/browser-core/test'
-import { addExperimentalFeatures, clocksNow, ExperimentalFeature } from '@datadog/browser-core'
+import { addExperimentalFeatures, clocksNow, ExperimentalFeature, generateUUID } from '@datadog/browser-core'
 import { collectAndValidateRawRumEvents, mockPageStateHistory } from '../../../test'
 import type { RawRumEvent, RawRumVitalEvent } from '../../rawRumEvent.types'
 import { VitalType, RumEventType } from '../../rawRumEvent.types'
@@ -37,8 +37,7 @@ describe('vitalCollection', () => {
         stopDurationVital(cbSpy, vitalsState, vitalRef)
 
         expect(cbSpy).toHaveBeenCalledOnceWith(
-          jasmine.objectContaining({ name: 'foo', duration: 100 }),
-          jasmine.any(String)
+          jasmine.objectContaining({ id: jasmine.any(String), name: 'foo', duration: 100 })
         )
       })
 
@@ -49,10 +48,7 @@ describe('vitalCollection', () => {
         clock.tick(100)
         stopDurationVital(cbSpy, vitalsState, 'foo')
 
-        expect(cbSpy).toHaveBeenCalledOnceWith(
-          jasmine.objectContaining({ name: 'foo', duration: 100 }),
-          jasmine.any(String)
-        )
+        expect(cbSpy).toHaveBeenCalledOnceWith(jasmine.objectContaining({ name: 'foo', duration: 100 }))
       })
 
       it('should only create a single duration vital from a vital name', () => {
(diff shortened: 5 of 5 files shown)


packages/rum/src/domain/profiling/vitalHistory.ts
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  const startedVitals = new Map<string, ValueHistoryEntry<VitalContext>>()
+
+  lifeCycle.subscribe(LifeCycleEventType.VITAL_STARTED, (vitalStart) => {
+    const startedVitalHistoryEntry = history.add(
+      {
+        id: vitalStart.id,
+        startClocks: vitalStart.startClocks,
+        duration: 0 as Duration,
```
thomasbertet: I wonder if we should have another value for "non-stopped" events 🤔 
Relying on 0 seems like it could be error prone and the BE that process should be aware 0 means "not stopped".
Maybe just `undefined` would work instead ? WDYT ?

MandragoreVR: Good question... could we maybe settle for something like -1, though, to not have to re-do a `rum-events-format` PR to allow `undefined` here?

thomasbertet: Yes, but we will need to adjust the logic, I kinda forgot about this when I did the BE logic : https://github.com/DataDog/profiling-backend/pull/7969/changes#diff-fe24f465d52fe19d3a52bfd86ac3eaa5e441e710eca6bb9c26a14cac5ce169e4R339-R346 

You can see I'm just using the `duration` field without taking into account if the duration is 0 or undefined. 
Let's have -1 as signal for unfinished events. I'll prepare the BE to support that value.

thomasbertet: Actually maybe that's cleaner if that's undefined, even if we have to do another PR for the format, WDYT ?

MandragoreVR: Works for me, I just updated!

packages/rum/src/domain/profiling/actionHistory.ts
```diff
@@ -0,0 +1,64 @@
...
+      },
+      actionStart.startClocks.relative
+    )
+
+    startedActions.set(actionStart.id, startedActionHistoryEntry)
+  })
+
+  lifeCycle.subscribe(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, ({ rawRumEvent, startClocks, duration }) => {
+    if (rawRumEvent.type === 'action') {
+      const durationForEntry = duration ?? (0 as Duration)
```
amortemousque, line 40: You can use `duration!` like we do in `longTaskHistory`

packages/rum/src/domain/profiling/actionHistory.ts
```diff
@@ -0,0 +1,64 @@
...
+          actionHistoryEntry.value.duration = duration!
+          actionHistoryEntry.close(addDuration(startClocks.relative, durationForEntry))
+          startedActions.delete(rawRumEvent.action.id)
+        }
+      } else {
+        history
+          .add(
+            {
+              id: rawRumEvent.action.id,
+              label: '',
```
amortemousque, line 53: **❓ question:** ‏Is it expected not to have a label? Shouldn't it be the action name?

packages/rum/src/domain/profiling/actionHistory.ts
```diff
@@ -0,0 +1,64 @@
...
+  duration?: Duration
+  startClocks: ClocksState
+}
+
+export function createActionHistory(lifeCycle: LifeCycle) {
+  const history = createValueHistory<ActionContext>({
+    expireDelay: ACTION_ID_HISTORY_TIME_OUT_DELAY,
+  })
+
+  const startedActions = new Map<string, ValueHistoryEntry<ActionContext>>()
```
amortemousque: **💬 suggestion:** You can use only the `history` and do       `history.findAll(startClocks.relative).find((action) => action.id === rawRumEvent.action.id)`to locate the startedAction you want to update.

packages/rum-core/src/domain/action/trackManualActions.ts
```diff
@@ -60,6 +61,8 @@ export function trackManualActions(lifeCycle: LifeCycle, onManualActionCompleted
       },
       { isChildEvent: isActionChildEvent }
     )
+
+    lifeCycle.notify(LifeCycleEventType.ACTION_STARTED, startedManualAction)
```
amortemousque, line 65: **💬 suggestion:** ‏You could make this part of eventTracker so you have the lifecycle event for click and manual actions. We also are planning to use eventTracker for vitals.

packages/rum/src/domain/profiling/actionHistory.ts
MandragoreVR: The thing is that the duration is not always defined here since we can have punctual actions (or actions that are still ongoing) that don't have a duration. In these cases, we do need to default to 0 for the `close` method of the history, but we keep the `undefined` duration for the profiling metadata (which is not perfectly reflected, I will update that)

packages/rum/src/domain/profiling/actionHistory.ts
MandragoreVR: The issue here is that actions' names might be redacted by the customers in `beforeSend`, and we currently have no way of knowing this (or at least no easy way). After talking about this with @thomasbertet, we decided that we would be fine without the action names, but I preferred leaving the field as an empty string here instead of removing it because we will still try to add it later!

packages/rum/src/domain/profiling/actionHistory.ts
MandragoreVR: The thing is that `history.findAll` returns the actual values and not the history entries (on which I can call `.close`), except if I missed a different method that gives me the history entry?

packages/rum-core/src/domain/action/trackManualActions.ts
MandragoreVR: I don't really have a way to know which event type I'm dealing with in the `start` method of the event tracker, right? So do you imagine one of the two following things:
- `start` takes an optional function executed before returning, and that would in our case emit the life cycle event?
- we pass an event type at the creation of the event tracker so that we know what we're dealing with (which might actually have other uses later, it can be useful to know which event type we're dealing with 🤷‍♂️)

packages/rum/src/domain/profiling/actionHistory.ts
amortemousque: Ah yes :) In that case, let’s add a method on the [ValueHistory](https://github.com/DataDog/browser-sdk/blob/5e88cac6b9e730232d4536ad29a5bb2422d1fa17/packages/core/src/tools/valueHistory.ts#L23), for instance `history.getEntry(startClocks.relative)`
It would return the entry that matches the given `startTime`. This lets us remove `startedActions` and simplify the code like so:
```ts
if (rawRumEvent.type === 'action') {
  const entry =
    history.getEntry(startClocks.relative) ??
    history.add({ id: rawRumEvent.action.id, label: '', startClocks }, startClocks.relative)

  entry.value.duration = duration!
  entry.close(addDuration(startClocks.relative, duration ?? (0 as Duration)))
}
```

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -171,11 +181,15 @@ export function startDurationVital(
   // To avoid memory leaks caused by the creation of numerous references (e.g., from improper useEffect implementations), we use a WeakMap.
   vitalsByReference.set(reference, vital)
 
+  if (lifeCycle) {
```
amortemousque: **💬 suggestion:** ‏Instead of having an optional `lifeCycle`, you could send the event here [here](https://github.com/DataDog/browser-sdk/blob/48e21b77e01ff3b3f5a7c3adeac2ec50b4f7d740/packages/rum-core/src/domain/vital/vitalCollection.ts#L154). Like so 
```ts
startDurationVital: (name: string, options: DurationVitalOptions = {}) => {
  const ref = startDurationVital(customVitalsState, name, options)
  lifeCycle.notify(LifeCycleEventType.VITAL_STARTED, customVitalsState.vitalsByReference.get(ref)!)
  return ref
},
```

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -145,19 +152,22 @@ export function startVitalCollection(
...
 
 export function startDurationVital(
   { vitalsByName, vitalsByReference }: CustomVitalsState,
+  lifeCycle: LifeCycle | undefined, // the lifecycle might not be defined in case datadogRum is not initialized yet
   name: string,
   options: DurationVitalOptions = {}
 ) {
+  const vitalId = generateUUID()
   const vital = {
+    id: vitalId,
```
amortemousque: **🥜 nitpick:** ‏
```suggestion
const vital = {
    id: generateUUID(),
```

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -116,6 +117,12 @@ export function startVitalCollection(
     }
   }
 
+  function addDurationVitalWithId(vital: DurationVital, vitalId: string) {
+    if (isValid(vital)) {
+      lifeCycle.notify(LifeCycleEventType.RAW_RUM_EVENT_COLLECTED, processVital(vital, vitalId))
+    }
+  }
+
```
amortemousque: **❓ question:** ‏Why do you need this function? 
Adding the `id` to `BaseVital` to follows the same logic as the other vital attributes should be good.

packages/rum-core/src/domain/action/trackManualActions.ts
amortemousque: Yes right, it’s a bit tricky. Let’s keep it simple.

You can keep it here, but we also need to handle click actions. So it should be added [here](https://github.com/DataDog/browser-sdk/blob/5e88cac6b9e730232d4536ad29a5bb2422d1fa17/packages/rum-core/src/domain/action/trackClickActions.ts#L293)￼ as well.

packages/rum/src/domain/profiling/actionHistory.ts
amortemousque: I see. We usually remove unnecessary attributes to avoid confusion. Would it be too much work to remove it?

packages/rum/src/domain/profiling/actionHistory.ts
amortemousque: Oh yes, my bad :)

packages/rum/src/domain/profiling/actionHistory.ts
MandragoreVR: Well I was planning to try a follow-up to see if can get the action name in some cases when it's not redacted. Is that okay if I remove it later, only if my follow-up doesn't work?

(Otherwise, I wouldn't say it's _too much_ work, just that we need to do a PR on `rum-events-format` to remove it, and probably on `profiling-backend` as well since it expects a label I think)

thomasbertet: Removing it would also need a BE PR to make it optional. Let's keep it for now :)

packages/rum-core/src/domain/action/trackManualActions.ts
MandragoreVR: Oh I didn't think click actions could have a duration, so I thought they would only be sent through `LifeCycleEventType.RAW_RUM_EVENT_COLLECTED`, let me add it 👌

packages/rum/src/domain/profiling/actionHistory.ts
MandragoreVR: got it, just updated, is that what you had in mind?

packages/rum/src/domain/profiling/actionHistory.ts
amortemousque: I’m not convinced by this decision, but at least add a comment explaining why it’s empty ;)

packages/rum-core/src/domain/vital/vitalCollection.ts
MandragoreVR: yeah but I didn't want to allow the customers to be able to set the vital id manually since it would override our internal logic... do you think that's okay to let them do it anyway?

amortemousque: The exposed signature should use `AddDurationVital` not  `DurationVital` so it should be fine
 https://datadoghq.dev/browser-sdk/interfaces/_datadog_browser-rum.DatadogRum.html#adddurationvital

MandragoreVR: that would only be a typescript blocker right? not an actual blocker per se... 

but I'm fine with making the change anyway, just updated 👌

PR #4149, opened 2026-02-06
✅ improve unit test setup

BenoitZugmeyer:
## Motivation

For a long time, our unit test suite did not support to be built as a single bundle -- each .spec.ts files produced a separate bundle with their dependencies, so module-level variables were not shared between spec files.

This constraint caused maintenance issues in the past ([see](https://github.com/codymikol/karma-webpack/issues/587)). It also prevents us from running global cleanups in the "forEach.spec.ts" file.

## Changes

* Make sure all tests are properly cleaned-up on exit
* Remove the `runtimeChunck: false` option so karma-webpack is building mutualizing common dependencies between spec files
* Also, introduce a `--seed` argument so we can run the spec files for a given seed without modifying the config file (useful for letting agents investigating test failures)

## Test instructions

If CI is green, it should be good

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
+++ b/packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
@@ -1,5 +1,5 @@
 import { mockClock, getSessionState, registerCleanupTask } from '../../../../test'
-import { setCookie, deleteCookie, getCookie, getCurrentSite } from '../../../browser/cookie'
+import { setCookie, deleteCookie, getCookie } from '../../../browser/cookie'
 import type { SessionState } from '../sessionState'
 import { validateAndBuildConfiguration } from '../../configuration'
 import type { InitConfiguration } from '../../configuration'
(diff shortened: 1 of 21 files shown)


AGENTS.md
```diff
@@ -24,6 +24,9 @@ yarn test:unit
 # Run specific test file
 yarn test:unit --spec packages/core/src/path/to/feature.spec.ts
 
+# Run tests on a specific seed
+yarn test:unit --seed 123
```
thomas-lebeau, line 28: **👏 praise:** ‏Nice!

packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
```diff
@@ -93,28 +93,31 @@ describe('session in cookie strategy', () => {
...
       },
     ].forEach(({ description, initConfiguration, cookieString }) => {
       it(description, () => {
         const cookieSetSpy = spyOnProperty(document, 'cookie', 'set')
         selectCookieStrategy(initConfiguration)
-        expect(cookieSetSpy.calls.argsFor(0)[0]).toMatch(cookieString)
+        expect(cookieSetSpy).toHaveBeenCalled()
+        for (const call of cookieSetSpy.calls.all()) {
+          expect(call.args[0]).toMatch(cookieString)
+        }
```
BenoitZugmeyer, line 118: **note:** this change was needed because before that PR, the `getCurrentSite` cache had a high chance of being defined when those tests were executed, so cookie accesses to get the current site were not taken into account. Now that the current site cache is properly reset between tests, this test was failing consistently.

packages/core/src/tools/valueHistory.ts
```diff
@@ -143,9 +143,20 @@ export function createValueHistory<Value>({
...
 
   return { add, find, closeActive, findAll, reset, stop }
 }
+
+/**
+ * Reset all global state. This is useful for testing to ensure clean state between tests.
+ *
+ * @internal
+ */
+export function resetValueHistoryGlobals() {
```
thomas-lebeau: **💭 thought:** ‏I wonder if we should have a way to prevent using these function outside of tests. Like using a prefix (i.e. `test_only_resetValueHistoryGlobals`) and a linting rule preventing to call this function outside of `.spec.ts`

PR #4152, opened 2026-02-09
🐛 [RUM Profiler] Fix profiler stuck when session expires

thomasbertet:
## Motivation

The profiler can get stuck in "stopped" status after periods of inactivity, preventing it from collecting profiling data. Two race conditions were identified:

1. **Session expires while profiler is paused:** When the tab is hidden, the profiler pauses. If the session expires during this time, `stopProfilerInstance` returned early without updating the instance state from 'paused' to 'stopped', causing the session renewal check to fail.

2. **Session renews while stop is in progress:** When `SESSION_RENEWED` fires while the async `stopProfiling` was still executing, the check `instance.state === 'stopped'` failed because the state hadn't been updated yet.

### Fix 1: Handle paused state in `stopProfilerInstance`
When `stopProfilerInstance` is called while the profiler is paused, properly transition the state to 'stopped' with the appropriate `stateReason`.

### Fix 2: Sync state changes with fire-and-forget data collection
Instead of using `await` to wait for profiler to stop (with data being sent), we use a sync `stopProfiling` :
- `stopProfilerInstance` and `pauseProfilerInstance` now update state **synchronously**
- Data collection (`collectProfilerInstance`) continues in the background as fire-and-forget
- This eliminates the race condition by design: when `SESSION_RENEWED` fires, the state is already 'stopped'
- Changed `stop()` return type from `Promise<void>` to `void`

This approach is simpler and more robust than the flag-based solution (commit 2 in this PR) because:
- No flags to track or reset
- State transitions are immediate and predictable
- No risk of forgetting to handle edge cases in flag logic

### Unit tests 
Since we changed the `stop` and `stopProfilerInstance` functions to be sync instead of async, and we don't wait for the profiles to be sent, we needed to update the unit tests to account for: 
- `stop` is updating synchronously the status of the profiler. 
- Checking `isStopped` no longer tells you the latest "Profile" was sent. 
So overall we needed to await the requests to make sure it was properly sent. 

## Test instructions

1. Start the profiler
2. Switch to another tab (profiler pauses)
3. Wait for session to expire (15 min inactivity) or manually trigger `SESSION_EXPIRED`
4. Trigger session renewal (or return to tab and interact)
5. Verify profiler restarts

For the race condition test:
1. Fire `SESSION_EXPIRED` immediately followed by `SESSION_RENEWED`
2. Verify profiler restarts immediately (state is already 'stopped' when renewal fires)

Both of theses have unit test to cover it.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/profiler.spec.ts
+++ b/packages/rum/src/domain/profiling/profiler.spec.ts
@@ -19,6 +19,7 @@ import {
   DEFAULT_FETCH_MOCK,
   readFormDataRequest,
   mockClock,
+  waitNextMicrotask,
 } from '@datadog/browser-core/test'
 import { LONG_TASK_ID_HISTORY_TIME_OUT_DELAY } from 'packages/rum-core/src/domain/longTask/longTaskCollection'
 import { createRumSessionManagerMock, mockRumConfiguration, mockViewHistory } from '../../../../rum-core/test'
@@ -135,12 +136,10 @@ describe('profiler', () => {
 
     expect(profilingContextManager.get()?.status).toBe('running')
 
-    // Stop collection of profile.
+    // Stop collection of profile (sync - state changes immediately)
     profiler.stop()
 
-    // Wait for stop of collection.
-    await waitForBoolean(() => profiler.isStopped())
-
+    expect(profiler.isStopped()).toBe(true)
     expect(profilingContextManager.get()?.status).toBe('stopped')
 
     // Wait for data collection to complete (async fire-and-forget)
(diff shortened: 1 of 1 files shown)


packages/rum/src/domain/profiling/profiler.spec.ts
```diff
@@ -238,9 +247,14 @@ describe('profiler', () => {
...
 
     // stop the second profiling session
-    await profiler.stop()
+    profiler.stop()
     await waitForBoolean(() => profiler.isStopped())
 
+    // Data collection uses Promises (microtasks), not setTimeout.
+    // With mockClock(), we can't use waitForBoolean (which polls via setTimeout).
+    // Instead, flush microtasks by awaiting a resolved Promise.
+    await Promise.resolve()
```
BenoitZugmeyer: **suggestion:** use `await waitNextMicrotask()` for expliciteness

PR #4157, opened 2026-02-10
♻️ Make session manager injectable and isolate for v7 rewrite

thomas-lebeau:
## Motivation

Prepare the session manager layer for a from-scratch rewrite in v7. The existing session manager internals (cookies, storage, core session store) are tightly coupled to boot-level tests, making it hard to replace the implementation without breaking unrelated tests.

This PR decouples all non-session-manager code from session internals, so the session manager can be rewritten in isolation.

### 1. Injectable session manager factory

Move the stub-vs-real session manager branching from `createPreStartStrategy` to the caller (`rumPublicApi` / `logsPublicApi`) via a factory parameter:

```mermaid
graph LR
    A[rumPublicApi / logsPublicApi] -->|ternary: bridge ? stub : real| B[factory]
    B -->|passed as param| C[createPreStartStrategy]
    C -->|calls factory| D[session manager instance]
```

- **BREAKING CHANGE:** When no storage strategy is available (cookies/localStorage unavailable), neither SDK will send data. Previously Logs would still start without storage — now it requires explicit `sessionPersistence: 'memory'` to work without storage. In worker environments, Logs uses the real session manager with memory persistence (we can revisit that when we star using CookieStore as this seems supported in service workers). RUM is explicitly blocked with a warning.
- `createPreStartStrategy` receives one factory, just calls it — no branching logic
- Callers use: `canUseEventBridge() ? stub : real`

### 2. Prove isolation via gut-then-restore

To verify that all non-session-manager tests use mocks (not real implementations), the session manager bodies were temporarily gutted and their unit tests skipped (`xdescribe`). All 3071 remaining tests passed with zero failures, proving the isolation is complete. The implementations and tests were then fully restored.

### 3. Mock infrastructure

- Boot tests (`preStartRum.spec`, `preStartLogs.spec`) inject mock factories — no more `stopSessionManager()` cleanup or cookie manipulation
- Public API tests (`rumPublicApi.spec`, `logsPublicApi.spec`) inject mock session managers via optional parameter on `makeRumPublicApi` / `makeLogsPublicApi`
- `mockRumSessionManager` / `mockLogsSessionManager` updated with factory wrappers

### 4. Worker environment support

Use a defense-in-depth strategy for worker environments, handling workers at multiple layers:

- **Storage layer**: `selectSessionStoreStrategyType` forces `SessionPersistence.MEMORY` in workers — cookies and localStorage are unavailable in worker contexts
- **Session manager**: Skip DOM event tracking (`trackActivity`, `trackVisibility`, `trackResume`) in workers — no DOM events to listen to
- **RUM**: Block entirely in `preStartRum` with an explicit warning — RUM features (DOM observation, user interactions) are not applicable in workers
- **BREAKING CHANGE:** Logs Uses the real session manager (not a stub) — the storage and session manager layers handle the worker context, so Logs works normally with memory-backed sessions. this means that now logs will have a proper session Id

New exports from `@datadog/browser-core`: `selectSessionStoreStrategyType`, `SessionStoreStrategyType`
`startSessionManager` now takes `sessionStoreStrategyType` as a direct parameter.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/sessionStore.ts
+++ b/packages/core/src/domain/session/sessionStore.ts
@@ -48,19 +48,7 @@ export const STORAGE_POLL_DELAY = ONE_SECOND
 export function selectSessionStoreStrategyType(
   initConfiguration: InitConfiguration
 ): SessionStoreStrategyType | undefined {
-  let sessionPersistence = initConfiguration.sessionPersistence
-
-  /**
-   * Force memory persistence in worker environment since cookie and localStorage are not available
-   * TODO: make it work when we start using Cookie Store API
-   *
-   * @see https://developer.mozilla.org/en-US/docs/Web/API/CookieStore
-   */
-  if (isWorkerEnvironment) {
-    sessionPersistence = SessionPersistence.MEMORY
-  }
-
-  const persistenceList = normalizePersistenceList(sessionPersistence, initConfiguration)
+  const persistenceList = normalizePersistenceList(initConfiguration.sessionPersistence, initConfiguration)
 
   for (const persistence of persistenceList) {
     const strategyType = selectStrategyForPersistence(persistence, initConfiguration)
@@ -82,6 +70,13 @@ function normalizePersistenceList(
 
   if (sessionPersistence !== undefined) {
     return [sessionPersistence]
+  }
+
+  // In worker environments, default to memory since cookie and localStorage are not available
+  // TODO: make it work when we start using Cookie Store API
+  // @see https://developer.mozilla.org/en-US/docs/Web/API/CookieStore
+  if (isWorkerEnvironment) {
+    return [SessionPersistence.MEMORY]
   }
 
   // Legacy default behavior: cookie first, with optional localStorage fallback


packages/logs/src/domain/logsSessionManager.ts
```diff
@@ -46,13 +46,15 @@ export function startLogsSessionManager(
...
-  const session = isTracked ? {} : undefined
-  return {
-    findTrackedSession: () => session,
+export function startLogsSessionManagerStub(
+  _configuration: LogsConfiguration,
+  _trackingConsentState: TrackingConsentState,
+  onReady: (sessionManager: LogsSessionManager) => void
+): void {
+  onReady({
+    findTrackedSession: () => ({}),
```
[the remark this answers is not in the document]
thomas-lebeau: Good catch. Sampling decision should be handled by the other side of bridge (i.e.  the mobile SDK), similarly than RUM

packages/core/src/domain/session/sessionStore.ts
```diff
@@ -47,7 +48,17 @@ export const STORAGE_POLL_DELAY = ONE_SECOND
...
+
+  /**
+   * Force memory persistence in worker environment since cookie and localStorage are not available
+   * TODO: make it work when we start using Cookie Store API
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/CookieStore
+   */
+  if (isWorkerEnvironment) {
+    sessionPersistence = SessionPersistence.MEMORY
+  }
```
BenoitZugmeyer: **suggestion:** maybe don't force it, but only adjust the default?

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -100,6 +100,11 @@ export function createPreStartStrategy(
 
       cachedConfiguration = configuration
 
+      if (!canUseEventBridge() && !configuration.sessionStoreStrategyType) {
+        display.warn('No storage available for session. We will not send any data.')
+        return
+      }
```
BenoitZugmeyer: **suggestion:** Move this warning in the session manager?

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -111,15 +116,12 @@ export function createPreStartStrategy(
...
         mockable(startTelemetry)(TelemetryService.LOGS, configuration, hooks)
-        if (configuration.sessionStoreStrategyType && !canUseEventBridge() && !willSyntheticsInjectRum()) {
-          startLogsSessionManager(configuration, trackingConsentState, (newSessionManager) => {
-            sessionManager = newSessionManager
-            tryStartLogs()
-          })
-        } else {
-          sessionManager = startLogsSessionManagerStub(configuration)
+        const startSessionManagerFn =
+          canUseEventBridge() || isWorkerEnvironment ? startLogsSessionManagerStub : mockable(startLogsSessionManager)
```
BenoitZugmeyer: **question:** Is the `isWorkerEnvironment` check here still useful since session manager now supports workers?

BenoitZugmeyer (approved):

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -105,13 +111,15 @@ export function startSessionManager<TrackingType extends string>(
...
     })
 
-    trackActivity(configuration, () => {
-      if (trackingConsentState.isGranted()) {
-        sessionStore.expandOrRenewSession()
-      }
-    })
-    trackVisibility(configuration, () => sessionStore.expandSession())
-    trackResume(configuration, () => sessionStore.restartSession())
+    if (!isWorkerEnvironment) {
```
mormubis, line 114: Enabling the session manager to work on service workers seem to increase the complexity.


**💬 suggestion:** Should we move the `trackActivity`/`trackVisibility`/`trackResume` to an upper level, an call `sessionManager.expandOrRenewSession`?‏

thomas-lebeau: The complexity was already here, it has just been moved from `preStartRum` and `preStartLogs` to the session Manager (we were using a stub session Manager instead of the real one)

Moving that back one level up will risk duplicating the logic (LOGS vs RUM, even though RUM does not work in worker env). I think it's fine, so far it's just one if statement but we will need to keep an eye on this.

In v8, is we manage to reconcile session manager and only have one instance, we might want to do differently then.

I will keep note on this though, maybe we can be smarter and guard by feature instead of `isWorkerEnvironment` (i.e. `document` is present,...)

PR #4162, opened 2026-02-11
♻️ Refactor e2e service worker setup into a composable builder pattern

thomas-lebeau:
## Motivation

The e2e service worker test infrastructure was tightly coupled and hard to extend. Adding new SDK configurations (e.g. RUM in a worker) required modifying multiple options flags (`importScript`, `nativeLog`) and threading them through the test framework. The `interactWithWorker` API forced tests to go through `postMessage` with the active service worker, making test code indirect and harder to read.

## Changes

- **New `createWorker()` builder** (`test/e2e/lib/framework/createWorker.ts`): Composable builder pattern that mirrors the existing `createTest()` API. Workers can be configured with `.withRum()` and `.withLogs()` independently, replacing the previous boolean flags.
- **Replace `interactWithWorker` with `evaluateInWorker`**: Instead of sending messages to the worker via `postMessage`, `evaluateInWorker` sends code as a string and executes it via `new Function()` inside the worker. This allows tests to call SDK APIs (e.g. `DD_LOGS.logger.log(...)`) directly in the worker scope.
- **Dynamic worker script generation**: `workerSetup()` in `pageSetups.ts` now conditionally includes Logs and/or RUM SDK imports based on the worker configuration, and always registers a message listener for the `evaluate` protocol.
- **More explicit tests**: Each test now specifies its own Logs/RUM configuration, so options like `forwardConsoleLogs` are only enabled in the test that actually uses `console.log`, making the intent clearer.
- **New RUM-in-worker tests** (`test/e2e/scenario/rum/init.scenario.ts`): Verifies that initializing RUM inside a service worker gracefully warns about missing session storage instead of crashing.
- **Updated existing logs worker tests** to use the new `createWorker().withLogs()` builder and `evaluateInWorker` API.

## Test instructions

```bash
yarn test:e2e:init
yarn test:e2e -g "service worker"
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/scenario/rum/init.scenario.ts
+++ b/test/e2e/scenario/rum/init.scenario.ts
@@ -323,7 +323,9 @@ test.describe('Service workers Rum', () => {
       test.skip(browserName !== 'chromium', 'Non-Chromium browsers do not support ES modules in Service Workers')
 
       await flushEvents()
-      expectNoRumEventsWithSessionWarning(intakeRegistry, withBrowserLogs)
+
+      expect(intakeRegistry.rumEvents).toHaveLength(0)
+      withBrowserLogs((logs) => expectNoSessionWarning(browserName, logs))
     })
 
   createTest('service worker with worker rum - importScripts')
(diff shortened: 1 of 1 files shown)


test/e2e/lib/framework/createWorker.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface WorkerOptions {
+  importScripts?: boolean
+}
+
+export function createWorker(options: WorkerOptions = {}) {
+  return new Worker(options)
+}
+
+export class Worker {
```
mormubis, line 13:
```suggestion
export class TestWorker {
```

Just in case clashes with the TypeScript Worker definition.

test/e2e/lib/framework/createTest.ts
```diff
@@ -362,8 +347,22 @@ function createTestContext(
         browserLogsManager.clear()
       }
     },
-    interactWithWorker: async (cb: (worker: ServiceWorker) => void) => {
-      await page.evaluate(`(${cb.toString()})(window.myServiceWorker.active)`)
+    evaluateInWorker: async (fn: () => void) => {
+      await page.evaluate(async (code) => {
+        const { active, installing, waiting } = window.myServiceWorker
+        const worker = active ?? (await waitForActivation(installing ?? waiting!))
+        worker.postMessage({ __type: 'evaluate', code })
```
mormubis, line 354: Not my cup of tea, but if it works I'm ok with it. Evaluating code is always confusing and if there are any errors is difficult to debug when they are happening.

PR #4163, opened 2026-02-11
♻️ [PANA-5945] Support child list InsertionCursors

sethfowler-datadog:
## Motivation

We've already landed support for recording full snapshots as Change records behind an experiment flag. The next step is to introduce support for incremental snapshots, as well. This requires making improvements to a number of different aspects of the recording code, so this PR is the first in a series that will work toward the goal of supporting incremental snapshots.

## Changes

The Change record version of the DOM serialization code tracks its position in the DOM using an `InsertionCursor`. The `InsertionCursor`'s role is to convert the insertion position of each serialized node into an encoded form used by Change records; any location in the DOM can be specified using a single integer.

Today, the only factory function that exists for `InsertionCursor` objects creates them initially positioned at the root of the DOM. That's all we needed to support full snapshots. However, for incremental snapshots, we need to start serialization at an arbitrary point in the DOM; in other words, we need to be able to create an  `InsertionCursor` that initially points to a position in the child list of some existing, already-serialized DOM node. This allows us to use these `InsertionCursor`s to serialize new DOM subtrees as they are added.

This PR adds a factory function for `InsertionCursor`s called `createChildInsertionCursor()` that positions the new `InsertionCursor` within an existing child list. It also updates the tests to test this new functionality.

### DIFF
--- a/packages/rum/src/domain/record/serialization/insertionCursor.spec.ts
+++ b/packages/rum/src/domain/record/serialization/insertionCursor.spec.ts
@@ -1,15 +1,13 @@
 import type { InsertionPoint } from '../../../types'
-import type { NodeId } from '../itemIds'
+import type { NodeId, NodeIds } from '../itemIds'
 import { createNodeIds } from '../itemIds'
-import type { InsertionCursor } from './insertionCursor'
-import { createRootInsertionCursor } from './insertionCursor'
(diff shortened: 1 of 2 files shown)


gonzalezreal (approved):
LGTM 🚀

PR #4164, opened 2026-02-11
♻️ [PANA-5948] Make some small improvements to ItemId management

sethfowler-datadog:
## Motivation

We've already landed support for recording full snapshots as Change records behind an experiment flag. The next step is to introduce support for incremental snapshots, as well. This requires making improvements to a number of different aspects of the recording code. This PR is the second in a series that will work toward the goal of supporting incremental snapshots; for the previous PR, see #4163.

## Changes

This PR makes some small improvements to the code surrounding the `ItemIds` interface, which is used to track the identifiers assigned to various kinds of DOM-related objects during recording.

In particular:
1. A `delete()` method has been added, making it possible to delete an item id. This allows you to reinsert the same item and assign it a new id, which is something that the new serialization algorithm requires.
2. A `nextId` getter has been added, making it possible to determine what the next item id will be before actually assigning it. This, too, is helpful for implementing the new serialization algorithm.
3. The existing `clear()` method now does nothing if no ids have been assigned. This is just a small performance optimization.

Related code has been updated:
1. In `startFullSnapshots.ts`, we now always call `scope.resetIds()` when generating a Change record, regardless of whether we're generating the initial full snapshot or a subsequent one. We avoided calling it for the first full snapshot as a small performance optimization, but now `ItemIds#clear()` avoids the unnecessary work automatically.
2. In `recordingScope.ts`, the implementation of `scope.resetIds()` now calls `clear()` on the `ItemIds` instances that are _currently_ stored on the `RecordingScope`, instead of just the ones that were passed in when the `RecordingScope` was constructed. This is helpful for testing the new serialization algorithm, since it allows us to wrap these `ItemIds` instance in test code with version that check assertions.
3. Tests have been added for the new methods. (The behavior change in `clear()` is not visible, so there's no test, but the existing tests continue to pass.)

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/itemIds.spec.ts
+++ b/packages/rum/src/domain/record/itemIds.spec.ts
@@ -51,6 +51,33 @@ describe('ItemIds', () => {
         })
       })
 
+      describe('delete', () => {
+        it('allows an item to be re-inserted and receive a new id', () => {
+          const item = createItem()
+          expect(itemIds.getOrInsert(item)).toBe(firstId)
+          expect(itemIds.getOrInsert(item)).toBe(firstId)
+
+          itemIds.delete(item)
+
+          expect(itemIds.getOrInsert(item)).toBe((firstId + 1) as ItemId)
+          expect(itemIds.getOrInsert(item)).toBe((firstId + 1) as ItemId)
+        })
+
+        it('does not change the next assigned id', () => {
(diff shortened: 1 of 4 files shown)


Tizi42 (approved):
Looks good. The PR description was helpful, thanks!

PR #4168, opened 2026-02-12
✅ fix flaky test

BenoitZugmeyer:
## Motivation

`viewportObservable.spec.ts` changes the page to add scrollbars, causing some asynchronous events to be fired *after* the test has completed. If `recorder.spec.ts` is executed just after, its test might fail because it'll record those asynchronous events.

This is the same issue we had with `scroll.spec.ts`. In the past, we tried to be smart and predict which events will be triggered, and wait for them. But it's a bit complex and not reliable on every browsers.

## Changes

This commit simplifies this workaround by waiting for two frames. When scrolling/changing viewport, the browser will first paint it, and then emit events. The second frame should reliably come after events are emitted (tested in Chrome, Firefox and Safari).

## Test instructions

Reproduce flaky test:

```
git checkout 56a15f139dec81f010396e6a32e172b2291efcc4
yarn test:unit --seed 26598
```

Confirm that it's gone when running the tests with the same seed after applying this PR changes

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/test/registerCleanupTask.ts
+++ b/packages/core/test/registerCleanupTask.ts
@@ -1,9 +1,13 @@
-const cleanupTasks: Array<() => void> = []
+type CleanupTask = () => unknown
 
-export function registerCleanupTask(task: () => void) {
+const cleanupTasks: CleanupTask[] = []
+
+export function registerCleanupTask(task: CleanupTask) {
   cleanupTasks.unshift(task)
 }
 
-afterEach(() => {
-  cleanupTasks.splice(0).forEach((task) => task())
+afterEach(async () => {
+  for (const task of cleanupTasks.splice(0)) {
+    await task()
+  }
 })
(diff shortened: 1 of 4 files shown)


BeltranBulbarellaDD (approved):

packages/core/test/registerCleanupTask.ts
```diff
@@ -1,9 +1,13 @@
-const cleanupTasks: Array<() => void> = []
+type CleanupTask = () => unknown
```
thomas-lebeau, line 1: **🥜 nitpick:** ‏
```suggestion
type CleanupTask = () => void | Promise<void>
```

thomas-lebeau (approved):

packages/core/test/registerCleanupTask.ts
BenoitZugmeyer: That was my initial thought, but then it lead to a typechecking issue https://gitlab.ddbuild.io/DataDog/browser-sdk/-/jobs/1420187767

Sooo... I just used unknown because we don't really care of the result?

thomas-lebeau: yeah, it's fine

PR #4180, opened 2026-02-13
[RUM-14619] Add setViewLoadingTime() public API

MaelNamNam:
## Motivation

Allow developers to manually report view loading time for complex async patterns (SPAs with deferred fetching, skeleton screens) where auto-detection is inaccurate.

Brings Browser SDK to parity with iOS/Android SDKs.

## Changes

New experimental public API: `DD_RUM.setViewLoadingTime()`

- Marked as `[Experimental]` via JSDoc annotation (no runtime feature flag required)
- Loading time = elapsed since current view start
- First call suppresses auto-detected loading time
- Last call always wins
- Pre-start buffering supported (calls before `init()` are drained with preserved timestamps)
- Works with `startView()` / `manuallyTrackViews` — timer resets for each new view
- Telemetry sends `feature: 'addViewLoadingTime'` only (browser omits `overwritten`, `no_view`, and `no_active_view` — kept optional in schema for mobile compat)

**Note:** The `enableExperimentalFeatures: ['set_view_loading_time']` runtime gate has been removed. The API is available without any flag, consistent with other experimental APIs (`startAction`, `stopAction`, `startResource`, `stopResource`).

**Upstream dep:** [rum-events-format#352](https://github.com/DataDog/rum-events-format/pull/352) has been merged, submodule updated to `5ddfd3b`, types regenerated, and `[MANUAL]` annotation removed from this PR. `AddViewLoadingTime` is now auto-generated from the schema.

## Test instructions

```bash
yarn test:unit --spec packages/rum-core/src/domain/view/trackViews.spec.ts
yarn test:unit --spec packages/rum-core/src/boot/rumPublicApi.spec.ts
yarn test:unit --spec packages/rum-core/src/boot/preStartRum.spec.ts
yarn test:e2e:init && yarn test:e2e -g "manual loading time"
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/boot/preStartRum.spec.ts
+++ b/packages/rum-core/src/boot/preStartRum.spec.ts
@@ -613,9 +613,9 @@ describe('preStartRum', () => {
       doStartRumSpy.and.returnValue({ setLoadingTime: setLoadingTimeSpy } as unknown as StartRumResult)
 
       const timestamp = 123 as TimeStamp
-      strategy.setLoadingTime(timestamp, false)
+      strategy.setLoadingTime(timestamp)
       strategy.init(DEFAULT_INIT_CONFIGURATION, PUBLIC_API)
-      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(timestamp, false)
+      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(timestamp)
     })
 
     it('setLoadingTime should preserve call timestamp', () => {
@@ -629,7 +629,7 @@ describe('preStartRum', () => {
       clock.tick(20)
       strategy.init(DEFAULT_INIT_CONFIGURATION, PUBLIC_API)
 
-      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(jasmine.any(Number), false)
+      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(jasmine.any(Number))
       // Verify the timestamp was captured at call time (tick 10), not at drain time (tick 30)
       const capturedTimestamp = setLoadingTimeSpy.calls.argsFor(0)[0] as number
       expect(capturedTimestamp).toBe(clock.timeStamp(10))
--- a/packages/rum-core/src/boot/preStartRum.ts
+++ b/packages/rum-core/src/boot/preStartRum.ts
@@ -242,8 +242,8 @@ export function createPreStartStrategy(
       bufferApiCalls.add((startRumResult) => startRumResult.addTiming(name, time))
     },
 
-    setLoadingTime: ((callTimestamp: TimeStamp, overwrite = false) => {
-      bufferApiCalls.add((startRumResult) => startRumResult.setLoadingTime(callTimestamp, overwrite))
+    setLoadingTime: ((callTimestamp: TimeStamp) => {
+      bufferApiCalls.add((startRumResult) => startRumResult.setLoadingTime(callTimestamp))
     }) as Strategy['setLoadingTime'],
 
     startView(options, startClocks = clocksNow()) {
--- a/packages/rum-core/src/boot/rumPublicApi.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.ts
@@ -68,13 +68,6 @@ export interface StartRecordingOptions {
   force: boolean
 }
 
-export interface SetViewLoadingTimeOptions {
-  /**
-   * Set to `true` to replace a previously set loading time.
-   */
-  overwrite?: boolean
-}
-
 /**
  * Public API for the RUM browser SDK.
  *
@@ -248,13 +241,11 @@ export interface RumPublicApi extends PublicApi {
    * [Experimental] Manually set the current view's loading time.
    *
    * Call this method when the view has finished loading. The loading time is computed as the
-   * elapsed time since the view started. By default, the first call sets the loading time and
-   * subsequent calls are no-ops. Use `{ overwrite: true }` to replace a previously set value.
+   * elapsed time since the view started. Each call replaces any previously set value (last-call-wins).
    *
    * @category Data Collection
-   * @param options - Options. Set `overwrite: true` to replace a previously set loading time.
    */
-  setViewLoadingTime: (options?: SetViewLoadingTimeOptions) => void
+  setViewLoadingTime: () => void
 
   /**
    * Set the global context information to all events, stored in `@context`
@@ -794,9 +785,9 @@ export function makeRumPublicApi(
       strategy.addTiming(sanitize(name)!, time as RelativeTime | TimeStamp | undefined)
     }),
 
-    setViewLoadingTime: monitor((options?: SetViewLoadingTimeOptions) => {
+    setViewLoadingTime: monitor(() => {
       const callTimestamp = timeStampNow()
-      strategy.setLoadingTime(callTimestamp, options?.overwrite ?? false)
+      strategy.setLoadingTime(callTimestamp)
       addTelemetryUsage({
         feature: 'addViewLoadingTime',
       })
--- a/packages/rum-core/src/domain/view/trackViews.ts
+++ b/packages/rum-core/src/domain/view/trackViews.ts
@@ -179,8 +179,7 @@ export function trackViews(
     addTiming: (name: string, time: RelativeTime | TimeStamp = timeStampNow()) => {
       currentView.addTiming(name, time)
     },
-    setLoadingTime: (callTimestamp?: TimeStamp, overwrite?: boolean) =>
-      currentView.setLoadingTime(callTimestamp, overwrite),
+    setLoadingTime: (callTimestamp?: TimeStamp) => currentView.setLoadingTime(callTimestamp),
     startView: (options?: ViewOptions, startClocks?: ClocksState) => {
       currentView.end({ endClocks: startClocks })
       currentView = startNewView(ViewLoadingType.ROUTE_CHANGE, startClocks, options)
--- a/packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
+++ b/packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
@@ -82,11 +82,8 @@ export function trackCommonViewMetrics(
       commonViewMetrics.interactionToNextPaint = getInteractionToNextPaint()
       return commonViewMetrics
     },
-    setLoadingTime: (callTimestamp?: TimeStamp, overwrite = false) => {
+    setLoadingTime: (callTimestamp?: TimeStamp) => {
       if (viewEnded) {
-        return
-      }
-      if (hasManualLoadingTime && !overwrite) {
         return
       }
       const loadingTime = elapsed(viewStart.timeStamp, callTimestamp ?? timeStampNow())
--- a/test/e2e/scenario/rum/views.scenario.ts
+++ b/test/e2e/scenario/rum/views.scenario.ts
@@ -87,7 +87,7 @@ test.describe('rum views', () => {
         expect(viewEvent!.view.loading_time).toBeGreaterThanOrEqual(200 * 1e6)
       })
 
-    createTest('overwrites manual loading time when called with overwrite option')
+    createTest('overwrites manual loading time on subsequent calls (last-call-wins)')
       .withRum()
       .withBody(SPINNER)
       .run(async ({ flushEvents, intakeRegistry, page }) => {
@@ -99,7 +99,7 @@ test.describe('rum views', () => {
               }, 200)
 
               setTimeout(() => {
-                window.DD_RUM!.setViewLoadingTime({ overwrite: true })
+                window.DD_RUM!.setViewLoadingTime()
                 resolve()
               }, 500)
             })
@@ -108,7 +108,7 @@ test.describe('rum views', () => {
         await flushEvents()
         const viewEvent = intakeRegistry.rumViewEvents.at(-1)
         expect(viewEvent).toBeDefined()
-        // Should reflect the second (overwritten) value (~500ms), not the first (~200ms)
+        // Should reflect the second value (~500ms), not the first (~200ms)
         expect(viewEvent!.view.loading_time).toBeGreaterThanOrEqual(500 * 1e6)
       })
 
--- a/packages/rum-core/src/boot/rumPublicApi.spec.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.spec.ts
@@ -569,20 +569,12 @@ describe('rum public api', () => {
       }))
     })
 
-    it('should call setLoadingTime with timestamp and no overwrite by default', () => {
+    it('should call setLoadingTime with timestamp', () => {
       rumPublicApi.init(DEFAULT_INIT_CONFIGURATION)
 
       rumPublicApi.setViewLoadingTime()
 
-      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(jasmine.any(Number), false)
-    })
-
-    it('should pass overwrite true when specified', () => {
-      rumPublicApi.init(DEFAULT_INIT_CONFIGURATION)
-
-      rumPublicApi.setViewLoadingTime({ overwrite: true })
-
-      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(jasmine.any(Number), true)
+      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(jasmine.any(Number))
     })
 
     it('should not throw when called before init', () => {
--- a/packages/rum-core/src/domain/view/trackViews.spec.ts
+++ b/packages/rum-core/src/domain/view/trackViews.spec.ts
@@ -807,7 +807,7 @@ describe('manual loading time', () => {
     expect(lastUpdate.commonViewMetrics.loadingTime).toBe(clock.relative(500))
   })
 
-  it('should not overwrite loading time by default on subsequent calls', () => {
+  it('should overwrite loading time on subsequent calls (last-call-wins)', () => {
     const { getViewUpdate, getViewUpdateCount, setLoadingTime } = viewTest
 
     clock.tick(100)
@@ -818,24 +818,6 @@ describe('manual loading time', () => {
 
     clock.tick(200)
     setLoadingTime()
-
-    clock.tick(THROTTLE_VIEW_UPDATE_PERIOD)
-
-    const lastUpdate = getViewUpdate(getViewUpdateCount() - 1)
-    expect(lastUpdate.commonViewMetrics.loadingTime).toBe(firstValue)
-  })
-
-  it('should overwrite loading time when overwrite is true', () => {
-    const { getViewUpdate, getViewUpdateCount, setLoadingTime } = viewTest
-
-    clock.tick(100)
-    setLoadingTime()
-
-    clock.tick(THROTTLE_VIEW_UPDATE_PERIOD)
-    const firstValue = getViewUpdate(getViewUpdateCount() - 1).commonViewMetrics.loadingTime
-
-    clock.tick(200)
-    setLoadingTime(undefined, true)
 
     clock.tick(THROTTLE_VIEW_UPDATE_PERIOD)
 
@@ -891,7 +873,7 @@ describe('manual loading time', () => {
     expect(lastUpdate.commonViewMetrics.loadingTime).toBe(manualValue)
   })
 
-  it('should start with clean overwrite state on new view', () => {
+  it('should start with clean loading time state on new view', () => {
     const { getViewUpdate, getViewUpdateCount, startView, setLoadingTime } = viewTest
 
     clock.tick(100)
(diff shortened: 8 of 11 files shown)


packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -378,6 +383,19 @@ function newView(
       customTimings[sanitizeTiming(name)] = relativeTime
       scheduleViewUpdate()
     },
+    addLoadingTime(callTimestamp?: TimeStamp, overwrite = false) {
+      if (endClocks) {
+        return { no_active_view: true as const }
```
thomas-lebeau: **🔨 warning:** ‏These are not following the standard naming convention. we use camelCase and convert what need to be to pascal_case just before sending the event.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
```diff
@@ -74,5 +77,26 @@ export function trackCommonViewMetrics(
       commonViewMetrics.interactionToNextPaint = getInteractionToNextPaint()
       return commonViewMetrics
     },
+    setManualLoadingTime: (loadingTime: Duration) => {
+      const hadPreviousManual = isLoadingTimeSetManually
+      const hadAutoValue = !isLoadingTimeSetManually && commonViewMetrics.loadingTime !== undefined
+
+      // Stop auto-detection entirely on first manual call (CONTEXT decision)
```
thomas-lebeau: **❓ question:** ‏what does `(CONTEXT decision)` means?

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
```diff
@@ -74,5 +77,26 @@ export function trackCommonViewMetrics(
...
+      if (!isLoadingTimeSetManually) {
+        stopLoadingTimeTracking()
+      }
+
+      isLoadingTimeSetManually = true
+      commonViewMetrics.loadingTime = loadingTime
+
+      // Return overwrite source for Phase 2 telemetry
```
thomas-lebeau: **❓ question:** ‏what's Phase 2 telemetry?

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -727,6 +743,20 @@ export function makeRumPublicApi(
...
+      const result = strategy.addLoadingTime(callTimestamp, options?.overwrite ?? false)
+
+      // For pre-start buffered calls, result is undefined so we emit best-guess default values.
+      const isResult = result && typeof result === 'object'
+      addTelemetryUsage({
+        feature: 'addViewLoadingTime',
+        no_view: false,
+        no_active_view: isResult && 'no_active_view' in result,
+        overwritten: isResult && 'overwritten' in result && result.overwritten !== false,
+      })
```
thomas-lebeau, line 793: **❓ question:** TelemetryUsage is meant to understand which API the customer are using and with which option when available. I don't think we should return run-time analysis like this.


```suggestion
      addTelemetryUsage({
        feature: 'addViewLoadingTime',
        overwrite
      })
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -215,6 +216,20 @@ export interface RumPublicApi extends PublicApi {
...
   addTiming: (name: string, time?: number) => void
 
+  /**
+   * Manually set the current view's loading time.
+   *
+   * Call this method when the view has finished loading. The loading time is computed as the
+   * elapsed time since the view started. By default, the first call sets the loading time and
+   * subsequent calls are no-ops. Use `{ overwrite: true }` to replace a previously set value.
+   *
+   * See [Override RUM View Loading Time](https://docs.datadoghq.com/real_user_monitoring/browser/advanced_configuration/) for further information.
```
thomas-lebeau: **🔨 warning:** "Override RUM View Loading Time" does not exist at that link ‏

packages/core/src/domain/telemetry/telemetryEvent.types.ts
```diff
@@ -532,11 +532,13 @@ export type TelemetryCommonFeaturesUsage =
 /**
  * Schema of browser specific features usage
  */
+// [MANUAL] AddViewLoadingTime added pending upstream rum-events-format schema sync
```
thomas-lebeau: **❓ question:** ‏I did not find any PR related to that in rum-event-format

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -727,6 +743,20 @@ export function makeRumPublicApi(
       strategy.addTiming(sanitize(name)!, time as RelativeTime | TimeStamp | undefined)
     }),
 
+    addViewLoadingTime: monitor((options?: { overwrite?: boolean }) => {
+      const callTimestamp = timeStampNow()
+      const result = strategy.addLoadingTime(callTimestamp, options?.overwrite ?? false)
+
+      // For pre-start buffered calls, result is undefined so we emit best-guess default values.
+      const isResult = result && typeof result === 'object'
```
thomas-lebeau: **💬 suggestion:** ‏use Types to prevent runtime assertion

packages/core/src/domain/telemetry/telemetryEvent.types.ts
MaelNamNam: Opened: https://github.com/DataDog/rum-events-format/pull/352

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
MaelNamNam: Woops, sorry, it was an internal note reminding why we keep auto-detection running instead of disabling it. I removed it in in 1bb569eff.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
MaelNamNam: Sorry again, it was a reference to our internal development phasing (telemetry wiring came after the core logic). I removed in 1bb569eff.

packages/rum-core/src/boot/rumPublicApi.ts
MaelNamNam: Removed the broken link in 1bb569eff.

packages/rum-core/src/boot/rumPublicApi.ts
MaelNamNam: Good point. I went a different direction though — I aligned the telemetry with the mobile SDK schema instead, which tracks runtime state (`no_view`, `no_active_view`, `overwritten`). This matches what iOS/Android already report for `AddViewLoadingTime`. See c25c762fd.

packages/rum-core/src/boot/rumPublicApi.ts
MaelNamNam: Kept the runtime assertion — it feeds the `no_active_view` and `overwritten` telemetry fields that are now aligned with the mobile SDK schema. See c25c762fd.

packages/core/src/domain/telemetry/telemetryEvent.types.ts
MaelNamNam: Opened as https://github.com/DataDog/rum-events-format/pull/352. Moved `AddViewLoadingTime` to the common schema (shared with mobile) with `no_view`, `no_active_view`, `overwritten` fields.

packages/rum-core/src/domain/view/trackViews.ts
MaelNamNam: Fixed in 46d9511ab — renamed to camelCase internally (`noActiveView`, `overwritten`), conversion to snake_case now happens at the `addTelemetryUsage` boundary in rumPublicApi.ts.

.gitignore
```diff
@@ -33,3 +33,4 @@ playwright-report/
 # Claude Code local files
 *.local.md
 .claude/settings.local.json
+.planning/
```
bcaudan: **💬 suggestion:** ‏as everyone could choose different locations for where they store plans, it could be preferable to configure a global gitignore file with your own preferences rather than adding it to all the repositories you contribute to.

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -771,6 +792,15 @@ export function makeRumPublicApi(
       strategy.addTiming(sanitize(name)!, time as RelativeTime | TimeStamp | undefined)
     }),
 
+    setViewLoadingTime: monitor((options?: SetViewLoadingTimeOptions) => {
+      const callTimestamp = timeStampNow()
+      const result = strategy.setLoadingTime(callTimestamp, options?.overwrite ?? false)
+      addTelemetryUsage({
+        feature: 'addViewLoadingTime',
+        overwritten: result?.overwritten ?? false,
```
bcaudan: **❗ warning:** ‏‏preStartStrategy does not return anything so `result` is always undefined if the API is called before init.

cf https://github.com/DataDog/rum-events-format/pull/352#issuecomment-3998963616, unless we really want to monitor the actual behavior of this feature, I'd be in favor of: 
- only monitoring the API and API option usage
- drop the code computing and returning the overwritten result

sandbox/set-view-loading-time.html
bcaudan, line 1: **💬 suggestion:** ‏We usually don't commit specific feature sandbox files, it would be a burden to maintain.

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -376,6 +380,20 @@ function newView(
...
+      if (hasManualLoadingTime && !overwrite) {
+        return { overwritten: false }
+      }
+      const overwritten = hasManualLoadingTime
+      const loadingTime = elapsed(startClocks.timeStamp, callTimestamp ?? timeStampNow())
+      setManualLoadingTime(loadingTime)
+      hasManualLoadingTime = true
+      scheduleViewUpdate()
+      return { overwritten }
+    },
```
bcaudan: **💬 suggestion:** scattering the logic between `trackViews` and `commonViewMetrics` make things harder to understand (in an already quite complex spot 😬)
We should probably:
- move all the needed logic to `commonViewMetrics`, as it is where loading time logic is already orchestrated
- enrich the exposed `setViewEnd` to notify the end of the view for both manual loading time and INP

.gitignore
MaelNamNam: Good point. It's done

sandbox/set-view-loading-time.html
MaelNamNam: Noted, thanks. Removed in 712ed76e4.

packages/rum-core/src/domain/view/trackViews.ts
MaelNamNam: Done. All manual loading time business logic (state, guards, elapsed calculation) moved into 
`trackCommonViewMetrics`. `trackViews.setLoadingTime` is now a 3-line forwarder, and `setViewEnd` is enriched to handle both INP and manual loading time end-of-view notification. See commits be0997573 and def5efcae.

packages/rum-core/src/boot/rumPublicApi.ts
MaelNamNam: Fixed!

packages/core/src/domain/telemetry/telemetryEvent.types.ts
```diff
@@ -948,21 +950,15 @@ export interface StopResource {
 }
 export interface AddViewLoadingTime {
```
mormubis: Since you're already hand-editing this block under `[MANUAL]`, worth renaming the interface to `SetViewLoadingTime` to match the feature string (`'set-view-loading-time'`) and the public API (`setViewLoadingTime`). `AddViewLoadingTime` is a leftover from the old name.

packages/rum-core/src/boot/preStartRum.spec.ts
```diff
@@ -608,6 +608,33 @@ describe('preStartRum', () => {
...
+      clock.tick(10)
+      strategy.setLoadingTime()
+
+      clock.tick(20)
+      strategy.init(DEFAULT_INIT_CONFIGURATION, PUBLIC_API)
+
+      expect(setLoadingTimeSpy).toHaveBeenCalledOnceWith(jasmine.any(Number), false)
+      // Verify the timestamp was captured at call time (tick 10), not at drain time (tick 30)
+      const capturedTimestamp = setLoadingTimeSpy.calls.argsFor(0)[0] as number
+      expect(capturedTimestamp).toBeLessThan(clock.timeStamp(30))
```
mormubis: This assertion is too loose — it would pass even if the timestamp was captured at t=29. A regression that deferred capture to just before drain time wouldn't be caught. `toBe(clock.timeStamp(10))` would actually pin down what the comment claims to verify.

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -241,6 +241,10 @@ export function createPreStartStrategy(
       bufferApiCalls.add((startRumResult) => startRumResult.addTiming(name, time))
     },
 
+    setLoadingTime: ((callTimestamp = timeStampNow(), overwrite = false) => {
```
mormubis: This default never fires — `rumPublicApi` always passes an explicit timestamp. Misleads readers into thinking pre-start has its own independent capture. Either drop it or add a comment explaining it's just a defensive fallback.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
```diff
@@ -69,10 +74,28 @@ export function trackCommonViewMetrics(
     },
     stopINPTracking,
     setLoadEvent,
-    setViewEnd,
+    setViewEnd: (viewEndTime: RelativeTime) => {
+      viewEnded = true
+      inpSetViewEnd(viewEndTime)
```
mormubis: Nit: `setINPViewEnd` reads more naturally than `inpSetViewEnd`.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+import { createMutationRecord, mockGlobalPerformanceBuffer, mockRumConfiguration } from '../../../../test'
+import { PAGE_ACTIVITY_END_DELAY, PAGE_ACTIVITY_VALIDATION_DELAY } from '../../waitPageActivityEnd'
+import { LifeCycle } from '../../lifeCycle'
+import type { RumMutationRecord } from '../../../browser/domMutationObservable'
+import { trackCommonViewMetrics } from './trackCommonViewMetrics'
+
+const BEFORE_PAGE_ACTIVITY_VALIDATION_DELAY = (PAGE_ACTIVITY_VALIDATION_DELAY * 0.8) as Duration
+const AFTER_PAGE_ACTIVITY_END_DELAY = PAGE_ACTIVITY_END_DELAY * 1.1
+
+describe('trackCommonViewMetrics', () => {
```
mormubis, line 16: Nit: a one-liner explaining why this spec exists separately from `trackViews.spec.ts` (e.g. "tests the internal module in isolation to avoid setting up the full view stack for edge cases") would help future readers land faster.

test/e2e/scenario/rum/views.scenario.ts
```diff
@@ -66,6 +66,51 @@ test.describe('rum views', () => {
...
+              }, 500)
+            })
+        )
+
+        await flushEvents()
+        const viewEvent = intakeRegistry.rumViewEvents.at(-1)
+        expect(viewEvent).toBeDefined()
+        // Should reflect the second (overwritten) value (~500ms), not the first (~200ms)
+        expect(viewEvent!.view.loading_time).toBeGreaterThanOrEqual(500 * 1e6)
+      })
```
mormubis, line 113: Both new tests call `setViewLoadingTime()` after init. The buffered pre-start path (`setViewLoadingTime()` → `init()`) has unique timing behavior and is only covered at unit level — worth adding an E2E sanity check for it too.

packages/core/src/domain/telemetry/telemetryEvent.types.ts
MaelNamNam: Woooooops, sorry, I missed this one. Thanks & renamed in 4bb65b3.

packages/rum-core/src/boot/preStartRum.spec.ts
MaelNamNam: Thanks. I pinned it to `toBe(clock.timeStamp(10))` as you suggested in 4a6331e.

packages/rum-core/src/boot/preStartRum.ts
MaelNamNam: Got it, I removed the default so `callTimestamp` is now a required `TimeStamp` param. See bc37a25.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
MaelNamNam: Done. I Added a one-liner in eba54d6.

test/e2e/scenario/rum/views.scenario.ts
MaelNamNam: Thanks! It's done in 72c0bf3.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.ts
MaelNamNam: Much better, thanks, I renamed it to `setINPViewEnd` in a177d40

packages/rum-core/src/boot/preStartRum.ts
MaelNamNam: And fixed the test in 75431bb 😅

packages/core/src/domain/telemetry/telemetryEvent.types.ts
MaelNamNam: Actually, I went too fast on this and I'm reverting it (cf 00a28dc): the telemetry event is called `add-view-loading-time` because it was define as such in the Mobile SDKs. As this `setViewLoadingTime` is an experimental API, I feel it's best to re-use already existing event names rather than add a new one just for Browser that might eventually get removed.

packages/rum-core/src/domain/view/trackViews.ts
```diff
@@ -376,6 +379,11 @@ function newView(
       customTimings[sanitizeTiming(name)] = relativeTime
       scheduleViewUpdate()
     },
+    setLoadingTime(callTimestamp?: TimeStamp, overwrite = false) {
+      if (setLoadingTimeOnMetrics(callTimestamp, overwrite)) {
+        scheduleViewUpdate()
+      }
+    },
```
bcaudan: **💬 suggestion:** ‏what about scheduling view update from inside common view metrics?
it would:
- avoid this extra wrapper, `currentview.setLoadingTime` could directly be `commonViewMetrics.setLoadingTime`
- avoid to return a boolean from `commonViewMetrics.setLoadingTime`

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  beforeEach(() => {
+    mockGlobalPerformanceBuffer()
+    clock = mockClock()
+    domMutationObservable = new Observable()
+    windowOpenObservable = new Observable()
+    scheduleViewUpdateSpy = jasmine.createSpy('scheduleViewUpdate')
+  })
+
+  describe('manual loading time suppresses auto-detected loading time callback', () => {
+    it('should ignore auto-detected loading time when manual loading time was already set', () => {
```
bcaudan, line 32: **❓ question:** ‏is it not already covered by `should suppress auto-detected loading time after manual call`?

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+      clock.tick(100)
+      const secondResult = setLoadingTime(undefined, true)
+      expect(secondResult).toBe(false)
+
+      expect(getCommonViewMetrics().loadingTime).toBe(firstValue)
+    })
+  })
+
+  describe('setLoadingTime overwrite logic', () => {
+    it('should reject setLoadingTime without overwrite after first call', () => {
```
bcaudan: **❓ question:** ‏is it not already covered by `should not overwrite loading time by default on subsequent calls`?

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+      const firstValue = getCommonViewMetrics().loadingTime
+
+      clock.tick(200)
+      const result = setLoadingTime()
+      expect(result).toBe(false)
+
+      expect(getCommonViewMetrics().loadingTime).toBe(firstValue)
+    })
+
+    it('should allow overwrite before view end', () => {
```
bcaudan: **❓ question:** ‏is it not already covered by `should overwrite loading time when overwrite is true`?

packages/rum-core/src/domain/view/trackViews.ts
MaelNamNam: Updated in 85fa995. `setLoadingTime` now calls `scheduleViewUpdate` internally and no longer returns a boolean. The wrapper in `trackViews` is replaced by a direct alias.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
MaelNamNam: Sorry, I missed this one. Removed in 85fa995.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
MaelNamNam: Sorry and same, removed in 85fa995.

packages/rum-core/src/domain/view/viewMetrics/trackCommonViewMetrics.spec.ts
MaelNamNam: Not entirely, this test covers a specific race condition where `setLoadEvent` fires after a manual loading time was already set, which triggers the auto-detection callback. The `trackViews` test doesn't exercise that path. Do you feel I should delete this test too?

bcaudan: No, let's keep it then

mormubis (approved):
LGTM, thanks for the changes!

PR #4184, opened 2026-02-13
✨ Add tab.id to all RUM and Logs events

mormubis:
## Motivation

The Browser SDK shares session state across tabs via cookies/localStorage, but has no way to identify which specific tab produced a given event. Adding `tab.id` enables distinguishing events from different tabs within the same session, enabling tab-level correlation between RUM and Logs.

## Changes

- **New `tabContext` module** (`packages/core/src/domain/contexts/tabContext.ts`): Generates a UUID v4 per browser tab, persisted in `sessionStorage` so it survives same-tab navigations and reloads. Falls back to an in-memory UUID (cached at module level) when `sessionStorage` is unavailable (Web Workers, sandboxed iframes, quota exceeded), ensuring RUM and Logs always share the same `tab.id` within a page lifecycle.
- **Wired into RUM and Logs**: Both SDKs call `startTabContext(hooks)` which registers an `Assemble` hook contributing `{ tab: { id: "<uuid>" } }` to every event.
- **Unit tests**: Hook output, ID consistency across triggers, sessionStorage persistence/reuse, fallback sharing across SDK instances, and error paths (`getItem` throws, `setItem` throws).

### Field placement

`tab.id` sits at the top level alongside `view`, `session`, `usr`, etc. — it has a different lifecycle than a view (a single tab spans multiple views via SPA navigation) and is customer-facing, unlike the `_dd` internal namespace.

> **Note:** A schema PR in `rum-events-format` is required to formally define `tab` in the event schema. That is tracked separately; the field is emitted correctly today via the open-ended index signatures in the generated type files.

### Behavior

| Scenario | Behavior |
|---|---|
| New tab | Generates UUID, persists to `sessionStorage` |
| Page reload / same-tab navigation | Reuses UUID from `sessionStorage` |
| `sessionStorage` unavailable | Shared in-memory UUID for the page lifecycle |
| Tab duplication | Both tabs share the same ID (known tradeoff — inherent to `sessionStorage`-based schemes) |

## Test instructions

```bash
yarn test:unit --spec packages/core/src/domain/contexts/tabContext.spec.ts
yarn test:unit --spec packages/rum-core/src/domain/assembly.spec.ts
yarn test:unit --spec packages/logs/src/boot/startLogs.spec.ts
yarn typecheck
```

**Manual verification (sandbox):**

1. Run `yarn dev` → open `http://localhost:8080` in two separate tabs
2. In each tab's DevTools console, paste:
   ```js
   window.__ddBrowserSdkExtensionCallback = (msg) => {
     if (msg.type === 'rum' || msg.type === 'logs')
       console.log(msg.type, msg.payload.type ?? 'log', '→ tab.id:', msg.payload.tab?.id)
   }
   ```
3. Observe: each tab emits a different UUID; reloading a tab preserves its UUID.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/contexts/tabContext.spec.ts
+++ b/packages/core/src/domain/contexts/tabContext.spec.ts
@@ -10,6 +10,7 @@ describe('tabContext', () => {
   let hooks: Hooks
 
   beforeEach(() => {
+    resetCachedTabId()
     hooks = createHooks()
     registerCleanupTask(() => {
       sessionStorage.removeItem(TAB_ID_STORAGE_KEY)


packages/core/src/domain/contexts/tabContext.ts
```diff
@@ -0,0 +1,34 @@
...
+    const existingId = sessionStorage.getItem(TAB_ID_STORAGE_KEY)
+    if (existingId) {
+      return existingId
+    }
+    const newId = generateUUID()
+    sessionStorage.setItem(TAB_ID_STORAGE_KEY, newId)
+    return newId
+  } catch {
+    // sessionStorage unavailable (e.g. Web Worker, sandboxed iframe, quota exceeded)
+    return generateUUID()
```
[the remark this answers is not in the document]
mormubis: Also wanted to clarify why `browser_tab_id` lives under `_dd` rather than `view`: the tab ID has a different lifecycle than a view — a single tab can have many views (SPA navigations), while the tab ID persists across them via `sessionStorage`. Placing it under `view` would be misleading since `view.id` changes per navigation. `_dd` is the right home as internal SDK metadata, alongside `browser_sdk_version` and `sdk_name`.

packages/core/src/domain/contexts/tabContext.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  hooks.register(HookNames.Assemble, () => ({
+    _dd: {
+      browser_tab_id: tabId,
+    },
+  }))
+
+  return {
+    getTabId: () => tabId,
+  }
```
BenoitZugmeyer: **💬 suggestion:** ‏this is just used in tests, right? Maybe use the hook in the tests instead. Like [this](https://github.com/DataDog/browser-sdk/blob/a40c729967904cc619a1a6c78e63a45a9d811ee4/packages/rum-core/src/domain/contexts/defaultContext.spec.ts#L18) for example

packages/rum-core/src/rumEvent.types.ts
BenoitZugmeyer, line 1: The new field should be added in rum-events-format

packages/core/src/domain/contexts/tabContext.ts
BenoitZugmeyer: Suggestion: add a module-level variable to store the tab id, so Logs and RUM share the same id when installed via NPM. 

```ts
let cachedTabId: string | undefined

function retrieveOrCreateTabId(): string {
  if (!cachedTabId) {
     ...
     cachedTabId = ...
  }

  return cachedTabId
}
```

packages/core/src/domain/contexts/tabContext.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const TAB_ID_STORAGE_KEY = '_dd_tab_id'
+
+// Cached fallback ID for environments where sessionStorage is unavailable (e.g. sandboxed
+// iframes, Web Workers). Ensures RUM and Logs share the same tab ID within a page lifecycle
+// even when storage access throws.
+let fallbackTabId: string | undefined
+
+export function startTabContext(hooks: AbstractHooks) {
+  const tabId = mockable(retrieveOrCreateTabId)()
```
BenoitZugmeyer: **nitpick:** you use `mockable()` but you never actually mock it in tests

packages/core/src/domain/contexts/tabContext.ts
```diff
@@ -0,0 +1,39 @@
...
+    const newId = generateUUID()
+    sessionStorage.setItem(TAB_ID_STORAGE_KEY, newId)
+    return newId
+  } catch {
+    // sessionStorage unavailable (e.g. Web Worker, sandboxed iframe, quota exceeded)
+    if (!fallbackTabId) {
+      fallbackTabId = generateUUID()
+    }
+    return fallbackTabId
+  }
```
BenoitZugmeyer, line 43: **nitpick:** I was thinking, maybe use a "cache" pattern instead of just a fallback. It could be simpler to follow and a tiny bit safer since we would access the session storage only once when RUM and Logs are used together in NPM setups. Maybe something like:

```ts
if (!cachedTabId) {
  cachedTabId = getOrCreateIdInSessionStorage() || generateUUID()
}
return cachedTabId
```

packages/core/src/domain/contexts/tabContext.ts
mormubis: @BenoitZugmeyer I thought this was the new standard in case someone needs to mock it. But if I don't apply it now, nobody will be able to do it, right?

BenoitZugmeyer: Ah! I don't think we should preemptively make things mockable. If you don't make it mockable now, we can always make it later (when we need to)

mormubis: Done — removed `mockable()` and call `retrieveOrCreateTabId()` directly. Added `resetCachedTabId()` (exported for test cleanup only) so specs can reset the in-memory cache between tests via `registerCleanupTask`.

packages/core/src/domain/contexts/tabContext.ts
mormubis: Done — implemented exactly this pattern. `retrieveOrCreateTabId()` now caches unconditionally (not just in the fallback path), and `getOrCreateIdInSessionStorage()` is a private helper that returns `undefined` on failure so the `?? generateUUID()` fallback kicks in. SessionStorage is accessed at most once per page lifecycle regardless of how many SDK packages are initialized.

PR #4194, opened 2026-02-17
✨ make all sampling decisions deterministic

thomas-lebeau:
## Motivation

Decouple sampling decisions from session storage. Currently, the session store computes and persists tracking types (e.g. `rum=1`, `logs=0`) alongside the session ID. This means sampling logic lives deep in the core session store, and both RUM and Logs depend on it to make sampling decisions at session creation time.

This PR moves sampling responsibility to each product (RUM, Logs), where it's computed on demand using the deterministic `isSampled()` function. Since sampling is deterministic based on session ID + sample rate, there's no need to persist the decision — it can be recomputed at any time.

Additionally, it corrects a monotone correlation issue in chained deterministic sampling: when the same hash is compared against two thresholds (session → replay/tracing/profiling), the effective child rate must be adjusted to `parentRate × childRate / 100` so that `childRate%` of tracked sessions get the child feature.

## Changes

**Move sampler from `rum-core` to `core`**
- Moved `sampler.ts` from `packages/rum-core/src/domain/sampler/` to `packages/core/src/domain/` so both RUM and Logs can use it
- Updated imports in `tracer.ts` and `profilerApi.ts`

**Simplify core session manager and store**
- `SessionManager` no longer carries a `TrackingType` generic, `productKey`, or `computeTrackingType` callback
- `SessionStore` always generates a session ID on creation (no more "not tracked = no ID" logic)
- Removed `SESSION_NOT_TRACKED` checks and tracking type persistence from session state
- `SessionContext` no longer has a `trackingType` field
- `hasSessionInCache()` now checks for `id` instead of product key
- `isSessionInCacheOutdated()` only compares session IDs (no more tracking type comparison)

**Move sampling to each product**
- **RUM** (`rumSessionManager.ts`): `computeTrackingType()` now takes `(configuration, sessionId)` and calls `isSampled()` at query time
- **Logs** (`logsSessionManager.ts`): Same pattern
- Both stubs updated to generate a UUID for deterministic sampling consistency
- Removed `RUM_SESSION_KEY`, `LOGS_SESSION_KEY` constants and `hasValidRumSession`/`hasValidLoggerSession` validators

**Apply correction factor for chained deterministic sampling**
- Added `correctedChildSampleRate(parentRate, childRate)` in `core/sampler.ts`
- Applied at 3 call sites: replay (`rumSessionManager.ts`), tracing (`tracer.ts`), profiling (`profilerApi.ts`)
- Wrapped `isProfilingSupported` in `mockable()` to enable testing the profiler sampling path

## Test instructions

1. `yarn test:unit --spec packages/core/src/domain/sampler.spec.ts` — unit tests for `correctedChildSampleRate` and `isSampled`
2. `yarn test:unit --spec packages/rum-core/src/domain/rumSessionManager.spec.ts` — deterministic sampling with correction for replay
3. `yarn test:unit --spec packages/rum-core/src/domain/tracing/tracer.spec.ts` — correction for tracing
4. `yarn test:unit --spec packages/rum/src/boot/profilerApi.spec.ts` — correction for profiling
5. `yarn typecheck` — no type errors

Correction tests use a `MID_HASH_UUID` (hash ~50.7%) with `sessionSampleRate=60, childRate=60`. The corrected rate (36) rejects while the uncorrected rate (60) would pass — tests fail if the correction is removed.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -110,23 +109,14 @@ export function startRumSessionManagerStub(
   })
 }
 
-function computeTrackingType(configuration: RumConfiguration, rawTrackingType?: string) {
-  if (hasValidRumSession(rawTrackingType)) {
-    return rawTrackingType
-  }
-  if (!performDraw(configuration.sessionSampleRate)) {
+function computeTrackingType(configuration: RumConfiguration, sessionId: string): RumTrackingType {
+  if (!isSampled(sessionId, configuration.sessionSampleRate)) {
```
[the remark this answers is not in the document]
thomas-lebeau: Session IDs have always been UUIDs in this codebase — they are generated by \`generateUUID()\` in \`sessionStore.ts\`. Even legacy cookies from \`oldCookiesMigration.ts\` contain UUIDs originally generated by the same function in older SDK versions. There are no non-UUID session IDs to guard against.

packages/logs/src/domain/logsSessionManager.ts
```diff
@@ -24,49 +22,48 @@ export function startLogsSessionManager(
...
 ): void {
+  // Generate a session ID for deterministic sampling in stub mode
+  const stubSessionId = generateUUID()
+  const isTracked = computeTrackingType(configuration, stubSessionId) === LoggerTrackingType.TRACKED
+  const session: LogsSession | undefined = isTracked ? { id: stubSessionId } : undefined
   onReady({
-    findTrackedSession: () => ({}),
+    findTrackedSession: () => session,
     expireObservable: new Observable(),
   })
```
BenoitZugmeyer: **issue:** session manager stub should not do sampling

packages/rum-core/src/domain/tracing/tracer.spec.ts
```diff
@@ -259,6 +259,26 @@ describe('tracer', () => {
...
 
+    it('should apply the correction factor for chained sampling on the trace sample rate', () => {
+      if (!window.BigInt) {
+        pending('BigInt is not supported')
+      }
+
+      // MID_HASH_UUID has a hash of ~50.7%. With sessionSampleRate=60 and traceSampleRate=60:
+      // - Without correction: isSampled(id, 60) → true (50.7 < 60)
+      // - With correction: isSampled(id, 60*60/100=36) → false (50.7 > 36)
+      const MID_HASH_UUID = '88ef85ab-7902-45f0-b93b-2def1ec3e5fe'
```
BenoitZugmeyer: **suggestion:** move those test constants to a common module (maybe in `packages/core/test/sampling.ts`)

packages/rum-core/src/domain/rumSessionManager.ts
```diff
@@ -49,44 +48,40 @@ export function startRumSessionManager(
...
+        }
+      }
+    })
+
+    onReady({
+      findTrackedSession: (startTime) => {
+        const session = sessionManager.findSession(startTime)
+        if (!session || session.id === 'invalid') {
+          return
+        }
```
BenoitZugmeyer, line 66: **thought:** Those `invalid` session id is somewhat bothering, but I guess it'll be removed in a follow-up step?

packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -125,7 +131,10 @@ function injectHeadersIfTracingAllowed(
     return
   }
 
-  const traceSampled = isSampled(session.id, configuration.traceSampleRate)
+  const traceSampled = isSampled(
+    session.id,
+    correctedChildSampleRate(configuration.sessionSampleRate, configuration.traceSampleRate)
```
mormubis, line 136: IIUC, the new session manager will start regardless of the product and won't be tied to RUM. Hence, do we need child sample rates? Can we not have plain sample rates that are easier to understand?

thomas-lebeau: Traces still depend on RUM, if rum don't start, we don't collect traces.
It's true that we could imagine a separate Trace product, but that's a separate topic.
currently we sample traces as: _N% of collected resources will have traces_. with deterministic sampling we will have _N% of sampled session will have traces_

packages/rum-core/src/domain/rumSessionManager.ts
thomas-lebeau: I didn't think about it yet, but I'm taking note 👍

PR #4195, opened 2026-02-17
🔥 Remove betaEncodeCookieOptions and always encode cookie options

thomas-lebeau:
> ⚠️ Stacked on #4157 — review that PR first

## Motivation

The `betaEncodeCookieOptions` feature flag was introduced as a beta option to encode cookie options into the cookie value, mitigating micro-session issues caused by conflicting cookie configurations (e.g., when RUM and Logs SDKs are initialized with different `trackSessionAcrossSubdomains` or `usePartitionedCrossSiteSessionCookie` values).

This option has proven stable and is now ready to be promoted to the default behavior. Removing the beta flag simplifies the codebase and ensures all users benefit from encoded cookie options.

## Changes

- Remove the `betaEncodeCookieOptions` init configuration option and its associated `Configuration` field
- Always encode cookie options into the session cookie value (previous behavior when `betaEncodeCookieOptions: true`)
- Simplify `retrieveSessionCookie` to always use the encoded cookie retrieval path, removing the `configuration` parameter
- Remove `betaEncodeCookieOptions` from serialized configuration and default context telemetry
- Update unit tests to reflect that cookie options are always encoded (cookie values now include `&c=0` suffix)
- Simplify E2E tests for conflicting cookie options: remove the `true`/`false` loop and `test.fail()` guard since the behavior is now always correct

## Test instructions

- Run unit tests: `yarn test:unit --spec packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts`
- Run E2E session store tests: `yarn test:e2e -g "Session Stores"`
- Verify no regression in session handling with different cookie configurations

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
+++ b/packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
@@ -74,7 +74,24 @@ function storeSessionCookie(
  * If there is no match, return the first cookie, because that's how `getCookie()` works
  */
 export function retrieveSessionCookie(cookieOptions: CookieOptions): SessionState {
-  return retrieveSessionCookieFromEncodedCookie(cookieOptions)
+  const cookies = getCookies(SESSION_STORE_KEY)
+  const opts = encodeCookieOptions(cookieOptions)
+
+  let sessionState: SessionState | undefined
+
+  // reverse the cookies so that if there is no match, the cookie returned is the first one
+  for (const cookie of cookies.reverse()) {
+    sessionState = toSessionState(cookie)
+
+    if (sessionState.c === opts) {
+      break
+    }
+  }
+
+  // remove the cookie options from the session state as this is not part of the session state
+  delete sessionState?.c
+
+  return sessionState ?? {}
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -77,14 +73,8 @@ function storeSessionCookie(
...
-  if (configuration.betaEncodeCookieOptions) {
-    return retrieveSessionCookieFromEncodedCookie(cookieOptions)
-  }
-
-  const sessionString = getCookie(SESSION_STORE_KEY)
-  const sessionState = toSessionState(sessionString)
-  return sessionState
+export function retrieveSessionCookie(cookieOptions: CookieOptions): SessionState {
+  return retrieveSessionCookieFromEncodedCookie(cookieOptions)
 }
```
BenoitZugmeyer, line 95: **nitpick:** Inline `retrieveSessionCookieFromEncodedCookie`

PR #4196, opened 2026-02-17
👷 Migrate unit tests from Karma/Jasmine to Vitest

mormubis:
## Motivation

Karma is deprecated. This migrates the full unit test suite to Vitest 4.x with browser mode (Playwright).

## Changes

Most of the diff is mechanical spec file replacements (`jasmine.createSpy()` → `vi.fn()`, etc.). The interesting parts:

`mockClock` had to be rewritten because Vitest's `vi.useFakeTimers` fakes `performance` by default, which breaks every perf-related test. The `toFake` list now explicitly excludes it.

`allJsonSchemas` was using webpack's `require.context` to load schema files. Replaced with Vite's `import.meta.glob`.

BrowserStack needed `bs-local.com` as server host because Safari replaces localhost with its own domain, breaking cookie access in vitest's iframe. Same issue we already have with ServiceWorker tests in E2E. Cookie tests are still skipped on Safari because this alone doesn't fully fix it.

The `unit-bs` CI job needs Playwright installed with system deps (`--with-deps`) and 4 CPUs to handle 5 concurrent remote browsers without timing out.

`vite@8.0.8` is pinned in the developer extension because `@vitejs/plugin-react` can't resolve `vite/internal` in newer patch versions.

3784 unit tests pass, 18628 BrowserStack tests across 5 browsers.

## Test instructions

```bash
yarn test:unit
yarn typecheck
yarn lint
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging


vitest.config.ts
```diff
@@ -0,0 +1,80 @@
...
+  },
+
+  define: buildEnvDefines,
+
+  optimizeDeps: {
+    include: ['pako'],
+  },
+
+  test: {
+    browser: {
```
mormubis, line 60: Browser mode with Playwright. Locally it runs headless Chromium. On BrowserStack, Playwright connects to remote browsers via WebSocket (`wss://cdp.browserstack.com/playwright`), so the old browser versions (Chrome 63, Edge 80, etc.) are real browsers hosted by BrowserStack, not local Playwright binaries. The aliases mirror `tsconfig.base.json` so vitest resolves packages the same way TypeScript does.

packages/core/test/emulate/mockClock.ts
```diff
@@ -1,19 +1,34 @@
...
+  // at page load, so it doesn't suffer from timing races between performance.now()
+  // and Date.now() captures. This matches getNavigationStart() in timeUtils.ts.
+  const timeOrigin = performance.timing.navigationStart
 
-  const timeOrigin = performance.timing.navigationStart // @see getNavigationStart() in timeUtils.ts
+  // Exclude 'performance' from faking so our override sticks on the real object
+  // AND so performance.timing.navigationStart remains accessible.
+  // When performance IS faked, @sinonjs/fake-timers replaces the object and our
+  // override silently fails (the fake performance.now() starts at 0).
+  vi.useFakeTimers({
```
mormubis, line 17: `performance` is excluded from `toFake`. If we fake it, `performance.now()` returns fake time but the browser's internal performance observer still uses real time. Breaks every perf-related test.

packages/core/test/collectAsyncCalls.ts
```diff
@@ -1,35 +1,54 @@
...
-export function collectAsyncCalls<F extends jasmine.Func>(
-  spy: jasmine.Spy<F>,
+export interface MockCalls<F extends (...args: any[]) => any> {
+  all(): Array<{ args: Parameters<F>; returnValue: ReturnType<F> }>
+  count(): number
+  argsFor(index: number): Parameters<F>
+  mostRecent(): { args: Parameters<F>; returnValue: ReturnType<F> }
+}
+
+export function collectAsyncCalls<F extends (...args: any[]) => any>(
```
mormubis, line 10: Vitest equivalent of the old Jasmine `calls.all()` pattern. Wraps a `vi.fn()` and resolves a promise once the expected number of calls is reached.

packages/browser-rum-core/test/allJsonSchemas.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference types="vite/client" />
+// Load all JSON schema files from the rum-events-format submodule.
+// Uses Vite's import.meta.glob (replaces webpack's require.context).
+const schemaModules = import.meta.glob('../../../rum-events-format/schemas/**/*.json', { eager: true })
```
mormubis: `require.context` → `import.meta.glob`. Same result, Vite API instead of webpack.

vitest.bs.config.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  optimizeDeps: {
+    include: ['pako'],
+  },
+
+  test: {
+    browser: {
+      enabled: true,
+      provider: playwright(),
+      // Use bs-local.com instead of localhost so Safari on BrowserStack can access cookies.
```
mormubis, line 125: Safari on BrowserStack replaces localhost with `bs-local.com`, which breaks cookie access in vitest's iframe. Same issue we already have with ServiceWorker tests in E2E.

packages/rum-core/src/browser/domMutationObservable.spec.ts
```diff
@@ -8,27 +9,28 @@ const DOM_MUTATION_OBSERVABLE_DURATION = 16
...
-
-      setTimeout(() => {
-        expect(counter).toBe(expectedMutations)
-        root.parentNode!.removeChild(root)
-        domMutationSubscription.unsubscribe()
-        done()
-      }, DOM_MUTATION_OBSERVABLE_DURATION)
-    }
+    return () =>
+      new Promise<void>((resolve) => {
```
mormubis, line 13: Vitest doesn't support `done()` callbacks. Converted to `new Promise<void>`.

packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
```diff
@@ -7,6 +8,13 @@ import { SESSION_COOKIE_EXPIRATION_DELAY, SESSION_EXPIRATION_DELAY, SESSION_TIME
 import { buildCookieOptions, selectCookieStrategy, initCookieStrategy } from './sessionInCookie'
 import { SESSION_STORE_KEY } from './sessionStoreStrategy'
 
+// Safari on BrowserStack cannot access cookies because vitest runs tests in an iframe
+// and BrowserStack replaces localhost with bs-local.com, triggering Safari's ITP restrictions.
```
mormubis, line 12: Safari on BrowserStack still can't access cookies in the vitest iframe even with `bs-local.com`. Skipped on Safari only, runs fine on the other 4 browsers.

packages/core/src/tools/instrumentMethod.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { vi, beforeEach, describe, expect, it } from 'vitest'
```
mormubis, line 1: Typical spec migration. `jasmine.createSpy()` → `vi.fn()`, `spyOn` → `vi.spyOn`, etc. All 257 spec files follow this same pattern.

PR #4201, opened 2026-02-18
⚗️ Partial view updates (experimental)

mormubis:
## Motivation

Every periodic view update was sending the full view payload even when only one counter changed. Benchmarks showed 50–90% of the data was redundant. This implements the partial view updates RFC to reduce bandwidth by sending only changed fields in subsequent view events.

Aligned with [rum-events-format #355](https://github.com/DataDog/rum-events-format/pull/355) (now merged).

## Changes

When `partial_view_updates` is enabled, the SDK sends the first event per `view.id` as a full `view`, then sends `view_update` diffs with only changed fields. The diff runs post-assembly in `startRumBatch.ts` so `beforeSend` always sees the full event (backward-compatible). `view_update` events bypass the assembly pipeline intentionally, they are a bandwidth optimization and not a customer-visible event type.

A full `view` checkpoint is sent every 100 updates for backend recovery. Checkpoints can be disabled with `partial_view_updates_no_checkpoint`.

`view_update` events use `batch.add` instead of `upsert`, so a batch can contain a full `view` followed by `view_update` events. This is intentional: if we consolidated them, we wouldn't be able to tell if the backend missed an intermediate update or it was never sent.

## Test instructions

```bash
yarn test:unit
yarn test:e2e:init && yarn test:e2e -g "partial view"
```

Or manually:

```js
datadogRum.init({
  enableExperimentalFeatures: ['partial_view_updates'],
})
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -49,6 +49,10 @@ export function startRumAssembly(
       ...VIEW_MODIFIABLE_FIELD_PATHS,
       ...ROOT_MODIFIABLE_FIELD_PATHS,
     },
+    // view_update events are created post-assembly in startRumBatch.ts and never go through
+    // this pipeline — they intentionally bypass beforeSend. This entry is required by the
+    // exhaustive type but is never reached in practice.
+    [RumEventType.VIEW_UPDATE]: {},
     [RumEventType.ERROR]: {
       'error.message': 'string',
       'error.stack': 'string',
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -49,6 +49,10 @@ export function startRumAssembly(
       ...VIEW_MODIFIABLE_FIELD_PATHS,
       ...ROOT_MODIFIABLE_FIELD_PATHS,
     },
+    // view_update events are created post-assembly in startRumBatch.ts and never reach this pipeline.
+    // The full view already went through assembly (as RumEventType.VIEW), so any beforeSend
+    // modifications (e.g. PII scrubbing) are already reflected in the view_update diff.
+    [RumEventType.VIEW_UPDATE]: {},
     [RumEventType.ERROR]: {
       'error.message': 'string',
       'error.stack': 'string',
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -50,8 +50,10 @@ export function startRumAssembly(
       ...ROOT_MODIFIABLE_FIELD_PATHS,
     },
     // view_update events are created post-assembly in startRumBatch.ts and never go through
-    // this pipeline — they intentionally bypass beforeSend. This entry is required by the
-    // exhaustive type but is never reached in practice.
+    // this pipeline. The full view event already went through assembly (as RumEventType.VIEW),
+    // so any beforeSend modifications (e.g. PII scrubbing on view.performance.lcp.resource_url)
+    // are already reflected in the view_update diff. This entry is required by the exhaustive
+    // type but is never reached in practice.
     [RumEventType.VIEW_UPDATE]: {},
     [RumEventType.ERROR]: {
       'error.message': 'string',
--- a/packages/rum-core/src/domain/assembly.ts
+++ b/packages/rum-core/src/domain/assembly.ts
@@ -49,12 +49,6 @@ export function startRumAssembly(
       ...VIEW_MODIFIABLE_FIELD_PATHS,
       ...ROOT_MODIFIABLE_FIELD_PATHS,
     },
-    // view_update events are created post-assembly in startRumBatch.ts and never go through
-    // this pipeline. The full view event already went through assembly (as RumEventType.VIEW),
-    // so any beforeSend modifications (e.g. PII scrubbing on view.performance.lcp.resource_url)
-    // are already reflected in the view_update diff. This entry is required by the exhaustive
-    // type but is never reached in practice.
-    [RumEventType.VIEW_UPDATE]: {},
     [RumEventType.ERROR]: {
       'error.message': 'string',
       'error.stack': 'string',
--- a/packages/rum-core/src/domain/trackEventCounts.ts
+++ b/packages/rum-core/src/domain/trackEventCounts.ts
@@ -30,7 +30,7 @@ export function trackEventCounts({
   }
 
   const subscription = lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event): void => {
-    if (event.type === 'view' || event.type === 'vital' || !isChildEvent(event)) {
+    if (event.type === 'view' || event.type === 'view_update' || event.type === 'vital' || !isChildEvent(event)) {
       return
     }
     switch (event.type) {
--- a/packages/rum-core/src/domain/trackEventCounts.ts
+++ b/packages/rum-core/src/domain/trackEventCounts.ts
@@ -30,7 +30,7 @@ export function trackEventCounts({
   }
 
   const subscription = lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event): void => {
-    if (event.type === 'view' || event.type === 'view_update' || event.type === 'vital' || !isChildEvent(event)) {
+    if (event.type === 'view' || event.type === 'vital' || !isChildEvent(event)) {
       return
     }
     switch (event.type) {
--- a/packages/rum-core/src/domain/view/viewDiff.spec.ts
+++ b/packages/rum-core/src/domain/view/viewDiff.spec.ts
@@ -1,0 +1,116 @@
+import { isEqual, diffMerge } from './viewDiff'
+
+describe('isEqual', () => {
+  it('should return true for identical primitives', () => {
+    expect(isEqual(1, 1)).toBe(true)
+    expect(isEqual('a', 'a')).toBe(true)
+    expect(isEqual(true, true)).toBe(true)
+    expect(isEqual(null, null)).toBe(true)
+    expect(isEqual(undefined, undefined)).toBe(true)
+  })
+
+  it('should return false for different primitives', () => {
+    expect(isEqual(1, 2)).toBe(false)
+    expect(isEqual('a', 'b')).toBe(false)
+    expect(isEqual(true, false)).toBe(false)
+    expect(isEqual(null, undefined)).toBe(false)
+  })
+
+  it('should return true for deeply equal objects', () => {
+    expect(isEqual({ a: 1, b: { c: 2 } }, { a: 1, b: { c: 2 } })).toBe(true)
+  })
+
+  it('should return false for objects with different values', () => {
+    expect(isEqual({ a: 1 }, { a: 2 })).toBe(false)
+  })
+
+  it('should return false for objects with different keys', () => {
+    expect(isEqual({ a: 1 }, { b: 1 })).toBe(false)
+  })
+
+  it('should return true for equal arrays', () => {
+    expect(isEqual([1, 2, 3], [1, 2, 3])).toBe(true)
+  })
+
+  it('should return false for arrays with different lengths', () => {
+    expect(isEqual([1, 2], [1, 2, 3])).toBe(false)
+  })
+
+  it('should return false for arrays with different values', () => {
+    expect(isEqual([1, 2, 3], [1, 2, 4])).toBe(false)
+  })
+
+  it('should return false when comparing array to non-array', () => {
+    expect(isEqual([1], { 0: 1 })).toBe(false)
+  })
+
+  it('should return false for type mismatch', () => {
+    expect(isEqual(1, '1')).toBe(false)
+  })
+})
+
+describe('diffMerge', () => {
+  it('should return undefined when there are no changes', () => {
+    const result = diffMerge({ a: 1, b: 'x' }, { a: 1, b: 'x' })
+    expect(result).toBeUndefined()
+  })
+
+  it('should return changed primitive fields', () => {
+    const result = diffMerge({ a: 1, b: 2 }, { a: 1, b: 1 })
+    expect(result).toEqual({ b: 2 })
+  })
+
+  it('should include new fields not present in lastSent', () => {
+    const result = diffMerge({ a: 1, b: 2 }, { a: 1 })
+    expect(result).toEqual({ b: 2 })
+  })
+
+  it('should set null for deleted keys', () => {
+    const result = diffMerge({ a: 1 }, { a: 1, b: 2 })
+    expect(result).toEqual({ b: null })
+  })
+
+  it('should recursively diff nested objects', () => {
+    const result = diffMerge({ nested: { x: 1, y: 2 } }, { nested: { x: 1, y: 1 } })
+    expect(result).toEqual({ nested: { y: 2 } })
+  })
+
+  it('should return undefined for unchanged nested objects', () => {
+    const result = diffMerge({ nested: { x: 1 } }, { nested: { x: 1 } })
+    expect(result).toBeUndefined()
+  })
+
+  it('should include new nested objects', () => {
+    const result = diffMerge({ nested: { x: 1 } }, {})
+    expect(result).toEqual({ nested: { x: 1 } })
+  })
+
+  describe('replaceKeys option', () => {
+    it('should use full replace strategy for specified keys', () => {
+      const result = diffMerge({ arr: [1, 2, 3] }, { arr: [1, 2] }, { replaceKeys: new Set(['arr']) })
+      expect(result).toEqual({ arr: [1, 2, 3] })
+    })
+
+    it('should not include replace key if unchanged', () => {
+      const result = diffMerge({ arr: [1, 2] }, { arr: [1, 2] }, { replaceKeys: new Set(['arr']) })
+      expect(result).toBeUndefined()
+    })
+  })
+
+  describe('appendKeys option', () => {
+    it('should append only new trailing elements for array keys', () => {
+      const result = diffMerge({ items: [1, 2, 3] }, { items: [1, 2] }, { appendKeys: new Set(['items']) })
+      expect(result).toEqual({ items: [3] })
+    })
+
+    it('should include full array when it first appears', () => {
+      const result = diffMerge({ items: [1, 2] }, {}, { appendKeys: new Set(['items']) })
+      expect(result).toEqual({ items: [1, 2] })
+    })
+
+    it('should not include append key if array has not grown', () => {
+      const result = diffMerge({ items: [1, 2] }, { items: [1, 2] }, { appendKeys: new Set(['items']) })
+      expect(result).toBeUndefined()
+    })
+  })
+})
--- a/packages/rum-core/src/domain/view/viewDiff.spec.ts
+++ b/packages/rum-core/src/domain/view/viewDiff.spec.ts
@@ -28,6 +28,10 @@ describe('isEqual', () => {
     expect(isEqual({ a: 1 }, { b: 1 })).toBe(false)
   })
 
+  it('should return true for objects with same keys in different order', () => {
+    expect(isEqual({ a: 1, b: 2 }, { b: 2, a: 1 })).toBe(true)
+  })
+
   it('should return true for equal arrays', () => {
     expect(isEqual([1, 2, 3], [1, 2, 3])).toBe(true)
   })
--- a/packages/rum-core/src/domain/view/viewDiff.ts
+++ b/packages/rum-core/src/domain/view/viewDiff.ts
@@ -1,24 +1,14 @@
+import { isIndexableObject } from '@datadog/browser-core'
+
 /**
  * Compare two values for deep equality
  */
 export function isEqual(a: unknown, b: unknown): boolean {
-  // Reference equality
   if (a === b) {
     return true
   }
 
-  // Handle null/undefined
-  if (a === null || b === null || a === undefined || b === undefined) {
-    return a === b
-  }
-
-  // Type mismatch
-  if (typeof a !== typeof b) {
-    return false
-  }
-
-  // Primitives
-  if (typeof a !== 'object') {
+  if (a === null || typeof a !== 'object' || b === null || typeof b !== 'object') {
     return a === b
   }
 
@@ -30,12 +20,11 @@ export function isEqual(a: unknown, b: unknown): boolean {
     return a.every((val, idx) => isEqual(val, b[idx]))
   }
 
-  // One is array, other is not
   if (Array.isArray(a) || Array.isArray(b)) {
     return false
   }
 
-  // Objects
+  // Plain objects
   const aObj = a as Record<string, unknown>
   const bObj = b as Record<string, unknown>
   const aKeys = Object.keys(aObj)
@@ -54,6 +43,7 @@ export function isEqual(a: unknown, b: unknown): boolean {
 export interface DiffMergeOptions {
   replaceKeys?: Set<string>
   appendKeys?: Set<string>
+  ignoreKeys?: Set<string>
 }
 
 /**
@@ -62,6 +52,7 @@ export interface DiffMergeOptions {
  *
  * Default strategy is REPLACE (isEqual check). Exceptions:
  * - Both values are plain objects and key is not in replaceKeys: recurse (MERGE)
+ *   Sub-paths (e.g. 'view.custom_timings') are propagated to the recursive call.
  * - Both values are arrays and key is in appendKeys: include only new trailing elements (APPEND)
  */
 export function diffMerge(
@@ -72,22 +63,23 @@ export function diffMerge(
   const result: Record<string, unknown> = {}
   const replaceKeys = options?.replaceKeys ?? new Set<string>()
   const appendKeys = options?.appendKeys ?? new Set<string>()
+  const ignoreKeys = options?.ignoreKeys ?? new Set<string>()
 
   for (const key of Object.keys(current)) {
+    if (ignoreKeys.has(key)) {
+      continue
+    }
+
     const currentVal = current[key]
     const lastSentVal = lastSent[key]
 
-    if (
-      !replaceKeys.has(key) &&
-      currentVal !== null &&
-      typeof currentVal === 'object' &&
-      !Array.isArray(currentVal) &&
-      lastSentVal !== null &&
-      typeof lastSentVal === 'object' &&
-      !Array.isArray(lastSentVal)
-    ) {
+    if (!replaceKeys.has(key) && isIndexableObject(currentVal) && isIndexableObject(lastSentVal)) {
       // Both are plain objects and not marked for replace: recurse (MERGE)
-      const nestedDiff = diffMerge(currentVal as Record<string, unknown>, lastSentVal as Record<string, unknown>)
+      const nestedDiff = diffMerge(currentVal, lastSentVal, {
+        replaceKeys: extractSubPaths(replaceKeys, key),
+        appendKeys: extractSubPaths(appendKeys, key),
+        ignoreKeys: extractSubPaths(ignoreKeys, key),
+      })
       if (nestedDiff) {
         result[key] = nestedDiff
       }
@@ -110,4 +102,14 @@ export function diffMerge(
   }
 
   return Object.keys(result).length > 0 ? result : undefined
+}
+
+function extractSubPaths(keys: Set<string>, prefix: string): Set<string> {
+  const result = new Set<string>()
+  for (const key of keys) {
+    if (key.startsWith(`${prefix}.`)) {
+      result.add(key.slice(prefix.length + 1))
+    }
+  }
+  return result
 }
--- a/packages/rum-core/src/domain/view/viewDiff.ts
+++ b/packages/rum-core/src/domain/view/viewDiff.ts
@@ -1,0 +1,145 @@
+import { isEmptyObject } from '@datadog/browser-core'
+
+/**
+ * Compare two values for deep equality
+ */
+export function isEqual(a: unknown, b: unknown): boolean {
+  // Reference equality
+  if (a === b) {
+    return true
+  }
+
+  // Handle null/undefined
+  if (a === null || b === null || a === undefined || b === undefined) {
+    return a === b
+  }
+
+  // Type mismatch
+  if (typeof a !== typeof b) {
+    return false
+  }
+
+  // Primitives
+  if (typeof a !== 'object') {
+    return a === b
+  }
+
+  // Arrays
+  if (Array.isArray(a) && Array.isArray(b)) {
+    if (a.length !== b.length) {
+      return false
+    }
+    return a.every((val, idx) => isEqual(val, b[idx]))
+  }
+
+  // One is array, other is not
+  if (Array.isArray(a) || Array.isArray(b)) {
+    return false
+  }
+
+  // Objects
+  const aObj = a as Record<string, unknown>
+  const bObj = b as Record<string, unknown>
+  const aKeys = Object.keys(aObj)
+  const bKeys = Object.keys(bObj)
+
+  if (aKeys.length !== bKeys.length) {
+    return false
+  }
+
+  return aKeys.every((key) => bKeys.includes(key) && isEqual(aObj[key], bObj[key]))
+}
+
+/**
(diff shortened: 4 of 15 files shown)


packages/rum-core/src/domain/contexts/sourceCodeContext.ts
```diff
@@ -54,6 +54,10 @@ export function startSourceCodeContext(hooks: Hooks) {
   hooks.register(HookNames.Assemble, ({ domainContext, rawRumEvent }): DefaultRumEventAttributes | SKIPPED => {
     buildContextByFile()
 
+    if (rawRumEvent.type === 'view_update') {
+      return SKIPPED
+    }
```
mormubis, line 55: **Why is `view_update` explicitly skipped here, but not `view`?**

`view_update` gets an explicit guard because these events are emitted by SDK-internal timers and observers — user code never triggers them, so `handlingStack` is never set on their domain context. Without this guard, they'd go through `getSourceUrl()` only to hit `computeStackTrace({ stack: undefined })` and return `SKIPPED` anyway. The explicit guard makes the intent clear and short-circuits pointless work.

`view` events don't need the same guard because they *can* have a meaningful `handlingStack` — specifically when a user explicitly calls `datadogRum.startView()`, which captures the call stack. In that case `getSourceUrl()` can resolve a real file URL and the view event legitimately receives source code context. When no `handlingStack` is present (the common case), it naturally returns `SKIPPED` via the `contextByFile.get(url)` miss at the end.

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,127 @@
...
+  'source',
+  'ddtags',
+  'context',
+  'connectivity',
+  'usr',
+  'device',
+  'privacy',
+] as const
+
+export function computeAssembledViewDiff(
```
BenoitZugmeyer: **suggestion:** This function contains a lot of special cases. It could be nice to have something a bit more declarative:

```ts
export function computeAssembledViewDiff(current: RumViewEvent, last: RumViewEvent): AssembledRumEvent | undefined {
  const currentObj = current as unknown as Record<string, unknown>
  const lastObj = last as unknown as Record<string, unknown>

  const diff = diffMerge(currentObj, lastObj, {
    replaceKeys: new Set('view.custom_timings'),
    appendKeys: new Set(['_dd.page_states']),
    ignoreKeys: new Set(['_dd.document_version']),
  })

  return (
    diff &&
    combine(diff, {
      date: current.date,
      application: current.application,
      session: current.session,
      view: {
        id: current.view.id,
        url: current.view.url,
      },
      _dd: {
        document_version: current._dd.document_version,
        format_version: 2,
      },
      type: RumEventType.VIEW_UPDATE,
    })
  )
}
```

packages/rum-core/src/domain/view/viewDiff.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  // Type mismatch
+  if (typeof a !== typeof b) {
+    return false
+  }
+
+  // Primitives
+  if (typeof a !== 'object') {
+    return a === b
+  }
```
BenoitZugmeyer: **suggestion:** all this could be simplified as:

```ts
  if (a === null || typeof a !== 'object') {
    return a === b
  }
```

packages/rum-core/src/domain/view/viewDiff.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (currentVal !== null && typeof currentVal !== 'object') {
+      if (currentVal !== lastSentVal) {
+        result[key] = currentVal
+      }
+      continue
+    }
+
+    // Handle null explicitly
+    if (currentVal === null) {
+      if (currentVal !== lastSentVal) {
```
BenoitZugmeyer: **suggestion:** all this logic can be greatly simplified by assuming that we always replace the whole value unless:
* both values are objects and our key is not in `replaceKeys` (in which case we recurse)
* both values are arrays and our key is in `appendKeys` (in which case we slice)

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -28,12 +145,73 @@ export function startRumBatch(
     }),
   })
 
+  let lastSentView: AssembledRumEvent | undefined
+  let currentViewId: string | undefined
```
BenoitZugmeyer: **suggestion:** keep only `lastSentView`, remove `currentViewId` so we have a single source of truth.

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -28,12 +145,73 @@ export function startRumBatch(
...
+    // Note: view_update events are created here, post-assembly, and go directly to batch.add().
+    // They intentionally bypass RAW_RUM_EVENT_COLLECTED → assembly → RUM_EVENT_COLLECTED, which
+    // means they skip beforeSend entirely. view_update is an internal bandwidth optimization —
+    // not a customer-visible event type, and not modifiable via beforeSend.
+    if (!lastSentView) {
+      // Safety fallback (should not happen in practice)
+      lastSentView = serverRumEvent
+      batch.upsert(serverRumEvent, viewId)
+      return
+    }
```
BenoitZugmeyer: **suggestion:** all this could be simplified with a single condition `shouldSendViewUpdate`:

```ts
const shouldSendViewUdpate = lastSentView && viewId === currentViewId && serverRumEvent.view.is_active && viewUpdatesSinceCheckpoint > PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL

if (shouldSendViewUpdate) {
  viewUpdatesSinceCheckpoint += 1
  const diff = ...
  ...
} else {
  viewUpdatesSinceCheckpoint = 0
  batch.upsert(serverRumEvent, viewId)
}
```

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -28,12 +145,73 @@ export function startRumBatch(
...
+
+    const diff = computeAssembledViewDiff(serverRumEvent, lastSentView)
+    lastSentView = serverRumEvent
+    if (diff) {
+      sendToExtension('rum', diff)
+      batch.add(diff)
```
BenoitZugmeyer, line 140: **thought:** We should think a bit more about the upsert strategy. Here, we might end up having a full view + view updates in the same batch, which is less efficient than before.

mormubis: Yes! But in first instance, this would allow us to detect if we miss any intermediate events. If we skip (or consolidate) any view updates, we won't be able to tell if the backend missed them or they were never sent.

I'll leave a comment in the description.

packages/rum-core/src/domain/view/viewDiff.ts
mormubis: There's a bug here. If `a = {}` and `b = null`, `a === null || typeof a !== 'object'` is false and we reach `Object.keys(null)`, which throws. The `b === null` check is still needed. Also, I feel it makes it really difficult to understand all the checks, isn't it?

packages/rum-core/src/transport/startRumBatch.ts
mormubis: Done! Using `lastSentView?.view.id` instead.

packages/rum-core/src/transport/startRumBatch.ts
mormubis: I like it. The concern is the always-required fields (`view.id`, `view.url`, `_dd.document_version`) would be implicit, and `combine` doesn't exist yet. Do you want me to go that route?

packages/rum-core/src/transport/startRumBatch.ts
mormubis: I prefer early returns. Each path (new view, view end, checkpoint, diff) is readable on its own. I'm not sure the single condition improves it. What do you think?

packages/rum-core/src/domain/view/viewDiff.ts
mormubis: Done!

packages/rum-core/src/transport/startRumBatch.ts
BenoitZugmeyer: * not sure what you mean by required fields being implicit? all the fields defined in the object passed to `combine` would be enforced. Maybe a comment would help clarifying that those are required?

* `combine` does exist, we use it extensively

packages/rum-core/src/domain/view/viewDiff.ts
BenoitZugmeyer: maybe `if (a === null || type a !== 'object' || b === null || typeof b !== 'object') return a===b` ?

BenoitZugmeyer: or maybe use [`isIndexableObject`](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/utils/typeUtils.ts#L34-L36)

```ts
if (!isIndexableObject(a) || !isIndexableObject(b)) return a===b
```

packages/rum-core/src/transport/startRumBatch.ts
BenoitZugmeyer: Well, a bit of logic is repeated in you current code, which makes things unnecessarily lengthy. Also there this some case:

```ts
 if (!lastSentView) {
      // Safety fallback (should not happen in practice)
      lastSentView = serverRumEvent
      batch.upsert(serverRumEvent, viewId)
      return
    }
```

which seem incomplete because it doesn't reset `viewUpdatesSinceCheckpoint`. It's probably fine as you stated that it shouldn't happen in practice, but then either it happens and we should make sure it behaves correctly, or it shouldn't and we remove it.

It feels like you are struggling because of imprecise typings. How can we improve the schema so types are better defined?

packages/rum-core/src/domain/view/viewDiff.ts
mormubis: I went with the inline version. `isIndexableObject` returns false for arrays, so `isEqual([1, 2], [1, 2])` with different references returns false. Any array in `replaceKeys` would always end up in the diff even when unchanged. What do you think?

packages/rum-core/src/transport/startRumBatch.ts
mormubis: Removed the fallback. Replaced `as any` with `as RumViewEvent` since `is_active` is already typed there. Not sure if there's a better way to improve the schema to avoid the cast entirely. What do you think?

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -0,0 +1,208 @@
...
+  view: { id: string; is_active?: boolean; [key: string]: unknown }
+  _dd: { document_version: number; [key: string]: unknown }
+  [key: string]: unknown
+}
+
+// Helper: extract view_update events from all RUM events
+// (intakeRegistry.rumViewEvents only returns type==='view')
+function getViewUpdateEvents(intakeRegistry: IntakeRegistry): ViewUpdateEvent[] {
+  return intakeRegistry.rumEvents.filter((e) => (e.type as string) === 'view_update') as unknown as ViewUpdateEvent[]
+}
```
thomas-lebeau: **💬 suggestion:** ‏view_update is here to stay, let's just add `rumViewUpdateEvents` to `intakeRegistry`, it will be nicer

test/e2e/scenario/rum/actions.scenario.ts
```diff
@@ -237,8 +237,8 @@ test.describe('action collection', () => {
       const viewEvents = intakeRegistry.rumViewEvents
       const originalViewEvent = viewEvents.find((view) => view.view.url.endsWith('/'))!
       const otherViewEvent = viewEvents.find((view) => view.view.url.endsWith('/other-view'))!
-      expect(originalViewEvent.view.action.count).toBe(1)
-      expect(otherViewEvent.view.action.count).toBe(0)
+      expect(originalViewEvent.view.action?.count).toBe(1)
+      expect(otherViewEvent.view.action?.count).toBe(0)
```
thomas-lebeau: **❓ question:** ‏isn't `view.action` required? as per the schema?

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -0,0 +1,208 @@
...
+      ].sort((a, b) => a._dd.document_version - b._dd.document_version)
+
+      expect(allViewRelatedEvents.length).toBeGreaterThanOrEqual(2)
+
+      // Verify monotonic increase
+      for (let i = 1; i < allViewRelatedEvents.length; i++) {
+        expect(allViewRelatedEvents[i]._dd.document_version).toBeGreaterThan(
+          allViewRelatedEvents[i - 1]._dd.document_version
+        )
+      }
```
thomas-lebeau: **🔨 warning:** ‏If you sort them (line 66, of course you'll have them sorted)

also, because of batching and retry, this is not guarantee to arrive in order, is it?

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -0,0 +1,208 @@
...
+      <a id="nav-link">Navigate</a>
+      <script>
+        document.getElementById('nav-link').addEventListener('click', () => {
+          history.pushState(null, '', '/new-page')
+        })
+      </script>
+    `)
+    .run(async ({ intakeRegistry, flushEvents, page }) => {
+      // Trigger a route change to create a new view
+      await page.click('#nav-link')
```
thomas-lebeau: **💬 suggestion:** ‏can't we do simpler:

```suggestion
      await page.goto('/new-page')
```

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -0,0 +1,208 @@
...
+      <a id="nav-link">Navigate</a>
+      <script>
+        document.getElementById('nav-link').addEventListener('click', () => {
+          history.pushState(null, '', '/other-page')
+        })
+      </script>
+    `)
+    .run(async ({ intakeRegistry, flushEvents, page }) => {
+      // Navigate to trigger view end on the first view
+      await page.click('#nav-link')
```
thomas-lebeau:
```suggestion
      await page.goto('/other-page')
```

packages/rum-core/src/domain/view/viewDiff.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  it('should return false for different primitives', () => {
+    expect(isEqual(1, 2)).toBe(false)
+    expect(isEqual('a', 'b')).toBe(false)
+    expect(isEqual(true, false)).toBe(false)
+    expect(isEqual(null, undefined)).toBe(false)
+  })
+
+  it('should return true for deeply equal objects', () => {
+    expect(isEqual({ a: 1, b: { c: 2 } }, { a: 1, b: { c: 2 } })).toBe(true)
```
thomas-lebeau, line 20: **💬 suggestion:** ‏should we have a test for two identical object but with the key in a different order?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -49,6 +49,10 @@ export function startRumAssembly(
       ...VIEW_MODIFIABLE_FIELD_PATHS,
       ...ROOT_MODIFIABLE_FIELD_PATHS,
     },
+    // view_update events are created post-assembly in startRumBatch.ts and never go through
+    // this pipeline — they intentionally bypass beforeSend. This entry is required by the
+    // exhaustive type but is never reached in practice.
+    [RumEventType.VIEW_UPDATE]: {},
```
thomas-lebeau, line 55: **💬 suggestion:** ‏let's add a mention about the fact that the full view already went through this (as `RumEventType.VIEW`) so the modified fields are already modified eventually in beforeSend. (e.g. if customer overwrite `view.performance.lcp.resource_url` to remove PII, the view update will contain this overwrite already)

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,71 @@
...
+  })
+
+  if (!diff) {
+    return undefined
+  }
+
+  const currentView = currentObj.view as Record<string, unknown>
+  const currentDd = currentObj._dd as Record<string, unknown>
+
+  // Merge always-required fields on top of the diff for backend routing
```
thomas-lebeau: **💬 suggestion:** ‏
```suggestion
  // Restore the ignoreKeys — backend needs them on every event
```

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,71 @@
...
 import type { LifeCycle } from '../domain/lifeCycle'
 import { LifeCycleEventType } from '../domain/lifeCycle'
 import type { AssembledRumEvent } from '../rawRumEvent.types'
+import type { RumViewEvent } from '../rumEvent.types'
 import { RumEventType } from '../rawRumEvent.types'
+import { diffMerge } from '../domain/view/viewDiff'
+
+export const PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL = 100
+
+export function computeAssembledViewDiff(
```
thomas-lebeau: **🥜 nitpick:** maybe rename to ‏`assembleViewUpdateEvent()`, or `computeAssembledViewUpdateEvent()` or `computeViewUpdateEvent()`

test/e2e/scenario/rum/actions.scenario.ts
mormubis: It's optional in the generated types. The `?.` is just TS complaining because `RumViewUpdateEvent` joined the union. `rumViewEvents` filters for `type === 'view'` so `action` is always there at runtime.

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -28,12 +89,61 @@ export function startRumBatch(
...
+      return
+    }
+
+    const viewId = serverRumEvent.view.id
+
+    // New view started
+    if (viewId !== lastSentView?.view.id) {
+      lastSentView = serverRumEvent
+      viewUpdatesSinceCheckpoint = 0
+      batch.upsert(serverRumEvent, viewId)
```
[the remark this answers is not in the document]
thomas-lebeau: This sound relevant, does the backend supports this properly?

test/e2e/lib/framework/intakeRegistry.ts
```diff
@@ -193,6 +197,10 @@ function isRumViewEvent(event: RumEvent): event is RumViewEvent {
   return event.type === 'view'
 }
 
+function isRumViewUpdateEvent(event: RumEvent): boolean {
+  return (event.type as string) === 'view_update'
+}
```
thomas-lebeau: **💬 suggestion:** ‏for consistency with the other methods and to avoid typing each `intakeRegistry.rumViewUpdateEvents` 

```suggestion
function isRumViewUpdateEvent(event: RumEvent): event is RumViewUpdateEvent {
  return event.type === 'view_update'
}
```

test/e2e/scenario/rum/actions.scenario.ts
thomas-lebeau: but these are `RumViewEvent` here, and we did not changed this type, did we? What am I missing?

mormubis: Sorry, that was in an old version of the types (where they were mark as optional). You are right it didn't change.

packages/rum-core/src/transport/startRumBatch.ts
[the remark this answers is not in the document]
mormubis: The backend handles out of order events as long as they go in the same batch. They get divided into multiple tracks and ordered by `document_version`, not payload position.

thomas-lebeau (approved):

packages/rum-core/src/transport/startRumBatch.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+describe('startRumBatch partial_view_updates routing', () => {
+  beforeEach(() => {
+    addExperimentalFeatures([ExperimentalFeature.PARTIAL_VIEW_UPDATES])
+    registerCleanupTask(resetExperimentalFeatures)
+  })
+
+  it('PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL should be 100', () => {
+    expect(PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL).toBe(100)
+  })
+})
```
BenoitZugmeyer: **suggestion:** unit testing constants is unnecessary

packages/rum-core/src/domain/assembly.ts
```diff
@@ -49,6 +49,12 @@ export function startRumAssembly(
       ...VIEW_MODIFIABLE_FIELD_PATHS,
       ...ROOT_MODIFIABLE_FIELD_PATHS,
     },
+    // view_update events are created post-assembly in startRumBatch.ts and never go through
+    // this pipeline. The full view event already went through assembly (as RumEventType.VIEW),
+    // so any beforeSend modifications (e.g. PII scrubbing on view.performance.lcp.resource_url)
+    // are already reflected in the view_update diff. This entry is required by the exhaustive
+    // type but is never reached in practice.
+    [RumEventType.VIEW_UPDATE]: {},
```
BenoitZugmeyer: **suggestion:** remove this. We never have view updates coming through the lifecycle.

packages/rum-core/src/domain/trackEventCounts.ts
```diff
@@ -30,7 +30,7 @@ export function trackEventCounts({
   }
 
   const subscription = lifeCycle.subscribe(LifeCycleEventType.RUM_EVENT_COLLECTED, (event): void => {
-    if (event.type === 'view' || event.type === 'vital' || !isChildEvent(event)) {
+    if (event.type === 'view' || event.type === 'view_update' || event.type === 'vital' || !isChildEvent(event)) {
```
BenoitZugmeyer: **suggestion:** same here, view updates are not coming there, we can remove this condition

packages/rum-core/src/transport/startRumBatch.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+    })
+    const result = assembleViewUpdateEvent(current, last)!
+
+    expect(result.type).toBe(RumEventType.VIEW_UPDATE)
+    expect((result as any).application).toEqual({ id: 'app-1' })
+    expect((result as any).session).toEqual({ id: 'sess-1', type: 'user' })
+    expect((result.view as any).id).toBe('view-1')
+    expect((result.view as any).url).toBe('/home')
+    expect((result._dd as any).document_version).toBe(2)
+    expect((result._dd as any).format_version).toBe(2)
```
BenoitZugmeyer: **suggestion:** could we avoid those `as any`? Something like:

```ts
    expect(result.type).toBe(RumEventType.VIEW_UPDATE)
    const viewUpdate = result as RumViewUpdateEvent
    expect(viewUpdate.application).toEqual({ id: 'app-1' })
    expect(viewUpdate.session).toEqual({ id: 'sess-1', type: 'user' })
    expect(viewUpdate.view.id).toBe('view-1')
    expect(viewUpdate.view.url).toBe('/home')
    expect(viewUpdate._dd.document_version).toBe(2)
    expect(viewUpdate._dd.format_version).toBe(2)
```

packages/rum-core/src/transport/startRumBatch.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+        time_spent: 5000,
+      },
+    })
+    const result = assembleViewUpdateEvent(current, last)!
+
+    expect((result.view as any).action).toEqual({ count: 3 }) // changed
+    expect((result.view as any).time_spent).toBe(5000) // changed
+    expect((result.view as any).error).toBeUndefined() // unchanged, stripped
+    expect((result.view as any).name).toBeUndefined() // unchanged, stripped
+    expect((result.view as any).url).toBe('/home') // required routing field, always present
```
BenoitZugmeyer: **suggestion:** same here

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,71 @@
...
+export function assembleViewUpdateEvent(
+  current: AssembledRumEvent,
+  last: AssembledRumEvent
+): AssembledRumEvent | undefined {
+  const currentObj = current as unknown as Record<string, unknown>
+  const lastObj = last as unknown as Record<string, unknown>
+
+  const diff = diffMerge(currentObj, lastObj, {
+    // context, connectivity, usr, device, privacy are objects — use REPLACE to avoid partial updates
+    replaceKeys: new Set(['view.custom_timings', 'context', 'connectivity', 'usr', 'device', 'privacy']),
```
BenoitZugmeyer, line 28: **question:** I'm surprised we want to REPLACE connectivity, device and privacy. Is there any backend spec for that?

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,71 @@
...
+import type { RumViewEvent } from '../rumEvent.types'
 import { RumEventType } from '../rawRumEvent.types'
+import { diffMerge } from '../domain/view/viewDiff'
+
+export const PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL = 100
+
+export function assembleViewUpdateEvent(
+  current: AssembledRumEvent,
+  last: AssembledRumEvent
+): AssembledRumEvent | undefined {
```
BenoitZugmeyer: **suggestion:** Types around this function could be greatly improved like this:

```ts
export function assembleViewUpdateEvent(current: RumViewEvent, last: RumViewEvent): RumViewUpdateEvent | undefined {
  const diff = diffMerge(current, last, {
    // context, connectivity, usr, device, privacy are objects — use REPLACE to avoid partial updates
    replaceKeys: new Set(['view.custom_timings', 'context', 'connectivity', 'usr', 'device', 'privacy']),
    appendKeys: new Set(['_dd.page_states']),
    // Ignore always-required fields — they are added back via combine regardless of changes
    ignoreKeys: new Set([
      'date',
      'type',
      'application',
      'session',
      'view.id',
      'view.url',
      '_dd.document_version',
      '_dd.format_version',
    ]),
  })

  if (!diff) {
    return undefined
  }

  // Restore the ignoreKeys — backend needs them on every event
  return combine(diff, {
    type: RumEventType.VIEW_UPDATE,
    date: current.date,
    application: current.application,
    session: current.session,
    view: {
      id: current.view.id,
      url: current.view.url,
    },
    _dd: {
      document_version: current._dd.document_version,
      format_version: current._dd.format_version,
    },
  })
}

```

By using the proper types `RumViewEvent` and `RumViewUpadetEvent`, you can remove all casts.

packages/rum-core/src/transport/startRumBatch.ts
```diff
@@ -28,12 +89,61 @@ export function startRumBatch(
     }),
   })
 
+  let lastSentView: AssembledRumEvent | undefined
```
BenoitZugmeyer: **issue:** keeping track of a single view is an issue because it is possible to get updates for a view after it becomes inactive

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -0,0 +1,168 @@
...
+      // The checkpoint (every 100 updates) uses batch.upsert with the same viewId which would
+      // replace the initial view if they share a batch — flushing first prevents that.
+      // Dispatching beforeunload triggers the SDK batch send without navigating away.
+      await page.evaluate(() => window.dispatchEvent(new Event('beforeunload')))
+      await waitForRequests(page)
+
+      // Use setViewName to trigger unthrottled view updates (unlike addAction which is
+      // throttled to THROTTLE_VIEW_UPDATE_PERIOD=3s, setViewName calls triggerViewUpdate directly).
+      // We need more than PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL (100) updates to trigger a checkpoint.
+      // All calls are batched in a single evaluate to avoid 102 round-trips to the browser.
```
BenoitZugmeyer, line 154: **suggestion:** This is likely a bug, no? `setViewName` should schedule a view update instead of sending the update directly.

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      // Should have at least one view_update
+      const viewUpdateEvents = intakeRegistry.rumViewUpdateEvents
+      expect(viewUpdateEvents.length).toBeGreaterThanOrEqual(1)
+
+      // All events share the same view.id
+      const viewId = viewEvents[0].view.id
+      for (const update of viewUpdateEvents) {
+        expect(update.view.id).toBe(viewId)
+      }
```
BenoitZugmeyer: **suggestion:** add an assertion to show that a view update contains the updated `view.action.count` field, so we understand why we have an `addAction` above

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -34,6 +36,7 @@ export type RumEventType = (typeof RumEventType)[keyof typeof RumEventType]
 
 export type AssembledRumEvent = (
   | RumViewEvent
+  | RumViewUpdateEvent
```
BenoitZugmeyer: **suggestion:** This types depicts the RUM events that are assembled by `assembly.ts`. Since ViewUpdate are not assembled by `assembly.ts`, it shouldn't be listed there.

packages/rum-core/src/transport/startRumBatch.ts
mormubis: The [SDK Implementation Guide](https://datadoghq.atlassian.net/wiki/spaces/~6318f9106a8bb4eda678a4b6/pages/6280675805/SDK+Implementation+Guide+VIEW_UPDATE+Events) defines which fields use REPLACE vs MERGE. `connectivity`, `device` and `privacy` are REPLACE there.

packages/rum-core/src/transport/startRumBatch.ts
mormubis: As we discussed offline, there can only be one active view. When `startView` is called, the previous view gets `is_active: false` immediately. The view-end always sends a full upsert before the diff logic. One edge case: if a late view-end for A arrives after B started, it clears B's `lastSentView`. Not a correctness issue but B's next update would be a full upsert instead of a diff. I can fix that by only clearing when the ending view matches the tracked view.

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
mormubis: I think you're right. `setViewName` calls `triggerViewUpdate` directly instead of `scheduleViewUpdate`. I'll file it separately and fix forward.

packages/rum-core/src/transport/startRumBatch.spec.ts
mormubis: Removed!

packages/rum-core/src/transport/startRumBatch.ts
BenoitZugmeyer: Alright!

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -27,7 +26,6 @@ export const RumEventType = {
   ERROR: 'error',
   LONG_TASK: 'long_task',
   VIEW: 'view',
-  VIEW_UPDATE: 'view_update',
```
BenoitZugmeyer: **nitpick:** I wonder why you removed this variant?

BenoitZugmeyer (approved):

packages/rum-core/src/transport/startRumBatch.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+      },
+    })
+    const currentService = current.service
+    assembleViewUpdateEvent(current, last)
+
+    expect(current.service).toBe(currentService)
+  })
+})
+
+describe('startRumBatch partial_view_updates routing', () => {
```
BenoitZugmeyer: **suggestion:** this `describe` is empty, you could remove it

BenoitZugmeyer (approved):

packages/rum-core/src/rawRumEvent.types.ts
mormubis: Added it back!

PR #4204, opened 2026-02-18
♻️ Simplify session managers by removing product-specific wrappers

thomas-lebeau:
## Motivation

The `rumSessionManager` and `logsSessionManager` were thin wrappers around the core `SessionManager` that added minimal product-specific logic (tracked session lookup and session replay state computation). This indirection made the session management code harder to follow and maintain, with each product duplicating similar patterns.

This PR consolidates session management by moving the shared concepts (`findTrackedSession`, `TrackedSession`, `startSessionManagerStub`) directly into the core `SessionManager` and extracting session replay state computation into a dedicated module. This eliminates the product-specific session manager files entirely.

## Changes

**Core (`packages/core`)**
- Added `findTrackedSession()`, `TrackedSession` type, and `startSessionManagerStub()` to the core `SessionManager`
- Exported these new utilities from the core package index

**RUM (`packages/rum-core`, `packages/rum`)**
- Extracted `SessionReplayState` and `computeSessionReplayState()` into a new `sessionReplayState.ts` module
- Removed `rumSessionManager.ts` entirely (133 lines deleted)
- Updated all RUM consumers to use the core `SessionManager` directly plus `computeSessionReplayState` where needed
- Simplified `rumSessionManager.spec.ts` to test only session replay state computation

**Logs (`packages/logs`)**
- Removed `logsSessionManager.ts` entirely (69 lines deleted)
- Updated all Logs consumers to use the core `SessionManager` directly
- Simplified mock session manager in tests

**Impact on types throughout**
- Replaced `RumSessionManager` / `LogsSessionManager` types with `SessionManager` from core
- Updated function signatures, test mocks, and context providers across both products

## Test instructions

- `yarn test:unit` -- all existing tests should pass
- Review that session replay state logic (forced replay, sampling) is preserved correctly in the extracted `sessionReplayState.ts`

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/boot/recorderApi.spec.ts
+++ b/packages/rum/src/boot/recorderApi.spec.ts
@@ -383,6 +383,10 @@ describe('makeRecorderApi', () => {
     describe('when session renewal change the tracking type', () => {
       describe('from WITHOUT_REPLAY to WITH_REPLAY', () => {
         beforeEach(() => {
+          // These tests rely on deterministic hash-based sampling which requires BigInt
+          if (typeof BigInt === 'undefined') {
+            pending('BigInt not supported')
+          }
           sessionManager.setId(HIGH_HASH_UUID)
           setupRecorderApi({ sessionManager, sessionReplaySampleRate: 50 })
         })
@@ -446,6 +450,9 @@ describe('makeRecorderApi', () => {
 
       describe('from WITH_REPLAY to WITHOUT_REPLAY', () => {
         beforeEach(() => {
+          if (typeof BigInt === 'undefined') {
+            pending('BigInt not supported')
+          }
           setupRecorderApi({ sessionManager, sessionReplaySampleRate: 50 })
         })
 
@@ -555,6 +562,9 @@ describe('makeRecorderApi', () => {
 
       describe('from untracked to WITHOUT_REPLAY', () => {
         beforeEach(() => {
+          if (typeof BigInt === 'undefined') {
+            pending('BigInt not supported')
+          }
           sessionManager.setNotTracked()
           setupRecorderApi({ sessionManager, sessionReplaySampleRate: 50 })
         })


packages/core/src/domain/session/sessionManager.ts
```diff
@@ -112,6 +132,20 @@ export function startSessionManager(
...
+          return
+        }
+        if (!isSampled(session.id, sampleRate)) {
+          return
+        }
+        return {
+          id: session.id,
+          anonymousId: session.anonymousId,
+          isReplayForced: session.isReplayForced,
+        }
```
BenoitZugmeyer: **suggestion:** `return session`

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -369,20 +377,29 @@ describe('makeRecorderApi', () => {
...
+      configuration = setupRecorderApi({ sessionManager })
     })
 
+    function setReplayOn() {
+      configuration.sessionReplaySampleRate = 100
+    }
+
+    function setReplayOff() {
+      configuration.sessionReplaySampleRate = 0
+    }
```
BenoitZugmeyer: **suggestion:** we don't usually mutate the configuration. `setupRecorderApi` already support providing a `sessionReplaySampleRate`. What about calling `setupRecorderApi` in each tests instead of `beforeEach`, so you can define the sample rate to use?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -141,6 +165,33 @@ export function startSessionManager(
...
+    isReplayForced: false,
+    anonymousId: undefined,
+  }
+  onReady({
+    findSession: () => sessionContext,
+    findTrackedSession: () => ({
+      id: stubSessionId,
+      anonymousId: sessionContext.anonymousId,
+      isReplayForced: sessionContext.isReplayForced,
+    }),
```
BenoitZugmeyer: **suggestion:** 
```suggestion
    findTrackedSession: () => sessionContext,
```

packages/rum/src/boot/postStartStrategy.ts
```diff
@@ -151,16 +146,20 @@ export function createPostStartStrategy(
...
 
-function canStartRecording(session: RumSession | undefined, options?: StartRecordingOptions) {
-  return !session || (session.sessionReplay === SessionReplayState.OFF && (!options || !options.force))
+function canStartRecording(
+  session: SessionContext | undefined,
+  replayState: SessionReplayState | undefined,
+  options?: StartRecordingOptions
+) {
+  return !session || (replayState === SessionReplayState.OFF && (!options || !options.force))
 }
```
BenoitZugmeyer: **suggestion:** remove the `session` argument, as it's redundant with `replayState` (`replayState` is undefined if `session` is undefined)

```ts
function canStartRecording(
  replayState: SessionReplayState | undefined,
  options?: StartRecordingOptions
) {
  return !replayState || (replayState === SessionReplayState.OFF && (!options || !options.force))
}
```

Also, it might be nice to change the function so it matches its name:

```ts
function canStartRecording(
  replayState: SessionReplayState | undefined,
  options?: StartRecordingOptions
) {
  return replayState && (replayState !== SessionReplayState.OFF || options?.force)
}
```
(and use it like `if (!canStartRecording(...))`

sethfowler-datadog: Wow, I hadn't noticed that inconsistency before. Yes, it'd be a huge win to make this function match its name!

packages/rum/src/boot/recorderApi.spec.ts
```diff
@@ -383,6 +383,10 @@ describe('makeRecorderApi', () => {
     describe('when session renewal change the tracking type', () => {
       describe('from WITHOUT_REPLAY to WITH_REPLAY', () => {
         beforeEach(() => {
+          // These tests rely on deterministic hash-based sampling which requires BigInt
+          if (typeof BigInt === 'undefined') {
+            pending('BigInt not supported')
+          }
```
thomas-lebeau, line 389: Currently If the browser does not supported BigInt we fallback to non-deterministic sampling, so this test is also non deterministic.

for v7 we plan to drop support for browser that don't support BigInt so we will be able to clean that up in a latter PR.

packages/rum-core/src/boot/preStartRum.ts
```diff
@@ -184,11 +185,17 @@ export function createPreStartStrategy(
...
       }
 
-      const startSessionManagerFn = canUseEventBridge() ? startRumSessionManagerStub : mockable(startRumSessionManager)
-      startSessionManagerFn(configuration, trackingConsentState, (newSessionManager) => {
-        sessionManager = newSessionManager
-        tryStartRum()
-      })
+      if (canUseEventBridge()) {
+        startSessionManagerStub((newSessionManager) => {
+          sessionManager = newSessionManager
```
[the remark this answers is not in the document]
thomas-lebeau: fixed, thanks

PR #4225, opened 2026-02-23
💥 [RUM-970] do not collect network errors for aborted requests

BenoitZugmeyer:
## Motivation

Requests canceled by the application via `AbortController.abort()` (or `xhr.abort()`) are intentional — sending them as network errors is unexpected and noisy for customers. This is option 1 from RUM-970, implemented now that we're in a major version.

## Changes

- `networkErrorCollection.ts`: exclude aborted requests from `isNetworkError()` by checking `!request.isAborted` (applies to both XHR and Fetch)

## Test instructions

1. Open the sandbox (`yarn dev`, navigate to `http://localhost:8080`)
2. With `forwardErrorsToLogs: true`, run `fetch('https://httpbin.org/status/500')` → should produce a network error log
3. Run the same request with `AbortController.abort()` → should produce no log

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/logs/src/domain/networkError/networkErrorCollection.ts
+++ b/packages/logs/src/domain/networkError/networkErrorCollection.ts
@@ -57,7 +57,6 @@ export function startNetworkErrorCollection(configuration: LogsConfiguration, li
         : request.responseBody || 'Failed to load'
 
     const domainContext: LogsEventDomainContext<typeof ErrorSource.NETWORK> = {
-      isAborted: request.isAborted,
       handlingStack: request.handlingStack,
     }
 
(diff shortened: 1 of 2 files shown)


packages/logs/src/domain/networkError/networkErrorCollection.ts
rgaignault: ❓ Question : do we still need this  property ?

BenoitZugmeyer: hum good point, I guess `isAborted` will always be false then. Let's remove it

PR #4226, opened 2026-02-23
💥 enable propagateTraceBaggage by default

BenoitZugmeyer:
## Motivation

The `baggage` header propagates RUM session context (session ID, user ID, account ID) to backend services for distributed tracing correlation. Requiring customers to explicitly opt in creates unnecessary friction — it should be on by default, with an explicit opt-out for those who don't want it.

## Changes

- `configuration.ts`: change default from `false` to `true` by using `!== false` instead of `!!`
- `tracer.spec.ts`: update expected headers to include `baggage` by default; use `toPlainObject()` for the `Request` case to avoid alphabetical ordering sensitivity
- `tracing.scenario.ts`: update e2e scenarios — baggage is now expected in `checkRequestHeaders` by default, rename the old "propagate trace baggage" test to cover user/account specifically, add a new test for `propagateTraceBaggage: false`

## Test instructions

1. Open the sandbox (`yarn dev`, navigate to `http://localhost:8080`)
2. Add `allowedTracingUrls: ['https://httpbin.org'], service: 'sandbox'` to `DD_RUM.init()` in `sandbox/index.html`
3. Run `fetch('https://httpbin.org/headers').then(r => r.json()).then(j => console.log(j.headers))` — verify `Baggage: session.id=...` is present
4. Add `propagateTraceBaggage: false` — reload and repeat — verify `Baggage` is absent

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/scenario/rum/tracing.scenario.ts
+++ b/test/e2e/scenario/rum/tracing.scenario.ts
@@ -69,11 +69,10 @@ test.describe('tracing', () => {
       checkTraceAssociatedToRumEvent(intakeRegistry)
     })
 
-  createTest('propagate trace baggage')
+  createTest('propagate trace baggage with user and account')
     .withRum({
       service: 'service',
       allowedTracingUrls: ['LOCATION_ORIGIN'],
-      propagateTraceBaggage: true,
       enableExperimentalFeatures: ['user_account_trace_header'],
     })
     .run(async ({ intakeRegistry, flushEvents, page }) => {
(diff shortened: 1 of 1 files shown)


test/e2e/scenario/rum/tracing.scenario.ts
```diff
@@ -88,7 +87,28 @@ test.describe('tracing', () => {
...
+        window
+          .fetch('/headers')
+          .then((response) => response.text())
+          .catch(() => new Error('Fetch request failed!'))
+      )
+      const headers = parseHeaders(rawHeaders)
+      checkRequestHeaders(headers, { withBaggage: false })
       await flushEvents()
       checkTraceAssociatedToRumEvent(intakeRegistry)
     })
```
thomas-lebeau, line 114: **❓ question:** ‏what happen when:
- `propagateTraceBaggage: true` and no user and account context set?
- `propagateTraceBaggage: false` and user and account context is set

should we also test for these?

BenoitZugmeyer: > * `propagateTraceBaggage: true` and no user and account context set?

This is already covered by various test cases, since `propagateTraceBaggage: true` is the default now. [This assertion](https://github.com/DataDog/browser-sdk/blob/001e331368f5c711756458aa5c8dcfaf8dd2b0cd/test/e2e/scenario/rum/tracing.scenario.ts#L139) ensure the baggage header is set with the session id.

> * `propagateTraceBaggage: false` and user and account context is set

I don't think we really need to test for it, since in this case the header should just not be present?

thomas-lebeau (approved):

PR #4234, opened 2026-02-25
[NextJS] Nextjs app router integration (AI optimistic)

lierniel:
# ✨ Add Next.js App Router integration for `@datadog/browser-rum-react`

## Motivation

Next.js App Router is a widely adopted routing paradigm (Next.js 13.4+), but RUM view tracking requires manual instrumentation today. This integration provides a `<DatadogRumProvider>` component that automatically tracks client-side navigations as RUM views with parameterized route names (e.g., `/users/:id`), following the same sub-package pattern as the existing React Router integrations.

## Changes

**Plugin configuration (`packages/rum-react/src/domain/reactPlugin.ts`):**
- Add `nextAppRouter?: boolean` option to `ReactPluginConfiguration`
- Automatically set `trackViewsManually = true` when `nextAppRouter` is enabled
- Warn when both `router` and `nextAppRouter` are enabled simultaneously
- Report `nextAppRouter` in configuration telemetry

**View name computation (`packages/rum-react/src/domain/nextjs/computeNextViewName.ts`):**
- Segment-aware param replacement: reconstructs parameterized route patterns from `pathname` + `params` (e.g., `/users/123` + `{ id: '123' }` → `/users/:id`)
- Handles catch-all segments (`[...slug]` → `/:slug`), multi-occurrence params, and edge cases (empty/undefined values)
- Array params processed before string params to avoid partial replacements

**Provider component (`packages/rum-react/src/domain/nextjs/datadogRumProvider.tsx`):**
- Client-side `DatadogRumProvider` component that reads `usePathname()` / `useParams()` from `next/navigation`
- Uses ref-based deduplication to prevent duplicate views on re-renders
- Defers view start via `onRumInit` subscriber pattern (safe for any init order)
- Deduplicates by computed view name (not raw pathname), so `/users/123` → `/users/456` does not start a new view if both resolve to `/users/:id`

**Sub-package entry point:**
- `packages/rum-react/nextjs/package.json` — shim pointing to compiled artifacts (same pattern as `react-router-v6/`, `react-router-v7/`)
- `packages/rum-react/src/entries/nextjs.ts` — entry point exporting `DatadogRumProvider`

**Sample test app (`test/apps/nextjs-app/`):**
- Simulates Next.js routing via `mockNextNavigation.tsx` (mock `usePathname`, `useParams`, `useNavigate` backed by React context)
- Routes: home (`/`), dynamic param (`/user/:id`), tracked component (`/tracked`), error boundary (`/error`)
- Uses `DatadogRumProvider`, `ErrorBoundary`, `UNSTABLE_ReactComponentTracker` from the SDK
- Registered in `scripts/build/build-test-apps.ts` and `test/e2e/lib/framework/sdkBuilds.ts`

**E2E tests (`test/e2e/scenario/reactPlugin.scenario.ts`):**
- View name tracking: navigate to `/user/42`, verify last view name is `/user/:id`
- Component render vital: navigate to tracked page, verify `TrackedPage` vital with duration
- Error boundary: trigger error, verify error event with `framework: 'react'` and component stack

## Test instructions

1. Run unit tests: `yarn test:unit --spec packages/rum-react/src/domain/nextjs/computeNextViewName.spec.ts` and the other spec files in `packages/rum-react/src/domain/nextjs/`
2. Run plugin tests: `yarn test:unit --spec packages/rum-react/src/domain/reactPlugin.spec.ts`
3. Build test apps: `yarn build:apps`
4. Run E2E tests: `yarn test:e2e -g "Next.js App Router"`
5. Verify typecheck: `yarn typecheck`

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-react/src/domain/nextjs/datadogRumProvider.spec.tsx
```diff
@@ -0,0 +1,144 @@
...
+    expect(startViewSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it('starts a new view when the view name changes', () => {
+    mockPathname = '/users/123'
+    mockParams = { id: '123' }
+
+    let forceUpdate: () => void
+
+    function App() {
```
BeltranBulbarellaDD, line 75: I'm not sure I love the idea of doing a component in a unit test. I would prefer to put it in a e2e test and here test behaviour.

packages/rum-react/src/domain/nextjs/datadogRumProvider.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+  if (viewNameRef.current !== viewName) {
+    viewNameRef.current = viewName
+    startNextjsView(viewName)
+  }
+
+  return <>{children}</>
+}
+
+DatadogRumProvider.displayName = 'DatadogRumProvider'
```
BeltranBulbarellaDD, line 22: **Question:** What is `displayName` this for?

packages/rum-react/src/entries/nextjs.ts
```diff
@@ -0,0 +1,32 @@
...
+ * })
+ *
+ * // app/layout.tsx
+ * import { DatadogRumProvider } from '@datadog/browser-rum-react/nextjs'
+ *
+ * export default function RootLayout({ children }: { children: React.ReactNode }) {
+ *   return (
+ *     <html>
+ *       <body>
+ *         <DatadogRumProvider>{children}</DatadogRumProvider>
```
BeltranBulbarellaDD, line 25: I am stealing the following question since it was asked to me too:
**Question:** Why is this a Provider if it's not wrapping anything? Could it be a function call too?

packages/rum-react/package.json
```diff
@@ -38,6 +42,7 @@
   "devDependencies": {
     "@types/react": "19.2.14",
     "@types/react-dom": "19.2.3",
+    "next": "15.2.2",
```
BeltranBulbarellaDD, line 45: **Question:**: So we will support next >= 15.2.2. Shouldn't we try to support sooner versions too? 
**Suggestion:**: For example support version >= 13?

test/apps/nextjs-app/app.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { useState } from 'react'
+import ReactDOM from 'react-dom/client'
+import { datadogRum } from '@datadog/browser-rum'
+import {
+  reactPlugin,
+  UNSTABLE_ReactComponentTracker as ReactComponentTracker,
+  ErrorBoundary,
```
BeltranBulbarellaDD, line 7: **Suggestion:** The thing about this is that NextJS has a different signature to report errors. 
`export default function ErrorPage({
  error,
  reset,
}:`
So the React Error Boundary would not have the whole context. So we would need to make something for it.

test/apps/nextjs-app/app.tsx
```diff
@@ -0,0 +1,142 @@
...
+    RUM_CONTEXT?: any
+  }
+}
+
+datadogRum.init({ ...window.RUM_CONFIGURATION, plugins: [reactPlugin({ nextAppRouter: true })] })
+if (window.RUM_CONTEXT) {
+  datadogRum.setGlobalContext(window.RUM_CONTEXT)
+}
+
+function Link({
```
BeltranBulbarellaDD, line 24: **Question:** Isn't this a Next component? Why do we have to re-declare it?

test/apps/nextjs-app/app.tsx
```diff
@@ -0,0 +1,142 @@
...
+          </div>
+        )}
+      >
+        <ComponentWithErrorButton />
+      </ErrorBoundary>
+    </div>
+  )
+}
+
+function App() {
```
BeltranBulbarellaDD, line 116: This is not how App Router works for routes. Each file is it's own route. [Doc](https://nextjs.org/docs/app/getting-started/project-structure#routing-files)

BeltranBulbarellaDD: This is important so that we use NextJS specific behaviour like client and server components and even prefetching and test against it.

test/apps/nextjs-app/mockNextNavigation.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export function useParams(): Params {
+  return useContext(NavigationContext).params
+}
+
+export function useNavigate() {
+  return useContext(NavigationContext).navigate
+}
+
+export function NavigationProvider({
```
BeltranBulbarellaDD, line 33: **Question:** What is this component for? Would customer need to create this file themselves? Shouldn't this be handled on the SDK side?

test/e2e/scenario/reactPlugin.scenario.ts
```diff
@@ -61,4 +63,59 @@ test.describe('react plugin', () => {
         })
     })
   }
+
+  for (const { appName, description } of nextApps) {
+    test.describe(`with ${description}`, () => {
+      createTest('should define a view name with DatadogRumProvider')
+        .withRum()
+        .withReactApp(appName)
```
BeltranBulbarellaDD, line 71: **Question:** Why are using ReactApp and not NextApp. IIUC, here you are serving the static html files to the tests, a NextApp needs it's own server to run so as to leverage NextJS features like prefetching, server components, cache, etc.

test/e2e/scenario/reactPlugin.scenario.ts
```diff
@@ -61,4 +63,59 @@ test.describe('react plugin', () => {
...
+      createTest('should capture react error from error boundary')
+        .withRum()
+        .withReactApp(appName)
+        .run(async ({ page, flushEvents, intakeRegistry, browserName, withBrowserLogs }) => {
+          await page.click('text=Go to Error')
+          await page.click('#error-button')
+
+          // Firefox may delay dispatching error events from React error boundaries,
+          // causing flushEvents() to miss them, this timeout ensures the RUM event is captured.
+          if (browserName === 'firefox') {
```
BeltranBulbarellaDD, line 102: I'm not soo sure about this.

packages/rum-react/src/domain/nextjs/datadogRumProvider.tsx
lierniel: It's just for the React Browser Extension to be able to display the component name in the React Virtual DOM tree. This is optional but nice to have

packages/rum-react/src/entries/nextjs.ts
lierniel: Yeah, indeed, it can be implemented as a hook for example. We can't make it a pure function call because we have to call hooks in it, but hooks approach could work, I agree

packages/rum-react/package.json
lierniel: No, that's just the version we use inside our integration. As it is defined above in the peerDependencies section, this approach supports all the versions >= 13

test/apps/nextjs-app/app.tsx
lierniel: Yeah, tbh I didn't look too much into sample app implementation made by AI, so it probably has a lots of problems which you raised

PR #4250, opened 2026-02-27
Fix OpenFeature resolution reason (STATIC vs DEFAULT)

leoromanovsky:
## Motivation

The OpenFeature provider always returned `reason: 'DEFAULT'` even when a precomputed flag was successfully resolved. Per the [OpenFeature spec](https://openfeature.dev/specification/types#resolution-details), `STATIC` should be used when values come from a pre-evaluated store, and `DEFAULT` only when no flag was found.

## Changes

- Add `hasFlag(flagKey)` method to `PrecomputeClient` to check flag existence
- Use `StandardResolutionReasons` constants from `@openfeature/web-sdk` instead of string literals
- Return `STATIC` reason when a flag is resolved from the store
- Return `DEFAULT` reason when falling back to the default value (no client, or flag not found)
- Replace existing default-only provider tests with comprehensive tests covering both resolved (STATIC) and missing (DEFAULT) scenarios using a real `PrecomputeClient`

## Decisions

- Used `STATIC` over `CACHED` since precomputed values are a static evaluation store, not a cache of dynamic evaluations
- Added `hasFlag()` as a lightweight existence check rather than changing the assignment method signatures

### DIFF
--- a/packages/flagging/src/openfeature/provider.spec.ts
+++ b/packages/flagging/src/openfeature/provider.spec.ts
@@ -1,13 +1,13 @@
 import { StandardResolutionReasons, type EvaluationContext, type Logger } from '@openfeature/core'
+import mockPrecomputedConfig from '../../test/data/precomputed-v1-deobfuscated.json'
+import { offlinePrecomputedInit } from '../precomputeClient'
 import { DatadogProvider } from './provider'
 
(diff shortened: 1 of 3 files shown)


leoromanovsky:
Closing — tracked in follow-up plan

PR #4254, opened 2026-02-27
✨ NextJS App Router Integration

BeltranBulbarellaDD:
## Motivation

Adds NextJS App Router Plugin implementation to use in `instrumentation-client.js` documented in [Readme](https://github.com/DataDog/browser-sdk/pull/4254/changes#diff-098d85b1eae4111bd186dbc0c1b549d9ea9c3f0a05bfb763c4533c12f9c472fa)

## Changes
Discussed in this [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6247321353/RFC+-+Browser-SDK+NextJS+Integration)

Introduces `@datadog/browser-rum-nextjs`, a new package for instrumenting Next.js App Router applications with Datadog RUM.

  **Package** (`packages/rum-nextjs/`)
  - `nextjsPlugin() `— RUM plugin that enables manual view tracking.
  - `DatadogAppRouter `— Client component that starts RUM views on route changes, using usePathname() and useParams() to compute normalized view names
  (e.g. /user/42 → /user/[id], /docs/a/b → /docs/[...slug])
  - `onRouterTransitionStart` — Re-exported by the user from `instrumentation-client.ts` so Next.js calls it on client-side navigations, enabling RUM to
  capture the correct URL (including query parameters) before the browser updates `window.location`
  - `computeViewNameFromParams` — Normalizes dynamic route segments back into Next.js route patterns, processing catch-all params before regular params
   to handle overlapping values

  **E2E infrastructure**
  - Adds a Next.js App Router test app (`test/apps/nextjs-app-router/`) with dynamic routes (/user/[id], /guides/[...slug])
  - Configures the E2E framework to build and serve the Next.js app alongside existing test apps
  - Adds `withNextjsApp()` test builder method for Next.js-specific E2E scenarios

  **E2E tests**
  - Initial view tracking with correct loading_type
  - SPA navigation with `loading_time` measurement
  - Route normalization across dynamic and catch-all routes, with URL and referrer assertions

[Next Steps ](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/edit-v2/6096454155)

## Test instructions

Unit tests in rum-nextjs should pass.
New e2e [tests](https://github.com/DataDog/browser-sdk/pull/4254/changes#diff-9b2401066a98d985a3d1d288ae7a4bf78380e3d873b1c78d4eb9f3fa0c58f81a) should pass

## Checklist

- [X] Tested locally
- [ ] Tested on staging
- [X] Updated documentation and/or relevant AGENTS.md file

### DIFF
--- a/packages/rum-nextjs/src/domain/computeViewNameFromParams.spec.ts
+++ b/packages/rum-nextjs/src/domain/computeViewNameFromParams.spec.ts
@@ -24,7 +24,7 @@ describe('computeViewNameFromParams', () => {
     ['catch-all does not match partial segments',  '/docs-extra/a/b',       { slug: ['a', 'b'] },                      '/docs-extra/[...slug]'],
     // Catch-all processed before regular params (prevents a shared value being consumed by the regular param first)
     ['catch-all takes priority over same-value regular param', '/x/a/b', { id: 'a', slug: ['a', 'b'] }, '/x/[...slug]'],
-    // Ambiguous cases (known limitations — deterministic but not guaranteed to match route pattern)
+    // Repeated param values
     ['same value for multiple params: assigned left-to-right', '/toto/toto', { user: 'toto', view: 'toto' }, '/[user]/[view]'],
     ['param value appears in static segment: first occurrence taken', '/toto/toto', { user: 'toto' }, '/[user]/toto'],
     // Edge cases
--- a/test/e2e/scenario/nextjsPlugin.scenario.ts
+++ b/test/e2e/scenario/nextjsPlugin.scenario.ts
@@ -64,6 +64,8 @@ test.describe('nextjs app router', () => {
 
       await page.click('text=Go to User 42')
       await page.waitForURL('**/user/42?admin=true')
+
+      // Test that the sidebar content is shown on the page
       expect(await page.textContent('[data-testid="sidebar"]')).toContain('Sidebar: User 42')
 
       await page.click('text=Back to Home')


packages/core/src/tools/instrumentMethod.spec.ts
BeltranBulbarellaDD, line 1: This is because of [this](https://github.com/microsoft/TypeScript/issues/43094)

packages/rum-core/src/domain/tracing/identifier.spec.ts
BeltranBulbarellaDD, line 1: Same, because of [this](https://github.com/microsoft/TypeScript/issues/43094)

packages/rum-nextjs/src/domain/datadogRumProvider.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { computeViewNameFromParams } from './computeViewNameFromParams'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogRumProvider({ children }: { children: React.ReactNode }) {
+  const pathname = usePathname()
+  const params = useParams()
+
+  useEffect(() => {
+    startNextjsView(computeViewNameFromParams(pathname, params as Record<string, string | string[] | undefined>))
+  }, [pathname])
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Leftover code from old changes. Fixed it.

eslint.config.mjs
```diff
@@ -461,6 +463,19 @@ export default tseslint.config(
...
 
+  {
+    files: ['test/apps/nextjs-app-router/**'],
+    languageOptions: {
+      globals: globals.browser,
+    },
+    rules: {
+      'import/no-default-export': 'off',
+      'import/no-unresolved': 'off',
+      'unicorn/filename-case': 'off',
```
BeltranBulbarellaDD: Needed for `instrumentation-client.js`

eslint.config.mjs
```diff
@@ -461,6 +463,19 @@ export default tseslint.config(
...
   },
 
+  {
+    files: ['test/apps/nextjs-app-router/**'],
+    languageOptions: {
+      globals: globals.browser,
+    },
+    rules: {
+      'import/no-default-export': 'off',
+      'import/no-unresolved': 'off',
```
BeltranBulbarellaDD: For `next/link`, `next/navigation`

eslint.config.mjs
```diff
@@ -461,6 +463,19 @@ export default tseslint.config(
     },
   },
 
+  {
+    files: ['test/apps/nextjs-app-router/**'],
+    languageOptions: {
+      globals: globals.browser,
+    },
+    rules: {
+      'import/no-default-export': 'off',
```
BeltranBulbarellaDD: For `export default` in `Page.tsx`

packages/rum-nextjs/src/domain/datadogRumProvider.tsx
BenoitZugmeyer: **suggestion:** Don't use an effect hook to start a view. It's run too late. Is there any other way to look for a route change in Next? Else, if we really need to do it within a React component, let's maybe do something like:

```ts
function DatadogRumProvider() {
   const pathname = usePathname()
   const params = useParams()
   const previousPathName = useRef(pathname)
   if (previousPathName.value !== pathname) {
      startNextjsView(...)
   }

  ..
}
```

packages/rum-nextjs/src/domain/datadogRumProvider.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { usePathname, useParams } from 'next/navigation'
+import { computeViewNameFromParams } from './computeViewNameFromParams'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogRumProvider({ children }: { children: React.ReactNode }) {
+  const pathname = usePathname()
+  const params = useParams()
+
+  useEffect(() => {
+    startNextjsView(computeViewNameFromParams(pathname, params as Record<string, string | string[] | undefined>))
```
BenoitZugmeyer: **suggestion:**  Can we use Next.js types instead of casting things here?

packages/rum-nextjs/src/domain/datadogRumProvider.tsx
```diff
@@ -0,0 +1,17 @@
+'use client'
+
+import React, { useEffect } from 'react'
+import { usePathname, useParams } from 'next/navigation'
+import { computeViewNameFromParams } from './computeViewNameFromParams'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogRumProvider({ children }: { children: React.ReactNode }) {
```
BenoitZugmeyer: **thought:** I don't know if it's really a "Provider"... it doesn't provide any context. I'm not sure we really need to use this component to wrap the customer app. Do we have plans for the future?

packages/rum-nextjs/src/domain/nextjsPlugin.ts
```diff
@@ -0,0 +1,70 @@
+import type { RumPlugin, RumPublicApi, StartRumResult } from '@datadog/browser-rum-core'
+
+export interface NextjsPluginConfiguration {
+  router: 'app'
+}
```
BenoitZugmeyer: **question:** the `router` property seems unused, will we need it in a future PR?

packages/rum-nextjs/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@datadog/browser-rum-nextjs",
```
BenoitZugmeyer, line 2: **thought:** I wonder if it shouldn't be named `browser-rum-next`, not `-nextjs`. Although I feel the name isn't specific enough. Maybe `browser-rum-next-plugin`

packages/rum-nextjs/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+datadogRum.init({
+  applicationId: '<APP_ID>',
+  clientToken: '<CLIENT_TOKEN>',
+  site: 'datadoghq.com',
+  plugins: [nextjsPlugin({ router: 'app' })],
+})
+```
+
+## 2. Create a client-side providers component
```
BenoitZugmeyer: **question:** Is this a common pattern? Isn't the `'use client'` within the `datadogRumProvider.ts` module enough?

eslint.config.mjs
BenoitZugmeyer: I don't think it's useful? (I remove this rule and see no eslint issue)

test/e2e/lib/framework/createTest.ts
```diff
@@ -325,12 +334,23 @@ function createTestContext(
...
+    }
+    if (context) {
+      baseUrl.searchParams.set('rum-context', JSON.stringify(context))
+    }
+  } else {
+    baseUrl = new URL(basePath, servers.base.origin)
+    if (hostName) {
+      baseUrl.hostname = hostName
+    }
   }
```
BenoitZugmeyer: thought: this is a bit of a mess :/ every new environment is adding a bit of custom logic here, we should come up with a cleaner pattern.

I try to clean things up a bit here: https://github.com/DataDog/browser-sdk/commit/47fa4be64a2cd8a2ec40739e096de59929185d1f

But really, I feel like the `setup` functions that return a simple HTML string is too limiting now. A setup should be able to represent a whole test environment, including:

* the ability to spawn HTTP servers + define URLs to reach them (ex: adjust the base URL origin)
* the ability to install extensions
* spawning playwright, electron, puppeteer, android emulator... instances

Maybe "setups" should be closer to what we do with extensions

packages/rum-nextjs/README.md
BenoitZugmeyer, line 1: **suggestion:** to avoid maintaining this doc in two places, move it to the integrations-extra repo, like the rum-react package: https://github.com/DataDog/integrations-extras/blob/master/rum_react/README.md

packages/rum-nextjs/README.md
```diff
@@ -0,0 +1,64 @@
+# RUM Browser Monitoring - NEXTJS integration
+
+This package provides NextJS App Router integration for Datadog Browser RUM.
+
+To use please read [instrumentation-client docs][1]
```
drichards-87:
```suggestion
To use it, read the [instrumentation-client docs][1].
```

packages/rum-nextjs/src/domain/datadogRumProvider.tsx
BeltranBulbarellaDD: We can make it a hook or function. I talked about this in the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6247321353/RFC+-+Browser-SDK+NextJS+Integration#Option-1%3A-Expose-a-Provider-Recommended) that it is common react convention to have providers.

packages/rum-nextjs/src/domain/nextjsPlugin.ts
BeltranBulbarellaDD: Yes, we will have Pages router too.

packages/rum-nextjs/README.md
BeltranBulbarellaDD: Discussed in the RFC but tldr:

> because components that use use client still [render partially on the server](https://github.com/reactwg/server-components/discussions/4), as explained in the [NextJS docs on client components](https://nextjs.org/docs/app/building-your-application/rendering/client-components).

So we need to add it for safety.

BeltranBulbarellaDD: But this will be needed for versions of nextJS that don't support `instrumentation-client` (<v15.3). I'll remove the Provider and import it directly.

test/e2e/lib/framework/createTest.ts
BeltranBulbarellaDD: Let me know what you think now! I had to added `baseUrl.search = parsed.search` because of profiling tests which navigate to `'/?js-profiling=true'`

eslint.config.mjs
BeltranBulbarellaDD: Maybe I can revise but see [this job](https://gitlab.ddbuild.io/DataDog/browser-sdk/-/jobs/1469218858).

packages/rum-nextjs/README.md
BeltranBulbarellaDD: I didn't know about the existence of that repository. For the uuid part, is there a specific way to do it? ([e.g](https://github.com/DataDog/integrations-extras/blob/7f1d551644296780e385b5fa210ed237c9f5a985/rum_react/manifest.json#L3))

BeltranBulbarellaDD:
@codex pls review

packages/rum-nextjs/src/domain/nextjsPlugin.ts
BeltranBulbarellaDD: As discussed. I'll remove it. We can discuss when we add Pages Router how to export the 2 different providers.

packages/rum-nextjs/README.md
BeltranBulbarellaDD: Made the draft [PR](https://github.com/DataDog/integrations-extras/pull/2918). Will move it forward once we merge!

test/e2e/lib/helpers/playwright.ts
```diff
@@ -4,6 +4,8 @@ import { getBuildInfos } from '../../../envUtils.ts'
 import packageJson from '../../../../package.json' with { type: 'json' }
 
 export const DEV_SERVER_BASE_URL = `http://localhost:${process.env.DEV_SERVER_PORT}`
+export const NEXTJS_APP_ROUTER_PORT = process.env.NEXTJS_APP_ROUTER_PORT!
+export const NEXTJS_APP_ROUTER_BASE_URL = `http://localhost:${NEXTJS_APP_ROUTER_PORT}`
```
BenoitZugmeyer: **nitpick:** `NEXTJS_APP_ROUTER_BASE_URL` isn't used anymore

BenoitZugmeyer (approved):

packages/rum-nextjs/src/domain/nextjsPlugin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+let globalPublicApi: RumPublicApi | undefined
+let globalAddEvent: StartRumResult['addEvent'] | undefined
+
+const onRumInitSubscribers: InitSubscriber[] = []
+const onRumStartSubscribers: StartSubscriber[] = []
+
+export function nextjsPlugin(): NextjsPlugin {
+  return {
+    name: 'nextjs',
```
BenoitZugmeyer, line 18: **suggestion:** `next`?

packages/rum-nextjs/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@datadog/browser-rum-next-plugin",
```
BenoitZugmeyer: **suggestion:** rename the directory as well (`rum-next-plugin`)

packages/rum-nextjs/src/domain/datadogRumProvider.tsx
```diff
@@ -0,0 +1,19 @@
+'use client'
+
+import React, { useRef } from 'react'
+import { usePathname, useParams } from 'next/navigation'
+import { computeViewNameFromParams } from './computeViewNameFromParams'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogRumProvider({ children }: { children: React.ReactNode }) {
```
BenoitZugmeyer: **suggestion:** Now that the `router` option is removed, maybe rename this component to `DatadogRumAppRouterProvider`?

packages/rum-nextjs/src/domain/nextjsPlugin.ts
BeltranBulbarellaDD: https://nextjs.org/
I have my doubts. 😄

tsconfig.base.json
```diff
@@ -35,6 +35,8 @@
       "@datadog/browser-rum-react/react-router-v6": ["./packages/rum-react/src/entries/reactRouterV6"],
       "@datadog/browser-rum-react/react-router-v7": ["./packages/rum-react/src/entries/reactRouterV7"],
 
+      "@datadog/browser-rum-nextjs": ["./packages/rum-nextjs/src/entries/main"],
```
rgaignault, line 38: 💬 Suggestion : Rename it `datadog/browser-rum-next-plugin`

packages/rum-nextjs/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+import { datadogRum } from '@datadog/browser-rum'
+import { nextjsPlugin } from '@datadog/browser-rum-next-plugin'
+
+datadogRum.init({
+  applicationId: '<APP_ID>',
+  clientToken: '<CLIENT_TOKEN>',
+  site: 'datadoghq.com',
+  plugins: [nextjsPlugin({ router: 'app' })],
```
rgaignault: ❓ question : does nextJsPlugin take this argument ? 
doesn't seem so from the code

packages/rum-nextjs/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "scripts": {
+    "build": "node ../../scripts/build/build-package.ts --modules",
+    "prepack": "npm run build"
+  },
+  "dependencies": {
+    "@datadog/browser-core": "6.28.1",
+    "@datadog/browser-rum-core": "6.28.1"
+  },
+  "peerDependencies": {
+    "next": ">=12.0.0",
```
rgaignault: ❓ Question : why do we not use next 15 as stated in the readme ?

BeltranBulbarellaDD: It's a valid concern. I did it because we still want to support all versions but I am not married with this idea. Do you think it would be best to change it?

packages/rum-nextjs/README.md
BeltranBulbarellaDD: Not anymore! Good catch.

packages/rum-nextjs/src/domain/nextjsPlugin.ts
BenoitZugmeyer: Hum hum, right! In any case, we should be coherent between folder/package name/plugin name.

packages/rum-nextjs/package.json
BeltranBulbarellaDD: Yeah, also `"name": "@datadog/browser-rum-react",` WDYT? Stay consistent? Like: `browser-rum-nextjs`

BenoitZugmeyer: Are you suggesting `@datadog/browser-rum-next`? Why not, let's discuss in the guild?

packages/rum-nextjs/src/domain/nextjsPlugin.ts
```diff
@@ -0,0 +1,63 @@
...
+    onRumStart({ addEvent }) {
+      globalAddEvent = addEvent
+      for (const subscriber of onRumStartSubscribers) {
+        if (addEvent) {
+          subscriber(addEvent)
+        }
+      }
+    },
+  } satisfies RumPlugin
+}
```
rgaignault, line 36: suggestion 💬 : The check is redundant, you can have it before the loop

```suggestion
    onRumStart({ addEvent }) {
      globalAddEvent = addEvent
      if (addEvent) {                                                                                                
         for (const subscriber of onRumStartSubscribers) {
              subscriber(addEvent)
         }
      }
    },
  } satisfies RumPlugin
}
```

test/apps/nextjs-app-router/app/layout.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const metadata: Metadata = {
+  title: 'Next.js App Router Test',
+  description: 'Test app for Datadog RUM Next.js integration',
+}
+
+export default function RootLayout({ children }: { children: React.ReactNode }) {
+  return (
+    <html lang="en">
+      <DatadogRumAppRouterProvider>
```
rgaignault: Question ❓  : Are you sure this HTML is correct ? I think the body should be directly nested into the html

eslint.config.mjs
thomas-lebeau: it does not makes a lot of sense, no? 
- if the module exist then the rule should not be a problem, 
- if the module does not exists, then deactivating the rule is just hiding the problem.

Also, if yarn link pass locally, but not on the CI, maybe the problem is somewhere else and we should fix that instead of just ignoring it.

packages/rum-nextjs/src/domain/datadogAppRouter.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogAppRouter() {
+  const pathname = usePathname()
+  const params = useParams()
+  const previousPathname = useRef<string | null>(null)
+
+  if (previousPathname.current !== pathname) {
+    previousPathname.current = pathname
+    startNextjsView(computeViewNameFromParams(pathname, params))
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: This is not possible because we have to init in `instrumentation-client.js` so the `init` is never delayed. For `Pages Router` this will be handled differently

PR #4270, opened 2026-03-04
✨ [RUM-6798] Capture resource content type

bdibon:
## Motivation

Web Browsers are about to support the `PerformanceResourceTiming.contentType` property, this will enable the browser-sdk to add a content type to resources that have not been manually retrieved via the `fetch` or `XMLHttpRequest` APIs.

## Changes

- Update typing to reflect the new `contentType` property on resource entries
- Add types on the RUM Event to reflect the recently added `resource.response.headers` (see https://github.com/DataDog/rum-events-format/pull/351)
- Process the `PerformanceResourceTiming` and set `resource.response.headers["content-type"]` when relevant
- Add unit tests
- Add e2e tests

## Test instructions

Send events to intake (from FIREFOX!), your resources (but manual fetch/XHR calls) will have the new content-type field.

## Checklist

- [x] Tested locally
- [x] Tested on staging
- [ ] ~~Updated documentation and/or relevant AGENTS.md file~~

[RUM-6798]: https://datadoghq.atlassian.net/browse/RUM-6798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/e2e/scenario/rum/resources.scenario.ts
+++ b/test/e2e/scenario/rum/resources.scenario.ts
@@ -291,8 +291,13 @@ test.describe('rum resources', () => {
 
     createTest('collect resource response content-type for XHR resources')
       .withRum()
-      .run(async ({ intakeRegistry, flushEvents, page }) => {
-        test.skip(true, 'not implemented yet')
+      .run(async ({ intakeRegistry, flushEvents, page, browserName }) => {
+        // Firefox 129+ can use resource timing entries to collect xhr content-type,
+        // other browsers will have to retrieve the it from the response headers
+        test.skip(
+          isContentTypeAvailableInPerformanceEntry(test, browserName) === false,
+          'contentType is not available in this browser'
+        )
 
         await page.evaluate(
           () =>
@@ -315,8 +320,13 @@ test.describe('rum resources', () => {
 
     createTest('collect resource response content-type for fetch resources')
       .withRum()
-      .run(async ({ intakeRegistry, flushEvents, page }) => {
-        test.skip(true, 'not implemented yet')
+      .run(async ({ intakeRegistry, flushEvents, page, browserName }) => {
+        // Firefox 129+ can use resource timing entries to collect fetch content-type
+        // other browsers will have to retrieve the it from the response headers
+        test.skip(
+          isContentTypeAvailableInPerformanceEntry(test, browserName) === false,
+          'contentType is not available in this browser'
+        )
 
         await page.evaluate(() => fetch('/ok'))
 


packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -193,6 +195,30 @@ function computeGraphQlMetaData(
...
+  entry: RumPerformanceResourceTiming | undefined,
+  request: RequestCompleteEvent | undefined
+): ResourceHTTPResponse | undefined {
+  if (!entry && !request) {
+    return undefined
+  }
+
+  const rawContentTypeResponseHeader =
+    request?.xhr?.getResponseHeader('content-type') ?? request?.response?.headers?.get('content-type')
+  const contentTypeResponseHeader = rawContentTypeResponseHeader?.split(';')[0].trim()
```
bdibon: TODO: Confirm with mobile if they also strip charset and boundary

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -85,6 +86,15 @@ export interface RawRumResourceEvent {
...
 }
 
+interface ResourceHTTPResponseDefaultHeaders {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  'content-type'?: string
+}
+
+export interface ResourceHTTPResponse {
+  headers?: ResourceHTTPResponseDefaultHeaders & Record<string, string>
+}
```
bcaudan: **🥜 naming nitpick:** IMO we could drop the HTTP as we don't make any difference in the schema for now. If we find useful to have it, we should at least align with the case used elsewhere in the code base, cf `HttpResponse`.‏

packages/rum-core/src/rawRumEvent.types.ts
```diff
@@ -85,6 +86,15 @@ export interface RawRumResourceEvent {
   context?: Context
 }
 
+interface ResourceHTTPResponseDefaultHeaders {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  'content-type'?: string
+}
```
bcaudan: **🥜 nitpick:** Probably fine for now but if we plan to add more headers there it could be nice to find a more expressive naming than default. ‏

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -255,6 +256,42 @@ test.describe('rum resources', () => {
...
     })
 
+  test.describe('resource response content-type', () => {
+    createTest('collect resource response content-type in Firefox')
+      .withRum()
+      .withHead(html`<link rel="stylesheet" href="/empty.css" />`)
+      .run(async ({ intakeRegistry, flushEvents, browserName }) => {
+        const browserVersion = (test.info().project.metadata as BrowserConfiguration).version
+        test.skip(browserName !== 'firefox', 'contentType is only available in Firefox')
+        test.skip(browserName === 'firefox' && Number(browserVersion) < 129, 'contentType requires Firefox >= 129')
```
bcaudan: **💬 suggestion:** ‏looking at our e2e browser conf, we are currently testing on Firefox 119 🤔
What about not focusing on the browser support in the test names but instead have one test for static resources and one test for xhr/fetch?
It would allow us to enrich the xhr/fetch scenario later with network header capture.
We could extract a `isContentTypeAvailableInPerformanceEntry(test)` that we should update as the browser support evolve.

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -144,6 +144,7 @@ function assembleResource(
     : computeRequestDuration(pageStateHistory, startClocks, request!.duration)
 
   const graphql = request && computeGraphQlMetaData(request, configuration)
+  const response = computeResourceEntryHTTPResponse(entry)
```
bcaudan: **🥜 nitpick:** for consistency sake, we could extract the entry check from the method‏

```suggestion
const graphql = request && computeGraphQlMetaData(request, configuration)
  const response = entry && computeResourceEntryHTTPResponse(entry)
```

packages/rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -193,6 +195,26 @@ function computeGraphQlMetaData(
   return extractGraphQlMetadata(request, graphQlConfig)
 }
 
+function computeResourceEntryHTTPResponse(
+  entry: RumPerformanceResourceTiming | undefined
+): ResourceHTTPResponse | undefined {
```
bcaudan: **👏 praise:** Nice to have extracted that from the start!‏

packages/rum-core/src/domain/resource/resourceCollection.spec.ts
```diff
@@ -273,6 +275,41 @@ describe('resourceCollection', () => {
     })
   })
 
+  describe('HTTP response metadata enrichment', () => {
+    it('should extract content-type from performance entry when no request is available', () => {
```
bcaudan: **💬 suggestion:** or am I missing something with the request?‏

```suggestion
    it('should extract content-type from performance entry when available', () => {
```

packages/rum-core/src/domain/resource/resourceCollection.ts
bdibon: I would do it the other way around, encapsulate the `request` check in the `computeGraphQLMetadata`, wdyt?

packages/rum-core/src/rawRumEvent.types.ts
bdibon: I took it from the iOS RFC that references a default captured headers list. Would `ResourceHTTPResponseDefaultCapturedHeaders` be better or are you thinking to something more compact?

packages/rum-core/src/rawRumEvent.types.ts
bdibon: So, `ResourceResponse` and `DefaultResourceHeaders`?

packages/rum-core/src/domain/resource/resourceCollection.ts
bcaudan: we have this pattern at other places in the same method as well, no strong feeling on one way or the other, slight preference on being consistent though.

packages/rum-core/src/rawRumEvent.types.ts
bcaudan: `Captured` would add some intent, not sure about the meaning of `Default` though. We could iterate on the naming when we'll add the capture of headers. 
Maybe `ResourceResponseHeaders` is just fine for now.

packages/rum-core/src/rawRumEvent.types.ts
bcaudan: `ResourceResponse` looks good to me.
For the headers, cf [other comment](https://github.com/DataDog/browser-sdk/pull/4270#discussion_r2890909367)

test/e2e/lib/helpers/browser.ts
```diff
@@ -90,3 +99,12 @@ export async function sendXhr(page: Page, url: string, headers: string[][] = [])
...
   return result.response
 }
+
+export function isContentTypeAvailableInPerformanceEntry(
+  test: TestType<PlaywrightTestArgs & PlaywrightTestOptions, PlaywrightWorkerArgs & PlaywrightWorkerOptions>,
+  browserName: ReturnType<typeof getBrowserName>
+) {
+  const browserVersion = (test.info().project.metadata as BrowserConfiguration).version
+
+  return browserName === 'firefox' && Number(browserVersion) >= 129
```
thomas-lebeau, line 110: **❓ question:** Our ci we have two job:
- `e2e`: only run on chromium headless
- `e2e-bs`: run on firefox 119 (oldest version available that we support)

This mean that this test will never execute on our CI. Are we ok with that?‏

thomas-lebeau, line 110: **❓ question:**  Do we know of any timeline for support in other browsers? How do we keep track of this so we can update this method to reflect the new browser support?

test/e2e/scenario/rum/resources.scenario.ts
```diff
@@ -274,21 +275,44 @@ test.describe('rum resources', () => {
...
+    createTest('collect resource response content-type for XHR resources')
       .withRum()
-      .withHead(html`<link rel="stylesheet" href="/empty.css" />`)
-      .run(async ({ intakeRegistry, flushEvents, browserName }) => {
-        const browserVersion = (test.info().project.metadata as BrowserConfiguration).version
-        test.skip(
-          browserName === 'firefox' && Number(browserVersion) >= 129,
-          'contentType is available in Firefox >= 129'
+      .run(async ({ intakeRegistry, flushEvents, page }) => {
+        test.skip(true, 'not implemented yet')
```
bcaudan: **💬 suggestion:** ‏actually, we should probably have it on firefox as we get performance entries for xhr and fetch as well

test/e2e/lib/helpers/browser.ts
bdibon: We're adding a new feature that depends on a not widely (yet) supported browser API, so I think it's sound to have a test for it (that we can still run locally).

bdibon: No, we don't.

- Chrome has it available under the `enable-experimental-web-platform-features` flag
- Safari labelled it as `position:support`, so there's hope they will implement it someday

thomas-lebeau: what about adding a comment to make this more explicit, something like _currently our CI only test on firefox 119 so this test only run locally with `yarn test:e2e --project=firefox`._

test/e2e/scenario/rum/resources.scenario.ts
bdibon: This would reflect the current state of things better.

But when we will implement the response headers capture, it will be "overridden":
- we will use the resource timing API to collect content-type on static resources
- we will use the response header to collect content-type on fetch/xhr resources

IMO the current e2e tests are more clear, as they highlight what the new resource timing content-type brings to the table.

bcaudan: > But when we will implement the response headers capture, it will be "overridden":

we could adjust the e2e test when we will be there, depending on how we will implement that.
FMU, the response header capture feature could be behind a configuration option, so there will still be some cases where the value from performance response timing will be used for xhr/fetch.

thomas-lebeau (approved):

PR #4273, opened 2026-03-04
🐛 retry transient 503 errors in telemetry error checking

thomas-lebeau:
## Motivation

The deploy-to-prod CI job polls the Datadog Logs Analytics API every 60 seconds for 30 minutes after deployment. A transient `503 Service Unavailable` response (e.g. from rate limiting) crashes the script immediately since `fetchHandlingError` throws on any non-OK response with no retry logic.

## Changes

Add retry logic in `queryLogsApi` (`scripts/deploy/lib/checkTelemetryErrors.ts`):

- Wrap the `fetchHandlingError` call in a retry loop (up to 3 attempts)
- On 503, wait `RATE_LIMIT_DELAY_MS` (6s) before retrying
- Use existing `findError`/`FetchError` utilities to detect the error type
- Log each retry with `printLog` for CI visibility
- Re-throw immediately on non-503 errors or after exhausting retries

## Test instructions

- `yarn typecheck` passes
- `yarn test:script` passes

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/deploy/lib/checkTelemetryErrors.ts
+++ b/scripts/deploy/lib/checkTelemetryErrors.ts
@@ -1,8 +1,8 @@
 /**
  * Check telemetry errors
  */
-import { Agent } from 'undici'
-import { printLog, fetchHandlingError, timeout, findError, FetchError } from '../../lib/executionUtils.ts'
+import { Agent, fetch, type Response } from 'undici'
+import { printLog, timeout, createFetchError } from '../../lib/executionUtils.ts'
 import { getTelemetryOrgApiKey, getTelemetryOrgApplicationKey } from '../../lib/secrets.ts'
 import { getDatacenterMetadata } from '../../lib/datacenter.ts'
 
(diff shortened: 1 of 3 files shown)


scripts/deploy/lib/checkTelemetryErrors.ts
```diff
@@ -110,54 +111,69 @@ async function queryLogsApi(
...
+      ) {
+        throw new Error(`Unexpected response from the API: ${JSON.stringify(data)}`)
+      }
+
+      return data.data.buckets
+    } catch (error) {
+      const fetchError = findError(error, FetchError)
+      if (!fetchError || fetchError.response.status !== 503 || attempt === MAX_RETRIES) {
+        throw error
+      }
```
bcaudan: **💬 suggestion:** ‏it seems that we are using the first throw to control the execution flow.
it could be easier to follow / maintain with a structure like:


```ts
for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
  const response = await queryLogs()
  if (isValid(response)) {
    return response.data.buckets
  } else if (shouldRetry(response)) {
    printLog(...)
    await timeout(...)
  } else {
    throw new Error(...)
  }
}
```

bcaudan (approved):
LGTM

PR #4275, opened 2026-03-04
👷 remove usages of "lerna" in `yarn release`

BenoitZugmeyer:
## Motivation

Preparatory work toward removing lerna from the project. This PR replaces `lerna version` with a
custom TypeScript script, and moves the version source of truth from `lerna.json` to root
`package.json`. `lerna.json` is kept in sync so the existing publish pipeline is unaffected.

## Changes

The new `yarn release <version>` command creates a release branch, bumps all package versions and
`@datadog/*` dependencies across the monorepo, opens the changelog editor, then commits and tags.
It also prints next steps (push, open PR, trigger deployment).

The old `lerna version` hook machinery (`cmd_release`, `cmd_version`, `update-peer-dependency-versions.ts`) is removed, as is the split between `lerna.json` and `package.json` as version sources.

## Test instructions

- Run `yarn release 1.2.3` on a test branch — should create a `release/v1.2.3` branch, bump versions in all `package.json` files and `lerna.json`, open the changelog editor, then commit and tag `v1.2.3`

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/lib/filesUtils.ts
+++ b/scripts/lib/filesUtils.ts
@@ -58,7 +58,7 @@ export interface PackageJsonInfo {
 
 interface PackageJson {
   name?: string
-  private?: string
+  private?: boolean
   version?: string
   dependencies?: Record<string, string>
   devDependencies?: Record<string, string>


scripts/release/prepare-release.ts
```diff
@@ -0,0 +1,80 @@
...
+Next steps:
+  1. git push origin ${releaseBranch}
+  2. Open a PR from ${releaseBranch} to main
+  3. Ask for a review and merge the PR
+  4. Go to the CI and trigger the deployment pipeline
+`)
+})
+
+function setVersionInPackageJsonFiles(version: string) {
+  const packageJsonFiles = findPackageJsonFiles()
```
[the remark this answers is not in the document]
BenoitZugmeyer: This is wrong. codex is confused because it thinks `*/package.json` will be shell-expanded, but the `command` nodejs utility does not uses a shell (by design!) and passes the argument to git. This is the equivalent of running:

```
git ls-files -- 'package.json' '*/package.json'
```

scripts/release/prepare-release.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (command`git status --porcelain`.run().trim()) {
+    throw new Error('Git working tree is not clean')
+  }
+
+  const releaseBranch = `release/v${version}`
+  const currentBranch = command`git branch --show-current`.run().trim()
+  if (currentBranch !== releaseBranch) {
+    printLog(`Checking out ${releaseBranch} branch...`)
+    command`git checkout -b ${releaseBranch} main`.run()
+  }
```
bcaudan, line 32: **❓ question:** ‏is it lerna that required the branch to be pushed to origin before the release?

BenoitZugmeyer: Yes, it was lerna. I always wondered why this was a requirement, I don't see a good reason why we would need this. This was really annoying.

scripts/release/prepare-release.ts
```diff
@@ -0,0 +1,80 @@
...
+  3. Ask for a review and merge the PR
+  4. Go to the CI and trigger the deployment pipeline
+`)
+})
+
+function setVersionInPackageJsonFiles(version: string) {
+  const packageJsonFiles = findPackageJsonFiles()
+
+  for (const { path: absolutePath, content } of packageJsonFiles) {
+    if (isSemanticVersion(content.version)) {
```
thomas-lebeau, line 63: **❓ question:** ‏Do we care to check if the _previous_ version is semantic?

scripts/release/prepare-release.ts
```diff
@@ -0,0 +1,80 @@
...
+  updateLernaJson(version)
+  command`node ./scripts/release/generate-changelog/index.ts`.withLogs().run()
+  command`git add -u`.run()
+  command`git commit -m v${version}`.run()
+  command`git tag -a v${version} -m v${version}`.run()
+
+  printLog(`Release v${version} prepared successfully!`)
+  printLog(`
+Next steps:
+  1. git push origin ${releaseBranch}
```
thomas-lebeau: **❓ question:** ‏why not pushing the branch automatically?

scripts/release/prepare-release.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  if (command`git status --porcelain`.run().trim()) {
+    throw new Error('Git working tree is not clean')
+  }
+
+  const releaseBranch = `release/v${version}`
+  const currentBranch = command`git branch --show-current`.run().trim()
+  if (currentBranch !== releaseBranch) {
+    printLog(`Checking out ${releaseBranch} branch...`)
+    command`git checkout -b ${releaseBranch} main`.run()
```
thomas-lebeau: **🔨 warning:** ‏

```suggestion
    command`git fetch origin`.run()
    command`git checkout -b ${releaseBranch} origin/main`.run()
```

scripts/release/prepare-release.ts
BenoitZugmeyer: Yes, we don't want to set the version when [`*`](https://github.com/DataDog/browser-sdk/blob/d1022bbe23fa0ac7c94bb511bf5ba629a28e4fd3/test/apps/vanilla/package.json#L10-L11) or [`workspace:*`](https://github.com/DataDog/browser-sdk/blob/d1022bbe23fa0ac7c94bb511bf5ba629a28e4fd3/developer-extension/package.json#L17-L18) is used.

scripts/release/prepare-release.ts
BenoitZugmeyer: I don't mind, but maybe I'll add a `--no-push` option to be able to try the script without pushing?

thomas-lebeau: or maybe a `--push` option that we can add to the the command in package.json

scripts/release/prepare-release.ts
```diff
@@ -0,0 +1,80 @@
...
+
+function updateLernaJson(version: string) {
+  const lernaJsonPath = `${import.meta.dirname}/../../lerna.json`
+  const content = JSON.parse(fs.readFileSync(lernaJsonPath, 'utf8'))
+  content.version = version
+  fs.writeFileSync(lernaJsonPath, `${JSON.stringify(content, null, 2)}\n`)
+}
+
+function isSemanticVersion(input: string | undefined): boolean {
+  return input !== undefined && /^\d+\.\d+\.\d+/.test(input)
```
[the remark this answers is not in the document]
BenoitZugmeyer: Fixed, anchored the regex with `$`.

scripts/release/prepare-release.ts
BenoitZugmeyer: Applied, thanks!

scripts/release/prepare-release.ts
BenoitZugmeyer: Went with `--push`, wired it up in the `release` script in `package.json` so `yarn release <version>` always pushes, but you can run the script directly without `--push` to test locally.

thomas-lebeau (approved):

scripts/lib/filesUtils.ts
```diff
@@ -50,10 +50,19 @@ export async function modifyFile(
...
+export interface PackageJsonInfo {
   relativePath: string
   path: string
-  content: any
+  content: PackageJson
+}
+
+interface PackageJson {
+  name?: string
+  private?: string
```
rgaignault: 💬 Suggestion : this should be a boolean

scripts/cli
```diff
@@ -55,27 +55,6 @@ cmd_build_json2type () {
...
-
-cmd_version () {
-  node ./scripts/release/generate-changelog/index.ts
-  node ./scripts/release/update-peer-dependency-versions.ts
-  yarn run pack
-  # keep test apps lockfiles up to date
-  for app_package_path in $(git ls-files test/apps/*/package.json); do
-    app_dir=$(dirname "$app_package_path");
-    cd "$app_dir"
-    yarn up
```
rgaignault, line 74: ❓ Question : why did we drop the test app lockfile update

scripts/lib/filesUtils.ts
BenoitZugmeyer: Good catch! Will be fixing this in the next push.

scripts/cli
BenoitZugmeyer: The lockfile update was necessary when test apps referenced fixed `@datadog/` versions in their `package.json` (so bumping versions would modify those files and invalidate the lockfile). Now that test apps use `*` or `workspace:*`, `setVersionInPackageJsonFiles()` never modifies them, so there is nothing to sync in the lockfile.

rgaignault: I see, probably nitpick then but that mean the lockfile version would be "freeze" while it resolve with the other package json version ?

rgaignault: like resolve with 6.31 while lockfile is still 6.30, but it's probably just cosmetic anyway

BenoitZugmeyer: Huum right, some test apps are still specifying a fixed version. Let me get rid of this :)

BenoitZugmeyer: Actually no, there is no more fixed version in lockfiles.

PR #4276, opened 2026-03-04
👷 replace lerna publish with yarn npm publish

BenoitZugmeyer:
## Motivation

This is a follow-up of https://github.com/DataDog/browser-sdk/pull/4275

Second step toward removing lerna from the project. This PR replaces `lerna publish` with `yarn npm publish` to publish NPM packages.

## Changes

* The publish script now uses `yarn workspaces foreach npm publish` and gains a `--dry-run` option. The old lerna-based script is preserved as `publish-npm-legacy.ts` just in case things don't work we can still publish using lerna. 

* Replaced `.npmignore` with `files` package.json field, because `yarn pack` does not support how we use `.npmignore` (we exclude, include, then re-exclude files)

* The `check-packages.ts` script is updated:
  * use `yarn pack` for dry-run validation
  * now also verifies that packed files match the `files` field. I'm not totally sure this is useful, but 🤷 
  * assert that LICENSE, README.md and package.json are included
  
* Add a `LICENSE` file to all packages, because `yarn npm publish` does not pull it from the repository root like lerna did.

## Test instructions

- Run `node scripts/deploy/publish-npm.ts --dry-run` — should list files that would be published without actually publishing
- Run `node scripts/check-packages.ts` — should pass without errors

I used a small script to compare files listed by `node scripts/deploy/publish-npm.ts --dry-run` and the current packages published on NPM to ensure everything looks good.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/lib/command.spec.ts
+++ b/scripts/lib/command.spec.ts
@@ -1,0 +1,49 @@
+import { test } from 'node:test'
+import assert from 'node:assert'
+import { parseCommandTemplateArguments } from './command.ts'
+
+test('splits on white spaces', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo bar`, ['foo', 'bar'])
+})
+
+test('interpolates a variable as a separate argument when surrounded by spaces', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo ${'bar'} baz`, ['foo', 'bar', 'baz'])
+})
+
+test('appends a variable to the previous argument when no leading space', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo${'bar'} baz`, ['foobar', 'baz'])
+})
+
+test('prepends a variable to the next argument when no trailing space', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo ${'bar'}baz`, ['foo', 'barbaz'])
+})
+
+test('merges variable into surrounding argument when no spaces on either side', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo${'bar'}baz`, ['foobarbaz'])
+})
+
+test('does not split a variable value on white spaces', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo ${'bar baz'}`, ['foo', 'bar baz'])
+})
+
+test('spreads array variables as separate arguments', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo ${['bar', 'baz']}`, ['foo', 'bar', 'baz'])
+})
+
+test('keeps empty string as a distinct argument', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo ${''} bar`, ['foo', '', 'bar'])
+})
+
+test('ignores empty array, producing no extra arguments', () => {
+  assert.deepStrictEqual(parseCommandTemplateArguments`foo ${[]} bar`, ['foo', 'bar'])
+})
+
+test('real-world: git commit with message', () => {
+  const commitMessage = 'my commit message'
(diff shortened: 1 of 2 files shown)


scripts/deploy/publish-npm.ts
```diff
@@ -1,14 +1,23 @@
...
+      'dry-run': { type: 'boolean', default: false },
+    },
+  })
 
-  printLog('Publishing')
-  // eslint-disable-next-line no-template-curly-in-string
-  fs.writeFileSync('.npmrc', '//registry.npmjs.org/:_authToken=${NPM_TOKEN}')
-  command`yarn lerna publish from-package --yes`.withEnvironment({ NPM_TOKEN: getNpmToken() }).withLogs().run()
+  printLog(dryRun ? 'Publishing (dry run)' : 'Publishing')
+  command`yarn workspaces foreach --verbose --all --topological --no-private npm publish --access public ${dryRun ? ['--dry-run'] : []}`
```
thomas-lebeau: **🥜 nitpick:** ‏
```suggestion
  command`yarn workspaces foreach --verbose --all --topological --no-private npm publish --access public ${dryRun ? '--dry-run' : ''}`
```

scripts/check-packages.ts
```diff
@@ -27,28 +19,33 @@ runMain(() => {
 function checkBrowserSdkPackage(packagePath: string) {
   const packageJson = getPackageJson(packagePath)
 
-  if (packageJson?.private) {
-    printWarning(`Skipping private package ${packageJson.name}`)
-    return true
-  }
```
thomas-lebeau, line 33: **❓ question:** ‏We don't need this anymore?

BenoitZugmeyer: This was for the 'flagging' package that we removed recently.

I removed this check because it ignored the new integration packages that we are about to release. Better catch malformed packages early than late.

scripts/deploy/publish-npm.ts
BenoitZugmeyer: It's not quite the same.

```ts
command`ls ${''}` // passes an empty argument, similar to `ls ''`
command`ls ${[]}` // passes no argument
```

thomas-lebeau: > `command`ls ${''}` -> passes an empty argument, similar to `ls ''``

really? shouldn't it be 
```js
command`ls ${"''"}`
```
to output `ls ''` ?

BenoitZugmeyer: Maybe I wasn't quite clear.

```ts
command`ls ${''}`    // passes an empty argument, similar to running `ls ''` in a shell
command`ls ${'foo'}` // passes `foo` as argument, similar to running `ls 'foo'` in a shell
command`ls ''`       // passes `''` as an argument, similar to running `ls "''"` in a shell
command`ls ${"''"}`  // same as above
command`ls ${[]}`    // passes no argument
```

This is a feature, not a bug 😄  I had security concerns about shell injections when spawning commands in our scripts. So `command` does not spawn shells and instead passes values directly to the executed binary. It might seem confusing at first but it removes a whole source of problems.

I added a few tests to illustrate the behavior: https://github.com/DataDog/browser-sdk/pull/4276/commits/f90939f2a91e819ec4a02b9bdd9c4e1e83db03ce#diff-7933dd4283324d4158a016dd9934b38c391f7c61b99fdebf8d00be9134900746R34-R39

domalessi (approved):

scripts/deploy/publish-npm.ts
thomas-lebeau: thank you! I did not realized 
```
command`ls ${''}`
// is different from
command`ls `
```

unlike 
```
`ls ${''}` === `ls ` // true
```

Thanks for adding the tests

PR #4278, opened 2026-03-04
🔊 add debug context to "Too many requests" telemetry

BenoitZugmeyer:
## Motivation

The `requestRegistry` can overflow when `trackEarlyRequests` is enabled,
triggering a "Too many requests" telemetry event. One known scenario is
`fetch()` called with an already-aborted `AbortSignal` — the browser fires
no `PerformanceResourceTiming` entry, so the request is never matched and
accumulates permanently. But other causes may exist in the wild. This PR
adds more context to the telemetry to help understand what is actually
happening.

## Changes

Adds an `isAbortedOnStart` flag to fetch requests (true when the `AbortSignal`
is already aborted at call time) and propagates it through to
`RequestCompleteEvent`.

Enriches the "Too many requests" telemetry with a debug context (gated behind
the `too_many_requests_investigation` experimental feature): counts of aborted,
aborted-on-start, XHR and fetch requests; how many lack a matching
`PerformanceResourceTiming` entry; and the age of the oldest request both from
start time and from end time. The telemetry is also capped to fire only once
per registry instance.

## Test instructions

Enable `trackEarlyRequests: true` and `enableExperimentalFeatures:
['too_many_requests_investigation']`, then fire many `fetch()` calls with
an already-aborted `AbortController` signal. The "Too many requests" debug
telemetry event should appear in the intake with the context fields populated.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/requestCollection.ts
+++ b/packages/rum-core/src/domain/requestCollection.ts
@@ -98,9 +98,9 @@ export function trackXhr(lifeCycle: LifeCycle, configuration: RumConfiguration, 
       case 'complete':
         tracer.clearTracingIfNeeded(context)
         lifeCycle.notify(LifeCycleEventType.REQUEST_COMPLETED, {
+          ...context,
           type: RequestType.XHR,
           isAbortedOnStart: false,
-          ...context,
         })
         break
     }
@@ -136,10 +136,10 @@ export function trackFetch(lifeCycle: LifeCycle, configuration: RumConfiguration
       case 'resolve':
         tracer.clearTracingIfNeeded(context)
         lifeCycle.notify(LifeCycleEventType.REQUEST_COMPLETED, {
+          ...context,
           duration: elapsed(context.startClocks.timeStamp, timeStampNow()),
           type: RequestType.FETCH,
           requestBody: context.init?.body,
-          ...context,
         })
         break
     }


packages/core/src/browser/fetchObservable.ts
```diff
@@ -105,6 +106,7 @@ function beforeSend(
     startClocks,
     url,
     handlingStack,
+    isAbortedOnStart: init?.signal?.aborted || false,
```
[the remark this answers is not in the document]
BenoitZugmeyer: Fixed — now checks both `input.signal?.aborted` (when `input` is a `Request`) and `init?.signal?.aborted`, combined with `||` so both paths are covered.

packages/rum-core/src/domain/resource/requestRegistry.ts
```diff
@@ -15,13 +22,57 @@ export const MAX_REQUESTS = 1000
...
+              abortedCount++
+            }
+            if (r.isAbortedOnStart) {
+              abortedOnStartCount++
+            }
+            if (r.type === RequestType.XHR) {
+              xhrCount++
+            }
+            const entries = performance.getEntriesByName(r.url, 'resource') as PerformanceResourceTiming[]
+            const hasMatchingEntry = entries.some((e) => e.startTime >= r.startClocks.relative)
```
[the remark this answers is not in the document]
BenoitZugmeyer: We are fine with an approximation here — the goal is to get a rough sense of how many requests lack a timing entry, not an exact count.

packages/rum-core/src/domain/resource/requestRegistry.ts
```diff
@@ -15,13 +22,57 @@ export const MAX_REQUESTS = 1000
...
+            if (r.isAborted) {
+              abortedCount++
+            }
+            if (r.isAbortedOnStart) {
+              abortedOnStartCount++
+            }
+            if (r.type === RequestType.XHR) {
+              xhrCount++
+            }
+            const entries = performance.getEntriesByName(r.url, 'resource') as PerformanceResourceTiming[]
```
mormubis, line 53: **💬 suggestion:** ‏We could use the `globalObject` to access `performance`. Decoupling the file from the browser.

packages/rum-core/src/domain/requestCollection.ts
```diff
@@ -97,21 +98,9 @@ export function trackXhr(lifeCycle: LifeCycle, configuration: RumConfiguration,
...
-          traceSampled: context.traceSampled,
           type: RequestType.XHR,
-          url: context.url,
-          xhr: context.xhr,
-          isAborted: context.isAborted,
-          handlingStack: context.handlingStack,
-          requestBody: context.requestBody,
-          responseBody: context.responseBody,
+          isAbortedOnStart: false,
+          ...context,
```
mormubis: **❓ question:** ‏Do we plan to context override `type` or `isAbortedOnStart`? If not, I think it'd be safer to spread context at the beginning.

mormubis (approved):

packages/rum-core/src/domain/resource/requestRegistry.ts
BenoitZugmeyer: True, but also I don't intend to ever enable this flag out of web-ui :D

packages/rum-core/src/domain/requestCollection.ts
BenoitZugmeyer: Good catch — moved `...context` before the explicit fields so `type` and `isAbortedOnStart` always override anything in context. Will be pushing this shortly.

PR #4287, opened 2026-03-05
✨ [PANA-6283] Support incremental mutation Change records

sethfowler-datadog:
## Motivation

Change records should support incremental mutations. After quite a few stacked PRs, it's time to make it happen.

## Changes

This PR adds a new serialization implementation for `MutationObserver` records, `serializeMutationsAsChange.ts`. In broads stroke, the implementation is similar to the existing `serializeMutations.ts`, but the output is a Change record instead of a record in the older V1 format.

I've tried to make the new code a little bit simpler than the old code. Many of the simplifications are possible because of a nice property of Change records: we never reuse old node ids. In particular, this means:
* A new or moved node is always assigned a new node id that's greater than any existing node id. The algorithm does not need to build a data structure to track which nodes it has already serialized; it can simply compare their node id with the highest node id that existed before the current serialization operation started.
* A removed node's id can simply be "forgotten", since we will never need it again. This means that the algorithm does not need to build a data structure to track which nodes have been removed; it can simply check whether a node currently has a node id assigned.

Because at this stage my focus is on ensuring that the Change serialization code matches the V1 serialization algorithm's output, the testing approach this PR uses is "instrument" the existing incremental mutation tests with a helper class, `SerializationVerifier`, that observes each serialization performed by the existing tests. When `SerializationVerifier` sees that a V1 serialization has occurred, it:
1. Performs the same serialization using the Change serialization code.
2. Converts the Change output to the older V1 format.
3.  Verifies that the two match.

For full snapshots, we check for byte-for-byte identical results. For incremental snapshots, we tolerate some small differences, which are documented in the code. These checks are quite strict, and combined with the extensive manual validation I've performed, I'm confident that Change serialization algorithm for both full and incremental snapshots is performing well.

## Test instructions

1. Check out this branch locally.
6. Run `yarn dev`.
7. Visit staging with the browser SDK extension running.
8. Edit the config using the extension to ensure that the list of experimental feature flags includes `use_incremental_change_records`.
9. Surf around a bit and use `DD_RUM.getSessionReplayLink()` in the console to get a link to your session.
10. Verify that it plays back correctly. (I've already merged playback support, so you don't need to do anything special to view it.)

Unfortunately you can't just rely on the live session replay feature within the browser SDK extension, because I haven't yet updated the extension with the latest replay code. I'll update this comment when that changes.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
+++ b/packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
@@ -45,8 +45,13 @@ function processMutations(mutations: RumMutationRecord[], transaction: ChangeSer
   const characterDataMutations = new Map<Node, OldValue>()
   const removedNodes = new Set<Node>()
 
-  // Collect the changes made by this sequence of mutations. The mutations tell us which
-  // nodes have been changed, but they don't
+  // Collect the changes made by this sequence of mutations. It's important to think of
+  // the mutations as telling us which parts of the tree are dirty, but not necessarily as
+  // specifying exactly which changes have occurred; trying to be too clever will result
+  // in incorrect behavior. For example, if we see a sequence of mutations where a node is
+  // removed and then added back in the same position, one might be tempted to "cancel
+  // out" these two changes and ignore both mutations, but that could cause us to miss
+  // changes to the node that occurred while it was detached.
   for (const mutation of mutations) {
     switch (mutation.type) {
       case 'attributes': {
@@ -146,8 +151,8 @@ function processAddedNodes(
 
     const parentId = nodeIds.get(parentNode)
     if (parentId === undefined) {
-      // This node's parent hasn't been serialized (e.g. because it's hidden),
-      // so we shouldn't serialize this node either.
+      // This node's parent hasn't been serialized (e.g. because it has privacy level
+      // HIDDEN), so we shouldn't serialize this node either.
       continue
     }
 
@@ -219,7 +224,7 @@ function processAttributeMutations(
   firstNewNodeId: NodeId,
   nodePrivacyLevelCache: NodePrivacyLevelCache,
   transaction: ChangeSerializationTransaction
-) {
+): void {
   const nodeIds = transaction.scope.nodeIds
 
   for (const [node, attributeNames] of mutations) {


packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
```diff
@@ -0,0 +1,293 @@
...
+type OldValue = string | null
+
+function processMutations(mutations: RumMutationRecord[], transaction: ChangeSerializationTransaction): void {
+  const addedNodes = new Set<Node>()
+  const attributeMutations = new Map<Element, Map<AttributeName, OldValue>>()
+  const characterDataMutations = new Map<Node, OldValue>()
+  const removedNodes = new Set<Node>()
+
+  // Collect the changes made by this sequence of mutations. The mutations tell us which
+  // nodes have been changed, but they don't
```
yannickadam: We're missing the end of the comment

packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
```diff
@@ -0,0 +1,293 @@
...
+    transaction.setText(nodeId, content)
+  }
+}
+
+function processAttributeMutations(
+  mutations: Map<Element, Map<AttributeName, OldValue>>,
+  firstNewNodeId: NodeId,
+  nodePrivacyLevelCache: NodePrivacyLevelCache,
+  transaction: ChangeSerializationTransaction
+) {
```
yannickadam: nit
```suggestion
): void {
```

packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    const parentNode = getParentNode(node)
+    if (!parentNode) {
+      continue // This node has no parent.
+    }
+
+    const parentId = nodeIds.get(parentNode)
+    if (parentId === undefined) {
+      // This node's parent hasn't been serialized (e.g. because it's hidden),
```
yannickadam: When we say hidden here, I believe we're talking about privacy, right?

packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
```diff
@@ -0,0 +1,293 @@
...
+  for (const mutation of mutations) {
+    switch (mutation.type) {
+      case 'attributes': {
+        const node = mutation.target
+        let attributes = attributeMutations.get(node)
+        if (!attributes) {
+          attributes = new Map<AttributeName, OldValue>()
+          attributeMutations.set(node, attributes)
+        }
+        const attributeName = mutation.attributeName!
```
yannickadam, line 64: Could we use a discriminated union for RumMutationRecord so we don't need `!` here?

packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
sethfowler-datadog: Yeah, `NodePrivacyLevel.HIDDEN`. I'll update the comment to clarify.

packages/rum/src/domain/record/serialization/serializeMutationsAsChange.ts
sethfowler-datadog: We already are; `RumMutationRecord` is a union of `RumCharacterDataMutationRecord`, `RumAttributesMutationRecord`, and `RumChildListMutationRecord`, discriminated by a `type` field. The problem is that `RumAttributesMutationRecord` is wrongly defined this way:
```ts
export interface RumAttributesMutationRecord {
  type: 'attributes'
  target: Element
  oldValue: string | null
  attributeName: string | null
}
```

It should be defined like this:
```ts
export interface RumAttributesMutationRecord {
  type: 'attributes'
  target: Element
  oldValue: string | null
  attributeName: string
}
```

The file that defines these types has different CODEOWNERS, so I can't easily fix it in this PR, but I'll open a followup to fix it.

yannickadam: 👍

PR #4290, opened 2026-03-07
✨ NextJS Pages Router Integration

BeltranBulbarellaDD:
## Motivation

Adds NextJS Pages Router support to `@datadog/browser-rum-nextjs`, complementing the existing App Router integration.
Small [one pager](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6345097478/Small+RFC+-+Browser-SDK+NextJS+Pages+Router+Integration) on the implementation

## Changes

**Package (`packages/rum-nextjs/`)**

  - `DatadogPagesRouter` — Client component that starts RUM views on route changes, using `useRouter().pathname` from `next/router` which already provides the normalized route pattern (e.g. /user/[id], /guides/[...slug]). 
  - Updated README with Pages Router usage instructions alongside existing App Router docs.

  **E2E infrastructure**

  - Adds a Next.js Pages Router test app (`test/apps/nextjs-pages-router/`) with dynamic routes (`/user/[id]`,` /guides/[...slug]`)
  - Configures the E2E framework to build and serve the Pages Router app alongside existing test apps
  - Adds `withNextjsPagesApp()` test builder method for Pages Router-specific E2E scenarios

  **E2E tests**

Same as for App Router:
  - Initial view tracking with correct loading_type
  - SPA navigation with loading_time measurement
  - Route normalization across dynamic and catch-all routes, with URL and referrer assertions

## Test instructions

- Existing unit tests in rum-nextjs should pass: `yarn test:unit --spec packages/rum-nextjs/src/domain/nextjsPlugin.spec.ts`
  - Typecheck should pass: `yarn typecheck`
  - New E2E tests should pass: `yarn test:e2e -g "pages router"`
  - 
## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-nextjs/src/domain/datadogPagesRouter.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogPagesRouter() {
+  const router = useRouter()
+  const previousPathname = useRef<string | null>(null)
+
+  if (previousPathname.current !== router.pathname) {
+    previousPathname.current = router.pathname
+    // router.pathname already contains the route pattern (e.g., "/user/[id]")
+    startNextjsView(router.pathname)
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Tested, this is a valid concern

packages/rum-nextjs/src/domain/nextJSRouter/datadogPagesRouter.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useRouter } from 'next/router'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogPagesRouter() {
+  const router = useRouter()
+  const previousAsPath = useRef<string | null>(null)
+
+  if (previousAsPath.current !== router.asPath) {
+    previousAsPath.current = router.asPath
+    // router.pathname is the route pattern (e.g., "/user/[id]") — used as the view name
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Bot has a point
https://nextjs.org/docs/pages/building-your-application/rendering/automatic-static-optimization#caveats

packages/rum-nextjs/src/domain/datadogPagesRouter.tsx
```diff
@@ -0,0 +1,20 @@
+'use client'
```
mormubis: `'use client'` is an App Router / RSC thing — Pages Router doesn't have server components so this does nothing. Might confuse someone working in a hybrid app who wonders what server rendering rules apply here.

packages/rum-nextjs/src/domain/datadogPagesRouter.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { useRef } from 'react'
+import { useRouter } from 'next/router'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogPagesRouter() {
+  const router = useRouter()
+  const previousAsPath = useRef<string | null>(null)
+
+  if (previousAsPath.current !== router.asPath) {
```
mormubis: Pages Router [docs warn](https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object) not to use `router.asPath` until `router.isReady` is true — on SSG pages it can be wrong before hydration. When `isReady` flips, `asPath` updates and this fires a second `startNextjsView` for what's still the initial load.

```tsx
if (!router.isReady) return null
```

test/e2e/scenario/nextjsPlugin.scenario.ts
```diff
@@ -1,61 +1,115 @@
...
-      await page.waitForURL('**/user/42?admin=true')
-
-      await flushEvents()
-
-      const viewEvents = intakeRegistry.rumViewEvents
-      const homeView = viewEvents.find((e) => e.view.name === '/' && e.view.loading_type === 'initial_load')
-      expect(homeView).toBeDefined()
-    })
-  createTest('should normalize dynamic routes and preserve real URLs and referrers')
+test.describe('nextjs pages router', () => {
```
mormubis: Two `test.describe('nextjs pages router', ...)` blocks in this file — one from the `forEach`, one here. They show up as separate groups in the test output. Could fold this into the loop with `if (router === 'pages')`, or keep everything separate and drop the loop entirely.

packages/rum-nextjs/src/domain/nextJSRouter/datadogPagesRouter.tsx
```diff
@@ -0,0 +1,25 @@
+import { useRef } from 'react'
+import { useRouter } from 'next/router'
+import { startNextjsView } from './nextjsPlugin'
+
+export function DatadogPagesRouter() {
```
rgaignault, line 6: 💬 Suggestion : Don't we need any uni test for this ?

test/e2e/scenario/nextjsPlugin.scenario.ts
```diff
@@ -1,108 +1,157 @@
...
+        const userView = intakeRegistry.rumViewEvents.find((e) => e.view.name === '/user/[id]')
+        expect(userView).toBeDefined()
+
+        // No view should have been created for the query-param-only navigation
+        const spuriousView = intakeRegistry.rumViewEvents.find((e) => e.view.url?.includes('admin=false'))
+        expect(spuriousView).toBeUndefined()
+      })
+  })
+
+  if (router === 'pages') {
```
rgaignault, line 110: ❓ Question : Why those test are not nested under a "descrbibe" ?

packages/rum-nextjs/src/domain/nextJSRouter/datadogPagesRouter.tsx
BeltranBulbarellaDD: I think most importantly are the e2e that test this behavior. I could definitely add unit tests. WDYT?

test/e2e/scenario/nextjsPlugin.scenario.ts
BeltranBulbarellaDD: Sort of because of [this](https://github.com/DataDog/browser-sdk/pull/4290#discussion_r2906531620). I had them in their own describe but that repeated titles so I merged them.
But IIUC they are nested under [this](https://github.com/DataDog/browser-sdk/pull/4290/changes#diff-9b2401066a98d985a3d1d288ae7a4bf78380e3d873b1c78d4eb9f3fa0c58f81aR7) describe

test/e2e/playwright.base.config.ts
```diff
@@ -51,5 +51,13 @@ export const config: Config = {
         stdout: /- Local:\s+http:\/\/localhost:(?<nextjs_app_router_port>\d+)/,
       },
     },
+    {
+      stdout: 'pipe' as const,
+      cwd: path.join(__dirname, '../apps/nextjs-pages-router'),
```
amortemousque: **❓ question:** ‏Do you really need a new application? Couldn’t you serve both Page Router and App Router pages from a single app?

packages/rum-nextjs/src/domain/nextJSRouter/datadogPagesRouter.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const router = useRouter()
+  const previousPath = useRef<string | null>(null)
+
+  if (!router.isReady) {
+    return null
+  }
+
+  // Extract the path portion of asPath (without query params or hash) to detect navigations.
+  const path = router.asPath.split(/[?#]/)[0]
```
amortemousque, line 15: **❓ question:** ‏Is asPath deprecated https://nextjs.org/docs/app/guides/migrating/app-router-migration#:~:text=asPath%20has%20been%20removed%20because%20the%20concept%20of%20as%20has%20been%20removed%20from%20the%20new%20router.?

BeltranBulbarellaDD: Nope, that's the App Router docs right? This is the correct [link](https://nextjs.org/docs/pages/api-reference/functions/use-router#:~:text=asPath%3A%20String%20%2D%20The%20path%20as%20shown%20in%20the%20browser%20including%20the%20search%20params%20and%20respecting%20the%20trailingSlash%20configuration.%20basePath%20and%20locale%20are%20not%20included.)

test/e2e/playwright.base.config.ts
BeltranBulbarellaDD: I think it could be possible. [Link](https://stackoverflow.com/a/76774999). It's not recommended and there might be a side effect. As first though we would have to have different routes for both (not a problem), my only concern is mostly `instrumentation-client` since Pages Router does not export `onRouterTransitionStart`. I'll implement it and see if any issues arise.

packages/rum-nextjs/src/domain/nextJSRouter/datadogPagesRouter.tsx
BeltranBulbarellaDD: And while i'm adding changing the e2e setup I could add these tests too. For both `App` and `Pages`

rgaignault: Yes defintely, always a good idea to have unit test also

PR #4296, opened 2026-03-09
✨ [RUM-14736] Add size to ResourceStopOptions

BeltranBulbarellaDD:
## Motivation

`stopResource `accepts a size parameter in cross-platform SDKs (e.g. Flutter), but the Browser SDK's `ResourceStopOptions` had no corresponding field.

As a result, manually tracked web resources silently dropped the caller-provided byte count.

## Changes

  - Added `size?: number` to `ResourceStopOptions` in `trackManualResources.ts`
  - Wired size through `emitResource` into r`awRumEvent.resource.size`
  - Forwarded `options.size` in the `stopResource` public API handler in `rumPublicApi.ts`
  
## Test instructions

Unit tests cover the new field:
- `trackManualResources.spec.ts` — verifies `resource.size` is set when size is passed to `stopResource`, and `undefined` when omitted
  - `rumPublicApi.spec.ts` — verifies size is forwarded from the public API through to the strategy's `stopResource` call

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/resource/trackManualResources.ts
+++ b/packages/rum-core/src/domain/resource/trackManualResources.ts
@@ -109,7 +109,7 @@ export function trackManualResources(lifeCycle: LifeCycle, resourceTracker: Even
       rawRumEvent,
       startClocks: stopped.startClocks,
       duration,
-      domainContext: { isManual: true as const },
+      domainContext: { isManual: true },
     })
   }
 


packages/rum-core/src/domain/resource/trackManualResources.ts
```diff
@@ -114,7 +121,7 @@ export function trackManualResources(lifeCycle: LifeCycle, resourceTracker: Even
       return
     }
 
-    emitResource(stopped.id, stopped.startClocks, stopped, options.statusCode, stopClocks)
+    emitResource(stopped.id, stopped.startClocks, stopped, options.statusCode, options.size, stopClocks)
```
BenoitZugmeyer: **💬 suggestion:** ‏inline `emitResource` so you don't have to pass more and more arguments to a function that is only called at a single place

BenoitZugmeyer (approved):

PR #4298, opened 2026-03-09
[RUM-13793] 🐛 Skip potential sanitize updates on unaltered fields

bdibon:
## Motivation

When a RumResourceEvent has a `resource.url` field that is too long one of two things can happen:

1. **beforeSend is used:** the `sanitize()` function will return `undefined` and the `limitModification` function will set the `resource.url` field to `undefined`, effectively deleting it. This caused problem on web-ui (now fixed).

2. **beforeSend is not used:** the `createBatch()` function will discard the message that contains the RumResourceEvent, we're loosing data.

---
<details>

<summary>Reflection behind the changes</summary>

### Sending invalid data

The problem that surfaced to our users is now fixed, web-ui doesn't break when a RumResourceEvent has no `resource.url`.

The SDK shouldn't send invalid data to intake, this can be fixed in several ways:

- For the specific `resource.url` field we can simply truncate it instead of setting it to undefined, then "how much do we truncate?", at lest enough for the message to intake to not be discarded
- Have a generic fix that replaces every field sanitized to `undefined` by either `""` or `{}`

### Inconsistent behavior

Also note the SDK is behaving in an inconsistent manner:

- if you don't pass a `beforeSend()` in the RumInitConfiguration, events with unsafe fields (large URL for example) will be discarded by the transport layer. **The event is discarded**
- if you do pass a `beforeSend()` that doesn't alter the `event`, unsafe fields will be sanitized, which mean they will be either set to undefined (string too long), or truncated (the object is too big); **The event or context is updated while the user never intended to do it.**

### Summary

Then we have two problems to solve:
1. invalid data sent to intake
2. inconsistent behavior with `beforeSend()`, 

We solve 1. by not erasing required fields.

We solve 2. by not applying the sanitize process to fields the user hasn't modified, this requires us to compare the original value `object[field]` with the new `value` in `setNestedValue()`. This is easy for string values, but the implementation might be more involved for objects and potentially slow.

For this PR, I've have decided to fix 1 completely and 2 partially.

For string values, we don't go through the sanitize process if they didn't change.

For object values, we still go through it, at the risk of tempering the value if it's a large object (serialization wise), meaning the resulting object might be truncated. I think this is an acceptable outcome given the low probability of occurence (objects whose serialized version length is superior to 225 280 characters).

</details>

## Changes

After second feedback round:

- Skip sanitize() calls for untouched string values
- Add a related unit test

<details>
<summary>Changes history</summary>

After first feedback round:

- ~~When a field is erased by the sanitize function, just replace it with `""` or `{}` depending on the `fieldType`~~
- ~~Add a related unit test~~

Before feedback:

- ~~Save resource.url before beforeSend runs, so the original value can be restored if the user callback corrupts it~~
- ~~Truncate resource.url to 32 KiB if it exceeds that limit (or was set to undefined by beforeSend), with a display.warn message~~
- ~~Add two unit tests in assembly.spec.ts covering truncation both with and without a beforeSend callback.~~

</details>

## Test instructions

- Manual test (use the sandbox):

```html
<script>
  fetch('/' + 'a'.repeat(300_000))
</script>
```

- Unit tests:

```sh
yarn test:unit --spec 'packages/rum-core/src/domain/limitModification.spec.ts'
```

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/limitModification.spec.ts
+++ b/packages/rum-core/src/domain/limitModification.spec.ts
@@ -1,5 +1,5 @@
 import type { Context } from '@datadog/browser-core'
-import { display, noop, objectEntries } from '@datadog/browser-core'
+import { display, noop, objectEntries, SANITIZE_DEFAULT_MAX_CHARACTER_COUNT } from '@datadog/browser-core'
 import type { ModifiableFieldPaths } from './limitModification'
 import { limitModification } from './limitModification'
 
@@ -197,7 +197,7 @@ describe('limitModification', () => {
 
   it("should not reset unsafe literal fields that the user didn't alter", () => {
     spyOn(display, 'warn')
-    const wayTooLongUrl = `/${'a'.repeat(300_000)}`
+    const wayTooLongUrl = `/${'a'.repeat(SANITIZE_DEFAULT_MAX_CHARACTER_COUNT + 1)}`
     const object: Context = { resource: { url: wayTooLongUrl } }
     const modifier = noop
     limitModification(object, { 'resource.url': 'string' }, modifier)
@@ -206,7 +206,7 @@ describe('limitModification', () => {
 
   it('should sanitize object fields (fieldType "object") even with a noop modifier', () => {
     spyOn(display, 'warn')
-    const wayTooLongUrl = `/${'a'.repeat(300_000)}`
+    const wayTooLongUrl = `/${'a'.repeat(SANITIZE_DEFAULT_MAX_CHARACTER_COUNT + 1)}`
     const object: Context = {
       context: {
         response: {
--- a/packages/core/src/tools/serialisation/sanitize.ts
+++ b/packages/core/src/tools/serialisation/sanitize.ts
@@ -27,7 +27,7 @@ interface SanitizedEvent extends Context {
 
 // The maximum size of a single event is 256KiB. By default, we ensure that user-provided data
 // going through sanitize fits inside our events, while leaving room for other contexts, metadata, ...
-const SANITIZE_DEFAULT_MAX_CHARACTER_COUNT = 220 * ONE_KIBI_BYTE
+export const SANITIZE_DEFAULT_MAX_CHARACTER_COUNT = 220 * ONE_KIBI_BYTE
 
 // Symbol for the root element of the JSONPath used for visited objects
 const JSON_PATH_ROOT_ELEMENT = '$'


packages/rum-core/src/domain/assembly.ts
```diff
@@ -141,3 +147,27 @@ function shouldSend(
...
+    return { resourceUrl: (event as RumResourceEvent).resource.url }
+  }
+  return {}
+}
+
+function ensureLegalPayload(event: AssembledRumEvent, savedFields: SavedFields) {
+  if (event.type === RumEventType.RESOURCE) {
+    const resourceEvent = event as RumResourceEvent
+    // eslint-disable-next-line eqeqeq
+    if (resourceEvent.resource.url == undefined || resourceEvent.resource.url.length > URL_BYTES_LIMIT) {
```
bcaudan: **❓ question:** ‏any reason to disable the lint rule?

packages/rum-core/src/domain/assembly.ts
```diff
@@ -141,3 +147,27 @@ function shouldSend(
...
+      resourceEvent.resource.url = savedFields.resourceUrl!.slice(0, URL_BYTES_LIMIT)
+      display.warn(`Resource URL is too long, truncated to ${URL_BYTES_LIMIT} characters`)
```
bcaudan: **💬 suggestion:** ‏let's mock the warn in the unit tests to avoid flooding the console

packages/rum-core/src/domain/assembly.ts
```diff
@@ -109,7 +111,11 @@ export function startRumAssembly(
         ddtags: buildTags(configuration).join(','),
       }) as AssembledRumEvent
 
+      const savedFields = saveFieldsOfInterest(serverRumEvent)
+
       if (shouldSend(serverRumEvent, configuration.beforeSend, domainContext, eventRateLimiters)) {
+        ensureLegalPayload(serverRumEvent, savedFields)
```
bcaudan: **💭 thought:** it seems that we are working around the issue by saving the resource url of the event here and doing the truncation after the beforeSend, I won't probably scale well if we want to add more fields.

Since other too big fields could lead to events being dropped, should we try to do the same thing for all fields regardless of the event type?
FMU, the intent for sanitize was to prevent wrong values provided by customers, could we limit the sanitization to only the values that have changed to avoid the confusing behavior?

I think it could be nice to dig a bit deeper on what is happening here and how we would like to solve this.

packages/rum-core/src/domain/assembly.ts
bdibon: Good point!

packages/rum-core/src/domain/assembly.ts
bdibon: > could we limit the sanitization to only the values that have changed to avoid the confusing behavior?

Indeed, I think it's the problem we need to fix first here. I have addressed this issue in the last commit.

packages/rum-core/src/domain/limitModification.ts
```diff
@@ -52,6 +52,16 @@ function setNestedValue(
   value: unknown,
   fieldType: 'string' | 'object'
 ) {
+  // When the value is the same as object[field], and it's unsafe,
+  // we don't want to tamper with it.
+  // This is to have a consistent SDK behavior when the user doesn't provide a beforeSend function
```
bcaudan: **🥜 nitpick:** ‏attempt to clarify + avoid mentioning beforeSend as `limitModification` is not coupled with to `beforeSend` usage

```suggestion
  // Do not sanitize fields that have not been touched by the modifier.
  // This is to have a consistent SDK behavior wether limitModification is called or not on the original object.
```

packages/rum-core/src/domain/limitModification.ts
```diff
@@ -52,6 +52,16 @@ function setNestedValue(
...
+  // and when it provides one that doesn't modify the event.context.
+  // This fix only works for strings, for example when the event.context is untouched,
+  // it will still go through the sanitize process. It's acceptable given the unlikely eventuality.
```
bcaudan: **❓ question:** ‏I don't get what you have in mind around `event.context`. Could you clarify?

packages/rum-core/src/domain/limitModification.spec.ts
```diff
@@ -194,6 +194,18 @@ describe('limitModification', () => {
...
+
+  it("should not reset unsafe fields that the user didn't alter", () => {
+    spyOn(display, 'warn')
+    const wayTooLongUrl = `/${'a'.repeat(300_000)}`
+    const object: Context = { resource: { url: wayTooLongUrl } }
+    const modifier = function () {
+      /* Identity modifier */
+    }
+
+    limitModification(object, { 'resource.url': 'string' }, modifier)
```
bcaudan: **💬 suggestion:** ‏you could use `noop` utility directly


```suggestion
   limitModification(object, { 'resource.url': 'string' }, noop)
```

packages/rum-core/src/domain/limitModification.ts
bdibon: The hard part of this "skip sanitization unless the value has changed" is to compare values that are objects, here `object[field] === value`. For a RumEvent the only "object" field that is modifiable by the user is `context`, see https://github.com/DataDog/browser-sdk/blob/d8c5490e69374d8e22c17c38ea2c298b50e852aa/packages/rum-core/src/domain/assembly.ts#L27-L29

bcaudan: oh ok, I think I get it now.
With the current implementation, If the field is an object, updating a value of this object won't trigger a sanitize but updating the object reference will 🤔
What about using `deepEqual` when the field is an object? 
This way it is more "Sanitize if it has been modified".

Regardless of the approach, we should add tests on these behaviors as well.

bdibon: > With the current implementation, If the field is an object, updating a value of this object won't trigger a sanitize but updating the object reference will 🤔

The `object[field] === value` will never be true here for `fieldType = "object"`, because we end up comparing the original object field with the one from the deep clone.

> What about using deepEqual when the field is an object?
> This way it is more "Sanitize if it has been modified".

I am not sure it's worth to do the deepEqual check, let's review this together.

**It's not free.**

deepEqual can be an expensive operation:
- at runtime, both in time and memory, depending on the shape of the object; plus we would apply it on every event
- implementation wise: we introduce more complexity

I think the bundle size increase would be insignificant here.

**What's the value?**

If we don't do the deepEqual, the following awkward flow might happen:

beforeSend() in config > "cloneContext" is unsafe (larger than 220 Kib) > beforeSend made no update to cloneContext > Unexpected Modified RumEvent Context

This is a bad outcome because the context of the RumEvent has been modified implicitly.

I think it is mitigated:
- how likely is it to happen?
- the event will likely be discarded by the transport layer (as it would have been originally)

What do you think?

bcaudan: > The object[field] === value will never be true here for fieldType = "object", because we end up comparing the original object field with the one from the deep clone.

oh yeah, indeed.

> deepEqual can be an expensive operation

agree on end user resource impact, implementation wise I may miss the complexity that you anticipate

> the following awkward flow might happen

yeah, I think we can live without it.

OK for me with current implementation but:
- let's make it clear the this function does not sanitize modifiable field of types object, may be worth moving all the aded comments to the function description
- let's have a test on that behavior

packages/rum-core/src/domain/limitModification.spec.ts
```diff
@@ -194,6 +194,38 @@ describe('limitModification', () => {
     limitModification(object, { bar: 'object' }, modifier)
     expect(() => JSON.stringify(object)).not.toThrowError()
   })
+
+  it("should not reset unsafe literal fields that the user didn't alter", () => {
+    spyOn(display, 'warn')
+    const wayTooLongUrl = `/${'a'.repeat(300_000)}`
```
BenoitZugmeyer: **suggestion:** Instead of using a "magic number", could you use something like:

```suggestion
    const wayTooLongUrl = `/${'a'.repeat(SANITIZE_DEFAULT_MAX_CHARACTER_COUNT + 1)}`
```
(import the constant from `packages/core/src/tools/serialisation/sanitize.ts`)

PR #4301, opened 2026-03-09
✨ Add @datadog/browser-rum-vue integration package

mormubis:
## Motivation

Vue doesn't have a first-class Datadog RUM integration yet. Users who want automatic view tracking, error capture with component context, or component performance measurement have to wire it all up by hand. This adds `@datadog/browser-rum-vue`.

## Changes

Three things in the new package, mirroring what `rum-react` does for React:

**Error tracking** — `addVueError` matches Vue's `app.config.errorHandler` signature exactly, so you can just do `app.config.errorHandler = addVueError`. No wrapper. Vue's `info` string (like `"mounted hook"`) goes into `component_stack`.

**Vue Router v4** — swap `import { createRouter } from 'vue-router'` for `import { createRouter } from '@datadog/browser-rum-vue/vue-router'` and you get automatic view tracking with parameterized route names like `/users/:id`. One thing that took a bit of thought: `afterEach` fires for blocked navigations too (e.g. an auth guard turning someone away), so we filter those out — details in a comment on the diff.

**Component timing** — `UNSTABLE_useVueComponentTracker('MyComponent')` in `setup()` measures mount and update duration via lifecycle hooks. Reports as `vueComponentRender` vitals, same name as the React one, so cross-framework dashboards just work.

Also a Vite sample app at `test/apps/vue-router-app` if you want to poke at it.

## Test instructions

1. Build the packages and sample app:
   ```bash
   yarn build:apps
   ```
2. Start the dev server:
   ```bash
   cd test/apps/vue-router-app && yarn dev
   ```
3. Open the app and verify:
   - Navigating between **Home → User 42 → Tracked → Error** creates a new RUM view for each route (names: `/`, `/user/:id`, `/tracked`, `/error`)
   - Clicking "Trigger Error" sends an error event with `context.framework: "vue"` and `component_stack: "v-on handler"`
   - Visiting the Tracked page shows a `vueComponentRender` vital in Custom Vitals

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [x] Added unit tests for this change (19 tests across 5 spec files)
- [ ] Added e2e/integration tests (planned follow-up once the API is reviewed)
- [ ] Updated documentation (will come once the API is settled)

### DIFF
--- a/test/unit/karma.base.conf.js
+++ b/test/unit/karma.base.conf.js
@@ -126,7 +126,7 @@ function overrideTsLoaderRule(module) {
   // We use swc-loader to transpile some dependencies that are using syntax not compatible with browsers we use for testing
   module.rules.push({
     test: /\.m?js$/,
-    include: /node_modules\/(react|react-router-dom|react-dom|react-router|turbo-stream)/,
+    include: /node_modules\/(react|react-router-dom|react-dom|react-router|turbo-stream|vue-router)/,
     use: {
       loader: 'swc-loader',
       options: {


packages/rum-vue/src/domain/router/vueRouter.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { RouterOptions, Router } from 'vue-router'
+import { startVueRouterView } from './startVueRouterView'
+
+export function createRouter(options: RouterOptions): Router {
+  const router = originalCreateRouter(options)
+
+  // afterEach fires for the initial navigation when the app is mounted via app.use(router).
+  // In tests without mounting, an explicit router.push() is needed to trigger the hook.
+  router.afterEach((to, _from, failure) => {
+    if (failure && !isNavigationFailure(failure, NavigationFailureType.duplicated)) {
```
mormubis, line 11: Vue's `afterEach` fires even for navigations that got blocked by a guard — not just successful ones. The `failure` parameter tells you why it didn't complete. We skip `cancelled` (preempted by another navigation) and `aborted` (guard explicitly blocked it). `duplicated` (same route you're already on) we let through, since those aren't really failures — the URL just didn't change because it was already there.

packages/rum-vue/src/domain/router/startVueRouterView.ts
```diff
@@ -0,0 +1,42 @@
...
+  for (const routeRecord of matched) {
+    const path = routeRecord.path
+    if (!path) {
+      continue
+    }
+
+    // Note: Vue Router normalizes all paths in the matched array to absolute paths,
+    // so the relative-path branch below is purely defensive and not expected to be
+    // hit in practice. It mirrors the React Router implementation for consistency.
+    if (path.startsWith('/')) {
```
mormubis, line 31: Vue Router always gives you absolute paths in `matched`, unlike React Router which can give you relative fragments. So the `else` branch here will never actually run — it's just mirroring the React implementation. Didn't want to delete it and diverge unnecessarily.

PR #4315, opened 2026-03-11
⚗️ Add Angular Router integration

rgaignault:
## Motivation

The RUM Browser SDK supports React but has no Angular equivalent. Angular is widely used and customers need RUM views aligned with their Angular route definitions — using route patterns like `/article/:articleId` instead of resolved URLs like `/article/2`.

This PR adds `@datadog/browser-rum-angular`, starting with Angular Router integration as described in the [Angular integration RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6355583242/RFC+Angular+router+integration).

## Changes

**New package: `@datadog/browser-rum-angular`**

- `angularPlugin({ router: true })` — RUM plugin that enables manual view tracking when router integration is active
- `provideDatadogRouter()` — Angular provider that listens to `GuardsCheckEnd` events, computes the view name from the route snapshot tree, and starts a new RUM view
- `computeRouteViewName()` — Walks the primary outlet chain to build a stable view name from route patterns (e.g. `/users/:id/posts/:postId`), skipping layout wrappers and named outlets

Uses `GuardsCheckEnd` rather than `NavigationEnd` so that resolver data fetches are attributed to the correct view.

**Test app & E2E**

- Angular 15 test app with static, parameterized, and lazy-loaded routes
- AOT compilation via `@ngtools/webpack` + Angular linker (no `unsafe-eval` needed)
- E2E scenarios validating view names for navigation and nested routes

## Test instructions

```bash
yarn test:unit --spec packages/rum-angular
yarn test:e2e -g "angular plugin"
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/.github/workflows/deploy-generated-docs.yml
+++ b/.github/workflows/deploy-generated-docs.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Upload artifact
         uses: actions/upload-pages-artifact@7b1f4a764d45c48632c6b24a0339c27f5614fb0b # v4.0.0
         with:
-          path: './docs'
+          path: './generated-docs'
 
       - name: Deploy to GitHub Pages
         id: deployment
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ specs.log
 /test-report/
 browserstack.err
 package.tgz
-docs/
+generated-docs/
 *.tsbuildinfo
 /developer-extension/dist
 /developer-extension/.wxt
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,5 +1,5 @@
 variables:
-  CURRENT_STAGING: staging-12
+  CURRENT_STAGING: staging-13
   APP: 'browser-sdk'
   CURRENT_CI_IMAGE: 102
   BUILD_STABLE_REGISTRY: 'registry.ddbuild.io'
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,6 +8,6 @@ test/**/dist
 test/**/.next
 test/apps/nextjs/next-env.d.ts
 yarn.lock
-/docs
+/generated-docs
 /developer-extension/.output
 /developer-extension/.wxt
--- a/.wokeignore
+++ b/.wokeignore
@@ -1,7 +1,7 @@
 **/dist/**
 yarn.lock
 .yarn/releases
-docs
+generated-docs
 
 # code import
 packages/worker/src/domain/deflate.js
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,6 +60,8 @@ packages/
 
 developer-extension/ # Chrome DevTools extension
 
+docs/                # Repository documentation
+
 test/
 ├── apps/            # Test apps for E2E and performance testing
 ├── e2e/             # Playwright E2E test scenarios
@@ -68,6 +70,17 @@ test/
 
 scripts/             # Build, deploy, release automation
 ```
+
+## Documentation
+
+For deeper context, see:
+
+- `docs/DEVELOPMENT.md` — commit conventions, dependency management, schema generation, TS compatibility
+- `docs/CONVENTIONS.md` — coding style, file organization, size control
+- `docs/ARCHITECTURE.md` — package dependencies, data pipeline
+- `docs/TESTING.md` — local testing strategies, debugging flaky tests
+- `scripts/AGENTS.md` — writing and organizing scripts in this codebase
+- `test/e2e/AGENTS.md` — E2E test writing guide (createTest builder, IntakeRegistry, patterns)
 
 ## Critical Patterns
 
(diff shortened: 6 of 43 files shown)


test/e2e/scenario/angularPlugin.scenario.ts
```diff
@@ -0,0 +1,51 @@
+import { test, expect } from '@playwright/test'
+import { createTest } from '../lib/framework'
+
+test.describe('angular plugin', () => {
+  createTest('should define a view name based on the route')
+    .withRum()
+    .withReactApp('angular-app')
```
rgaignault: 💭 Suggestion : Let's add a `withAngularApp`

Something Like this 

```
  withAngularApp(appName: string) {
    this.setups = [{ factory: (options, servers) => reactSetup(options, servers, appName) }]
    return this
  }
```

packages/rum-angular/src/domain/angularRouter/provideDatadogRouter.ts
```diff
@@ -0,0 +1,56 @@
...
+  return makeEnvironmentProviders([
+    {
+      // Needed for Angular v15 support (provideEnvironmentInitializer requires v16+)
+      provide: ENVIRONMENT_INITIALIZER,
+      multi: true,
+      useFactory: () => {
+        const router = inject(Router)
+
+        return () => {
+          // No unsubscribe needed as its for the full app lifecycle and because DestroyRef requires v16+
```
rgaignault, line 39: Though 💭 : Should we not wrap the router subscription in onRumInit like react ?

amortemousque: Indeed! nice catch

test/e2e/scenario/angularPlugin.scenario.ts
amortemousque: Oh my, that’s an ugly slope. Thanks for the review.
I generalized `withReactApp` to `withApp(appName)`.

amortemousque (approved):

packages/rum-angular/src/domain/angularPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+      for (const subscriber of onRumInitSubscribers) {
+        subscriber(globalConfiguration, globalPublicApi)
+      }
+      if (configuration.router) {
+        initConfiguration.trackViewsManually = true
+      }
+    },
+    onRumStart({ addError }) {
+      globalAddError = addError
+      for (const subscriber of onRumStartSubscribers) {
```
BeltranBulbarellaDD: Same as you said to me, we can move the if outside.

test/apps/angular-app/main.ts
```diff
@@ -0,0 +1,87 @@
...
+  { path: '', component: InitialRouteComponent },
+  { path: 'parameterized/:id', component: ParameterizedRouteComponent },
+  { path: 'parent', loadChildren: () => Promise.resolve(nestedRoutes) },
+  { path: '**', component: WildcardRouteComponent },
+]
+
+// Bootstrap - dynamically create root element (E2E framework serves bare HTML)
+const rootElement = document.createElement('app-root')
+document.body.appendChild(rootElement)
+// eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
BeltranBulbarellaDD: Why is this an unsafe-call? Does it mean customers have to add this too?

test/e2e/scenario/angularPlugin.scenario.ts
```diff
@@ -0,0 +1,54 @@
...
+test.describe('angular plugin', () => {
+  createTest('should define a view name based on the route')
+    .withRum()
+    .withApp('angular-app')
+    .run(async ({ page, flushEvents, intakeRegistry }) => {
+      await page.click('text=Go to Parameterized Route')
+      await flushEvents()
+      const viewEvents = intakeRegistry.rumViewEvents
+      expect(viewEvents.length).toBeGreaterThan(0)
+      const lastView = viewEvents[viewEvents.length - 1]
```
BeltranBulbarellaDD, line 13: Can we use `view. referrer`? instead of `[viewEvents. length - 1]`

packages/rum-angular/src/domain/angularRouter/provideDatadogRouter.ts
```diff
@@ -0,0 +1,53 @@
+import type { EnvironmentProviders } from '@angular/core'
+// eslint-disable-next-line local-rules/disallow-side-effects
```
BenoitZugmeyer: **suggestion:** Remove those comments, mark those new packages as without sideeffect [in the rule](https://github.com/DataDog/browser-sdk/blob/fcbf67aa4e2a17924cac39793adbf97cea6ace18/eslint-local-rules/disallowSideEffects.js#L35-L42)

packages/rum-angular/src/domain/angularRouter/startAngularView.ts
```diff
@@ -0,0 +1,47 @@
...
+      segments.push(path)
+    }
+
+    // Follow primary outlet only — named outlets (e.g. sidebar) are excluded
+    current = current.children.find((child) => child.outlet === PRIMARY_OUTLET)
+  }
+
+  if (segments.length === 0) {
+    return '/'
+  }
```
BenoitZugmeyer: **nitpick:** this is unnecessary

packages/rum-angular/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@angular/router": ">=15 <=21",
+    "rxjs": ">=7"
+  },
+  "peerDependenciesMeta": {
+    "@angular/core": {
+      "optional": true
+    },
+    "@angular/router": {
+      "optional": true
+    }
```
BenoitZugmeyer: **suggestion:** those dependencies are directly imported by the main entry point. It shouldn't be optional.

test/apps/angular-app/webpack.config.js
BenoitZugmeyer, line 1: **suggestion:** this is not the canonical way to build angular apps. What about using `ng`?

test/e2e/lib/framework/serverApps/mock.ts
```diff
@@ -196,7 +196,7 @@ export function createMockServerApp(
     }
   })
 
-  app.get(/(?<appName>app|react-[\w-]+).js$/, (req, res) => {
+  app.get(/(?<appName>app|react-[\w-]+|angular-[\w-]+).js$/, (req, res) => {
```
BenoitZugmeyer, line 199: **thought:** this doesn't scale... we should probably do something about it at some point.

amortemousque: Agree, let's revisit in another PR. I'll create a card ;)

test/apps/angular-app/webpack.config.js
amortemousque: Yes, I agree. As discussed, switching to a canonical approach would require much more work. We should revisit how we handle the e2e test setup. Let’s decouple that from this task. I’ll keep it as is if that’s okay for you.

test/e2e/scenario/angularPlugin.scenario.ts
amortemousque: what do you mean exactly?

test/apps/angular-app/main.ts
amortemousque: This is no longer required. I removed the comment

test/e2e/scenario/angularPlugin.scenario.ts
BeltranBulbarellaDD: Oh no my bad, I though you wanted to check the previous view!. All good!

PR #4316, opened 2026-03-12
♻️ NextJS- Rename test app and update e2e tests. 

BeltranBulbarellaDD:
## Motivation

The `datadogRouters.spec.ts` unit tests required maintaining karma webpack aliases and stub files for `next/navigation` and `next/router.` These tests are already covered by the E2E suite, so the stubs and unit tests are unnecessary overhead. The test app was also renamed to `nextjs` now that it covers both the app and pages router.

## Changes

- Renamed test app `nextjs-app-router` → `nextjs`
- Removed `datadogRouters.spec.ts` unit tests — router behaviour is covered by E2E
- Removed `next/navigation` and `next/router` karma stubs (and the webpack aliases that loaded them)
- Restructured E2E scenarios under a single nextjs - router describe block, iterating each router config inside both app and pages router test cases
 
## Test instructions

Unit and E2E tests should pass.
`yarn test:unit`
`yarn test:e2e -g "nextjs"`

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/scenario/nextjsPlugin.scenario.ts
+++ b/test/e2e/scenario/nextjsPlugin.scenario.ts
@@ -7,14 +7,12 @@ const routerConfigs = [
     router: 'app' as const,
     viewPrefix: '',
     homeUrlPattern: '**/',
-    clientErrorMessage: 'Client error from error-test',
   },
   {
     name: 'nextjs pages router',
     router: 'pages' as const,
     viewPrefix: '/pages-router',
     homeUrlPattern: /\/pages-router(\?|$)/,
-    clientErrorMessage: 'Pages Router error from NextjsErrorBoundary',
   },
 ]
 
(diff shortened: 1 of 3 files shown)


test/e2e/scenario/nextjsPlugin.scenario.ts
```diff
@@ -1,165 +1,181 @@
...
+    viewPrefix: '',
+    homeUrlPattern: '**/',
+    clientErrorMessage: 'Client error from error-test',
+  },
   {
     name: 'nextjs pages router',
     router: 'pages' as const,
     viewPrefix: '/pages-router',
     homeUrlPattern: /\/pages-router(\?|$)/,
+    clientErrorMessage: 'Pages Router error from NextjsErrorBoundary',
```
rgaignault: 💬 Suggestion : Is this used somewhere ? (maybe for a future PR ? ) if not consider removing it

rgaignault (approved):

test/e2e/scenario/nextjsPlugin.scenario.ts
BeltranBulbarellaDD: Oh yeah. Nice catch. It will be used in the next PR when we are testing errors. I forgot to remove it! But now we can keep it since it's for next PR 🙂

BeltranBulbarellaDD: Made an update with an extra test so I removed too unneeded this param.

PR #4308, opened 2026-03-10
✨ NextJS Add addNextJSError component

BeltranBulbarellaDD:
## Motivation

Discussed in this [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6247321353/RFC+-+Browser-SDK+NextJS+Integration) specifically [here](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6247321353/RFC+-+Browser-SDK+NextJS+Integration#Next.js-App-Router-Error-Handling)

## Changes
(AI written)

Adds `addNextjsError(error, context?)` and `NextjsErrorBoundary` to `@datadog/browser-rum-nextjs` — the integration points for reporting errors from Next.js error boundaries to Datadog RUM.

**`addNextjsError(error, context?)`**
- Tags every error with `framework: 'nextjs'` context
- Adds `nextjs.digest` context when the error carries a `digest` property (server component errors — the hash links client-side errors to server-side logs)
- Uses `onRumStart` deferred callback so calls before SDK init are buffered and replayed
- Reserved fields (`framework`, `nextjs.digest`) are always applied last in the context merge, so user-provided context cannot overwrite them

**`NextjsErrorBoundary`**
- React class component wrapping `addNextjsError` for Pages Router client-side error catching
- Accepts a `fallback` render prop (`NextjsErrorBoundaryFallback`) and an optional `onError` callback
- Automatically calls `addNextjsError` in `componentDidCatch`

**Test app & E2E**
- Renamed test app from `nextjs-app-router` → `nextjs` (now covers both App Router and Pages Router)
- Added Pages Router test pages: `pages-router/` index, user, guides, error boundary test
- E2E scenarios cover: router navigation (app + pages), client error (app + pages), server error with digest (app router), global error (app router), `NextjsErrorBoundary` catch + reset (pages router)

Both APIs exported from `@datadog/browser-rum-nextjs` main entry.

## Test instructions
(AI written)

```bash
# Unit tests
yarn test:unit --spec packages/rum-nextjs/src/domain/error/addNextjsError.spec.ts

# E2E tests
yarn test:e2e -g "nextjs"
```
Navigating into `test/apps/nextjs` you can do `yarn dev` or `yarn start` (for production) and use the developer extension to test the events.

## Checklist

- [X] Tested locally
- [ ] Tested on staging
- [X] Added unit tests for this change.
- [X] Added e2e/integration tests for this change.
- [X] Updated documentation and/or relevant AGENTS.md file

PR #4317, opened 2026-03-12
♻️ React- Replace addEvent for AddError, Export Error Boundary

BeltranBulbarellaDD:
## Motivation

The `addReactError` function was building and dispatching a raw RUM error event directly, duplicating logic already handled by `addError` in `rum-core`. 
Additionally, `ErrorBoundary` was not composable, there was no way to create a customised error boundary that reports errors through a different mechanism (e.g. for the upcoming `Next.js` integration)

The idea is to refactor `addReactError` and `ErrorBoundary` without changing functionality.

## Changes

- `reactPlugin.ts`: replaced `addEvent` with `addError` as the value passed to `onRumStart` subscribers, react plugin no longer needs access to the raw event API
- `addReactError.ts`: simplified to delegate directly to `addError()`, removing the manual `computeRawError` / event construction
- `errorBoundary.ts`: introduced `BaseErrorBoundary` abstract class and `createErrorBoundary`(`reportError`, `displayName`) factory, so error boundaries can be composed with any error-reporting callback; `ErrorBoundary` now extends `BaseErrorBoundary`
- `error/index.ts`: exported `createErrorBoundary`
- Added `src/entries/errorBoundary.ts` entry point and `error-boundary/package.json` for the `@datadog/browser-rum-react/error-boundary` subpath export
- Updated `initializeReactPlugin` test helper and all affected specs accordingly

Usage of the changes in `Error Boundary` can be seen [here](https://github.com/DataDog/browser-sdk/pull/4308/changes#diff-99317a0ecf39572e525afa087a77bc1860c8779b15be54efec465d099a791cb4)

## Test instructions

Old unit and e2e test should pass. 
Added extra validation unit and e2e tests for the new `addError` function.

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-react/src/domain/error/errorBoundary.ts
+++ b/packages/rum-react/src/domain/error/errorBoundary.ts
@@ -21,46 +21,40 @@ export type State =
 
 const INITIAL_STATE: State = { didCatch: false, error: null }
 
-// eslint-disable-next-line no-restricted-syntax
-export abstract class BaseErrorBoundary extends React.Component<ErrorBoundaryProps, State> {
-  constructor(props: ErrorBoundaryProps) {
-    super(props)
-    this.state = INITIAL_STATE
-  }
-
-  static getDerivedStateFromError(error: Error): State {
-    return { didCatch: true, error }
-  }
-
-  resetError = () => {
-    this.setState(INITIAL_STATE)
-  }
-
-  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
-    this.reportError(error, errorInfo)
-  }
-
-  render() {
-    if (this.state.didCatch) {
-      return React.createElement(this.props.fallback, {
-        error: this.state.error,
-        resetError: this.resetError,
-      })
-    }
-
-    return this.props.children
-  }
-
-  protected abstract reportError(error: Error, errorInfo: ErrorInfo): void
-}
-
 export function createErrorBoundary(
   reportError: (error: Error, errorInfo: ErrorInfo) => void,
   displayName = 'ErrorBoundary'
 ) {
-  const DatadogErrorBoundary: React.ComponentClass<ErrorBoundaryProps> = class DatadogErrorBoundary extends BaseErrorBoundary {
-    protected reportError(error: Error, errorInfo: ErrorInfo) {
+  const DatadogErrorBoundary: React.ComponentClass<ErrorBoundaryProps> = class DatadogErrorBoundary extends React.Component<
+    ErrorBoundaryProps,
+    State
+  > {
+    constructor(props: ErrorBoundaryProps) {
+      super(props)
+      this.state = INITIAL_STATE
+    }
+
+    static getDerivedStateFromError(error: Error): State {
+      return { didCatch: true, error }
+    }
+
+    resetError = () => {
+      this.setState(INITIAL_STATE)
+    }
+
+    componentDidCatch(error: Error, errorInfo: ErrorInfo) {
       reportError(error, errorInfo)
+    }
+
+    render() {
+      if (this.state.didCatch) {
+        return React.createElement(this.props.fallback, {
+          error: this.state.error,
+          resetError: this.resetError,
+        })
+      }
+
+      return this.props.children
     }
   }
 
@@ -84,9 +78,5 @@ export function createErrorBoundary(
  * </ErrorBoundary>
  * ```
  */
-// eslint-disable-next-line no-restricted-syntax
-export class ErrorBoundary extends BaseErrorBoundary {
-  protected reportError(error: Error, errorInfo: ErrorInfo) {
-    addReactError(error, errorInfo)
-  }
-}
+// eslint-disable-next-line local-rules/disallow-side-effects
+export const ErrorBoundary = createErrorBoundary(addReactError)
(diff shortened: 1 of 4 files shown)


packages/rum-react/src/domain/error/addReactError.ts
```diff
@@ -35,43 +26,15 @@ import { onRumStart } from '../reactPlugin'
...
-        originalError: error,
+      addError({
+        error,
         handlingStack,
         componentStack: info.componentStack ?? undefined,
         startClocks,
-        source: ErrorSource.CUSTOM,
-        handling: ErrorHandling.HANDLED,
-        nonErrorPrefix: NonErrorPrefix.PROVIDED,
+        context: { ...(error as Error & { dd_context?: Context }).dd_context, framework: 'react' },
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: It's intended behaviour.

packages/rum-react/src/domain/error/errorBoundary.ts
```diff
@@ -65,4 +50,43 @@ export class ErrorBoundary extends React.Component<ErrorBoundaryProps, State> {
...
+) {
+  const DatadogErrorBoundary: React.ComponentClass<ErrorBoundaryProps> = class DatadogErrorBoundary extends BaseErrorBoundary {
+    protected reportError(error: Error, errorInfo: ErrorInfo) {
+      reportError(error, errorInfo)
+    }
+  }
+
+  DatadogErrorBoundary.displayName = displayName
+
+  return DatadogErrorBoundary
```
amortemousque: **💬 suggestion:** ‏We could remove some complexity by returning the class directly from here. What do you think?

```suggestion
export function createErrorBoundary(
  reportError: (error: Error, errorInfo: ErrorInfo) => void,
  displayName = 'ErrorBoundary'
) {
  const DatadogErrorBoundary: React.ComponentClass<ErrorBoundaryProps> = class extends React.Component<
    ErrorBoundaryProps,
    State
  > {
    constructor(props: ErrorBoundaryProps) {
      super(props)
      this.state = INITIAL_STATE
    }

    static getDerivedStateFromError(error: Error): State {
      return { didCatch: true, error }
    }

    resetError = () => {
      this.setState(INITIAL_STATE)
    }

    componentDidCatch(error: Error, errorInfo: ErrorInfo) {
      reportError(error, errorInfo)
    }

    render() {
      if (this.state.didCatch) {
        return React.createElement(this.props.fallback, {
          error: this.state.error,
          resetError: this.resetError,
        })
      }

      return this.props.children
    }
  }

  DatadogErrorBoundary.displayName = displayName

  return DatadogErrorBoundary
}
```

packages/rum-react/src/entries/internal.ts
amortemousque, line 1: **❓ question:** ‏ Why do you need this in `/entries`?

BeltranBulbarellaDD: This way, users can use `createErrorBoundary` without needing to set up the full `ReactPlugin`

amortemousque: Interesting point 🤔

Setting up the ReactPlugin gives us telemetry on how many customers use the React integration, which is valuable. So I would not do it unless customers clearly ask for it.

packages/rum-react/src/domain/error/errorBoundary.ts
BeltranBulbarellaDD: Hm I don't think it's possible as `CallExpression can have side effects when the module is evaluated. Maybe move it in a function declaration?`. This way `createErrorBoundary(addReactError)` can break tree-shaking. WDYT? Should we keep it?

packages/rum-react/src/entries/internal.ts
BeltranBulbarellaDD: When you say:
>  So I would not do it unless customers clearly ask for it.

You mean having it's own entrypoint? 

I believe this approach of having an entry point is good because of the same issue as above that we don't want to depend on Rum-React which can have side-effects. And we can gather telemetry on usage of NextJS.

packages/rum-react/error-boundary/package.json
BenoitZugmeyer, line 1: **suggestion:** rename this entry point as "internal". We might export more things to share with nextjs, and we don't want to confuse people looking for the error boundary.

packages/rum-react/src/domain/error/errorBoundary.ts
BenoitZugmeyer: Let's remove that function. Export the `BaseErrorBoundary` class and extend it like you do below.

packages/rum-react/src/domain/error/errorBoundary.ts
```diff
@@ -65,4 +50,43 @@ export class ErrorBoundary extends React.Component<ErrorBoundaryProps, State> {
 
     return this.props.children
   }
+
+  protected abstract reportError(error: Error, errorInfo: ErrorInfo): void
+}
+
+export function createErrorBoundary(
```
mormubis: Relevant to the open discussion about exporting `BaseErrorBoundary` for direct extension: React explicitly recommends against component inheritance hierarchies — ["we haven't found any use cases where we would recommend creating component inheritance hierarchies"](https://legacy.reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance). Might be worth factoring that in when deciding which API to ship.

PR #4319, opened 2026-03-12
✨ Add `view_update` support to developer extension

mormubis:
## Motivation

With the `partial_view_updates` experimental feature enabled, the SDK now emits `view_update` events alongside regular `view` events. Without this change, those events would silently appear in the extension as unstyled/unlabeled — no color, no description, just a raw JSON blob. Since `view_update` is the whole point of the feature, engineers debugging it need to see these events clearly.

## Changes

Added a `ViewUpdateDescription` component that shows the document version, view name, and a count of changed fields in the event. The color map now recognizes `view_update` (same blue as `view`), and event sorting handles the optional `date` field that `view_update` carries per the schema.

Also fixed a pre-existing crash in `getViewName` — `new URL(relativeUrl)` throws when the URL is a path like `/home`, which is common in SPAs. Wrapped it in a try/catch and fall back to the raw string.

## Test instructions

1. Enable the feature flag: `datadogRum.init({ enableExperimentalFeatures: ['partial_view_updates'] })`
2. Open the Browser SDK developer extension
3. Interact with the page to generate view updates
4. Verify `view_update` events appear in blue with a description like `View update v3 · /home · 2 fields changed`

## Checklist

- [x] Tested locally
- [ ] Tested on staging

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
+++ b/developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
@@ -460,12 +460,5 @@ function Emphasis({ children }: { children: ReactNode }) {
 }
 
 function getViewName(view: { name?: string; url: string }) {
-  if (view.name) {
-    return view.name
-  }
-  try {
-    return new URL(view.url).pathname
-  } catch {
-    return view.url
-  }
+  return view.name ?? new URL(view.url).pathname
 }
--- a/developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
+++ b/developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
@@ -341,18 +341,22 @@ function ViewDescription({ event }: { event: RumViewEvent }) {
   )
 }
 
-// view.id is the only field always present in a view_update diff (routing field)
-const VIEW_UPDATE_REQUIRED_KEYS = new Set(['id'])
+// view.id is always present in a view_update diff as a routing field to identify the view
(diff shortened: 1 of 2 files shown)


developer-extension/src/panel/components/tabs/eventsTab/eventRow.tsx
```diff
@@ -425,5 +456,12 @@ function Emphasis({ children }: { children: ReactNode }) {
...
 function getViewName(view: { name?: string; url: string }) {
-  return `${view.name || new URL(view.url).pathname}`
+  if (view.name) {
+    return view.name
+  }
+  try {
+    return new URL(view.url).pathname
+  } catch {
+    return view.url
+  }
```
BenoitZugmeyer: **question:** in which case the url isn't an url?

developer-extension/src/panel/hooks/useEvents/eventCollection.ts
```diff
@@ -37,17 +37,17 @@ export function startEventCollection(strategy: EventCollectionStrategy, onEvents
 function compareEvents(a: SdkEvent, b: SdkEvent) {
   // Sort events chronologically
   if (a.date !== b.date) {
-    return b.date - a.date
+    return (b.date ?? 0) - (a.date ?? 0)
```
BenoitZugmeyer: **issue:** this will make 'view update' order to be displayed at the bottom of the list, even for new views/SPA changes

BenoitZugmeyer (approved):

thomas-lebeau (approved):

PR #4325, opened 2026-03-13
✨ scaffold @datadog/browser-rum-vue package

mormubis:
## Motivation

First of four PRs adding Vue 3 RUM integration. This one is just the empty package shell and monorepo plumbing. Nothing user-facing yet, but the next three PRs need this to exist first.

## Changes

Creates the \`@datadog/browser-rum-vue\` package skeleton: \`package.json\`, empty entry stubs, the \`vue-router-v4\` sub-entry for the router integration in PR 3.

The rest is infrastructure:
- tsconfig path mappings and ESLint allowlists so the monorepo tooling knows about the new package
- license CSV entries for \`vue\`, \`vue-router\`, \`@vue/test-utils\`, \`@vitejs/plugin-vue\`
- \`swc-loader\` rule to transpile vue-router for BrowserStack. Vue-router ships its devtools as a \`.mjs\` file with optional chaining, which older BS browsers can't parse. Same fix as the React packages already in that config.

## Test instructions

No tests here, it's all plumbing. The test suite should pass unchanged.

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [ ] Added unit tests (nothing to test, empty stubs)
- [ ] Added e2e/integration tests
- [ ] Updated documentation

PR #4327, opened 2026-03-13
✨ add vuePlugin and addVueError

mormubis:
## Motivation

Second of five PRs (#4325 is the first one). First one with actual user-facing code. After this, Vue apps can report errors to Datadog with Vue-specific context.

## Changes

`vuePlugin()` is the RUM plugin Vue apps register via `datadogRum.init({ plugins: [vuePlugin()] })`. It uses the same module-level subscriber pattern as `reactPlugin`: `onVueInit` and `onVueStart` hooks let features queue work until the plugin initializes, so call order doesn't matter. `vuePlugin({ router: true })` additionally sets `trackViewsManually: true` to disable the SDK's automatic URL tracking, which PR 3 takes over with parameterized route names.

`addVueError` matches Vue's `app.config.errorHandler` signature exactly, so it can be assigned directly without a wrapper:

```ts
app.config.errorHandler = addVueError
```

Vue's `info` string (e.g. `"mounted hook"`, `"v-on handler"`) maps to `component_stack`. `context.framework` is set to `"vue"`. Errors thrown before `datadogRum.init()` are queued and flushed when RUM starts.

The implementation uses `addError` from `StartRumResult` directly, the same pattern introduced in #4343 for `addNextjsError`, rather than constructing the event manually with `addEvent`.

## Test instructions

```bash
yarn test:unit --spec packages/rum-vue/src/domain/vuePlugin.spec.ts
yarn test:unit --spec packages/rum-vue/src/domain/error/addVueError.spec.ts
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [ ] Added e2e/integration tests
- [ ] Updated documentation

### DIFF
--- a/packages/rum-vue/src/domain/error/addVueError.ts
+++ b/packages/rum-vue/src/domain/error/addVueError.ts
@@ -1,0 +1,41 @@
+import type { ComponentPublicInstance } from 'vue'
+import { callMonitored, clocksNow, createHandlingStack } from '@datadog/browser-core'
+import { onVueStart } from '../vuePlugin'
+
+/**
+ * Add a Vue error to the RUM session.
+ *
+ * @category Error
+ * @example
+ * ```ts
+ * import { createApp } from 'vue'
+ * import { addVueError } from '@datadog/browser-rum-vue'
+ *
+ * const app = createApp(App)
+ * // Report all Vue errors to Datadog automatically
+ * app.config.errorHandler = addVueError
+ * ```
+ */
+export function addVueError(
+  error: unknown,
+  // Required by Vue's app.config.errorHandler signature, but not used by the SDK
+  _instance: ComponentPublicInstance | null,
+  info: string
+) {
+  const handlingStack = createHandlingStack('vue error')
+  const startClocks = clocksNow()
+  onVueStart((addError) => {
+    callMonitored(() => {
+      addError({
+        error,
+        handlingStack,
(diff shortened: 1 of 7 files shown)


packages/rum-vue/src/domain/vuePlugin.ts
```diff
@@ -0,0 +1,68 @@
...
+      }
+      if (configuration.router) {
+        initConfiguration.trackViewsManually = true
+      }
+    },
+    onRumStart({ addError }) {
+      globalAddError = addError
+      for (const subscriber of onRumStartSubscribers) {
+        if (addError) {
+          subscriber(addError)
```
BeltranBulbarellaDD: You can move the check outside of the for loop

packages/rum-vue/src/domain/error/addVueError.ts
```diff
@@ -0,0 +1,42 @@
...
+  const startClocks = clocksNow()
+  onVueStart((addError) => {
+    callMonitored(() => {
+      addError({
+        error,
+        handlingStack,
+        componentStack: info || undefined,
+        startClocks,
+        context: {
+          ...(typeof error === 'object' && error !== null ? (error as { dd_context?: Context }).dd_context : undefined),
```
BeltranBulbarellaDD: Is this type needed? I think you could simplify

PR #4328, opened 2026-03-13
⚗️ add Vue Router v4 view tracking

mormubis:
## Motivation

After this PR, navigating between routes in a Vue app automatically creates RUM views with parameterized names instead of raw URLs. It also improves the `component_stack` in Vue error tracking to include the full parent component hierarchy rather than just the component name.

## Changes

Exports a wrapped `createRouter` from `@datadog/browser-rum-vue/vue-router-v4` as a drop-in for the standard `vue-router` import:

```ts
import { createRouter } from '@datadog/browser-rum-vue/vue-router-v4'
```

On each navigation, `afterEach` calls `startView()` with the parameterized route name (`/users/:id` rather than `/users/42`). Failed navigations (blocked by guards, cancelled, duplicated) and query-only changes are filtered out.

`component_stack` now walks the `ComponentInternalInstance` parent chain to build the full hierarchy, matching what Vue DevTools shows.

The PR includes a `vue-router-app` test app and E2E tests for both features.

Requires `vuePlugin({ router: true })` alongside this import. That sets `trackViewsManually: true` at init so the SDK's built-in URL tracking does not race with the router integration.

## Test instructions

```bash
yarn test:unit --spec packages/rum-vue/src/domain/router/startVueRouterView.spec.ts
yarn test:unit --spec packages/rum-vue/src/domain/router/vueRouter.spec.ts
yarn test:e2e:init
yarn test:e2e -g "vue plugin"
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum-vue/src/domain/router/vueRouter.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { RouterOptions, Router } from 'vue-router'
+import { startVueRouterView } from './startVueRouterView'
+
+export function createRouter(options: RouterOptions): Router {
+  const router = originalCreateRouter(options)
+
+  // afterEach fires for the initial navigation when the app is mounted via app.use(router).
+  // In tests without mounting, an explicit router.push() is needed to trigger the hook.
+  router.afterEach((to, _from, failure) => {
+    if (failure && !isNavigationFailure(failure, NavigationFailureType.duplicated)) {
```
mormubis: Vue Router's `afterEach` fires even for navigations blocked by a guard, not just successful ones. The `failure` parameter tells you why. We skip `cancelled` (preempted by another navigation) and `aborted` (guard explicitly blocked it). `duplicated` (same route you're already on) we let through - that's not really a failure, the URL just didn't change because it was already there.

packages/rum-vue/src/domain/router/startVueRouterView.ts
```diff
@@ -0,0 +1,42 @@
+import { display } from '@datadog/browser-core'
+import type { RouteLocationMatched } from 'vue-router'
+import { onVueInit } from '../vuePlugin'
+
+export function startVueRouterView(matched: RouteLocationMatched[]) {
+  onVueInit((configuration, rumPublicApi) => {
```
amortemousque: **💬 suggestion:** For consistency with React, `onVueInit` could be renamed to onRumInit. The same likely applies to `onVueStart`.

packages/rum-vue/src/domain/router/startVueRouterView.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!configuration.router) {
+      display.warn('`router: true` is missing from the vue plugin configuration, the view will not be tracked.')
+      return
+    }
+    rumPublicApi.startView(computeViewName(matched))
+  })
+}
+
+export function computeViewName(matched: RouteLocationMatched[]): string {
```
amortemousque, line 15: **💬 suggestion:** ‏I think we are missing a case for [catch all routes](https://router.vuejs.org/guide/essentials/dynamic-matching.html#Catch-all-404-Not-found-Route)(aka wildcard). For those, in [React](https://github.com/DataDog/browser-sdk/blob/11234b4e4319c5408c1c77d5306cb13dff39d01c/packages/rum-react/src/domain/reactRouter/startReactRouterView.spec.ts#L98) we keep the path with the params.

rgaignault, line 15: ❓ question : don't we need a logic to resolve "catch-all" pattern like in  react with this function ? 

https://github.com/DataDog/browser-sdk/blob/05e66f993162708daf6b231b834f3378af69604d/packages/rum-react/src/domain/reactRouter/startReactRouterView.ts#L28

packages/rum-vue/src/domain/router/startVueRouterView.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    const warnSpy = spyOn(display, 'warn')
+    initializeVuePlugin({ configuration: {} })
+    startVueRouterView([] as unknown as RouteLocationMatched[])
+    expect(warnSpy).toHaveBeenCalledOnceWith(
+      '`router: true` is missing from the vue plugin configuration, the view will not be tracked.'
+    )
+  })
+})
+
+describe('computeViewName', () => {
```
amortemousque, line 29: **🥜 nitpick:** ‏‏It would be great to align how these tests are written across integrations. I found, using an array of cases, like in the [React tests](https://github.com/DataDog/browser-sdk/blob/11234b4e4319c5408c1c77d5306cb13dff39d01c/packages/rum-react/src/domain/reactRouter/startReactRouterView.spec.ts#L80), works well.
I tried to apply the same approach for [Angular](https://github.com/DataDog/browser-sdk/blob/2d5431978df69de58d77b6274d1801ed6e31a31e/packages/rum-angular/src/domain/angularRouter/startAngularView.spec.ts#L72). Wdyt?

test/e2e/lib/helpers/playwright.ts
```diff
@@ -5,6 +5,7 @@ import packageJson from '../../../../package.json' with { type: 'json' }
 
 export const DEV_SERVER_BASE_URL = `http://localhost:${process.env.DEV_SERVER_PORT}`
 export const NEXTJS_APP_ROUTER_PORT = process.env.NEXTJS_APP_ROUTER_PORT!
+export const VUE_ROUTER_APP_PORT = '5175'
```
BeltranBulbarellaDD: Can we do something to not have to hardcode the port?

test/e2e/lib/framework/createTest.ts
```diff
@@ -111,6 +111,20 @@ class TestBuilder {
     return this
   }
 
+  withVueApp() {
```
BeltranBulbarellaDD, line 114: Maybe we could reuse the repeated code with `nextjs` here?

test/e2e/scenario/vuePlugin.scenario.ts
```diff
@@ -0,0 +1,35 @@
+import { test, expect } from '@playwright/test'
+import { createTest } from '../lib/framework'
+
+test.describe('vue plugin', () => {
```
BeltranBulbarellaDD, line 4: This is more for me but maybe we could try to have the same sets of test for all plugins and run them together? I'll bring it up in the guild.

test/e2e/playwright.base.config.ts
```diff
@@ -53,5 +53,12 @@ export const config: Config = {
         stdout: /- Local:\s+http:\/\/localhost:(?<nextjs_app_router_port>\d+)/,
       },
     },
+    {
+      stdout: 'pipe' as const,
+      cwd: path.join(__dirname, '../apps/vue-router-app'),
+      command: isLocal ? 'yarn dev' : 'yarn preview',
```
BeltranBulbarellaDD, line 60: What do you think of only running the production one?

test/e2e/lib/helpers/playwright.ts
mormubis: I did a similar thing to NextJS

test/e2e/scenario/vuePlugin.scenario.ts
mormubis: @BeltranBulbarellaDD  I didn't want to touch `withNextjsApp` in this PR. I can do a follow-up. Thoughts?

BeltranBulbarellaDD: Sounds good! We'll discuss later!

packages/rum-vue/src/domain/router/startVueRouterView.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  const cases: Array<[string, Array<{ path: string }>, string]> = [
+    // description,                       matched paths,                                                    expected
+    ['empty matched array',               [],                                                               ''],
+    ['simple route',                      [{ path: '/users' }],                                            '/users'],
+    ['nested routes',                     [{ path: '/users' }, { path: '/users/:id' }],                   '/users/:id'],
+    ['ignores records without a path',    [{ path: '/users' }, { path: '' }, { path: '/users/:id' }],     '/users/:id'],
+    // Vue Router 4 catch-all routes use /:pathMatch(.*)*  (no bare * wildcard like React Router).
+    // We keep the route pattern as-is — it is already a meaningful identifier that groups all
+    // unmatched paths together, unlike React Router's * which needs substitution to be readable.
+    ['catch-all route',                   [{ path: '/:pathMatch(.*)*' }],                                 '/:pathMatch(.*)*'],
```
amortemousque: **💬 suggestion:** We should probably align catch-all case to other integrations (see [React](https://github.com/DataDog/browser-sdk/blob/53c9b00041153a3da833c8e7e5e9730ebaab47b7/packages/rum-react/src/domain/reactRouter/startReactRouterView.spec.ts#L99)).
Nit: could show the path (/users/1) next to or instead of the description, like React.

PR #4343, opened 2026-03-16
✨  NextJS- addNextjsError component

BeltranBulbarellaDD:
## Motivation

This PR adds error reporting support to `@datadog/browser-rum-nextjs` for the App Router. 

## Changes

**packages/rum-nextjs**
  - Added `addNextjsError(error, errorInfo?)` — a new public API, mirroring `addReactError` from `rum-react`, for reporting errors from Next.js App Router error boundaries (`error.tsx`, `global-error.tsx`) and from createRoot's onUncaughtError / onRecoverableError hooks
  - Refactored `nextjsPlugin` to use `addError` instead of `addEvent`, aligning with how reactPlugin works
  - Exported `addNextjsError` from the package entry point

  **packages/core**
  - Added `'nextjs error'` as a valid `createHandlingStack` type

  **test/apps/nextjs**
  - Added App Router test pages: `/error-test`, `/error-test/server-error`, `/global-error-test`, and `global-error.tsx` to exercise client, server digest, and global error scenarios

  **test/e2e**
  - Added E2E scenarios covering:
    - Client-side error reported via `addNextjsError`
    - Server error with digest reported via `addNextjsError`
    - Global error via `global-error.tsx`
  - Error scenarios scoped to App Router only

## Test instructions

You can build the test app:
`yarn install && yarn build && yarn build:apps
  cd test/apps/nextjs/
  yarn dev
`

## Checklist

- [X] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-nextjs/src/domain/error/addNextjsError.ts
+++ b/packages/rum-nextjs/src/domain/error/addNextjsError.ts
@@ -3,6 +3,25 @@ import type { Context } from '@datadog/browser-core'
 import type { ErrorInfo } from 'react'
 import { onRumStart } from '../nextjsPlugin'
 
+/**
+ * Add a Next.js error to the RUM session.
+ *
+ * @category Error
+ * @example
+ * ```ts
+ * // app/error.tsx (or app/global-error.tsx)
+ * 'use client'
+ * import { useEffect } from 'react'
+ * import { addNextjsError } from '@datadog/browser-rum-nextjs'
+ *
+ * export default function Error({ error }: { error: Error & { digest?: string } }) {
+ *   useEffect(() => {
+ *     addNextjsError(error)
+ *   }, [error])
+ *   return <div>Something went wrong</div>
+ * }
+ * ```
+ */
 export function addNextjsError(error: Error & { digest?: string }, errorInfo?: ErrorInfo) {
   const handlingStack = createHandlingStack('nextjs error')
   const startClocks = clocksNow()


test/apps/nextjs/app/error-test/clientErrorThrower.tsx
amortemousque, line 1: **🥜 nitpick:** ‏Do we need to isolate clientErrorThrower?
It’s only used in page.tsx. Keeping the code directly in page.tsx would simplify the structure.

test/e2e/scenario/nextjsPlugin.scenario.ts
```diff
@@ -197,3 +198,71 @@ test.describe('nextjs - router', () => {
...
+      .run(async ({ page, flushEvents, intakeRegistry, withBrowserLogs }) => {
+        await page.click('text=Go to Error Test')
+        await page.waitForURL(`**${viewPrefix}/error-test`)
+
+        await page.click('[data-testid="trigger-error"]')
+        await page.waitForSelector('[data-testid="error-boundary"]')
+
+        await flushEvents()
+
+        // React StrictMode double-fires useEffect in dev mode, so we may get 2 errors
```
amortemousque, line 218: **❓ question:** ‏Any reason to use dev mode?

packages/rum-nextjs/src/domain/nextjsPlugin.ts
```diff
@@ -24,11 +24,11 @@ export function nextjsPlugin(): NextjsPlugin {
...
     },
-    onRumStart({ addEvent }) {
-      globalAddEvent = addEvent
-      if (addEvent) {
-        for (const subscriber of onRumStartSubscribers) {
-          subscriber(addEvent)
+    onRumStart({ addError }) {
+      globalAddError = addError
+      for (const subscriber of onRumStartSubscribers) {
+        if (addError) {
```
amortemousque: **💬 suggestion:** ‏Put this check outside of the loop

packages/rum-nextjs/src/domain/error/addNextjsError.ts
```diff
@@ -0,0 +1,24 @@
...
+  onRumStart((addError) => {
+    callMonitored(() => {
+      addError({
+        error,
+        handlingStack,
+        componentStack: errorInfo?.componentStack ?? undefined,
+        startClocks,
+        context: {
+          ...(error as Error & { dd_context?: Context }).dd_context,
+          ...(error.digest && { nextjs: { digest: error.digest } }),
```
amortemousque, line 37: **❓ question:** ‏Should `nextjs.digest` part of the `rum-evens-format`

test/e2e/scenario/nextjsPlugin.scenario.ts
BeltranBulbarellaDD: In local we use dev and in the CI we use the prod one. I thought it would cover all cases but we could just do prod. WDYT?

amortemousque: Shouldn’t this double fire also happen with the React plugin? How do we handle it there?

BeltranBulbarellaDD: Yeah we do the same "optimistic check" this is what React does:
`toBeGreaterThan(0)`

packages/rum-nextjs/src/domain/error/addNextjsError.ts
```diff
@@ -0,0 +1,24 @@
+import { callMonitored, clocksNow, createHandlingStack } from '@datadog/browser-core'
+import type { Context } from '@datadog/browser-core'
+import type { ErrorInfo } from 'react'
+import { onRumStart } from '../nextjsPlugin'
+
+export function addNextjsError(error: Error & { digest?: string }, errorInfo?: ErrorInfo) {
```
mormubis, line 25: `addReactError` has a JSDoc with a usage example. What do you think about adding one here too? It's a public API and it would help people know where to call it from (`useEffect` in `error.tsx`, `createRoot` hooks, etc.).

packages/rum-nextjs/src/domain/error/addNextjsError.ts
```diff
@@ -0,0 +1,24 @@
...
+  const handlingStack = createHandlingStack('nextjs error')
+  const startClocks = clocksNow()
+  onRumStart((addError) => {
+    callMonitored(() => {
+      addError({
+        error,
+        handlingStack,
```
mormubis, line 32: The `?? undefined` only matters if `componentStack` can be `null`. React typings do allow it, but is that actually possible here?

BeltranBulbarellaDD: I think so, typescript complains because of:
`interface ErrorInfo {
        /**
         * Captures which component contained the exception, and its ancestors.
         * 
         */
        componentStack?: string | null;
    }
  `

test/e2e/scenario/nextjsPlugin.scenario.ts
amortemousque: Okay so you can keep it like this :)

PR #4346, opened 2026-03-17
👷 NextJS- Make NextJS plugin private

BeltranBulbarellaDD:
## Motivation

Like [PR](https://github.com/DataDog/browser-sdk/pull/4350), makes package private so as to not release it.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-nextjs/package.json
+++ b/packages/rum-nextjs/package.json
@@ -10,8 +10,8 @@
     "prepack": "npm run build"
   },
   "dependencies": {
-    "@datadog/browser-core": "*",
-    "@datadog/browser-rum-core": "*"
+    "@datadog/browser-core": "6.31.0",
+    "@datadog/browser-rum-core": "6.31.0"
   },
   "peerDependencies": {
     "next": ">=13.0.0",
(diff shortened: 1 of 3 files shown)


packages/rum-nextjs/package.json
```diff
@@ -10,8 +10,8 @@
     "prepack": "npm run build"
   },
   "dependencies": {
-    "@datadog/browser-core": "6.31.0",
-    "@datadog/browser-rum-core": "6.31.0"
+    "@datadog/browser-core": "*",
+    "@datadog/browser-rum-core": "*"
```
amortemousque: **❓ question:** ‏Why this change?

BeltranBulbarellaDD: IMO Simplicity. We don't need a version since [checkPackageDependencyVersions](https://github.com/DataDog/browser-sdk/blob/d307cf0b06e91d2cba584f5aa5508506ac85eb9f/scripts/lib/checkBrowserSdkPackageJsonFiles.ts#L20) will early return on private packages, so we just resolve to the workspace version.

PR #4348, opened 2026-03-17
♻️ Rewrite SessionManager with reactive strategy interface

thomas-lebeau:
## Motivation

The existing `SessionManager` and `SessionStore` architecture has accumulated complexity through layered abstractions (`sessionStore.ts`, `sessionStoreOperations.ts`, product-specific wrappers). This makes it difficult to adopt modern browser APIs like the CookieStore API and Web Locks, and the synchronous mutex-based approach creates contention issues.

This PR rewrites the session management layer with a reactive strategy interface, preparing the foundation for async-first storage (CookieStore API) in v7.

## Changes

- **Redefine `SessionStoreStrategy` interface** — strategies now own their state and expose it reactively via an observable, replacing the synchronous read/write/mutex pattern
- **Rewrite all three storage strategies** (cookie, localStorage, memory) to implement the new reactive interface
- **Cookie strategy** uses Web Locks for cross-tab coordination and `cookieObservable` for change detection, replacing the polling-based mutex
- **Introduce `CookieAccess` abstraction** to decouple cookie read/write from the `document.cookie` API, preparing for CookieStore API adoption
- **Merge `SessionStore` into `SessionManager`** — remove `sessionStoreOperations.ts` and consolidate session lifecycle logic; extract strategy selection into `sessionStore.ts`
- **Make `SessionManager` initialization fully async** — `startSessionManager` accepts an `onReady` callback, with `setSessionState` async across all strategies
- **Remove product-specific session manager wrappers** (`rumSessionManager.ts`, `logsSessionManager.ts`) — the generic `SessionManager` now handles all products
- **Extract `startTelemetrySessionContext`** — defer telemetry configuration until the session is ready
- **Move `cookieObservable` from core to rum-core** — scoped to where it's actually used
- **Adapt rumPublicApi tests** to the new async session manager (buffered API calls, clock-based assertions)

## Test instructions

```bash
yarn test:unit --spec packages/core/src/domain/session/sessionManager.spec.ts
yarn test:unit --spec packages/core/src/domain/session/sessionStore.spec.ts
yarn test:unit --spec packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
yarn test:unit --spec packages/core/src/domain/session/storeStrategies/sessionInLocalStorage.spec.ts
yarn test:unit --spec packages/core/src/domain/session/storeStrategies/sessionInMemory.spec.ts
yarn test:unit --spec packages/rum-core/src/boot/rumPublicApi.spec.ts
yarn test:unit --spec packages/rum-core/src/boot/preStartRum.spec.ts
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/browser/cookieAccess.spec.ts
+++ b/packages/core/src/browser/cookieAccess.spec.ts
@@ -70,6 +70,8 @@ describe('cookieAccess', () => {
           clock,
           async flushObservable(this: void, spy: jasmine.Spy) {
             await collectAsyncCalls(spy, 1)
+            // Reset the spy calls to avoid throwing on unexpected calls during teardown
+            registerCleanupTask(() => spy.calls.reset())
           },
           async setCookieWithCleanup(
             this: void,
--- a/packages/core/test/collectAsyncCalls.ts
+++ b/packages/core/test/collectAsyncCalls.ts
@@ -1,5 +1,4 @@
 import { getCurrentJasmineSpec } from './getCurrentJasmineSpec'
-import { registerCleanupTask } from './registerCleanupTask'
 
 const originalPlanForGuard = new WeakMap<() => void, () => void>()
 
@@ -34,10 +33,6 @@ export function collectAsyncCalls<F extends jasmine.Func>(
     }) as F
     originalPlanForGuard.set(guard, originalPlan)
     spy.and.callFake(guard)
-    registerCleanupTask(() => {
-      spy.and.callFake(originalPlan)
-      originalPlanForGuard.delete(guard)
-    })
   })
 }
 


packages/core/src/browser/cookieAccess.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+  const original = Object.getOwnPropertyDescriptor(window, 'cookieStore')
+  Object.defineProperty(window, 'cookieStore', { get: () => mockStore, configurable: true })
+  registerCleanupTask(() => {
+    if (original) {
+      Object.defineProperty(window, 'cookieStore', original)
+    } else {
+      delete (window as CookieStoreWindow).cookieStore
+    }
+  })
+}
```
BenoitZugmeyer: **nitpick:** might be simpler to use `mockable` (`mockable(window.cookieStore)` // `replaceMockable(window.cookieStore, undefined)`

packages/core/src/browser/cookieAccess.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+import { deleteCookie, getCookie, setCookie } from './cookie'
+import type { CookieStoreWindow } from './browser.types'
+import { createCookieAccess, WATCH_COOKIE_INTERVAL_DELAY } from './cookieAccess'
+
+const COOKIE_NAME = 'test_cookie'
+const COOKIE_OPTIONS = { secure: false, crossSite: false, partitioned: false }
+const MOCK_CONFIGURATION = { allowUntrustedEvents: true } as Configuration
+
+describe('cookieAccess', () => {
+  describe('document.cookie fallback', () => {
```
BenoitZugmeyer: **thought:** I think we could have common tests for both implementations. Also, we could use the actual cookie store instead of a mock

```ts
const setups = [
  { title: 'document.cookie fallback', setup() { disableCookieStore() } },
  { title: 'cookieStore', setup() { if (!window.cookieStore) pending("No cookie store available") }
]

for (const { title, setup } of setups) {
  describe(title, () => {
    beforeEach(() => setup())

    it('should pas the current values to callback', () => {
      setCookie(...)
  
      const cookieAccess = createCookieAccess(COOKIE_NAME, MOCK_CONFIGURATION, COOKIE_OPTIONS)

      let capturedValues: string[] | undefined
      await cookieAccess.getAllAndSet((values) => {
        capturedValues = values
        return { value: 'new', expireDelay: 1000 }
      })

      expect(capturedValues).toEqual(['value1'])
...
```

packages/core/src/domain/session/storeStrategies/sessionInCookie.spec.ts
```diff
@@ -1,62 +1,239 @@
...
 import { SESSION_COOKIE_EXPIRATION_DELAY, SESSION_EXPIRATION_DELAY, SESSION_TIME_OUT_DELAY } from '../sessionConstants'
 import { buildCookieOptions, selectCookieStrategy, initCookieStrategy } from './sessionInCookie'
 import { SESSION_STORE_KEY } from './sessionStoreStrategy'
 
 const DEFAULT_INIT_CONFIGURATION = { clientToken: 'abc', trackAnonymousUser: true }
 
+/**
+ * Disable the CookieStore API so the cookie observable falls back to polling.
+ * This makes the tests deterministic since we can control time with mockClock.
+ */
```
BenoitZugmeyer: **suggestion:** mock the `cookieAccess`, so you can inject your own "cookieAccess" object, without actually relying on cookies.

```ts
const cookieAccess = mockable(createCookieAccess)(SESSION_STORE_KEY, configuration, cookieOptions)
```

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -23,74 +24,85 @@ export function selectCookieStrategy(initConfiguration: InitConfiguration): Sess
...
-      storeSessionCookie(
-        cookieOptions,
-        configuration,
-        getExpiredSessionState(sessionState, configuration),
-        SESSION_TIME_OUT_DELAY
-      ),
-  }
+// Promise chain serializes calls when Web Locks are unavailable
+// eslint-disable-next-line local-rules/disallow-side-effects
+let pendingChain = Promise.resolve()
```
BenoitZugmeyer: **suggestion:** lazily create the `Promise.resolve()`

```ts
let pendingChain: Promise<void> | undefined
```

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -23,74 +24,85 @@ export function selectCookieStrategy(initConfiguration: InitConfiguration): Sess
...
+      } else {
+        pendingChain = pendingChain.then(() => applyAndWrite(fn))
+        await pendingChain
+      }
+    },
+    sessionObservable,
+  }
+}
 
-  // reverse the cookies so that if there is no match, the cookie returned is the first one
```
BenoitZugmeyer, line 79: **nitpick:** I found that comment useful to justify the "reverse"

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -23,74 +24,85 @@ export function selectCookieStrategy(initConfiguration: InitConfiguration): Sess
...
+function computeExpireDelay(trackAnonymousUser: boolean, sessionState: SessionState): number {
+  // Cookie expiration logic:
+  // - trackAnonymousUser=true: always use 1 year (keep cookie alive for device ID)
+  // - trackAnonymousUser=false + active session: 15 min (activity-based renewal)
+  // - trackAnonymousUser=false + expired session: 4h (absolute timeout)
+  return trackAnonymousUser
+    ? SESSION_COOKIE_EXPIRATION_DELAY
+    : sessionState.isExpired
+      ? SESSION_TIME_OUT_DELAY
+      : SESSION_EXPIRATION_DELAY
```
BenoitZugmeyer: **question:** Do you think the `isExpired` distinction actually needed? I think it was needed in the past because we didn't store the expiration date in the cookie, but we've been doing it for a long time now. Keeping 4 hours cookies would make things simpler.

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -23,74 +24,85 @@ export function selectCookieStrategy(initConfiguration: InitConfiguration): Sess
...
+}
+
+function buildSessionString(sessionState: SessionState, cookieOptions: CookieOptions): string {
+  const sessionStateWithOptions = {
+    ...sessionState,
+    ...(!isEmptyObject(sessionState) ? { c: encodeCookieOptions(cookieOptions) } : {}),
+  }
+  return toSessionString(sessionStateWithOptions)
```
BenoitZugmeyer: **nitpick:** this might be a bit more simple:
```ts
function buildSessionString(sessionState: SessionState, cookieOptions: CookieOptions): string {
  return toSessionString(isEmptyObject(sessionState) ? sessionState : { ...sessionState, c: : encodeCookieOptions(cookieOptions) })
```

packages/core/src/domain/session/storeStrategies/sessionInCookie.ts
```diff
@@ -115,10 +127,10 @@ function encodeCookieOptions(cookieOptions: CookieOptions): string {
 
   /* eslint-disable no-bitwise */
   let byte = 0
-  byte |= SESSION_COOKIE_VERSION << 5 // Store version in upper 3 bits
-  byte |= domainCount << 1 // Store domain count in next 4 bits
-  byte |= cookieOptions.crossSite ? 1 : 0 // Store useCrossSiteScripting in next bit
```
BenoitZugmeyer, line 120: **nitpick:** I liked those comments

packages/core/src/domain/session/storeStrategies/sessionInLocalStorage.ts
```diff
@@ -21,24 +21,29 @@ export function selectLocalStorageStrategy(): SessionStoreStrategyType | undefin
...
+      if (event.key === SESSION_STORE_KEY && event.storageArea === localStorage) {
+        observable.notify(toSessionState(event.newValue))
+      }
+    }
+    // eslint-disable-next-line local-rules/disallow-zone-js-patched-values
+    window.addEventListener('storage', listener)
+    return () => {
+      // eslint-disable-next-line local-rules/disallow-zone-js-patched-values
+      window.removeEventListener('storage', listener)
+    }
```
BenoitZugmeyer: **suggestion:** don't disable lint rules, use our `addEventListener`

```ts
return addEventListener(window, 'storage', (event) => {
  ...
}).stop
```

packages/core/src/domain/session/storeStrategies/sessionInLocalStorage.ts
```diff
@@ -21,24 +21,29 @@ export function selectLocalStorageStrategy(): SessionStoreStrategyType | undefin
...
-    isLockEnabled: false,
-    persistSession: persistInLocalStorage,
-    retrieveSession: retrieveSessionFromLocalStorage,
-    expireSession: (sessionState: SessionState) => expireSessionFromLocalStorage(sessionState, configuration),
+    setSessionState(fn: (sessionState: SessionState) => SessionState): Promise<void> {
+      const currentState = toSessionState(localStorage.getItem(SESSION_STORE_KEY))
+      const newState = fn(currentState)
+      localStorage.setItem(SESSION_STORE_KEY, toSessionString(newState))
+      sessionObservable.notify(newState)
+      return Promise.resolve()
```
BenoitZugmeyer: **suggestion:** move the notify() after resolve() like [we do for cookies](https://github.com/DataDog/browser-sdk/pull/4348/changes#diff-ce896138afb9ef512234f4ba3ed7200ce4fde99aac39badccdb6945cb84ce6b5R103)?
(same suggestion for `sessionInMemory` and `fakeSessionStoreStrategy`)

packages/core/src/domain/session/storeStrategies/sessionInMemory.ts
```diff
@@ -1,47 +1,41 @@
...
-import { getExpiredSessionState } from '../sessionState'
 import { shallowClone } from '../../../tools/utils/objectUtils'
 import type { SessionStoreStrategy, SessionStoreStrategyType } from './sessionStoreStrategy'
 
-/**
- * Key used to store session state in the global object.
- * This allows RUM and Logs SDKs to share the same session when using memory storage.
- */
 export const MEMORY_SESSION_STORE_KEY = '_DD_SESSION'
+const MEMORY_SESSION_OBSERVABLE_KEY = '_DD_SESSION_OBSERVABLE'
```
BenoitZugmeyer: **suggestion:** 
* use a single global variable (avoid polluting global scope for nothing)
* expose just an 'onChange' callback instead of an Observable. Exposing an observable is problematic if we'll want to change anything on it someday and multiple versions of the SDK are used on the same page.

```
_DD_SESSION= { id, onChange }
```

packages/core/src/domain/session/storeStrategies/sessionInMemory.ts
```diff
@@ -1,47 +1,41 @@
...
-function expireSessionFromMemory(previousSessionState: SessionState, configuration: Configuration) {
-  persistInMemory(getExpiredSessionState(previousSessionState, configuration))
+  return {
+    setSessionState(fn: (sessionState: SessionState) => SessionState): Promise<void> {
+      const currentState = globalObject[MEMORY_SESSION_STORE_KEY]
+        ? shallowClone(globalObject[MEMORY_SESSION_STORE_KEY])
+        : {}
+      const newState = fn(currentState)
+      globalObject[MEMORY_SESSION_STORE_KEY] = shallowClone(newState)
+      sessionObservable.notify(shallowClone(newState))
```
BenoitZugmeyer: **nitpick:** maybe we don't need those two `shallowClone`

packages/core/src/domain/session/sessionManager.spec.ts
```diff
@@ -1,81 +1,119 @@
...
+    // Trigger user activity to create an actual session (fresh init starts as expired)
+    document.dispatchEvent(createNewEvent(DOM_EVENT.CLICK))
+
+    return sessionManager
+  }
+
+  /**
+   * Same as startSessionManagerWithDefaults but skips the automatic activity trigger.
+   * Used for tests that need to observe the raw initialization behavior.
+   */
```
BenoitZugmeyer: **suggestion:** maybe just introduce an option to `startSessionManagerWithDefaults` instead of introducing a second function

packages/core/src/tools/observable.ts
```diff
@@ -52,8 +52,11 @@ export function mergeObservables<T>(...observables: Array<Observable<T>>) {
...
   private buffer: T[] = []
 
-  constructor(private maxBufferSize: number) {
-    super()
+  constructor(
+    private maxBufferSize: number,
+    onFirstSubscribe?: (observable: Observable<T>) => (() => void) | void
+  ) {
+    super(onFirstSubscribe)
   }
```
BenoitZugmeyer, line 60: **thought:** I'm not sure about this. Not having the `onFirstSubscribe` on buffered observable was a design decision, because it makes less sense to have lazy observable when they are buffered. I understand that you use it for xhr/fetch/console early data collection but it feels like a hack: we don't actually want lazy initialization, since we immediately `.subscribe(noop)`. Maybe we can clean things up if/when we more things to `bufferedData.ts`.

packages/core/test/collectAsyncCalls.ts
```diff
@@ -22,7 +28,24 @@ export function collectAsyncCalls<F extends jasmine.Func>(
...
+          checkCalls()
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+          return result
+        } catch (e) {
+          checkCalls()
+          throw e
+        }
+      }
       checkCalls()
     }) as F)
```
BenoitZugmeyer: **suggestion:** This is quite complex. Maybe simplify a bit? Do we need the  `EXTRA_CALL_GUARD` / `shouldCallPreviousPlan` logic at all? Can't we just call  run the previous plan unconditionally?

```ts

    const previousPlan: F = (spy.and as unknown as { plan: F }).plan

    spy.and.callFake(((...args: Parameters<F>) => {
      checkCalls()
      return previousPlan(...args)
    }) as F)
```

packages/logs/src/boot/preStartLogs.ts
```diff
@@ -99,11 +100,12 @@ export function createPreStartStrategy(
...
-      // Instrument fetch to track network requests
-      // This is needed in case the consent is not granted and some customer
-      // library (Apollo Client) is storing uninstrumented fetch to be used later
-      // The subscrption is needed so that the instrumentation process is completed
+      // Instrument fetch and console early so events fired synchronously after
+      // init() are captured and buffered for replay when startLogs() subscribes.
       initFetchObservable().subscribe(noop)
+      if (configuration.forwardConsoleLogs.length) {
+        initConsoleObservable(configuration.forwardConsoleLogs).subscribe(noop)
+      }
```
BenoitZugmeyer, line 108: **thought:** it could be interesting to do this the same way we do for runtime errors, in `packages/core/src/domain/bufferedData.ts`.

Advantages are:
* we collect things even earlier, at module evaluation
* we have a proper "unbuffer" strategy to avoid leaking memory

Note that it could be done in a separate PR

packages/logs/src/boot/startLogs.spec.ts
```diff
@@ -175,7 +170,7 @@ describe('logs', () => {
       handleLog({ status: StatusType.info, message: 'message 1' }, logger)
 
       sessionManager.expire()
-      clock.tick(STORAGE_POLL_DELAY * 2)
+      clock.tick(ONE_SECOND * 2)
```
BenoitZugmeyer: **suggestion:** use a more explicit delay? What are those 2 seconds?

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -750,67 +761,55 @@ describe('rum public api', () => {
   })
 
   describe('startDurationVital', () => {
-    it('should call startDurationVital on the startRum result', async () => {
```
BenoitZugmeyer, line 753: **question:** Why those tests are removed?

packages/rum-core/src/boot/rumPublicApi.spec.ts
```diff
@@ -554,34 +556,38 @@ describe('rum public api', () => {
...
-      ;({ rumPublicApi } = makeRumPublicApiWithDefaults({
+      ;({ rumPublicApi, startRumSpy } = makeRumPublicApiWithDefaults({
         startRumResult: {
           addTiming: addTimingSpy,
         },
       }))
     })
 
     it('should add custom timings', async () => {
+      const clock = mockClock()
```
BenoitZugmeyer, line 572: **question:** why do we need this mockClock?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
+  const { throttled: throttledExpandOrRenew, cancel: cancelExpandOrRenew } = throttle(() => {
+    strategy.setSessionState((state) => expandOrRenew(state, configuration)).catch(monitorError)
+  }, ONE_SECOND)
+  stopCallbacks.push(cancelExpandOrRenew)
+
+  let stopped = false
+  stopCallbacks.push(() => {
+    stopped = true
+  })
+  ;(async () => {
```
BenoitZugmeyer, line 75: **suggestion:** would it make sense to turn the whole `startSessionManager` async and remove the `onReady` callback?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
-  sessionStore.expandOrRenewSession(() => {
-    const hasConsent = trackingConsentState.isGranted()
-    if (!hasConsent) {
-      sessionStore.expire(hasConsent)
+  let previousState: SessionState = {}
+
+  const { throttled: throttledExpandOrRenew, cancel: cancelExpandOrRenew } = throttle(() => {
+    strategy.setSessionState((state) => expandOrRenew(state, configuration)).catch(monitorError)
+  }, ONE_SECOND)
+  stopCallbacks.push(cancelExpandOrRenew)
```
BenoitZugmeyer, line 69: **suggestion:** since this is only used in `setupSessionTracking`, let's move it there

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
 const SESSION_CONTEXT_TIMEOUT_DELAY = SESSION_TIME_OUT_DELAY
 let stopCallbacks: Array<() => void> = []
 
+// Session state helper functions
```
BenoitZugmeyer: **suggestion:** It looks like those helper function could sit in `sessionState.ts` where a few similar functions are there

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
+      return state
     })
-    sessionStore.sessionStateUpdateObservable.subscribe(({ newState }) => {
-      // mutate the session state in the history
-      const currentContext = sessionContextHistory.find()
-      if (currentContext) {
-        currentContext.isReplayForced = !!newState.forcedReplay
+    if (isSessionInExpiredState(state)) {
+      state = getExpiredSessionState(state, configuration)
+    }
```
BenoitZugmeyer: **suggestion:** FMU we can't be in expired state here since we are doing "expandOrRenew"

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
+
+  function subscribeToSessionChanges(initialState: SessionState) {
+    previousState = initialState
+    sessionContextHistory.add(buildSessionContext(initialState), clocksOrigin().relative)
+
+    scheduleExpirationTimeout(initialState)
+
+    const subscription = strategy.sessionObservable.subscribe((newState) => {
+      if (isSessionInExpiredState(newState)) {
+        newState = getExpiredSessionState(newState, configuration)
```
BenoitZugmeyer: **question:** this is quite confusing. Why do we do this?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
       }
+      scheduleExpirationTimeout(newState)
+      handleStateChange(newState)
+      previousState = newState
```
BenoitZugmeyer: **suggestion:** move `previousState = newState` in `handleStateChange`

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
-      if (currentContext) {
-        currentContext.isReplayForced = !!newState.forcedReplay
+    if (isSessionInExpiredState(state)) {
+      state = getExpiredSessionState(state, configuration)
+    }
+    return state
+  }
+
+  function subscribeToSessionChanges(initialState: SessionState) {
+    previousState = initialState
```
BenoitZugmeyer: **suggestion:** instead of having a `previousState` variable, could we use `sessionContextHistory.find()` to get the latest known session?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -134,18 +211,62 @@ export function startSessionManager(
...
 
-    if (!session?.id) {
-      reportUnexpectedSessionState(configuration).catch(() => void 0) // Ignore errors
+  function expire() {
+    cancelExpandOrRenew()
+    // Update in-memory state synchronously so events stop being collected immediately
+    const expiredState = getExpiredSessionState(previousState, configuration)
+    if (!trackingConsentState.isGranted()) {
+      delete expiredState.anonymousId
+    }
```
BenoitZugmeyer, line 199: **thought:** the `anonymousId` logic is all over the place... maybe we can do something about it

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -48,80 +43,162 @@ export const VISIBILITY_CHECK_DELAY = ONE_MINUTE
...
+  }
+
+  delete state.isExpired
+  expandSessionState(state)
+
+  return state
+}
+
+function expandOnly(state: SessionState): SessionState {
+  if (isSessionInExpiredState(state) || isSessionInNotStartedState(state) || !state.id) {
```
BenoitZugmeyer: **thought:** A sanitized / strongly typed `SessionState` would help clarify the flow. We shouldn't have to check for various properties to understand the state we're in.

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -34,7 +30,6 @@ export interface SessionManager {
   expireObservable: Observable<void>
   sessionStateUpdateObservable: Observable<{ previousState: SessionState; newState: SessionState }>
```
BenoitZugmeyer: **suggestion:** remove `sessionStateUpdateObservable` as it doesn't seem used anymore

packages/core/src/domain/session/sessionManager.ts
thomas-lebeau: I think yes, I plan to have a look as a separate PR

packages/rum-core/src/boot/rumPublicApi.spec.ts
thomas-lebeau: `startDurationVital` and `stopDurationVital` are not buffered. instead they call addAction when used before the session Manager is initialized.

Maybe We should have two test, for calling it before the SM is started and after.

packages/rum-core/src/boot/rumPublicApi.spec.ts
thomas-lebeau: because `rumPublicApi.addTiming` is now buffered (SM is not ready yet). so the call to addTiming that occur after the SM is ready contain a timing calculated when buffered (`arg[1]`) instead of `undefined` that would let set the timing at the time the SM is ready.

we need to mock the clock so we can predict the timing `getTimeStamp(clock.relative(0))`

packages/core/src/domain/session/sessionManager.ts
thomas-lebeau: Hmm, `cancelExpandOrRenew` is also used in expire(), so it's not only used in `setupSessionTracking`.

packages/rum-core/src/boot/rumPublicApi.spec.ts
BenoitZugmeyer: Aah I see. It's coming soon https://github.com/DataDog/browser-sdk/pull/4403 :)

packages/core/src/domain/session/sessionManager.ts
thomas-lebeau: This is not outdated, this function has been moved to `sessionState.ts`

packages/core/src/domain/session/storeStrategies/sessionInLocalStorage.ts
```diff
@@ -22,23 +24,24 @@ export function selectLocalStorageStrategy(): SessionStoreStrategyType | undefin
...
-    persistSession: persistInLocalStorage,
-    retrieveSession: retrieveSessionFromLocalStorage,
-    expireSession: (sessionState: SessionState) => expireSessionFromLocalStorage(sessionState, configuration),
+    setSessionState(fn: (sessionState: SessionState) => SessionState): Promise<void> {
+      const currentState = toSessionState(localStorage.getItem(SESSION_STORE_KEY))
+      const newState = fn(currentState)
+      localStorage.setItem(SESSION_STORE_KEY, toSessionString(newState))
+      const result = Promise.resolve()
+      sessionObservable.notify(newState)
+      return result
```
BenoitZugmeyer: **suggestion:** 

```suggestion
      await Promise.resolve()
      sessionObservable.notify(newState)
```

Or maybe remove the asynchronicity altogether, as it might not be that useful.

(same for `sessionInMemory` and `fakeSessionStoreStrategy`)

packages/core/src/browser/cookieAccess.spec.ts
```diff
@@ -250,25 +250,9 @@ describe('cookieAccess', () => {
...
-  const original = Object.getOwnPropertyDescriptor(window, 'cookieStore')
-  Object.defineProperty(window, 'cookieStore', { get: () => undefined, configurable: true })
-  registerCleanupTask(() => {
-    if (original) {
-      Object.defineProperty(window, 'cookieStore', original)
-    } else {
-      delete (window as CookieStoreWindow).cookieStore
-    }
-  })
+  replaceMockable(getCookieStore, () => undefined)
```
BenoitZugmeyer: **nitpick:** can't you do `mockable(window.cookieStore)` directly? Do you need to use a getter?

PR #4352, opened 2026-03-17
⚗ NextJS- add nextjs error boundary component

BeltranBulbarellaDD:
## Motivation

Adds React error boundary to NextJS pages router. 

## Changes

Adds `ErrorBoundary` — a React error boundary component that wraps `createErrorBoundary` from `@datadog/browser-rum-react`, pre-wired to call `addNextjsError`.

Exports `ErrorBoundaryFallback` and `ErrorBoundaryProps` types alongside it.

Also adds `@datadog/browser-rum-react` as an optional peer dependency of `rum-nextjs`, and allows `@datadog/browser-rum-react/internal` through the `disallow-side-effects` ESLint rule.

## Test instructions

Test app — new pages for both routers to exercise all error paths:

| Scenario | Implementation |
|--------|--------|
| Pages Router client error | ErrorBoundary in `pages-router/error-test.tsx` |
| App Router client error | `error.tsx` segment boundary calling `addNextjsError` |
| App Router server error (with digest) | Server component throw → `error.tsx` calling `addNextjsError`  |
| App Router global error | Root-level throw → `global-error.tsx` calling `addNextjsError` | 

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/apps/angular-app/yarn.lock
+++ b/test/apps/angular-app/yarn.lock
@@ -296,14 +296,14 @@ __metadata:
 
 "@datadog/browser-core@file:../../../packages/core/package.tgz::locator=angular-app%40workspace%3A.":
   version: 6.31.0
-  resolution: "@datadog/browser-core@file:../../../packages/core/package.tgz#../../../packages/core/package.tgz::hash=d110af&locator=angular-app%40workspace%3A."
-  checksum: 10c0/1e3797a56310fde499578396bd2594c4e0ac891610e7d57ddac2164a8c29a97177b3eac938f6a8480192bb31b5d798acaf2ea26de1778047d43c010ff0124dc0
+  resolution: "@datadog/browser-core@file:../../../packages/core/package.tgz#../../../packages/core/package.tgz::hash=6201e3&locator=angular-app%40workspace%3A."
+  checksum: 10c0/a019713374971a2dd35c63318cecd76e38d08348a24ca5b88abb8fca8501d17589e187bd4e68c7358f3523dd3fd08d110b307647854a620b0f3371f2c7db14ca
   languageName: node
   linkType: hard
 
 "@datadog/browser-rum-angular@file:../../../packages/rum-angular/package.tgz::locator=angular-app%40workspace%3A.":
   version: 0.0.0
-  resolution: "@datadog/browser-rum-angular@file:../../../packages/rum-angular/package.tgz#../../../packages/rum-angular/package.tgz::hash=1db391&locator=angular-app%40workspace%3A."
+  resolution: "@datadog/browser-rum-angular@file:../../../packages/rum-angular/package.tgz#../../../packages/rum-angular/package.tgz::hash=004229&locator=angular-app%40workspace%3A."
   dependencies:
     "@datadog/browser-core": "npm:6.31.0"
     "@datadog/browser-rum-core": "npm:6.31.0"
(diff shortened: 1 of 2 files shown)


packages/rum-nextjs/src/domain/error/errorBoundary.spec.tsx
BeltranBulbarellaDD, line 1: Since all these tests are the same for rum-react (except for the last one) should I remove them?

package.json
```diff
@@ -17,12 +17,11 @@
     "build:apps": "node scripts/build/build-test-apps.ts",
     "build:docs:json": "typedoc --logLevel Verbose --json ./docs.json",
     "build:docs:html": "typedoc --out ./docs",
-    "pack": "yarn workspaces foreach --all --parallel --include '@datadog/*' exec yarn pack",
+    "pack": "yarn workspaces foreach --all --parallel --topological --include '@datadog/*' exec yarn pack",
```
BeltranBulbarellaDD, line 20: This is required because rum-nextjs needs rum-react

test/e2e/playwright.base.config.ts
```diff
@@ -48,7 +48,7 @@ export const config: Config = {
       name: 'nextjs app router',
       stdout: 'pipe' as const,
       cwd: path.join(__dirname, '../apps/nextjs'),
-      command: isLocal ? 'yarn dev' : 'yarn start',
+      command: 'yarn start',
```
BeltranBulbarellaDD, line 51: This is for 2 reasons: 
1. So that StrictMode does not double-fires in dev mode.
2. The dev error overlay does not appear ([ex](https://github.com/vercel/next.js/issues/47124))

package.json
BenoitZugmeyer: I'm not sure about this... `rum` depends on `rum-code` which depends on `core` and that was not an issue before.

test/e2e/scenario/nextjsPlugin.scenario.ts
```diff
@@ -200,69 +201,78 @@ test.describe('nextjs - router', () => {
...
-          expect(browserLogs.length).toBeGreaterThan(0)
-        })
-      })
+            const customErrors = intakeRegistry.rumErrorEvents.filter((e) => e.error.source === 'custom')
+            expect(customErrors).toHaveLength(1)
+            expect(customErrors[0].error.handling_stack).toBeDefined()
+            expect(customErrors[0].context).toMatchObject({
+              framework: 'nextjs',
+              nextjs: { digest: expect.any(String) },
+            })
```
BenoitZugmeyer, line 249: **suggestion:** 
```ts
          expect(errorEvent.error.component_stack).toBeDefined()
```

packages/rum-react/internal.d.ts
BenoitZugmeyer, line 1: **suggestion:** this should not be needed. During development, [the path points to the correct source](https://github.com/DataDog/browser-sdk/blob/fcbf67aa4e2a17924cac39793adbf97cea6ace18/tsconfig.base.json#L33), and for production builds the [package.json is pointing to the right .d.ts file](https://github.com/DataDog/browser-sdk/blob/fcbf67aa4e2a17924cac39793adbf97cea6ace18/packages/rum/internal/package.json#L1)

packages/rum-nextjs/package.json
```diff
@@ -14,10 +14,14 @@
     "@datadog/browser-rum-core": "6.31.0"
   },
   "peerDependencies": {
+    "@datadog/browser-rum-react": "6.31.0",
```
BenoitZugmeyer: **suggestion:** `@datadog/browser-rum-react` should be a direct dependencies, not a peer dep or a dev dep. Customers should not need to install the react integration themselves

.claude/skills/datadog-sdk-event-inspection/SKILL.md
thomas-lebeau, line 1: sounds like this as nothing to with this PR

test/e2e/scenario/nextjsPlugin.scenario.ts
BeltranBulbarellaDD: But we are not testing the same thing right? I don't mind the change I just wanted to test the nextjs specific values.

.claude/skills/datadog-sdk-event-inspection/SKILL.md
BeltranBulbarellaDD: Yeah I'll remove it, it was used for the investigation on Firefox.

packages/rum-react/internal.d.ts
BeltranBulbarellaDD: Aa okay, thanks for the info!

package.json
BeltranBulbarellaDD: Hm let's discuss!

BeltranBulbarellaDD: Discussed [here](https://dd.slack.com/archives/C07D07KLS2E/p1773932458377529?thread_ts=1773860415.280269&cid=C07D07KLS2E)

BenoitZugmeyer: Heeey so after a bit of investigation I think the issue comes from the integration packages having a ["prepack" script](https://github.com/DataDog/browser-sdk/blob/7807ff3eb14418ff864bf8a992ba65a135fb8225/packages/rum-react/package.json#L10). This script is called during `yarn pack`. It builds the package before packing it. This was useful when dogfooding the `rum-react` package in web-ui before publishing it  (see https://github.com/DataDog/browser-sdk/pull/3182)

I *think* it makes sense and it should be added to all packages, not just integrations. By doing that, we won't need to run `yarn build` before running `yarn run pack`, the build will always be done during pack. As you noted, this requires to use the `--topological` flag when packing, similarly to the `build` script. I'll work on this in a separate PR. In the meantime, please add back the `--topological` option and remove the `internal.d.ts` file.

test/e2e/scenario/nextjsPlugin.scenario.ts
BenoitZugmeyer: Nevermind. It seems like Next.js doesn't expose the component stack for those `error.tsx` files

packages/rum-nextjs/src/domain/error/errorBoundary.spec.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+
+import { disableJasmineUncaughtExceptionTracking, ignoreConsoleLogs } from '@datadog/browser-core/test'
+import { appendComponent } from '../../../../rum-react/test/appendComponent'
+import { initReactOldBrowsersSupport } from '../../../../rum-react/test/reactOldBrowsersSupport'
```
mormubis, line 5: The test imports `appendComponent` and `initReactOldBrowsersSupport` via relative paths across package boundaries (`../../../../rum-react/test/...`). Is there a plan to extract these into a shared test utility, or is this the accepted pattern? It works, just looks fragile if packages ever move around.

BeltranBulbarellaDD: Agreed. Maybe we can expose them as shared utility, that would be better.

PR #4354, opened 2026-03-17
🔥 remove internalAnalyticsSubdomain and INTAKE_SITE_FED_STAGING

BenoitZugmeyer:
## Motivation

Two internal SDK behaviors — `internalAnalyticsSubdomain` init option and the `INTAKE_SITE_FED_STAGING` special-casing — were workarounds for routing data to Datadog-internal intake endpoints. These are now handled in web-ui using the standard `proxy` initialization parameter, making them unnecessary.

## Changes

- Remove `internalAnalyticsSubdomain` init option (modified intake host and reversed query parameter order for internal use)
- Remove `INTAKE_SITE_FED_STAGING` constant and its `http-intake.logs.dd0g-gov.com` special-casing in `buildEndpointHost`

## Test instructions

No user-visible behavior change. These were internal-only behaviors not exposed to customers.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -2,7 +2,7 @@ import type { Payload } from '../../transport'
 import { timeStampNow } from '../../tools/utils/timeUtils'
 import { normalizeUrl } from '../../tools/utils/urlPolyfill'
 import { generateUUID } from '../../tools/utils/stringUtils'
-import { INTAKE_SITE_FED_STAGING, INTAKE_SITE_US1 } from '../intakeSites'
+import { INTAKE_SITE_US1 } from '../intakeSites'
 import type { InitConfiguration } from './configuration'
 
 // replaced at build time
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/configuration/endpointBuilder.spec.ts
```diff
@@ -38,15 +38,6 @@ describe('endpointBuilder', () => {
...
     })
 
-    it('should not start with ddsource for internal analytics mode', () => {
-      const url = createEndpointBuilder({ ...initConfiguration, internalAnalyticsSubdomain: 'foo' }, 'rum').build(
-        'fetch',
-        DEFAULT_PAYLOAD
-      )
-      expect(url).not.toContain('/rum?ddsource')
-      expect(url).toContain('ddsource=browser')
-    })
```
thomas-lebeau, line 48: I'm not sure why we were doing this (found [RUMF-1464](https://datadoghq.atlassian.net/browse/RUMF-1464), but it's not very clear), but this does not seems to be replicated in web-ui.

Image

BenoitZugmeyer: Good catch. I'll replicate in web-ui

PR #4449, opened 2026-04-07
Add Live Debugger package

watson:
## Motivation

Introduce the `@datadog/browser-debugger` package to enable [Live Debugger](https://docs.datadoghq.com/dynamic_instrumentation/) in browser applications. This gives frontend developers the ability to add log probes to running applications, evaluate conditions, and inspect runtime state — all without redeploying or modifying source code.

> **Note:** This package is intended for internal Datadog use only until validated in production. It follows the same convention as `@datadog/browser-core`, `@datadog/browser-rum-core`, and `@datadog/browser-worker` — published to npm with each release (to keep versions in sync), but excluded from generated documentation via `typedoc.json`.

### New package: `packages/debugger`

A new `@datadog/browser-debugger` package that provides the full probe execution pipeline:

- **`domain/api.ts`** — Core instrumentation hooks (`onEntry`, `onReturn`, `onThrow`) that execute probes when instrumented functions are called, including condition evaluation, snapshot capture, template message rendering, and rate limiting.
- **`domain/activeEntries.ts`** — Tracks per-probe execution stacks for correlating entry/return/throw events, extracted to break the dependency cycle between `api.ts` and `probes.ts`.
- **`domain/probes.ts`** — Probe lifecycle management (add, remove, clear) with per-probe and global snapshot rate limiting. Compiles probe conditions and template segments on registration.
- **`domain/capture.ts`** — Deep value capture for arguments, locals, return values, and thrown errors with configurable reference depth, collection size limits, and string length limits.
- **`domain/expression.ts`** — Expression compiler that parses JSON expression trees (comparisons, logical operators, member access, string operations, etc.) into executable functions.
- **`domain/condition.ts`** — Probe condition evaluator that compiles and caches condition expressions.
- **`domain/template.ts`** — Template segment compiler and evaluator for rendering dynamic probe messages with runtime context.
- **`domain/stacktrace.ts`** — Stack trace capture and parsing from `Error` objects.
- **`domain/deliveryApi.ts`** — Polling-based probe delivery client that fetches probe updates/deletions from the Delivery API using a cursor for incremental sync.
- **`transport/startDebuggerBatch.ts`** — Transport layer that reuses `@datadog/browser-core`'s batch/flush infrastructure to send debugger snapshots to the logs intake.
- **`entries/main.ts`** — Public API surface (`datadogDebugger.init()`). Exposes `$dd_entry`/`$dd_return`/`$dd_throw`/`$dd_probes` hooks on `globalThis` for instrumented code. Defines the global `DD_DEBUGGER` object.

### Changes to `@datadog/browser-core`

- Added `'dd_debugger'` as a valid source in configuration and transport types, mapped to `'browser'` for the SDK source.
- Exported `computeTransportConfiguration` and the `Batch` type so the debugger package can create its own transport.

### E2E test framework and scenarios

- **`test/e2e/scenario/debugger.scenario.ts`** — 7 E2E test scenarios covering: basic snapshot sending, argument/return value capture, exception capture on throw, template message evaluation with expression segments, condition evaluation (both met and not met), and RUM correlation.
- **E2E framework extensions** — Added `.withDebugger()` builder method to `createTest()`, `DebuggerIntakeRequest` type and `intakeRegistry.debuggerEvents` for asserting on debugger events, debugger page setups for CDN/bundle/npm modes, and default debugger configuration.
- **`test/apps/vanilla/app.ts`** — Added `@datadog/browser-debugger` import and `DEBUGGER_INIT` support so debugger E2E tests work in the npm setup.

### Performance benchmarks

- **`test/apps/instrumentation-overhead/`** — Webpack test app for measuring instrumentation overhead with instrumented vs. uninstrumented function variants.
- **`test/performance/scenarios/instrumentationOverhead.scenario.ts`** — Benchmark scenario that stress-tests 10M function calls to measure the overhead of debugger instrumentation hooks.
- **`test/performance/createBenchmarkTest.ts`** — Extended with `instrumented_no_probes` and `instrumented_with_probes` scenario configurations and a dedicated `injectDebugger` function.

### Tooling

- Updated `scripts/build/build-test-apps.ts` to include the new test app and to use resolution paths when installing peer dependencies (needed for unpublished packages like `@datadog/browser-debugger` that only exist locally as `.tgz` files).
- Updated `scripts/dev-server/lib/server.ts` to serve the debugger bundle.
- Added debugger entry point to ESLint side-effects allowlist.

## Test instructions

1. **Unit tests**: `yarn test:unit --spec "packages/debugger/src/**/*.spec.ts"`
2. **E2E tests**: `yarn test:e2e:init && yarn test:e2e -g "debugger"`
3. **Performance testing**: `yarn build:apps && yarn test:performance`

## Checklist

- [x] Tested locally
- ~~Tested on staging~~ — N/A, this is a new pre-production package not yet deployed to any environment


watson:
* **#4449** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4449?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 5  dependent PRs ([#4480](https://github.com/DataDog/browser-sdk/pull/4480) <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4480?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4503](https://github.com/DataDog/browser-sdk/pull/4503) <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4503?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4508](https://github.com/DataDog/browser-sdk/pull/4508) <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4508?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 2 others) 👈 <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4449?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4513** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4513?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/debugger/src/domain/api.ts
```diff
@@ -0,0 +1,331 @@
...
+    method: probe.where.methodName,
+    version: debuggerApi.version,
+    // thread_id: 1,
+    thread_name: threadName,
+  }
+
+  // Get the RUM internal context for trace correlation
+  const rumContext = rumApi?.getInternalContext?.()
+  const dd = {
+    trace_id: rumContext?.session_id,
```
mormubis: `session_id` is being used as `trace_id`. Is this intentional?

packages/debugger/src/domain/stacktrace.ts
```diff
@@ -0,0 +1,61 @@
...
+  columnNumber: number
+}
+
+/**
+ * Capture the current stack trace
+ *
+ * @param skipFrames - Number of frames to skip from the top of the stack (default: 0)
+ * @returns Array of stack frames
+ */
+export function captureStackTrace(skipFrames = 0): StackFrame[] {
```
mormubis, line 18: We already have stack trace parsing in `@datadog/browser-core`. Can we reuse it instead of a new implementation?

packages/debugger/src/domain/api.ts
watson: Fixed 👍

packages/debugger/src/domain/stacktrace.ts
watson: Fixed 👍

mormubis (approved):
I'd get another approval before merging. We usually look for at least two approvals.

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -331,10 +333,14 @@ export interface Configuration extends TransportConfiguration {
...
   // internal
   sdkVersion: string | undefined
-  source: 'browser' | 'flutter' | 'unity'
+  source: SdkSource
   variant: string | undefined
 }
 
+function toSdkSource(source: TransportConfiguration['source']): SdkSource {
+  return source === 'dd_debugger' ? 'browser' : source
+}
```
thomas-lebeau: **❓ question:** ‏why setting the source to `dd_debugger` to set it back to browser later?

I think we should keep `dd_debugger` (or just `debugger` for consistency) and your team should maintain their own monitors, dashboards, ...

test/e2e/lib/types/global.ts
```diff
@@ -1,10 +1,12 @@
 import type { LogsGlobal } from '@datadog/browser-logs'
+import type { DebuggerPublicApi } from '@datadog/browser-debugger'
 import type { RumGlobal } from '@datadog/browser-rum'
 
 declare global {
   interface Window {
     DD_LOGS?: LogsGlobal
     DD_RUM?: RumGlobal
+    DD_DEBUGGER?: DebuggerPublicApi
```
thomas-lebeau: **💬 suggestion:** the types `LogsGlobal` and `RumGlobal` are deprecated in favor of `DatadogLogs` and `DatadogRum`. what about using `DatadogDebugger` for consistency?

packages/debugger/README.md
```diff
@@ -0,0 +1,35 @@
...
+  applicationId: '<DATADOG_APPLICATION_ID>',
+  clientToken: '<DATADOG_CLIENT_TOKEN>',
+  site: '<DATADOG_SITE>',
+  service: 'my-web-application',
+  //  env: 'production',
+  //  version: '1.0.0',
+})
+```
+
+If [Datadog RUM][3] is also initialized on the page, debugger snapshots automatically include RUM context (session, view, user action) without any additional configuration.
```
thomas-lebeau: **❓ question:** ‏How does this works? I must have missed something, I didn't see how this sdk collects sessions

watson: I think the text here should probably be slightly modified to read better, but the gist is that we tie the probe result to the current trace/span: https://github.com/DataDog/browser-sdk/pull/4449/files#diff-88cda10a3adeb04ab7401bf422650f85908ff2261c31b4411959b7c80be817d4R330-R341

thomas-lebeau: 🤔 these fields (`traceId` and `spanId`) don't exist in [`RumInternalContext`](https://github.com/DataDog/browser-sdk/blob/12866483a831acbaad98ad73be49085e8ffed173/packages/core/src/domain/contexts/rumInternalContext.type.ts), is it really working?

watson: Ha interesting. I was actually looking for ways I could test this in an integration test without mocking it, but couldn't find a good way to do it, so I settled for a unit test after my AI had added it. Turns out my gut feeling was correct that this needed better testing

thomas-lebeau: We use `session.id` to correlate RUM event with logs, profiles and session replay.

packages/debugger/src/domain/api.ts
```diff
@@ -0,0 +1,349 @@
...
+  const rumContext = globalObj.DD_RUM?.getInternalContext?.()
+  const traceId = getStringContextValue(rumContext, 'trace_id')
+  const spanId = getStringContextValue(rumContext, 'span_id')
+
+  return traceId && spanId
+    ? {
+        trace_id: traceId,
+        span_id: spanId,
+      }
+    : undefined
```
BenoitZugmeyer: **issue**: trace_id and span_id don't exist in `getInternalContext`

packages/debugger/src/domain/capture.ts
```diff
@@ -0,0 +1,497 @@
...
+  notCapturedReason?: string
+  fields?: Record<string, CapturedValue>
+  elements?: CapturedValue[]
+  entries?: Array<[CapturedValue, CapturedValue]>
+}
+
+const hasReplaceAll = typeof (String.prototype as any).replaceAll === 'function'
+const replaceDots = hasReplaceAll
+  ? (str: string) => (str as string & { replaceAll: (s: string, r: string) => string }).replaceAll('.', '_')
+  : (str: string) => str.replace(/\./g, '_')
```
BenoitZugmeyer:
```suggestion
const replaceDots = (str: string) => str.replace(/\./g, '_')
```

packages/debugger/src/domain/condition.ts
```diff
@@ -0,0 +1,73 @@
...
+  const fnBody = `return ${condition}`
+  const functionCache = new Map<string, (...args: any[]) => boolean>()
+
+  return {
+    evaluate: (contextKeys: string[]) => {
+      const cacheKey = contextKeys.join(',')
+      let fn = functionCache.get(cacheKey)
+      if (!fn) {
+        // eslint-disable-next-line no-new-func, @typescript-eslint/no-implied-eval
+        fn = new Function(...contextKeys, fnBody) as (...args: any[]) => boolean
```
BenoitZugmeyer, line 35: *issue:* eval is evil, we shouldn't rely on eval if possible. Best practice is often to use CSP that block such functionality.

Is there something specific in the language that requires "inline" JavaScript code? Else we can probably parse its AST and run it using JS without actually generating JS code.

packages/debugger/src/domain/deliveryApi.ts
```diff
@@ -0,0 +1,146 @@
+import type { TimeoutId } from '@datadog/browser-core'
+import { display, fetch, getGlobalObject, mockable, setInterval, clearInterval } from '@datadog/browser-core'
+import { addProbe, removeProbe } from './probes'
+import type { Probe } from './probes'
+
+declare const __BUILD_ENV__SDK_VERSION__: string
+
+const DELIVERY_API_PATH = '/api/ui/debugger/probe-delivery'
```
BenoitZugmeyer, line 8: **Question:** what is this route? Is this supposed to live on the customer website?

packages/debugger/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@datadog/browser-debugger",
+  "version": "6.32.0",
```
BenoitZugmeyer: **suggestion:**: mark this as `"private": true` so we don't publish it until it's ready for customer consumpiton

test/performance/createBenchmarkTest.ts
```diff
@@ -116,6 +128,88 @@ async function injectSDK(page: Page, scenarioConfiguration: ScenarioConfiguratio
   )
 }
 
+async function injectDebugger(page: Page, scenarioConfiguration: ScenarioConfiguration, _scenarioName: string) {
+  // Set flag for app to use instrumented functions
+  await page.addInitScript(() => {
+    ;(window as any).USE_INSTRUMENTED = true
+  })
+
+  // Load debugger SDK (using local build for now)
```
BenoitZugmeyer: **question:** why not using the actual SDK?

test/e2e/scenario/debugger.scenario.ts
```diff
@@ -0,0 +1,304 @@
...
+      }
+    }
+  })
+}
+
+test.describe('debugger', () => {
+  createTest('send debugger snapshot when instrumented function is called')
+    .withDebugger()
+    .run(async ({ intakeRegistry, flushEvents, page, browserName, servers }) => {
+      test.skip(browserName !== 'chromium', 'Debugger tests require Chromium')
```
BenoitZugmeyer: **question** why debugger tests require Chromium?

test/e2e/lib/framework/intakeProxyMiddleware.ts
```diff
@@ -143,6 +162,23 @@ async function readRumOrLogsIntakeRequest(
   }
 }
 
+async function readDebuggerIntakeRequest(
```
BenoitZugmeyer: **suggestion:** this is a copy of the `readRumOrLogsIntakeRequest` function. You can just re-use it

packages/debugger/src/domain/capture.ts
watson: I used this implementation for performance reasons, as I assumed that in some browsers, it might be faster to use `replaceAll` over `replace` with a regular expression. What is your reason for wanting to simplify this?

packages/debugger/src/domain/deliveryApi.ts
watson: This is only a temporary path, that will only work when running inside of web-ui. I have a follow-up PR to generalize it: https://github.com/DataDog/browser-sdk/pull/4480

I on purpose didn't include it in this PR because it was already in review by the time I made that change. If you prefer, I can merge that PR into this one?

packages/debugger/src/domain/api.ts
watson: I'll remove it for now and deal with adding session_id in a follow up PR

test/e2e/scenario/debugger.scenario.ts
watson: Honestly, I can't remember why I did this. I remember running into some issue but I'm not sure anymore. I'll try to re-enable it and see what happens.

test/performance/createBenchmarkTest.ts
watson: I think this is from when I was experimenting with the structure of the code to make the overhead as small as possible when it was a no-op instrumentation. After then building the actual implementation, I didn't go back and update this to use that. I'll fix it.

packages/debugger/src/domain/api.ts
watson: Removed 👍

packages/debugger/package.json
watson: Fixed 👍

test/performance/createBenchmarkTest.ts
watson: Updated 👍

test/e2e/lib/framework/intakeProxyMiddleware.ts
watson: Fixed 👍

packages/debugger/src/domain/condition.ts
watson: Yeah, I know 😅  This is unfortunately the best way I've found to make this code performant. The input to `new Function()` is only coming from the probe definitions which are fetched from the `debugger-delivery-api` service in `dd-source`. Users of the website should not have access to manipulate this input. I previously brought this up with the security team and they have given 👍

watson: Background: Probes can have conditions that need to be evaluated at runtime to know if the probe should trigger or not. These conditions are defined in the probe definition in a custom DSL. This DSL needs to be translated to JavaScript so we can execute it every time the instrumented method is called. That's why it needs to be performant.

packages/debugger/README.md
watson: Removed for now. See also this related discussion: https://github.com/DataDog/browser-sdk/pull/4449#discussion_r3181031089

test/e2e/lib/types/global.ts
watson: Fixed 👍

test/e2e/scenario/debugger.scenario.ts
watson: Seems to work fine on all browsers 👍

packages/core/src/domain/configuration/configuration.ts
watson: Yeah, I agree: The round-trip really was a bit messy as written. The two roles `source` was playing got conflated:

- **URL routing source** (`?ddsource=...`): `dd_debugger` matters here, and this is what the existing monitors / dashboards / `source:dd_debugger` queries in logs-backend and debugger-backend key on. That was already preserved end-to-end.
- **RUM event SDK source** (the `source` field on RUM events): this is what `Configuration.source` (`SdkSource`) feeds into. The RUM events schema only accepts `'browser' | 'flutter' | 'unity'`, so we can't keep `'dd_debugger'` here without breaking event validation.

The `toSdkSource` helper was a defensive type bridge between those two, but the Debugger SDK never actually goes through `validateAndBuildConfiguration` (it calls `computeTransportConfiguration` directly), so it was mostly dead code that just made the intent harder to read.

I pushed a follow-up commit 09376b12b27f4e7cd243dd2584b99d6745667080 that drops `'dd_debugger'` from `InitConfiguration.source` and moves it to an explicit `sourceOverride` argument on `computeTransportConfiguration`. That removes `toSdkSource`, makes the two concepts visibly separate, and keeps wire behavior identical (URLs still go out with `?ddsource=dd_debugger&...&dd-evp-origin=browser`, monitors are unaffected).

Happy to keep that commit if you like the new shape or if you'd rather have the cleanup done separately (or differently), let me know and I can revert it 😊

test/e2e/scenario/debugger.scenario.ts
watson: I was wrong. Firefox has some issues: https://gitlab.ddbuild.io/DataDog/browser-sdk/-/jobs/1664026268

I'll investigate

watson: Just pushed a commit to try and increase the timeouts: 038eb976fafacca40dd06d5cb07c976f2d65ff68

watson: I wonder if this is related to the performance degradation we know exists for Firefox 🤔 (as Firefox can't optimize the functions as well as other JavaScript engines can)

packages/debugger/src/domain/capture.ts
BenoitZugmeyer: > What is your reason for wanting to simplify this?

Simple is better, also smaller bundle size. But if you measured it and think the performance benefit is worth it, it's be fine, just mention it in a comment maybe?

watson: In [a naive benchmark](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBMUZFZE1CcDZPU0JlTGVRMWQzWEoiLCJjb2RlIjoiJ2Zvb2JhcicucmVwbGFjZShEQVRBLCAnXycpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InJlcGxhY2UifSx7ImlkIjoibEROZzgwTE04Wk5zc1JNakRsRHlVIiwiY29kZSI6Iidmb29iYXInLnJlcGxhY2UoL1xcLi9nLCAnXycpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InJlcGxhY2UgKGlubGluZSByZWdleCkifSx7ImlkIjoiaDBoQXotczdkazZBeTVfUC13c3oyIiwiY29kZSI6Iidmb29iYXInLnJlcGxhY2VBbGwoJy4nLCAnXycpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InJlcGxhY2VBbGwifV0sImNvbmZpZyI6eyJuYW1lIjoiIiwicGFyYWxsZWwiOnRydWUsImRhdGFDb2RlIjoicmV0dXJuIC9cXC4vZyIsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX19fQ) it's about 20% faster to use `replaceAll` in this scenario in Google Chrome.

watson: I'll add a comment 👍

packages/debugger/src/domain/condition.ts
BenoitZugmeyer: I see! Sounds good then, we'll just need to document it for users

packages/debugger/src/domain/deliveryApi.ts
BenoitZugmeyer: Ok. Let's not add more things to that pr, a second pr is fine

packages/debugger/src/domain/condition.ts
watson: Yeah, I plan to add extensive documentation about this before this goes live (after we have dogfooded it outselves)

packages/core/src/domain/configuration/configuration.ts
```diff
@@ -331,7 +333,7 @@ export interface Configuration extends TransportConfiguration {
 
   // internal
   sdkVersion: string | undefined
-  source: 'browser' | 'flutter' | 'unity'
+  source: SdkSource
```
BenoitZugmeyer: **suggestion:** remove this property

`source` should be typed as `TransportSource`. Also, `source` is already declared in `TransportConfiguration`, so maybe we don't need to redeclare it here

BenoitZugmeyer (approved):

packages/core/src/domain/configuration/configuration.ts
watson: Pushed a fix: 0f3dfe2ab3f6d39dd160a651ff9b0fab490e4d34 (If I understood your suggestion correctly)

test/e2e/scenario/debugger.scenario.ts
watson: Found the bug: It was a race condition in the e2e test which has now been fixed

packages/core/src/domain/configuration/configuration.ts
thomas-lebeau: great, that's a nice improvement, thank you!

thomas-lebeau (approved):

PR #4475, opened 2026-04-14
🐛 Prevent cross-tab session adoption after expiry

thomas-lebeau:
## Motivation

In v7, when a session expires in a multi-tab scenario, Tab 2 could silently adopt a new session created by Tab 1 without any user interaction. This is a behavior change from v6, where each tab was required to renew its session through its own user interaction after expiry. This PR restores the v6 behavior.

## Changes

- Track session expiry state per tab with a `sessionExpired` flag in the session manager
- After expiry, ignore new sessions written to the cookie by other tabs — each tab must renew via its own user interaction (click, tracking consent grant, or explicit `expandOrRenew` call)
- Update the existing unit test to assert the non-adoption behavior
- Add a multi-tab E2E test covering the full lifecycle: shared session, expiry, independent renewal, and eventual session convergence

## Test instructions

1. `yarn test:unit --spec packages/core/src/domain/session/sessionManager.spec.ts`
2. `yarn test:e2e -g "multi-tab sessions"`

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/sessionManager.ts
+++ b/packages/core/src/domain/session/sessionManager.ts
@@ -102,7 +102,10 @@ export function startSessionManager(
   })
   stopCallbacks.push(() => sessionContextHistory.stop())
 
+  let sessionExpired = false
+
   const { throttled: throttledExpandOrRenew, cancel: cancelExpandOrRenew } = throttle(() => {
+    sessionExpired = false
     strategy.setSessionState((state) => expandOrRenew(state, configuration)).catch(monitorError)
   }, ONE_SECOND)
   stopCallbacks.push(cancelExpandOrRenew)
@@ -178,6 +181,7 @@ export function startSessionManager(
   function setupSessionTracking() {
     trackingConsentState.observable.subscribe(() => {
       if (trackingConsentState.isGranted()) {
+        sessionExpired = false
         strategy.setSessionState((state) => expandOrRenew(state, configuration)).catch(monitorError)
       } else {
         expire()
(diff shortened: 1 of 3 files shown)


packages/core/src/domain/session/sessionManager.ts
```diff
@@ -244,11 +248,18 @@ export function startSessionManager(
...
+      if (!hasSession) {
+        sessionExpired = true
+      }
       expireObservable.notify()
       sessionContextHistory.closeActive(relativeNow())
     }
 
     if (hasSession && (!hadSession || sessionIdChanged)) {
+      if (sessionExpired) {
+        // Don't adopt another tab's session — this tab needs its own user interaction to renew
```
BenoitZugmeyer, line 260: **nitpick:** maybe rename the variable `hadUserInteraction`?

test/e2e/scenario/rum/multiTabSessions.scenario.ts
```diff
@@ -0,0 +1,58 @@
...
+
+      // Wait for the new session to be created and the next poll cycle
+      await page.waitForTimeout(1100)
+
+      const tab1Renewed = await page.evaluate(() => window.DD_RUM!.getInternalContext()?.session_id)
+      expect(tab1Renewed).toBeDefined()
+      expect(tab1Renewed).not.toBe(tab1Session)
+
+      // Wait for Tab 2's poll cycle to pick up the cookie change
+      await page2.waitForTimeout(1100)
```
BenoitZugmeyer: **question:** do we need this second timeout?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -244,11 +248,18 @@ export function startSessionManager(
         locksAvailable: Boolean(globalThis.navigator?.locks),
         cookieStoreAvailable: Boolean(globalThis.cookieStore),
       }
+      if (!hasSession) {
+        sessionExpired = true
+      }
```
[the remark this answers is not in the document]
thomas-lebeau: This is expected — from Tab B's perspective it never observed the expired state, so this is a normal cross-tab session ID change. This matches v6 behavior.

packages/core/src/domain/session/sessionManager.ts
thomas-lebeau: I think `sessionExpired` better describes the tab's state ("session expired and not yet renewed"). This variable is not updated when there is a user interaction.

BenoitZugmeyer (approved):

PR #4477, opened 2026-04-14
⚗️ Add Nuxt error tracking

BeltranBulbarellaDD:
## Motivation

This PR adds first-class Nuxt error tracking, covering both Vue rendering/lifecycle errors (via `vueApp.config.errorHandler`) and Nuxt startup errors thrown in plugins or app hooks (via the `app:error` hook).

### New public APIs (exported from `@datadog/browser-rum-nuxt`)

- **`addNuxtError(error, instance, info)`** — a low-level function compatible with both `vueApp.config.errorHandler` and Nuxt's `app:error` hook. Intended for users who want manual control. It:
  - Captures a handling stack at call time
  - Builds a component stack string from the Vue component hierarchy and the `info` string
  - Merges a `dd_context` property from the error object (if present) into the event context
  - Tags all events with `framework: 'nuxt'`
  - Buffers errors thrown before RUM initialises and flushes them once `datadogRum.init` completes

- **`setupNuxtErrorHandling(nuxtApp)`** — a convenience helper that sets up both handlers automatically. It:
  - Wraps the existing `vueApp.config.errorHandler` (preserving Nuxt's own error page handler)
  - Hooks into `app:error` for errors thrown during plugin/app startup
  - Uses `WeakSet`-based deduplication to prevent double-reporting when the same error bubbles through both `errorHandler` and `app:error`

### Internal changes

- `nuxtRumPlugin` now implements the `onRumStart` lifecycle hook to receive and distribute the `addError` function to buffered subscribers
- `createHandlingStack` extended with a `'nuxt error'` source label

## Test instructions

1. **Unit tests:**
   ```bash
   yarn test:unit --spec packages/rum-nuxt/src/domain/error/addNuxtError.spec.ts
   ```

2. **E2E tests:**
   ```bash
   yarn test:e2e:init
   yarn test:e2e -g "nuxt error"
   ```

3. **Manual:**
   - Start the Nuxt test app and navigate to `/error-test`. Click "Trigger Error" and verify a single RUM error event is collected with `error.source: custom`, `context.framework: nuxt`, and a `handling_stack`.
   - Navigate to `/startup-error` and verify a single RUM error event is collected for the startup error thrown in the plugin, again with `context.framework: nuxt` and a `handling_stack`.

## Checklist

- [X] Tested locally
- [ ] Tested on staging


BeltranBulbarellaDD:

packages/rum-nuxt/src/domain/error/setupNuxtErrorHandling.ts
```diff
@@ -0,0 +1,36 @@
...
+  // Wrap existing errorHandler rather than replacing it, so Nuxt's own
+  // handleVueError (which drives the error page) is still called.
+  const original = nuxtApp.vueApp.config.errorHandler
+  nuxtApp.vueApp.config.errorHandler = (error, instance, info) => {
+    deduplicatedAddNuxtError(error, instance, info)
+    original?.(error, instance, info)
+  }
+
+  nuxtApp.hook('app:error', (err) => {
+    deduplicatedAddNuxtError(err, null, '')
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: I don't agree. 
> on client-side, it will throw a non-fatal error for you to handle. ([Link](https://nuxt.com/docs/4.x/api/utils/create-error#in-vue-app)) 

And that's what we do, we handle it

packages/rum-nuxt/src/domain/error/setupNuxtErrorHandling.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  nuxtApp.hook('app:error', (err) => {
+    deduplicatedReportError(err, null, '')
+  })
+}
+
+function callOnce(func: NuxtErrorReporter) {
+  const seen = new WeakSet<object>()
+  return (error: unknown, instance: ComponentPublicInstance | null, info: string) => {
```
mormubis, line 53: `callOnce` reads like it calls the function at most once total (like lodash's `once`), but it actually deduplicates per error object. Maybe `deduplicateByError` or `withDeduplication`? What do you think?

packages/rum-nuxt/src/entries/main.ts
```diff
@@ -1,2 +1,4 @@
 export type { NuxtPlugin, NuxtPluginConfiguration } from '../domain/nuxtPlugin'
 export { nuxtRumPlugin } from '../domain/nuxtPlugin'
+export { addNuxtError } from '../domain/error/addNuxtError'
+export { setupNuxtErrorHandling } from '../domain/error/setupNuxtErrorHandling'
```
mormubis: The only call site for `setupNuxtErrorHandling` is `nuxtPlugin.ts`. I don't think we need to export it. Vue and Next.js only expose the self-contained `addXError` function and keep the wiring internal. Should we do the same here?

mormubis (commented):
The description mentions filtering out H3Error/statusCode from `app:error`, but we updated the `app:error` hook. Should we update the description?

mormubis (approved):

PR #4491, opened 2026-04-15
👷 Ensure test apps use peerDependencies for @datadog/* packages

thomas-lebeau:
## Motivation

Test apps that list `@datadog/*` packages in `dependencies` cause `yarn install` to resolve them from the registry instead of using local workspace versions. This leads to dirty git trees after building (updated lockfiles, unexpected resolutions) and makes builds non-reproducible. The [convention](https://github.com/DataDog/browser-sdk/pull/4028) is to use `peerDependencies` so local packages are only installed at build time via `build-test-apps`.

see also: #4387

## Changes

- Add a `check-packages` rule that enforces `peerDependencies` for `@datadog/*` packages in test apps.
- Migrate `nuxt-app` from `dependencies` to `peerDependencies` for `@datadog/*` packages.

## Test instructions

1. Run `yarn node scripts/check-packages.ts` — should pass.
2. Add a `@datadog/*` package to `dependencies` in any test app's `package.json` — `check-packages` should fail with a clear error.
3. Run `yarn build:apps --app nuxt-app` — should not leave the directory dirty.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/build/build-test-apps.ts
+++ b/scripts/build/build-test-apps.ts
@@ -124,7 +124,9 @@ function buildApp(appName: string) {
   if (packageJson.peerDependencies) {
     // For each peer dependency, install it
     for (const [name] of Object.entries(packageJson.peerDependencies)) {
-      command`yarn add -D ${name}`.withCurrentWorkingDirectory(appPath).run()
+      const resolution = packageJson.resolutions?.[name]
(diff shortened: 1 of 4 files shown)


bcaudan (approved):
LGTM

BeltranBulbarellaDD (approved):

PR #4506, opened 2026-04-20
📝 Add README for rum-nuxt package

BeltranBulbarellaDD:
## Motivation

Now that Routing and Erorr handling has been merged in Nuxt, we can add its full documentation. 

## Changes

Adds `readme.md` file

## Checklist

- [ ] Tested locally
- [ ] Tested on staging
- [X] Updated documentation and/or relevant AGENTS.md file

### DIFF
--- a/packages/rum-nuxt/README.md
+++ b/packages/rum-nuxt/README.md
@@ -67,7 +67,6 @@ import { addNuxtError } from '@datadog/browser-rum-nuxt'
 
 onErrorCaptured((error, instance, info) => {
   addNuxtError(error, instance, info)
-  return false
 })
 </script>
 ```
--- a/packages/rum-nuxt/README.md
+++ b/packages/rum-nuxt/README.md
@@ -1,3 +1,111 @@
 # RUM Browser Monitoring - Nuxt integration
 
-This package provides Nuxt-specific integrations for Datadog Browser RUM.
+**Note**: This integration is in Preview. Features and configuration are subject to change.
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


packages/rum-nuxt/README.md
```diff
@@ -1,3 +1,110 @@
...
+```ts
+import { datadogRum } from '@datadog/browser-rum'
+import { nuxtRumPlugin } from '@datadog/browser-rum-nuxt'
+import { defineNuxtPlugin, useNuxtApp, useRouter } from 'nuxt/app'
+
+export default defineNuxtPlugin({
+  name: 'datadog-rum',
+  enforce: 'pre',
+  setup() {
+    datadogRum.init({
```
mormubis, line 37: I've just found out that `defineNuxtPlugin`, `useRouter`, and `useNuxtApp` are auto-imported in Nuxt, so the `from 'nuxt/app'` line isn't needed. https://nuxt.com/docs/guide/concepts/auto-imports#explicit-imports. What do you think? Should we adapt the example to Nuxt auto-import?

mormubis (approved):
Looks good!

amortemousque (approved):

packages/rum-nuxt/README.md
BeltranBulbarellaDD: I think both ways are fine. IMO I would prefer to leave it as the IDE considers it a linter error, even thought the application will build just fine. I think it does not hurt and prevents an import from the wrong component in case a developer wants to get rid of the linting issue locally.

PR #4507, opened 2026-04-20
✨ [PANA-6624] Enable the new session replay serialization algorithm by default

sethfowler-datadog:
## Motivation

This PR makes the new session replay serialization algorithm the default in V7.

## Changes

- The experimental feature flags have been removed.
- The internal API functions now return records in the new format. (I've confirmed that all callers can handle this.)
- The recording code no longer directly depends on the old serialization code, allowing it to be excluded from the bundle. (Some tests do still depend on it, so the code can't be removed yet, but we'll get there.)
- Some session replay tests had two variants: one version tested the old format, and one version tested the new one. In cases where the variant testing the old format no longer works (e.g. the E2E tests, which relied on the removed experimental feature flags), that variant has been removed.
- Many other session replay tests have been updated.

## Test instructions

It should be enough to test recording with `yarn dev` in the usual way.

## Checklist

- [x] Tested locally
- [x] Tested on staging


sethfowler-datadog:
Effects on bundle size for `rum_recorder`:

| Format | Status Quo | This PR | Improvement |
| -------- | -----------| --------| ------------------ |
| Uncompressed | 27.10 KiB | 21.27 KiB | 21.5% |
| Compressed | 9.27 KiB | 7.73 KiB | 16.6% |

packages/rum/src/domain/record/startFullSnapshots.ts
```diff
@@ -28,7 +23,7 @@ export function startFullSnapshots(
   emitStats: EmitStatsCallback,
   flushMutations: () => void,
   scope: RecordingScope,
-  serialize: SerializeFullSnapshotCallback = defaultSerializeFullSnapshotCallback()
+  serialize: SerializeFullSnapshotCallback = serializeFullSnapshotAsChange
```
gonzalezreal, line 26: Do we still need the `serialize` parameter here?

It looks like it was added in #4185 so tests could switch between v1 and v2 serializers during the migration. Now that `v2` is the only path, I couldn't find any caller passing a non-default value.

If you agree, the parameter and its type could be dropped and line 74 could just call `serializeFullSnapshotAsChange`. Happy to leave it if you'd rather keep the hook around.

gonzalezreal (approved):
🚀

mormubis (approved):

packages/rum/src/domain/record/startFullSnapshots.ts
sethfowler-datadog: Indeed, we can remove this one; only the incremental snapshot version is still being used. I've pushed a commit to remove this parameter.

PR #4516, opened 2026-04-21
RUM-15702 Add trace sampling decision to DatadogEventBridge

arroz:
## Motivation

With deterministic sampling, the decision of sampling requests on native clients became consistent, based on the sampling rate and session ID. However, when a native client loads an instrumented web site on a WebView, there is a disconnect between the sampling decision used by the browser SDK of the page being viewed in the WebView, and the native client's decision. This cannot be fixed by post-processing on the native client when the data is funneled through it, since the browser SDK needs to know if headers should be injected on requests, or not. More details [here](https://datadoghq.atlassian.net/browse/RUM-15702) (internal).

To fix this problem, this patch introduces a way for native clients to bridge the sampling decision.

This is the Browser SDK counterpart to https://github.com/DataDog/dd-sdk-ios/pull/2859.

## Changes

This patch introduces an additional function in `DatadogEventBridge`: `getIsTraceSampled()`. The native clients can inject 3 values: `true` (if requests should be sampled), `false` (if they should not), or `null` if no decision was made (for example, if first party host tracing is not enabled). For the two initial values, sampling rate is set, respectively, to 100 and 0. For `null`, we fallback to the previous behavior, where the browser SDK is responsible for the sampling rate (through configuration, etc).

Note: I'm not a frontend developer, these changes were mostly made using AI. Please review them taking this into account.

## Test instructions

This can only be tested using a native client SDK with the new functionality. I'm going to submit an iOS PR after this one with such functionality enabled.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/transport/eventBridge.ts
+++ b/packages/core/src/transport/eventBridge.ts
@@ -8,6 +8,7 @@ export interface BrowserWindowWithEventBridge extends Window {
 export interface DatadogEventBridge {
   getCapabilities?(): string
   getPrivacyLevel?(): DefaultPrivacyLevel
+  getIsTraceSampled?(): string
   getAllowedWebViewHosts(): string
   send(msg: string): void
 }
@@ -29,6 +30,12 @@ export function getEventBridge<T, E>() {
     },
     getPrivacyLevel() {
       return eventBridgeGlobal.getPrivacyLevel?.()
+    },
+    getIsTraceSampled() {
+      const value = eventBridgeGlobal.getIsTraceSampled?.()
+      if (value !== undefined) {
+        return value === 'true'
+      }
     },
     getAllowedWebViewHosts() {
       return JSON.parse(eventBridgeGlobal.getAllowedWebViewHosts()) as string[]
--- a/packages/core/src/transport/eventBridge.ts
+++ b/packages/core/src/transport/eventBridge.ts
@@ -8,7 +8,7 @@ export interface BrowserWindowWithEventBridge extends Window {
 export interface DatadogEventBridge {
   getCapabilities?(): string
   getPrivacyLevel?(): DefaultPrivacyLevel
-  getIsTraceSampled?(): string
+  getIsTraceSampled?(): boolean | null
   getAllowedWebViewHosts(): string
   send(msg: string): void
 }
@@ -33,8 +33,8 @@ export function getEventBridge<T, E>() {
     },
     getIsTraceSampled() {
       const value = eventBridgeGlobal.getIsTraceSampled?.()
-      if (value !== undefined) {
-        return value === 'true'
+      if (value === true || value === false) {
+        return value
       }
     },
     getAllowedWebViewHosts() {
--- a/packages/core/src/transport/eventBridge.ts
+++ b/packages/core/src/transport/eventBridge.ts
@@ -8,7 +8,7 @@ export interface BrowserWindowWithEventBridge extends Window {
 export interface DatadogEventBridge {
   getCapabilities?(): string
   getPrivacyLevel?(): DefaultPrivacyLevel
-  getIsTraceSampled?(): string
+  getIsTraceSampled?(): 'true' | 'false' | 'null'
   getAllowedWebViewHosts(): string
   send(msg: string): void
 }
@@ -32,15 +32,7 @@ export function getEventBridge<T, E>() {
       return eventBridgeGlobal.getPrivacyLevel?.()
     },
     getIsTraceSampled() {
-      // Bridge returns 'true', 'false', or no decision ('null', null, undefined, missing method).
-      // Normalize to boolean or undefined so callers only check one thing.
-      const value = eventBridgeGlobal.getIsTraceSampled?.()
-      if (value === 'true') {
-        return true
-      }
-      if (value === 'false') {
-        return false
-      }
+      return JSON.parse(eventBridgeGlobal.getIsTraceSampled?.() || 'null') as boolean | null
     },
     getAllowedWebViewHosts() {
       return JSON.parse(eventBridgeGlobal.getAllowedWebViewHosts()) as string[]
--- a/packages/core/src/transport/eventBridge.ts
+++ b/packages/core/src/transport/eventBridge.ts
@@ -8,7 +8,7 @@ export interface BrowserWindowWithEventBridge extends Window {
 export interface DatadogEventBridge {
   getCapabilities?(): string
   getPrivacyLevel?(): DefaultPrivacyLevel
-  getIsTraceSampled?(): boolean | null
+  getIsTraceSampled?(): string
   getAllowedWebViewHosts(): string
   send(msg: string): void
 }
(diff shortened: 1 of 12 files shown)


arroz:
recheck

arroz:

arroz:
Converting to draft to address the situation when a session roll over.

packages/core/src/transport/eventBridge.ts
```diff
@@ -30,6 +31,12 @@ export function getEventBridge<T, E>() {
     getPrivacyLevel() {
       return eventBridgeGlobal.getPrivacyLevel?.()
     },
+    getIsTraceSampled() {
+      const value = eventBridgeGlobal.getIsTraceSampled?.()
+      if (value === true || value === false) {
+        return value
+      }
```
thomas-lebeau: **💬 suggestion:** ‏let's keep thing simple:

```suggestion
      return eventBridgeGlobal.getIsTraceSampled?.()
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -8,6 +8,7 @@ export interface BrowserWindowWithEventBridge extends Window {
 export interface DatadogEventBridge {
   getCapabilities?(): string
   getPrivacyLevel?(): DefaultPrivacyLevel
+  getIsTraceSampled?(): boolean | null
```
thomas-lebeau:
```suggestion
  getIsTraceSampled?(): boolean
```

thomas-lebeau (commented):
Hi @arroz, Instead of exposing a new public API, it would be easier to read the bridge directly in the [tracer](https://github.com/DataDog/browser-sdk/blob/68209b3c9ff4e27321b4f196e862f3ee5c45e8ea/packages/rum-core/src/domain/tracing/tracer.ts#L128), something like:

```ts
  const traceSampled = canUseEventBridge() ? getEventBridge()?.getIsTraceSampled() : isSampled(session.id, configuration.traceSampleRate)
```

packages/core/src/transport/eventBridge.ts
arroz: Does that work correctly when `getIsTraceSampled()` returns `null` or does not exist? Sorry if the answer is trivial, I never coded TS and my JS is very rusty.

packages/core/src/transport/eventBridge.ts
arroz: Same question here. Don't know the nuances between `null` and `undefined`.

packages/core/src/transport/eventBridge.ts
thomas-lebeau: It will be best if it does not return `null`, any reason to not return a boolean?

otherwise, let's do that:
```suggestion
      return eventBridgeGlobal.getIsTraceSampled?.() ?? false
```

arroz: It needs to return `null`. The `null` has meaning, it's the case where we cannot make a decision (because, for example, first party host tracking is not configured), so we leave the decision to the browser SDK.

EDIT: So we cannot default to `false` also as in your suggestion, otherwise we would be preventing the Browser SDK from deciding.

thomas-lebeau: But the SDK cannot decide without knowing the session id.
So I think the event bridge should expose `getSessionId()` and maybe `getTraceSampleRate()` (unless we rely on the SDK's own configuration)

Note that deterministic sampling is not supported in SDK v6, it will be starting SDK v7 (ETA very soon)

arroz: Currently the SDK makes a decision. The ask I have is to override that decision form the native SDKs if we do have a decision, otherwise revert to the previous behaviour and allow the browser SDK to decide for itself.

In the future we will need to inject the session ID and keep it updated on rollovers for deterministic sampling. But since this is an urgent ask, the guidance we got from Benoit is to just inject the decision, if we have one.

packages/core/src/transport/eventBridge.ts
arroz: As discussed, it needs to return `null` as well.

packages/rum-core/src/domain/tracing/tracer.ts
```diff
@@ -125,7 +125,9 @@ function injectHeadersIfTracingAllowed(
     return
   }
 
-  const traceSampled = isSampled(session.id, configuration.traceSampleRate)
+  const traceSampled = canUseEventBridge()
+    ? (getEventBridge()?.getIsTraceSampled() ?? isSampled(session.id, configuration.traceSampleRate))
+    : isSampled(session.id, configuration.traceSampleRate)
```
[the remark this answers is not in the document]
arroz: This is OK, because we over-write `_dd.rule_psr` on the native side when the events are piped through. The rate coming from the Browser SDK does not matter in bridged situations.

packages/core/src/transport/eventBridge.ts
```diff
@@ -30,6 +31,17 @@ export function getEventBridge<T, E>() {
     getPrivacyLevel() {
       return eventBridgeGlobal.getPrivacyLevel?.()
     },
+    getIsTraceSampled() {
+      // Bridge returns 'true', 'false', or no decision ('null', null, undefined, missing method).
+      // Normalize to boolean or undefined so callers only check one thing.
```
thomas-lebeau: **💬 suggestion:** ‏we can enforce this comment in `DatadogEventBridge` type
```suggestion
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -8,6 +8,7 @@ export interface BrowserWindowWithEventBridge extends Window {
 export interface DatadogEventBridge {
   getCapabilities?(): string
   getPrivacyLevel?(): DefaultPrivacyLevel
+  getIsTraceSampled?(): string
```
thomas-lebeau:
```suggestion
  getIsTraceSampled?(): 'true' | 'false' | 'null'
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -30,6 +31,17 @@ export function getEventBridge<T, E>() {
...
+    getIsTraceSampled() {
+      // Bridge returns 'true', 'false', or no decision ('null', null, undefined, missing method).
+      // Normalize to boolean or undefined so callers only check one thing.
+      const value = eventBridgeGlobal.getIsTraceSampled?.()
+      if (value === 'true') {
+        return true
+      }
+      if (value === 'false') {
+        return false
+      }
```
thomas-lebeau: **💬 suggestion:** ‏Then we can simplify this:

```suggestion
      return JSON.parse(eventBridgeGlobal.getIsTraceSampled?.() || 'null') as boolean | null
```

packages/core/src/transport/eventBridge.ts
```diff
@@ -30,6 +31,17 @@ export function getEventBridge<T, E>() {
     getPrivacyLevel() {
       return eventBridgeGlobal.getPrivacyLevel?.()
     },
+    getIsTraceSampled() {
```
thomas-lebeau, line 34: **💭 thought:** ‏we should probably add some unit for this

arroz: Done, 4 tests were added to make sure parsing works properly.

thomas-lebeau (approved):

mormubis (approved):

PR #4518, opened 2026-04-22
feat: add Salesforce POC with view tracking and e2e tests

BeltranBulbarellaDD:
## Checklist

- [ ] Tested locally
- [ ] Tested on staging


packages/core/src/browser/addEventListener.ts
```diff
@@ -132,10 +133,20 @@ export function addEventListeners<Target extends EventTarget, EventName extends
     window.EventTarget && eventTarget instanceof EventTarget ? window.EventTarget.prototype : eventTarget
 
   const add = getZoneJsOriginalValue(listenerTarget, 'addEventListener')
+  if (typeof add !== 'function') {
```
amortemousque, line 139: **❓ question:**   Is `addEventListener` not supported in Salesforce? I don’t see it mentioned in the RFC. If it’s only missing on certain targets, we should probably handle the fallback at a higher level.

packages/core/src/domain/report/reportObservable.ts
```diff
@@ -60,11 +60,15 @@ function createReportObservable(reportTypes: ReportType[]) {
 
 function createCspViolationReportObservable(configuration: Configuration) {
   return new Observable<RawReportError>((observable) => {
-    const { stop } = addEventListener(configuration, document, DOM_EVENT.SECURITY_POLICY_VIOLATION, (event) => {
-      observable.notify(buildRawReportErrorFromCspViolation(event))
-    })
+    try {
```
amortemousque, line 63: **❓ question:** ‏Does this try/catch mean [SECURITY_POLICY_VIOLATION](https://developer.mozilla.org/en-US/docs/Web/API/SecurityPolicyViolationEvent) is not supported by Salesforce?

packages/rum-core/src/domain/getSelectorFromElement.ts
```diff
@@ -52,6 +52,18 @@ interface SubtreeTarget {
   target: Element
 }
 
+export function escapeCssValue(value: string) {
```
amortemousque, line 55: # #
**❓ question:** ‏Have you looked into a CSS.escape polyfill for cases where CSS.escape isn’t supported?

packages/rum/src/domain/salesforce/resourcePlugin.ts
amortemousque, line 1: **❓ question:** ‏Why do we need a resourcePlugin? From the RFC, I thought a fallback polling performance.getEntriesByType('resource') would be enough?

packages/rum-core/src/browser/cookieObservable.ts
```diff
@@ -27,22 +27,27 @@ export function createCookieObservable(configuration: Configuration, cookieName:
...
-      DOM_EVENT.CHANGE,
-      (event) => {
-        // Based on our experimentation, we're assuming that entries for the same cookie cannot be in both the 'changed' and 'deleted' arrays.
-        // However, due to ambiguity in the specification, we asked for clarification: https://github.com/WICG/cookie-store/issues/226
-        const changeEvent =
-          event.changed.find((event) => event.name === cookieName) ||
-          event.deleted.find((event) => event.name === cookieName)
-        if (changeEvent) {
-          callback(changeEvent.value)
+    try {
```
amortemousque, line 30: **💬 suggestion:** ‏Instead of using try/catch, you can check whether (window as CookieStoreWindow).cookieStore exists.

packages/rum-core/src/domain/contexts/urlContexts.ts
```diff
@@ -41,8 +42,11 @@ export function startUrlContexts(
   let previousViewUrl: string | undefined
 
   lifeCycle.subscribe(LifeCycleEventType.BEFORE_VIEW_CREATED, ({ startClocks, url }) => {
-    const locationHref = mockable(location).href
+    const locationHref = mockable(getGlobalObject<Window>().location)?.href
     const viewUrl = url !== undefined ? buildUrl(url, locationHref).href : locationHref
+    if (!viewUrl) {
```
amortemousque, line 47: **❓ question:** ‏In which case viewUrl can be undefined?

packages/core/src/tools/instrumentMethod.ts
```diff
@@ -73,6 +73,11 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
   onPreCall: (this: null, callInfos: InstrumentedMethodCall<TARGET, METHOD>) => void,
   { computeHandlingStack }: { computeHandlingStack?: boolean } = {}
 ) {
+  const methodDescriptor = findDescriptorInPrototypeChain(targetPrototype, method)
```
amortemousque: **❓ question:** ‏This is a very low-level check. Can you explain in which feature context unstrumentMethod fails?

packages/core/src/domain/report/reportObservable.ts
BeltranBulbarellaDD: Screenshot 2026-05-26 at 22 25 53
Exactly. I'll add it to the RFC

packages/core/src/browser/addEventListener.ts
BeltranBulbarellaDD: Agreed! Salesforce supports `addEventListener` I was trying to mute a telemetry error: `Cannot read properties of undefined (reading 'href')`. I'll move the check.

packages/rum/src/domain/salesforce/resourcePlugin.ts
BeltranBulbarellaDD: Not needed anymore, we have PerformanceObserver.

BeltranBulbarellaDD:
closing for https://github.com/DataDog/browser-sdk/pull/4726

PR #4521, opened 2026-04-22
feat(core): export createEndpointBuilder and add flagevaluation TrackType

typotter:
## Summary

- Exports `createEndpointBuilder` and `TrackType` from the `@datadog/browser-core` package root
- Adds `'flagevaluation'` to the `TrackType` union

## Motivation

`@datadog/openfeature-browser` needs to build a custom endpoint for flag evaluation tracking events (`/api/v2/flagevaluation`). Because `createEndpointBuilder` was not exported from the package root, it was accessing it via a deep internal subpath import:

```ts
import { createEndpointBuilder, type TrackType } from '@datadog/browser-core/cjs/domain/configuration'
```

This caused a `/cjs/` specifier to appear verbatim in the ESM build output, pulling CJS modules into the ESM dependency graph and breaking tree-shaking for consumers. A post-build rewrite script was added as a workaround (DataDog/openfeature-js-client#247).

With this change, `@datadog/openfeature-browser` can use the clean public import:

```ts
import { createEndpointBuilder, type TrackType } from '@datadog/browser-core'
```

and the workaround can be removed.

## Notes

`'flagevaluation'` is intentionally not wired into `TransportConfiguration` or `computeEndpointBuilders` — external packages call `createEndpointBuilder` directly to manage their own endpoint builders. A comment in `endpointBuilder.ts` explains this.

## Test plan

- [ ] Existing tests pass (no behavioral change — exports only)
- [ ] TypeScript compilation succeeds with the new exports


BenoitZugmeyer (approved):

gogusarov (approved):

typotter:
Closing — obviated by DataDog/openfeature-js-client#261.

PR #4523, opened 2026-04-23
[RUM-15280] ✨ Expose DEFAULT_TRACKED_RESOURCE_HEADERS on RumPublicApi

bdibon:
## Motivation

Dogfooding feedback on `trackResourceHeaders`: customers using the CDN bundle have no runtime access to the default header list, so they can't extend it (e.g. `[...defaults, 'x-my-header']`) without hardcoding. Named exports from `entries/main.ts` reach NPM users but not the `window.DD_RUM` global, so the constant needs to be attached to the public API object itself.

## Changes

- Expose `DEFAULT_TRACKED_RESOURCE_HEADERS` as a property on `RumPublicApi` (marked `@hidden` for now), so CDN users can access it as `window.DD_RUM.DEFAULT_TRACKED_RESOURCE_HEADERS`.
- Update the `trackResourceHeaders` JSDoc example in `RumInitConfiguration` to use `...DEFAULT_TRACKED_RESOURCE_HEADERS.map((h) => ({ name: h }))` instead of the `name: (h) => ...includes(h)` predicate form — matches how customers actually want to extend the defaults.

Note: this introduces a non-function, non-state property on `RumPublicApi` for the first time; every other member is either a method, a `ContextManager`, `initConfiguration`, or `version`.

## Test instructions

1. Check out the branch and run `yarn build` (or `yarn dev`).
2. Load the sandbox page with the CDN bundle in a browser.
3. In DevTools console, run:
   ```js
   window.DD_RUM.DEFAULT_TRACKED_RESOURCE_HEADERS
   ```
   Expect the array `['cache-control', 'etag', 'age', 'expires', 'content-type', 'content-encoding', 'vary', 'content-length', 'server-timing', 'x-cache']`.
4. Verify it can be used to extend the config:
   ```js
   window.DD_RUM.init({
     applicationId: '<APP_ID>',
     clientToken: '<CLIENT_TOKEN>',
     trackResourceHeaders: [
       ...window.DD_RUM.DEFAULT_TRACKED_RESOURCE_HEADERS.map((h) => ({ name: h })),
       { name: 'x-request-id' },
     ],
   })
   ```
   and that a resource event carries the expected headers.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

[RUM-15273]: https://datadoghq.atlassian.net/browse/RUM-15273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/rum-core/src/boot/rumPublicApi.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.ts
@@ -45,6 +45,7 @@ import type { ViewHistory } from '../domain/contexts/viewHistory'
 import type { RumSessionManager } from '../domain/rumSessionManager'
 import type { ReplayStats } from '../rawRumEvent.types'
 import { ActionType, VitalType } from '../rawRumEvent.types'
+import { DEFAULT_TRACKED_RESOURCE_HEADERS } from '../domain/configuration'
 import type { RumConfiguration, RumInitConfiguration } from '../domain/configuration'
 import type { ViewOptions } from '../domain/view/trackViews'
 import type {
@@ -513,6 +514,13 @@ export interface RumPublicApi extends PublicApi {
    * @hidden // TODO: replace by @since when GA
    */
   failFeatureOperation: (name: string, failureReason: FailureReason, options?: FeatureOperationOptions) => void
+
+  /**
+   * List of default headers to use in the {@link RumInitConfiguration.trackResourceHeaders | trackResourceHeaders} option.
+   *
+   * @hidden
+   */
+  DEFAULT_TRACKED_RESOURCE_HEADERS: typeof DEFAULT_TRACKED_RESOURCE_HEADERS
 }
 
 export interface RecorderApi {
@@ -962,6 +970,7 @@ export function makeRumPublicApi(
       addTelemetryUsage({ feature: 'add-operation-step-vital', action_type: 'fail' })
       strategy.addOperationStepVital(name, 'end', options, failureReason)
     }),
+    DEFAULT_TRACKED_RESOURCE_HEADERS,
   })
 
   return rumPublicApi
--- a/packages/rum-core/src/boot/rumPublicApi.ts
+++ b/packages/rum-core/src/boot/rumPublicApi.ts
@@ -516,7 +516,7 @@ export interface RumPublicApi extends PublicApi {
   failFeatureOperation: (name: string, failureReason: FailureReason, options?: FeatureOperationOptions) => void
 
   /**
-   * List of default headers to use in the {@link RumInitConfiguration.trackResourceHeaders | trackResourceHeaders} option.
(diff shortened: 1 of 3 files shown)


thomas-lebeau (approved):

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -513,6 +514,13 @@ export interface RumPublicApi extends PublicApi {
    * @hidden // TODO: replace by @since when GA
    */
   failFeatureOperation: (name: string, failureReason: FailureReason, options?: FeatureOperationOptions) => void
+
+  /**
+   * List of default headers to use in the {@link RumInitConfiguration.trackResourceHeaders | trackResourceHeaders} option.
```
bcaudan: **💬 suggestion:** I found the phrasing a bit confusing‏

```suggestion
  /**
   * List of default headers used by the {@link RumInitConfiguration.trackResourceHeaders | trackResourceHeaders} option. See configuration example for extending them.
```

packages/rum-core/src/boot/rumPublicApi.ts
```diff
@@ -962,6 +970,7 @@ export function makeRumPublicApi(
       addTelemetryUsage({ feature: 'add-operation-step-vital', action_type: 'fail' })
       strategy.addOperationStepVital(name, 'end', options, failureReason)
     }),
+    DEFAULT_TRACKED_RESOURCE_HEADERS,
```
bcaudan, line 973: **💭 thought:** ‏Should we prevent modifications of this variable? or should we actually allow customers to update it directly?

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -268,7 +268,7 @@ export interface RumInitConfiguration extends InitConfiguration {
    * @example
    * // Collect default headers plus custom ones for all URLs
    * trackResourceHeaders: [
-   *   { name: (h) => DEFAULT_TRACKED_RESOURCE_HEADERS.includes(h) },
+   *   ...DEFAULT_TRACKED_RESOURCE_HEADERS.map((h) => ({ name: h })),
```
bcaudan, line 300: **💬 suggestion:** ‏for someone landing on this doc, it is not obvious that this variable is actually exposed from the global.

bcaudan, line 300: **💬 suggestion:** ‏it could be nice to have an e2e test to exercise this new public API.

packages/rum-core/src/boot/rumPublicApi.ts
bdibon: If a user modifies this variable this would likely be intentional and the consequences wouldn't be dramatic. 

Even if we freeze it, users can still mess with custom headers if they really want to provide weird input.

I would leave it as is.

bcaudan (approved):

PR #4525, opened 2026-04-23
⚗️ 🐛 validate feature-operation name against schema character set

Valpertui:
## Motivation

The authoritative `_vital-common-schema.json` restricts `vital.name` to the character set `[\w.@$-]*` (letters, digits, `_`, `.`, `@`, `$`, `-`) and the backend enforces this as a server-side regex on the facet path. The Browser SDK previously had no input validation on feature-operation vitals — blank names, whitespace-only names, and names with disallowed characters were all silently accepted and shipped to intake.

This PR brings the Browser SDK in line with the cross-SDK contract shared with iOS, Android, Roku, C++, and Electron

## Changes

Adds validation in `addOperationStepVital` (inside the existing `FEATURE_OPERATION_VITAL` feature-flag gate):

- **Blank / whitespace-only `name`** → `display.warn` (`"Feature operation name cannot be empty or blank. Event will not be sent."`), event dropped.
- **Name with characters outside `[\w.@$-]*`** → `display.warn` quoting the exact backend pattern, event still emitted. The backend is the source of truth on the policy, so a client-side drop would force customers to bump the SDK if the rule is ever relaxed.
- **Valid `name`** → silent emit.
- **`operation_key`** is not subject to this rule (no schema restriction).

JS `\w` is always ASCII so no special handling is needed — the non-ASCII test case `'ログイン'` pins that ASCII-only semantics are retained.

## Test instructions

```
yarn test:unit --spec packages/rum-core/src/domain/vital/vitalCollection.spec.ts
```

The new tests exercise blank-drop, disallowed-character warn+emit (including Unicode and emoji), valid silent emit, and confirm `operation_key` with spaces/slashes still emits silently.

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [x] Updated documentation and/or relevant AGENTS.md file (CHANGELOG Unreleased entry)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@
 
 ---
 
+## Unreleased
+
+**Public Changes:**
+
+- ⚗️🐛 fix(rum): validate `startFeatureOperation` / `succeedFeatureOperation` / `failFeatureOperation` `name`. Blank/empty names are dropped with a warning (matches the backend's own non-empty precondition). Names that fail the backend-accepted character-set pattern `[\w.@$-]*` also warn via `display.warn` but are still emitted, so a future backend policy relaxation does not force an SDK bump. Still gated behind the `feature_operation_vital` experimental flag. [RUM]
+
+---
+
 ## v6.32.0
 
 **Public Changes:**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,14 +13,6 @@
 
 ---
 
-## Unreleased
-
-**Public Changes:**
-
-- ⚗️🐛 fix(rum): validate `startFeatureOperation` / `succeedFeatureOperation` / `failFeatureOperation` `name`. Blank/empty names are dropped with a warning (matches the backend's own non-empty precondition). Names that fail the backend-accepted character-set pattern `[\w.@$-]*` also warn via `display.warn` but are still emitted, so a future backend policy relaxation does not force an SDK bump. Still gated behind the `feature_operation_vital` experimental flag. [RUM]
-
----
-
 ## v6.32.0
 
 **Public Changes:**
--- a/packages/rum-core/src/domain/vital/vitalCollection.spec.ts
+++ b/packages/rum-core/src/domain/vital/vitalCollection.spec.ts
@@ -271,12 +271,6 @@ describe('vitalCollection', () => {
         })
       })
 
-      // Mirrors the backend's `[\w.@$-]*` server-side validation regex. Names
-      // that fail the pattern generate a `display.warn` but the event is
-      // still emitted — the backend is the single source of truth, so a
-      // client-side drop would force a customer SDK bump if the policy is
-      // ever relaxed. Blank/empty names are dropped with a warning instead,
-      // matching the backend's own non-empty precondition.
       describe('operation name character set', () => {
         beforeEach(() => {
           addExperimentalFeatures([ExperimentalFeature.FEATURE_OPERATION_VITAL])
--- a/packages/rum-core/src/domain/vital/vitalCollection.spec.ts
+++ b/packages/rum-core/src/domain/vital/vitalCollection.spec.ts
@@ -1,6 +1,6 @@
 import type { Duration } from '@datadog/browser-core'
 import { mockClock, type Clock } from '@datadog/browser-core/test'
-import { addExperimentalFeatures, clocksNow, ExperimentalFeature, generateUUID } from '@datadog/browser-core'
+import { addExperimentalFeatures, clocksNow, display, ExperimentalFeature, generateUUID } from '@datadog/browser-core'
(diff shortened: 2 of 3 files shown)


packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -249,3 +254,30 @@ function processVital(vital: DurationVital | OperationStepVital): RawRumEventCol
     domainContext: handlingStack ? { handlingStack } : {},
   }
 }
+
+/**
+ * Validates a feature-operation `vital.name`.
+ *
```
thomas-lebeau: **🥜 nitpick:** The function name already convey this information clearly enough‏
```suggestion
```

packages/rum-core/src/domain/vital/vitalCollection.spec.ts
```diff
@@ -271,6 +271,60 @@ describe('vitalCollection', () => {
         })
       })
 
+      // Mirrors the backend's `[\w.@$-]*` server-side validation regex. Names
+      // that fail the pattern generate a `display.warn` but the event is
+      // still emitted — the backend is the single source of truth, so a
+      // client-side drop would force a customer SDK bump if the policy is
+      // ever relaxed. Blank/empty names are dropped with a warning instead,
+      // matching the backend's own non-empty precondition.
```
thomas-lebeau: **💬 suggestion:** ‏this is already documented at the implementation level, I don't think we need to repeat it here

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -249,3 +254,30 @@ function processVital(vital: DurationVital | OperationStepVital): RawRumEventCol
...
+ * Blank / empty names are rejected (the backend rejects them with its own
+ * non-empty precondition before evaluating the character-set regex). Names
+ * that fail the backend's `[\w.@$-]*` character-set regex trigger a warning
+ * but the event is still emitted — the backend is the source of truth on
+ * character-set policy, so client-side drop would force a customer SDK bump
+ * if the rule is ever relaxed.
```
thomas-lebeau, line 264: **👏 praise:** nicely thought!
> so client-side drop would force a customer SDK bump if the rule is ever relaxed.

‏

CHANGELOG.md
```diff
@@ -13,55 +13,13 @@
...
-- 🔥 remove lerna ([#4412](https://github.com/DataDog/browser-sdk/pull/4412))
-- ⚗️ [RUM-15226] Capture network headers ([#4372](https://github.com/DataDog/browser-sdk/pull/4372)) [LOGS] [RUM] [RUM-ANGULAR] [RUM-NEXTJS] [RUM-NUXT] [RUM-SLIM] [RUM-VUE] [WORKER]
-- 🧹 Sync Rum-Events-Format for GraphQL ([#4417](https://github.com/DataDog/browser-sdk/pull/4417)) [RUM] [RUM-ANGULAR] [RUM-NEXTJS] [RUM-NUXT] [RUM-SLIM] [RUM-VUE]
+## Unreleased
+
+**Public Changes:**
+
+- ⚗️🐛 fix(rum): validate `startFeatureOperation` / `succeedFeatureOperation` / `failFeatureOperation` `name`. Blank/empty names are dropped with a warning (matches the backend's own non-empty precondition). Names that fail the backend-accepted character-set pattern `[\w.@$-]*` also warn via `display.warn` but are still emitted, so a future backend policy relaxation does not force an SDK bump. Still gated behind the `feature_operation_vital` experimental flag. [RUM]
+
+---
```
thomas-lebeau: **🔨 warning:** ‏This file is generated automatically during release, it should not be edited manually (nor with AI)


```suggestion
```

Valpertui:

Valpertui:
recheck

thomas-lebeau (approved):

packages/rum-core/src/domain/vital/vitalCollection.ts
```diff
@@ -249,3 +254,28 @@ function processVital(vital: DurationVital | OperationStepVital): RawRumEventCol
...
+ *
+ * Returns `true` when the event should be emitted.
+ */
+const BACKEND_OPERATION_NAME_REGEX = /^[\w.@$-]*$/
+
+function validateOperationName(name: string): boolean {
+  if (typeof name !== 'string' || name.trim().length === 0) {
+    display.warn('Feature operation name cannot be empty or blank. Event will not be sent.')
```
bcaudan, line 272: **💬 suggestion:** ‏Similar to [what has been done on electron](https://github.com/DataDog/electron-sdk/pull/102/changes#diff-bc6c819ddb6fef87d5b180fc6a8ea0bc0618daea33151130940e74902f379065R187-R188), we should probably display an error if we won't send the event.

PR #4527, opened 2026-04-27
fix 🐛: fix regexp for cookie parsing

gogusarov:
## Motivation

Currently used helper does not take into account the parsing of cookies which have empty values (eg `"foo=; hello=world"`) which leads to several problems:

1. sdk not starting due to having a corrupt cookie (`cookie` is a default session strategy and we check if this strategy is available in [areCookiesAuthorized](https://github.com/DataDog/browser-sdk/blob/8d9036eac1d1f9c8e4856c119f67c6efd7564b3f/packages/core/src/browser/cookie.ts#L67) by tryint to set a test cookie value & then read from this cookie; this check will fail if empty cookie goes before the test cookie)
2. sdk reporting incorrect cookie values

## Changes

Fixed common utility function, updated corresponding unit test

## Test instructions

https://regex101.com/r/3K03iT/1 - buggy behavior
https://regex101.com/r/3K03iT/2 - behavior after fix

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/browser/cookie.ts
+++ b/packages/core/src/browser/cookie.ts
@@ -31,7 +31,13 @@ export function setCookie(name: string, value: string, expireDelay: number = 0, 
  * If there are multiple cookies with the same name, returns the first one
  */
 export function getCookie(name: string) {
-  return findCommaSeparatedValue(document.cookie, name)
+  const value = findCommaSeparatedValue(document.cookie, name)
+
+  if (value === '') {
+    return
+  }
+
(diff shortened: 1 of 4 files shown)


mormubis (changes requested):
Can you make a regression test?

BenoitZugmeyer (approved):

gogusarov:

mormubis (approved):

PR #4528, opened 2026-04-27
feat(core): expose createFlagEvaluationEndpointBuilder for external SDK packages

typotter:
## Problem

External packages (specifically `@datadog/openfeature-browser` in DataDog/openfeature-js-client) need to build intake endpoint URLs for the `flagevaluation` track type. Currently there is no public API for this — the package resorts to a deep private subpath import (`@datadog/browser-core/cjs/domain/configuration`) to access `createEndpointBuilder` directly.

## Solution

Add a purpose-specific public wrapper `createFlagEvaluationEndpointBuilder` instead of exporting the generic `createEndpointBuilder`.

This approach:
- Hides `TrackType` and the `'flagevaluation'` literal from external callers
- Provides a single point of change if the track type name ever changes
- Keeps `createEndpointBuilder` and `TrackType` off the public surface of `@datadog/browser-core`

## Changes

**`packages/core/src/domain/configuration/endpointBuilder.ts`**
- Updated comment on `TrackType` to reference the new public function
- Added `createFlagEvaluationEndpointBuilder` as an exported function

**`packages/core/src/domain/configuration/index.ts`**
- Added `createFlagEvaluationEndpointBuilder` to the internal barrel

**`packages/core/src/index.ts`**
- Added `createFlagEvaluationEndpointBuilder` to public exports
- Removed `createEndpointBuilder` and `TrackType` from public exports (kept internal)
- `EndpointBuilder` type remains exported so consumers can annotate variables

## Notes

- A companion PR in DataDog/openfeature-js-client (TBD) will update `@datadog/openfeature-browser` to use this function instead of the deep subpath import.
- `createEndpointBuilder` and `TrackType` remain available from the internal barrel for use by other packages in this repo.

### DIFF
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -8,6 +8,10 @@ import type { InitConfiguration } from './configuration'
 // replaced at build time
 declare const __BUILD_ENV__SDK_VERSION__: string
 
+// Note: 'flagevaluation' is intentionally not wired into TransportConfiguration or computeEndpointBuilders.
+// It is used by external SDK packages (e.g. @datadog/openfeature-browser) via the public
+// createFlagEvaluationEndpointBuilder function. Adding it here makes the intake path
+// (/api/v2/flagevaluation) an explicit part of the contract.
 export type TrackType = 'logs' | 'rum' | 'replay' | 'profile' | 'exposures' | 'flagevaluation'
 export type ApiType =
   | 'fetch'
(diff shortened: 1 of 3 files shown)


typotter:
Closing — investigation revealed flagEvaluationEndpointBuilder is already part of TransportConfiguration (since PR #3976) and available via validateAndBuildConfiguration. The explicit createEndpointBuilder call in openfeature-js-client was dead code overridden by ...baseConfiguration spread. No new public API needed.

PR #4544, opened 2026-04-29
🐛 [PANA-7217] Limit the size of the string table in session replay recordings

sethfowler-datadog:
## Motivation

The new DOM serialization algorithm in v7 of the browser SDK constructs a string table, allowing strings to be deduplicated across records. This greatly improves the space efficiency of session replay recordings.

There's a known issue with this string table: if a view lasts for a very long time, the string table can grow to be quite large. We should apply a limit to the size of the string table to ensure that we limit the amount of memory the recording process uses.

In the short term, the easiest approach is to just stop adding entries to the string table once it grows above an arbitrary maximum size, and that's what this PR implements. However, this is a bit of a bummer, because it means that views that hit this threshold will not benefit from the string table as much as they otherwise could. It would likely be better to reset the string table when it gets too large, repopulating it from that point. That change is a bit more complicated, since it requires a tweak to the data format (we should be explicit about when the reset happens to ensure that the recorder and the player agree), so I don't want to start with it, but I plan to revisit this issue in the near future.

## Changes

This PR adds a `StringIdConstants.SOFT_MAX_SIZE` constant defining the size limit for the string table. This limit is set arbitrarily for now; I plan to add telemetry to gather data about how large these string tables are in practice, and I'll adjust the limit in the future based on that data.

The size limit is enforced by `ChangeEncoder`, which is the infrastructure that actually replaces strings in the recorded event stream with references to the string table. The data format was designed to allow either a literal string or a string table reference in any position where a string can be used, so it's fine if `ChangeEncoder` simply declines to replace strings once the size limit is hit. Even if the string table is maxed out, a string can still be replaced with a string table reference if the string is already in the table, so we still get value from the string table even in this situation.

## Checklist

- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/packages/rum/src/domain/record/serialization/changeEncoder.spec.ts
+++ b/packages/rum/src/domain/record/serialization/changeEncoder.spec.ts
@@ -141,6 +141,45 @@ describe('ChangeEncoder', () => {
     })
   })
 
+  describe('soft max size of the string table', () => {
+    // Simulates a full string table without actually inserting a million entries.
+    const simulateFullStringTable = () => {
+      Object.defineProperty(stringIds, 'size', { get: () => 1_000_000, configurable: true })
+    }
+
+    it('does not add new strings to the table once the soft max size is reached', () => {
+      simulateFullStringTable()
+
+      encoder.add(ChangeType.Text, [0, 'new-string'])
+      const changes = encoder.flush()
+
+      // The new string remains as a literal in the change data and no AddString is emitted.
+      expect(changes).toEqual([[ChangeType.Text, [0, 'new-string']]])
+    })
+
+    it('still reuses existing string ids once the soft max size is reached', () => {
+      const existingId = stringIds.getOrInsert('existing')
+      simulateFullStringTable()
+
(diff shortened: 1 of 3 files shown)


packages/rum/src/domain/record/serialization/changeEncoder.spec.ts
```diff
@@ -141,6 +141,45 @@ describe('ChangeEncoder', () => {
     })
   })
 
+  describe('soft max size of the string table', () => {
+    // Simulates a full string table without actually inserting a million entries.
+    const simulateFullStringTable = () => {
+      Object.defineProperty(stringIds, 'size', { get: () => 1_000_000, configurable: true })
```
BenoitZugmeyer: **nitpick:** 
```suggestion
      Object.defineProperty(stringIds, 'size', { get: () => StringIdConstants.SOFT_MAX_SIZE, configurable: true })
```

BenoitZugmeyer (approved):

mormubis (approved):

PR #4566, opened 2026-05-06
♻️ [Profiler] Remove sample-count threshold from profiler discard logic

thomasbertet:
## Motivation

Short-lived non-SPA navigations (full-page loads with moderate scripting) were being silently discarded by the profiler because they accumulated fewer than 50 stack-trace samples (~500ms of scripting activity), even when the page was active long enough to meet the duration threshold. This caused an empty \"No profiles\" state in the Profiling UI for use cases like shopist.io where users click quickly across pages.

Additionally, filtering on sample count introduced a selection bias in aggregations: only CPU-heavy pages were retained, skewing flame graphs away from scripting-light but potentially slow navigations.

The trade-off here is collecting more profiles for a slightly better coverage of short-lived navigations. 
Nonetheless, we keep discarding profiles <5s if they don't have Long Tasks in them.

## Changes

- Removed the `minNumberOfSamples` check from the profile discard logic in `profiler.ts`. Profiles are now kept or discarded based solely on duration (`minProfileDurationMs`) and long-task presence.
- Removed the `minNumberOfSamples` field from `RUMProfilerConfiguration` and `DEFAULT_RUM_PROFILER_CONFIGURATION`.
- Added unit tests covering the three discard cases: below duration threshold with no long tasks (discarded), below duration threshold with a long task (sent), and above duration threshold (sent).

## Test instructions

1. Load a non-SPA page with the RUM SDK and profiler enabled.
2. Navigate away quickly (within a few seconds) before reaching the 5s duration threshold.
3. Verify a profile is sent if a long task was captured, discarded otherwise.
4. Run `yarn test:unit --spec packages/rum/src/domain/profiling/profiler.spec.ts` — 20 tests should pass.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging


thomas-lebeau (commented):
Can you edit the PR title to follow [our convention](https://github.com/DataDog/browser-sdk/blob/main/docs/DEVELOPMENT.md#commit-messages-and-pull-request-titles). This help generate clean changelog and release notes.

other than that, LGTM 👍

thomas-lebeau (approved):

PR #4569, opened 2026-05-06
🐛 Drain SDK idle-callback queue before flushing in e2e tests

thomas-lebeau:
## Motivation

E2E tests (notably the microfrontend fetch/XHR scenarios) were flaky because the SDK defers some bookkeeping — particularly resource event emission — onto an idle-callback task queue. The previous `waitForRequests` helper just slept 200ms via `setTimeout`, which was not always enough for those deferred tasks to run before the test asserted on captured events.

## Changes

- Replace the fixed 200ms `setTimeout` in `waitForRequests` with two `requestIdleCallback` round-trips to drain the SDK's idle-callback queue (including tasks enqueued by the first batch).
- Keep a 500ms watchdog as a fallback for environments where `requestIdleCallback` is throttled or unavailable (e.g. the empty `/flush` page during teardown).

## Test instructions

- Run the E2E suite: `yarn test:e2e`
- Re-run microfrontend fetch/XHR scenarios multiple times to confirm the flake no longer reproduces.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/test/e2e/lib/framework/waitForRequests.ts
+++ b/test/e2e/lib/framework/waitForRequests.ts
@@ -21,16 +21,9 @@ export async function waitForRequests(page: Page) {
         // first batch of tasks are also processed. A 500ms watchdog covers pages where
         // requestIdleCallback is throttled or unavailable (e.g. the empty /flush page during
         // teardown).
-        let done = false
-        const finish = () => {
-          if (!done) {
-            done = true
-            resolve()
-          }
-        }
-        setTimeout(finish, 500)
+        setTimeout(resolve, 500)
         const ric = window.requestIdleCallback?.bind(window) ?? ((cb: IdleRequestCallback) => setTimeout(cb, 50))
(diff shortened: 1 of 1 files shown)


test/e2e/lib/framework/waitForRequests.ts
```diff
@@ -8,16 +8,29 @@ import { waitForServersIdle } from './httpServers'
...
+        // first batch of tasks are also processed. A 500ms watchdog covers pages where
+        // requestIdleCallback is throttled or unavailable (e.g. the empty /flush page during
+        // teardown).
+        let done = false
+        const finish = () => {
+          if (!done) {
+            done = true
+            resolve()
+          }
+        }
```
BenoitZugmeyer: **suggestion:** this is unnecessary, just use `resolve` instead of `finish`: it doesn't matter if `resolve` is called twice

BenoitZugmeyer (approved):

mormubis (approved):

PR #4574, opened 2026-05-07
feat(graphql): extract operationType from the req. body

### LINKED ISSUE #4563: Providing operationType for persisted GraphQL operations
For persisted operations `query` value in the request body does not contain the `query|mutation|subscription` value.

It'd great if the GraphQL integration can check for a `operationType` property in the request body, so providing one could be used to populate the respective field in the resource event (similar to `operationName`)?

I can open a PR for it, wanted to get the initial impression of the idea beforehand.

mortargrind:
Resolves: #4563 

## Motivation

Persisted GraphQL operations do not have a `query` property that contains the full operation string. As a result `operationType` cannot be extracted for them.

## Changes

An alternative way of extracting `operationType` from a dedicated `operationType` param is added. The current way of extracting the value has higher precedence over this new alternative.

## Test instructions

Manual testing: If a GraphQL request has `operationType` in its body or in query params as `'query'|'mutation'|'subscription'` and `operationType` cannot be extracted from the `query` property.

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/resource/graphql.spec.ts
+++ b/packages/rum-core/src/domain/resource/graphql.spec.ts
@@ -267,6 +267,91 @@ describe('GraphQL detection and metadata extraction', () => {
       )
       expect(result).toBeUndefined()
     })
+
+    it('should use operationType from body when query field is absent', () => {
+      const requestBody = JSON.stringify({
+        operationName: 'CreateUser',
+        operationType: 'mutation',
+        variables: { name: 'Alice' },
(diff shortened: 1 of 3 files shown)


mortargrind:

rgaignault (commented):
LGTM

rgaignault (approved):

BenoitZugmeyer (approved):

PR #4605, opened 2026-05-11
🐛 await re-grant if consent is revoked during session manager init

thomas-lebeau:
## Motivation

Addresses [a P2 review comment](https://github.com/DataDog/browser-sdk/pull/4425#discussion_r3218615540) on #4425.

When tracking consent is revoked while `startSessionManager` is awaiting the cookie lock / Cookie Store call (the "async initialization" window), the previous code called `expire()` and resolved with `undefined`. The RUM and Logs pre-start strategies (`preStartRum.ts:180`, `preStartLogs.ts:109`) invoke `startSessionManager` from `trackingConsentState.onGrantedOnce`, which fires **exactly once** and then unsubscribes (`trackingConsent.ts:30-41`). As a result, a customer who re-granted consent later in the same page could **never** bring the SDK back — the session manager was permanently stuck in pre-start state.

Note: once the manager is installed, `setupSessionTracking` already handles revoke/grant cycles correctly. The bug is strictly the gap between `startSessionManager` starting and `setupSessionTracking` subscribing.

## Changes

- `packages/core/src/domain/session/sessionManager.ts`: replace the early-return-on-revoked-consent branch with a wait loop that mirrors `setupSessionTracking`'s handler — call `expire()` to mark the cookie expired during the gap (so other tabs don't pick up an orphan session ID), wait for the next grant via `onGrantedOnce`, then re-resolve the initial state before installing the manager. If the manager is stopped while waiting, exit cleanly on the next grant.
- `packages/core/src/domain/session/sessionManager.spec.ts`:
  - Replace the previous "resolves with undefined on revocation during init" test (which codified the old behavior) with one verifying the promise **stays pending** while consent stays revoked.
  - Add a test verifying the manager **does** install once consent is granted again after being revoked mid-init.
  - Add a test verifying `expire()` is called (additional `setSessionState` write) while consent stays revoked.
  - Add a `flushMicrotasks` helper (10 cycles) so chained awaits — `setSessionState` resolution, the `.catch(monitorError)` chain, and the post-await continuation — all run before assertions.
- `test/e2e/scenario/trackingConsent.scenario.ts`: add an E2E test reproducing the customer-facing scenario — init with `trackingConsent: 'not-granted'`, synchronously grant-then-revoke (the revoke lands inside the async init window), wait, then re-grant. Asserts the cookie is marked expired during the revoked gap and that RUM view events are captured after the re-grant.

## Test instructions

- `yarn test:unit --spec packages/core/src/domain/session/sessionManager.spec.ts` — the three tests under `tracking consent` pass. Reverting `sessionManager.ts` makes them fail.
- `yarn test:e2e -g "recovers when consent is revoked during session manager init"` — the new E2E test passes; reverting the fix makes it fail (no RUM view events captured after re-grant).
- `yarn test:unit` — full suite stays green (4287 tests).
- `yarn typecheck` / `yarn lint` clean.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


thomas-lebeau:
@codex review

packages/core/src/domain/session/sessionManager.spec.ts
```diff
@@ -31,6 +31,15 @@ import type { SessionStoreStrategyType } from './storeStrategies/sessionStoreStr
 import type { SessionState } from './sessionState'
 import { EXPIRED } from './sessionState'
 
+// Flush several microtask cycles so chained awaits inside startSessionManager
+// (resolveInitialState's await, the .catch chain, and the post-await continuation)
+// have a chance to run before assertions.
+async function flushMicrotasks(): Promise<void> {
```
gogusarov, line 37: `question`

you decided not to go for classic 
```
const flushMicrotasks = (): Promise<void> => new Promise((resolve) => {
    setTimeout(() => {
      resolve();
    })
  })

```
because of `mockClock` overhead?

packages/core/src/domain/session/sessionManager.ts
```diff
@@ -132,9 +132,18 @@ export async function startSessionManager(
 
   // Consent is always granted when the session manager is started, but it may
   // be revoked during the async initialization (e.g., while waiting for cookie lock).
-  if (!trackingConsentState.isGranted()) {
+  // Mirror setupSessionTracking's revoke/grant handler until the manager is installed:
+  // expire the session in storage, wait for the next grant, then re-resolve.
+  while (!trackingConsentState.isGranted()) {
```
gogusarov, line 137: `question`

cycle here and not a single `if` to protect from another possible consent revocation while `resolveInititialState` is awaited?

thomas-lebeau: if the consent is revoked while we await for `resolveInitialState`, then the while condition is true, so we expire the session (started from`resolveInitialState`) and await for consent again.

packages/core/src/domain/session/sessionManager.spec.ts
thomas-lebeau: yes, that's correct

gogusarov (approved):

BenoitZugmeyer (approved):

PR #4606, opened 2026-05-11
✨ implement remote configuration async loading & caching

gogusarov:
## Motivation

Support remote configuration asynchronous loading & caching mechanism according to the [RFC](https://datadoghq.atlassian.net/wiki/spaces/~71202051361f9d78364c1fbe5093e2c942b83f/pages/6579224798/RFC+Stale-While-Revalidate+Caching+for+Remote+Configuration)

## Changes

Added async remote configuration fetching flow where config isn't awaited during `DD_RUM.init` but read synchronously from a versioned localStorage cache (`dd_rc_<remoteConfigurationId>`) and applied on the same page load. A background fetch runs in parallel and refreshes the cache for the next page load, so updates roll out one load late but startup is unblocked. Cache outcomes (hit/miss/error) and fetch outcomes (success/failure) are reported via the existing telemetry metrics. Unit tests cover the new cache module and the read/apply/fetch flow; the e2e scenarios were rewritten to wait for the cache to be populated and reload the page before asserting.

Blocking fetch flow was kept for backward compatibility with existing customer setup

 ### New init configuration API

  - `remoteConfiguration: { id: string; sync?: boolean }` — new top-level field.
  Default behavior is async (cache-and-reload). Pass `sync: true` to opt into the
  legacy blocking fetch through the new API.
  - `remoteConfigurationId: string` — legacy field preserved for backward
  compatibility with existing customer setups. Always triggers the blocking fetch
  flow.

## Test instructions

1. add `remote-config.json` file to sandbox directory
2. update parameters, passed to `DD_RUM.init` adding `remoteConfiguration: { id: 'remote-config-test' }`, `remoteConfigurationProxy: '/remote-config.json'`
3. check that remote config from file is requested and written to localStorage on the first page load (localStorage.getItem('dd_rc_remote-config-test')
4. check that `DD_RUM.getInitConfiguration()` returns initial configuration on the first page load
5. check that `DD_RUM.getInitConfiguration()` returns remote configuration on the second page load

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/configuration/configuration.ts
+++ b/packages/rum-core/src/domain/configuration/configuration.ts
@@ -16,6 +16,7 @@ import type { RumEventDomainContext } from '../../domainContext.types'
 import type { RumEvent } from '../../rumEvent.types'
 import type { RumPlugin } from '../plugins'
 import type { PropagatorType, TracingOption } from '../tracing/tracer.types'
+import { getRemoteConfigurationId } from './remoteConfiguration'
 
 export const DEFAULT_PROPAGATOR_TYPES: PropagatorType[] = ['tracecontext', 'datadog']
 
@@ -676,7 +677,7 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
       ...plugin.getConfigurationTelemetry?.(),
     })),
     track_feature_flags_for_events: configuration.trackFeatureFlagsForEvents,
-    remote_configuration_id: configuration.remoteConfigurationId ?? configuration.remoteConfiguration?.id,
+    remote_configuration_id: getRemoteConfigurationId(configuration),
     profiling_sample_rate: configuration.profilingSampleRate,
     use_remote_configuration_proxy: !!configuration.remoteConfigurationProxy,
     track_resource_headers: getTrackResourceHeadersTelemetryValue(configuration.trackResourceHeaders),
--- a/packages/rum-core/src/domain/configuration/configuration.ts
+++ b/packages/rum-core/src/domain/configuration/configuration.ts
@@ -138,11 +138,21 @@ export interface RumInitConfiguration extends InitConfiguration {
   compressIntakeRequests?: boolean | undefined
 
   /**
-   * [Internal option] Id of the remote configuration
+   * [Internal option] Id of the remote configuration.
+   * Prefer `remoteConfiguration.id` for the non-blocking cache-and-reload path.
    *
    * @internal
    */
   remoteConfigurationId?: string | undefined
+
+  /**
+   * [Internal option] Remote configuration descriptor. By default the SDK reads a cached
+   * configuration synchronously and refreshes it in the background. Set `sync: true` to fall back
+   * to the legacy blocking fetch.
+   *
+   * @internal
+   */
+  remoteConfiguration?: { id: string; sync?: boolean } | undefined
 
   /**
    * [Internal option] set a proxy URL for the remote configuration
@@ -666,7 +676,7 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
       ...plugin.getConfigurationTelemetry?.(),
     })),
     track_feature_flags_for_events: configuration.trackFeatureFlagsForEvents,
-    remote_configuration_id: configuration.remoteConfigurationId,
+    remote_configuration_id: configuration.remoteConfigurationId ?? configuration.remoteConfiguration?.id,
     profiling_sample_rate: configuration.profilingSampleRate,
     use_remote_configuration_proxy: !!configuration.remoteConfigurationProxy,
     track_resource_headers: getTrackResourceHeadersTelemetryValue(configuration.trackResourceHeaders),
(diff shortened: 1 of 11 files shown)


packages/rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -312,3 +294,45 @@ export function buildEndpoint(configuration: RumInitConfiguration) {
...
+  supportedContextManagers: SupportedContextManagers
+): RumInitConfiguration {
+  const configurationCache = createConfigurationCache({
+    remoteConfigurationId: initConfiguration.remoteConfigurationId!,
+  })
+  const metrics = initMetrics()
+
+  const cacheResult = configurationCache.read()
+
+  metrics.increment('cache', CACHE_STATUS_TO_METRIC_MAP[cacheResult.status])
```
gogusarov, line 360: @mormubis would it be ok to leave metrics keys the way they currently are (`success` / `missing` / `failure`) or should they be exactly the same as in your [RFC](https://datadoghq.atlassian.net/wiki/spaces/~71202051361f9d78364c1fbe5093e2c942b83f/pages/6579224798/RFC+Stale-While-Revalidate+Caching+for+Remote+Configuration) (`hit`/`miss`/`error`)

test/e2e/scenario/rum/remoteConfiguration.scenario.ts
```diff
@@ -47,7 +48,7 @@ test.describe('remote configuration', () => {
         version: { rcSerializedType: 'dynamic', strategy: 'dom', selector: '#version' },
       },
     })
-    .withBody(html`<span id="version">123</span>`)
+    .withHead(html`<span id="version">123</span>`)
```
gogusarov: I updated these tests using `withHead` so that they pass, but there’s an issue here that we should definitely address later

Remote Configuration currently supports dynamic options, one of which is `dom`. If we use synchronous bundle setup and Remote Configuration is loaded from the cache immediately, this strategy will always fail because SDK initialization happens before the DOM is available

thomas-lebeau: That's a good point, but it might break existing customer setup if we don't address this now, no?

We can change the test, but we can't change our customer setup

test/e2e/scenario/rum/remoteConfiguration.scenario.ts
```diff
@@ -231,13 +204,11 @@ test.describe('remote configuration', () => {
     })
 })
 
+/* The background fetch on the initial page load writes the remote configuration into localStorage;
+ * reloading lets the SDK pick it up synchronously on the next init().
```
thomas-lebeau: **💬 suggestion:** I think with your change, all the test are now relying on the config being cached and we don't test that the remote config works properly even without cache.

I would rather keep `waitForRemoteConfigurationToBeApplied` as before and add a separate test specifically for the caching mechanism.

test/e2e/scenario/rum/remoteConfiguration.scenario.ts
gogusarov: It seems like it's not something we should solve immediately because currently remote config is only available for SSI clients and from what I know they have hardcoded v6 version and they can't update on their own until we update the injector

thomas-lebeau:
can you edit the PR title to follow our [convention](https://github.com/DataDog/browser-sdk/blob/main/docs/DEVELOPMENT.md#commit-messages-and-pull-request-titles) please!

packages/rum-core/src/domain/configuration/remoteConfiguration.ts
mormubis: I'm ok with any metrics. These are perfect.

packages/rum-core/src/domain/configuration/configuration.ts
```diff
@@ -666,7 +676,7 @@ export function serializeRumConfiguration(configuration: RumInitConfiguration) {
       ...plugin.getConfigurationTelemetry?.(),
     })),
     track_feature_flags_for_events: configuration.trackFeatureFlagsForEvents,
-    remote_configuration_id: configuration.remoteConfigurationId,
+    remote_configuration_id: configuration.remoteConfigurationId ?? configuration.remoteConfiguration?.id,
```
mormubis: `remote_configuration_id` resolves the ID in the opposite order from `getRemoteConfigurationId`. I'd use the helper here too:

```ts
remote_configuration_id: getRemoteConfigurationId(configuration),
```

Nobody would pass both options, but this way there's a single source of truth for the resolution. What do you think?

mormubis (approved):

packages/rum-core/src/domain/configuration/configuration.ts
gogusarov: agree, will fix

gogusarov: @datadog fix this

gogusarov: bruh

PR #4613, opened 2026-05-13
✅ fix flaky cookieAccess tests in slow CI

watson:
## Motivation

The `document.cookie fallback` variant of `cookieAccess.spec` is occasionally flaky on CI. The failure mode is `should notify when cookie is changed externally` reporting:

> Expected spy observer to have been called once. It was called 0 times.

The test is unrelated to any specific feature work — it can be reproduced from `main` on slow CI runs.

## Changes

The root cause is a real-time / mock-time mismatch. When `mockClock()` is installed, `new Date()` returns the mocked time. `setCookie()` computes the cookie's expiry from `new Date()` + `expireDelay`, but since `mockClock()` initializes the mocked time to the current real time, the absolute expiry the browser sees is `REAL_T_install + expireDelay`. The browser then expires cookies against **real** wall-clock time — it doesn't know anything about the mock.

With the previous `expireDelay = 1000` ms, any CI run slow enough for >1 s of real time to elapse between `mockClock()` and the polling interval reading `document.cookie` would see the cookie already expired. `getCookies()` returned `[]`, matched `previousCookieValues = []` captured at construction, and the observable never notified.

Fix in `packages/core/src/browser/cookieAccess.spec.ts`: bump the test cookie `expireDelay` from `1000` ms to `ONE_MINUTE` at every call site, leaving real time comfortable headroom to elapse during a slow test without the cookie disappearing.

No production code changes — this is test-only.

## Test instructions

To stress the random ordering (including the originally failing seed `00743`):

```bash
for seed in 00001 00100 00743 12345 99999 50000; do
  yarn test:unit --spec packages/core/src/browser/cookieAccess.spec.ts --seed $seed
done
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/browser/cookieAccess.spec.ts
+++ b/packages/core/src/browser/cookieAccess.spec.ts
@@ -2,7 +2,7 @@ import type { Clock } from '../../test'
 import { collectAsyncCalls, mockClock, registerCleanupTask, replaceMockable } from '../../test'
 import type { Configuration } from '../domain/configuration'
 import { detectVersion, isChromium } from '../tools/utils/browserDetection'
-import { dateNow } from '../tools/utils/timeUtils'
+import { dateNow, ONE_MINUTE } from '../tools/utils/timeUtils'
 import type { CookieOptions } from './cookie'
 import { deleteCookie, getCookie, setCookie } from './cookie'
 import type { CookieStoreWindow } from './browser.types'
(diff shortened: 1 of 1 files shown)


watson:
* **#4613** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4613?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4613?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/core/src/browser/cookieAccess.spec.ts
```diff
@@ -33,6 +33,7 @@ describe('cookieAccess', () => {
     {
       title: 'document.cookie fallback',
       setup: () => {
+        registerCleanupTask(() => deleteCookie(COOKIE_NAME))
```
thomas-lebeau: Supposedly we already delete all cookies in the [global forEach](https://github.com/DataDog/browser-sdk/blob/9b44f2f874ba41db3a9fbf3627053b7804579ac8/packages/core/test/forEach.spec.ts#L25-L25)

If this is not working properly we should probably fix it there, no?

p-datadog (approved):

packages/core/src/browser/cookieAccess.spec.ts
watson: Good point. I removed those again

thomas-lebeau (approved):

PR #4617, opened 2026-05-13
✨ Add operation's start vital's id to profiles metadata

MandragoreVR:
## Motivation

As part of the Custom profiling initiative, custom profiles will be linked to Operation events. Since on the SDK side, Operation events are represented by a start vital and a stop vital, we decided (and this is already implemented as such in the iOS SDK) to tag the profiles overlapping with an operation with the start vital's id of that operation.

This PR implements this change.

## Changes

- Rework `vitalHistory`
	- Add vital `type` and `operationKey` to the data stored in the history
	- Nothing should have changed for duration vitals
	- For operation step vitals:
		- if we are receiving a start vital, we store it with an `undefined` duration to represent that it's still active
		- if we are receiving a stop vital, we look for the matching start vital. If it exists, we update the start vital with the duration, and close it in the history.
- Added `findAllEntries` to `valueHistory`, that is the same as `findAll` but returns the entries instead of the values.
- Since we now have the vital type in the vital history, I renamed the life cycle event to start duration vitals to something clearer.
- Added some tests 🧪

## Test instructions - Tested locally

When the operation is not over, here's what it looks like in the profile:

Screenshot 2026-05-13 at 14 58 14

And when it's over, same but with a duration:

Screenshot 2026-05-13 at 14 58 18

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum/src/domain/profiling/vitalHistory.ts
+++ b/packages/rum/src/domain/profiling/vitalHistory.ts
@@ -1,13 +1,15 @@
 import type { ClocksState, Duration } from '@datadog/browser-core'
-import { addDuration, createValueHistory, SESSION_TIME_OUT_DELAY } from '@datadog/browser-core'
+import { addDuration, createValueHistory, elapsed, SESSION_TIME_OUT_DELAY } from '@datadog/browser-core'
 import type { LifeCycle } from '@datadog/browser-rum-core'
-import { LifeCycleEventType } from '@datadog/browser-rum-core'
+import { LifeCycleEventType, VitalType } from '@datadog/browser-rum-core'
 
 export const VITAL_ID_HISTORY_TIME_OUT_DELAY = SESSION_TIME_OUT_DELAY
 
 export interface VitalContext {
   id: string
+  type: VitalType
   label: string
+  operationKey?: string
   duration?: Duration
   startClocks: ClocksState
 }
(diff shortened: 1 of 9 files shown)


BeltranBulbarellaDD:
@codex review?

BeltranBulbarellaDD (approved):

thomas-lebeau:
@MandragoreVR can you edit the PR title to follow our [convention](https://github.com/DataDog/browser-sdk/blob/main/docs/DEVELOPMENT.md#commit-messages-and-pull-request-titles) please!

packages/rum/src/domain/profiling/vitalHistory.ts
```diff
@@ -31,18 +34,49 @@ export function createVitalHistory(lifeCycle: LifeCycle) {
...
     if (rawRumEvent.type === 'vital') {
-      const historyEntry = history
-        .getEntries(startClocks.relative)
-        .find((entry) => entry.value.id === rawRumEvent.vital.id)
-
-      if (historyEntry) {
-        historyEntry.value.duration = duration!
-        historyEntry.close(addDuration(startClocks.relative, duration!))
+      if (rawRumEvent.vital.type === VitalType.OPERATION_STEP) {
+        if (rawRumEvent.vital.step_type === 'end') {
```
thomasbertet: kinda nit, but we could try to minimize cascading of `if`s, if we break sooner.

for instance:  ```if (rawRumEvent.type === 'vital') { ... }  else { ... }```

could be replaced with `if (rawRumEvent.type != 'vital') { return } 

The current logic is slightly hard to reason about (and read). Few comments may also help.

thomasbertet (approved):

packages/rum/src/domain/profiling/vitalHistory.ts
MandragoreVR: Good point indeed, I tried to improve the readability and added some comments [here](https://github.com/DataDog/browser-sdk/pull/4617/commits/11f2c37a349f0e4e16a417fe855ed3c3a088ae8b)!

PR #4622, opened 2026-05-14
🎨 Gate view_update events behind datadogMode in developer extension

mormubis:
## Motivation

Follow-up from #4319. `view_update` events are an implementation detail of partial view updates. Customers don't need to see them when debugging.

The full `view` event is still sent to the extension on every intermediate update via `RUM_EVENT_COLLECTED`, so views stay up to date in the list.

## Changes

The "Show only the latest View event" checkbox now controls `view_update` visibility:
- **Checked (default):** only the latest view is shown, no `view_update` events, no intermediate views
- **Unchecked:** all events shown, but if a `view_update` exists for a specific `view.id + document_version`, the corresponding full view is skipped to avoid redundancy

The checkbox itself is hidden behind `datadogMode`. Customers always get the default behavior.

**Known limitation:** the facet sidebar counts `view_update` events even when they are filtered from the list. This is a pre-existing limitation of how facet counts are computed from all collected events, not from the filtered set.

## Test instructions

1. Enable `partial_view_updates` experimental feature and open the developer extension
2. With datadogMode off: `view_update` events don't appear, `view` events update normally
3. Enable datadogMode, checkbox checked: same as above
4. Enable datadogMode, uncheck "Show only the latest View event": `view_update` diffs appear, corresponding full views are hidden

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [ ] Added unit tests for this change

### DIFF
--- a/developer-extension/src/panel/components/tabs/eventsTab/eventsTabSide.tsx
+++ b/developer-extension/src/panel/components/tabs/eventsTab/eventsTabSide.tsx
@@ -1,22 +1,23 @@
 import { Box, Checkbox } from '@mantine/core'
 import React from 'react'
 import type { EventFilters, FacetRegistry } from '../../../hooks/useEvents'
+import { useSettings } from '../../../hooks/useSettings'
 import { FacetList } from './facetList'
 
 export function EventsTabSide({
   facetRegistry,
   filters,
   onFiltersChange,
-  datadogMode,
 }: {
   facetRegistry?: FacetRegistry
   filters: EventFilters
   onFiltersChange: (filters: EventFilters) => void
-  datadogMode: boolean
 }) {
+  const [settings] = useSettings()
+
   return (
     <Box ml="md">
-      {datadogMode && (
+      {settings.datadogMode && (
         <Checkbox
           label={'Show only the latest View\xa0event'}
           checked={!filters.outdatedVersions}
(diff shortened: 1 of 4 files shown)


thomas-lebeau (commented):
shouldn't this use "show only the last view Event"?
- if it's checked, the view update are merged into the view event.
- if it's no checked, then we show each view and view update events.

Then if we want we can hide this checkbox when in non-datadog mode.

we need to make sure to keep the compatibility with both mechanism (old _view update_ and the new one)


Image

mormubis:
> shouldn't this use "show only the last view Event"?
> 
> * if it's checked, the view update are merged into the view event.
> * if it's no checked, then we show each view and view update events.
> 
> Then if we want we can hide this checkbox when in non-datadog mode.
> 
> we need to make sure to keep the compatibility with both mechanism (old _view update_ and the new one)
> 
> Image

@thomas-lebeau Updated! The "Show only the latest View event" checkbox now controls view_update visibility. When checked (default), only the latest view is shown, no view_updates. When unchecked, both show, but if a view_update exists for a specific document_version, the corresponding full view is skipped to avoid redundancy. The checkbox itself is hidden behind datadogMode. What do you think?

developer-extension/src/panel/components/tabs/eventsTab/eventsTabSide.tsx
```diff
@@ -7,19 +7,23 @@ export function EventsTabSide({
   facetRegistry,
   filters,
   onFiltersChange,
+  datadogMode,
 }: {
   facetRegistry?: FacetRegistry
   filters: EventFilters
   onFiltersChange: (filters: EventFilters) => void
+  datadogMode: boolean
```
BenoitZugmeyer: **suggestion:** use `useSettings`, no need to drill down settings values

BenoitZugmeyer (approved):

gogusarov (approved):

PR #4624, opened 2026-05-14
♻️ detect cookie API availability when selecting session store strategy

thomas-lebeau:
## Motivation

In WebKit (Safari) on `localhost`-based subdomains, `window.cookieStore` is exposed and `cookieStore.set(...)` resolves without throwing — but the cookie is silently dropped because `localhost` is not a registrable domain per [RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761#section-6.3) (Safari rejects explicit `domain=` attributes for non-registrable hosts). This breaks session persistence end-to-end on the latest WebKit in our e2e setup (which uses `foo.bar.localhost`), even though `document.cookie` writes still succeed.

Previously, the SDK preferred the CookieStore API whenever `window.cookieStore` was present, with no end-to-end verification that a cookie actually round-trips. This change makes the session store strategy probe each API by writing and reading back a test cookie, and records the chosen API on the strategy so the runtime uses the same one that was verified during initialization.

## Changes

- Move `areCookiesAuthorized` into `cookieAccess.ts` and make it async, taking a `CookieAccess` factory so it can verify each API end-to-end.
- Split `createCookieAccess` into `createCookieStoreAccess` and `createDocumentCookieAccess`.
- `selectCookieStrategy` probes the CookieStore API first and falls back to `document.cookie`, recording the chosen `cookieApi` on the strategy type.
- Thread `cookieApi` through `CookieSessionStoreStrategyType` and `initCookieStrategy` so the runtime strategy uses the probed API.
- `selectSessionStoreStrategyType` is now async and takes `Configuration` (with `cookieOptions`/`sessionPersistence`); `sessionManager` awaits it during startup.
- Drop `areCookiesAuthorized` from the public core index export.
- CI: run latest Firefox and WebKit alongside Chromium; default local `yarn test:e2e` to Chromium for faster feedback.

## Test instructions

- Reproduce the original WebKit failure (before this change) and verify it's fixed by running the sessionStore e2e suite on WebKit:
  ```
  yarn test:e2e:ci --project=webkit -g sessionStore
  ```
  Expected: 11 passed, 1 skipped (the firefox-only iframe test).
- Run unit tests: `yarn test:unit --spec packages/core/src/browser/cookieAccess.spec.ts` and `yarn test:unit --spec packages/core/src/domain/session/sessionStore.spec.ts`.
- Verify in a browser with `cookieStore` available (e.g. Chromium) that the SDK selects the cookieStore API; in WebKit on localhost, verify it falls back to `document.cookie`.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/session/sessionManager.spec.ts
+++ b/packages/core/src/domain/session/sessionManager.spec.ts
@@ -25,14 +25,19 @@ import {
   TRACKED_SESSION_MAX_AGE,
   VISIBILITY_CHECK_DELAY,
 } from './sessionManager'
-import { getSessionStoreStrategy } from './sessionStore'
+import { getSessionStoreStrategy, selectSessionStoreStrategyType } from './sessionStore'
 import { SESSION_EXPIRATION_DELAY, SESSION_TIME_OUT_DELAY, SessionPersistence } from './sessionConstants'
 import type { SessionStoreStrategyType } from './storeStrategies/sessionStoreStrategy'
+import { CookieApi } from './storeStrategies/sessionStoreStrategy'
 import type { SessionState } from './sessionState'
 import { EXPIRED } from './sessionState'
 
 describe('startSessionManager', () => {
-  const STORE_TYPE: SessionStoreStrategyType = { type: SessionPersistence.COOKIE, cookieOptions: {} }
+  const STORE_TYPE: SessionStoreStrategyType = {
+    type: SessionPersistence.COOKIE,
+    cookieOptions: {},
+    cookieApi: CookieApi.DOCUMENT_COOKIE,
+  }
   let fakeStrategy: ReturnType<typeof createFakeSessionStoreStrategy>
   let clock: Clock
   let sessionObservableSpy!: jasmine.Spy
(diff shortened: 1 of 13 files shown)


packages/core/src/index.ts
```diff
@@ -102,14 +102,7 @@ export {
...
 export type { Context, ContextArray, ContextValue } from './tools/serialisation/context'
-export {
-  areCookiesAuthorized,
-  getCookie,
-  getInitCookie,
-  setCookie,
-  deleteCookie,
-  resetInitCookies,
-} from './browser/cookie'
+export { getCookie, getInitCookie, setCookie, deleteCookie, resetInitCookies } from './browser/cookie'
```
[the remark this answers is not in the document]
thomas-lebeau: This is fine, people are not supposed to use `@datadog/browser-core` and I did not find any other package in the dataDog github org that uses it.

packages/core/src/domain/session/sessionManager.spec.ts
```diff
@@ -537,13 +546,15 @@ describe('startSessionManager', () => {
...
       const sessionManagerPromise = startSessionManager(
         {
-          sessionStoreStrategyType: STORE_TYPE,
           sessionSampleRate: 100,
           trackAnonymousUser: false,
         } as Configuration,
         trackingConsentState
       )
 
+      // Allow startSessionManager to await selectSessionStoreStrategyType and call setSessionState
```
gogusarov, line 556: `minor`

maybe it's time to add smth like `flushPromises` helper to test utils? I didn't find anything similar

gogusarov (approved):

packages/core/src/domain/session/sessionManager.spec.ts
thomas-lebeau: you're right, looks like we have `waitNextMicrotask()` already, so using that.

gogusarov (approved):

packages/core/src/domain/session/storeStrategies/sessionStoreStrategy.ts
```diff
@@ -6,8 +6,20 @@ import type { Observable } from '../../../tools/observable'
 export const SESSION_STORE_KEY = '_dd_s_v2'
 export const LEGACY_SESSION_STORE_KEY = '_dd_s'
 
+export const CookieApi = {
+  COOKIE_STORE: 'cookieStore',
+  DOCUMENT_COOKIE: 'documentCookie',
```
BenoitZugmeyer: **suggestion:** use a const enum, with default "numbered" values?

BenoitZugmeyer (approved):

PR #4628, opened 2026-05-14
🎨 [PANA-7353] Clean up naming in recording code

sethfowler-datadog:
## Motivation

I'm about to start pushing a major new round of session replay improvements. Before that process starts, I'd like to take a moment to perform some cleanup. In SDK v6, the SDK included two serialization algorithms: the original one (`v1`) and the then-experimental replacement (`change`). In SDK v7, we've totally replaced the `v1` algorithm with the `change` algorithm, but the split that previously existing remains visible in the naming of many functions and types which specify that they generate "change" output or work with the "change" codepath. Now that there's only one codepath, including "change" in all these names is unnecessary. Let's clean it up to make things a little less verbose.

## Changes

This PR removes the term "change" everywhere in the recording code where it occurs unnecessarily. The term still appears, but only in places where it involves serialization logic for the `Change` record type itself; it's no longer spread throughout the code base.

This is a pure refactoring change; no functionality has been altered.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/rum/src/domain/record/serialization/serializationTransaction.ts
```diff
@@ -33,16 +33,16 @@ export const enum SerializationKind {
...
-export type ChangeSerializationTransactionCallback = (transaction: ChangeSerializationTransaction) => void
+export type SerializationTransactionCallback = (transaction: SerializationTransaction) => void
 
 /**
- * ChangeSerializationTransaction is used to build and emit a BrowserChangeRecord
+ * SerializationTransaction is used to build and emit a BrowserChangeRecord
  * containing a serialized snapshot of the DOM. Unlike SerializationTransaction, it
  * doesn't support emitting arbitrary BrowserRecords; instead, the builder methods it
  * exposes are used to construct a single BrowserChangeRecord which is emitted at the end
  * of the transaction.
```
gonzalezreal: 🤏🏽 This comment has become nonsensical 😄. The original compared `ChangeSerializationTransaction` to a (now-gone) sibling `SerializationTransaction`.

packages/rum/src/domain/record/test/serializeHtml.specHelper.ts
gonzalezreal: nit: I see this import has been changed to `'@datadog/browser-core/test'` in other spec files. Should we normalize all of them?

gonzalezreal (approved):
LGTM! Just a couple of nits

packages/rum/src/domain/record/serialization/serializationTransaction.ts
sethfowler-datadog: Haha, good catch. I'll fix that.

packages/rum/src/domain/record/test/serializeHtml.specHelper.ts
sethfowler-datadog: Yeah, it'd be good to keep that consistent; I'll fix that. Unfortunately the LSP for TypeScript loves to auto-insert these unidiomatic paths...

PR #4629, opened 2026-05-14
♻️ [PANA-7375] Eliminate unnecessary conditional node serialization

sethfowler-datadog:
## Motivation

There's an awkward edge case that exists today: in very narrow circumstances, we decline to serialize nodes when "bulk serializing" a DOM subtree based on certain conditions, but we _will_ serialize those exact same nodes if they're mutated later.

This breaks some invariants that you'd expect to hold. For example, you'd expect that if you encounter an unserialized node, then one of these three things must be true:
1. The node's parent also has not been serialized.
2. The node has a different privacy level than its parent, such that it's excluded from serialization.
3. The node has been newly added to the document since we last processed mutations.

Conditional serialization adds a fourth, non-obvious case to this list.

Similarly (and this is a case I had reason to care about recently), you might expect that this sequence of actions:
1. Populate the document.
2. Take a full snapshot.
3. Set the text content of a text node to 'foo'.
4. Take an incremental snapshot.

Would be guaranteed to produce the same final rendering as this sequence of actions:
1. Populate the document.
2. Set the text content of a text node to 'foo'.
3. Take a full snapshot.

However, today, this isn't necessarily so! That's because we apply conditional serialization to text nodes, and changing the text content of a text node can change whether it's included in the full snapshot. If it's not included, subsequent incremental mutations won't be captured either, even if the condition is satisfied as a result of the change. However, the next time a full snapshot is taken, the node will reappear -- hence the reason why the two sequences of action above can produce different results.

We should view this as a bug. There are two ways we could fix it:
1. We could detect when the conditions that led us to decline to serialize a node have changed, and consider whether we need to serialize it again.
2. We could just remove the unnecessary conditional serialization logic for these nodes.

I've chosen path #2; given the very limited circumstances in which we apply conditional serialization, and the very limited benefit we obtain from it, I think the best path forward is simply to remove it and eliminate this source of complexity.

## Changes

We conditionally decline to serialize nodes in two circumstances today:
1. Text nodes which are direct children of `<head>` are not serialized if they contain only whitespace.
2. Text nodes which are direct children of `<datalist>`, `<select>`, or `<optgroup>`are not serialized if they contain only whitespace and they would otherwise be masked.

In both of these situations, we now unconditionally serialize the nodes.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/rum-core/src/domain/privacy.ts
+++ b/packages/rum-core/src/domain/privacy.ts
@@ -220,37 +220,23 @@ export function censorText(text: string): string {
   return text.replace(/\S/g, TEXT_MASKING_CHAR)
 }
 
-export function getTextContent(textNode: Node, parentNodePrivacyLevel: NodePrivacyLevel): string | undefined {
+export function getTextContent(textNode: Node, parentNodePrivacyLevel: NodePrivacyLevel): string {
   // The parent node may not be a html element which has a tagName attribute.
   // So just let it be undefined which is ok in this use case.
   const parentTagName = textNode.parentElement?.tagName
   let textContent = textNode.textContent || ''
-
-  const shouldIgnoreWhiteSpace = parentTagName === 'HEAD'
-  if (shouldIgnoreWhiteSpace && !textContent.trim()) {
-    return
-  }
-
   const nodePrivacyLevel = parentNodePrivacyLevel
 
-  const isScript = parentTagName === 'SCRIPT'
-
-  if (isScript) {
+  if (parentTagName === 'SCRIPT') {
     // For perf reasons, we don't record script (heuristic)
     textContent = CENSORED_STRING_MARK
   } else if (nodePrivacyLevel === NodePrivacyLevel.HIDDEN) {
     // Should never occur, but just in case, we set to CENSORED_MARK.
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/domain/privacy.ts
```diff
@@ -220,44 +220,31 @@ export function censorText(text: string): string {
   return text.replace(/\S/g, TEXT_MASKING_CHAR)
 }
 
-export function getTextContent(textNode: Node, parentNodePrivacyLevel: NodePrivacyLevel): string | undefined {
+export function getTextContent(textNode: Node, parentNodePrivacyLevel: NodePrivacyLevel): string {
```
mormubis, line 223: I think we still need the node for mutation tracking, but the serialized content could be empty instead of the actual whitespace. No need to send whitespace that is invisible in `<head>`. What do you think?

mormubis (approved):

packages/rum-core/src/domain/privacy.ts
sethfowler-datadog: Yeah, I think that's reasonable; the important thing is really just that the node is actually generated. I'll push a commit that makes that change.

PR #4630, opened 2026-05-15
🔍 [REPRO – DO NOT MERGE] reproduce early-resource-timings flake on CI

thomas-lebeau:
## Motivation

The E2E test `rum resources › retrieve early requests timings` (async / npm / bundle variants at `test/e2e/scenario/rum/resources.scenario.ts:61`) is one of the top flaky tests this week on `webkit-pinned` — 28 / 24 / 21 occurrences over the last 7 days across many branches.

Working theory: WebKit clamps `PerformanceResourceTiming` timestamps and `performance.now()` to 1ms resolution. When the entire `/empty.css` request (a zero-byte response) fits inside one 1ms tick, every relevant timestamp clamps to the same value. The SDK's existing Safari fallback in `computeResourceEntryDuration` (`packages/rum-core/src/domain/resource/resourceUtils.ts:76`) is guarded by `startTime < responseEnd`, which fails when they're equal, so the SDK truthfully emits `duration: 0` — and the test asserts strictly `> 0`.

Locally on macOS, 50/50 repeats passed but every captured entry confirmed 1ms clamping. The failure presumably needs CI's faster Linux loopback to make collapse probable. **This branch verifies that.**

## Changes

⚠️ Temporary repro branch — must NOT be merged. Closing once we have data.

- Instrument the early-timings test (`test/e2e/scenario/rum/resources.scenario.ts`) to log:
  - The raw `PerformanceResourceTiming` for `empty.css` (all 12 timestamps + size fields).
  - Four consecutive `performance.now()` samples to expose clock granularity.
  - The SDK-emitted `resource.duration`, `download`, `first_byte`, `size`, `transfer_size`.
- Narrow `.gitlab-ci.yml` e2e job to:
  - only `webkit-pinned` (skip chromium/firefox/etc.)
  - only the `retrieve early requests timings` test via `-g`
  - `--repeat-each=50` × 3 sub-tests = 150 attempts
  - `--retries=0` so each repeat is an independent observation

## Test instructions

1. Wait for the GitLab `e2e: [webkit-pinned]` job to finish.
2. In the job log, look for `[EARLY-TIMINGS-DEBUG webkit]` lines.
3. Expected confirmation of the hypothesis:
   - `nowSamples` shows 4 identical integer-ms values (clock clamped to 1ms).
   - On failing attempts: `startTime === responseStart === responseEnd` in `raw=` and the corresponding `sdk=` shows `duration: 0`.
   - On passing attempts: at least one timestamp differs by ≥1ms; SDK shows `duration: ≥1ms`.
4. Failure count (compared to the local 0/50) tells us whether CI's faster loopback is the differentiator.

## Checklist

- [ ] Tested locally (50/50 passed — that's the point of the CI run)
- [ ] Tested on staging

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -244,8 +244,8 @@ e2e:
   interruptible: true
   parallel:
     matrix:
-      # REPRO branch: only webkit-pinned, the browser where the flake reproduces.
-      - BROWSER: [webkit-pinned]
+      # REPRO branch: webkit-pinned (where the flake reproduces) + chromium (comparison).
+      - BROWSER: [webkit-pinned, chromium]
   artifacts:
     when: always
     reports:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -256,8 +256,10 @@ e2e:
     - yarn build:apps
     # Browsers are pre-installed in the CI image. If playwright (current or pinned 1.40.1)
     # is upgraded without rebuilding the image, this job will crash — rebuild the image to fix it.
-    # REPRO branch: only the early-timings test, 50 repeats, no retries so each repeat is independent.
-    - FORCE_COLOR=1 PW_BROWSER=$BROWSER yarn test:e2e --project=$BROWSER -g "retrieve early requests timings" --repeat-each=50 --retries=0
+    # REPRO branch: only the early-timings test. workers=1 to serialize stdout (otherwise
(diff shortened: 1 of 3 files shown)


thomas-lebeau:
Repro confirmed — see analysis in PR description. Closing without merging; fix follows in a separate PR.

thomas-lebeau:
Re-running the same repro setup (webkit-pinned + chromium, repeat-each=30, retries=0) with the 50ms /empty.css delay applied to verify the fix lands 0 failures.

PR #4633, opened 2026-05-15
👷 add api-performance-benchmark scheduled job + rename test-performance to bundle-size

thomas-lebeau:
## Motivation

Two unrelated cleanups:

1. The `test-performance` GitLab job now only computes bundle sizes (the CPU and memory measurements it used to compute were tightly coupled to the per-PR comment flow). Rename the job to match.
2. The per-API CPU/memory measurements still have value as a regression signal on `main`, but they were originally compared against a base commit for every PR. Keep the measurement, drop the comparison, and run it on the existing `performance-benchmark-scheduled` schedule instead of on every PR.

## Changes

- Rename CI job `test-performance` → `bundle-size`.
- Add new CI job `api-performance-benchmark`, gated by `TARGET_TASK_NAME == "performance-benchmark-scheduled"` so it shares the cadence of the existing benchmark schedule.
- Add `scripts/api-performance/`:
  - `lib/cpuPerformance.ts` — triggers the Datadog synthetic test that drives the playground (`/performance/cpu`); the synthetic test reports its per-API CPU metrics to Datadog directly.
  - `lib/memoryPerformance.ts` — launches Chrome via puppeteer, clicks each playground button 30× per API while sampling the heap profile, and reports the median SDK-attributable bytes per API to Datadog.
  - `lib/constants.ts` — the 7-entry `TESTS` list (RUM addGlobalContext / addAction / addError / addTiming / startView / startStop session-replay-recording, Logs logMessage).
  - `lib/reportToDatadog.ts` — local copy of the helper (lint forbids importing across `scripts/*/lib/` directories).
- Drop the legacy `scripts/performance/lib/{cpuPerformance,memoryPerformance,constants}.ts` files and the CPU/Memory sections from the PR comment formatter (+ its spec).

## Notes

- Earlier iterations of this PR added per-API scenarios under `test/performance/scenarios/`, but they produced order-of-magnitude-different numbers from the old job (scenario-total vs per-call methodology). Reverted in favor of preserving the original measurement approach.
- The new job runs on the same schedule as `performance-benchmark` — change the cadence in GitLab's CI/CD → Schedules → `performance-benchmark-scheduled`.
- Metrics flow to Datadog: CPU from the synthetic test (its existing dashboards), memory as logs tagged per `TESTS.property` from `scripts/api-performance/lib/memoryPerformance.ts`.

## Test instructions

- Check the next GitLab pipeline: the `bundle-size` job runs (renamed from `test-performance`) and posts the bundle-size PR comment as before.
- On the next `performance-benchmark-scheduled` run, the new `api-performance-benchmark` job runs alongside `performance-benchmark` and pushes memory metrics to Datadog (look for SDK memory consumption logs from the `browser-sdk` service / `ci` env on main).

## Checklist

- [x] Tested locally (typecheck, lint, script test pass)
- [ ] Tested on staging


thomas-lebeau:
@codex review

rgaignault (approved):

mormubis (approved):

PR #4514, opened 2026-04-21
✨[Profiler] Quota check

thomasbertet:
## Motivation

Adds client-side enforcement of the per-org profiling quota by calling the quota admission API before allowing profiling data to be sent. Ticket: [PROF-13798](https://datadoghq.atlassian.net/browse/PROF-13798).

## Changes

- **New `quotaCheck.ts`**: calls `GET https://quota.browser-intake-<site>/api/v2/profiling/quota?session_id=<id>` with a `DD-CLIENT-TOKEN` header. Returns a `QuotaResult { decision: 'quota_ok' | 'quota_ko', reason: QuotaReason }`. Client-side 5 s timeout via `AbortController` + `Promise.race`. Uses the Zone.js-safe `fetch` and `setTimeout` wrappers from `@datadog/browser-core`. Fail-open on timeout (`reason: 'timeout'`), network error and unparseable response (`reason: 'api-error'`).

- **`QuotaReason` type**: `BackendQuotaReason` (exact strings from the API: `quota_ok`, `quota_exceeded`, `org_disabled`, `backend_unavailable`, `backend_client_not_initialized`, `undefined`) union with `FrontendQuotaReason` (`timeout`, `api-error`).

- **`profiler.ts`**: profiler starts recording immediately (optimistic), quota check fires in parallel. On `quota_ko` decision: profiler stops, trace is discarded (no data sent), `_dd.profiling.quota_reason` is set on RUM events with the specific reason. A generation counter prevents stale results from a prior session applying to a new one. `SESSION_RENEWED` restarts the profiler (and re-checks quota) when previously stopped due to `quota_ko`.

- **`rumProfiler.types.ts`**: `RumProfilerStoppedInstance.stateReason` uses `'quota_ko'` to cover all quota-denied outcomes.

- **`transportConfiguration.ts`**: exposes `clientToken` on `TransportConfiguration` (and thus `RumConfiguration`) so it is accessible from the profiler lazy-loaded chunk.

**Note:** `_dd.profiling.quota_reason` is passed with `as any` until the `rum-events-format` [schema is updated](https://github.com/DataDog/rum-events-format/pull/382).

## Test instructions

- [ ] Unit tests: `yarn test:unit --spec packages/rum/src/domain/profiling/quotaCheck.spec.ts` (13 tests)
- [ ] Unit tests: `yarn test:unit --spec packages/rum/src/domain/profiling/profiler.spec.ts` (28 tests, includes quota check scenarios)
[...]

PR #4638, opened 2026-05-18
[PROF-13798] ✨ Add subdomain support to ProxyFn for quota check

thomasbertet:
## Motivation

The quota admission endpoint lives on a subdomain (`quota.browser-intake-*`) that differs from the standard intake host. The existing `ProxyFn` signature has no way to express this, so proxy implementations can't route quota requests to the right backend.

Stacks on top of [#4514](https://github.com/DataDog/browser-sdk/pull/4514).

## Changes

- **`ProxyFn`**: adds optional `subdomain?: string` to the options object, allowing proxy implementations to target intake subdomains.
- **`checkProfilingQuota`**: passes `subdomain: 'quota'` when calling a function proxy.
- **proxy-as-string path**: appends `ddforwardSubdomain=quota` to the forwarded URL so the proxy server can identify quota requests without inspecting the full path.
- **E2E intake server**: detects `ddforwardSubdomain=quota` and returns a static admit response.

## Note

`subdomain` is optional and defaults to nothing, so existing proxy implementations are unaffected.

### DIFF
--- a/packages/rum/src/domain/profiling/quotaCheck.ts
+++ b/packages/rum/src/domain/profiling/quotaCheck.ts
@@ -29,8 +29,21 @@ function parseQuotaResult(body: unknown, httpStatusFallback: QuotaResult): Quota
 }
 
 function buildQuotaUrl(configuration: RumConfiguration, sessionId: string): string {
+  const path = '/api/v2/profiling/quota'
+  const parameters = `session_id=${sessionId}`
+  const proxy = configuration.proxy
   const host = `quota.${buildEndpointHost({ site: configuration.site, clientToken: configuration.clientToken })}`
-  return `https://${host}/api/v2/profiling/quota?session_id=${sessionId}`
+
+  if (typeof proxy === 'string') {
+    // ddforwardSubdomain tells the proxy which intake subdomain to target.
+    return `${proxy}?ddforward=${encodeURIComponent(`${path}?${parameters}`)}&ddforwardSubdomain=quota`
+  }
+
+  if (typeof proxy === 'function') {
+    return proxy({ path, parameters, subdomain: 'quota' })
+  }
+
+  return `https://${host}${path}?${parameters}`
 }
 
 export function checkProfilingQuota(
--- a/test/e2e/lib/framework/serverApps/intake.ts
+++ b/test/e2e/lib/framework/serverApps/intake.ts
@@ -11,6 +11,16 @@ export function createIntakeServerApp(intakeRegistry: IntakeRegistry) {
 
   app.post('/', createIntakeProxyMiddleware({ onRequest: (request) => intakeRegistry.push(request) }))
 
+  // Quota admission check — always admit in the test environment.
+  // Triggered via proxy-as-string with ddforwardSubdomain=quota.
+  app.get('/', (req, res) => {
+    if (req.query.ddforwardSubdomain === 'quota') {
+      res.json({ data: { id: 'test', type: 'profiling-quota', attributes: { admitted: true, reason: 'quota_ok' } } })
+    } else {
+      res.status(404).end()
+    }
+  })
+
   app.post('/api/unstable/debugger/frontend/probes', (_req, res) => {
     res.json({ nextCursor: '', updates: debuggerProbes, deletions: [] })
   })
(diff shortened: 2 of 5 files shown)


test/e2e/lib/framework/serverApps/intake.ts
```diff
@@ -11,6 +11,16 @@ export function createIntakeServerApp(intakeRegistry: IntakeRegistry) {
 
   app.post('/', createIntakeProxyMiddleware({ onRequest: (request) => intakeRegistry.push(request) }))
 
+  // Quota admission check — always admit in the test environment.
+  // Triggered via proxy-as-string with ddforwardSubdomain=quota.
+  app.get('/', (req, res) => {
```
amortemousque, line 16: **❓ question:** ‏Do you need to keep the API mock in `profiling.scenario.ts` if you have this route defined?

amortemousque (approved):

test/e2e/lib/framework/serverApps/intake.ts
thomasbertet: I think we do because this one is targeting the proxy URL case ? 🤔

thomasbertet: Courtesy of our AI friend: 

```tsx
// pageSetups.ts
proxy: servers.intake.origin  // e.g. http://192.168.1.83:9240
````

Then in `quotaCheck.ts`, `buildQuotaUrl` runs with `typeof proxy === 'string'`:

```tsx
return `${proxy}?ddforward=${encodeURIComponent(`${path}?${parameters}`)}&ddforwardSubdomain=quota`
  // → http://192.168.1.83:9240?ddforward=%2Fapi%2Fv2%2Fprofiling%2Fquota%3Fsession_id%3D...&ddforwardSubdomain=quota
```
So the fetch goes to http://192.168.1.83:9240 — the intake server — not to quota.browser-intake-datadoghq.com.

  Meanwhile, the profiling.scenario.ts test.beforeEach has:
  ```tsx
await page.route('https://quota.browser-intake-datadoghq.com/api/v2/profiling/quota*', ...)
````

That pattern doesn't match http://192.168.1.83:9240?ddforward=... — so `page.route()` never intercepts in the proxy case.

The request actually lands on the intake server Express app. Without the GET handler, Express has no route for GET / → returns 404. Chrome logs that 404 in the browser console → E2E framework sees a console error → test fails.

So the `page.route()` mock is for the no-proxy case, and the GET handler in `intake.ts` is for the proxy case . They serve the same purpose in different scenarios.

Comment (amortemousque) [reply test/e2e/lib/framework/serverApps/intake.ts:16]: don't we always have proxy set in the e2e tests?

File: packages/rum/src/domain/profiling/quotaCheck.ts [hunk 2]
```diff
@@ -29,8 +29,21 @@ function parseQuotaResult(body: unknown, httpStatusFallback: QuotaResult): Quota
...
 function buildQuotaUrl(configuration: RumConfiguration, sessionId: string): string {
+  const path = '/api/v2/profiling/quota'
+  const parameters = `session_id=${sessionId}`
+  const proxy = configuration.proxy
   const host = `quota.${buildEndpointHost({ site: configuration.site, clientToken: configuration.clientToken })}`
-  return `https://${host}/api/v2/profiling/quota?session_id=${sessionId}`
+
+  if (typeof proxy === 'string') {
+    // ddforwardSubdomain tells the proxy which intake subdomain to target.
+    return `${proxy}?ddforward=${encodeURIComponent(`${path}?${parameters}`)}&ddforwardSubdomain=quota`
```
amortemousque, line 39: **💭 thought:** ‏I’m wondering whether that would be a breaking change, depending on how customers parse the params in their proxy.🤷‍♂️

thomasbertet: It would be surprising that adding a query param ends up breaking something, but I guess you never know 🤔 It's the less invasive change we can make as it's not touching the `ddforward` param.

test/e2e/lib/framework/serverApps/intake.ts
thomasbertet: Seems we do! I cleaned-up there: https://github.com/DataDog/browser-sdk/pull/4656

PR #4652, opened 2026-05-19
fix(view-update): exclude _dd.application from VIEW_UPDATE payloads

andreicoj:
## Problem

\`_dd.application\` contains four SDK init-time config booleans set once at SDK initialisation and constant for the session:
- \`rum_enabled\`
- \`product_analytics_enabled\`
- \`error_focused_mode_enabled\`
- \`product_analytics_preview_enabled\`

\`computeAssembledViewDiff\` uses \`diffMerge\` which recurses into nested objects and naturally omits unchanged fields — so these fields are already excluded in the common case. However they are not in \`ignoreKeys\`, which leaves the intent implicit and risks future regressions if \`ignoreKeys\` logic or \`diffMerge\` changes.

Backend observability data (PR #136915, \`sdk_source:browser\`, 18.5 h, ~341 K VIEW_UPDATEs on staging) confirmed 32.4% redundancy across all four fields — arriving unchanged on ~1 in 3 VIEW_UPDATEs driven by periodic checkpoint VIEW resets.

## Solution

Add \`_dd.application\` to the \`ignoreKeys\` set in \`computeAssembledViewDiff\`. The baseline VIEW always carries the full \`_dd.application\` object; VIEW_UPDATEs never need to re-send it.

## Changes

- \`startRumBatch.ts\`: add \`'_dd.application'\` to \`ignoreKeys\`
- \`startRumBatch.spec.ts\`: add test verifying \`_dd.application\` is absent from diff output

### DIFF
--- a/packages/rum-core/src/transport/startRumBatch.spec.ts
+++ b/packages/rum-core/src/transport/startRumBatch.spec.ts
@@ -205,6 +205,20 @@ describe('computeAssembledViewDiff', () => {
 
     expect(current.service).toBe(currentService)
   })
+
+  it('should exclude _dd.application from view_update (SDK init-time config, never changes)', () => {
+    const application = { rum_enabled: true, product_analytics_enabled: false }
+    const last = makeAssembledView({ _dd: { document_version: 1, format_version: 2, sdk_name: 'rum', configuration: { start_session_replay_recording_manually: false }, application } })
+    const current = makeAssembledView({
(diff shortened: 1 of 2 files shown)


andreicoj:
Closing — the fix belongs in the backend. _dd.application.* is stamped onto every RUM event (including VIEW_UPDATE) by RumApplicationEnabledProductMarker in the intake pipeline, not by the SDK. The SDK never sends these fields.

PR #4660, opened 2026-05-21
🐛 Forward initial loader errors to RouterProvider's onError

### LINKED ISSUE #4657: 🐛 onError callback doesn't get called with datadog react router wrapper
**Describe the bug**

Product=rum

Recently, in react-router an [issue](https://github.com/remix-run/react-router/issues/14941) has been fixed in 7.15.1. But if we want to use createBrowserRouter with the datadog rum wrapper in the @datadog/browser-rum-react library, the onError callback does not get called.

**To Reproduce**

A sandbox where it can be reproduced:

https://codesandbox.io/p/devbox/trusting-feynman-kzjsqr?file=%2Fsrc%2Fmain.tsx

Using the original createBrowserRouter works (the log in the onError callback appears), but if we use the datadog one it does not.

**Expected behavior**

The onError callback should be called / the log inside should appear.

Thanks a lot!

rgaignault:
## Motivation

`wrapCreateRouter` subscribes synchronously when the router is created, before `<RouterProvider>` mounts. As a result, the initial loader-error notification reaches our subscriber first and never makes it to RouterProvider, its `onError` callback is silently skipped.

Two paths lead to this, both broken by our wrapper:
- `react-router` ≥7.15.1 buffers the initial state update when no subscriber is attached and replays it one-shot to the first subscriber.
- In React's concurrent mode, the loader-error notification can fire after our subscribe but before `RouterProvider`'s deferred `useLayoutEffect`.

Fixes #4657.

## Changes

- `wrapCreateRouter` now detects when a notification carries `newErrors` and forwards it (one-shot) to the next subscriber via a temporary `router.subscribe` wrapper. RouterProvider's `setState` is then called with the same payload, triggering `onError` as expected.
- Adds the missing second argument (`opts`) to the subscriber type so we can pass it through verbatim.
- Bumps `react-router` in the v7 test app to `7.15.1` and adds a regression e2e test that mounts the SPA with a route whose loader throws synchronously and asserts that `onError` fires.

## Test instructions

```bash
yarn test:unit --spec packages/rum-react/src/domain/reactRouter/createRouter.spec.ts
yarn test:e2e -g "should call RouterProvider onError when initial route loader throws synchronously"
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/build/build-test-apps.ts
+++ b/scripts/build/build-test-apps.ts
@@ -158,13 +158,40 @@ async function buildReactRouterv7App() {
   await modifyFile(path.join(appPath, 'package.json'), (content: string) =>
     content
       .replace(/"name": "react-router-v6-app"/, '"name": "react-router-v7-app"')
-      .replace(/"react-router-dom": "[^"]*"/, '"react-router": "7.0.2"')
+      .replace(/"react-router-dom": "[^"]*"/, '"react-router": "7.15.1"')
   )
 
   await modifyFile(path.join(appPath, 'app.tsx'), (content: string) =>
     content
       .replace('@datadog/browser-rum-react/react-router-v6', '@datadog/browser-rum-react/react-router-v7')
       .replace("from 'react-router-dom'", "from 'react-router'")
+      // Inject a loader that throws synchronously when ?test-loader-error is present,
+      // and an onError marker on RouterProvider. Used by the regression test for #4657.
+      .replace(
+        /({\s*index: true,\s*Component: HomePage,\s*},)/,
+        `{
+        index: true,
+        loader: ({ request }: { request: Request }) => {
+          if (new URL(request.url).searchParams.has('test-loader-error')) {
+            throw new Error('Synchronous loader error')
+          }
+          return null
+        },
+        Component: HomePage,
+      },`
+      )
+      .replace(
+        /<RouterProvider router={router} \/>/,
+        `<RouterProvider
+      router={router}
+      onError={(error: unknown) => {
+        const el = document.createElement('div')
+        el.setAttribute('data-testid', 'on-error-fired')
+        el.textContent = (error as Error).message ?? String(error)
+        document.body.appendChild(el)
+      }}
+    />`
+      )
   )
 
   await modifyFile(path.join(appPath, 'webpack.config.js'), (content: string) =>
--- a/scripts/build/build-test-apps.ts
+++ b/scripts/build/build-test-apps.ts
@@ -165,21 +165,8 @@ async function buildReactRouterv7App() {
     content
       .replace('@datadog/browser-rum-react/react-router-v6', '@datadog/browser-rum-react/react-router-v7')
       .replace("from 'react-router-dom'", "from 'react-router'")
-      // Inject a loader that throws synchronously when ?test-loader-error is present,
-      // and an onError marker on RouterProvider. Used by the regression test for #4657.
-      .replace(
-        /({\s*index: true,\s*Component: HomePage,\s*},)/,
-        `{
-        index: true,
-        loader: ({ request }: { request: Request }) => {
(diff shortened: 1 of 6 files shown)


scripts/build/build-test-apps.ts
```diff
@@ -158,13 +158,40 @@ async function buildReactRouterv7App() {
...
+        /<RouterProvider router={router} \/>/,
+        `<RouterProvider
+      router={router}
+      onError={(error: unknown) => {
+        const el = document.createElement('div')
+        el.setAttribute('data-testid', 'on-error-fired')
+        el.textContent = (error as Error).message ?? String(error)
+        document.body.appendChild(el)
+      }}
+    />`
```
thomas-lebeau: **💬 suggestion:** ‏let's build that inside the test-app itself please

test/e2e/scenario/plugins/reactPlugin.scenario.ts
```diff
@@ -94,6 +94,20 @@ test.describe('plugin: react', () => {
             expect(browserLogs.filter((log) => log.level === 'error').length).toBeGreaterThan(0)
           })
         })
+
+      if (appName === 'react-router-v7-app') {
```
thomas-lebeau, line 98: **❓ question:** ‏don't we want to test that on v6 too?

rgaignault: I don't think we want since it's a v7 related bug

thomas-lebeau: yes, the bug was on v7, but the question is more is the feature (i.e. what is tested) supported in earlier version? in other world, if the test pass in v6, I see no reason to skip it.

scripts/build/build-test-apps.ts
```diff
@@ -165,21 +165,8 @@ async function buildReactRouterv7App() {
...
-          if (new URL(request.url).searchParams.has('test-loader-error')) {
-            throw new Error('Synchronous loader error')
-          }
-          return null
-        },
-        Component: HomePage,
-      },`
-      )
+      // Add an onError marker on RouterProvider — v7-only prop, exercised by the
+      // regression test for https://github.com/DataDog/browser-sdk/issues/4657.
```
thomas-lebeau, line 169: **🔨 warning:** ‏This comment became irrelevent here

test/e2e/scenario/plugins/reactPlugin.scenario.ts
rgaignault: v6 doesn't expose onError on <RouterProvider>
See : https://github.com/remix-run/react-router/blob/e185e80dbcec9f4f7d9c8b1eb14c8f09870826ef/packages/react-router/lib/components.tsx#L59

scripts/build/build-test-apps.ts
rgaignault: what do you mean by irrelevant ?

test/e2e/scenario/plugins/reactPlugin.scenario.ts
thomas-lebeau: so what happens when there is a synchronous error? they are not collected?

rgaignault: They surface via the route's errorElement not a callback. v7 keeps that path and adds onError on top, but the errorElement path doesn't go through subscriber notifications, so our wrapper doesn't intercept it. 
https://reactrouter.com/6.30.0/route/error-element

BeltranBulbarellaDD (approved):

scripts/build/build-test-apps.ts
thomas-lebeau: sorry, I didn't read properly, you're right

thomas-lebeau (approved):

PR #4663, opened 2026-05-22
⚗️ auto-disable Live Debugger after prolonged Delivery API outage

watson:
## Motivation

If the Live Debugger Delivery API becomes unavailable (network outage, server-side issue, misconfiguration, …) the SDK currently keeps polling it forever, logging an error on every failed request. Aside from the noise, this means stale probes stay installed indefinitely with no way to update or remove them.

We want to tolerate transient glitches (the common case) but automatically and permanently disable the Live Debugger when the Delivery API has been unreachable for long enough that something is clearly wrong.

## Changes

Added a circuit breaker on the Delivery API polling loop in `packages/debugger/src/domain/deliveryApi.ts`:

- **Tolerated transparently**: network errors (`fetch` throws) and 5xx responses, as long as they don't span more than 5 continuous minutes.
- **Recovery**: any 2xx response resets the failure window — short glitches are forgiven.
- **Trips immediately on 4xx**: these are config issues (bad client token, wrong service, …) that won't self-recover; no point waiting.
- **Trips after 5 minutes** of continuous network errors or 5xx.
- **On trip**: stops polling, calls `clearProbes()` (uninstalls every active probe), emits a `display.warn` for developers, and fires `addTelemetryDebug('Delivery API circuit breaker tripped', { reason })` for ops visibility. The trip is permanent for the page lifetime.

The 5-minute threshold is configurable via a new public init option:

```ts
datadogDebugger.init({
  clientToken: '…',
  service: 'my-app',
  maxUnreachableDuration: 60_000, // shut down after 1 min instead
})
```

`maxUnreachableDuration` must be a positive finite number; invalid values fall back to the 5-minute default (same defensive pattern as `pollInterval` and the probe budget rates).

## Test instructions

Unit tests cover the full behavior (see `packages/debugger/src/domain/deliveryApi.spec.ts`):

- Polling continues while failures last less than 5 minutes
- Polling stops after 5 minutes of continuous network failures
- Polling stops after 5 minutes of continuous 5xx responses
- 4xx responses trip immediately (no grace period)
- A successful response resets the failure window
- Active probes are cleared on trip
- `display.warn` is emitted on trip
- A configured `maxUnreachableDuration` is honored
- Invalid `maxUnreachableDuration` (e.g. `-1`) falls back to the default

```bash
yarn test:unit --spec packages/debugger/src/domain/deliveryApi.spec.ts
```

To exercise it manually, point a debugger-enabled sandbox at a proxy that returns 503 (or block the request in DevTools), set `maxUnreachableDuration: 10_000` for quick feedback, and observe the warning + `clearProbes` after the window elapses.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/debugger/src/domain/deliveryApi.spec.ts
+++ b/packages/debugger/src/domain/deliveryApi.spec.ts
@@ -383,6 +383,44 @@ describe('deliveryApi', () => {
       expect(fetchSpy).toHaveBeenCalledTimes(1)
     })
 
+    // 408 (Request Timeout) and 429 (Too Many Requests) are 4xx statuses that
+    // are nevertheless expected to recover on their own - the request just took
+    // too long, or the client/proxy is temporarily rate-limited. They should
+    // follow the same transient-failure path as 5xx rather than trip the breaker
+    // on a single response. This matches core's transport retry predicate
+    // (`shouldRetryRequest` in `sendWithRetryStrategy.ts`).
+    for (const status of [408, 429]) {
+      it(`should treat ${status} as a transient failure rather than tripping immediately`, async () => {
+        respondWith({}, status)
+        startDeliveryApiPolling(makeConfig({ pollInterval: POLL_INTERVAL_MS }))
+        await flushPromises()
+
+        // Just shy of five minutes of continuous failures - should still be polling.
+        const ticks = Math.floor((FIVE_MINUTES_MS - POLL_INTERVAL_MS) / POLL_INTERVAL_MS)
+        for (let i = 0; i < ticks; i++) {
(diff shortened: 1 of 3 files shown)


watson:
* **#4663** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4663?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4663?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jpbempel (approved):

PR #4676, opened 2026-05-26
✨ Add initial Salesforce support.

BeltranBulbarellaDD:
## Motivation

Make the SDK not crash when instrumenting in Salesforce lightning environments.

## Changes

First 2 commits make the SDK not crash under Salesforce/LWC environment. ([d32f6ea](https://github.com/DataDog/browser-sdk/pull/4676/commits/d32f6eae054448825d9e44f7371418834d960cf9)
[c02e9bf](https://github.com/DataDog/browser-sdk/pull/4676/commits/c02e9bfe2e7026f41815088220f86d56ffc78dc7))

And this commit allows the SDK to report correctly resources and have APM correlation [0d952b5](https://github.com/DataDog/browser-sdk/pull/4676/commits/0d952b563745b53910cc01da0edbc70ff0c91d7a)

This PR allows to have a Salesforce dedicated entrypoint, and build the bundle so we can upload as a Static Resource in the Salesforce Org. 

`reportObservable`, `sessionInCookie`, `instrumentMethod`, `cookieObservable`, `urlContexts` changes all make the SDK not crash, each one has comments explaining why.

`viewNameTracker` exists because we cannot instrument `history` methods so we have to poll `location`. This file is a WIP i'll improve it. 

Salesforce application can be found [here](https://github.com/DataDog/browser-sdk-salesforce-POC-app/pull/5)

## Test instructions

[Lightning App](https://connect-fun-8831-dev-ed.scratch.lightning.force.com/lightning/page/home)
User: [test-ripmjgp4afsc@example.com](mailto:test-ripmjgp4afsc@example.com)
Pass: wwchjcpxdtXi5!ercrsj

[Experience Cloud](https://connect-fun-8831-dev-ed.scratch.my.site.com/ebikes/s/)
[RUM App](https://app.datadoghq.com/rum/sessions?query=%40type%3Asession%20%40application.id%3A1397744d-34f4-4a6a-a735-801e31c18221&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&refresh_mode=sliding&from_ts=1779372695209&to_ts=1779459095209&live=true)

On the top left, on the 9 dots, --> View All --> you can see both apps. Lightning and Experience Cloud.

Don't forget to make in the Dev Extension to change Event collection strategy to Requests

Screenshot 2026-05-27 at 00 06 00

## Checklist

I'm missing adding tests

- [x] Tested locally
- [ ] Tested on staging


BeltranBulbarellaDD:
@codex review

packages/rum-slim/src/domain/salesforce/viewNameTracker.ts
BeltranBulbarellaDD, line 1: I'm rewriting this file

packages/browser-rum-core/src/browser/cookieObservable.ts
amortemousque, line 1: **💬 suggestion:** There is already a fallback strategy implemented. It would be easier to extend it like this:

```ts
const detectCookieChangeStrategy = isCookieStoreSupported 
    ? listenToCookieStoreChange(configuration)
    : watchCookieFallback

function isCookieStoreSupported() {
  const cookieStore = (window as CookieStoreWindow).cookieStore
  if (!cookieStore) {
    return false
  }
  try {
  // Salesforce exposes cookieStore but does not support change events
    addEventListener(configuration, cookieStore, DOM_EVENT.CHANGE, noop).stop()
    return true
  } catch {
    return false
  }
}
```

amortemousque:
Generally speaking, I would avoid adding low-level try/catch blocks in generic methods like `instrumentMethod`, because we might become blind to unexpected issues or regressions. 

Case by case: 
- Cookie observable: We already have a fallback, so we can extend the support check by verifying whether subscribing to the "change" event throws.
- Report (with CSP):  we already check window.ReportingObserver existence for unsupported browsers. We could extend this by checking whether securitypolicyviolation is readonly before instrumentation.
- View tracking: we should explore integrating with lightning/navigation and rely on manual view tracking, similar to other integrations.
- History/location
    - With manual view tracking, locationChangeObservable would receive both the view name and URL changes when a view starts.
    - We could then simply check whether pushState is readonly and skip the instrumentation if it is.
    - In any case, once manual view tracking is enabled, listening to History API changes becomes secondary. We could almost skip it entirely, but avoiding breaking changes is probably safer.

packages/core/src/browser/cookieAccess.ts
```diff
@@ -162,3 +163,12 @@ export function createDocumentCookieAccess(
...
 }
+
+// Salesforce LWS does not support the change event of CookieStore objects. https://developer.salesforce.com/tools/lws-distortion-viewer
+export function isCookieStoreSupported(configuration: Configuration): boolean {
+  const cookieStore = globalObject.cookieStore
+  if (!cookieStore) {
+    return false
+  }
+  return isEventSupported(configuration, cookieStore, DOM_EVENT.CHANGE, noop)
+}
```
amortemousque:
```suggestion
export function isCookieStoreSupported(configuration: Configuration): boolean {
  return globalObject.cookieStore && isEventSupported(configuration, cookieStore, DOM_EVENT.CHANGE, noop)
}
```

BeltranBulbarellaDD:
Closing for https://github.com/DataDog/browser-sdk/pull/4726

PR #4683, opened 2026-05-27
♻️ unify endpoint URL building with buildEndpointUrl

BenoitZugmeyer:
## Motivation

Customers using proxy routing should get consistent and predictable intake behavior across Browser SDK features.  
This change improves consistency of URL construction so proxy routing, subdomain forwarding, and site-based fallback behave the same way in core ingestion paths, profiling quota checks, and remote configuration fetches.

## Changes

- unify endpoint URL construction through `buildEndpointUrl` in core configuration logic
- switch profiling quota endpoint generation to use the shared endpoint URL builder
- switch remote configuration endpoint generation to use the same shared builder (including `sdk-configuration` subdomain)
- extend endpoint URL tests to cover:
  - string and relative proxy URLs
  - subdomain forwarding with proxies
  - proxy function delegation
  - default-site fallback behavior
  - empty-parameters URL behavior
- add/adjust profiling quota tests to validate normalized proxy URL behavior

## Test instructions

1. Run the sandbox with a proxy path configured (for example `proxy: '/proxy'` in SDK init).
2. Open the sandbox in the browser and verify in Network that intake requests go through `/proxy` and include `ddforward`.
3. Verify profiling quota requests include `ddforwardSubdomain=quota` when proxying is enabled.
4. Disable proxy and verify requests target the expected Datadog intake host based on `site`.
5. Configure remote configuration (with an id) and verify the fetched endpoint uses the `sdk-configuration` subdomain path.
6. Validate that requests still succeed when query parameters are empty for endpoints that don’t require them.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/domain/configuration/endpointBuilder.ts
+++ b/packages/core/src/domain/configuration/endpointBuilder.ts
@@ -2,8 +2,9 @@ import type { Payload } from '../../transport'
 import { timeStampNow } from '../../tools/utils/timeUtils'
 import { normalizeUrl } from '../../tools/utils/urlPolyfill'
 import { generateUUID } from '../../tools/utils/stringUtils'
+import type { Site } from '../intakeSites'
 import { INTAKE_SITE_US1 } from '../intakeSites'
-import type { InitConfiguration } from './configuration'
+import type { InitConfiguration, ProxyFn } from './configuration'
 
 // replaced at build time
 declare const __BUILD_ENV__SDK_VERSION__: string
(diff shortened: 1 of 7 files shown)


packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -39,45 +40,59 @@ export function createEndpointBuilder(
...
+    let url = `${normalizeUrl(proxy)}?ddforward=${encodeURIComponent(pathAndParameters)}`
+    if (subdomain) {
+      url += `&ddforwardSubdomain=${subdomain}`
+    }
+    return url
   }
+
   if (typeof proxy === 'function') {
-    return (parameters) => proxy({ path, parameters })
+    return proxy({ path, parameters, subdomain })
```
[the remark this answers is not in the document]
BenoitZugmeyer: It's fine.

rgaignault (approved):

packages/core/src/domain/configuration/endpointBuilder.ts
```diff
@@ -39,45 +40,59 @@ export function createEndpointBuilder(
...
-  trackType: TrackType
-): (parameters: string) => string {
-  const path = `/api/v2/${trackType}`
-  const proxy = initConfiguration.proxy
+export interface BuildEndpointUrlOptions {
+  proxy?: string | ProxyFn
+  site: Site | undefined
+  subdomain?: string
+  path: string
+  parameters: string
```
bcaudan: **💬 suggestion:** ‏What about making `parameters` optional? it would avoid to pass an empty string

bcaudan (approved):

mormubis (approved):

gogusarov (approved):

packages/core/src/domain/configuration/endpointBuilder.ts
BenoitZugmeyer: Thanks for the suggestion, ✅

thomas-lebeau (approved):

PR #4693, opened 2026-05-28
👷 add CI check for unreplaced __BUILD_ENV__ placeholders

rgaignault:
## Motivation

Guards against the class of bug where a release shipped with `__BUILD_ENV__SDK_VERSION__` left in NPM output because the post-processing step only replaced the first occurrence. The replacement step now uses `replaceAll()`, but nothing verifies the result.

## Changes

Adds a per-package check in `scripts/check-packages.ts` that greps the built `cjs`, `esm` and `bundle` directories for any leftover `BUILD_ENV` placeholder and fails the build if any are found.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


rgaignault:
@codex review

scripts/check-build-env-placeholders.ts
```diff
@@ -0,0 +1,68 @@
...
+        scannedFiles += 1
+        const matches = content.match(PLACEHOLDER_RE)
+        if (matches) {
+          hits.push({ file, placeholders: matches })
+        }
+      }
+    }
+  }
+  return { hits, scannedFiles }
+}
```
BenoitZugmeyer: **suggestion:** 

```ts
function checkBuildEnvPlaceholders(packagePath: string) {
  try {
    command`grep -r -q BUILD_ENV cjs esm bundle`.withCurrentDirectory(packagePath).run()
  } catch {
    return
  }
  throw new Error(`Found unreplaced __BUILD_ENV__ placeholders in built package ${packagePath}`)
}
```

-> put this in `check-package.ts`. It should be good enough, no need for a whole new script reimplementing grep.

BenoitZugmeyer (changes requested):

BenoitZugmeyer (approved):

PR #4695, opened 2026-05-28
♻️ improve `globalObject` definition and usage

BenoitZugmeyer:
## Motivation

Two goals:
* Make the GlobalObject type customizable, so we don't *have* to rely on typescript native types. Concrete example is: the `location` object is not always defined in practice, we can make it optional in our types (in a future PR).
* Get rid of the useless `getGlobalObject` function

## Changes

- replace `getGlobalObject()` usages with `globalObject`
- adjust the `GlobalObject` type to make it easier to understand and customizable

## Test instructions

1. Run `yarn dev` and open the sandbox.
2. Initialize `DD_RUM` + `DD_LOGS` in `sandbox/index.html`.
3. In console, verify both globals exist and basic calls work.
4. Trigger an error and confirm no SDK crash.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/core/src/tools/globalObject.ts
```diff
@@ -1,23 +1,33 @@
...
- * re-evaluating the heavyweight fallback logic in `getGlobalObject()`.
+ * Cached reference to the global object so it can be imported and re-used.
  */
-// eslint-disable-next-line local-rules/disallow-side-effects
-export const globalObject = getGlobalObject<GlobalObject>()
+export const globalObject = globalThis as GlobalObject
+
+export function getGlobalObject<T = GlobalObject>() {
+  return globalObject as T
+}
```
thomas-lebeau: **❓ question:** ‏do we still need this function?

packages/debugger/src/entries/main.spec.ts
```diff
@@ -7,7 +7,7 @@ import type { BrowserWindow } from './main'
 import { datadogDebugger } from './main'
 
 describe('datadogDebugger', () => {
-  const browserWindow: BrowserWindow = window
+  const browserWindow = window as BrowserWindow
```
thomas-lebeau, line 10: **❓ question:** ‏not sure if it's intended, but sometimes you use `window as BrowserWindow` and sometimes `globalThis as BrowserWindow`. Any specific reason not to use globalObject everywhere?

packages/rum/src/domain/profiling/profilingSupported.ts
thomas-lebeau, line 6: **💬 suggestion:** ‏what about adding `Profiler` to `globalObject` type definition, then we could simplify this

```ts
export function isProfilingSupported(): boolean {
  return globalObject.Profiler !== undefined
}
```

packages/rum/src/entries/main.ts
```diff
@@ -95,7 +95,7 @@ export const datadogRum = makeRumPublicApi(recorderApi, profilerApi, {
   sdkName: 'rum',
 })
 
-interface BrowserWindow extends Window {
+interface BrowserWindow {
```
thomas-lebeau, line 98: **💭 thought:** we have multiple `BrowserWindow` types defined differently, I wonder if we should add `DD_RUM` and other globals to `globalObject` to avoid all these casting

packages/rum/test/mockProfiler.ts
```diff
@@ -1,9 +1,9 @@
 import { registerCleanupTask } from '@datadog/browser-core/test'
-import { getGlobalObject } from '@datadog/browser-core'
+import { globalObject } from '@datadog/browser-core'
 import type { Profiler, ProfilerTrace, ProfilerInitOptions } from '../src/domain/profiling/types'
 
 export function mockProfiler(mockedTrace: ProfilerTrace) {
-  const globalThis = getGlobalObject()
+  const globalThis = globalObject
   // Save original Profiler class to restore it during cleanup.
   const originalProfiler = (globalThis as any).Profiler
```
thomas-lebeau:
```suggestion
// Save original Profiler class to restore it during cleanup.
const originalProfiler = globalObject.Profiler
```

test/e2e/lib/types/global.ts
```diff
@@ -1,11 +1,11 @@
...
 import type { DatadogDebugger } from '@datadog/browser-debugger'
-import type { RumGlobal } from '@datadog/browser-rum'
+import type { DatadogRum } from '@datadog/browser-rum'
 
 declare global {
   interface Window {
-    DD_LOGS?: LogsGlobal
-    DD_RUM?: RumGlobal
+    DD_LOGS?: DatadogLogs
+    DD_RUM?: DatadogRum
```
thomas-lebeau, line 8: **👏 praise:** ‏nice, thanks!

test/e2e/scenario/logs.scenario.ts
```diff
@@ -22,7 +19,7 @@ test.describe('logs', () => {
         test.skip(browserName !== 'chromium', 'Non-Chromium browsers do not support ES modules in Service Workers')
 
         await evaluateInWorker(() => {
-          DD_LOGS!.logger.log('Some message')
+          self.DD_LOGS!.logger.log('Some message')
```
thomas-lebeau, line 22: **🥜 nitpick:** ‏use `window` like all other test in this file. (or use `self` everywhere

packages/core/src/tools/globalObject.ts
BenoitZugmeyer: Good call

test/e2e/scenario/logs.scenario.ts
BenoitZugmeyer: It's in the worker, we don't have `window` here. I think it's fine to use `self` in worker and `window` elsewhere

packages/rum/src/entries/main.ts
BenoitZugmeyer: I think *some* casting is fine, so we don't have to pull all domain types in core.

packages/debugger/src/entries/main.spec.ts
BenoitZugmeyer: I don't have a specific reason. I think we used to use `window` before because `globalThis` was kind of new.

I improved the situation a bit to make sure we always favor `globalObject` instead of `globalThis`.

Although, I don't want to enforce `globalObject` in favor of `window`, as:
* there are 300+ `window` references in the codebase
* there are subtle differences between `window` and `globalThis`, so I don't want to risk a migration now (see https://github.com/DataDog/browser-sdk/blob/27a74a2cd2abb17c712de4d2d5ba5448e1617dbb/packages/core/src/tools/queueMicrotask.ts#L5-L9)

As we are trying to support environments without `window`, I think we'll use `globalObject` more and more in the future, and we might enforce it at some point.

thomas-lebeau (approved):

mormubis (approved):

PR #4703, opened 2026-06-01
✨RUM-16635 Add wildcard host pattern matching to WebView event bridge

barboraplasovska:
## Motivation

Part of [RUM-15826](https://datadoghq.atlassian.net/browse/RUM-15826). iOS SDK counterpart https://github.com/DataDog/dd-sdk-ios/pull/2963.

## Changes

 Adds a new optional `getAllowedWebViewHostPatterns?()` method to the `DatadogEventBridge` interface. When present, `canUseEventBridge()` uses wildcard pattern matching via plain string operations instead of the legacy exact + subdomain-suffix path. 
 
Invalid patterns (more than one `*`) produce a `display.error` console warning so customers see an actionable signal. Old Browser SDK versions receiving wildcard entries from a new mobile SDK will silently ignore them (literal match fails) while plain host entries continue to bridge normally.

## Test instructions

CI should pass.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

[RUM-15826]: https://datadoghq.atlassian.net/browse/RUM-15826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/browser-core/src/transport/eventBridge.spec.ts
+++ b/packages/browser-core/src/transport/eventBridge.spec.ts
@@ -1,7 +1,7 @@
 import { mockEventBridge } from '../../test'
 import { DefaultPrivacyLevel } from '../domain/configuration'
 import type { BrowserWindowWithEventBridge } from './eventBridge'
-import { getEventBridge, canUseEventBridge, BridgeCapability, bridgeSupports } from './eventBridge'
+import { getEventBridge, canUseEventBridge, matchesWildcardPattern, BridgeCapability, bridgeSupports } from './eventBridge'
 
 describe('canUseEventBridge', () => {
   const allowedWebViewHosts = ['foo.bar']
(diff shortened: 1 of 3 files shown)


barboraplasovska:

packages/browser-core/src/transport/eventBridge.ts
```diff
@@ -53,13 +53,28 @@ export function canUseEventBridge(currentHost = globalObject.location?.hostname)
...
+  }
+  const parts = pattern.split('*')
+  if (parts.length !== 2) {
+    return false
+  }
+  const [prefix, suffix] = parts
+  return host.length >= prefix.length + suffix.length && host.startsWith(prefix) && host.endsWith(suffix)
+}
+
 function getEventBridgeGlobal() {
```
mormubis, line 71: The RFC says invalid patterns (more than one `*`) should produce a `display.error` warning so customers get a console signal when a pattern is misconfigured. Right now it silently returns `false`, which matches the PR description but deviates from what was spec'd. Should we stick to the RFC?

mormubis (changes requested):

packages/browser-core/src/transport/eventBridge.ts
barboraplasovska: My mistake! Will fix it

packages/browser-core/src/transport/eventBridge.spec.ts
```diff
@@ -31,6 +38,58 @@ describe('canUseEventBridge', () => {
...
+  const cases: Array<[string, string, boolean]> = [
+    // [host,                          pattern,                  expected]
+    ['app.shopist.io',                 '*.shopist.io',           true],
+    ['preview-abc.shopist.io',         '*.shopist.io',           true],
+    ['shopist.io',                     '*.shopist.io',           false], // apex not matched by subdomain wildcard
+    ['shopist.io',                     'shopist.io',             true],  // exact match
+    ['app.shopist.io',                 'shopist.io',             false], // exact only, no subdomain
+    ['preview-abc123.shopist.io',      'preview-*.shopist.io',   true],
+    ['app.shopist.io',                 'preview-*.shopist.io',   false],
+    ['evil.com',                       '*.shopist.io',           false],
```
mormubis, line 46:
```suggestion
    ['evil.com',                       '*.shopist.io',           false],
    ['app.shopist.io.evil.com',                       '*.shopist.io',                       false],
```

mormubis (approved):

packages/browser-core/src/transport/eventBridge.ts
```diff
@@ -53,13 +54,30 @@ export function canUseEventBridge(currentHost = globalObject.location?.hostname)
   const bridge = getEventBridge()
 
   return (
-    !!bridge &&
-    bridge
-      .getAllowedWebViewHosts()
-      .some((allowedHost) => currentHost === allowedHost || currentHost.endsWith(`.${allowedHost}`))
+    bridge?.getAllowedWebViewHosts().some((entry) => {
```
gogusarov: `suggestion` 

we currently have a couple of duplicating checks both inside `some` iterator and inside `matchesWildcardPattern` function. I suggest to combine them into one function and use early returns for readability

```
export function canUseEventBridge(currentHost = globalObject.location?.hostname): boolean {
    return bridge?.getAllowedWebViewHosts().some((entry) =>
  matchesHostEntry(currentHost, entry)) ?? false
}

export function matchesHostEntry(host: string, entry: string): boolean
  {
    if (!entry.includes('*')) {
      return host === entry || host.endsWith(`.${entry}`)
    }

    const parts = entry.split('*')
    if (parts.length !== 2) {
      display.error(`Invalid WebView host pattern "${entry}": only one
  wildcard (*) is supported.`)
      return false
    }

    const [prefix, suffix] = parts

    return host.length >= prefix.length + suffix.length &&
  host.startsWith(prefix) && host.endsWith(suffix)
}

```

gogusarov (approved):

PR #4714, opened 2026-06-03
⚗️ Add instrumentConstructor utility

bdibon:
## Motivation

In order to instrument WebSocket connections, we need to wrap the `WebSocket` constructor, so far the `instrumentMethod` only supports instrumenting regular methods. This PR adds another utility `instrumentConstructor` supporting the instrumentation of constructor functions.

## Changes

- Add an `instrumentConstructor` utility that supports instrumenting constructor functions
- Factorize logic between `instrumentMethod` and new `instrumentConstructor`utils
- Add exhaustive unit tests for `instrumentConstructor`

## Test instructions

`yarn test:unit --spec packages/core/src/tools/instrumentMethod.spec.ts`

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-core/src/tools/instrumentMethod.spec.ts
+++ b/packages/browser-core/src/tools/instrumentMethod.spec.ts
@@ -199,6 +199,8 @@ describe('instrumentMethod', () => {
 })
 
 describe('instrumentConstructor', () => {
+  const THIRD_PARTY_CONSTRUCTOR_TAG = 42
+
   class MyClass {
     value: number
     constructor(value: number) {
@@ -237,6 +239,36 @@ describe('instrumentConstructor', () => {
     })
     expect(instance.value).toBe(42)
     expect(instance.getValue()).toBe(42)
+  })
+
+  it('allows other instrumentations from third parties', () => {
+    const container = { MyClass }
+    const instrumentationSpy = jasmine.createSpy()
+    instrumentConstructor(container, 'MyClass', instrumentationSpy)
+
+    thirdPartyConstructorWrap(container)
+
+    const instance = new container.MyClass(7) as MyClass & { thirdPartyTag: number }
+
+    expect(instance.value).toBe(7)
+    expect(instance.thirdPartyTag).toBe(THIRD_PARTY_CONSTRUCTOR_TAG)
+    expect(instrumentationSpy).toHaveBeenCalled()
+  })
+
+  it('computes the handling stack', () => {
+    const container = { MyClass }
+    const instrumentationSpy = jasmine.createSpy()
+    instrumentConstructor(container, 'MyClass', instrumentationSpy, { computeHandlingStack: true })
+
+    function foo() {
+      new container.MyClass(1)
+    }
+
+    foo()
+
+    expect(instrumentationSpy.calls.mostRecent().args[0].handlingStack).toEqual(
+      jasmine.stringMatching(/^HandlingStack: instrumented constructor\n {2}at foo @/)
+    )
   })
 
   it('does not call the instrumentation when the constructor is invoked without new', () => {
@@ -412,6 +444,37 @@ describe('instrumentConstructor', () => {
       expect(instance.value).toBe(5)
     })
 
+    describe('when the constructor has been instrumented by a third party', () => {
+      it('should preserve third party instrumentation after stop()', () => {
+        const container = { MyClass }
+        const { stop } = instrumentConstructor(container, 'MyClass', noop)
+
+        thirdPartyConstructorWrap(container)
+        const wrappedConstructor = container.MyClass
+
+        stop()
+
+        expect(container.MyClass).toBe(wrappedConstructor)
+      })
+
+      it('does not call the instrumentation when constructing after stop()', () => {
+        const container = { MyClass }
+        const instrumentationSpy = jasmine.createSpy()
+        const { stop } = instrumentConstructor(container, 'MyClass', instrumentationSpy)
+
+        thirdPartyConstructorWrap(container)
+
+        stop()
+
+        const instance = new container.MyClass(5) as MyClass & { thirdPartyTag: number }
+
+        expect(instrumentationSpy).not.toHaveBeenCalled()
+        expect(instance.value).toBe(5)
+        expect(instance.thirdPartyTag).toBe(THIRD_PARTY_CONSTRUCTOR_TAG)
+        expect(instance instanceof MyClass).toBeTrue()
+      })
+    })
+
     it('preserves new.target when constructing a subclass of the instrumented constructor after stop()', () => {
       const container = { MyClass }
       const { stop } = instrumentConstructor(container, 'MyClass', noop)
@@ -491,6 +554,22 @@ describe('instrumentConstructor', () => {
       expect(instance.x).toBe(2)
     })
   })
+
+  /**
+   * Wraps the current `MyClass` on `container` (typically the SDK-instrumented constructor) in a
+   * Proxy that still delegates construction, then tags instances — analogous to
+   * `thirdPartyInstrumentation` in the `instrumentMethod` suite.
+   */
+  function thirdPartyConstructorWrap(container: { MyClass: typeof MyClass }) {
+    const inner = container.MyClass
+    container.MyClass = new Proxy(inner, {
+      construct(target, argArray, newTarget) {
+        const instance = Reflect.construct(target, argArray, newTarget) as MyClass & { thirdPartyTag: number }
+        instance.thirdPartyTag = THIRD_PARTY_CONSTRUCTOR_TAG
+        return instance
+      },
+    })
+  }
 })
 
 describe('instrumentSetter', () => {
--- a/packages/browser-core/src/tools/instrumentMethod.spec.ts
+++ b/packages/browser-core/src/tools/instrumentMethod.spec.ts
@@ -180,6 +180,72 @@ describe('instrumentMethod', () => {
     })
   })
 
+  describe('constructor support', () => {
+    class MyClass {
+      value: number
+      constructor(value: number) {
+        this.value = value
+      }
+
+      getValue() {
+        return this.value
+      }
+    }
+
+    it('works when the instrumented method is called with new', () => {
+      const container = { MyClass }
+      const instrumentationSpy = jasmine.createSpy()
+      instrumentMethod(container, 'MyClass', instrumentationSpy)
+
+      const instance = new container.MyClass(42)
+
+      expect(instrumentationSpy).toHaveBeenCalledOnceWith({
+        target: jasmine.any(MyClass),
+        parameters: [42],
+        onPostCall: jasmine.any(Function),
+        handlingStack: undefined,
+      })
+      expect(instance.value).toBe(42)
+      expect(instance.getValue()).toBe(42)
+    })
+
+    it('preserves instanceof on the constructed instance', () => {
+      const container = { MyClass }
+      instrumentMethod(container, 'MyClass', noop)
+
+      const instance = new container.MyClass(1)
+
+      expect(instance instanceof container.MyClass).toBeTrue()
+    })
+
+    it('passes the constructed instance to onPostCall', () => {
+      const container = { MyClass }
+      const postCallCallbackSpy = jasmine.createSpy()
+      instrumentMethod(container, 'MyClass', ({ onPostCall }) => onPostCall(postCallCallbackSpy))
+
+      const instance = new container.MyClass(7)
+
+      expect(postCallCallbackSpy).toHaveBeenCalledOnceWith(instance)
+      expect((postCallCallbackSpy.calls.mostRecent().args[0] as MyClass).value).toBe(7)
+    })
+
+    it('works correctly after stop() when a third party has re-instrumented', () => {
+      const container = { MyClass }
+      const { stop } = instrumentMethod(container, 'MyClass', noop)
+
+      // Third party wraps our instrumentation
+      const OurInstrumentation = container.MyClass
+      container.MyClass = class extends OurInstrumentation {}
+
+      stop()
+
+      const instance = new container.MyClass(99)
+
+      expect(instance instanceof MyClass).toBeTrue()
+      expect(instance.value).toBe(99)
+    })
+  })
+
   function thirdPartyInstrumentation(object: { method?: () => number; onevent?: () => void }) {
     const originalMethod = object.method
     if (typeof originalMethod === 'function') {
--- a/packages/browser-core/src/tools/instrumentMethod.spec.ts
+++ b/packages/browser-core/src/tools/instrumentMethod.spec.ts
@@ -511,6 +511,26 @@ describe('instrumentConstructor', () => {
       const instance = new container.MyClass(2)
       expect(instance.constructor).toBe(MyClass)
     })
+
+    it('does not clobber prototype.constructor when a third party repoints it before stop()', () => {
+      const container = { MyClass }
+      const { stop } = instrumentConstructor(container, 'MyClass', noop)
+
+      function thirdPartyCanonicalCtor() {
+        /* third party keeps identity checks aligned with their wrapper */
+      }
+
+      // Simulate another layer repointing the shared prototype after our instrumentation.
+      const constructorDescriptor = Object.getOwnPropertyDescriptor(MyClass.prototype, 'constructor')!
+      Object.defineProperty(MyClass.prototype, 'constructor', {
+        ...constructorDescriptor,
+        value: thirdPartyCanonicalCtor,
+      })
+
+      stop()
+
+      expect(MyClass.prototype.constructor).toBe(thirdPartyCanonicalCtor)
+    })
   })
 
   /**
--- a/packages/browser-core/src/tools/instrumentMethod.spec.ts
+++ b/packages/browser-core/src/tools/instrumentMethod.spec.ts
@@ -1,4 +1,4 @@
-import { mockClock, mockZoneJs } from '../../test'
+import { mockClock, mockZoneJs, registerCleanupTask } from '../../test'
 import type { Clock, MockZoneJs } from '../../test'
 import type { InstrumentedMethodCall } from './instrumentMethod'
 import { instrumentConstructor, instrumentMethod, instrumentSetter } from './instrumentMethod'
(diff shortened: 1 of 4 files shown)


bdibon:
* **#4718** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4718?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4717** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4716** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4716?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4715** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4714** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4714?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -3,6 +3,17 @@ import { callMonitored } from './monitor'
 import { noop } from './utils/functionUtils'
 import { createHandlingStack } from './stackTrace/handlingStack'
 
+/* Resolves parameters from either a call or construct signature, so instrumentMethod can handle
+ * both regular methods and constructors (e.g. `new WebSocket()`).
```
bcaudan: **🥜 nitpick:** ‏What about replacing `WebSocket` references by something like `MyClass`?
No need to reference the usages in utility functions.

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -85,13 +96,16 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
 
   let stopped = false
 
-  const instrumentation = function (this: TARGET): ReturnType<TARGET[METHOD]> {
+  const instrumentation = function (this: TARGET): MethodReturnType<TARGET[METHOD]> {
     if (stopped) {
+      if (new.target) {
+        return Reflect.construct(original, arguments as unknown as MethodParameters<TARGET[METHOD]>)
```
bcaudan: **❓ question:** ‏the other construct call uses a third parameter `new.target`, should it be used in this call as well?

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -18,7 +29,7 @@ export interface InstrumentedMethodCall<TARGET extends { [key: string]: any }, M
    *
    * Note: if needed, parameters can be mutated by the instrumentation
    */
-  parameters: Parameters<TARGET[METHOD]>
+  parameters: MethodParameters<TARGET[METHOD]>
```
bcaudan: **❓ question:** ‏the `TARGET[METHOD]` type or even the `target` property do not really make sense in a constructor context, no?
Could it be worth considering an `instrumentConstructor` function instead with some shared helpers?

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -117,6 +131,32 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
   }
 
+  // Preserve the original prototype so that instanceof checks against the instrumented global work.
+  // e.g. `new window.WebSocket(url) instanceof window.WebSocket` must remain true.
+  if (typeof original === 'function' && original.prototype) {
+    instrumentation.prototype = original.prototype
+  }
+
+  // Preserve static members (class fields, static methods, etc.) on the instrumented function.
+  if (typeof original === 'function') {
```
bcaudan: **🥜 nitpick:** ‏what about moving the prototype logic inside the other function guard below?

packages/browser-core/src/tools/instrumentMethod.ts
bdibon: They don't in a regular constructor context, but all the constructor we're likely to instrument will be instrumented on the globalObject (window.WebSocket for example).

I will try to make this split and see if it reads better.

packages/browser-core/src/tools/instrumentMethod.ts
bdibon: No, this is called when the instrumentation has been stopped, but the consumer might still have a reference to it, then we want to behave as if we're not here.

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -130,6 +238,76 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+    return
+  }
+
+  // Preserve the original prototype so that instanceof checks against the instrumented global work.
+  // e.g. `new MyClass() instanceof MyClass` must remain true.
+  instrumentation.prototype = original.prototype
+
+  // Preserve static members (class fields, static methods, etc.) on the instrumented function.
+  const excludedKeys = new Set(['prototype', 'length', 'name', 'arguments', 'caller'])
+  for (const key of Object.getOwnPropertyNames(original)) {
```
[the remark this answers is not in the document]
bdibon: Good point, but the usage of `instrumentConstructor` is the instrumentation of `WebSocket` and it doesn't have any inherited static members.

Plus the fix would be to use `Object.setPrototypeOf` which might degrade performance.

I've chosen to document the limitation.

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -83,41 +112,124 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+  if (typeof original !== 'function') {
+    return { stop: noop }
+  }
+
+  return replaceWithInstrumentation(target, name, original, (isStopped) => {
+    const instrumentation = function (this: TARGET): ConstructorInstanceOf<TARGET[NAME]> {
+      // When `new` is used on this instrumented property, `new.target` is this wrapper. Passing
+      // it through to Reflect.construct would expose the wrong new.target inside the original
+      // body. If a subclass extends the wrapper, or a third party wraps us and their class is
+      // `new`ed, `new.target` is that outer constructor and must be preserved.
```
bcaudan: **🥜 nitpick:** it is what you mean, right?
‏
```suggestion
      // body. If a subclass extends the wrapper, or a third party wraps us and their class is
      // instantiated, `new.target` is that outer constructor and must be preserved.
```

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -83,41 +112,124 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
 
-    if (postCallCallback) {
-      callMonitored(postCallCallback, null, [result])
+      const parameters = Array.from(arguments) as ConstructorParametersOf<TARGET[NAME]>
+
+      return notifyInstrumentation<InstrumentedConstructorCall<TARGET[NAME]>, ConstructorInstanceOf<TARGET[NAME]>>(
+        onPreCall,
+        {
+          parameters,
+          handlingStack: computeHandlingStack ? createHandlingStack('instrumented method') : undefined,
```
bcaudan: **🥜 nitpick:** it could be handy for debugging‏

```suggestion
          handlingStack: computeHandlingStack ? createHandlingStack('instrumented constructor') : undefined,
```

packages/browser-core/src/tools/instrumentMethod.spec.ts
```diff
@@ -198,6 +198,190 @@ describe('instrumentMethod', () => {
   }
 })
 
+describe('instrumentConstructor', () => {
```
bcaudan, line 201: **💬 suggestion:** ‏comparing with instrumentMethod tests, it could be nice to have some equivalents on:
- computeHandlingStack
- third party compatibility

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -83,41 +112,124 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
-        },
-        handlingStack: computeHandlingStack ? createHandlingStack('instrumented method') : undefined,
-      },
-    ])
+  if (typeof original !== 'function') {
+    return { stop: noop }
+  }
+
+  return replaceWithInstrumentation(target, name, original, (isStopped) => {
+    const instrumentation = function (this: TARGET): ConstructorInstanceOf<TARGET[NAME]> {
```
bcaudan: **💬 suggestion:** ‏`this` is actually not `TARGET` here but since we don't use it, we could just remove it

```suggestion
    const instrumentation = function (): ConstructorInstanceOf<TARGET[NAME]> {
```

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -130,6 +242,83 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+  if (!original.prototype) {
+    return
+  }
+
+  // Preserve the original prototype so that instanceof checks against the instrumented global work.
+  // e.g. `new MyClass() instanceof MyClass` must remain true.
+  instrumentation.prototype = original.prototype
+
+  // Preserve own static members (class fields, static methods, etc.) on the instrumented function.
+  const excludedKeys = new Set(['prototype', 'length', 'name', 'arguments', 'caller'])
```
bcaudan: **💬 suggestion:** ‏Not sure to get why each of them are excluded, especially name/length, it could be worth adding a comment.

bdibon: Sure! It's because we don't want to overwrite the ones of the subclass.

bdibon: Edit: subclass in our case is the instrumentation and the goal is to make it look like it's the original, so nevermind, I will remove this.

packages/browser-core/src/tools/instrumentMethod.ts
bdibon: For the types, I've decided to rename `TARGET` to `CONTAINER` which is more accurate in our context (instrumenting global constructors).

I am using `this` when I replicate the native behavior for constructor called without `new` (i.e runtime throws an exception). I probably could have used `undefined` instead, but I preferred to be defensive and not to temper with what the caller of the instrumentation was doing. In this context, `this` is unknown.

packages/browser-core/src/tools/instrumentMethod.ts
bdibon: I've removed these checks, I had somehow lost track that we're delivering an instrumentation that should be as close as possible to the original.

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -130,6 +265,87 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+   * acceptable for the globals we instrument (e.g. `WebSocket`, whose own statics are the only ones
+   * that matter), but we would need to delegate the static prototype chain to support subclassed
+   * constructors with meaningful inherited statics.
+   */
+  for (const key of ([] as PropertyKey[]).concat(
+    Object.getOwnPropertyNames(original),
+    Object.getOwnPropertySymbols(original)
+  )) {
+    const descriptor = Object.getOwnPropertyDescriptor(original, key) as PropertyDescriptor
+    Object.defineProperty(instrumentation, key, descriptor)
```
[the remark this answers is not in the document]
bdibon: This shouldn't be a problem:
- use case is pretty rare
- we'll run instrumentation early, and in most cases the SDK lives until the end

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -130,6 +265,87 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+
+  if (postCallCallback) {
+    callMonitored(postCallCallback, null, [result])
+  }
+
+  return result
+}
+
+interface InstrumentationCallbacks<RESULT> {
+  onPostCall: (callback: (result: RESULT) => void) => void
```
mormubis, line 295: After instrumentation, `instance.constructor === window.WebSocket` returns `false` because `prototype.constructor` still points to the original. I can't think of real-world code that checks this, so it's probably fine, but wanted to confirm it's a known acceptable limitation. What do you think?

bdibon: You are right for instances that have been constructed before instrumentation, I have documented it [here](https://github.com/DataDog/browser-sdk/blob/8d89c4834ebf130d0d6cddfc4ad37c5055cdd788/packages/browser-core/src/tools/instrumentMethod.ts#L330-L340).

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -83,41 +114,145 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+ *  instrumentConstructor(window, 'WebSocket', ({ parameters, onPostCall }) => {
+ *    console.log('Before constructing WebSocket with parameters', parameters)
+ *
+ *    onPostCall((instance) => {
+ *      console.log('Constructed WebSocket instance', instance)
+ *    })
+ *  })
+ */
+export function instrumentConstructor<CONTAINER extends { [key: string]: any }, CONSTRUCTOR extends keyof CONTAINER>(
+  container: CONTAINER,
```
bcaudan, line 175: **💬 suggestion:** ‏is there a difference between `CONTAINER` and `TARGET`? if not, let's use the same term

bcaudan (approved):

packages/browser-core/src/tools/instrumentMethod.ts
bdibon: It's semantics.

With `instrumentMethod` the target will also be the `this` value of instrumentation / original, so `this` is `TARGET`, I can wrap my head around this one.

With `instrumentConstructor`, in the normal usage (when it's called with `new`), the `this` value refers to the instance being constructed, "target" feels weird in this case, since we're really interested in the (global) object holding the reference to the function we're instrumenting.

Wdyt?

bcaudan: Could it make sense to use `CONTAINER` for both `instrumentMethod` and `instrumentConstructor`? 
Since the type parameter describes the object holding the reference and the `this` value is already surfaced through `target` in `InstrumentedMethodCall` for callers who need it.

bdibon: I would leave it as is, it makes sense to have targetPrototype / TARGET for instrumentMethod and container / CONTAINER for instrumentConstructor.

I'll wait for a third pair of eyes to come by.

packages/browser-core/src/tools/instrumentMethod.spec.ts
```diff
@@ -198,6 +198,358 @@ describe('instrumentMethod', () => {
...
+
+  beforeEach(() => {
+    // Several tests instrument without calling `stop()`; they share this `MyClass` and leave
+    // `MyClass.prototype.constructor` pointing at a stale wrapper, which breaks random-order runs.
+    Object.defineProperty(MyClass.prototype, 'constructor', {
+      configurable: initialMyClassPrototypeConstructor.configurable!,
+      enumerable: initialMyClassPrototypeConstructor.enumerable!,
+      writable: initialMyClassPrototypeConstructor.writable,
+      value: MyClass,
+    })
```
BenoitZugmeyer: **suggestion:** just re-define the class in `beforeEach`.

```ts
MyClass = class {
 ...
}
```

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -83,41 +114,145 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
     }
   }
 
-  let stopped = false
+  return replaceWithInstrumentation(
```
BenoitZugmeyer: **suggestion:** slight suggestion, but I find the callback nesting a bit hard to follow, and maybe if we don't factorize `isStopped` between `instrumentMethod` and `instrumentConstructor` it could be easier to read, see [this](https://github.com/DataDog/browser-sdk/commit/f244faf010cbb1812c0682f0543f60b52d43b811). Bonus: you can use `replaceWithInstrumentation` to patch the prototype constructor

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -83,41 +114,145 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
-        parameters,
-        onPostCall: (callback) => {
-          postCallCallback = callback
-        },
-        handlingStack: computeHandlingStack ? createHandlingStack('instrumented method') : undefined,
-      },
-    ])
+  let restorePrototypeConstructor: () => void = noop
+
+  const { stop: replaceStop } = replaceWithInstrumentation(container, constructor, original, (isStopped) => {
```
BenoitZugmeyer: **suggestion:** rename `replaceStop` with `restoreOriginal`

BenoitZugmeyer (commented):
I'm impressed!

packages/browser-core/src/tools/instrumentMethod.ts
```diff
@@ -130,6 +259,87 @@ export function instrumentMethod<TARGET extends { [key: string]: any }, METHOD e
...
+   * which narrows the window where pre-instrument instances exist.
+   */
+  const originalConstructor = original.prototype.constructor
+  original.prototype.constructor = instrumentation
+
+  return () => {
+    if (original.prototype.constructor === instrumentation) {
+      original.prototype.constructor = originalConstructor
+    }
+  }
```
BenoitZugmeyer: **nitpick:** you could use `replaceWithInstrumentation` here, [see](https://github.com/DataDog/browser-sdk/commit/f244faf010cbb1812c0682f0543f60b52d43b811#diff-039c831bb64c90da6f2c367869896d95e3a2f82ffe286d1edec58dbaa102828dR345-R350)

BenoitZugmeyer (approved):

PR #4717, opened 2026-06-03
✨ Add trackWebSockets configuration option (default: false)

bdibon:
## Motivation

Make WebSocket collection opt-in and gated behind an experimental feature flag.

## Changes

- Add the `trackWebSockets` init option (default `false`) to the RUM configuration.
- Add the `TRACK_WEB_SOCKETS` experimental feature flag.

## Test instructions

`yarn test:unit --spec packages/rum-core/src/domain/configuration/configuration.spec.ts`

## Checklist

- [ ] Tested locally
- [ ] Tested on staging
- [x] Added unit tests for this change.
- [ ] Added e2e/integration tests for this change.
- [ ] Updated documentation and/or relevant AGENTS.md file

PR #4716, opened 2026-06-03
🔧 Sync rum-events-format with WebSocket resource schema

bdibon:
## Motivation

Add the `websocket` resource type and its fields to the RUM event format so WebSocket connections can be reported as resource events.

> **Blocked:** depends on the upstream `rum-events-format` schema PR landing first. Kept in draft until then.

## Changes

- Bump the `rum-events-format` submodule to include the websocket resource sub-schema (`d6ba7f7` -> `a597e6d`).
- Regenerate `rumEvent.types.ts` and add the corresponding raw event shape in `rawRumEvent.types.ts`.
- Add `ResourceType.WEBSOCKET` to the core resource utils.

## Test instructions

`yarn typecheck`

## Checklist

- [ ] Tested locally
- [ ] Tested on staging
- [ ] Added unit tests for this change.
- [ ] Added e2e/integration tests for this change.
- [ ] Updated documentation and/or relevant AGENTS.md file

PR #4715, opened 2026-06-03
⚗️ Add webSocketObservable for WebSocket lifecycle events

bdibon:
## Motivation

Provide a low-level observable that emits WebSocket lifecycle events (connecting, open, message in/out, closed). This is the browser-level building block consumed by the future WebSocket collection layer.

## Changes

- Add `webSocketObservable` (`initWebSocketObservable`), instrumenting the `WebSocket` constructor and `send`, and listening to `open` / `message` / `close` / `error` events.
- Add matching unit tests.
- Add `WebSocketEventMap` support to `addEventListener`.
- Export the observable and its context types from `@datadog/browser-core`.

## Test instructions

`yarn test:unit --spec packages/core/src/browser/webSocketObservable.spec.ts`

## Checklist

- [ ] Tested locally
- [ ] Tested on staging
- [x] Added unit tests for this change.
- [ ] Added e2e/integration tests for this change.
- [ ] Updated documentation and/or relevant AGENTS.md file

PR #4718, opened 2026-06-03
⚗️ Collect WebSocket resource events

bdibon:
## Motivation

WebSocket connections are a blind spot in RUM resource monitoring. This PR adds WebSocket tracking behind the `track_websockets` experimental feature flag, collecting per-connection metrics (message counts, sizes, timing, handshake duration, close codes) and reporting them as `resource` events with `resource.type: "websocket"`.

> [!IMPORTANT]
> This PR introduces a prototype for collecting WebSocket data, the model is not stable, some fields might be removed / renamed or added in a future PR.

## Changes

**`packages/browser-core`**
- `webSocketObservable`: Instruments the `WebSocket` constructor and `send` method via `instrumentConstructor`/`instrumentMethod` to emit lifecycle events: `connecting`, `open`, `message-in`, `message-out`, `closed`. Tracks payload sizes and clock timestamps for each event.
- Exports the observable from the package index.
- Adds `ResourceType.WEBSOCKET = 'websocket'` to the resource type enum.
- Adds `ExperimentalFeature.TRACK_WEBSOCKETS = 'track_websockets'` experimental flag.

**`packages/browser-rum-core`**
- `webSocketCollection`: Consumes `webSocketObservable` and maintains an in-memory registry of open connections. Aggregates metrics (message counts/sizes, first-message offsets, longest inbound silence, buffered amount peaks) and emits `WEBSOCKET_COMPLETED` lifecycle events on close or session expiry.
- `resourceCollection`: Subscribes to `WEBSOCKET_COMPLETED` and assembles `RumResourceEvent` objects with `resource.type: "websocket"` and a `resource.websocket` sub-object containing the full connection metrics.
- `lifeCycle`: Adds the `WEBSOCKET_COMPLETED` event type and its payload type.
- `rawRumEvent.types`: Adds `WebSocketResourceProperties` interface and wires it into `RawRumResourceEvent`.
- `startRum`: Starts `webSocketCollection` when the `TRACK_WEBSOCKETS` experimental feature is enabled.

**`scripts/dev-server`**
- Adds `Access-Control-Allow-Origin: *` response header to support cross-origin WebSocket testing from the sandbox.

## Test instructions

Run unit tests:

```bash
yarn test:unit --spec packages/browser-core/src/browser/webSocketObservable.spec.ts
yarn test:unit --spec packages/browser-rum-core/src/domain/webSocketCollection.spec.ts
yarn test:unit --spec packages/browser-rum-core/src/domain/resource/resourceCollection.spec.ts
```

Manual end-to-end test:

```bash
# Terminal 1: Start dev server
yarn dev-server start

# Terminal 2: Spin up a local WebSocket echo server (uses the ws package already in the repo)
yarn node -e "
  const WebSocket = require('ws');
  const wss = new WebSocket.Server({ port: 8765 });
  wss.on('connection', ws => { ws.on('message', msg => ws.send(msg)); });
  console.log('WS echo server on :8765');
"
```

Open `http://localhost:8080` and replace `sandbox/index.html` with:

```html
<script>
  DD_RUM.init({
    applicationId: '...',
    clientToken: '...',
    trackResources: true,
    enableExperimentalFeatures: ['track_websockets'],
    proxy: '/proxy',
  })
</script>
<button onclick="
  let ws = new WebSocket('ws://localhost:8765');
  ws.onopen = () => { ws.send('hello'); ws.send('world'); };
  ws.onmessage = e => console.log('in:', e.data);
  setTimeout(() => ws.close(1000, 'done'), 2000);
">Test WebSocket</button>
```

Click the button, wait ~3 seconds, then open a new tab to flush events:

```bash
yarn dev-server intake rum-resources | jq 'select(.resource.type == "websocket")'
```

Expected output — a resource event with:
```json
{
  "resource": {
    "type": "websocket",
    "url": "ws://localhost:8765",
    "websocket": {
      "handshake_succeeded": true,
      "messages_in": { "count": 2, "size": 22 },
      "messages_out": { "count": 2, "size": 22 },
      "close_code": 1000,
      "was_clean": true,
      "tracking_end_reason": "close_event"
    }
  }
}
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging


bdibon:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4718?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#4718** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4718?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4718?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4717** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4716** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4716?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4715** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4714** <a href="https://app.graphite.com/github/pr/DataDog/browser-sdk/4714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -108,6 +113,55 @@ export function startResourceCollection(lifeCycle: LifeCycle, configuration: Rum
...
+  event: WebSocketCompleteEvent
+): RawRumEventCollectedData<RawRumResourceEvent> | undefined {
+  const duration = elapsed(event.startClocks.timeStamp, event.endClocks.timeStamp)
+
+  const rawRumEvent: RawRumResourceEvent = {
+    date: event.startClocks.timeStamp,
+    type: RumEventType.RESOURCE,
+    resource: {
+      id: generateUUID(),
+      type: ResourceType.WEBSOCKET,
```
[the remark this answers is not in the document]
bdibon: This is a prototype and the data collected is subject to change, downstream validation has been handled.

packages/browser-core/src/browser/webSocketObservable.ts
```diff
@@ -0,0 +1,184 @@
...
+        })
+      }
+    )
+
+    const { stop: stopInstrumentingSend } = instrumentMethod(
+      globalObject.WebSocket.prototype,
+      'send',
+      ({ target: instance, parameters: [data], onPostCall }) => {
+        const size = computePayloadSize(data)
+        const bufferedAmountPreSend = instance.bufferedAmount
```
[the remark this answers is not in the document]
bdibon: Sampling `bufferedAmountPreSend` allows us to spot back pressure issues, if we instrument the post-call path, we'll just see the `bufferedAmount` grow by the size of the message that has just been sent.

packages/browser-rum-core/src/boot/startRum.ts
```diff
@@ -230,6 +233,11 @@ export function startRumEventCollection(
 
   const vitalCollection = startVitalCollection(lifeCycle, pageStateHistory)
 
+  if (isExperimentalFeatureEnabled(ExperimentalFeature.TRACK_WEB_SOCKETS)) {
+    const webSocketCollection = startWebSocketCollection(lifeCycle, viewHistory, vitalCollection.addDurationVital)
+    cleanupTasks.push(webSocketCollection.stop)
```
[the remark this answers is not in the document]
bdibon: In practice, the `stop()` cleanup isn't ran.

Also, the only way `resourceCollection.stop()` prevents a websocket event from being collected is when there's a corresponding task in `taskQueue` which is an unlikely window.

packages/browser-core/src/browser/webSocketObservable.ts
```diff
@@ -0,0 +1,181 @@
...
+      ({ parameters, onPostCall }) => {
+        const url = String(parameters[0])
+        const protocols = parameters[1]
+        const startClocks = clocksNow()
+        onPostCall((instance) => {
+          observable.notify({
+            state: 'connecting',
+            instance,
+            url,
+            ...(protocols !== undefined ? { protocols } : {}),
```
bcaudan: **💬 suggestion:** or is it needed?‏

```suggestion
            protocols,
```

Same for similar patterns.

packages/browser-core/src/browser/webSocketObservable.ts
```diff
@@ -0,0 +1,181 @@
...
+    })
+
+    stopMessage()
+    stopClose()
+  })
+}
+
+function computePayloadSize(data: unknown): number {
+  if (typeof data === 'string') {
+    return new TextEncoder().encode(data).byteLength
```
bcaudan: **💬 suggestion:** ‏what about using `computeBytesCount` utility?

packages/browser-core/src/browser/webSocketObservable.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+  if (data instanceof ArrayBuffer) {
+    return data.byteLength
+  }
+  if (ArrayBuffer.isView(data)) {
+    return data.byteLength
+  }
```
bcaudan: **🥜 nitpick:** ‏what about grouping similar returns?

```suggestion
  if (data instanceof ArrayBuffer || ArrayBuffer.isView(data)) {
    return data.byteLength
  }
```

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
bcaudan, line 1: **💬 suggestion:** ‏what about moving this file and its tests to the resource directory?

packages/browser-core/src/domain/resourceUtils.ts
```diff
@@ -8,6 +8,7 @@ export const ResourceType = {
   IMAGE: 'image',
   FONT: 'font',
   MEDIA: 'media',
+  WEBSOCKET: 'websocket',
```
bcaudan, line 11: **💬 suggestion:** ‏It could be worth adding this type in the dev extension as well

Image

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -108,6 +113,56 @@ export function startResourceCollection(lifeCycle: LifeCycle, configuration: Rum
...
+    _dd: {
+      discarded: !configuration.trackResources,
+    },
+  }
+
+  return {
+    startClocks: event.startClocks,
+    duration,
+    rawRumEvent,
+    domainContext: {},
```
bcaudan: **💭 thought:** ‏could it be worth providing the web socket instance in the domain context?
if yes, it could be added later down the road

packages/browser-rum-core/src/rawRumEvent.types.ts
```diff
@@ -86,6 +87,29 @@ export interface RawRumResourceEvent {
...
+  close_reason?: string
+  was_clean?: boolean
+  messages_in: { count: number; size: number }
+  messages_out: { count: number; size: number }
+  time_to_first_message_in?: ServerDuration
+  time_to_first_message_out?: ServerDuration
+  last_message_in_at?: TimeStamp
+  longest_inbound_silence: ServerDuration
+  inbound_idle_duration_before_close?: ServerDuration
+  buffered_amount_max?: number
```
bcaudan: **💬 suggestion:** it seems to always be defined

‏
```suggestion
  buffered_amount_max: number
```

packages/browser-core/src/index.ts
```diff
@@ -126,6 +126,15 @@ export type { XhrCompleteContext, XhrStartContext, XhrContext } from './browser/
 export { initXhrObservable } from './browser/xhrObservable'
 export type { FetchResolveContext, FetchStartContext, FetchContext } from './browser/fetchObservable'
 export { initFetchObservable, ResponseBodyAction } from './browser/fetchObservable'
+export type {
+  WebSocketContext,
+  WebSocketConnectingContext,
+  WebSocketOpenContext,
+  WebSocketMessageInContext,
+  WebSocketMessageOutContext,
+  WebSocketClosedContext as WebSocketCloseContext,
```
bcaudan: **💬 suggestion:** we should keep the same naming‏

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,251 @@
...
+export function startWebSocketCollection(
+  lifeCycle: LifeCycle,
+  viewHistory: ViewHistory,
+  addDurationVital: (vital: DurationVital) => void
+) {
+  const tracker = trackWebSocket(lifeCycle, initWebSocketObservable(), viewHistory, addDurationVital)
+
+  const sessionExpiredSubscription = lifeCycle.subscribe(LifeCycleEventType.SESSION_EXPIRED, () => {
+    tracker.flushOpenConnections('session_end')
+  })
```
bcaudan, line 75: **❓ question:** ‏what do we want to do for websocket connection that spans across multiple sessions?

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,251 @@
...
+  bufferedAmountMax: number
+  setupDuration?: Duration
+}
+
+export interface WebSocketConnectionTracker {
+  flushOpenConnections: (reason: WebSocketTrackingEndReason) => void
+  stop: () => void
+}
+
+export function startWebSocketCollection(
```
bcaudan, line 63: **💬 suggestion:** ‏IMO, we should have some e2e tests to:
- exercise a real instrumentation
- assert the expected events are emitted: vital + resource
- ensure it reacts as intended to sdk lifecycle (start/end view, session expiration/reewal)

packages/browser-core/src/browser/webSocketObservable.ts
bdibon: I will remove the usage of this pattern, the idea was to not pass objects with `[key, undefined]` entries, but:
- this is not dramatic as it gets stripped out by the `sanitize()` function in the end,
- and it's never used in the codebase.

bdibon: However, for this specific line, I will make a shallowCopy of the array before passing it around as a safety measure so we don't risk mutating it.

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
bdibon: Should we move `packages/browser-rum-core/src/domain/requestCollection.ts` as well?

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
bdibon: I don't see any value right now. What do you have in mind?

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
bdibon: For the prototype we'll just cut the tracking with the reason "session_end".

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,251 @@
...
+  lifeCycle: LifeCycle,
+  viewHistory: ViewHistory,
+  addDurationVital: (vital: DurationVital) => void
+) {
+  const tracker = trackWebSocket(lifeCycle, initWebSocketObservable(), viewHistory, addDurationVital)
+
+  const sessionExpiredSubscription = lifeCycle.subscribe(LifeCycleEventType.SESSION_EXPIRED, () => {
+    tracker.flushOpenConnections('session_end')
```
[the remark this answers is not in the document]
bdibon: We will address this in a future PR.

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -108,6 +113,56 @@ export function startResourceCollection(lifeCycle: LifeCycle, configuration: Rum
...
+        was_clean: event.wasClean,
+        messages_in: event.messagesIn,
+        messages_out: event.messagesOut,
+        time_to_first_message_in: toServerDuration(event.firstMessageInOffset),
+        time_to_first_message_out: toServerDuration(event.firstMessageOutOffset),
+        last_message_in_at: event.lastMessageInAt,
+        longest_inbound_silence: toServerDuration(event.longestInboundSilence),
+        inbound_idle_duration_before_close: toServerDuration(event.inboundIdleDurationBeforeClose),
+        buffered_amount_max: event.bufferedAmountMax,
+        protocol: event.protocol,
```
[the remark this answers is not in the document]
bdibon: The `resource.websocket.protocol` is supposed to contain a protocol name which is benign. Now if the client uses the `protocols` argument of the `WebSocket` constructor to smuggle auth tokens and then the server echoes it back to the client instead of the chosen protocol, the legitimate fix would be to fix it on the server.

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,251 @@
...
+        addDurationVital({
+          id: connectionId,
+          name: WEBSOCKET_CONNECTING_VITAL_NAME,
+          type: VitalType.DURATION,
+          startClocks: context.startClocks,
+          duration: 0 as Duration,
+          context: sanitize({
+            url: context.url,
+            protocols: context.protocols,
+            startViewId,
```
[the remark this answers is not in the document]
bdibon: It seems logic to disable websocket collection when trackResources is false, so I am enforcing this behavior now with an e2e test documenting it. However if / when we establish a link between RUM and APM websocket connections, we'll have to revert this.

packages/browser-rum-core/src/domain/lifeCycle.ts
```diff
@@ -22,6 +23,7 @@ export const enum LifeCycleEventType {
   AFTER_VIEW_ENDED,
   REQUEST_STARTED,
   REQUEST_COMPLETED,
+  WEBSOCKET_COMPLETED,
```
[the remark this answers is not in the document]
bdibon: This shouldn't be a problem as consumer packages have the same version, they would inline the correct values.

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,251 @@
...
+export function startWebSocketCollection(
+  lifeCycle: LifeCycle,
+  viewHistory: ViewHistory,
+  addDurationVital: (vital: DurationVital) => void
+) {
+  const tracker = trackWebSocket(lifeCycle, initWebSocketObservable(), viewHistory, addDurationVital)
+
+  const sessionExpiredSubscription = lifeCycle.subscribe(LifeCycleEventType.SESSION_EXPIRED, () => {
+    tracker.flushOpenConnections('session_end')
+  })
```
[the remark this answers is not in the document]
bdibon: Let's address this in an upcoming PR.

packages/browser-core/src/domain/resourceUtils.ts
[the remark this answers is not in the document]
bdibon: Adding `{ domainContext: { isManual: false } }` to raw websocket events.

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
bcaudan: I'd say yes

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
bcaudan: It is what we do with other events linked to a fetch or xhr instance for example.
It can allow customers to attach extra attributes that we don't collect automatically if they feel the need.

bdibon: There's only two fields that we don't capture right now: `extensions` and `binaryType`.

However the data model might change, so for the sake of parity. Also, there might be use cases for customers to attach their own computed attributes. Adding the instance!

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
bcaudan: Could we add some tests to enforce that we don't track events anymore after session renewal?
Maybe a comment as well could be handy.

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
bdibon: Only moved the webSocket files as requestCollection is used by other `domain/` modules i.e `waitPageActivityEnd`.

test/e2e/scenario/rum/websockets.scenario.ts
```diff
@@ -0,0 +1,217 @@
...
+      expect(ws.messages_in.count).toBe(1)
+      expect(ws.messages_in.size).toBe(KNOWN_IN_WS_MESSAGE.length)
+    })
+
+  createTest('websocket resource ends with close_event when the server closes the echo socket')
+    .withRum({ enableExperimentalFeatures: ['track_web_sockets'] })
+    .withBody(WEBSOCKET_TEST_BODY)
+    .run(async ({ intakeRegistry, flushEvents, page, servers }) => {
+      await page.locator('#ws-open').click()
+      await expect(page.locator('#ws-status')).toHaveText('open')
```
bcaudan: **💬 suggestion:** ‏it could be nice to extract a page object for this websocket body to allow to retrieve the different locator without scattering the different element ids across the scenarios.

Something like:

```ts
const { wsOpenButton, wsStatus } = await websocketPage(page)
await wsOpenButton.click()
await expect(page.locator(wsStatus).toHaveText('open')
```

test/e2e/scenario/rum/websockets.scenario.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function getWebSocketResources(events: RumResourceEvent[]): RumResourceEventWithWebSocket[] {
+  return events.filter(isWebSocketResource)
+}
+
+function getLastRumResourceEventWithWebSocket(events: RumResourceEvent[]): RumResourceEventWithWebSocket | undefined {
+  const list = getWebSocketResources(events)
+  return list.length === 0 ? undefined : list[list.length - 1]
+}
```
bcaudan, line 219: **💬 suggestion:** ‏It seems that we have this kind of helpers directly in intakeRegistry, any reason to do differently?

packages/browser-rum-core/src/domain/contexts/viewHistory.ts
```diff
@@ -18,7 +18,7 @@ export interface ViewHistoryEntry {
 }
 
 export interface ViewHistory {
-  findView: (startTime?: RelativeTime) => ViewHistoryEntry | undefined
+  findView: (startTime?: RelativeTime, options?: { returnInactive: boolean }) => ViewHistoryEntry | undefined
```
bcaudan, line 21: **💬 suggestion:** would it worth collecting the start view id at the beginning of the websocket instead of increasing the API?‏

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,254 @@
...
+}
+
+function recordMessageTiming(webSocket: WebSocketConnection, at: ClocksState, direction: 'in' | 'out') {
+  const handshakeFailed = webSocket.openClocks === undefined
+
+  if (handshakeFailed) {
+    return
+  }
+
+  const offset = elapsed(webSocket.openClocks!.timeStamp, at.timeStamp)
```
bcaudan: **💬 suggestion:** ‏non-null assertion after a null-check, what about extracting openClocks to allow typescript to correctly check that?

packages/browser-rum-core/src/domain/resource/webSocketCollection.spec.ts
```diff
@@ -0,0 +1,437 @@
...
+import { elapsed, relativeToClocks } from '@datadog/js-core/time'
+import { mockViewHistory } from '../../test'
+import { VitalType } from '../rawRumEvent.types'
+import type { ViewHistoryEntry } from './contexts/viewHistory'
+import { LifeCycle, LifeCycleEventType } from './lifeCycle'
+import type { DurationVital } from './vital/vitalCollection'
+import type { WebSocketCompleteEvent } from './webSocketCollection'
+import { startWebSocketCollection, trackWebSocket, WEBSOCKET_CONNECTING_VITAL_NAME } from './webSocketCollection'
+
+describe('webSocketCollection', () => {
```
bcaudan, line 14: **💬 suggestion:** ‏Could it worth having a test somewhere to ensure that it behaves correctly with multiple websockets?

test/e2e/scenario/rum/websockets.scenario.ts
bdibon: We don't have the same granularity, these utils help with websocket resources specifically, which is unlikely to be required by other test scenarios.

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -51,6 +53,9 @@ export function startResourceCollection(lifeCycle: LifeCycle, configuration: Rum
   const taskQueue = mockable(createTaskQueue)()
   const requestRegistry = createRequestRegistry(lifeCycle)
 
+  lifeCycle.subscribe(LifeCycleEventType.WEBSOCKET_COMPLETED, (event: WebSocketCompleteEvent) => {
+    handleResource(() => assembleWebSocketResource(event, configuration))
```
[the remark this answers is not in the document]
bdibon: Odds are very small for this to actually happen, plus we have the same problem for other resources.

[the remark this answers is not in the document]
bdibon: Edit: it can be consistently reproduced when the websocket is alive and the session is renewed, which might not be that rare. Adding a test case + a fix.

packages/browser-rum-core/src/domain/contexts/viewHistory.ts
bdibon: This is for the end_view_id, see https://github.com/DataDog/browser-sdk/pull/4718/changes#diff-3d1a097ccff522c83cf5a4696cd2de548b345f024819ed40bf6941e277eff188R87-R107

packages/browser-rum-core/src/domain/resource/webSocketCollection.spec.ts
bdibon: We have this https://github.com/DataDog/browser-sdk/pull/4718/changes#diff-0d6411be7f7527b6ab09c42ca6778a2eb2bb692590f2c9ef7863f124fe6d7699R145-R156 but the ws connections do not overlap.

I will add a more thorough test with overlapping connections.

packages/browser-rum-core/src/domainContext.types.ts
```diff
@@ -48,6 +49,12 @@ export interface RumManualResourceEventDomainContext {
   isManual: true
 }
 
+export interface RumWebSocketResourceEventDomainContext {
+  isManual: false
+  isWebSocket: true
+  webSocket: WebSocket
+}
```
bcaudan, line 56: **❗ warning:** like the other `DomainContext` types, this one should be exported from the different rum packages‏

test/e2e/lib/pages/webSocketPage.ts
bcaudan, line 1: **👏 praise:** ‏Nice!

test/e2e/scenario/rum/websockets.scenario.ts
```diff
@@ -0,0 +1,219 @@
...
+    .run(async ({ intakeRegistry, flushEvents, page }) => {
+      const ws = new WebSocketPage(page)
+
+      await ws.open()
+
+      await page.evaluate(() => {
+        window.DD_RUM!.stopSession()
+        // Generate user activity to trigger session renewal
+        document.documentElement.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }))
+      })
```
bcaudan, line 101: **💬 suggestion:** ‏could we use the `renewSession` helper directly?

packages/browser-rum-core/src/domain/contexts/viewHistory.ts
bcaudan: ok, got it

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+
+function buildCompletedEvent(
+  webSocket: WebSocketConnection,
+  endInfo: { at: ClocksState; code?: number; reason?: string; wasClean?: boolean },
+  trackingEndReason: WebSocketTrackingEndReason,
+  viewHistory: ViewHistory
+): WebSocketCompleteEvent {
+  const endClocks = endInfo.at
+  const endViewId =
```
bcaudan, line 238: **💬 suggestion:** ‏to clarify what is going on

```suggestion
  // On session_end, the websocket event can be received after the active view has been closed.
  const endViewId =
```

test/e2e/scenario/rum/websockets.scenario.ts
bdibon: No, it doesn't exercise the same race condition.

# Current Test

Here, we trigger a SESSION_EXPIRED lifecycle event, so websocketCollection emits a WebSocketCompleteEvent, which would then be dispatched asynchronously via the task queue (RED PATH). 

In parallel a click event is dispatched, leading to a SESSION_RENEWED being fired synchronously, which triggers the viewValueHistory to be reset. When websocket assembly task runs, it won't find any view matching the startTime of the event, so it will be silently dropped.

# renewSession

Calls expireSession which involves a similar code path, but this time the SESSION_RENEWED event is fired asynchronously which guarantees that the assembly task will have the time to run before the viewValueHistory is flushed.

bcaudan: no need to h1 😆
sounds good to me 👍

test/e2e/scenario/rum/websockets.scenario.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  createTest('websocket resource ends with close_event when the server closes the echo socket')
+    .withRum({ enableExperimentalFeatures: ['track_web_sockets'] })
+    .withBody(WebSocketPage.testBody())
+    .run(async ({ intakeRegistry, flushEvents, page, servers }) => {
+      const ws = new WebSocketPage(page)
+
+      await ws.open()
+
+      servers.base.app.closeEchoWebSockets!()
```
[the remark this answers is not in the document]
bdibon: It's ok, we don't expect any other test scenario to run tests involving websockets.

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
```diff
@@ -108,6 +118,60 @@ export function startResourceCollection(lifeCycle: LifeCycle, configuration: Rum
...
+        setup_duration: toServerDuration(event.setupDuration),
+      },
+    },
+    _dd: {
+      discarded: !configuration.trackResources,
+    },
+  }
+
+  return {
+    startClocks: event.startClocks,
```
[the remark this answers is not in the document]
bdibon: This is a legit concern, we will address this in a future PR.

bcaudan (approved):
Great work!

packages/browser-rum-core/src/domain/resource/resourceCollection.ts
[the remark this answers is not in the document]
bdibon: We'll address this in a future PR.

packages/browser-rum-core/src/domain/resource/webSocketCollection.ts
```diff
@@ -0,0 +1,260 @@
...
+        const webSocket = webSocketRegistry.get(context.instance)
+        if (!webSocket) {
+          return
+        }
+
+        webSocketRegistry.delete(context.instance)
+
+        lifeCycle.notify(
+          LifeCycleEventType.WEBSOCKET_COMPLETED,
+          buildCompletedEvent(webSocket, context, 'close_event', viewHistory)
```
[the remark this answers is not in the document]
bdibon: We'll handle this in a future PR.

packages/browser-core/src/browser/webSocketObservable.ts
```diff
@@ -0,0 +1,181 @@
...
+      ({ target: instance, parameters: [data], onPostCall }) => {
+        const size = computePayloadSize(data)
+        const bufferedAmountPreSend = instance.bufferedAmount
+
+        onPostCall(() => {
+          observable.notify({
+            state: 'message-out',
+            instance,
+            size,
+            bufferedAmountPreSend,
```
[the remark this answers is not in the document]
bdibon: We'll handle this in a future PR.

packages/browser-core/src/browser/webSocketObservable.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+function computePayloadSize(data: unknown): number {
+  if (typeof data === 'string') {
+    return computeBytesCount(data)
+  }
+  if (data instanceof ArrayBuffer || ArrayBuffer.isView(data)) {
+    return data.byteLength
+  }
+  if (typeof Blob !== 'undefined' && data instanceof Blob) {
```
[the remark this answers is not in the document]
bdibon: Too much of an edge case

PR #4726, opened 2026-06-04
✨ Add Browser RUM Salesforce support

BeltranBulbarellaDD:
## Motivation

Make the SDK not crash when instrumenting in Salesforce lightning environments.

## Changes

- Adds the support on the SDK so it not crashes on LWC environment. 

The main idea of this is to use `isEventSupported` to patch the 3 listeners LWC does not support. By using `isEventSupported` we can try to assign the listener and it fails safely. 
We do this check on 3 things: `change event of CookieStore`, `unhandledrejection event` and `securitypolicyviolation event listener`

The other 2 changes is to safely access `globalObject.location` instead of just `location` per Salesforce [docs](https://developer.salesforce.com/docs/platform/lightning-components-security/guide/lws-limitations.html#:~:text=The%20location%20property%20must%20be%20used%20with%20these%20accessors.). 

Also adds unit tests.

## Test instructions

Don't forget to make in the Dev Extension: `Event collection strategy` to `Requests`

Lightning application [URL](https://java-business-7601-dev-ed.scratch.lightning.force.com/lightning/page/home)
Username: 1780401668150_test-avrjsf381uzm@example.com
Password: Ufv7@chnbkeqmdssfjfn

Experience Cloud [URL](https://java-business-7601-dev-ed.scratch.my.site.com/ebikes/s/)

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/browser-core/src/browser/addEventListener.ts
```diff
@@ -146,3 +147,16 @@ export function addEventListeners<Target extends EventTarget, EventName extends
...
   }
 }
+
+export function isEventSupported<Target extends EventTarget, EventName extends keyof EventMapFor<Target> & string>(
+  eventTarget: Target,
+  eventName: EventName,
+  listener: (event: EventMapFor<Target>[EventName] & { type: EventName }) => void
+) {
+  try {
+    addEventListener({}, eventTarget, eventName, listener).stop()
```
BenoitZugmeyer: **suggestion:** 
```suggestion
    addEventListener({}, eventTarget, eventName, noop).stop()
```

packages/browser-rum-salesforce/src/entries/main.ts
BenoitZugmeyer, line 1: **suggestion:** let's not introduce the dedicated salesforce package until we have salesforce dedicated code. We can just tell customers to download the rum-slim bundle for now.

packages/browser-rum-slim/src/entries/main.ts
```diff
@@ -4,6 +4,8 @@ import type { RumPublicApi } from '@datadog/browser-rum-core'
 import { makeRumPublicApi } from '@datadog/browser-rum-core'
 import { makeRecorderApiStub } from '../boot/stubRecorderApi'
 import { makeProfilerApiStub } from '../boot/stubProfilerApi'
+export { makeRecorderApiStub } from '../boot/stubRecorderApi'
+export { makeProfilerApiStub } from '../boot/stubProfilerApi'
```
BenoitZugmeyer: **issue:** those should not be public

packages/browser-core/src/domain/error/trackRuntimeError.ts
```diff
@@ -44,6 +46,11 @@ export function instrumentOnError(callback: UnhandledErrorCallback) {
 }
 
 export function instrumentUnhandledRejection(callback: UnhandledErrorCallback) {
+  // Salesforce LWS does not support the unhandledrejection event. https://developer.salesforce.com/tools/lws-distortion-viewer
+  if (!isEventSupported(globalObject.window, DOM_EVENT.UNHANDLED_REJECTION, noop)) {
+    return { stop: noop }
+  }
```
BenoitZugmeyer, line 52: **thought:** I understand that we want to test for cookieStore "change" event explicitly, because else the session manager won't work correctly.

However, for other events, I wonder if it wouldn't be simpler to wrap the `addEventListener` in a try/catch and ignore event silently.

packages/browser-core/src/domain/error/trackRuntimeError.ts
```diff
@@ -44,6 +46,11 @@ export function instrumentOnError(callback: UnhandledErrorCallback) {
 }
 
 export function instrumentUnhandledRejection(callback: UnhandledErrorCallback) {
+  // Salesforce LWS does not support the unhandledrejection event. https://developer.salesforce.com/tools/lws-distortion-viewer
+  if (!isEventSupported(globalObject.window, DOM_EVENT.UNHANDLED_REJECTION, noop)) {
```
BenoitZugmeyer: **nitpick:** `globalObject.window` is a bit redundant 
```suggestion
  if (!isEventSupported(window, DOM_EVENT.UNHANDLED_REJECTION, noop)) {
```

packages/browser-core/src/domain/error/trackRuntimeError.ts
BeltranBulbarellaDD: I think this way just makes it clearer we are trying to attach the event in Salesforce. If we were to wrap the `addEventListener` we might catch/fail on other things without realizing it?

packages/browser-core/src/domain/error/trackRuntimeError.ts
BeltranBulbarellaDD: Actually it breaks compatibility. Do you think there is another way?
Screenshot 2026-06-04 at 18 20 23

packages/browser-rum-core/src/browser/locationChangeObservable.ts
```diff
@@ -17,7 +25,7 @@ export function createLocationChangeObservable(configuration: RumConfiguration)
       if (currentLocation.href === location.href) {
```
mormubis: Should we replace this `location.href` with `getLocation().href` too?

packages/browser-core/src/domain/error/trackRuntimeError.ts
BenoitZugmeyer: OK!

packages/browser-rum-slim/src/salesforce/README.md
BenoitZugmeyer, line 1: **suggestion:** is it the right place to put this doc? Could we put it in https://github.com/DataDog/integrations-extras instead?

packages/browser-core/src/domain/error/trackRuntimeError.ts
BenoitZugmeyer: If `window` is undefined, then `globalObject.window` is also undefined, and so it seems that `isEventSupported` works but the types are unsound.

We should make `globalObject.window` optional, so types are accurate

packages/browser-rum-slim/src/salesforce/README.md
BeltranBulbarellaDD: Yeah I agree that is a better place but for now, until we release completely and have support, I would prefer to keep it in the repository.

packages/browser-core/src/tools/globalObject.ts
```diff
@@ -23,6 +26,9 @@ export interface GlobalObject extends Omit<typeof globalThis, 'queueMicrotask' |
 
   // Profiler is not available in all browsers yet
   Profiler?: ProfilerConstructor
+
+  // window is not available in SSR environments
+  window?: Window
```
BenoitZugmeyer, line 31: **praise:** Nice!

BenoitZugmeyer (approved):

mormubis (approved):

PR #4727, opened 2026-06-04
✨ Introduce @datadog/js-core, a runtime-agnostic core package with independent versioning

thomas-lebeau:
## Motivation

External packages such as [`electron-sdk`](https://github.com/DataDog/electron-sdk) and [`openfeature-js-client`](https://github.com/DataDog/openfeature-js-client) currently import APIs from `@datadog/browser-core`, which was never designed as a stable public surface — breaking changes can ship in patch releases, making downstream integrations brittle.

This PR introduces **`@datadog/js-core`**, a small, semver-stable, runtime-agnostic core package intended for cross-SDK and external usage, as proposed in the RFC. It is the MVP first step: create the package, move a first utility into it, and wire up the release process.

## Changes

**New package `@datadog/js-core`**
- Sub-path exports via both the `exports` field (modern resolvers, native Node ESM/CJS) and physical `<name>/` directory fallbacks (legacy resolvers — webpack 4, old Node). The build emits `esm/package.json` with `{"type":"module"}` so Node correctly treats `esm/*.js` as ES modules.
- No root entry point — always import from a sub-path (e.g. `@datadog/js-core/time`).
- Versioned independently from `@datadog/browser-*` (starts at `0.0.1`).
- `README.md` + `AGENTS.md` documenting the no-breaking-changes, JSDoc, and sub-path rules.
- Moves `dateNow` out of `@datadog/browser-core` into `@datadog/js-core/time` and updates all call sites in `browser-core` and `browser-debugger` to import directly from the new package.
- Adds `@datadog/js-core/time` to the `disallowSideEffects` ESLint allowlist and a `tsconfig` path alias.
- Adds `--esm-type-module` flag to the shared `build-package.ts` script, used by js-core to emit `esm/package.json` post-build.

**Release process**
- `@datadog/js-core` is published in the **same release pipeline** as the browser packages, but keeps its **own version**.
- `prepare-release` bumps independently-versioned packages with a **patch increment** (e.g. `0.0.1` → `0.0.2`) instead of the synced release version, and updates dependents (`browser-core`, `browser-debugger`) to reference the new version.
- `check-release` validates that dependents reference the independent package's actual version.

## Test instructions

- `yarn build` — js-core builds, integrates into the full build, and emits `esm/package.json`.
- `yarn typecheck`, `yarn lint` — pass.
- `yarn test:script` — 47 pass / 0 fail (validates the release/deploy scripts).
- Release behavior was verified by simulating `prepare-release` + `check-release` against all package.json files: browser packages bump to the release version, js-core patches `0.0.1` → `0.0.2`, and `browser-core`'s `@datadog/js-core` dependency is updated and validated to match.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


thomas-lebeau:
@codex review

packages/js-core/AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+- `@param` for each parameter (if the name alone is not self-explanatory)
+- `@returns` describing the return value and its unit/format
+
+### Runtime-agnostic exports only
+
+Every export must work in both Node.js and browser environments. Do not add:
+
+- DOM APIs (`document`, `window`, `navigator`, `performance.timing`, etc.)
+- Node.js built-in modules (`fs`, `path`, `process`, etc.)
+- Any API that is not available in both runtimes
```
BenoitZugmeyer, line 39: **nitpick:** we'll probably have some runtime-related code, like some kind of `pageMayExitObservable` versions for Browser and Nodejs. Or transport level things.

We'll need to figure whether we want a [`browser` conditional export](https://nodejs.org/api/packages.html#community-conditions-definitions) so we have isomorphic packages, or we export both variants and the consumer uses one or the other.

packages/js-core/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The package uses sub-path exports. Import only what you need.
+
+### `@datadog/js-core/time`
+
+Time utilities that work in any JavaScript runtime.
+
+| Export      | Description                                                                                                     |
+| ----------- | --------------------------------------------------------------------------------------------------------------- |
+| `dateNow()` | Returns the current Unix timestamp in milliseconds. Prefer over `Date.now()` to guard against broken polyfills. |
```
BenoitZugmeyer, line 30: **thought:** for the future: we should have a API reference doc that is automatically generated. I don't want to maintain the README :D

scripts/lib/filesUtils.ts
```diff
@@ -65,6 +65,29 @@ interface PackageJson {
   peerDependencies?: Record<string, string>
 }
 
+/**
+ * Packages under `packages/` that are versioned independently from the synced Browser SDK release
+ * version. They are still built and published as part of the same release process, but their
+ * version is managed manually and is not bumped to the release version.
+ */
+export const INDEPENDENTLY_VERSIONED_PACKAGES = new Set<string>(['@datadog/js-core'])
```
BenoitZugmeyer: **nitpick:** this constant don't need to be exported

BenoitZugmeyer (approved):

packages/js-core/README.md
thomas-lebeau: I agree, for now I want to keep it simple. Claude can update the readme.

packages/js-core/AGENTS.md
thomas-lebeau: true, we probably want to have some integration testing too. but I'd rather think about that in a separate PR.

BenoitZugmeyer (approved):

PR #4747, opened 2026-06-08
✨ [RUM-16632] add profiling bridge mode support

bcaudan:
## Motivation

Profiling data is always sent directly to the intake via HTTP. In Electron and WebView environments where the native SDK acts as a proxy through the bridge, profiling requests would use dummy session and application IDs, resulting in rejections at the intake.

## Changes

- Add `BridgeCapability.PROFILES`: route profiling through the bridge when the native SDK advertises it, disable profiling silently when a bridge is present without that capability
- Refactor the profiler to be transport-agnostic: introduce an `emitPayload` callback so the profiler core no longer owns transport decisions
- Add `ProfilingPayload` as a shared domain type (`profile` + `trace`) passed to both the bridge emitter and the FormData emitter
- Session injection for the non-bridge path moves to `formDataEmitter`, keeping the bridge path session-free (the native SDK handles it)

## Test instructions

Unit tests cover bridge mode routing, session injection, and sampling. E2E test covers the bridge path end-to-end using the mock bridge.

## Checklist

- [x] Tested locally
- [x] Tested on staging


bcaudan:
@codex review

packages/browser-rum/src/boot/profilerApi.ts
```diff
@@ -49,7 +65,11 @@ export function makeProfilerApi(): ProfilerApi {
       return
     }
 
-    lazyLoadProfiler()
+    const emitPayload = isBridgeMode
+      ? createBridgeEmitter()
+      : createFormDataEmitter(configuration, lifeCycle, createEncoder)
```
BenoitZugmeyer: **suggestion:** do this in the lazy chunk (maybe in `createRumProfiler`), so it doesn't impact the main bundle size

packages/browser-rum/src/domain/profiling/types.ts
```diff
@@ -54,6 +54,11 @@ export interface RUMProfiler {
   isPaused: () => boolean
 }
 
+export interface ProfilingPayload {
+  profile: BrowserProfileEvent
```
BenoitZugmeyer, line 60: **suggestion:** call this property `event` to align the naming between transports?

bcaudan: The bridge "event" is actually the full payload:
```ts
{
  "eventType": "profile",
  "event": {
    "profile": { ...BrowserProfileEvent },
    "trace":   { ...BrowserProfilerTrace }
  }
}
```

packages/browser-rum/src/domain/profiling/datadogProfiler.ts
```diff
@@ -268,8 +266,8 @@ export function createRumProfiler(
         }
 
         handleProfilerTrace(
-          // Enrich trace with time and instance data
-          Object.assign(trace, {
+          // Enrich trace with time and instance data — create new object to avoid mutating the raw trace
+          Object.assign({} as BrowserProfilerTrace, trace, {
```
thomasbertet, line 296: I don't think this is necessary 🤔 . I don't think it hurts too much either though.

thomasbertet (approved):
Code LGTM.

packages/browser-rum/src/boot/profilerApi.ts
bcaudan: Updated:
Screenshot 2026-06-11 at 11 22 51

packages/browser-rum/src/domain/profiling/datadogProfiler.ts
```diff
@@ -349,9 +354,11 @@ export function createRumProfiler(
 
   function handleProfilerTrace(trace: BrowserProfilerTrace, startTime: RelativeTime): void {
     const sessionId = session.findTrackedSession(startTime)?.id
-    const payload = assembleProfilingPayload(trace, configuration, sessionId)
-
-    void transport.send(payload as unknown as TransportPayload)
+    const payload: ProfilingPayload = {
+      profile: buildProfileEvent(trace, configuration, sessionId),
```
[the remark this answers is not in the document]
bcaudan: That is intended, native sdk is responsible for providing the session id

packages/browser-rum/src/domain/profiling/datadogProfiler.ts
```diff
@@ -47,7 +50,9 @@ export function createRumProfiler(
   viewHistory: ViewHistory,
   profilerConfiguration: RUMProfilerConfiguration = DEFAULT_RUM_PROFILER_CONFIGURATION
 ): RUMProfiler {
-  const transport = createFormDataTransport(configuration, lifeCycle, createEncoder, DeflateEncoderStreamId.PROFILING)
+  const emitPayload = canUseEventBridge()
+    ? mockable(createBridgeEmitter)()
+    : mockable(createFormDataEmitter)(configuration, lifeCycle, createEncoder)
```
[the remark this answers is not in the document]
bcaudan: oh, good catch, @thomasbertet we should chat about that

amortemousque (approved):

packages/browser-rum/src/domain/profiling/types.ts
BenoitZugmeyer: So, you do this to avoid the `event: { event: ... }?

I would still rename `profile` to `event` in the payload to avoid back-and-forth between namings:

https://github.com/DataDog/browser-sdk/blob/1afb4bc37b700681a16a124e4aac319a59106363/test/e2e/lib/framework/intakeProxyMiddleware.ts#L246
https://github.com/DataDog/browser-sdk/blob/1afb4bc37b700681a16a124e4aac319a59106363/packages/browser-rum/src/domain/profiling/transport/formDataEmitter.ts#L14

But that's a nitpick, you can ignore :)

BenoitZugmeyer (approved):

packages/browser-rum/src/domain/profiling/types.ts
bcaudan: > So, you do this to avoid the `event: { event: ... }?

Yeah, it felt a bit confusing

packages/browser-rum/src/boot/profilerApi.ts
```diff
@@ -77,3 +77,14 @@ export function makeProfilerApi(): ProfilerApi {
     },
   }
 }
+
+function isProfilingSampled(configuration: RumConfiguration, session: SessionContext) {
+  if (canUseEventBridge()) {
+    // In bridge mode, native SDK owns the sampling decision, skip the rate check
+    return bridgeSupports(BridgeCapability.PROFILES)
```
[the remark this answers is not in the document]
bcaudan: `findTrackedSession` returns the dummy session in bridge mode

bcaudan:
> Apply bridge sampling rules on session renewal

@chatgpt-codex-connector that is indeed a known issue with the bridge, session expiration / renewal is not propagated from native SDK to browser SDK.
Definitely something to improve but outside the scope of this PR.

thomasbertet (approved):
LGTM

thomasbertet:
As discussed, one nice follow-up would be to not load the profiler chunk (by not sending the "PROFILES" capability) when we know the profiling sample rate is 0.

amortemousque (approved):

PR #4756, opened 2026-06-10
⚗️ Partial view updates — batch optimizations

mormubis:
## Motivation

The initial partial view updates implementation (already on main) sends diffs via `batch.add()`. Two inefficiencies: when the full VIEW is still in the current batch, intermediate updates add separate `view_update` events on top of it instead of replacing the VIEW in place. And when multiple updates land in the same batch, each one becomes a separate `view_update` instead of a single aggregate.

## Changes

`startRumBatch.ts` — refactored into `createViewBatchRouter` to make the routing logic unit-testable without the full batch setup. Two batch-level optimizations:

If the current batch already has a full VIEW, intermediate updates replace it in the upsert buffer (`batchHasFullView` path). No `view_update` emitted. Same as the non-experimental behavior.

If the batch was already flushed, updates compute an aggregate diff from `batchBase` (the last state the backend received) and upsert it under the same key. Multiple updates collapse to one `view_update` per batch.

## Test instructions

1. `yarn dev`, open `http://localhost:8080`
2. Add `enableExperimentalFeatures: ['partial_view_updates']` to the `DD_RUM.init()` call in `sandbox/index.html`
3. Open DevTools → Network tab, filter by `/proxy`

**Optimization 1 (no view_update emitted):**

4. In the console, call `DD_RUM.setViewName('test')` — update arrives while the VIEW is still in the batch
5. Flush: `window.dispatchEvent(new Event('beforeunload'))`
6. The proxy request should contain only a full VIEW, no `view_update`

**Optimization 2 (aggregate view_update sent):**

7. Flush first: `window.dispatchEvent(new Event('beforeunload'))`
8. In the console, call `DD_RUM.setViewName('test2')` — update arrives in a fresh batch
9. Flush again: `window.dispatchEvent(new Event('beforeunload'))`
10. The second proxy request should be noticeably smaller — it contains a `view_update` diff, not a full VIEW

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [ ] Updated documentation

### DIFF
--- a/packages/browser-rum-core/src/transport/startRumBatch.ts
+++ b/packages/browser-rum-core/src/transport/startRumBatch.ts
@@ -22,10 +22,7 @@ import { diffMerge } from '../domain/view/viewDiff'
 export const PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL = 100
 
 export function computeAssembledViewDiff(current: RumViewEvent, last: RumViewEvent): RumViewUpdateEvent | undefined {
-  const currentObj = current as unknown as Record<string, unknown>
-  const lastObj = last as unknown as Record<string, unknown>
-
-  const diff = diffMerge(currentObj, lastObj, {
+  const diff = diffMerge(current, last, {
     // context, connectivity, usr, device, privacy are objects — use REPLACE to avoid partial updates
     replaceKeys: new Set(['view.custom_timings', 'context', 'connectivity', 'usr', 'device', 'privacy']),
     appendKeys: new Set(['_dd.page_states']),
@@ -46,24 +43,21 @@ export function computeAssembledViewDiff(current: RumViewEvent, last: RumViewEve
     return undefined
   }
 
-  const currentView = currentObj.view as Record<string, unknown>
-  const currentDd = currentObj._dd as Record<string, unknown>
-
   // Merge always-required fields on top of the diff for backend routing
   return combine(diff, {
     type: RumEventType.VIEW_UPDATE,
-    date: currentObj.date,
-    application: currentObj.application,
-    session: currentObj.session,
+    date: current.date,
+    application: current.application,
+    session: current.session,
     view: {
-      id: currentView.id,
-      url: currentView.url,
+      id: current.view.id,
+      url: current.view.url,
     },
     _dd: {
-      document_version: currentDd.document_version,
-      format_version: currentDd.format_version,
+      document_version: current._dd.document_version,
+      format_version: current._dd.format_version,
     },
-  }) as unknown as RumViewUpdateEvent
+  })
 }
 
 /**
--- a/packages/browser-rum-core/src/transport/startRumBatch.ts
+++ b/packages/browser-rum-core/src/transport/startRumBatch.ts
@@ -15,17 +15,22 @@ import type { RumConfiguration } from '../domain/configuration'
 import type { LifeCycle } from '../domain/lifeCycle'
 import { LifeCycleEventType } from '../domain/lifeCycle'
 import type { AssembledRumEvent } from '../rawRumEvent.types'
-import type { RumViewEvent, RumViewUpdateEvent } from '../rumEvent.types'
+import type { RumViewUpdateEvent } from '../rumEvent.types'
 import { RumEventType } from '../rawRumEvent.types'
 import { diffMerge } from '../domain/view/viewDiff'
 
+type AssembledViewEvent = Extract<AssembledRumEvent, { type: 'view' }>
+
 export const PARTIAL_VIEW_UPDATE_CHECKPOINT_INTERVAL = 100
 
-export function assembleViewUpdateEvent(
-  current: RumViewEvent,
-  last: RumViewEvent
-): (RumViewUpdateEvent & Context) | undefined {
-  const diff = diffMerge(current, last, {
+export function computeAssembledViewDiff(
+  current: AssembledViewEvent,
+  last: AssembledViewEvent
+): RumViewUpdateEvent | undefined {
+  const currentObj = current as unknown as Record<string, unknown>
+  const lastObj = last as unknown as Record<string, unknown>
+
+  const diff = diffMerge(currentObj, lastObj, {
     // context, connectivity, usr, device, privacy are objects — use REPLACE to avoid partial updates
     replaceKeys: new Set(['view.custom_timings', 'context', 'connectivity', 'usr', 'device', 'privacy']),
     appendKeys: new Set(['_dd.page_states']),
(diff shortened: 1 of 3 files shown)


packages/browser-rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,182 @@
...
+  let batchBase: AssembledRumEvent | undefined
+  // True when the current batch already contains a full VIEW event for the active view.
+  // Reset to false on each flush.
+  let batchHasFullView = false
+  let viewUpdatesSinceCheckpoint = 0
+
+  // On flush: advance batchBase to lastSentView (what the backend now has) and clear the flag.
+  batch.flushController.flushObservable.subscribe(() => {
+    batchHasFullView = false
+    batchBase = lastSentView
```
mormubis, line 98: batchBase tracks what the backend actually received, not just the latest event we processed. It advances to lastSentView on flush, so aggregate diffs in the next batch are always computed against the right base. If we used lastSentView directly as the diff base without this reset, we'd start each new batch from the wrong state.

packages/browser-rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,182 @@
...
+        batch.upsert(serverRumEvent, viewId)
+        return
+      }
+    }
+
+    if (batchBase) {
+      const diff = computeAssembledViewDiff(serverRumEvent, batchBase)
+      if (diff) {
+        sendToExtension('rum', diff)
+        batch.upsert(diff, viewId)
```
mormubis: The view_update upsert uses the same key as the full VIEW (viewId). This means when a checkpoint fires, upsert(fullView, viewId) naturally replaces any pending view_update in the buffer without needing to explicitly remove it.

packages/browser-rum-core/src/transport/startRumBatch.ts
```diff
@@ -1,10 +1,182 @@
...
+    }
+
+    // Optimization 2: no full VIEW in the current batch — compute an aggregate diff from
+    // batchBase (last state the backend received) and upsert it under the same key as the VIEW.
+    // Each update replaces the previous aggregate, so the batch always contains at most one
+    // view_update per view ID.
+    //
+    // Note: view_update events are created here, post-assembly, bypassing
+    // RAW_RUM_EVENT_COLLECTED → assembly → RUM_EVENT_COLLECTED. They intentionally skip
+    // beforeSend — view_update is an internal bandwidth optimization, not a customer-visible type.
```
mormubis: view_update bypasses beforeSend by design, not by accident. Since these events are assembled here (not through the raw event pipeline), there's no natural hook. And since view_update is an internal bandwidth optimization rather than a customer-visible type, beforeSend modification wouldn't be meaningful.

packages/browser-rum-core/src/transport/startRumBatch.ts
```diff
@@ -46,21 +46,144 @@ export function assembleViewUpdateEvent(
...
     },
     _dd: {
-      document_version: current._dd.document_version,
-      format_version: current._dd.format_version,
+      document_version: currentDd.document_version,
+      format_version: currentDd.format_version,
     },
-  }) as RumViewUpdateEvent & Context
+  }) as unknown as RumViewUpdateEvent
+}
```
BenoitZugmeyer, line 61: **suggestion:** no cast needed

```ts
export function computeAssembledViewDiff(current: RumViewEvent, last: RumViewEvent): RumViewUpdateEvent | undefined {
  const diff = diffMerge(current, last, {
    // context, connectivity, usr, device, privacy are objects — use REPLACE to avoid partial updates
    replaceKeys: new Set(['view.custom_timings', 'context', 'connectivity', 'usr', 'device', 'privacy']),
    appendKeys: new Set(['_dd.page_states']),
    // Ignore always-required fields — they are added back via combine regardless of changes
    ignoreKeys: new Set([
      'date',
      'type',
      'application',
      'session',
      'view.id',
      'view.url',
      '_dd.document_version',
      '_dd.format_version',
    ]),
  })

  if (!diff) {
    return
  }

  // Merge always-required fields on top of the diff for backend routing
  return combine(diff, {
    type: RumEventType.VIEW_UPDATE,
    date: current.date,
    application: current.application,
    session: current.session,
    view: {
      id: current.view.id,
      url: current.view.url,
    },
    _dd: {
      document_version: current._dd.document_version,
      format_version: current._dd.format_version,
    },
  })
}

```

packages/browser-rum-core/src/transport/startRumBatch.ts
```diff
@@ -46,21 +46,144 @@ export function assembleViewUpdateEvent(
...
+
+      const viewId = serverRumEvent.view.id
+
+      // New view started
+      if (viewId !== lastSentView?.view.id) {
+        viewUpdatesSinceCheckpoint = 0
+        batch.upsert(serverRumEvent, viewId)
+        // State set after upsert: if upsert triggers a sync flush, the flush subscriber runs
+        // first (resetting batchHasFullView to false), then these assignments restore the
+        // correct state for the fresh batch that now contains this VIEW.
```
BenoitZugmeyer, line 123: **issue:** this makes everything more brittle than necessary, and in fact it is problematic, because `batch.upsert` can flush the batch with or without the view.

What would make the implementation easier is if you could have a hint of what is being flushed in the `flushObservable`, something like this:

```ts
const { unsubscribe } = batch.flushObservable.subscribe(({ upsertedKeys }) => {
  if (upsertedKeys.includes(lastSentView.view.id)) {
    batchHasFullView = false
  }
})

```

test/e2e/scenario/rum/partialViewUpdates.scenario.ts
```diff
@@ -1,15 +1,39 @@
...
+  view: { id: string; is_active?: boolean; [key: string]: unknown }
+  _dd: { document_version: number; [key: string]: unknown }
+  [key: string]: unknown
+}
+
+// Helper: extract view_update events from all RUM events
+// (intakeRegistry.rumViewEvents only returns type==='view')
+function getViewUpdateEvents(intakeRegistry: IntakeRegistry): ViewUpdateEvent[] {
+  return intakeRegistry.rumEvents.filter((e) => (e.type as string) === 'view_update') as unknown as ViewUpdateEvent[]
+}
```
BenoitZugmeyer, line 20: **suggestion:** https://github.com/DataDog/browser-sdk/pull/4201#discussion_r3275578701

BenoitZugmeyer (changes requested):
Let's talk about it

PR #4767, opened 2026-06-11
👷 migrate module builds from TypeScript compiler API to tsdown

thomas-lebeau:
## Motivation

The TypeScript compiler API (`ts.createProgram`) used for building package modules is slow and verbose. This was called out in a TODO comment in `build-package.ts`:

> TODO: in the future, consider building packages with something else than typescript (ex: rspack, tsdown...)

This migrates module transpilation to **tsdown** (backed by Rolldown/Rust).

## Changes

- **Transpile** package modules (CJS + ESM) with `tsdown` instead of `ts.createProgram`/`program.emit`.
  - `unbundle: true` preserves per-file output (matching tsc).
  - Per-format `outDir` keeps the `./cjs` and `./esm` layout. ESM now uses `.mjs`, so the `--esm-type-module` flag (which wrote `esm/package.json`) is removed.
  - `define` inlines build-time constants (`__BUILD_ENV__*__`) at transpile time, replacing the old `replaceBuildEnvInDirectory` post-processing.
- **Declarations are still emitted by `tsc`** (`--emitDeclarationOnly`, `dts: false` in tsdown). Rolldown's declaration *bundler* restructures modules in ways that break older TypeScript consumers — inline `type` modifiers (TS 4.5+), `.js`/`.mjs` extensions on declaration imports (unresolvable under `moduleResolution: node` < TS 4.7), and `export ... from` re-exports rewritten into `import` + local `export` (trips `isolatedModules` on re-exported `const enum`s). Keeping tsc for `.d.ts` makes declaration output identical to before, preserving the full TS 3.8+ compatibility matrix. Declarations are emitted only next to the CJS output, since every package's `types` field (and the `types` condition in `exports`) already points at `./cjs`.

## Test instructions

- `yarn build` → `yarn typecheck` — passes.
- `scripts/cli typecheck test/apps/vanilla` — passes.
- `yarn test:compat:tsc` — all checks pass (TS 3.8.2, isolatedModules, 4.1.6, latest, exactOptionalPropertyTypes, ESNext).
- `yarn build:apps` — all test apps build.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


package.json
```diff
@@ -12,13 +12,13 @@
   "type": "module",
   "scripts": {
     "postinstall": "scripts/cli init_submodule",
-    "build": "yarn workspaces foreach --all --parallel --topological run build",
-    "build:bundle": "yarn workspaces foreach --all --parallel run build:bundle",
+    "build": "yarn workspaces foreach --all --parallel --topological-dev run build",
+    "build:bundle": "yarn workspaces foreach --all --parallel --topological-dev run build:bundle",
```
BenoitZugmeyer: **question:** why do we need those `-dev`? Also `build:bundle` don't need `--topological` because all bundles are built including all source files, not built artifacts. Paths are resolved from `tsconfig.base.json`.

scripts/build/build-package.ts
```diff
@@ -90,74 +73,82 @@ async function buildBundle({ filename, verbose }: { filename: string; verbose: b
...
+      },
     },
-    include: ['./src'],
-    exclude: ['./src/**/*.spec.*', './src/**/*.specHelper.*'],
+    platform: 'neutral',
+    unbundle: true,
+    dts: false,
+    tsconfig: '../../tsconfig.base.json',
+    // Mark all non-relative imports as external (cross-package deps and node_modules)
+    deps: { neverBundle: /^[^./]/ },
```
BenoitZugmeyer: **thought:** this `deps` config looks fishy, we can probably remove it, the default should be good.

scripts/build/build-package.ts
```diff
@@ -90,74 +73,82 @@ async function buildBundle({ filename, verbose }: { filename: string; verbose: b
...
-      outDir,
-      paths: {},
+async function buildModules({ verbose }: { verbose: boolean }) {
+  // Transpile the source with tsdown (Rolldown). We let TypeScript emit the declaration files (see
+  // emitDeclarations) rather than tsdown, because Rolldown's declaration bundler restructures
+  // modules in ways that break compatibility with older TypeScript versions (e.g. inline `type`
+  // modifiers, rewritten re-exports).
+  await tsdownBuild({
+    clean: true,
+    entry: ['./src/**/*.ts', '!./src/**/*.spec.ts', '!./src/**/*.specHelper.ts'],
```
BenoitZugmeyer: **thought:** considering all files are entrypoints is a bit suboptimal, if we included only real entry points tsdown would be able to do some treeshaking for code that is only used in specs.

scripts/build/build-package.ts
```diff
@@ -90,74 +73,82 @@ async function buildBundle({ filename, verbose }: { filename: string; verbose: b
...
-    exclude: ['./src/**/*.spec.*', './src/**/*.specHelper.*'],
+    platform: 'neutral',
+    unbundle: true,
+    dts: false,
+    tsconfig: '../../tsconfig.base.json',
+    // Mark all non-relative imports as external (cross-package deps and node_modules)
+    deps: { neverBundle: /^[^./]/ },
+    define: Object.fromEntries(
+      referencedBuildEnvKeys().map((key) => [`__BUILD_ENV__${key}__`, JSON.stringify(getBuildEnvValue(key))])
+    ),
```
BenoitZugmeyer, line 121: **suggestion:** We can probably define all build env even if they are unused? 

```ts
    define: Object.fromEntries(buildEnvKeys.map((key) => [`__BUILD_ENV__${key}__`, getBuildEnvValue(key)])),
```

thomas-lebeau: The filtering is there specifically because `WORKER_STRING` isn't a pure value like the others — evaluating it has a side effect. `getBuildEnvValue('WORKER_STRING')` reads packages/browser-worker/bundle/worker.js (and rebuilds the worker if it's missing/stale). Only browser-rum actually references `__BUILD_ENV__WORKER_STRING__` (in deflateWorker.ts).

Since tsdown's define is eager — every listed key is computed up front, for every package — defining all keys unconditionally makes every package's build read the worker bundle, even though only browser-rum needs it. In the parallel yarn build that's a problem: browser-debugger/logs/etc. end up reading worker.js while browser-worker is concurrently rebuilding it (its build does fs.rm('./bundle') first), which gives a flaky ENOENT.

package.json
thomas-lebeau: because I added browser-worker as a dev dependency, in Theory this is needed because it need to builds the worker before building browser-rum. 

In practice, `needsWorkerRebuild()` in buildEnv deals about it. Ideally I would get rid of it, but it is still needed because unit tests don't build first.

scripts/build/build-package.ts
BenoitZugmeyer: I see! Thank you for the explanation. Makes sense then, however this worker string brings a lot of complexity, if we were able to build it fast and in-memory we wouldn't need to cache it on disk, so no dependency needed and no concurrent access to the files. Let's keep it in mind :)

scripts/build/build-package.ts
```diff
@@ -90,74 +73,92 @@ async function buildBundle({ filename, verbose }: { filename: string; verbose: b
...
-  if (diagnostics.length) {
-    printTypeScriptDiagnostics(diagnostics)
-    throw new Error('Failed to build package due to TypeScript errors')
+function getEntries() {
+  if (existsSync('./src/index.ts') || existsSync('./src/entries')) {
+    return ['./src/index.ts', './src/entries/*.ts']
   }
 
-  await replaceBuildEnvInDirectory(outDir, { verbose })
+  return ['./src/*.ts', './src/*/index.ts']
```
BenoitZugmeyer: **suggestion:** WDYT about always using `entries/*.ts` in js-core, so we can simplify this?

packages/browser-logs/src/domain/logger.ts
```diff
@@ -55,46 +55,47 @@ export class Logger {
...
-      const rawError = computeRawError({
-        originalError: error,
-        nonErrorPrefix: NonErrorPrefix.PROVIDED,
-        source: ErrorSource.LOGGER,
-        handling: ErrorHandling.HANDLED,
-        startClocks: clocksNow(),
-      })
-
-      context = combine(
+    callMonitored(() => {
```
BenoitZugmeyer, line 65: **nitpick:** we should be able to remove the `monitored` decorator no?

BenoitZugmeyer (approved):

packages/browser-logs/src/domain/logger.ts
thomas-lebeau: we could yes. I think it's good to have a decorator for any modern consumer that might want to use it.
To use it in the browser-sdk while building with tsdown we would need to add a dependency for the decorator pollyfill. tsc inline it OOTB, but not tsdown and I though it's not worth it complicating the build to inline it for just this one usage.

PR #4807, opened 2026-06-18
⚗️ Partial view updates — batch optimizations

mormubis:
## Motivation

The initial partial view updates implementation (already on main) sends diffs via `batch.add()`. Two inefficiencies: when the full VIEW is still in the current batch, intermediate updates add separate `view_update` events on top of it instead of replacing the VIEW in place. And when multiple updates land in the same batch, each one becomes a separate `view_update` instead of a single aggregate.

## Changes

`startRumBatch.ts` — routing logic extracted into `createBatchDispatcher` to make it unit-testable without the full batch setup. Two batch-level optimizations:

If the current batch already has a full VIEW, intermediate updates replace it in the upsert buffer (`batchHasFullView` path). No `view_update` emitted. Same as the non-experimental behavior.

If the batch was already flushed, updates compute an aggregate diff from `batchBase` (the last state the backend received) and upsert it under the same key. Multiple updates collapse to one `view_update` per batch.

`batchHasFullView` is set after `batch.upsert()` using `flushController.messagesCount > 0` — if a sync flush fired inside the upsert (e.g. messages limit), the VIEW was already sent and the batch is empty, so `batchHasFullView` stays false. This avoids the state race where a sync flush could incorrectly mark an empty batch as containing a full VIEW.

## Test instructions

1. `yarn dev`, open `http://localhost:8080`
2. Add `enableExperimentalFeatures: ['partial_view_updates']` to the `DD_RUM.init()` call in `sandbox/index.html`
3. Open DevTools → Network tab, filter by `/proxy`

**Optimization 1 (no view_update emitted):**

4. In the console, call `DD_RUM.setViewName('test')` — update arrives while the VIEW is still in the batch
5. Flush: `window.dispatchEvent(new Event('beforeunload'))`
6. The proxy request should contain only a full VIEW, no `view_update`

**Optimization 2 (aggregate view_update sent):**

7. Flush first: `window.dispatchEvent(new Event('beforeunload'))`
8. In the console, call `DD_RUM.setViewName('test2')` — update arrives in a fresh batch
9. Flush again: `window.dispatchEvent(new Event('beforeunload'))`
10. The second proxy request should be noticeably smaller — it contains a `view_update` diff, not a full VIEW

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [ ] Updated documentation


packages/browser-core/src/transport/batch.ts
```diff
@@ -115,6 +116,9 @@ export function createBatch({
   }
 
   return {
+    get isEmpty() {
+      return encoder.isEmpty && Object.keys(upsertBuffer).length === 0
```
BenoitZugmeyer: **nitpick:** this works, but `flushController.messagesCount > 0` should also do the trick?

BenoitZugmeyer (approved):

PR #4812, opened 2026-06-19
✨ [RUM-16925] attach debug IDs from DD_SOURCE_CODE_CONTEXT to error and long task events

amortemousque:
## Motivation

Part of the sourcemap unminification project. When a Datadog bundler plugin is used, it injects `DD_SOURCE_CODE_CONTEXT` on the window with per-bundle metadata (service, version, debug ID). This PR makes the RUM SDK read the `ddDebugId` field from that context and attach it to error and long task events as `_dd.debug_ids`, enabling the backend to correlate events with uploaded sourcemaps.

## Changes

- Move `sourceCodeContext` from `browser-rum-core` to `browser-core` so both  RUM and later Logs SDK can access it.
- Rename the MFE-specific context hook to `sourceCodeMfeContext` (clearer scope)
- In `computeRawError`: resolve `debug_ids` by looking up all stack frame URLs in `DD_SOURCE_CODE_CONTEXT` and attach them as `_dd.debug_ids` on error events
- In `longTaskCollection`: same lookup on `script.sourceURL` for Long Animation Frame events

## Test instructions

1. Set `window.DD_SOURCE_CODE_CONTEXT` with a stack entry whose top-frame URL matches a script URL on the page and a `ddDebugId` value
2. Trigger an error or long task from that script
3. Verify the intake event has `_dd.debug_ids: { [scriptUrl]: debugId }`

The E2E tests cover both flows automatically (`yarn test:e2e -g "debug ID"`).

## Checklist

- [x] Tested locally
- [ ] Tested on staging


packages/browser-rum-core/src/domain/error/errorCollection.ts
```diff
@@ -83,6 +83,7 @@ function processError(error: RawError): RawRumEventCollectedData<RawRumErrorEven
       fingerprint: error.fingerprint,
       csp: error.csp,
     },
+    _dd: { debug_ids: error.debugIds },
```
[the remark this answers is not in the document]
amortemousque: Right I need to fix the rum event format

packages/browser-rum-core/src/domain/contexts/sourceCodeContext.ts
```diff
@@ -1,85 +0,0 @@
...
-
-export function startSourceCodeContext(assembleHook: AssembleHook) {
-  const browserWindow = window as BrowserWindow
-  const contextByFile = new Map<StackFrameUrl, SourceCodeContext>()
-
-  function buildContextByFile() {
-    if (!browserWindow.DD_SOURCE_CODE_CONTEXT) {
-      return
-    }
-    addTelemetryUsage({ feature: 'source-code-context' })
```
rgaignault, line 23: 💭 Suggestion : We could carry the telemetry tracking over the new file

packages/browser-core/src/domain/sourceCodeContext.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    expect(getDebugIds([url])).toBeUndefined()
+  })
+
+  it('should omit URLs whose entry has no ddDebugId', () => {
+    mockSourceCodeContext({ [makeStack(url)]: { service: 'my-service' } })
+
+    expect(getDebugIds([url])).toBeUndefined()
+  })
+
+  it('should only matche the top frame URL, not deeper frames of the context stack', () => {
```
rgaignault: nitpick : match

packages/browser-core/src/domain/error/error.ts
```diff
@@ -48,6 +49,10 @@ function computeErrorBase({
   }
 }
 
+function getStackTraceUrls(stackTrace: StackTrace) {
+  return stackTrace.stack.map((frame) => frame.url).filter((url): url is string => !!url)
+}
+
 export function computeRawError({
```
rgaignault, line 77: 💬 Suggestion : since logs is not using debug ids, maybe we could add a flag ?

rgaignault (dismissed):

packages/browser-core/src/domain/error/error.ts
amortemousque: What do you mean by "add a flag"?

rgaignault: I meant adding a param that only RUM passes, so the lookup is skipped for Logs

packages/browser-core/src/domain/sourceCodeContext.ts
```diff
@@ -0,0 +1,60 @@
...
+const contextByUrl = new Map<string, SourceCodeContextEntry>()
+const processedStacks = new Set<string>()
+
+function syncSourceCodeContext() {
+  const sourceCodeContext = (window as BrowserWindow).DD_SOURCE_CODE_CONTEXT
+  if (!sourceCodeContext) {
+    return
+  }
+
+  for (const [stack, entry] of Object.entries(sourceCodeContext)) {
```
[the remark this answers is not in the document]
amortemousque: Yes, but in the future the Logs SDK will also read DD_SOURCE_CODE_CONTEXT, so if RUM clears it, Logs won’t be able to use it.

packages/browser-rum-core/src/domain/contexts/sourceCodeContext.ts
amortemousque: Indeed done :)

packages/browser-core/src/domain/error/error.ts
```diff
@@ -68,6 +74,7 @@ export function computeRawError({
     componentStack,
     originalError,
     ...errorBase,
+    debugIds: resolvedStackTrace ? getDebugIds(getStackTraceUrls(resolvedStackTrace)) : undefined,
```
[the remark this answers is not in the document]
amortemousque: Fix it https://github.com/DataDog/browser-sdk/pull/4812/changes/877eea7a437d4331b4ae2912ef9ed196a38b615f

packages/browser-core/src/domain/error/isError.ts
amortemousque, line 1: Prevent dependency cycle between error and telemetry modules.

packages/browser-core/src/domain/console/consoleObservable.ts
```diff
@@ -34,10 +35,13 @@ type ConsoleLogForApi<T extends ConsoleApiName> = T extends typeof ConsoleApiNam
...
 
-export function initConsoleObservable<T extends ConsoleApiName[]>(apis: T): Observable<ConsoleLogForApi<T[number]>> {
+export function initConsoleObservable<T extends ConsoleApiName[]>(
+  apis: T,
+  sourceCodeContext?: SourceCodeContext
+): Observable<ConsoleLogForApi<T[number]>> {
   const consoleObservables = apis.map((api) => {
     if (!consoleObservablesByApi[api]) {
-      consoleObservablesByApi[api] = createConsoleObservable(api) as any // we are sure that the observable created for this api will yield the expected ConsoleLog type
+      consoleObservablesByApi[api] = createConsoleObservable(api, sourceCodeContext) as any // we are sure that the observable created for this api will yield the expected ConsoleLog type
```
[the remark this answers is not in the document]
amortemousque: In the next PR, the Logs SDK will pass sourceCodeContext.

packages/browser-core/src/domain/sourceCodeContext.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+
+    if (!hasFiredTelemetry) {
+      hasFiredTelemetry = true
+      addTelemetryUsage({
+        feature: 'source-code-context',
+        use_debug_id: useDebugId,
+        use_service: useService,
+        use_version: useVersion,
```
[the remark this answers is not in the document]
amortemousque: It’s a reasonable approximation.

packages/browser-core/src/domain/error/error.ts
```diff
@@ -68,7 +107,8 @@ export function computeRawError({
     componentStack,
     originalError,
     ...errorBase,
-    causes: isError(originalError) ? flattenErrorCauses(originalError, source) : undefined,
+    debugIds: getErrorDebugIds(errorStackTrace, errorCauses),
```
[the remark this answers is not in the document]
amortemousque: If an error stack contains a frame from a chunk, that chunk has already started executing, so the injected snippet at the top of the chunk has already populated DD_SOURCE_CODE_CONTEXT.

packages/browser-rum-core/src/domain/contexts/sourceCodeMfeContext.ts
```diff
@@ -0,0 +1,45 @@
+import { computeStackTrace, getSourceCodeContext } from '@datadog/browser-core'
+import { SKIPPED } from '@datadog/js-core/assembly'
+import type { DefaultRumEventAttributes, AssembleHookParams, AssembleHook } from '../hooks'
+
+export function startSourceCodeMfeContext(assembleHook: AssembleHook) {
+  assembleHook.register(({ domainContext, rawRumEvent }): DefaultRumEventAttributes | SKIPPED => {
+    const url = getSourceUrl(domainContext, rawRumEvent)
```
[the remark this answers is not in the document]
amortemousque: Yes but no big deal

packages/browser-core/src/domain/error/error.ts
amortemousque: In practice, stack traces are small (10 frames by default in V8), so the lookup cost is negligible. Also I’ll add debugId to Logs in a follow-up PR :)

packages/browser-core/src/domain/sourceCodeContext.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export interface BrowserWindow {
+  DD_SOURCE_CODE_CONTEXT?: Record<string, SourceCodeContextEntry>
+}
+
+const contextByUrl = new Map<string, SourceCodeContextEntry>()
+const processedStacks = new Set<string>()
+
+function syncSourceCodeContext() {
+  const sourceCodeContext = (window as BrowserWindow).DD_SOURCE_CODE_CONTEXT
```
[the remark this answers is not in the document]
amortemousque: fixed https://github.com/DataDog/browser-sdk/pull/4812/changes/426e2e191105696f46ef6211a351229436894cf1

rgaignault (approved):

packages/browser-core/src/domain/sourceCodeContext.ts
```diff
@@ -0,0 +1,60 @@
+import { computeStackTrace } from '../tools/stackTrace/computeStackTrace'
+import { isEmptyObject } from '../tools/utils/objectUtils'
+
+export interface SourceCodeContextEntry {
+  service?: string
+  version?: string
+  ddDebugId?: string
+}
```
bcaudan, line 17: **💭 thought:** ‏could it be interesting to document the possible cases, like:
```ts
SourceCodeContextEntry = {
  service: string
  version?: string
} | {
  ddDebugId: string
} | {
  ddDebugId: string
  service: string
  version?: string
}
```

packages/browser-core/src/domain/sourceCodeContext.ts
```diff
@@ -0,0 +1,83 @@
...
+      use_service: useService,
+      use_version: useVersion,
+    })
+  }
+}
+
+export function getDebugIds(urls: string[]): { [url: string]: string } | undefined {
+  syncSourceCodeContext()
+
+  const debugIds: { [url: string]: string } = {}
```
bcaudan: **💬 suggestion:** ‏clearer naming


```suggestion
export function buildDebugIdByUrl(urls: string[]): { [url: string]: string } | undefined {
  syncSourceCodeContext()

  const debugIdByUrl: { [url: string]: string } = {}
```

packages/browser-core/src/domain/sourceCodeContext.spec.ts
```diff
@@ -65,6 +65,21 @@ describe('getSourceCodeContext', () => {
   it('should return undefined for an unknown URL', () => {
     expect(getSourceCodeContext(url)).toBeUndefined()
   })
+
+  it('should ignore updates to an existing URL after it has been read', () => {
```
bcaudan, line 69: **❓ question:** ‏what is the intent for this cache?

packages/browser-core/src/domain/error/error.ts
```diff
@@ -48,6 +46,33 @@ function computeErrorBase({
...
+export function flattenErrorCauses(
+  error: ErrorWithCause
+): Array<{ originalError: unknown; stackTrace: StackTrace | undefined }> {
+  const chain: Array<{ originalError: unknown; stackTrace: StackTrace | undefined }> = []
+  let current: unknown = error.cause
+
+  while (current !== undefined && current !== null && chain.length < 10) {
+    const stackTrace = isError(current) ? computeStackTrace(current) : undefined
+    chain.push({ originalError: current, stackTrace })
+    current = isError(current) ? (current as ErrorWithCause).cause : undefined
```
bcaudan: **🥜 nitpick:** ‏A single const `isCurrentError = isError(current)` at the top of the loop could avoid the double `isError` Call.

packages/browser-rum-core/src/domain/contexts/sourceCodeMfeContext.ts
```diff
@@ -0,0 +1,45 @@
+import { computeStackTrace, getSourceCodeContext } from '@datadog/browser-core'
+import { SKIPPED } from '@datadog/js-core/assembly'
+import type { DefaultRumEventAttributes, AssembleHookParams, AssembleHook } from '../hooks'
+
+export function startSourceCodeMfeContext(assembleHook: AssembleHook) {
```
bcaudan, line 13: **💬 suggestion:** ‏it is a bit implicit right now so it could be interesting to document the intent somewhere:
- Service/version attribution uses only the top frame URL.
- Debug IDs cover all bundles across the full cause chain, not just the top frame.

packages/browser-rum-core/src/domain/longTask/longTaskCollection.ts
```diff
@@ -71,9 +71,12 @@ function processEntry(
       },
     }
   }
+  const scriptUrls = entry.scripts.map((script) => script.sourceURL).filter((url): url is string => !!url)
+  const debugIds = getDebugIds(scriptUrls)
```
bcaudan: **💬 suggestion:** ‏what about adding test cases for that?

packages/browser-rum-core/src/domain/error/errorCollection.ts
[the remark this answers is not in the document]
bcaudan: @amortemousque what do you think of codex comment?

bcaudan, line 86: **💬 suggestion:** ‏would it be worth adding a test case with debug ids?

test/e2e/scenario/microfrontend.scenario.ts
```diff
@@ -430,7 +430,63 @@ test.describe('microfrontend', () => {
...
+    createTest('errors from console.error should have debug_id from source code context')
+      .withRum(RUM_CONFIG)
+      .withSetup(microfrontendSetup)
+      .run(async ({ intakeRegistry, flushEvents, page, withBrowserLogs }) => {
+        await page.click('#app1-console-error')
+        await flushEvents()
+
+        expect(intakeRegistry.rumErrorEvents).toHaveLength(2)
+        expect(intakeRegistry.rumErrorEvents[0]._dd?.debug_ids).toBeDefined()
+        expect(intakeRegistry.rumErrorEvents[1]._dd?.debug_ids).toBeDefined()
```
bcaudan: **💬 suggestion:** ‏could we assert on something more specific?

test/apps/microfrontend/webpack.app1.js
```diff
@@ -18,6 +18,7 @@ module.exports = webpackBase({
         sourceCodeContext: {
           service: 'mfe-app1-service',
           version: '1.0.0',
+          debugId: true,
```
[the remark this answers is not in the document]
amortemousque: I updated the plugin. It's now working

test/e2e/scenario/microfrontend.scenario.ts
```diff
@@ -430,7 +430,63 @@ test.describe('microfrontend', () => {
     })
   })
 
-  test.describe('Logs', () => {
+  test.describe('RUM debug_id attribution', () => {
+    createTest('errors from console.error should have debug_id from source code context')
+      .withRum(RUM_CONFIG)
+      .withSetup(microfrontendSetup)
+      .run(async ({ intakeRegistry, flushEvents, page, withBrowserLogs }) => {
+        await page.click('#app1-console-error')
```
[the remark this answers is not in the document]
amortemousque: Updated

packages/browser-core/src/domain/sourceCodeContext.spec.ts
amortemousque: This test already existed in the previous sourceCodeContext.ts, so I kept it. IIUC, the rationale was to keep attribution stable throughout the session, even if the URL context is mutated (e.g., when the cache is refreshed or by an unexpected third-party script). Do you think this is unnecessary?

packages/browser-rum-core/src/domain/error/errorCollection.ts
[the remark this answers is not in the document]
amortemousque: Yes, a legitimate one. I see two options:
1 Add _dd.debug_ids to the list of modifiable fields and let the customer manage it.
2 Build debug_ids after beforeSend. This changes the approach significantly but provides a better DX.

The downside of option 2 is that we’d need to reparse the stack traces.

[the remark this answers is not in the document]
amortemousque: After talking with @BenoitZugmeyer, I learned that COOP wants to add a URL redaction API. I’ll track that as a follow-up. In the meantime, I’ve added _dd.debug_ids to the list of modifiable fields, and I’ll update the documentation if the URL redaction API isn’t available by GA.

packages/browser-core/src/domain/sourceCodeContext.ts
amortemousque: I added a comment instead. I found the union more confusing than helpful. The LLM’s rationale was interesting 🙂 :
* External data — DD_SOURCE_CODE_CONTEXT is injected at runtime, so TypeScript can’t enforce a union on data it doesn’t construct.
* No discriminant — the three cases overlap, which makes type narrowing awkward.

packages/browser-rum-core/src/domain/assembly.ts
```diff
@@ -33,6 +33,7 @@ const MODIFIABLE_FIELD_PATHS_BY_EVENT: Record<AssembledRumEvent['type'], Modifia
     'error.handling_stack': 'string',
     'error.resource.url': 'string',
     'error.fingerprint': 'string',
+    '_dd.debug_ids': 'object',
```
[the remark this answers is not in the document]
amortemousque: I'll fix the schema.

packages/browser-rum-core/src/domain/error/errorCollection.ts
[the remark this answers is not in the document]
bcaudan: ok, great!

bcaudan (approved):
LGTM

PR #4825, opened 2026-06-26
✨ Use optional MatchHeader name to account for default headers

bdibon:
## Motivation

Enabling `trackResourceHeaders` with custom matchers while still collecting the built-in defaults previously required spreading `DEFAULT_TRACKED_RESOURCE_HEADERS` manually:

```ts
trackResourceHeaders: [
  ...DEFAULT_TRACKED_RESOURCE_HEADERS.map((name) => ({ name })),
  { name: 'x-request-id' },
]
```

This is verbose and forces customers to import (NPM) or reference (CDN) the `DEFAULT_TRACKED_RESOURCE_HEADERS` constant. This change adds an ergonomic way to pull in the defaults inline, by making `name` optional on a `MatchHeader` entry.

## Changes

- **`MatchHeader` type** (`configuration.ts`): `name` is now optional. A matcher without `name` targets the built-in default headers instead of a specific one, and can still be scoped with `url`, `location`, etc.
- **Validation** (`configuration.ts`): `validateAndBuildTrackResourceHeaders` now accepts entries without `name` as-is (instead of expanding them at config-build time), only validating `name` when present. Warning messages were adjusted accordingly (e.g. `trackResourceHeaders[0].name should be a MatchOption`).
- **Header filtering** (`resourceCollection.ts`): `filterHeaders` falls back to a `defaultHeadersMatchOption` (checks membership in `DEFAULT_TRACKED_RESOURCE_HEADERS`) when a matcher's `name` is absent, so the expansion now happens at collection time rather than at config validation time.
- **TSDoc**: documented that omitting `name` targets the default headers, with an example combining it with `location: 'response'`.
- **Deprecation**: marked the `DEFAULT_TRACKED_RESOURCE_HEADERS` field on the `DD_RUM` public API as `@deprecated`, now pointing to the optional-`name` matcher instead of the previously introduced `{ includeDefaults: true }` sentinel (which this commit replaces before it shipped). The field/value is kept to avoid a breaking change.
- **Tests**: updated/added cases for a bare `{}` matcher, `{ location: 'response' }` without `name`, combining it with a custom matcher, and `name: undefined` being treated like an absent `name`.
- **E2E**: updated the resources scenario to use `[{}, { name: 'x-request-id' }]` instead of spreading `DEFAULT_TRACKED_RESOURCE_HEADERS`.

New usage:

```ts
trackResourceHeaders: [{}, { name: 'x-request-id' }]
```

### Automated

```bash
yarn test:unit --spec packages/browser-rum-core/src/domain/configuration/configuration.spec.ts
yarn test:unit --spec packages/browser-rum-core/src/domain/resource/resourceCollection.spec.ts
yarn typecheck
yarn lint
```

### Manual (end-to-end)

Start the dev server and create a temporary sandbox page that combines a bare (default-headers) matcher with a custom matcher:

```bash
yarn dev-server start

cat > sandbox/test-default-headers.html << 'HTML'
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>Test default headers matcher</title>
    <script src="/datadog-rum.js"></script>
    <script>
      DD_RUM.init({
        clientToken: 'xxx',
        applicationId: 'xxx',
        proxy: '/proxy',
        trackResources: true,
        trackResourceHeaders: [{}, { name: 'x-custom-header' }],
      })
      function makeRequest() {
        fetch('/datadog-rum.js', { headers: { 'x-custom-header': 'hello' } })
      }
    </script>
  </head>
  <body>
    <button id="fetch-btn" onclick="makeRequest()">Fetch</button>
  </body>
</html>
HTML
```

Then exercise it (dev-server URL from `yarn dev-server status`):

```bash
yarn dev-server intake clear
agent-browser open http://localhost:8081/test-default-headers.html
agent-browser click '#fetch-btn'
# wait for the fetch's PerformanceResourceTiming, then flush
agent-browser reload
yarn dev-server intake rum-resources \
  | jq -c 'select(.resource.type=="fetch") | {request_headers: .resource.request.headers, response_headers: .resource.response.headers}'
```

Expected output — the custom matcher (`x-custom-header`) is captured on the request **and** the default headers (`content-type`, `content-length`, …) are captured on the response:

```json
{"request_headers":{"x-custom-header":"hello"},"response_headers":{"content-type":"text/javascript; charset=utf-8","content-length":"2429864"}}
```

Cleanup:

```bash
yarn dev-server stop
rm sandbox/test-default-headers.html
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-rum-core/src/domain/configuration/configuration.ts
+++ b/packages/browser-rum-core/src/domain/configuration/configuration.ts
@@ -22,31 +22,7 @@ export const DEFAULT_PROPAGATOR_TYPES: PropagatorType[] = ['tracecontext', 'data
 
 /**
  * Default list of headers collected on resource events when {@link RumInitConfiguration.trackResourceHeaders | trackResourceHeaders}
- * is set to `true`. Re-exported by the `@datadog/browser-rum` and `@datadog/browser-rum-slim` packages, and exposed on the
- * `DD_RUM` global object when the SDK is loaded via the CDN, so it can be referenced when building a custom matcher list.
- *
- * @example NPM
- * ```ts
- * import { datadogRum, DEFAULT_TRACKED_RESOURCE_HEADERS } from '@datadog/browser-rum'
- *
- * datadogRum.init({
- *   // ...
- *   trackResourceHeaders: [
- *     ...DEFAULT_TRACKED_RESOURCE_HEADERS.map((name) => ({ name })),
- *     { name: 'x-request-id' },
- *   ],
- * })
- * ```
- * @example CDN
- * ```ts
- * DD_RUM.init({
- *   // ...
- *   trackResourceHeaders: [
- *     ...DD_RUM.DEFAULT_TRACKED_RESOURCE_HEADERS.map((name) => ({ name })),
- *     { name: 'x-request-id' },
- *   ],
- * })
- * ```
+ * is set to `true`.
  */
 export const DEFAULT_TRACKED_RESOURCE_HEADERS = [
   'cache-control',
@@ -283,11 +259,9 @@ export interface RumInitConfiguration extends InitConfiguration {
    *
    * - `true`: collect {@link DEFAULT_TRACKED_RESOURCE_HEADERS} for all URLs, both directions
    * - `MatchHeader[]`: each {@link MatchHeader} targets a header name, with optional URL scope
-   * (`url`), value extraction (`extractor`), and `location`. By default, both request and
-   * response headers are captured; set `location` to `'request'` or `'response'` to restrict
-   * to one.
-   * - `{ includeDefaults: true }` (as an array entry): expands to {@link DEFAULT_TRACKED_RESOURCE_HEADERS}
-   * in place, so the defaults can be combined with custom matchers without listing them manually.
+   * (`url`), value extraction (`extractor`), and `location`. When `name` is omitted, the matcher
+   * applies to {@link DEFAULT_TRACKED_RESOURCE_HEADERS}. By default, both request and response
+   * headers are captured; set `location` to `'request'` or `'response'` to restrict to one.
    *
    * Headers whose names match a built-in sensitive-data pattern are always dropped, regardless
    * of the configured matchers. The pattern blocks headers whose names contain: `token`, `cookie`,
@@ -298,7 +272,10 @@ export interface RumInitConfiguration extends InitConfiguration {
    * @defaultValue false (disabled)
    * @example
    * // Collect default headers plus custom ones for all URLs
-   * trackResourceHeaders: [{ includeDefaults: true }, { name: 'x-request-id' }]
+   * trackResourceHeaders: [{}, { name: 'x-request-id' }]
+   * @example
+   * // Collect default headers from responses only
+   * trackResourceHeaders: [{ location: 'response' }]
    * @example
    * // URL-scoped rule: capture specific response headers only for calls to /api
    * trackResourceHeaders: [{ url: /\/api\//, name: 'cache-control', location: 'response' }]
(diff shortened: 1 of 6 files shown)


packages/browser-rum-core/src/domain/configuration/configuration.ts
```diff
@@ -294,18 +296,15 @@ export interface RumInitConfiguration extends InitConfiguration {
...
+   * trackResourceHeaders: [{ includeDefaults: true }, { name: 'x-request-id' }]
    * @example
    * // URL-scoped rule: capture specific response headers only for calls to /api
    * trackResourceHeaders: [{ url: /\/api\//, name: 'cache-control', location: 'response' }]
    * @example
    * // Extract a partial value from a header
    * trackResourceHeaders: [{ url: /\/api\//, name: 'server-timing', extractor: /dur=(\d+)/, location: 'response' }]
    */
-  trackResourceHeaders?: boolean | MatchHeader[] | undefined
+  trackResourceHeaders?: boolean | Array<MatchHeader | { includeDefaults: true }> | undefined
```
BenoitZugmeyer: **suggestion:** make `includeDefaults` a `MatchHeader` option?

bdibon: I don't think it's a good idea, if we want to do this we either have to:

A)

Add includeDefaults field on type MatchHeader, then we may end up with non-sensical entries such as `{ name: 'x-my-header', includeDefaults: true}`, we could make name optional, but we will have to handle possibly undefined matchHeader.name

What's nice in the current design is that validateAndBuildTrackResourceHeaders always resolves entries of the configuration option to a MatchHeader, which the rest of the SDK uses.

B)

We do something like `type MatchHeader = HeaderMatcher | IncludeDefaultsOption`, SDK uses internally the HeaderMatcher type. It adds a type whose name will be hard to not be confusing.

BenoitZugmeyer: I like the optional `MatchHeader.name` because:

* it is more powerful, i.e. the user can include default headers for a specific URL only, or specific "location". Solution B only allows to include the default headers on all URLs and locations

* it doesn't add a new data model/concept into the mix

> What's nice in the current design is that validateAndBuildTrackResourceHeaders always resolves entries of the configuration option to a MatchHeader, which the rest of the SDK uses.

Having an optional `name` is trivial to handle, I would argue that it makes things even simpler. See [this commit](https://github.com/DataDog/browser-sdk/commit/6a302b1ec8a98c63003aa54098944f7b41acd3c4) for illustration.

mormubis: I like that `{ includeDefaults: true }` is self-explanatory without looking at the docs. Could we extend it to also support `url` and `location` to cover the scoping use case? What do you think?

BenoitZugmeyer (commented):
See my comment [here](https://github.com/DataDog/browser-sdk/pull/4825#discussion_r3504401012)

packages/browser-rum-core/src/domain/configuration/configuration.ts
bdibon: I didn't think about doing it this way, it's powerful and introduces minimal changes, but I don't like that we have to rely on the convention that `{}` is equivalent to the `DEFAULT_TRACKED_RESOURCE_HEADERS`, this can be mitigated by presenting it as a consequence of a "omit the name to get default header names" behavior in the docs.

TLDR; implemented it!

bdibon: It introduces changes and a weird type alias for the internal usage of `MatchHeader` in the SDK. I have followed @BenoitZugmeyer solution which is a minimal change and shouldn't confuse users too much.

BenoitZugmeyer (approved):

PR #4833, opened 2026-06-29
✨ Add betaEnableViewUpdates config option

mormubis:
## Motivation

Enabling partial view updates currently requires setting `enableExperimentalFeatures: ['partial_view_updates']`, which is an internal mechanism not meant for customers. To share the feature with beta customers, we need a proper public-facing config option.

## Changes

Added `betaEnableViewUpdates?: boolean` to `RumInitConfiguration`. When set to `true`, the partial view updates optimization is enabled directly — `createBatchDispatcher` receives it as a plain boolean, no experimental feature flag involved. The `partial_view_updates` experimental flag was removed.

## Test instructions

```js
DD_RUM.init({
  // ...
  betaEnableViewUpdates: true,
})
```

Check the Network tab — intermediate view updates should appear as `view_update` events instead of full `view` events.

## Checklist

- [x] Tested locally
- [ ] Tested on staging
- [ ] Updated documentation

### DIFF
--- a/packages/browser-core/src/tools/experimentalFeatures.ts
+++ b/packages/browser-core/src/tools/experimentalFeatures.ts
@@ -15,7 +15,6 @@ import { objectHasValue } from './utils/objectUtils'
 // eslint-disable-next-line no-restricted-syntax
 export enum ExperimentalFeature {
   TRACK_INTAKE_REQUESTS = 'track_intake_requests',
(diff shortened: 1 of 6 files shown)


BenoitZugmeyer:
Can we just drop the experimental flag?

BenoitZugmeyer (approved):

thomas-lebeau (approved):

PR #4838, opened 2026-07-01
👷 sign chrome-bump commits with commit-headless

thomas-lebeau:
## Motivation

The chrome-bump CI job pushed its version-bump commit directly with `git push`, producing an unverified commit on the created branch. This currently works because we have admin rights to bypass the signed-commit policy on the repo, but we're about to lose that bypass — so CI-generated commits need to be properly signed (Verified) going forward.

## Changes

- Add a `commit` OctoSTS scope (main branch only) and a chainguard policy (`self.gitlab.commit.sts.yaml`) authorizing it for the GitLab CI job
- Install `commit-headless` in the CI image (Dockerfile) to create signed (Verified) commits through the GitHub API
- Add `pushSignedCommit()` in `gitUtils.ts`, which pushes HEAD as a new remote branch via `commit-headless push` and points the local branch at it, so a later `gh pr create` doesn't attempt its own unsigned push
- Use `pushSignedCommit()` in `bump-chrome-version.ts` instead of `git push`
- Bump `CURRENT_CI_IMAGE` to 117 for the updated Dockerfile

## Test instructions

It's hard to test before merging, but when it's merged we can:

- Manually run the `bump-chrome-version-scheduled` GitLab CI job with `CHROME_PACKAGE_VERSION` overridden to an old version (e.g. a version with a lower major than what's currently released), so the job detects a newer Chrome version and creates a bump PR
- Verify the job pushes a **Verified** commit on the new `bump-chrome-version-to-<major>` branch
- Verify the resulting PR is opened correctly against `main`
- Close the test PR/branch afterwards (no need to merge it)

## Checklist

- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/scripts/lib/gitUtils.ts
+++ b/scripts/lib/gitUtils.ts
@@ -90,6 +90,7 @@ export function pushSignedCommit(branch: string): void {
   using token = getGithubCommitToken()
   command`commit-headless push -T DataDog/browser-sdk --branch ${branch} --create-branch --head-sha ${headSha} --reset`
     .withEnvironment({ GITHUB_TOKEN: token.value })
+    .withLogs()
     .run()
   command`git branch --set-upstream-to=origin/${branch} ${branch}`.run()
 }
--- a/scripts/lib/gitUtils.ts
+++ b/scripts/lib/gitUtils.ts
@@ -77,9 +77,10 @@ export function createPullRequest(mainBranch: string, labels?: string[]) {
 
 /**
  * Push the current HEAD commit to a new remote branch as a signed (Verified) commit, using
- * commit-headless to create it through the GitHub API instead of a plain `git push`. The local
- * branch is then made to track the newly created remote branch, so callers (e.g. `gh pr create`)
- * don't attempt an unsigned push of their own.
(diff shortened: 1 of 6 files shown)


scripts/lib/gitUtils.ts
```diff
@@ -74,6 +75,25 @@ export function createPullRequest(mainBranch: string, labels?: string[]) {
...
+ * unsigned push of their own.
+ */
+export function pushSignedCommit(branch: string): void {
+  // `--create-branch` has no default branch point, so `--head-sha` must be given explicitly: the
+  // parent of HEAD, i.e. the commit the local branch was created from.
+  const headSha = command`git rev-parse HEAD^`.run().trim()
+
+  using token = getGithubCommitToken()
+  command`commit-headless push -T DataDog/browser-sdk --branch ${branch} --create-branch --head-sha ${headSha} --reset`
+    .withEnvironment({ GITHUB_TOKEN: token.value })
```
BeltranBulbarellaDD, line 92: Why not `withLogs`?

BeltranBulbarellaDD (approved):

PR #4839, opened 2026-07-02
🐛 fix logs not being sent long after session expiration

thomas-lebeau:
## Motivation

Session context history entries were evicted based on elapsed time (`SESSION_TIME_OUT_DELAY`). Once a session expired and enough time passed (~4h), `browser-logs` lost the ability to determine whether a session had ever been legitimately tracked, and stopped sending logs entirely.

## Changes

- `valueHistory`: support a `maxEntries`-only mode (no time-based expiration) alongside the existing `expireDelay` mode
- `sessionManager`: keep session context entries bounded by count (`MAX_SESSION_CONTEXT_HISTORY_ENTRIES`) instead of elapsed time, and add a `maxAge` option to `findTrackedSession` so callers can opt out of the default `TRACKED_SESSION_MAX_AGE` cutoff
- `browser-logs` `sessionContext`: use `maxAge: Infinity` when deciding whether to discard a log, so logs keep being sent indefinitely once a session was legitimately tracked, independent of the session id attached to the event (still subject to `TRACKED_SESSION_MAX_AGE` for the session id itself)
- Un-skip the e2e test that documented this bug (previously added with `test.fail()`)

## Test instructions

- `yarn test:unit --spec packages/browser-core/src/tools/valueHistory.spec.ts`
- `yarn test:unit --spec packages/browser-logs/src/domain/session/sessionManager.spec.ts`
- `yarn test:e2e -g "logs stop being sent"` (or the relevant test name in `test/e2e/scenario/logs.scenario.ts`)

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-logs/src/domain/contexts/sessionContext.ts
+++ b/packages/browser-logs/src/domain/contexts/sessionContext.ts
@@ -9,10 +9,16 @@ export function startSessionContext(
   sessionManager: SessionManager
 ) {
   hook.register(({ startTime }) => {
+    // Used to attach a (fresh, safe-to-reference) session id: subject to the default
+    // TRACKED_SESSION_MAX_AGE, so it becomes undefined once the session is too old to reference.
     const session = sessionManager.findTrackedSession(startTime)
 
+    // Used for the discard decision: unlike `session` above, this ignores session age (logs
+    // should keep being sent indefinitely, with or without a session, once a session was
+    // legitimately tracked here) but still respects sampling.
     const isSessionTracked = sessionManager.findTrackedSession(startTime, {
       returnInactive: true,
+      maxAge: Infinity,
     })
 
     if (!isSessionTracked) {
--- a/test/e2e/scenario/logs.scenario.ts
+++ b/test/e2e/scenario/logs.scenario.ts
@@ -1,4 +1,6 @@
 import { DEFAULT_REQUEST_ERROR_RESPONSE_LENGTH_LIMIT } from '@datadog/browser-logs/src/domain/configuration'
+import { ONE_HOUR, ONE_MINUTE } from '@datadog/js-core/time'
+import { SESSION_EXPIRATION_DELAY } from '@datadog/browser-core'
 import { test, expect } from '@playwright/test'
 import { createTest, createWorker } from '../lib/framework'
 import { APPLICATION_ID } from '../lib/helpers/configuration'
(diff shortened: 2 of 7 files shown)


test/e2e/scenario/logs.scenario.ts
```diff
@@ -354,4 +356,38 @@ test.describe('logs', () => {
       expect(intakeRegistry.logsEvents).toHaveLength(1)
       expect(intakeRegistry.logsEvents[0].foo).toBe('bar')
     })
+
+  test.describe('session expiration', () => {
+    createTest('logs should keep being sent forever, even long after the session has expired')
+      .withLogs()
+      .withMockClock()
+      .run(async ({ intakeRegistry, flushEvents, page }) => {
```
BenoitZugmeyer, line 364: **praise:** 👏

packages/browser-logs/src/domain/contexts/sessionContext.ts
```diff
@@ -9,10 +9,16 @@ export function startSessionContext(
...
+    // Used to attach a (fresh, safe-to-reference) session id: subject to the default
+    // TRACKED_SESSION_MAX_AGE, so it becomes undefined once the session is too old to reference.
     const session = sessionManager.findTrackedSession(startTime)
 
+    // Used for the discard decision: unlike `session` above, this ignores session age (logs
+    // should keep being sent indefinitely, with or without a session, once a session was
+    // legitimately tracked here) but still respects sampling.
     const isSessionTracked = sessionManager.findTrackedSession(startTime, {
       returnInactive: true,
+      maxAge: Infinity,
```
BenoitZugmeyer, line 21: **nitpick:** I wonder if we couldn't collapse the two options, as they are only used together (here). Maybe `returnInactive: true` could skip the age check? It would simplify a tiny bit. Not a huge deal either way.

BenoitZugmeyer (approved):

packages/browser-logs/src/domain/contexts/sessionContext.ts
thomas-lebeau: we certainly could but that's be less explicit, I think it's easier to understand like this. If you don't mind I'll keep it this way.

BenoitZugmeyer: I don't mind!

PR #4872, opened 2026-07-12
✅ Add Salesforce Experience Cloud testing app

BeltranBulbarellaDD:
## Motivation

Adds Experience Cloud testing application for Salesforce LWC. 

## Changes

Created new `sf-experience-app` which inits the SDK when query param `?init=true`. This way we can leverage this test app to test both regular init and head markup. 

Updated `build-test-apps` to build rum-slim into this application too. 

Updated and renamed `scripts/salesforce-lwc-apps.ts` to build and deploy also this app. 

I left out the CSP config and Intake file because we already deploy it to the sandbox org when building and deploying the lwc-app.

## Test instructions

N/A. CI should pass normally.

## Checklist

- [x] Tested locally
- [ ] Tested on staging


BeltranBulbarellaDD:
@codex review

BeltranBulbarellaDD:
@codex review

scripts/salesforce-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+const APPS: Record<AppKey, { dir: string; urlPath: string; buildBundle: boolean; siteName?: string }> = {
+  lwc: {
+    dir: resolve(repositoryRoot, 'test/apps/sf-lwc-app'),
+    urlPath: '/lightning/app/c__SF_LWC_App/page/home',
+    buildBundle: true,
+  },
+  'experience-cloud': {
+    dir: resolve(repositoryRoot, 'test/apps/sf-experience-app'),
+    urlPath: 'sfexperiencecloud',
+    buildBundle: false,
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Yeah this is expected. This is not a command that will be run on a clean checkout.

package.json
```diff
@@ -15,8 +15,8 @@
     "build": "yarn workspaces foreach --all --parallel --topological-dev run build",
     "build:bundle": "yarn workspaces foreach --all --parallel run build:bundle",
     "build:apps": "node scripts/build/build-test-apps.ts",
-    "salesforce:deploy-app": "yarn workspace @datadog/browser-rum-slim build:bundle && node --env-file-if-exists=.env scripts/salesforce-lwc-app.ts deploy-app",
```
BenoitZugmeyer, line 18: **question:** is `salesforce-lwc-app.ts` (singular) still used somewhere? Maybe we should delete it.

eslint.config.ts
```diff
@@ -53,6 +53,7 @@ export default defineConfig(
       'test/apps/nuxt-app',
       'test/apps/nuxt-vue-router-v4-app',
       'test/apps/sf-lwc-app',
+      'test/apps/sf-experience-app',
```
BenoitZugmeyer: **question:** remind me, why do we ignore the test app?

.prettierignore
```diff
@@ -13,6 +13,7 @@ yarn.lock
 /developer-extension/.output
 /developer-extension/.wxt
 /test/apps/sf-lwc-app/*
+/test/apps/sf-experience-app/*
```
BenoitZugmeyer, line 18: **question:** here too, why do we ignore the app?

scripts/salesforce-lwc-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+import { dirname, resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { tmpdir } from 'node:os'
+import { parseArgs } from 'node:util'
+
+import { printLog, runMain } from './lib/executionUtils.ts'
+import { getSfLwcClientId, getSfLwcInstanceUrl, getSfLwcJwtPrivateKey, getSfLwcUsername } from './lib/secrets.ts'
+import { command } from './lib/command.ts'
+
+const repositoryRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..')
```
BenoitZugmeyer: **suggestion:** just use `import.meta.dirname`

scripts/salesforce-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+
+const APP_KEYS: AppKey[] = ['lwc', 'experience-cloud']
+
+const APPS: Record<AppKey, { dir: string; urlPath: string; buildBundle: boolean; siteName?: string }> = {
+  lwc: {
+    dir: resolve(repositoryRoot, 'test/apps/sf-lwc-app'),
+    urlPath: '/lightning/app/c__SF_LWC_App/page/home',
+    buildBundle: true,
+  },
+  'experience-cloud': {
```
BenoitZugmeyer, line 23: **suggestion:** since this script supports both type of apps, maybe we can rename it to `salesforce-apps.ts`?

scripts/salesforce-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+    const path = new URL(urlPath, 'https://salesforce.local')
+
+    if (siteName) {
+      console.log(new URL(`${path.pathname}/`, getSalesforceSiteUrl()).href)
+      continue
+    }
+
+    console.log(new URL(`${path.pathname}${path.search}`, getSfLwcInstanceUrl()).href)
+  }
+}
```
BenoitZugmeyer, line 148: **suggestion:** it looks like `urlPath` is only used here. We could avoid this URL formatting logic if we just stored the url directly instead of `urlPath`.

```ts
const APPS: Record<AppKey, { dir: string; urlPath: string; buildBundle: boolean; siteName?: string }> = {
  lwc: {
    dir: resolve(repositoryRoot, 'test/apps/sf-lwc-app'),
    url: new URL('/lightning/app/c__SF_LWC_App/page/home', getSfLwcInstanceUrl()).href,
    buildBundle: true,
  },
  'experience-cloud': {
    dir: resolve(repositoryRoot, 'test/apps/sf-experience-app'),
    url: new URL('sfexperiencecloud/', getSalesforceSiteUrl()).href,
    buildBundle: false,
    siteName: 'SF Experience Cloud App',
  },
```

scripts/salesforce-lwc-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+}
+
+function deployApp(appKeys: AppKey[]) {
+  for (const appKey of appKeys) {
+    const { dir, buildBundle, siteName } = APPS[appKey]
+
+    if (buildBundle) {
+      printLog('Building RUM slim bundle...')
+      command`yarn workspace ${'@datadog/browser-rum-slim'} build:bundle`.withLogs().run()
```
BenoitZugmeyer: **nitpick:** Why do you need the string interpolation here? This should work just the same:


```suggestion
      command`yarn workspace @datadog/browser-rum-slim build:bundle`.withLogs().run()
```

scripts/salesforce-lwc-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+    throw new Error(
+      `Usage: node scripts/salesforce-lwc-apps.ts <${SUPPORTED_COMMANDS.join('|')}> [--app ${APP_KEYS.join('|')}]`
+    )
+  }
+
+  const commandName = positionals[0]
+  const appKeys = resolveAppKeys(values.app)
+
+  switch (commandName) {
+    // Deploy the app(s) to the Salesforce org. To be done only when an app is updated.
```
BenoitZugmeyer: **nitpick:** remove those comments, add them to the usage log instead

scripts/salesforce-apps.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (buildBundle) {
+      printLog('Building RUM slim bundle...')
+      command`yarn workspace ${'@datadog/browser-rum-slim'} build:bundle`.withLogs().run()
+    }
+
+    authenticate(TARGET_ORG, dir)
+    rmSync(resolve(dir, '.sf'), { recursive: true, force: true })
+
+    printLog(`Deploying Salesforce ${appKey} app to ${TARGET_ORG}...`)
+    command`sf project deploy start --target-org ${TARGET_ORG} --source-dir force-app --ignore-conflicts --concise`
```
BenoitZugmeyer, line 124: **question:** how the `sf` binary gets installed?

package.json
BeltranBulbarellaDD: Yeah, I forgot to delete it.

.prettierignore
BeltranBulbarellaDD: In this case it's because of this change that prettier does which breaks deployment. 

> Event handler should be an expression (13:9)

Screenshot 2026-07-17 at 09 32 04

scripts/salesforce-apps.ts
BeltranBulbarellaDD: True I missed it.

eslint.config.ts
BeltranBulbarellaDD: I tried something [here](https://github.com/DataDog/browser-sdk/blob/cbc8697c80add0c0476465e1bd21399481729576/eslint.config.ts), what do you think?

scripts/salesforce-apps.ts
```diff
@@ -0,0 +1,158 @@
...
+function printUrl(appKeys: AppKey[]): void {
+  for (const appKey of appKeys) {
+    const { siteName, url } = APPS[appKey]
+
+    if (siteName) {
+      console.log(url)
+      continue
+    }
+
+    console.log(url)
```
BenoitZugmeyer: **nitpick:** this is a bit pointless 😄

BenoitZugmeyer (approved):

PR #4877, opened 2026-07-14
✨ Extend remote configuration supported parameters

mormubis:
## Motivation

The multi-platform RFC defines the full set of fields that should be remotely configurable for the browser SDK. We were only handling a subset. This adds the missing ones from the RFC browser table marked as "\[Product\] Keep".

## Changes

Added 7 new fields to `SUPPORTED_RUM_FIELDS`: `telemetrySampleRate`, `trackAnonymousUser`, `trackUserInteractions`, `trackResources`, `trackLongTasks`, `traceContextInjection`, and `actionNameAttribute`.

Also added support for `profiling.sampleRate`, which lives in a separate top-level namespace in the schema (not under `rum`). To handle this, `applyRemoteConfiguration` now receives the full `RemoteConfiguration` instead of just the `rum` part. The cache format changed accordingly, so `CACHE_VERSION` is bumped to 2.

`logs.*` fields (`forwardErrorsToLogs`, `forwardConsoleLogs`, `forwardReports`) are not included here. They belong to `browser-logs`, not `browser-rum-core`, and need a separate approach to avoid a double fetch.

RFC: https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6675005886/RFC+-+Remote+Configuration+Parameters

## Test instructions

1. Set up a remote configuration with one of the new fields, e.g. `trackResources: false`
2. Initialize the SDK with that remote config ID
3. Verify the field is applied (resources should not be collected)
4. Check that fields not present in the remote config keep their init values

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
+++ b/packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
@@ -757,7 +757,7 @@ describe('remoteConfiguration', () => {
       })
     })
 
-    it('should apply the new RFC fields when present in the rum section', () => {
+    it('should apply telemetrySampleRate, trackResources, trackLongTasks, and actionNameAttribute from the rum section', () => {
       const result = applyRemoteConfiguration(
         DEFAULT_INIT_CONFIGURATION,
         {
--- a/packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
+++ b/packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
@@ -117,7 +117,19 @@ describe('remoteConfiguration', () => {
       })
     })
 
-    it('should throw an error if the remote config does not contain rum config', async () => {
+    it('should succeed if the remote config contains only a profiling section', async () => {
+      interceptor.withFetch(() =>
+        Promise.resolve({
+          ok: true,
+          json: () => Promise.resolve({ profiling: { sampleRate: 10 } }),
+        })
+      )
+
+      const fetchResult = await fetchRemoteConfiguration(configuration)
+      expect(fetchResult).toEqual({ ok: true, value: { profiling: { sampleRate: 10 } } })
+    })
+
+    it('should return an error if the remote config does not contain rum or profiling', async () => {
       interceptor.withFetch(() =>
         Promise.resolve({
           ok: true,
--- a/packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
+++ b/packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
@@ -6,13 +6,14 @@ import { appendElement } from '../../../test'
 import type { RumInitConfiguration } from './configuration'
 import {
   type RumRemoteConfiguration,
+  type RemoteConfiguration,
   initMetrics,
   applyRemoteConfiguration,
   buildEndpoint,
   fetchRemoteConfiguration,
   getRemoteConfiguration,
 } from './remoteConfiguration'
-import { buildCacheKey } from './remoteConfigurationCache'
+import { buildCacheKey, CACHE_VERSION } from './remoteConfigurationCache'
 
 const DEFAULT_INIT_CONFIGURATION: RumInitConfiguration = {
   clientToken: 'xxx',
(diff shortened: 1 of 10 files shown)


packages/browser-rum-core/src/domain/configuration/remoteConfigurationCache.ts
```diff
@@ -1,14 +1,14 @@
 import { timeStampNow } from '@datadog/js-core/time'
 import { tryJsonParse } from '@datadog/browser-core'
 import type { TimeStamp } from '@datadog/js-core/time'
-import type { RumRemoteConfiguration } from './remoteConfiguration'
+import type { RemoteConfiguration } from './remoteConfiguration'
 
-export const CACHE_VERSION = 1
+export const CACHE_VERSION = 2
```
mormubis, line 6: Previously the cache stored only the `rum` part of the config. Now it stores the full `RemoteConfiguration` (which also includes `profiling`). The version bump ensures old caches are discarded on upgrade rather than misread.

packages/browser-rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -86,15 +93,16 @@ export async function fetchAndApplyRemoteConfiguration(
 
 export function applyRemoteConfiguration(
   initConfiguration: RumInitConfiguration,
-  rumRemoteConfiguration: RumRemoteConfiguration & { [key: string]: unknown },
+  remoteConfiguration: RemoteConfiguration,
   supportedContextManagers: SupportedContextManagers,
   metrics: ReturnType<typeof initMetrics>
 ): RumInitConfiguration {
+  const rumRemoteConfiguration = remoteConfiguration.rum as RumRemoteConfiguration & { [key: string]: unknown }
```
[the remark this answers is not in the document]
bdibon: This looks like a legit issue @mormubis

bdibon (approved):
Besides Codex legit concern, it looks good to me!

packages/browser-rum-core/src/domain/configuration/remoteConfiguration.spec.ts
```diff
@@ -731,6 +756,38 @@ describe('remoteConfiguration', () => {
         )
       })
     })
+
+    it('should apply the new RFC fields when present in the rum section', () => {
```
gogusarov: `minor` 

it's better to change the name of the test, because if someone reads that in couple of months, they would have no clue what was the `old` rfc and what was considered `new` at the moment of creation of this test 🙂

packages/browser-rum-core/src/domain/configuration/remoteConfiguration.ts
```diff
@@ -86,24 +93,30 @@ export async function fetchAndApplyRemoteConfiguration(
...
+        appliedConfiguration[option] = resolveConfigurationProperty(rumRemoteConfiguration[option])
+      }
+    })
+    ;(Object.keys(supportedContextManagers) as Array<keyof SupportedContextManagers>).forEach((context) => {
+      if (rumRemoteConfiguration[context] !== undefined) {
+        resolveContextProperty(supportedContextManagers[context], rumRemoteConfiguration[context])
+      }
+    })
+  }
+  if (remoteConfiguration.profiling?.sampleRate !== undefined) {
```
gogusarov, line 117: `question`

I don't know much about profiling, but is `sampleRate` here the only property we would probably want to configure? I saw that it's the only one mentioned in RFC but want to double check. Because if not then probably it would be better to do smth similar to what we already had and introduce second white list for profiling and loop through it

gogusarov (approved):

packages/browser-rum-core/src/domain/configuration/remoteConfiguration.ts
mormubis: @gogusarov I'm not sure. Given that it's a different product, I would keep them separate, mainly because I know the schema will be very difficult to change if we need to add anything. At the same time, I understand that there won't be any more parameters.

What do you think?

gogusarov: Ok, let's keep it like that for now

packages/browser-rum-core/src/domain/configuration/remoteConfigurationCache.ts
aleksandr-gringauz: @mormubis 

question. Why do you need to bump the cache version? Does the schema stop being backward compatible?

mormubis: @aleksandr-gringauz Not really, it's just a precaution.

PR #4878, opened 2026-07-15
✨ Shopify Integration - implement dedicated bundle

lierniel:
## Motivation

Following the [[RFC] Datadog RUM Shopify integration — checkout coverage via Web Pixel](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6833636068), we're going with the Custom Pixel approach. Its core deliverable is a code snippet that customers paste into their Custom Pixel extension in Shopify Admin.
We want that pasted snippet to be as small as possible while still wiring up full Shopify event bindings into RUM. That calls for a dedicated browser-rum-shopify package — shipped as a versioned CDN bundle, referenced from the snippet — that encapsulates the event bindings and the SDK's context-specific initialization logic, instead of inlining that logic into the snippet itself. It also means we can ship fixes or improvements to the Shopify integration on the CDN without forcing customers to re-paste anything.

Find more details here: https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6955696402/browser-rum-shopify+package+implementation

## Changes

* Exposed `makeRecorderApi` and `makeProfilerApi` from `@datadog/browser-rum/internal` package, so I can reuse them in `browser-rum-shopify`'s main entry
* Implemented `browser-rum-shopify` package
* Updated deployments `packages` const to include newly added shopify package and updated deployments scripts` tests to reflect that change
* Extended computeBundleSize `packages` const with shopify bundle

**Out of scope for this PR**:
* Telemetry
* E2E tests

## Test instructions

Here is [Shopify Dev Store](https://dev-store-custom-pixel-y3yx2lca.myshopify.com/) (password is `napray`) I've created for testing during the local development. I connected `datadog-rum-shopify-skoshcheev.ts` bundle (which I deployed with `deploy-feature` gitlab task to staging) to it in the following way:

Theme Liquid:
```html
<head>
    ...
    <script>
        const DD_APPLICATION_ID      = '7a4c736c-0e91-49a5-aadc-b234a454ec7c';
        const DD_CLIENT_TOKEN        = 'pub3dddd74bb4e210cb05e310b5ca32037e';
        const DD_SITE                = 'datad0g.com';
        const DD_SERVICE             = 'service';
        const DD_ENV                 = 'prod';
        const DD_VERSION             = '1.0.0';
        const DD_SESSION_SAMPLE_RATE = 100;
        
        const sdkUrl = 'https://d3uc069fcn7uxw.cloudfront.net/datadog-rum-shopify-skoshcheev.js';
        
        window.DD_RUM = window.DD_RUM || {
          q: [],
          onReady(c) { this.q.push(c); },
        };
        
        const script = document.createElement('script');
        script.async = 1;
        script.crossOrigin = 'anonymous';
        script.src = sdkUrl;
        document.head.appendChild(script);
        
        DD_RUM.onReady(function(){
          DD_RUM.init({
            applicationId: DD_APPLICATION_ID,
            clientToken: DD_CLIENT_TOKEN,
            site: DD_SITE,
            service: DD_SERVICE,
            env: DD_ENV,
            version: DD_VERSION,
            sessionSampleRate: DD_SESSION_SAMPLE_RATE,
            sessionReplaySampleRate: 100,
            trackUserInteractions: true,
            trackResources: true,
            tracklongTasks: true,
            defaultPrivacylevel: 'mask-user-input',
          });
          DD_RUM.startSessionReplayRecording();
        })
    </script>
    ...
</head>
```

Custom Pixel:
```js
const DD_APPLICATION_ID      = '7a4c736c-0e91-49a5-aadc-b234a454ec7c';
const DD_CLIENT_TOKEN        = 'pub3dddd74bb4e210cb05e310b5ca32037e';
const DD_SITE                = 'datad0g.com';
const DD_SERVICE             = 'service';
const DD_ENV                 = 'prod';
const DD_VERSION             = '1.0.0';
const DD_SESSION_SAMPLE_RATE = 100;

const sdkUrl = 'https://d3uc069fcn7uxw.cloudfront.net/datadog-rum-shopify-skoshcheev.js';

window.DD_RUM = window.DD_RUM || {
  q: [],
  onReady(c) { this.q.push(c); },
};

const script = document.createElement('script');
script.crossOrigin = 'anonymous';
script.src = sdkUrl;
document.head.appendChild(script);

DD_RUM.onReady(function(){
  DD_RUM.init({
    applicationId: DD_APPLICATION_ID,
    clientToken: DD_CLIENT_TOKEN,
    site: DD_SITE,
    service: DD_SERVICE,
    env: DD_ENV,
    version: DD_VERSION,
    sessionSampleRate: DD_SESSION_SAMPLE_RATE,
    // pass Shopify analytics to SDK init options
    shopifyAnalytics: analytics,
  })
})
```

And here is Datadog RUM App, where you can track how events are sent: https://dd.datad0g.com/rum/sessions?query=%40type%3Asession%20%40application.id%3A7a4c736c-0e91-49a5-aadc-b234a454ec7c&agg_m=count&agg_m_source=base&agg_t=count&cols=&fromUser=false&refresh_mode=sliding&viz=stream&from_ts=1784162672472&to_ts=1784249072472&live=true

## Checklist

- [x] Tested locally
- [x] Tested on staging


lierniel:

packages/browser-rum-shopify/src/boot/patchSandboxedIframeApis.ts
```diff
@@ -0,0 +1,35 @@
...
+ * which silently breaks session tracking and view/URL resolution:
+ * - `cookieStore.getAll()` is blocked; `document.cookie` works fine as a fallback.
+ * - `navigator.locks.request()` throws a synchronous `SecurityError` (the Web Locks API is
+ * denied without Permissions Policy delegation), but the SDK only catches an async rejection.
+ * - `getPristineWindow()` creates a nested iframe to read an unpatched `URL` constructor; that
+ * nested iframe's `contentWindow` is cross-origin here, and reading a property off it throws.
+ * Shimming these away makes the SDK fall back to code paths that do work in this context
+ * (`document.cookie`, same-document promise chaining, `globalObject.URL`).
+ */
+export function patchSandboxedIframeApis() {
```
amortemousque, line 17: **💬 suggestion:** ‏What about handling these API accesses gracefully within the SDK where they’re called? That’s the strategy we’ve been using to support environments like Salesforce and Web Workers in the Logs SDK.

packages/browser-rum-shopify/src/boot/patchSandboxedIframeApis.ts
```diff
@@ -0,0 +1,35 @@
+import { globalObject } from '@datadog/browser-core'
+
+/**
+ * A Shopify Custom Pixel runs in a "lax" sandboxed iframe (`sandbox="allow-scripts"`, no
```
amortemousque, line 4: **💭 thought:** Nice findings! The Browser SDK is meant to run inside iframes, we are documenting it (https://datadoghq.dev/browser-sdk/interfaces/_datadog_browser-rum.RumInitConfiguration.html#usepartitionedcrosssitesessioncookie). So unsupported APIs should be handled by default in the SDK. This isn’t specific to this integration.

We could also add an end-to-end test that runs the SDK inside an iframe to prevent regressions.

packages/browser-rum-shopify/src/boot/patchSandboxedIframeApis.ts
BeltranBulbarellaDD: To add to this, we added [this function](https://github.com/DataDog/browser-sdk/blob/stephan.koshcheev/RUM-17232/shopify-integration-implement-dedicated-rum-sdk-package/packages/browser-core/src/browser/addEventListener.ts#L153-L156), and what we do is we try to support the event and if it fails we just noop it.

packages/browser-rum-shopify/src/domain/shopifyAnalytics.ts
```diff
@@ -0,0 +1,43 @@
...
+  id: string
+  timestamp: string
+  context?: {
+    document?: {
+      location?: { href?: string }
+      title?: string
+    }
+  }
+  data?: TData
+}
```
BeltranBulbarellaDD, line 16: Should this go in rum-events-format? (in the future?)

packages/browser-rum-shopify/src/domain/shopifyBindings.ts
```diff
@@ -0,0 +1,139 @@
+import type { RumPublicApi } from '@datadog/browser-rum-core'
+import type { ShopifyAnalyticsApi, ShopifyPixelEvent } from './shopifyAnalytics'
+
+export interface CheckoutData {
+  token?: string
+}
+
```
BeltranBulbarellaDD: Same though here about Rum-events-format.

BeltranBulbarellaDD (commented):
Overall it looks good to me!

packages/browser-rum-shopify/src/domain/shopifyAnalytics.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { globalObject } from '@datadog/browser-core'
+import type { ShopifyAnalyticsApi } from './shopifyAnalytics'
+import { getShopifyAnalytics } from './shopifyAnalytics'
+
+interface TestGlobal {
+  analytics?: ShopifyAnalyticsApi
+  _DATADOG_SHOPIFY_FORCE_PIXEL_CONTEXT?: boolean
```
amortemousque: **❓ question:** ‏Why do we need _DATADOG_SHOPIFY_FORCE_PIXEL_CONTEXT?

packages/browser-rum-shopify/src/domain/shopifyBindings.ts
```diff
@@ -0,0 +1,139 @@
...
+  // mapped as a custom action for funnel friction analysis.
+  analytics.subscribe('alert_displayed', (event: ShopifyPixelEvent<{ alert?: AlertData }>) => {
+    const alert = event.data?.alert
+    rumPublicApi.addAction('alert_displayed', {
+      message: alert?.message,
+      type: alert?.type,
+      target: alert?.target,
+      value: alert?.value,
+    })
+  })
```
amortemousque: **💬 suggestion:** I don’t think we should wire these domain events in the first iteration.
A few reasons:
* Custom actions are intended to capture customer-defined events. It's not ment for our integrations. Customers who want these events could always instrument them themselves.
* It’s not clear that custom actions are the right abstraction. Some of these events may fit better as Vitals or Operations, depending on how we want to surface them.
* Shipping default mappings also creates a long-term API commitment. We’d likely need an opt-out mechanism and a clear strategy for evolving those mappings over time.

I think we should keep the initial integration focused on the core instrumentation and revisit those event mappings once we have a clearer product direction and customer feedback.

packages/browser-rum-shopify/src/domain/shopifyAnalytics.spec.ts
lierniel: I've added it just as a flag to disable automatic detection of the Custom Pixel context. But now I see that it makes no sense to have it, so I'll remove it 👍

packages/browser-rum-shopify/src/domain/shopifyBindings.ts
lierniel: Yeah, makes sense, I'll remove the wirings. However, just to confirm, should we keep binding for Click and Error (`ui_extension_errored`) events?

packages/browser-rum-shopify/src/domain/shopifyAnalytics.ts
lierniel: I'm not sure if it makes sense to add it to rum-events-format. It's just the interface of Shopify internal events, not the extension of the RUM events. I'm listening to the Shopify event and then bind them to standard RUM events

But maybe I misunderstood the purpose of rum-events-format, let me know if I'm wrong :)

packages/browser-rum-shopify/src/boot/patchSandboxedIframeApis.ts
lierniel: Sure, but do you mind if I add a TODO here and tackle it in the following PRs? Don't wont this PR to handle too many things :)

packages/browser-rum-shopify/src/domain/shopifyBindings.ts
amortemousque: Yes, I think we can keep ui_extension_errored as it maps naturally to an Error event.

I have a few questions about the context fields:

1. Could you share a few examples of the values in these fields? That would help assess how they should map into the SDK.
2. We already populate error.type from stackTrace.name ([here](https://github.com/DataDog/browser-sdk/blob/75ce84ff5e2603e44f362e2fdb9e6be50a7ecf79/packages/browser-core/src/domain/error/error.ts#L27)⁠￼). Using another field called type here could be confusing.
3. Do we need the `extension` namespace? I’m not sure `appName` is specific to the extension?

lierniel: Yeah, I'll give it a closer look

packages/browser-rum-shopify/src/entries/main.ts
amortemousque, line 1: **💬 suggestion:** ‏I think we can simplify the boot flow:
1 - In main.ts we can merge the sandbox/storefront branches into one path, and creates DD_RUM synchronously instead of waiting on page_viewed:

```ts
const shopifyAnalytics = getShopifyAnalytics()
const isCustomPixelSandbox = !!shopifyAnalytics

const datadogRum = makeShopifyRumPublicApi(
  makeRumPublicApi(makeRecorderApi(), makeProfilerApi(), { sdkName: 'rum-shopify' }),
  isCustomPixelSandbox
)

defineGlobal(global, 'DD_RUM', datadogRum)

if (isCustomPixelSandbox) {
  patchSandboxedIframeApis()
  initShopifyBindings(datadogRum, shopifyAnalytics)
}

```
2 - Move checkout-path gating to the page_viewed binding instead. If the view is not started no events will flow:
```ts
analytics.subscribe('page_viewed', (event) => {
    const url = event.context?.document?.location?.href
    if (!url || !CHECKOUT_PATH.test(url)) {
      return
    }
    rumPublicApi.startView({
      name: event.context?.document?.title,
      url,
    })
  })
```
3 - Delete stubRecorderApi.ts and stubProfilerApi.ts and use profilingSampleRate: 0 and sessionReplaySampleRate: 0 instead.
```ts 
export function makeShopifyRumPublicApi(datadogRum: RumPublicApi, isSandbox: boolean): RumPublicApi {
  return {
    ...datadogRum,
    init(initConfiguration: RumInitConfiguration) {
      datadogRum.init(
        isSandbox
          ? {
              ...initConfiguration,
              trackViewsManually: true, // Views are started explicitly via startView()
              sessionReplaySampleRate: 0, // Session Replay is not usable in the Pixel sandbox iframe
              profilingSampleRate: 0, // Real User Profiling is not usable in the Pixel sandbox iframe
              trackUserInteractions: false, // Pixel sandbox iframe has no real DOM to interact with
              trackResources: false, // Iframe resources are not meaningful
              trackLongTasks: false, // PerformanceObserver tracks the empty iframe
            }
          : initConfiguration
      )
    },
  }
}

```

Also keep a single sdk_name: rum-shopify. We can always tell whether a session ran on a checkout page by looking at the view name and URL.

BeltranBulbarellaDD (approved):

packages/browser-rum-shopify/src/domain/shopifyBindings.ts
```diff
@@ -0,0 +1,65 @@
...
+  extensionTarget?: string
+  type?: string
+  appName?: string
+}
+
+// Matches /checkouts/*, /checkout, including locale-prefixed paths.
+// Storefront pages never match, so init() is a no-op there: those pages
+// already get a `DD_RUM` instance from the Theme Liquid snippet, and initializing here too would
+// create a second, independent SDK instance double-tracking the same page view.
+const CHECKOUT_PATH = /\/(([a-z]{2}(-[a-z0-9]+)?)\/)?(checkouts?)(\/|$)/i
```
BeltranBulbarellaDD, line 21: I tested this regex and it works good even for discount pages like:
`https://salesify-demo.myshopify.com/discount/welcome10?redirect=/checkout`
But for this page:
`https://crochetdaddy.myshopify.com/pages/checkout` which is a 404 and not a checkout page it will think it's a checkout path. It's not a big deal just wanted to acknowledge it. We can revisit the regex or maybe find another way to check the checkout.

packages/browser-rum-shopify/README.md
BeltranBulbarellaDD, line 1: IMO, it could be nice to expand a little the readme, like have some setup instructions. It could be done in a future PR.

BeltranBulbarellaDD (approved):
LGTM! Looking forward for the e2e tests.

amortemousque (approved):

packages/browser-rum-shopify/README.md
lierniel: Yeah, good idea! I don't yet prepared setup instructions, so I'll do it in a future PR

packages/browser-rum-shopify/src/domain/shopifyBindings.ts
lierniel: Thank you for testing, nice findings! 
That makes me think that testing current url agains CHECKOUT_PATH regex works only as a temporal short-term solution. I think we should try to establish connection with Theme Liquid's SDK instance and start Pixel iframe's SDK only if it failed. That was the initial idea - run Pixel iframe's SDK only if main SDK in Theme Liquid is not available and the current approach looks more like workaround

PR #4881, opened 2026-07-15
🐛 Keep primitive values unchanged during sanitization

rgaignault:
## Motivation

Primitive values should follow the primitive sanitization path even when their prototypes define `toJSON`.

## Changes

- Ignore inherited `toJSON` methods on primitive values.
- Add unit coverage for root and nested values.

## Test instructions

`yarn test:unit --spec packages/browser-core/src/tools/serialisation/sanitize.spec.ts`

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-core/src/tools/serialisation/sanitize.ts
+++ b/packages/browser-core/src/tools/serialisation/sanitize.ts
@@ -252,8 +252,13 @@ function sanitizeEvent(event: Event): SanitizedEvent {
  *
  */
 function tryToApplyToJSON(value: ExtendedContextValue) {
+  // Primitive values are sanitized directly and should not use inherited toJSON methods.
+  if ((typeof value !== 'object' || value === null) && typeof value !== 'function') {
+    return value
+  }
+
   const object = value as ObjectWithToJsonMethod
-  if (object && typeof object.toJSON === 'function') {
+  if (typeof object.toJSON === 'function') {
     try {
       return object.toJSON() as ExtendedContextValue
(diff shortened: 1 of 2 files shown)


bdibon (approved):

packages/browser-core/src/tools/serialisation/sanitize.ts
```diff
@@ -252,8 +252,13 @@ function sanitizeEvent(event: Event): SanitizedEvent {
  *
  */
 function tryToApplyToJSON(value: ExtendedContextValue) {
+  // Primitive values are sanitized directly and should not use inherited toJSON methods.
+  if ((typeof value !== 'object' || value === null) && typeof value !== 'function') {
```
BenoitZugmeyer, line 256: **Suggestion:** remove `&& typeof value !== 'function'` as:
* It's not a primitive, so the comment doesn't make sense
* It's not problematic, as we can have a function as WeakMap key

Alternative: at the very least, add a unit test to cover this.

rgaignault: I might be missing something, but isn’t it the opposite? Removing it would skip toJSON for functions, while keeping it lets them through as valid WeakMap keys.

BenoitZugmeyer: Ah I see, sorry

BenoitZugmeyer (approved):

Issue #4890, opened 2026-07-17
🐛 [rum-nuxt] Plugin keeps Nuxt's error-page handler armed after hydration, so post-hydration errors render the full-page error


mframos3:
**Describe the bug**

Product: `rum` (`@datadog/browser-rum-nuxt`, v7.x — confirmed still present on `main`).

The Nuxt RUM plugin keeps Nuxt's full-page error handler armed for the entire session, so **any uncaught error that happens after hydration renders the full-page error page** (`error.vue`), even non-fatal client errors that Nuxt would otherwise just log to the console.

**Root cause**

`setupNuxtErrorHandling` wraps `vueApp.config.errorHandler` and delegates to the original handler:

```ts
// packages/browser-rum-nuxt/src/domain/error/setupNuxtErrorHandling.ts
const original = nuxtApp.vueApp.config.errorHandler
nuxtApp.vueApp.config.errorHandler = (error, instance, info) => {
  deduplicatedReportError(error, instance, info)
  original?.(error, instance, info)
}
```

At `onInit` time (plugins run before mount), `original` is Nuxt's own `handleVueError`, which does `nuxt.payload.error ||= createError(error)` — i.e. it renders the full-page error. Nuxt intends this handler to be active **only during the initial render**, and removes it on `app:suspense:resolve`:

```js
// nuxt/dist/app/entry.js
vueApp.config.errorHandler = handleVueError;
nuxt.hook("app:suspense:resolve", () => {
  if (vueApp.config.errorHandler === handleVueError) {   // identity check
    vueApp.config.errorHandler = void 0;
  }
});
```

The removal is guarded by an **identity check**. Because the plugin replaced `config.errorHandler` with a wrapper, `config.errorHandler === handleVueError` is now `false`, Nuxt skips the removal, and the wrapper (which still calls `handleVueError`) stays wired for the whole session. Every post-hydration uncaught Vue error then propagates to it and renders the full-page error, instead of Nuxt's intended default of just logging it.

Note: startup errors are unaffected by the wrapping — they flow through Nuxt calling `handleVueError` directly in `entry.js` and through the plugin's `app:error` hook, not through `config.errorHandler`.

**To Reproduce**

1. A Nuxt 3 app with `nuxtRumPlugin` initialized (`nuxtApp` passed so error handling is set up).
[...]

PR #4899, opened 2026-07-21
🐛 [rum-nuxt] Fix Nuxt post-hydration error handling

BeltranBulbarellaDD:
## Motivation

[Issue](https://github.com/DataDog/browser-sdk/issues/4890).

Nuxt installs its default `vueApp.config.errorHandler` during initial rendering and removes it after hydration on `app:suspense:resolve`, as long as the handler was not overridden.

The Nuxt RUM plugin wraps that handler before hydration, which prevents Nuxt’s identity check from removing it. As a result, client-side Vue errors triggered after hydration keep delegating to Nuxt’s default error handler and render the full-page error.vue, even though Nuxt would normally only log them to the console.

## Changes

 - Stop delegating to Nuxt’s default error handler after `app:suspense:resolve`.
 - Continue reporting Vue errors to RUM after hydration.
 - Preserve the original behavior before hydration so initial render/startup errors can still trigger Nuxt’s error page.
 - Keep delegating to custom original Vue error handlers after hydration.
 - Add unit coverage for Nuxt default handler gating and custom handler preservation.
 - Added e2e test.

## Test instructions

1. Checkout main.
2. Start the Nuxt test app.
3. Open the Nuxt error test page without RUM.
4. Click `Trigger Error`.
5. Confirm the error is only logged to the console and the app does not render the Nuxt error page.
6. Open the same page with: `?rum-config={"applicationId":"test","clientToken":"test","site":"datadoghq.com"}`
7. Click `Trigger Error`.
8. On main, confirm the app renders the Nuxt 500 Internal Server Error page.
9. With this fix, confirm the error is reported to RUM but the app stays on the page and only logs the error to the console.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-rum-nuxt/src/domain/error/setupNuxtErrorHandling.spec.ts
+++ b/packages/browser-rum-nuxt/src/domain/error/setupNuxtErrorHandling.spec.ts
@@ -27,6 +27,70 @@ describe('setupNuxtErrorHandling', () => {
     expect(hookSpy).toHaveBeenCalledWith('app:error', jasmine.any(Function))
   })
 
+  it("stops calling Nuxt's default error handler after hydration", () => {
+    const reportErrorSpy = jasmine.createSpy()
+    const originalErrorHandlerSpy = jasmine.createSpy() as jasmine.Spy & { __nuxt_default?: true }
+    originalErrorHandlerSpy.__nuxt_default = true
+    let suspenseResolveCallback!: () => void
+    const nuxtApp = {
+      vueApp: {
+        config: {
+          errorHandler: originalErrorHandlerSpy,
+        },
+      },
+      hook: jasmine.createSpy().and.callFake((name: string, callback: () => void) => {
+        if (name === 'app:suspense:resolve') {
+          suspenseResolveCallback = callback
+        }
+      }),
+    } as unknown as NuxtApp
+
+    setupNuxtErrorHandling(nuxtApp, reportErrorSpy)
+
+    const initialError = new Error('initial')
+    const errorHandler = nuxtApp.vueApp.config.errorHandler as NonNullable<App['config']['errorHandler']>
+    errorHandler(initialError, null, 'mounted hook')
+    suspenseResolveCallback()
+
+    const postHydrationError = new Error('post hydration')
+    errorHandler(postHydrationError, null, 'native event handler')
+
+    expect(reportErrorSpy.calls.allArgs()).toEqual([
+      [initialError, null, 'mounted hook'],
+      [postHydrationError, null, 'native event handler'],
+    ])
+    expect(originalErrorHandlerSpy).toHaveBeenCalledOnceWith(initialError, null, 'mounted hook')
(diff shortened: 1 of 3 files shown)


BeltranBulbarellaDD:
@codex review

packages/browser-rum-nuxt/src/domain/error/setupNuxtErrorHandling.spec.ts
```diff
@@ -27,6 +27,70 @@ describe('setupNuxtErrorHandling', () => {
...
+        }
+      }),
+    } as unknown as NuxtApp
+
+    setupNuxtErrorHandling(nuxtApp, reportErrorSpy)
+
+    const initialError = new Error('initial')
+    const errorHandler = nuxtApp.vueApp.config.errorHandler as NonNullable<App['config']['errorHandler']>
+    errorHandler(initialError, null, 'mounted hook')
+    suspenseResolveCallback()
```
bdibon, line 54: **🥜 nitpick:** ‏Add line break for readability

```suggestion

    suspenseResolveCallback()
```

packages/browser-rum-nuxt/src/domain/error/setupNuxtErrorHandling.spec.ts
```diff
@@ -27,6 +27,70 @@ describe('setupNuxtErrorHandling', () => {
     expect(hookSpy).toHaveBeenCalledWith('app:error', jasmine.any(Function))
   })
 
+  it("stops calling Nuxt's default error handler after hydration", () => {
+    const reportErrorSpy = jasmine.createSpy()
+    const originalErrorHandlerSpy = jasmine.createSpy() as jasmine.Spy & { __nuxt_default?: true }
```
bdibon: **💬 suggestion:** ‏‏I would update naming or add a comment so it's easy to make the distinction while scanning the two tests between the "native" error handler and a "custom" one.

test/e2e/scenario/plugins/nuxtPlugin.scenario.ts
```diff
@@ -101,6 +101,23 @@ test.describe('plugin: nuxt error', () => {
             expect(browserLogs.filter((log) => log.level === 'error').length).toBeGreaterThan(0)
           })
         })
+      createTest('should not render the full-page error for non-fatal client errors after hydration')
```
bdibon: **💭 thought:** ‏I am struggling to make a connection between the test title and the test block. It should not render something but we're not asserting that something is not rendered. Maybe I just lack context.

bdibon (approved):
Good job! A couple of suggestions to improve tests readability.

test/e2e/scenario/plugins/nuxtPlugin.scenario.ts
BeltranBulbarellaDD: Added a comment.

mormubis (approved):

PR #4905, opened 2026-07-22
🔊 Track plugin usage via telemetry

mormubis:
## Motivation

Plugins are already captured in configuration telemetry via `serializeRumConfiguration`, but that only tells us what was configured before validation. There's no way to query which plugins customers have registered from usage telemetry. This adds a `use-plugin` event per plugin at init time.

## Changes

Calls `addTelemetryUsage({ feature: 'use-plugin', plugin_name })` for each entry in `initConfiguration.plugins`, right after `callPluginsMethod`. The cast to `RawTelemetryUsage` is needed because `use-plugin` isn't in the `rum-events-format` schema yet. A follow-up PR there will let us drop the cast once the type is generated.

## Test instructions

1. `yarn test:unit --spec packages/browser-rum-core/src/boot/preStartRum.spec.ts`

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-rum-core/src/boot/preStartRum.ts
+++ b/packages/browser-rum-core/src/boot/preStartRum.ts
@@ -1,6 +1,6 @@
 import { timeStampNow, clocksNow } from '@datadog/js-core/time'
 import type { TimeStamp } from '@datadog/js-core/time'
-import type { TrackingConsentState, DeflateWorker, Context, Telemetry, SessionManager } from '@datadog/browser-core'
+import type { TrackingConsentState, DeflateWorker, Context, Telemetry, SessionManager, RawTelemetryUsage } from '@datadog/browser-core'
 import {
   BufferedObservable,
   display,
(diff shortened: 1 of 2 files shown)


packages/browser-rum-core/src/boot/preStartRum.ts
```diff
@@ -243,6 +244,10 @@ export function createPreStartStrategy(
 
       callPluginsMethod(initConfiguration.plugins, 'onInit', { initConfiguration, publicApi })
 
+      initConfiguration.plugins?.forEach((plugin) => {
+        addTelemetryUsage({ feature: 'use-plugin', plugin_name: plugin.name } as unknown as RawTelemetryUsage)
```
mormubis, line 248: `use-plugin` is not in the rum-events-format schema yet, so we need the double cast to get past TypeScript. A PR to the upstream schema repo + yarn json-schemas:sync will let us drop this.

PR #4911, opened 2026-07-24
🐛 Align session-ended WebSocket teardown with the latest view's end clocks

bdibon:
## Motivation

When a RUM session expires, open WebSocket connections are finalized with `tracking_end_reason: 'session_end'`. The teardown used `clocksNow()` independently in both `trackViews` (to end the active view) and `webSocketCollection` (to close the websocket vital/resource). Because these two calls happened at slightly different times, the websocket's `end_time` could land **after** the view's `end_time` — causing the `websocket-closed` vital to be dropped during assembly and the resource's `end_time` to fall outside the lifespan of the view it was attached to.

## Changes

- Add a `SessionExpiredEvent` type (`{ endClocks: ClocksState }`) carried on the `SESSION_EXPIRED` lifecycle event, instead of `void`.
- `startRum` now captures `clocksNow()` once when the session expires and notifies `SESSION_EXPIRED` with those clocks, so every subscriber ends things at the exact same instant.
- `trackViews` ends the current view using the shared `endClocks` from `SESSION_EXPIRED` instead of computing its own.
- `webSocketCollection.flushOpenConnections` now accepts an optional `endClocks` param, using the shared clocks passed on `SESSION_EXPIRED` (falling back to `clocksNow()` for its own `stop()` call, where no shared clocks exist).
- Unit tests covering the new `SessionExpiredEvent` payload and the wired session-end path in `startRum`, `trackViews`, and `webSocketCollection`.
- E2E test (renamed to "collects the websocket-closed vital when the session expires") asserting the `websocket-closed` vital and resource `end_time` do not exceed the associated view's end time.

## Test instructions

```bash
yarn test:unit --spec packages/browser-rum-core/src/boot/startRum.spec.ts
yarn test:unit --spec packages/browser-rum-core/src/domain/view/trackViews.spec.ts
yarn test:unit --spec packages/browser-rum-core/src/domain/resource/webSocketCollection.spec.ts
yarn build:apps && yarn test:e2e -g "collects the websocket-closed vital when the session expires"
```

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-rum-core/src/domain/lifeCycle.ts
+++ b/packages/browser-rum-core/src/domain/lifeCycle.ts
@@ -10,6 +10,7 @@ import type { ViewEvent, ViewCreatedEvent, ViewEndedEvent, BeforeViewUpdateEvent
 import type { DurationVitalStart } from './vital/vitalCollection'
 import type { TrackedEventData } from './eventTracker'
 import type { ActionEventData } from './action/trackManualActions'
+import type { SessionExpiredEvent } from './session/session.types'
 
 export const enum LifeCycleEventType {
   // Contexts (like viewHistory) should be opened using prefixed BEFORE_XXX events and closed using prefixed AFTER_XXX events
@@ -93,7 +94,7 @@ export interface LifeCycleEventMap {
   [LifeCycleEventTypeAsConst.REQUEST_STARTED]: RequestStartEvent
   [LifeCycleEventTypeAsConst.REQUEST_COMPLETED]: RequestCompleteEvent
   [LifeCycleEventTypeAsConst.WEBSOCKET_COMPLETED]: WebSocketCompleteEvent
-  [LifeCycleEventTypeAsConst.SESSION_EXPIRED]: void
+  [LifeCycleEventTypeAsConst.SESSION_EXPIRED]: SessionExpiredEvent
   [LifeCycleEventTypeAsConst.SESSION_RENEWED]: void
   [LifeCycleEventTypeAsConst.PREPARE_URGENT_FLUSH]: UrgentFlushReason
   [LifeCycleEventTypeAsConst.RAW_RUM_EVENT_COLLECTED]: RawRumEventCollectedData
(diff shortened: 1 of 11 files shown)


packages/browser-rum-core/src/domain/lifeCycle.ts
```diff
@@ -93,7 +94,7 @@ export interface LifeCycleEventMap {
   [LifeCycleEventTypeAsConst.REQUEST_STARTED]: RequestStartEvent
   [LifeCycleEventTypeAsConst.REQUEST_COMPLETED]: RequestCompleteEvent
   [LifeCycleEventTypeAsConst.WEBSOCKET_COMPLETED]: WebSocketCompleteEvent
-  [LifeCycleEventTypeAsConst.SESSION_EXPIRED]: void
+  [LifeCycleEventTypeAsConst.SESSION_EXPIRED]: SessionExpiredEvent
```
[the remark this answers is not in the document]
bdibon: There are no other consumers.

test/e2e/scenario/rum/websockets.scenario.ts
```diff
@@ -87,6 +89,21 @@ test.describe('rum websockets', () => {
...
+      expect(closedVital!.session.id).toBe(wsWithSessionEnd!.session.id)
+      expect(closedVital!.view.id).toBe(wsWithSessionEnd!.resource.websocket.end_view_id)
+      expect(closedVital!.view.url).toBe(wsWithSessionEnd!.view.url)
+
+      const associatedView = intakeRegistry.rumViewEvents.find((event) => event.view.id === closedVital!.view.id)
+      expect(associatedView).toBeDefined()
+
+      const viewEndTime = associatedView!.date + associatedView!.view.time_spent / NANOSECONDS_PER_MILLISECOND
+      expect(closedVital!.date).toBeLessThanOrEqual(viewEndTime)
+      expect(wsWithSessionEnd!.resource.websocket.end_time).toBeLessThanOrEqual(viewEndTime)
```
bcaudan, line 106: **🥜 nitpick:** ‏Do we need to check that? I guess ensuring that we are collecting the closing vital should be enough for an e2e scenario

bcaudan (approved):

test/e2e/scenario/rum/websockets.scenario.ts
bdibon: It doesn't cost us much to have this and it prevents any regression where the vital is emitted with an incoherent timestamp.

PR #4912, opened 2026-07-24
✨ [FFL-2596] Feature Flags tab — overrides (stacked PR 3 of 4)

kellyw1806:
## Motivation

Stacked on **#4916** (catalog fetch). Allows users to create overrides for a flag's value (switch between variants) locally and have it take effect on the inspected page.

## Stack (review bottom-up)

| | PR | Base | What it adds | Status |
|---|---|---|---|---|
| 1 of 4 | #4913 · FFL-2597 | `main` | OAuth sign-in | ✅ merged |
| 2 of 4 | #4916 · FFL-2858 | `ffl-2597` | catalog browsing | ✅ merged |
| **3 of 4** | **#4912 (this)** · FFL-2596 | `main` | flag overrides | in review |
| 4 of 4 | FFL-2857 — branch `kelly.wang/ffl-2857` (PR not opened yet) | `ffl-2596` | team filter + token revocation | pending |

## Changes
- Override engine (`useFlagOverrides`) writing to the inspected page's `localStorage` per the `DatadogDevtools` wrapper contract; mutations serialized to avoid lost updates.
- Per-row variant buttons + revert, a manual override-by-key form, clear-all / save-and-reload.
- Page-lifecycle state machine (`loading | ready | error`) driven by `chrome.webNavigation`: hides the provider warning + blocks writes while the inspected page is navigating, and re-checks the wrapper marker over a short settle window after a reload — fixing a false "DatadogDevtools not detected" flash (see Design note).

## Design note — provider detection across a reload
Applying an override reloads the inspected page, and the `DatadogDevtools` wrapper's `initialize()` (which writes the `dd.dd_flag.devtools` marker) runs **asynchronously after load** — so reading the marker immediately after navigation briefly showed a false "not detected" warning.

We fix this **extension-side**: model the page as an explicit `loading | ready | error` lifecycle off `chrome.webNavigation`, and after navigation completes **re-check the marker for a short window** (resolving the moment it appears) rather than adding a lifecycle event to the SDK (which lives in the `openfeature-js-client` repo). Same result (no flash); the SDK event would be more deterministic, but we kept this a single-repo change to avoid a separate cross-repo review/release cycle given the merge target.

Per review (Beltran), the fix lands **in this PR** (not deferred) so what we ship is complete; the deterministic-event version is left as a documented follow-up, and the rare overlapping-navigation edge is noted as an accepted limitation in the code.

## Demo
This is a recording of what the three stacked PRs look like working together.

https://github.com/user-attachments/assets/8ad6ad56-875d-4b6c-a3fe-a48e748c9ee4

## Checklist
- [x] Tested locally
- [x] Tested on staging
- [x] Added unit tests for this change (`useFlagOverrides.spec.ts`).


kellyw1806:

developer-extension/src/panel/hooks/useFlagOverrides.ts
```diff
@@ -0,0 +1,173 @@
...
+import { createLogger } from '../../common/logger'
+import { evalInWindow } from '../evalInWindow'
+import type { FlagOverrideType } from './flagOverrideTypes'
+
+const logger = createLogger('useFlagOverrides')
+
+// Contract shared with @datadog/openfeature-browser's DatadogDevtools wrapper.
+// Keep these in sync with that package: the wrapper reads OVERRIDES_KEY once on
+// initialize() and writes DEVTOOLS_MARKER_KEY when it is composed into the provider stack.
+export const OVERRIDES_KEY = 'dd.dd_flag.overrides'
```
kellyw1806: These keys are a contract with `@datadog/openfeature-browser`'s `DatadogDevtools` wrapper so renaming them requires a coordinated change in that package, not a free rename.

developer-extension/src/panel/hooks/useFlagOverrides.ts
```diff
@@ -0,0 +1,173 @@
...
+export const DEVTOOLS_MARKER_KEY = 'dd.dd_flag.devtools'
+
+const REFRESH_INTERVAL = 2000
+
+export interface FlagOverride {
+  type: FlagOverrideType
+  // Any JSON value — objects/arrays are `object`; `null` is allowed (a flag/variant value can be
+  // null). The manual-entry form still rejects null via validateOverrideValue.
+  value: boolean | string | number | object | null
```
kellyw1806: Widened the override value to allow `null` (a variant value can legitimately be null). The manual-entry form still rejects null via `validateOverrideValue`. Is this okay or should variant-click also block null?

developer-extension/src/panel/hooks/useFlagOverrides.ts
```diff
@@ -0,0 +1,173 @@
...
+      // before a write can resolve after it and briefly revert the just-applied override.
+      if (seq === readSeq.current) {
+        setState((prev) => (deepEqual(prev, next) ? prev : next))
+      }
+    })
+  }, [])
+
+  useEffect(() => {
+    refresh()
+    const id = setInterval(refresh, REFRESH_INTERVAL)
```
kellyw1806: This polls the inspected page every 2s while the tab is mounted. Fine for a devtools panel IMO, but flagging in case prefer gating it on tab visibility.

developer-extension/src/panel/components/tabs/flagsTab/flagsTab.tsx
```diff
@@ -1,17 +1,46 @@
...
   const catalog = useFlagCatalog(auth)
-  const view = useFlagCatalogView(catalog.flags)
+  const view = useFlagCatalogView(catalog.flags, overrides)
+
+  const [pendingReload, setPendingReload] = useState(false)
+  const [addOpen, setAddOpen] = useState(false)
+
+  function applyOverride(flagKey: string, override: FlagOverride) {
+    setPendingReload(true)
+    void setOverride(flagKey, override)
```
greghuels: issue: This enables the reload button before the async override write has necessarily finished. `setOverride()` queues a read-modify-write through `evalInWindow`, but this promise is discarded, so a fast click on “Save Overrides and Refresh Page” can reload the inspected page before `writeOverrides()` updates `localStorage`; the `DatadogDevtools` wrapper then initializes with the stale overrides. Can we wait for the queued mutation to settle before enabling/reloading, or keep the button disabled while a write is in flight?

BeltranBulbarellaDD (commented):
I think polling is not the correct approach. Because for a moment, when we write the override to the `localStorage`, we reload the window and when the navigation is commited, the extension checks the `DatadogDevtools` which might not be initialized yet. So the warning appears, then the extension polls and removes the warning.

https://github.com/user-attachments/assets/b7ea9338-483a-44ea-a879-ef9150aa9b81

So I'm thinking we should have an explicit page lifecycle. 
Something like:
`type FlagPageState =
    | { status: 'loading' }
    | { status: 'ready'; overrides: FlagOverrides; devtoolsEnabled: boolean }
    | { status: 'error'; error: Error }`

On initial mount, we read the page state once. When `onBeforeNavigate fires`, we switch to loading, hide the provider warning, and block editing so overrides cannot be written while the inspected document is changing.

Once `onCompleted` or `onErrorOccurred` fires, we read the state again and unblock the tab. This replaces the navigation-related polling and avoids treating a temporary read failure as “provider not detected.”

Navigation events alone are not sufficient because `DatadogDevtools` is initialized asynchronously. The playground awaits `initFeatureFlags()` after initializing the Browser SDK, so `onCompleted` can fire before the provider wrapper has finished initializing. To handle this, the wrapper should emit a lifecycle event from `initialize()` and `onClose()`.
The extension relays that event through the existing content-script/background bridge and updates `devtoolsEnabled` immediately.

This leaves us with one initial read, one read after navigation, explicit updates after extension-driven writes, and provider lifecycle events. 

Let me know what you think!

developer-extension/src/panel/hooks/flagOverrideTypes.ts
```diff
@@ -0,0 +1,4 @@
+// The value type of a feature flag (and of an override for it), for the hooks layer. Intentionally
+// a separate copy of flagsTab's `FlagType`: the hooks layer must not import from the flagsTab
+// directory, so each layer declares its own copy of this small union.
+export type FlagOverrideType = 'BOOLEAN' | 'STRING' | 'INTEGER' | 'NUMERIC' | 'JSON'
```
BeltranBulbarellaDD: Could we reuse the one in `flagTypeConstants`? Or make a helper from both if hooks must not import from the tab?

developer-extension/src/panel/components/tabs/flagsTab/manualOverrideForm.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <Group justify="flex-end">
+        <Button size="xs" color="violet" onClick={submit}>
+          Apply override
+        </Button>
+      </Group>
+    </Stack>
+  )
+}
+
+function parseFormValue(type: FlagType, raw: boolean | string): FlagOverride['value'] {
```
BeltranBulbarellaDD, line 110: Could we reuse with the one in `flagCatalog.ts:48`

developer-extension/src/panel/hooks/useFlagOverrides.ts
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/**
+ * Returns an error message if the value doesn't match the declared type, or null if valid.
+ *
+ * `allowNull` accepts `null` (a valid JSON value that real flag variants can use) — the catalog
+ * passes it so a JSON `null` variant stays applyable; the manual-entry form leaves it off so a
+ * hand-typed empty value is still rejected. Either way a non-JSON `null` still fails the type check.
+ */
+export function validateOverrideValue(
```
BeltranBulbarellaDD: Same with `flagCatalog:48` and `manualOverrideForm:109` maybe we can reuse these 3.

developer-extension/src/panel/hooks/useFlagOverrides.ts
```diff
@@ -0,0 +1,357 @@
...
+}
+
+export async function deleteOverride(key: string): Promise<void> {
+  await applyOverrideStatement(`delete overrides[${JSON.stringify(key)}]`)
+}
+
+export async function clearAllOverrides(): Promise<void> {
+  await applyOverrideStatement('overrides = {}')
+}
+
```
BeltranBulbarellaDD: Could we have a `updateFlagState({ kind: 'set' | 'delete' | 'clear', ... })` to remove duplicate code.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.ts
```diff
@@ -131,6 +131,36 @@ export async function fetchFlagCatalog(
...
+        // Don't put the flag key in the message: this error is logged (console.error), and the panel
+        // forwards console logs to its own telemetry — the key is customer data.
+        throw new Error(`Failed to fetch flag: ${response.status} ${response.statusText}`)
+      }
+      const body = (await response.json()) as RawFeatureFlagsResponse
+      return mapResources(Array.isArray(body?.data) ? body.data : [])
+    })
+  )
+  return pages.flat()
+}
```
BeltranBulbarellaDD: `fetchFlagCatalog` and `fetchFlagsByKeys` are similar. Maybe we could have a `fetchFlagResources()` helper?

developer-extension/src/panel/hooks/useFlagOverrides.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+      latest = useFlagOverrides()
+      return null
+    }
+    act(() => root.render(React.createElement(Probe)))
+    registerCleanupTask(() => {
+      act(() => root.unmount())
+      ;(globalThis as any).chrome = previousChrome
+    })
+    return () => latest
+  }
```
BeltranBulbarellaDD: This is very similar to the one on line 143 we could reuse them.

developer-extension/src/panel/hooks/useFlagOverrides.ts
```diff
@@ -0,0 +1,357 @@
...
+    const next = await readFlagState()
+    if (mounted.current && next && seq === readSeq.current) {
+      setState((prev) => ({ ...prev, overrides: next.overrides, devtoolsEnabled: next.devtoolsEnabled }))
+    }
+  }, [])
+
+  // Reads the page and resolves the status, re-checking for the provider marker over a short window
+  // (see SETTLE_* above) so a not-yet-initialized wrapper doesn't read as "not detected".
+  const settle = useCallback(() => {
+    cancelSettle.current()
```
BeltranBulbarellaDD: I'd say this function is overlly complicated. If the provider lifecycle event cannot be added in this PR, we could do some improvements here. 
1. Extract polling into a plain async function, This replaces the nested settle()/tick()
2. Use one AbortController per page lifecycle, this can be seen in `src/panel/hooks/useDevServerStatus.ts`. So that on navigation or unmount, abort the current controller.
3. Make mutations return the resulting state that way we can get rid of `refreshOverrides`

BeltranBulbarellaDD (commented):
Overall, this looks great! Since this is the last PR in the series, I think it would be a good opportunity to bring everything together, reduce duplicated code, and simplify some of the state management.

I was thinking we could reorganize the Flags tab a little, something like:

  flagsTab/
  ├── flagsContext.tsx       # Connected tab state and actions
  ├── flagsRequests.ts       # Authenticated catalog requests
  ├── inspectedPageFlags.ts  # Read/write overrides and reload the page
  ├── flagTypes.ts           # Shared flag and override types
  ├── oauth.ts               # OAuth flow and token persistence
  ├── flagsTab.tsx           # Mostly rendering
  └── components/hooks...

The main change would be moving the orchestration out of `flagsTab.tsx` and into `FlagsContext`. The context would load the catalog, cache flag metadata, track inspected-page overrides, and expose actions such as apply, revert, clear, and reload. Those actions would use the reusable request modules and update the context with the resulting state.

This would keep the inspected page as the source of truth for overrides, avoid duplicating authentication and fetching logic, and leave the tab and its components focused on rendering state and invoking actions.

developer-extension/src/panel/hooks/useFlagOverrides.ts
kellyw1806: The shared read-modify-write is already extracted into applyOverrideStatement. the three functions are one-liners over it. Folding them into a kind switch would add a union + switch, which might end up adding complexity.

developer-extension/src/panel/hooks/useFlagOverrides.ts
kellyw1806: Merged the two parsers into parseTypedString. Kept validateOverrideValue separate (it validates a value, doesn't parse a string) but moved it into the same module so it's all in one place.

developer-extension/src/panel/hooks/useFlagOverrides.ts
kellyw1806: Happy to do 1 and3! For 2, i think we should leave cancellation as-is. useDevServerStatus aborts a fetch, but our read goes through inspectedWindow.eval, which cant be aborted. so an AbortController can't cancel the read, only discard the result afterward (but this is waht readSeq already does). this could introduce a stuck-on-loading risk.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
```diff
@@ -28,31 +24,95 @@ export function FlagCatalogBody({
...
+            flag={flag}
+            override={getOverride(overrides, flag.key)}
+            onSelectVariant={applyOverride}
+            onRevert={removeOverride}
+          />
+        ))}
+      </Box>
+    </>
+  )
+}
```
BeltranBulbarellaDD:
```suggestion
function FlagList({
  flags,
  borderColor,
  emptyMessage,
}: {
  flags: CatalogFlag[]
  borderColor: string
  emptyMessage?: ReactNode
}) {
  const { overrides, applyOverride, removeOverride } = useFlagsContext()

  return (
    <Box style={{ border: `1px solid ${borderColor}`, borderRadius: 'var(--mantine-radius-sm)' }}>
      {flags.length === 0 ? (
        <Text c="dimmed" p="md">
          {emptyMessage}
        </Text>
      ) : (
        flags.map((flag) => (
          <FlagRow
            key={flag.key}
            flag={flag}
            override={getOverride(overrides, flag.key)}
            onSelectVariant={applyOverride}
            onRevert={removeOverride}
          />
        ))
      )}
    </Box>
  )
}
```
Wdyt of something like this?
That way we can simplify above:
```
     <FlagList
        flags={bottomFlags}
        borderColor="var(--mantine-color-gray-2)"
        // `bottomFlags` is the page minus overridden flags (those are pinned above). Only call it
        // "no match" when the server total is 0; otherwise this page's flags are all overridden.
        emptyMessage={
          catalog.total === 0 ? 'No flags match.' : 'All flags on this page are overridden — see Local overrides above.'
        }
      />
    </>
```

developer-extension/src/panel/components/tabs/flagsTab/flagTypes.ts
```diff
@@ -0,0 +1,90 @@
...
+export type FlagType = (typeof FLAG_TYPES)[number]
+
+// Display labels matching the webapp's Type filter (NUMERIC shows as "Number").
+export const TYPE_LABELS = {
+  BOOLEAN: 'Boolean',
+  STRING: 'String',
+  INTEGER: 'Integer',
+  NUMERIC: 'Number',
+  JSON: 'JSON',
+} satisfies Record<FlagType, string>
```
BeltranBulbarellaDD:
```suggestion
interface FlagTypeConfig {
  label: string
  expectedJsType: 'boolean' | 'string' | 'number' | 'object'
  parseErrorMessage: string
}

// Display and validation behavior for each type. Keeping it in one descriptor makes FLAG_TYPES the
// ordered list while this remains the exhaustive source of per-type metadata.
export const FLAG_TYPE_CONFIG = {
  BOOLEAN: {
    label: 'Boolean',
    expectedJsType: 'boolean',
    parseErrorMessage: 'Enter true or false',
  },
  STRING: {
    label: 'String',
    expectedJsType: 'string',
    parseErrorMessage: 'Enter a value',
  },
  INTEGER: {
    label: 'Integer',
    expectedJsType: 'number',
    parseErrorMessage: 'Enter a whole number within the safe integer range',
  },
  NUMERIC: {
    label: 'Number',
    expectedJsType: 'number',
    parseErrorMessage: 'Enter a valid number',
  },
  JSON: {
    label: 'JSON',
    expectedJsType: 'object',
    parseErrorMessage: 'Enter valid JSON',
  },
} satisfies Record<FlagType, FlagTypeConfig>
```
Would this simplify? Then in other places we can do for example:
`label: FLAG_TYPE_CONFIG[type].label`

developer-extension/src/panel/components/tabs/flagsTab/inspectedPageFlags.ts
```diff
@@ -0,0 +1,116 @@
...
+export function sanitizeOverrides(overrides: Record<string, unknown>): FlagOverrides {
+  const sanitized: FlagOverrides = {}
+  for (const [key, entry] of Object.entries(overrides)) {
+    if (isFlagOverride(entry)) {
+      sanitized[key] = entry
+    }
+  }
+  return sanitized
+}
+
```
BeltranBulbarellaDD, line 53:
```suggestion

// Build every inspected-page storage operation from the same read/normalize prelude. Keeping this
// as one command builder prevents the read and mutation paths from interpreting malformed storage
// differently as the DatadogDevtools contract evolves.
function buildFlagStorageCommand(command: FlagStorageCommand): string {
  let mutation = ''
  switch (command.type) {
    case 'read':
      break
    case 'write':
      mutation = `overrides[${JSON.stringify(command.key)}] = ${JSON.stringify(command.override)}`
      break
    case 'delete':
      mutation = `delete overrides[${JSON.stringify(command.key)}]`
      break
    case 'clear':
      mutation = 'overrides = {}'
      break
  }

  const persist =
    command.type === 'read' ? '' : `localStorage.setItem(${JSON.stringify(OVERRIDES_KEY)}, JSON.stringify(overrides))`
  const readMarker =
    command.type === 'read' ? `localStorage.getItem(${JSON.stringify(DEVTOOLS_MARKER_KEY)}) === 'enabled'` : 'false'
  return `
    let overrides = {}
    try {
      const parsed = JSON.parse(localStorage.getItem(${JSON.stringify(OVERRIDES_KEY)}) || '{}')
      if (parsed && typeof parsed === 'object' && !Array.isArray(parsed)) {
        overrides = parsed
      }
    } catch (e) {}
    ${mutation}
    ${persist}
    const devtoolsEnabled = ${readMarker}
    return { overrides, devtoolsEnabled }
  `
}

function evaluateFlagStorage(command: FlagStorageCommand): Promise<RawFlagState> {
  return evalInWindow(buildFlagStorageCommand(command)) as Promise<RawFlagState>
}
```
I thought something like this.
`mutateFlagStorage({ type: 'write', key, override })`

developer-extension/src/panel/components/tabs/flagsTab/flagsContext.tsx
```diff
@@ -0,0 +1,419 @@
...
+    status: state.status,
+    overrides: state.overrides,
+    devtoolsEnabled: state.devtoolsEnabled,
+    error: state.error,
+    setOverride,
+    clearOverride,
+    clearAll,
+    reloadPage,
+  }
+}
```
BeltranBulbarellaDD: I know it goes a little bit against what I said before but maybe splitting this file? `useInspectedPageOverride` for everything above this and we keep the `flagsContext` with everything we have below.

BeltranBulbarellaDD (commented):
Let me know what you think! I'm trying to think ways of reuse/simplify the code a little bit and split it.

developer-extension/src/panel/components/tabs/flagsTab/inspectedPageFlags.ts
kellyw1806: Sounds good, just quick note: i pulled the duplicated read/normalize prelude into one constant, so the storage contract lives in one place. Kept them as two small functions instead of one builder that switches on a command type because i think that might add more branching

BeltranBulbarellaDD (approved):
LGTM! As long as codex is happy 😄

PR #4913, opened 2026-07-24
✨ [FFL-2597] Feature Flags tab — OAuth sign-in (stacked PR 1 of 3)

kellyw1806:
## Motivation

Part of the Feature Flags devtools integration (**FFL-2591**), split into a small stacked series for focused review. This PR is the **OAuth foundation**: sign in to Datadog. Browsing the catalog (#4916) and overriding flags (#4912) build on top.

## Stack (review bottom-up)

| | PR | Base | What it adds |
|---|---|---|---|
| **1 of 3** | **#4913 (this)** · FFL-2597 | `main` | OAuth sign-in |
| 2 of 3 | #4916 · FFL-2858 | `ffl-2597` | catalog browsing |
| 3 of 3 | #4912 · FFL-2596 | `ffl-2858` | flag overrides |

## Changes
- **OAuth** (`authorization_code` + **PKCE**, public client) against Datadog's first-party OAuth server; tokens stored in `chrome.storage.session` (cleared on session end, never persisted).
- Feature Flags tab wired into the DevTools panel; connect screen + connection header; persisted `flagsSite` setting; `identity` manifest permission.
- When connected, the tab shows the connection state; the catalog UI is added in #4916.

## Demo
This is what all three stacked PRs look like working together: 

https://github.com/user-attachments/assets/d540ac8c-9f48-48f4-be42-0b3de3fbef76

## Checklist
- [x] Tested locally
- [x] Tested on staging
- [x] Added unit tests for this change (`oauth.spec.ts` — `getFlagsApiHost`).
- [ ] Added e2e/integration tests for this change.
- [ ] Updated documentation and/or relevant AGENTS.md file

PR #4916, opened 2026-07-27
✨ [FFL-2858] Feature Flags tab — catalog browsing (stacked PR 2 of 3)

kellyw1806:
## Motivation

Stacked on **#4913** (OAuth, now merged). With sign-in in place, this PR lets users **browse their feature-flag catalog** from the extension. Overriding flags is the next PR (#4912).

## Stack (review bottom-up)

| | PR | Base | What it adds |
|---|---|---|---|
| 1 of 3 | #4913 · FFL-2597 | `main` | OAuth sign-in — ✅ merged |
| **2 of 3** | **#4916 (this)** · FFL-2858 | `main` | catalog browsing |
| 3 of 3 | #4912 · FFL-2596 | `ffl-2858` | flag overrides |

## Changes
- Browses the catalog via the OAuth-capable `GET /api/ui/ffe/feature-flags` endpoint, **server-side**: the extension fetches only the page it's showing and lets the endpoint apply **search / type / tag filtering + pagination** (`meta.page.total`) — it never loads the whole catalog.
- **Search** (debounced), **Type** filter (fixed set), and **Tags** as free-text with progressive autocomplete.
- OAuth token handling — `getValidAccessToken` refreshes transparently, with **single-flight refresh** so concurrent requests can't spend the rotating refresh token twice.
- **Privacy:** a RUM `beforeSend` strips customer data (flag names/keys/tags carried in query params) from this endpoint's resource events in the extension's own telemetry.
- Defensive variant-value parsing + dedupe-by-key.

## Recently changed
- **Pivoted to server-side pagination + filtering** (previously fetched the *entire* catalog and filtered client-side). Per review: the extension re-fetches on every panel open and the endpoint caps pages at 50, so loading ~2k flags meant many sequential requests. It now fetches one page per view.
- **Design decisions worth calling out:**
  - **Tags filter is free-text** (with suggestions accumulated from loaded pages) rather than a full dropdown — there's no tags/facets endpoint to enumerate all tags once we stopped loading the whole catalog. Type stays a real dropdown (fixed set).
  - **Search debounced at ~400ms** — responsive while avoiding a request per keystroke.
- **Addressed the Codex review:** single-flight token refresh (prevents a spurious disconnect when concurrent refreshes race near expiry), a RUM query-string scrub for the FFE endpoint, page-clamp when the catalog shrinks, remount on site change (avoids stale filters / another org's tag suggestions leaking in), and a raw-string fallback for an unknown `value_type`.

## Demo
This is what the three PRs look like working together:

https://github.com/user-attachments/assets/6ebece86-8c38-4bad-9366-9bf2dd643ca5

## Checklist
- [x] Tested locally
- [x] Tested on staging
- [x] Added unit tests for this change (`oauth.spec.ts` — `getValidAccessToken` incl. refresh/skew/coalescing; `flagCatalog.spec.ts`).

### DIFF
--- a/developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.spec.ts
+++ b/developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.spec.ts
@@ -1,0 +1,202 @@
+import type { CatalogFlag, FlagCatalogRequest } from './flagCatalog'
+import { fetchFlagCatalog } from './flagCatalog'
+
+describe('flagCatalog', () => {
+  describe('fetchFlagCatalog', () => {
+    const baseRequest: FlagCatalogRequest = { page: 1, pageSize: 20, search: '', typeFilter: [], tagFilter: [] }
+
+    function mockResponse(body: unknown) {
+      return spyOn(globalThis, 'fetch').and.returnValue(Promise.resolve(new Response(JSON.stringify(body))))
+    }
+
+    it('requests one page server-side (active-only, offset from page) and returns the server total', async () => {
+      const sampleFlag: CatalogFlag = {
+        key: 'flag-a',
+        name: 'Flag A',
+        type: 'BOOLEAN',
+        variants: [{ name: 'on', value: true }],
+        tags: ['x'],
+      }
+      const sampleTotal = 42
+      const spy = mockResponse({
+        data: [
+          {
+            attributes: {
+              key: sampleFlag.key,
+              name: sampleFlag.name,
+              value_type: sampleFlag.type,
+              // The API returns variant values as strings; parseVariantValue turns them back.
+              variants: sampleFlag.variants.map(({ name, value }) => ({ name, value: String(value) })),
+              tags: sampleFlag.tags,
+            },
+          },
+        ],
+        meta: { page: { total: sampleTotal } },
+      })
+
+      const page = await fetchFlagCatalog('tok', 'datad0g.com', { ...baseRequest, page: 3, pageSize: 20 })
+
+      const [requestUrl, requestInit] = spy.calls.argsFor(0) as [string, RequestInit]
+      const url = new URL(requestUrl)
+      expect(url.pathname).toBe('/api/ui/ffe/feature-flags')
+      expect(url.searchParams.get('page[limit]')).toBe('20')
+      expect(url.searchParams.get('page[offset]')).toBe('40') // (3 - 1) * 20
+      expect(url.searchParams.get('is_archived')).toBe('false')
+      expect((requestInit.headers as Record<string, string>).Authorization).toBe('Bearer tok')
+      expect(page.total).toBe(sampleTotal)
+      expect(page.flags).toEqual([sampleFlag])
+    })
+
+    it('sends search, value_type (repeated), and tags (repeated) as server-side filters', async () => {
+      const spy = mockResponse({ data: [], meta: { page: { total: 0 } } })
+
+      await fetchFlagCatalog('tok', 'datad0g.com', {
+        page: 1,
+        pageSize: 20,
+        search: 'checkout',
+        typeFilter: ['BOOLEAN', 'STRING'],
+        tagFilter: ['team:x', 'beta'],
+      })
+
+      const [requestUrl] = spy.calls.argsFor(0) as [string, RequestInit]
+      const url = new URL(requestUrl)
+      expect(url.searchParams.get('search')).toBe('checkout')
+      expect(url.searchParams.getAll('value_type')).toEqual(['BOOLEAN', 'STRING'])
+      expect(url.searchParams.getAll('tags')).toEqual(['team:x', 'beta'])
+    })
+
+    it('omits the search param when the term is empty', async () => {
+      const spy = mockResponse({ data: [], meta: { page: { total: 0 } } })
+      await fetchFlagCatalog('tok', 'datad0g.com', baseRequest)
+      const [requestUrl] = spy.calls.argsFor(0) as [string, RequestInit]
+      expect(new URL(requestUrl).searchParams.has('search')).toBe(false)
+    })
+
+    it('parses variant values by declared type, falling back to the raw string on bad input', async () => {
+      mockResponse({
+        data: [
+          {
+            attributes: {
+              key: 'f',
+              value_type: 'JSON',
+              variants: [
+                { name: 'ok', value: '{"a":1}' },
+                { name: 'bad', value: 'not json' },
+              ],
+            },
+          },
+          {
+            attributes: {
+              key: 'n',
+              value_type: 'INTEGER',
+              variants: [
+                { name: 'five', value: '5' },
+                { name: 'nan', value: 'abc' },
+              ],
+            },
+          },
+        ],
+        meta: { page: { total: 2 } },
+      })
+
+      const { flags } = await fetchFlagCatalog('tok', 'datad0g.com', baseRequest)
+      expect(flags[0].variants[0].value).toEqual({ a: 1 })
+      expect(flags[0].variants[1].value).toBe('not json')
+      expect(flags[1].variants[0].value).toBe(5)
+      expect(flags[1].variants[1].value).toBe('abc')
+    })
+
+    it('keeps malformed or out-of-range variant values raw instead of coercing them', async () => {
+      mockResponse({
+        data: [
+          {
+            attributes: {
+              key: 'b',
+              value_type: 'BOOLEAN',
+              variants: [
+                { name: 't', value: 'true' },
+                { name: 'f', value: 'false' },
+                { name: 'weird', value: 'True' },
+              ],
+            },
+          },
+          {
+            attributes: {
+              key: 'i',
+              value_type: 'INTEGER',
+              variants: [
+                { name: 'partial', value: '5abc' },
+                { name: 'float', value: '5.5' },
+                { name: 'unsafe', value: '9007199254740993' },
+              ],
+            },
+          },
+          {
+            attributes: {
+              key: 'd',
+              value_type: 'NUMERIC',
+              variants: [
+                { name: 'ok', value: '5.5' },
+                { name: 'partial', value: '5abc' },
+                { name: 'empty', value: '' },
+              ],
+            },
+          },
+        ],
+        meta: { page: { total: 3 } },
+      })
+
+      const { flags } = await fetchFlagCatalog('tok', 'datad0g.com', baseRequest)
+      const [booleanFlag, integer, numeric] = flags
+      expect(booleanFlag.variants.map((variant) => variant.value)).toEqual([true, false, 'True'])
+      expect(integer.variants.map((variant) => variant.value)).toEqual(['5abc', '5.5', '9007199254740993'])
+      expect(numeric.variants.map((variant) => variant.value)).toEqual([5.5, '5abc', ''])
+    })
+
+    it('dedupes flags sharing a key within a page, keeping the first occurrence', async () => {
+      mockResponse({
+        data: [
+          { attributes: { key: 'dup', name: 'First', value_type: 'STRING', variants: [], tags: [] } },
+          { attributes: { key: 'dup', name: 'Second', value_type: 'STRING', variants: [], tags: [] } },
+        ],
+        meta: { page: { total: 2 } },
+      })
+
+      const { flags } = await fetchFlagCatalog('tok', 'datad0g.com', baseRequest)
+      expect(flags.length).toBe(1)
+      expect(flags[0].name).toBe('First')
+    })
+
+    it('falls back to the key for a missing name and defaults tags/variants', async () => {
+      mockResponse({ data: [{ attributes: { key: 'no-name', value_type: 'STRING' } }], meta: { page: { total: 1 } } })
+
+      const { flags } = await fetchFlagCatalog('tok', 'datad0g.com', baseRequest)
+      expect(flags[0].name).toBe('no-name')
+      expect(flags[0].tags).toEqual([])
+      expect(flags[0].variants).toEqual([])
+    })
+
+    it('tolerates a response that omits data/meta, falling total back to the page length', async () => {
+      mockResponse({ errors: ['x'] })
+      const page = await fetchFlagCatalog('tok', 'datad0g.com', baseRequest)
+      expect(page.flags).toEqual([])
+      expect(page.total).toBe(0)
+    })
+
+    it('falls back total to the number of returned flags when meta.page.total is missing', async () => {
+      mockResponse({
+        data: [{ attributes: { key: 'a', value_type: 'STRING' } }, { attributes: { key: 'b', value_type: 'STRING' } }],
+      })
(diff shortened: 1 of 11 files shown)


developer-extension/src/panel/hooks/useSdkInfos.ts
```diff
@@ -69,39 +69,31 @@ async function getInfos(): Promise<SdkInfos> {
...
-          document.cookie
-            .split(';')
-            .map(cookie => cookie.match(/(\\S*?)=(.*)/)?.slice(1))
-            .filter(cookie => cookie !== undefined)
-        );
-        const cookieRawValue = cookies.get('_dd_s_v2') ?? cookies.get('_dd_s');
+        const cookieRawValue = document.cookie
+          .split(';')
+          .map(cookie => cookie.match(/(\\S*?)=(.*)/)?.slice(1) || [])
+          .find(([name, _]) => name === '_dd_s')
```
greghuels: issue: This regresses SDK v7 session detection by only reading the legacy `_dd_s` cookie. The current SDK writes `_dd_s_v2`, with `_dd_s` only as fallback, so the Info/Replay flows will lose the active session on current SDK pages. Can we restore the `_dd_s_v2` lookup from the base branch?

developer-extension/src/panel/hooks/useSdkInfos.ts
```diff
@@ -60,7 +60,7 @@ async function getInfos(): Promise<SdkInfos> {
       `
         // Helper to serialize objects while preserving function metadata
         function serializeWithFunctions(obj) {
```
greghuels, line 62: issue: This removes the `JSON.stringify(...) === undefined` guard from the base branch. When an SDK global exists but `getInitConfiguration` is missing or returns `undefined`, this now calls `JSON.parse(undefined)`, the whole `getInfos()` call falls into `catch`, and the Info tab goes blank. Can we restore the guard before parsing?

developer-extension/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/browser-sdk-developer-extension",
-  "version": "7.6.1",
+  "version": "7.5.0",
```
greghuels: issue: This downgrades only the developer extension from `7.6.1` back to `7.5.0`, while the repo and other packages stay on `7.6.1`. This looks like stale-branch churn from before the latest release bump; can we revert this line?

developer-extension/src/panel/components/tabs/flagsTab/oauth.spec.ts
```diff
@@ -16,4 +17,80 @@ describe('oauth', () => {
...
+    it('clears tokens and returns null when expired with no refresh token', async () => {
+      await storeTokens({ accessToken: 'old', expiresAt: Date.now() - 1 })
+
+      expect(await getValidAccessToken('datad0g.com')).toBeNull()
+      expect(await loadStoredTokens()).toBeNull()
+    })
+
+    afterEach(async () => {
+      await clearStoredTokens()
+    })
```
greghuels: nit: it looks like this repo prefers using "registerCleanupTask" instread of `afterEach`

BeltranBulbarellaDD:
@kellyw1806 In my opinion I would keep the stacked PR 2 and 3 in draft until the first one is approved. That way after changes are requested in the first one, this one remains clean for the reviewer.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export interface CatalogVariant {
+  name: string
+  value: CatalogVariantValue
+}
+
+export interface CatalogFlag {
+  key: string
+  // Human-friendly display name; falls back to the key when the API doesn't provide one.
+  name: string
```
BeltranBulbarellaDD:
```suggestion
  name: string
```

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+interface RawFeatureFlagsResponse {
+  data: RawFeatureFlagResource[]
+}
+
+// Variant values come back from the API as strings regardless of the flag's declared type. Falls
+// back to the raw string on unparseable input so one malformed variant can't blow up the mapping
+// of the entire catalog.
+function parseVariantValue(type: FlagOverrideType, rawValue: string): CatalogVariantValue {
```
BeltranBulbarellaDD:
```suggestion
import type { FlagType } from './flagTypeConstants'
import { getFlagsApiHost } from './oauth'

export interface CatalogFlag {
  key: string
  // Human-friendly display name; falls back to the key when the API doesn't provide one.
  name: string
  type: FlagType
  variants: Array<{
    name: string
    value: unknown
  }>
  tags: string[]
}

interface RawFeatureFlag {
  attributes: {
    key: string
    name?: string
    value_type: FlagType
    variants?: Array<{
      name: string
      value: string
    }>
    tags?: string[]
  }
}

// Variant values come back from the API as strings regardless of the flag's declared type. Falls
// back to the raw string on unparseable input so one malformed variant can't blow up the mapping
// of the entire catalog.
function parseVariantValue(type: FlagType, rawValue: string): unknown {
```
Since atm we are exporting but no using the other types, how would you feel of something like this? I think it simplifies a little bit the code.

developer-extension/src/panel/components/tabs/flagsTab/flagTypeConstants.ts
```diff
@@ -0,0 +1,17 @@
...
+export const FLAG_TYPES: FlagOverrideType[] = ['BOOLEAN', 'STRING', 'INTEGER', 'NUMERIC', 'JSON']
+
+// Display labels matching the webapp's Type filter (NUMERIC shows as "Number").
+export const TYPE_LABELS: Record<FlagOverrideType, string> = {
+  BOOLEAN: 'Boolean',
+  STRING: 'String',
+  INTEGER: 'Integer',
+  NUMERIC: 'Number',
+  JSON: 'JSON',
+}
```
BeltranBulbarellaDD:
```suggestion
export const FLAG_TYPES = ['BOOLEAN', 'STRING', 'INTEGER', 'NUMERIC', 'JSON'] as const

export type FlagType = (typeof FLAG_TYPES)[number]

// Display labels matching the webapp's Type filter (NUMERIC shows as "Number").
export const TYPE_LABELS = {
  BOOLEAN: 'Boolean',
  STRING: 'String',
  INTEGER: 'Integer',
  NUMERIC: 'Number',
  JSON: 'JSON',
} satisfies Record<FlagType, string>
```
Same thought here.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.ts
```diff
@@ -0,0 +1,148 @@
...
+  for (let page = 0; page < MAX_PAGES; page++) {
+    const url = new URL(`https://${host}/api/ui/ffe/feature-flags`)
+    url.searchParams.set('limit', String(PAGE_LIMIT))
+    url.searchParams.set('offset', String(offset))
+
+    const response = await fetch(url.toString(), {
+      headers: {
+        Authorization: `Bearer ${token}`,
+      },
+    })
```
BeltranBulbarellaDD:
```suggestion
  const url = new URL(`https://${host}/api/ui/ffe/feature-flags`)
  url.searchParams.set('page[limit]', String(request.pageSize))
  url.searchParams.set('page[offset]', String((request.page - 1) * request.pageSize))

  if (request.search) {
    url.searchParams.set('search', request.search)
  }
  for (const type of request.typeFilter) {
    url.searchParams.append('value_type', type)
  }
  for (const tag of request.tagFilter) {
    url.searchParams.append('tags', tag)
  }

  const response = await fetch(url.toString(), {
    headers: {
      Authorization: `Bearer ${token}`,
    },
  })
```
Instead of a `for loop` what do you think of this. 
Because reading the API docs: "Pagination is always applied: the response includes meta.page (with total) and pagination links." [Link](https://dd-api-docs.us1.staging.dog/internal/dd-api-docs). 
So no need to do the loop to get the total we can request the 20 we will show and offset on page change. I think this way it loads faster and does not get unnecessary info.

BeltranBulbarellaDD: And then we can really simplify the `useFlagCatalogView`

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+    [flags]
+  )
+
+  return (
+    <Stack gap="xs">
+      <TextInput
+        placeholder="Filter your feature flags"
+        leftSection={<IconSearch size={14} />}
+        value={view.search}
+        onChange={(event) => view.setSearch(event.currentTarget.value)}
```
BeltranBulbarellaDD, line 20: I think we could use a debounce here, like 300 ms so we don't run a search as soon as we write a character.

BeltranBulbarellaDD (commented):
Let me know what you think! I think these comments will make the flags load almost instantly.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.ts
BeltranBulbarellaDD: This was before the request suggestion below so you will have to extend my suggested types a little bit and add another type for the request probably.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.ts
kellyw1806: I think this might not work because it clashes with a couple of features the later PRs in this stack add, which both need the whole flag list loaded at once. For example, for overrides, you can override a flag's value locally, and we bump the ones you've overridden to the top so they're easy to find and the server can't do that (because it doesnt know which flags you overrode since they are stored locally in browser) A couple of the filters also break like the Tag and My Teams filter. So thinking maybe we keep the list loading client-side for now? Might also discuss this more with my team to see if they want to analyze this tradeoff

BeltranBulbarellaDD: Good point about locally overridden flags, but I don’t think caching every fetched flag in local storage solves it. Also I believe the endpoint supports search, type, regular-tag, and team-tag filtering server-side, so those filters shouldn’t require the complete catalog. 
As for the following PRs in my opinion is better to tackle it when it comes. Else it's hard to keep track of both at the same time. 

I'm thinking, what about having 2 parts?. The locally overriden flags in a `set()` above, when we open the extension we do a search for each to see if they match the server. And then below we have the rest. Where each time we do a pagination/search/filter we first check the  results agains the `set()` so we don't show them twice. 
This is because I am not confortable with fetching 2k feature flags on load each time. 

Screenshot 2026-07-31 at 10 08 32

BeltranBulbarellaDD (approved):
LGTM!

developer-extension/src/panel/hooks/flagOverrideTypes.ts
```diff
@@ -0,0 +1,4 @@
+// The value type of a feature flag (and of an override for it), for the hooks layer. Intentionally
+// a separate copy of flagsTab's `FlagType`: the hooks layer must not import from the flagsTab
+// directory, so each layer declares its own copy of this small union.
+export type FlagOverrideType = 'BOOLEAN' | 'STRING' | 'INTEGER' | 'NUMERIC' | 'JSON'
```
bdibon: **🥜 nitpick:** ‏It's a bit weird to add it now when it's not used.

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
```diff
@@ -190,3 +223,57 @@ export async function clearStoredTokens(): Promise<void> {
...
+  // skew would discard the last 60s of a still-valid token and force an unnecessary reconnect.
+  const skew = tokens.refreshToken ? EXPIRY_SKEW_MS : 0
+  if (Date.now() < tokens.expiresAt - skew) {
+    return tokens.accessToken
+  }
+  if (tokens.refreshToken) {
+    try {
+      // Coalesce concurrent refreshes into a single request (see pendingRefresh). Capture the shared
+      // promise in a local so the `finally` clearing pendingRefresh can't race the await below.
+      const refresh = (pendingRefresh ??= refreshTokens(site, tokens.refreshToken)
```
bdibon, line 253: **👏 praise:** ‏Didn't know about that one `??=`

developer-extension/src/panel/components/tabs/flagsTab/flagCatalog.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+      expect(url.pathname).toBe('/api/ui/ffe/feature-flags')
+      expect(url.searchParams.get('page[limit]')).toBe('20')
+      expect(url.searchParams.get('page[offset]')).toBe('40') // (3 - 1) * 20
+      expect(url.searchParams.get('is_archived')).toBe('false')
+      expect((requestInit.headers as Record<string, string>).Authorization).toBe('Bearer tok')
+      expect(page.total).toBe(42)
+      expect(page.flags).toEqual([
+        { key: 'flag-a', name: 'Flag A', type: 'BOOLEAN', variants: [{ name: 'on', value: true }], tags: ['x'] },
+      ])
+    })
```
bdibon, line 48: **💬 suggestion:** ‏I find it more readable when the data used in arrange and assert phases refer to a common label. This can be applied to other test cases.


```suggestion
     it('requests one page server-side (active-only, offset from page) and returns the server total', async () => {
      const sampleFlag: CatalogFlag = {
        key: 'flag-a',
        name: 'Flag A',
        type: 'BOOLEAN',
        variants: [{ name: 'on', value: true }],
        tags: ['x'],
      }
      const sampleTotal = 42

      const spy = mockResponse({
        data: [
          {
            attributes: {
              key: sampleFlag.key,
              name: sampleFlag.name,
              value_type: sampleFlag.type,
              variants: sampleFlag.variants.map(({ name, value }) => ({ name, value: String(value) })),
              tags: sampleFlag.tags,
            },
          },
        ],
        meta: { page: { total: sampleTotal } },
      })

      const page = await fetchFlagCatalog('tok', 'datad0g.com', { ...baseRequest, page: 3, pageSize: 20 })

      const [requestUrl, requestInit] = spy.calls.argsFor(0) as [string, RequestInit]
      const url = new URL(requestUrl)
      expect(url.pathname).toBe('/api/ui/ffe/feature-flags')
      expect(url.searchParams.get('page[limit]')).toBe('20')
      expect(url.searchParams.get('page[offset]')).toBe('40') // (3 - 1) * 20
      expect(url.searchParams.get('is_archived')).toBe('false')
      expect((requestInit.headers as Record<string, string>).Authorization).toBe('Bearer tok')
      expect(page.total).toBe(sampleTotal)
      expect(page.flags).toEqual([sampleFlag])
    })
```

bdibon (approved):
Good job!

PR #4920, opened 2026-07-27
✨ [RUM-16985] Capture wasm module build_ids and enrich error events

ImaneLargou:
## Motivation

Browser WebAssembly runtime errors currently lack the module metadata required for WASM symbolication.

This change identifies errors containing WebAssembly stack frames and reports the loaded WASM module URL and build ID. This allows the error-processing pipeline to distinguish ordinary browser errors from errors that should use WASM/DWARF symbolication.

Related upstream schema change: https://github.com/DataDog/rum-events-format/pull/427

## Changes

- Track WebAssembly modules loaded through:
   - `WebAssembly.instantiate`
   - `WebAssembly.instantiateStreaming`
   - `WebAssembly.compile`
   - `WebAssembly.compileStreaming`
- Extract module build IDs from the `build_id` custom section, with `external_debug_info` as a fallback.
- Detect WASM errors by checking the complete error stack for WASM-specific frames.
- Report the following metadata on WASM errors:
   - `error.source_type: "browser+wasm"`
   - `error.wasm_modules`, containing the module URL and build ID
- Add support to both RUM and Browser Logs.
- Keep ordinary JavaScript errors unchanged. In particular, `.js` frames do not qualify an error as WASM.
- Update the generated RUM event types from the corresponding `rum-events-format` schema change.

## Test instructions

Run the relevant unit tests:

- packages/browser-core/src/domain/wasmModules/wasmBinaryParser.spec.ts
- packages/browser-core/src/domain/wasmModules/wasmModuleTracking.spec.ts
- packages/browser-rum-core/src/domain/error/errorCollection.spec.ts
- packages/browser-logs/src/domain/createErrorFieldFromRawError.spec.ts

Run the E2E scenario:
```bash
yarn test:e2e -g "send WebAssembly runtime errors with module metadata"
```
The E2E test loads a real .wasm module using `WebAssembly.instantiateStreaming`, triggers a runtime error, and verifies that both RUM and Logs events contain:
```json
  {
    "source_type": "browser+wasm",
    "wasm_modules": [
      {
        "url": "<test module URL>",
        "build_id": "abcd"
      }
    ]
  }
```
## Checklist

- [x] Tested locally
- [x] Tested on staging


ImaneLargou:

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
```diff
@@ -0,0 +1,182 @@
...
+function recordModuleFromView(url: string, view: ArrayBufferView): void {
+  recordModule(url, view.buffer.slice(view.byteOffset, view.byteOffset + view.byteLength))
+}
+
+// Extracts build_id from a Response clone without consuming it for the caller.
+// Streaming compilation and metadata extraction happen in parallel, but the
+// wrapper only resolves once both are done. This guarantees that an error
+// thrown immediately by an exported function can reference the loaded module.
+function captureFromResponse(response: Response): Promise<void> {
+  const url = response.url || '<wasm-instantiate-streaming-no-url>'
```
mormubis: **❓ question:** ‏How could the URL be empty? Is this just because of the types?

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
```diff
@@ -0,0 +1,182 @@
...
+      return
+    }
+    stopped = true
+    trackingClients -= 1
+    if (trackingClients === 0) {
+      stopTracking?.()
+      stopTracking = undefined
+      registry.clear()
+    }
+  }
```
mormubis: **❓ question:** ‏How many clients could we expect tracking WebAssembly?

packages/browser-logs/src/entries/main.ts
```diff
@@ -7,10 +7,13 @@
  * @see [Browser Log Collection](https://docs.datadoghq.com/logs/log_collection/javascript/)
  */
 
-import { defineGlobal, globalObject } from '@datadog/browser-core'
+import { defineGlobal, globalObject, startWasmModuleTracking } from '@datadog/browser-core'
 import type { LogsPublicApi } from '../boot/logsPublicApi'
 import { makeLogsPublicApi } from '../boot/logsPublicApi'
 
+// Install WebAssembly hooks before deferred Logs initialization so eagerly loaded modules are captured.
+startWasmModuleTracking()
```
mormubis: **🔨 warning:** ‏The philosophy of the SDK is not to override any APIs unless we are allowed to track the user.

This should be deferred until SDK initialization.

packages/browser-rum-core/src/boot/startRum.ts
```diff
@@ -220,6 +221,13 @@ export function startRumEventCollection(
   const { stop: stopLongTaskCollection } = startLongTaskCollection(lifeCycle, configuration)
   cleanupTasks.push(stopLongTaskCollection)
 
+  // Intercept WebAssembly module loads to capture build_id. errorCollection
+  // reads it to set source_type='browser+wasm' and error.wasm_modules.
+  // Must start before any wasm load — RUM is initialised before the page's
+  // wasm fetch in typical setups.
+  const stopWasmModuleTracking = startWasmModuleTracking()
```
mormubis: **❓ question:** ‏Are we not initializing this in different places?

packages/browser-rum/src/entries/main.ts
```diff
@@ -7,13 +7,20 @@
...
 import { makeRecorderApi } from '../boot/recorderApi'
 import { createDeflateEncoder, startDeflateWorker } from '../domain/deflate'
 import { makeProfilerApi } from '../boot/profilerApi'
 
+// Install WebAssembly hooks synchronously at SDK script-load time. DD_RUM.init()
+// defers its startRum work to a microtask, which races against eagerly-loaded
+// wasm modules — by the time the deferred wrap installs, instantiateStreaming
+// may have already been called. Installing here (before any deferral) closes
+// that race.
+startWasmModuleTracking()
```
mormubis: **🔨 warning:** ‏Mentioned before.

packages/browser-rum-slim/src/entries/main.ts
```diff
@@ -1,10 +1,13 @@
 // Keep the following in sync with packages/browser-rum/src/entries/main.ts
-import { defineGlobal, globalObject } from '@datadog/browser-core'
+import { defineGlobal, globalObject, startWasmModuleTracking } from '@datadog/browser-core'
 import type { RumPublicApi } from '@datadog/browser-rum-core'
 import { makeRumPublicApi } from '@datadog/browser-rum-core'
 import { makeRecorderApiStub } from '../boot/stubRecorderApi'
 import { makeProfilerApiStub } from '../boot/stubProfilerApi'
 
+// Install WebAssembly hooks before deferred RUM initialization so eagerly loaded modules are captured.
+startWasmModuleTracking()
```
mormubis: **🔨 warning:** ‏Keeping track of all the places where this is automatically tracked.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: Once the module calls are removed, we expect at most two clients: RUM and Logs.

packages/browser-rum-core/src/boot/startRum.ts
ImaneLargou: Yes, this was part of my first tests at wasm module tracking but modules loaded before startRum runs were missed. Forgot to remove it 😅. I'll delete it.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: It can be empty when the response is created manually with `new Response(wasmBytes)` instead of coming from `fetch()`. This is probably uncommon, but the fallback handles that edge case. I can remove the fallback if we don’t want to send a synthetic URL.

packages/browser-logs/src/boot/preStartLogs.ts
```diff
@@ -113,21 +116,28 @@ export function createPreStartStrategy(
       trackingConsentState.onGrantedOnce(() => {
         startTrackingConsentContext(hooks, trackingConsentState)
         mockable(startTelemetry)(TelemetryService.LOGS, configuration, hooks.assembleTelemetry, sdkName)
+        stopWasmModuleTracking = mockable(startWasmModuleTracking)()
```
BenoitZugmeyer: **question:** why do we start module tracking here?

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
BenoitZugmeyer: **suggestion:** how much is it important to stop the tracking (outside of tests)? I think things might be simpler without this reference counting.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
```diff
@@ -0,0 +1,182 @@
...
+function installWasmModuleTracking(): () => void {
+  const origInstantiate = WebAssembly.instantiate
+  const origCompile = WebAssembly.compile
+  const origInstantiateStreaming = WebAssembly.instantiateStreaming
+  const origCompileStreaming = WebAssembly.compileStreaming
+
+  // Hook 1: instantiate(bytes | module, imports). For raw bytes, we can read
+  // build_id directly; for an already-compiled WebAssembly.Module we have no
+  // URL or bytes to inspect — register a placeholder.
+  WebAssembly.instantiate = function (this: typeof WebAssembly, source: any, importObject?: any) {
```
BenoitZugmeyer: **suggestion:** use `instrumentMethod` to instrument all methods.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
```diff
@@ -0,0 +1,182 @@
...
+
+function recordModuleFromView(url: string, view: ArrayBufferView): void {
+  recordModule(url, view.buffer.slice(view.byteOffset, view.byteOffset + view.byteLength))
+}
+
+// Extracts build_id from a Response clone without consuming it for the caller.
+// Streaming compilation and metadata extraction happen in parallel, but the
+// wrapper only resolves once both are done. This guarantees that an error
+// thrown immediately by an exported function can reference the loaded module.
+function captureFromResponse(response: Response): Promise<void> {
```
BenoitZugmeyer: **nitpick:** this might be better written as an `async` function

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
```diff
@@ -0,0 +1,182 @@
...
+        )
+      })
+    }
+  }
+
+  if (origCompileStreaming) {
+    WebAssembly.compileStreaming = function (source) {
+      return Promise.resolve(source).then((response: Response) => {
+        const capturePromise = captureFromResponse(response)
+        return Promise.all([origCompileStreaming.call(this, response), capturePromise]).then(([module]) => module)
```
BenoitZugmeyer: **issue:** using `instrumentMethod` here would make sure all received argument are passed to the original `compileStreaming` method (mdn [shows](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/compileStreaming_static) that it supports a `compileOptions` object)

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
```diff
@@ -0,0 +1,182 @@
...
+// thrown immediately by an exported function can reference the loaded module.
+function captureFromResponse(response: Response): Promise<void> {
+  const url = response.url || '<wasm-instantiate-streaming-no-url>'
+  if (registry.has(url)) {
+    return Promise.resolve()
+  }
+  try {
+    return response
+      .clone()
+      .arrayBuffer()
```
BenoitZugmeyer: **issue:** you are downloading the whole response in memory -- that will have an impact on large wasm files. Would it be possible to read only a limited amount of bytes instead?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -287,6 +287,36 @@ test.describe('logs', () => {
...
+    .run(async ({ baseUrl, intakeRegistry, flushEvents, page, withBrowserLogs }) => {
+      test.skip(
+        test.info().project.name === 'webkit' || test.info().project.name === 'chromium-pinned',
+        'These browser versions do not expose uncaught WebAssembly traps through the runtime error event'
+      )
+
+      await page.evaluate(async () => {
+        const { instance } = await WebAssembly.instantiateStreaming(fetch('/test-module.wasm'))
+
+        setTimeout(() => (instance.exports.run as () => void)())
```
BenoitZugmeyer, line 312: **question:** why do we need this `setTimeout`?

test/e2e/scenario/logs.scenario.ts
```diff
@@ -287,6 +287,36 @@ test.describe('logs', () => {
       })
     })
 
+  createTest('send WebAssembly runtime errors with module metadata')
+    .withRum()
```
BenoitZugmeyer: **suggestion:** don't add a rum test in logs.scenario. Move the test in a common scenario file, or add split the test (one for rum, one for logs)

packages/browser-logs/src/boot/preStartLogs.ts
ImaneLargou: It starts after consent but before the asynchronous session-manager startup to avoid missing modules loaded during that wait. When I was testing with rum, it seemed like modules were missing if we start later than this.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: Removing it.

test/e2e/scenario/logs.scenario.ts
ImaneLargou: The setTimeout makes the WASM trap occur as an uncaught browser error. If we call run() directly inside page.evaluate(), the exception is reported as an evaluation failure instead.

test/e2e/scenario/logs.scenario.ts
ImaneLargou: Tests are split into one for logs and one for rum now.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: Changed it to use instrumentMethod!

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: Changed to `instrumentMethod`

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: This function has been removed. We now extract the build ID directly from the compiled WebAssembly.Module, so we no longer clone or read the response body.

packages/browser-core/src/domain/wasmModules/wasmModuleTracking.ts
ImaneLargou: Updated the PR so now we extract the build ID directly from the compiled WebAssembly.Module, and we no longer clone or read the response body.

packages/browser-rum-core/src/domain/assembly.spec.ts
```diff
@@ -768,6 +768,41 @@ describe('rum assembly', () => {
       })
     })
   })
+
+  describe('assemble hook error override', () => {
```
BenoitZugmeyer: **suggestion:** I don't think this test is necessary anymore

packages/browser-logs/src/domain/logger.ts
```diff
@@ -78,10 +80,11 @@ export class Logger {
         })
 
         debugIds = rawError.debugIds
+        errorField = createErrorFieldFromRawError(rawError, { includeMessage: true })
 
         context = combine(
           {
-            error: createErrorFieldFromRawError(rawError, { includeMessage: true }),
+            error: errorField,
```
BenoitZugmeyer: **nitpick:** this change is not really useful anymore

packages/browser-plugin-wasm/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+ * `DD_LOGS.init({ plugins: [makeWasmPlugin()] })`, it intercepts WebAssembly module creation
+ * to record each module's URL and build ID, and enriches error events whose stack trace
+ * contains a WebAssembly frame with `source_type: 'browser+wasm'` and the list of loaded
+ * `wasm_modules`, so they can be symbolicated against the matching debug symbols.
+ *
+ * The WebAssembly hooks are installed once and stay active for the lifetime of the page.
+ *
+ * @experimental
+ */
+export function makeWasmPlugin(): WasmPlugin {
```
BenoitZugmeyer: **suggestion:** plugins don't usually have a `make` prefix. `wasmPlugin()` should be fine.

BenoitZugmeyer (approved):

mormubis (approved):

PR #4932, opened 2026-08-06
✨ [FFL-2857] Feature Flags tab — team filtering + token revocation (stacked PR 4 of 4)

kellyw1806:
## Motivation

Stacked on **#4912** (flag overrides). Adds team/creator filtering + per-flag descriptions to the Feature Flags tab, plus production OAuth support and token revocation on disconnect.

## Stack (review bottom-up)

| | PR | Base | What it adds | Status |
|---|---|---|---|---|
| 1 of 4 | #4913 · FFL-2597 | `main` | OAuth sign-in | ✅ merged |
| 2 of 4 | #4916 · FFL-2858 | `ffl-2597` | catalog browsing | ✅ merged |
| 3 of 4 | #4912 · FFL-2596 | `main` | flag overrides | in review |
| **4 of 4** | **this** · FFL-2857 | `ffl-2596` | team filter + token revocation | draft |

> ⚠️ Base is `kelly.wang/ffl-2596` (#4912) — merge that first. Kept as a **draft** until #4912 lands; this then retargets to `main` and gets a quick rebase.

## Changes
- **Team & creator filters (server-side):** "My feature flags" → `created_by=<uuid>`; "My teams" → `tags=team:<handle>`. Identity (user UUID + team handles) comes from `/api/v2/current_user` + `/api/v2/team?filter[me]=true`. "My teams" is a checkbox dropdown showing an "N selected" summary.
- **Per-flag description** in each row (catalog `description` attribute), with a "Show more" expander that only appears when the text is actually clamped.
- **Production OAuth:** per-site client id (`getClientId` — staging vs prod); site picker trimmed to **US1 + Staging**.
- **Token revocation on disconnect** (RFC 7009 `/oauth2/v1/revoke` — revokes the refresh token, clears local tokens either way).
- **Scope-retry:** falls back to the required scopes if the OAuth client isn't granted the optional `teams_read` (via `invalid_scope` redirect or an authorize page-load failure).

## DEMO 
Note, we can now choose between datadoghq.com and staging. There are also filters for My Feature Flags and My Teams. Descriptions are fetched as well. 

https://github.com/user-attachments/assets/4f7e302c-4997-455c-ae65-a79c42a40e63

## Checklist
- [x] Tested locally
- [x] Tested on staging

### DIFF
--- a/developer-extension/src/panel/components/panel.tsx
+++ b/developer-extension/src/panel/components/panel.tsx
@@ -54,11 +54,9 @@ export function Panel() {
           <Tabs.Tab value={PanelTabs.Replay}>
             <Text>Live replay</Text>
           </Tabs.Tab>
-          {settings.datadogMode && (
-            <Tabs.Tab value={PanelTabs.Flags}>
-              <Text>Feature Flags</Text>
-            </Tabs.Tab>
-          )}
+          <Tabs.Tab value={PanelTabs.Flags}>
+            <Text>Feature Flags</Text>
+          </Tabs.Tab>
           <Tabs.Tab
             value={PanelTabs.Settings}
             rightSection={
@@ -98,11 +96,9 @@ export function Panel() {
       <Tabs.Panel value={PanelTabs.Replay} className={classes.tab}>
         <ReplayTab />
       </Tabs.Panel>
-      {settings.datadogMode && (
-        <Tabs.Panel value={PanelTabs.Flags} className={classes.tab}>
-          <FlagsTab />
-        </Tabs.Panel>
-      )}
+      <Tabs.Panel value={PanelTabs.Flags} className={classes.tab}>
+        <FlagsTab />
+      </Tabs.Panel>
       <Tabs.Panel value={PanelTabs.Settings} className={classes.tab}>
         <SettingsTab />
       </Tabs.Panel>
--- a/developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
+++ b/developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
@@ -54,9 +54,8 @@ export function FlagCatalogBody() {
 }
 
 /**
- * The always-visible "Local overrides" section shown above the paginated catalog. Lists every
- * overridden flag (regardless of which catalog page it's on), so overrides are never buried by
- * pagination. The overridden flags' catalog data comes from the context (see useOverriddenFlags).
+ * Pinned above the paginated catalog, listing every overridden flag regardless of which page it's
+ * on, so overrides are never buried by pagination.
  */
 export function OverridesSection() {
   const { overriddenFlags } = useFlagsContext()
@@ -75,9 +74,10 @@ export function OverridesSection() {
   )
 }
 
-// Renders a bordered list of flag rows, or `emptyMessage` when there are none. Shared by the catalog
-// body and the "Local overrides" section — they differ only in border color and empty copy. Reads
-// the override state + actions from context so each row's wiring stays identical.
+/**
+ * A bordered list of flag rows, shared by the catalog body and the "Local overrides" section — they
+ * differ only in border color and empty copy.
+ */
 function FlagList({
   flags,
   borderColor,
@@ -126,15 +126,13 @@ function FlagRow({
     <Group
       justify="space-between"
       wrap="nowrap"
-      // Top-align so the variant buttons stay up beside the name/key instead of drifting to the
-      // vertical middle of a long description.
+      // Keeps the variant buttons beside the name/key rather than centred on a long description.
       align="flex-start"
       px="sm"
       py="sm"
       style={{
         borderBottom: '1px solid var(--mantine-color-default-border)',
-        // Mantine's scheme-aware subtle tint (same one variant="light" uses): light violet in light
-        // mode, a muted translucent violet in dark mode — not a full saturated fill.
+        // Mantine's scheme-aware subtle tint (the one variant="light" uses), not a saturated fill.
         backgroundColor: overridden ? 'var(--mantine-color-violet-light)' : undefined,
       }}
     >
@@ -160,10 +158,9 @@ function FlagRow({
         ) : (
           flag.variants.map((variant) => {
             const isActive = overridden && valuesEqual(override.value, variant.value)
-            // The catalog falls back to the raw string when a variant doesn't parse as its
-            // declared type (see parseVariantValue) — writing that through would violate the
-            // same contract validateOverrideValue enforces for manual overrides. `allowNull` keeps
-            // a legitimate JSON `null` variant applyable (a raw-string type mismatch still fails).
+            // The catalog keeps an unparseable variant as its raw string (see parseVariantValue), and
+            // writing that through would break the override type contract. `allowNull` keeps a
+            // legitimate JSON `null` variant applyable.
             const validationError = validateOverrideValue(flag.type, variant.value, { allowNull: true })
             return (
               <Button
@@ -187,19 +184,18 @@ function FlagRow({
   )
 }
 
-// Free-text descriptions can run long. Show a single line by default with a "Show more" toggle that
-// expands the rest inline. The toggle only appears when the one-line clamp actually hides something —
-// measured rather than guessed from length, since a short description can still wrap and a long one
-// might fit.
+/**
+ * Clamps a long description to one line behind a "Show more" toggle. The toggle appears only when
+ * the clamp actually hides something — measured rather than guessed from length, since a short
+ * description can still wrap and a long one might fit.
+ */
 function FlagDescription({ description }: { description: string }) {
   const [expanded, setExpanded] = useState(false)
   const [overflowing, setOverflowing] = useState(false)
   const textRef = useRef<HTMLParagraphElement>(null)
 
-  // Measure whether the collapsed description overflows one line, so we know to offer "Show more".
-  // Skip while expanded — the clamp is off then, so a measurement would read as "fits" and wrongly
-  // hide the toggle; `overflowing` keeps its collapsed value. The ResizeObserver re-measures when the
-  // panel width changes, so narrowing the DevTools panel surfaces a newly-clamped description's toggle.
+  // Skipped while expanded: the clamp is off then, so a measurement would read as "fits" and wrongly
+  // hide the toggle. The ResizeObserver re-measures when the panel width changes.
   useLayoutEffect(() => {
     const el = textRef.current
     if (!el || expanded) {
@@ -213,15 +209,12 @@ function FlagDescription({ description }: { description: string }) {
   }, [description, expanded])
 
   return (
-    // Extra top margin gives the description a touch more separation from the key above it than the
-    // name↔key gap, so the row reads as "title/key" then "description".
+    // Slightly more separation than the name↔key gap, so the row reads as "title/key" then "description".
     <Box mt={4}>
       <Text ref={textRef} size="xs" lineClamp={expanded ? undefined : 1}>
         {description}
       </Text>
       {overflowing && (
-        // Accent color in both states so it reads as the row's action. A hair smaller than the
-        // description text and sitting right beneath it, so the two read as clearly distinct.
         <Anchor
           component="button"
           type="button"
@@ -247,7 +240,7 @@ function FlagKey({ value }: { value: string }) {
           overflow: 'hidden',
           textOverflow: 'ellipsis',
           whiteSpace: 'nowrap',
-          // Negate the chip's own horizontal padding so the key text lines up with the flag name above.
+          // Negates the chip's own padding so the key lines up with the flag name above.
           paddingInline: 6,
           marginLeft: -6,
         }}
@@ -257,7 +250,6 @@ function FlagKey({ value }: { value: string }) {
       <CopyButton value={value}>
         {({ copied, copy }) => (
           <Tooltip label={copied ? 'Copied' : 'Copy key'} withArrow>
-            {/* Flip to violet on copy for a moment of feedback, then back to neutral grey. */}
             <ActionIcon
               size="xs"
               variant="subtle"
--- a/developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
+++ b/developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
@@ -1,6 +1,19 @@
-import { ActionIcon, Box, Button, Code, CopyButton, Group, Loader, Space, Text, Tooltip } from '@mantine/core'
+import {
+  ActionIcon,
+  Anchor,
+  Box,
+  Button,
+  Code,
+  CopyButton,
+  Group,
+  Loader,
+  Space,
+  Stack,
+  Text,
+  Tooltip,
+} from '@mantine/core'
 import { IconArrowBackUp, IconCopy } from '@tabler/icons-react'
-import React, { type ReactNode } from 'react'
+import React, { useLayoutEffect, useRef, useState, type ReactNode } from 'react'
 import type { CatalogFlag } from './flagsRequests'
 import { useFlagsContext } from './flagsContext'
 import { validateOverrideValue } from './flagTypes'
@@ -29,7 +42,7 @@ export function FlagCatalogBody() {
       <Space h="xs" />
       <FlagList
         flags={bottomFlags}
-        borderColor="var(--mantine-color-gray-2)"
+        borderColor="var(--mantine-color-default-border)"
         // `bottomFlags` is the page minus overridden flags (those are pinned above). Only call it "no
         // match" when the server total is 0; otherwise this page's flags are all overridden.
         emptyMessage={
@@ -57,7 +70,7 @@ export function OverridesSection() {
         Local overrides ({overriddenFlags.length})
       </Text>
       <Space h="xs" />
-      <FlagList flags={overriddenFlags} borderColor="var(--mantine-color-violet-2)" />
+      <FlagList flags={overriddenFlags} borderColor="var(--mantine-color-violet-outline)" />
     </>
   )
 }
@@ -113,20 +126,25 @@ function FlagRow({
     <Group
       justify="space-between"
       wrap="nowrap"
-      align="center"
+      // Top-align so the variant buttons stay up beside the name/key instead of drifting to the
+      // vertical middle of a long description.
+      align="flex-start"
       px="sm"
-      py="xs"
+      py="sm"
       style={{
-        borderBottom: '1px solid var(--mantine-color-gray-1)',
-        backgroundColor: overridden ? 'var(--mantine-color-violet-0)' : undefined,
+        borderBottom: '1px solid var(--mantine-color-default-border)',
+        // Mantine's scheme-aware subtle tint (same one variant="light" uses): light violet in light
+        // mode, a muted translucent violet in dark mode — not a full saturated fill.
+        backgroundColor: overridden ? 'var(--mantine-color-violet-light)' : undefined,
       }}
     >
-      <Box style={{ minWidth: 0, flex: 1 }}>
+      <Stack gap={6} style={{ minWidth: 0, flex: 1 }}>
         <Text size="sm" fw={600} truncate>
           {flag.name}
         </Text>
         <FlagKey value={flag.key} />
-      </Box>
+        {flag.description && <FlagDescription description={flag.description} />}
+      </Stack>
       <Group gap="xs" wrap="wrap" justify="flex-end" style={{ flexShrink: 0, maxWidth: '55%' }}>
         {overridden && (
           <Tooltip label="Revert override">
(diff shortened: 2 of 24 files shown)


developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
```diff
@@ -87,6 +238,10 @@ function FlagKey({ value }: { value: string }) {
           overflow: 'hidden',
           textOverflow: 'ellipsis',
           whiteSpace: 'nowrap',
+          paddingInline: 6,
+          // Pull the chip left by its own padding so the key TEXT lines up with the flag name above,
+          // while the grey box keeps its inner breathing room.
+          marginLeft: -6,
```
BeltranBulbarellaDD, line 243: This is weird? What is the issue here? Could we use flexbox?

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
```diff
@@ -39,10 +58,122 @@ export function FlagFilterBar() {
...
       </Group>
     </Stack>
   )
 }
+
+function MyFlagsSwitch({ view, identity }: { view: FlagCatalogView; identity: FlagIdentityState }) {
+  // Without the signed-in user's UUID there's nothing to compare a flag's creator against, so the
+  // filter could only ever match nothing. Disable it and say why, rather than offering a toggle whose
+  // only effect is to empty the list.
+  const unavailable = !identity.loading && !identity.userId
```
BeltranBulbarellaDD, line 71: I wonder, there is no way to see this component without being authenticated right? If we weren't authenticated the tab would fail before. I think we can remove these kinds of checks from the lower level components and in the root layout we check the auth, if not auth we show login screen, else all components are authenticated?

developer-extension/src/panel/components/tabs/flagsTab/flagIdentity.ts
```diff
@@ -0,0 +1,98 @@
...
+    `https://${getFlagsApiHost(site)}/api/v2/current_user`,
+    token,
+    'Current user request failed'
+  )
+  return body.data?.id ?? null
+}
+
+// The teams endpoint caps `page[size]` at 100; a user in more than 100 teams is unrealistic, so a
+// single page covers every real case.
+const TEAM_PAGE_SIZE = 100
```
BeltranBulbarellaDD: Still, how about fetching 10 and putting for example a search bar? That way we don't fill the page. Even if we showed 40 it would be hard to read unless we have a scroll.

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
BeltranBulbarellaDD: Or in the context.

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
```diff
@@ -3,12 +3,25 @@
...
 
 import { mockable } from '../../../../../../packages/browser-core/src/tools/mockable'
 
-const CLIENT_ID = '13c94d15-067d-4263-a309-be4811141419'
-const SCOPES = ['feature_flag_config_read', 'feature_flag_environment_config_read']
+// Staging (datad0g.com) and prod have separate registered OAuth clients. The prod client is
+// replicated across all prod DCs, so every non-staging site shares it. Both are public PKCE clients.
+const STAGING_CLIENT_ID = '13c94d15-067d-4263-a309-be4811141419'
+const PROD_CLIENT_ID = '2c19b57d-118a-4f52-bcfb-709503a68290'
+
```
BeltranBulbarellaDD, line 12: Please add in the confluence docs how can we get these values, and what would take to rotate them.

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
```diff
@@ -131,37 +137,90 @@ async function requestToken(host: string, body: URLSearchParams, fallbackRefresh
...
+// loginWithOAuth treats it the same as InvalidScopeError and retries without the optional scopes.
+class AuthorizationPageLoadError extends Error {}
+
 /**
  * Runs the interactive OAuth flow: opens Datadog's login/consent screen, then exchanges the
  * returned authorization code for tokens. Returns the tokens (caller is responsible for storing).
+ *
+ * Asks for the optional scopes first and retries without them if the client isn't granted them, so
+ * that a not-yet-updated OAuth client registration costs us one optional filter rather than the
+ * ability to sign in at all. Usually the retry isn't a second prompt (an unauthorized scope is
```
BeltranBulbarellaDD: I don't understand in what case could this happen?

developer-extension/src/panel/components/tabs/flagsTab/useFlagCatalogView.ts
```diff
@@ -37,28 +48,43 @@ export function useFlagCatalogView(): FlagCatalogView {
     return () => clearTimeout(id)
   }, [search])
 
+  // "My feature flags" filters by the signed-in user's creator UUID server-side. With no known user
+  // it resolves to no filter (the UI disables the toggle in that case). Deriving it here — rather than
+  // depending on currentUserId in the memo — keeps the request identity stable when the toggle is off,
+  // so identity resolving after connect doesn't trigger a redundant refetch.
```
BeltranBulbarellaDD: Again I don't get why this could happen. We would always resolve the identity after connect?

developer-extension/src/panel/components/tabs/flagsTab/useFlagCatalogView.ts
```diff
@@ -37,28 +48,43 @@ export function useFlagCatalogView(): FlagCatalogView {
...
+  const createdBy = myFlagsOnly && currentUserId ? currentUserId : null
```
BeltranBulbarellaDD, line 49: In this stage shouldn't `currentUserId` always be resolved?

developer-extension/src/panel/components/tabs/flagsTab/useFlagIdentity.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/**
+ * Loads the signed-in user's UUID and team handles, which the "My feature flags" and "My teams"
+ * filters need (see flagIdentity.ts). Refetches whenever the connection state or site changes.
+ *
+ * Failures are deliberately silent: identity only enables two optional filters, so a failed lookup
+ * leaves them disabled rather than surfacing an error over a catalog that works fine without them.
+ * Unlike useFlagCatalog this does not disconnect on an unusable token — the catalog fetch runs
+ * alongside and owns that decision, and having both react to it would risk a double disconnect.
```
BeltranBulbarellaDD: So we could fetch the catalog but not have a userId?

BeltranBulbarellaDD (commented):
Overall looks good! Left some comments.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
kellyw1806: This was a small nitpick by design where we wanted the description, ID key name and flag name to be aligned perfectly on the left for the text. The key is inside a Code chip (the grey box) and we want the key text to line up with the flag name, not the box's edge. the chip has its own padding so the text needed to sit around 6px to the left. its a very small design thing, we can remove it if it's too weird to keep, or i can also adjust the comment there. 

image

--- versus --- (this is without)

image

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
kellyw1806: Oh, this was not an auth check, but more of a UX guard. The "My Feature Flags" filter needs the user's id to work. If we don't know it yet (identity is still loading or failed), the toggle would do nothing when clicked, so i thought it would be better to just disable it and show a tooltip explaining why. what do you think?

developer-extension/src/panel/components/tabs/flagsTab/flagIdentity.ts
kellyw1806: i think the search bar is a good idea! i made it so that if you are in more than 12 teams (web-ui teams filter does the same thing, the min is 12 there so i used it here), then a search bar appears. otherwise, its the same as what its like currently. regarding fetching fewer, the filter returns the teams you're actually in, so if we fetched 10 and you're in more, the rest would just disappear from the filter. so I'd keep fetching all of them (100 is just a ceiling and nobody's in that many) and let the search box handle the readability?

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
kellyw1806: yeah looked at this again and i think this case doesn't really happen. we only had this retry for the teams_read scope. but just confirmed that teams load without it, so I removed the whole optional-scope/retry path. This comment is gone now.

developer-extension/src/panel/components/tabs/flagsTab/useFlagCatalogView.ts
kellyw1806: oh not always. connecting just gets us a token but getting who the user actually is, is a separate request that can still be loading, or can fail. we don't disconnect when it fails since the catalog works without it (this is for when its loading and the catalog can show). so there's a tiny window where we're connected but might not have a user id yet

developer-extension/src/panel/components/tabs/flagsTab/useFlagCatalogView.ts
kellyw1806: so for this, its not guaranteed here. when you connect, the catalog loads right away but your identity comes back a moment later on a separate request. if that request fails, it stays empty. so currentUserId can posisbly be null at this point, which is why the guard is here

developer-extension/src/panel/components/tabs/flagsTab/useFlagIdentity.ts
kellyw1806: yes, exactly. the catalog and the identity are two separate requests, and the catalog only needs the token. So the catalog can load fine while your identity is still coming back or has failed. that's the whole reason the "My feature flags" filter checks for a user id before doing anything.

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
BeltranBulbarellaDD: But could we be logged in here but no user id?. Even still, if we were to check this in the context for example we could use the same check right? So we wouldn't have to check it in the component but just get it from the context? That way we can re use it? Just sharing some thoughts, you are free to discard it 🙂

developer-extension/src/panel/components/tabs/flagsTab/flagIdentity.ts
BeltranBulbarellaDD: Aaa I get it. Okay perfect. Sounds good!

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
```diff
@@ -35,20 +44,13 @@ export interface FlagSite {
...
 export const FLAG_SITES: FlagSite[] = [
   { site: 'datadoghq.com', host: 'app.datadoghq.com', label: 'US1 (datadoghq.com)' },
-  { site: 'us3.datadoghq.com', host: 'us3.datadoghq.com', label: 'US3 (us3.datadoghq.com)' },
-  { site: 'us5.datadoghq.com', host: 'us5.datadoghq.com', label: 'US5 (us5.datadoghq.com)' },
-  { site: 'datadoghq.eu', host: 'app.datadoghq.eu', label: 'EU1 (datadoghq.eu)' },
-  { site: 'ap1.datadoghq.com', host: 'ap1.datadoghq.com', label: 'AP1 (ap1.datadoghq.com)' },
-  { site: 'ap2.datadoghq.com', host: 'ap2.datadoghq.com', label: 'AP2 (ap2.datadoghq.com)' },
-  { site: 'ddog-gov.com', host: 'ddog-gov.com', label: 'US1-FED (ddog-gov.com)' },
-  { site: 'us2.ddog-gov.com', host: 'us2.ddog-gov.com', label: 'US2-FED (us2.ddog-gov.com)' },
   { site: 'datad0g.com', host: 'dd.datad0g.com', label: 'Staging (datad0g.com)' },
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Nah no need. It's just us using it.

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
BeltranBulbarellaDD: I now understand that `catalog` and `identity` are separate requests. I wonder the above, how possible would it be to save it in the context and re use them.

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
```diff
@@ -3,12 +3,21 @@
...
+// replicated across all prod DCs, so every non-staging site shares it. Both are public PKCE clients.
+const STAGING_CLIENT_ID = '13c94d15-067d-4263-a309-be4811141419'
+const PROD_CLIENT_ID = '2c19b57d-118a-4f52-bcfb-709503a68290'
+
+function getClientId(site: string): string {
+  return site === 'datad0g.com' ? STAGING_CLIENT_ID : PROD_CLIENT_ID
+}
+// The only scopes we request. GET /api/v2/team needs no scope — team access is gated on the user's
+// Datadog permissions, not the token's scopes — so the "My teams" filter works without teams_read.
+const REQUIRED_SCOPES = ['feature_flag_config_read', 'feature_flag_environment_config_read']
```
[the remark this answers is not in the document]
kellyw1806: we tested this and teams_read loads fine on staging and prod.

developer-extension/src/panel/components/tabs/flagsTab/flagFilterBar.tsx
kellyw1806: hmm, the identity is already shared through the context (both the filter bar and catalog view read it from there), and the only thing left in the component is a one-line "do we have a user id yet" check used by a single toggle. so i think we might not need to try to find more places for context reuse here?

developer-extension/src/panel/components/panel.tsx
```diff
@@ -54,11 +54,9 @@ export function Panel() {
           <Tabs.Tab value={PanelTabs.Replay}>
             <Text>Live replay</Text>
           </Tabs.Tab>
-          {settings.datadogMode && (
-            <Tabs.Tab value={PanelTabs.Flags}>
-              <Text>Feature Flags</Text>
-            </Tabs.Tab>
-          )}
+          <Tabs.Tab value={PanelTabs.Flags}>
```
BeltranBulbarellaDD, line 57: Image
In the Settings tab when we override a value we show this icon. Would it be possible to do the same here? If on the page we are, we have overriden a FF we could show the icon?

developer-extension/src/panel/components/tabs/flagsTab/flagsTab.tsx
```diff
@@ -53,9 +53,13 @@ function ConnectedFlagsTab({ auth }: { auth: FlagAuthState }) {
   return (
     <TabBase
       top={
-        // No dd-privacy-allow here: the header/filter/catalog below render customer flag names,
-        // values, and tags, which must stay masked in the extension's own Session Replay.
-        <Box px="md">
+        // No px here: the TabBase Container already insets by `md`, so the header lines up with the
+        // catalog rows below (which get their `md` from their own Box). No dd-privacy-allow either —
+        // the header/filter/catalog render customer flag names, values, and tags, which must stay
+        // masked in the extension's own Session Replay.
```
BeltranBulbarellaDD: Is this needed? The comment.

developer-extension/src/panel/components/tabs/flagsTab/manualOverrideForm.tsx
BeltranBulbarellaDD, line 1: Can we surface the error to the user? Like if the flag cannot be created because there is already one?

https://github.com/user-attachments/assets/d8db5286-fbba-4693-9164-754f464557db

BeltranBulbarellaDD (commented):
Overall LGTM! Just some minor edge cases and I think it could be good to merge!

What is the expected behaviour when we are not logged in but we have a local overriden FF? Should it still apply? Cause we have it locally? Because I went to the FF test page and it was in dark mode even though I was unauthenticated.

developer-extension/src/panel/components/tabs/flagsTab/manualOverrideForm.tsx
BeltranBulbarellaDD: Also maybe we can move the disconnect button to the far right? I think where it is now it could be clicked by mistake. 
About the connected message it's a little bit weird because since it's in caps it's hard to tell it's staging. 
datad0g vs DATAD0G. But no problem for now I believe.

Screenshot 2026-08-13 at 11 31 34

developer-extension/src/panel/components/panel.tsx
kellyw1806: Hm, was thinking about this but i think it might not be worth the cost? it would need the override state at Panel level, so the extension would read the inspected page on every navigation for every user, including everyone who never opens Flags. That might be a lot of extension-wide overhead? i'm also going to add an alert on the Auth login page so that if there are overrides that exist, it will let you know. maybe that can help as well. having the override remain across disconnect and session restarts is intended, but wanted to add some sort of warning to users

image

developer-extension/src/panel/components/tabs/flagsTab/flagIdentity.ts
```diff
@@ -0,0 +1,87 @@
...
+import { getFlagsApiHost } from './oauth'
+
+export interface FlagIdentity {
+  /** UUID of the signed-in user, or null when it couldn't be resolved. */
+  userId: string | null
+  /** Handles of the teams the signed-in user belongs to, sorted for stable display. */
+  teamHandles: string[]
+  /** True when the team lookup was refused because the user lacks permission to read teams. */
+  teamsForbidden: boolean
+  /** True when the team lookup failed for another reason (network/server), distinct from an empty membership. */
```
BeltranBulbarellaDD: I feel the comments do not add any value here.

developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
```diff
@@ -113,20 +126,23 @@ function FlagRow({
...
+      // Keeps the variant buttons beside the name/key rather than centred on a long description.
+      align="flex-start"
       px="sm"
-      py="xs"
+      py="sm"
       style={{
-        borderBottom: '1px solid var(--mantine-color-gray-1)',
-        backgroundColor: overridden ? 'var(--mantine-color-violet-0)' : undefined,
+        borderBottom: '1px solid var(--mantine-color-default-border)',
+        // Mantine's scheme-aware subtle tint (the one variant="light" uses), not a saturated fill.
```
BeltranBulbarellaDD: no need I believe

developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
```diff
@@ -169,6 +184,52 @@ function FlagRow({
...
+    }
+    const measure = () => setOverflowing(el.scrollHeight > el.clientHeight)
+    measure()
+    const observer = new ResizeObserver(measure)
+    observer.observe(el)
+    return () => observer.disconnect()
+  }, [description, expanded])
+
+  return (
+    // Slightly more separation than the name↔key gap, so the row reads as "title/key" then "description".
```
BeltranBulbarellaDD: i don't think it adds value

developer-extension/src/panel/components/tabs/flagsTab/flagsRequests.ts
```diff
@@ -1,27 +1,45 @@
+import { fetchFfeJson } from './ffeApi'
 import { FLAG_TYPES, parseTypedString, type FlagType } from './flagTypes'
 import { getFlagsApiHost } from './oauth'
 
 export interface CatalogFlag {
   key: string
   name: string
+  /** Free-text description authored in the Datadog UI. Empty when the flag has none. */
```
BeltranBulbarellaDD: The comment is redundant i think?

developer-extension/src/panel/components/tabs/flagsTab/flagsRequests.ts
```diff
@@ -1,27 +1,45 @@
...
+ */
 export interface FlagCatalogRequest {
   page: number
   pageSize: number
   search: string
   typeFilter: string[]
   tagFilter: string[]
+  /** Team handles for the "My teams" filter, sent as `tags=team:<handle>`. */
+  teamFilter: string[]
+  /** The signed-in user's UUID when "My feature flags" is on, else null. Sent as `created_by`. */
```
BeltranBulbarellaDD: This comment is repeated above twice.

developer-extension/src/panel/components/tabs/flagsTab/oauth.ts
```diff
@@ -216,6 +227,55 @@ export async function clearStoredTokens(): Promise<void> {
...
+ */
+export async function revokeAndClearTokens(site: string): Promise<{ revoked: boolean }> {
+  const revoked = await tryRevokeGrant(site)
+  await clearStoredTokens()
+  return { revoked }
+}
+
+/**
+ * Revokes the refresh token — the renewable part of the grant. The access token is left to expire
+ * (RFC 7009 only *recommends* cascading revocation, so we don't rely on it) and is dropped locally
```
BeltranBulbarellaDD: I don't get this line.  Should we have a link to the RFC? or remove it?

developer-extension/src/panel/components/tabs/flagsTab/useFlagCatalogView.ts
```diff
@@ -2,8 +2,7 @@ import { useEffect, useMemo, useState } from 'react'
 import type { FlagCatalogRequest } from './flagsRequests'
 
 const CATALOG_PAGE_SIZE = 20
-// Wait out a typing burst before sending a search to the server, so we don't fire a request per
-// keystroke. Short enough to still feel responsive.
+// Long enough to wait out a typing burst, short enough to still feel responsive.
```
BeltranBulbarellaDD: No need, we know what a debounce is.

BeltranBulbarellaDD (approved):
Minor nits but LGTM!

PR #4940, opened 2026-08-12
🐛 Prevent duplicate Next.js RUM views from discarded renders

### LINKED ISSUE #4931: 🐛 Router components create duplicate RUM views: render-phase startView() guarded only by useRef
### Product
- `rum`
- `datadog/browser-sdk`
- `datadog/browser-rum-nextjs`

### Environment
- `@datadog/browser-rum` : 6.33.0
- `@datadog/browser-rum-nextjs` : 6.33.0
- `next` : 16.1.6 (App Router)
- `react/react-dom` : 19.2.4
- `reactStrictMode` : false

<br />

**Describe the bug**
DatadogAppRouter from @datadog/browser-rum-nextjs calls startNextjsView() during the render phase, and relies on a useRef guard alone to prevent duplicates.

```ts
export function DatadogAppRouter() {
  const pathname = mockable(usePathname)()
  const params = mockable(useParams)()
  const previousPathname = mockable(useRef)<string | null>(null)

  if (previousPathname.current !== pathname) {   // reading a ref during render
    previousPathname.current = pathname          // writing a ref during render
    startNextjsView(computeViewNameFromParams(pathname, params))  // side effect during render
  }

  return null
}
```

React is free to discard a render that never commits and start over from scratch. When that happens, a component that has not committed yet gets a brand-new useRef along with its new fiber, so the guard is reset — but the startView() call has already escaped and cannot be undone.

The guard rolls back. The side effect does not.

The result is one duplicate RUM view per discarded mount render.

Because nextjsPlugin sets initConfiguration.trackViewsManually = true in onInit, there is no dedupe layer underneath — one startView() call is one real, billable view.

This is documented usage, not misuse
The "App router usage" section of packages/browser-rum-nextjs/README.md instructs users to render DatadogAppRouter from the root layout:
> 2. Call the DatadogAppRouter component from your root layout.
> ```ts
> // app/layout.tsx — from the README
> export default function RootLayout({ children }) {
>   return (
>     <html lang="en">
>       <body>
>         <DatadogAppRouter />
>         {children}
>       </body>
>     </html>
>   )
> }
> ```

Our setup matches this example exactly, so the bug reproduces on the documented installation path.

**To Reproduce**
1. Set up a Next.js App Router app with nextjsPlugin() and <DatadogAppRouter />.
2. Render the following probe right next to <DatadogAppRouter /> (same parent, adjacent position):
```ts
'use client'
import { useEffect, useRef } from 'react'

let renderSeq = 0   // module scope — unaffected when a render is discarded

export const RestartProbe = () => {
  const instanceId = useRef<number | null>(null)

  if (instanceId.current === null) {
    renderSeq += 1              // only increments for a brand-new fiber
    instanceId.current = renderSeq
  }

  console.log(`[render] instance=${instanceId.current} t=${Math.round(performance.now())}`)

  useEffect(() => {
    console.log(`[commit] instance=${instanceId.current}`)
  }, [])

  return null
}
```
renderSeq lives in module scope, so discarding a render does not reset it. useRef, on the other hand, is recreated whenever a render is discarded before committing. That means an increasing instance number is a direct count of how many times the mount render was retried on a new fiber — which is exactly how many times DatadogAppRouter's previousPathname guard was reset.

3. Reload the page and check the console.
**Measured output (single page reload)**
```bash
[render] instance=1 t=2315
[render] instance=2 t=2357
[render] instance=3 t=2364
[render] instance=4 t=2372
[render] instance=5 t=2380
[render] instance=6 t=2386
[commit] instance=6
```

What this shows:
- 6 mount renders, each on a different fiber (instance goes 1 → 6)
- 1 commit, on instance=6 — only the last attempt committed; the first 5 were discarded
- instances 2 through 6 all happened within 29ms (t=2357 → 2386)

DatadogAppRouter sits in the same position and goes through the same 6 mount renders. Its useRef guard is recreated as null every time, so startNextjsView() fires 6 times — 5 of them from renders that never made it to the screen.

In the same session, the RUM dashboard showed the same view count increasing repeatedly for `/[locale]/videos/[id]`.

**Expected behavior**
A single page reload should create exactly one view.

`startNextjsView()` should only run from a committed render. A discarded render must not create a view.

BeltranBulbarellaDD:
## Motivation

Fixes #4931.

`DatadogAppRouter` used to call `startNextjsView()` while React rendered, with a `useRef` as its only guard. React may discard and retry an uncommitted render. Each retry gets a new ref, but the RUM side effect from the previous attempt has already escaped:

```
render attempt 1: useRef(null) -> startView("/user/42") -> discarded
render attempt 2: useRef(null) -> startView("/user/42") -> discarded
render attempt 3: useRef(null) -> startView("/user/42") -> committed
```

The result is several RUM views with distinct IDs for one displayed page.

Moving that ref to module scope would only hide part of the symptom and would make React state global. It would still perform `startView()` during render, so an abandoned render could create or rename a view for a route that never committed. React requires render to remain pure; effects run after commit. See [React's rule](https://react.dev/reference/rules/components-and-hooks-must-be-pure) and [Synchronizing with Effects](https://react.dev/learn/synchronizing-with-effects).

## Approach

We split view creation from view-name normalization:

1. **Initial load:** `nextjsPlugin.onInit()` starts one view immediately, outside React.
2. **Client navigation:** Next.js calls `onRouterTransitionStart()` before React renders. It starts the route-change view at navigation time, preserving loading timing for slow routes. This is the intended role of the [Next.js hook](https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation-client#router-navigation-tracking).
3. **Committed route:** `DatadogAppRouter` uses `useEffect` to call `setNextjsViewName()`. This only turns a concrete path such as `/user/42` into `/user/[id]`; it does not create a view or block paint.

```
onInit()                          -> startView("/user/42")        initial_load
onRouterTransitionStart("/slow") -> startView("/slow")            route_change, before commit
React commits "/user/42"          -> useEffect -> setViewName("/user/[id]")
```

This follows the same separation used by [PostHog's App Router component](https://github.com/PostHog/posthog-js/blob/7222accf674b6e005a92d6987174db61d39c9540/packages/next/src/client/PostHogPageView.tsx#L141-L161): derive route state in render and publish it from `useEffect`.

## Transition deduplication

Next.js can provide an opaque transition `event.id` when `experimental.instrumentationClientRouterTransitionEvents` is enabled. The plugin records the last ID and ignores a repeated callback for that same transition. Different IDs remain separate navigation attempts, including redirects. Next creates this ID in [`startRouterTransition`](https://github.com/vercel/next.js/blob/665d3458b616accd6e0ae942d794eb223273d616/packages/next/src/client/components/router-transition.ts#L62).

Sentry also starts navigation instrumentation from `onRouterTransitionStart` rather than waiting for React to commit. Its deduplication is specifically for overlap between its legacy router patch and the hook, not repeated hook calls: [handler](https://github.com/getsentry/sentry-javascript/blob/develop/packages/nextjs/src/client/routing/appRouterRoutingInstrumentation.ts#L119-L157), [hook entry point](https://github.com/getsentry/sentry-javascript/blob/develop/packages/nextjs/src/client/routing/appRouterRoutingInstrumentation.ts#L314-L323).

## Tests

- Unit: one initial view, view-name normalization, query/hash filtering, and repeated transition-ID deduplication.
- E2E: discarded initial renders produce one `initial_load` view.
- E2E: a slow route starts its view before the route commits.
- E2E: redirects retain one route-change view per transition.


packages/browser-rum-nextjs/src/domain/nextJSRouter/useStartNextjsView.ts
```diff
@@ -0,0 +1,13 @@
...
+import { startNextjsView } from '../nextjsPlugin'
+
+export function useStartNextjsView(path: string | null, viewName: string) {
+  const previousPath = useRef<string | null>(null)
+
+  useLayoutEffect(() => {
+    if (path !== null && previousPath.current !== path) {
+      previousPath.current = path
+      startNextjsView(viewName)
+    }
```
lierniel: Question:
Why not just `useEffect()` ? We're blocking the render by using `useLayoutEffect()`, which is not that I would expect from a monitoring tool

BeltranBulbarellaDD: The thing is that `useLayoutEffect`  runs before paint and just by doing this fix we are going to be starting views later that what we would like to. useEffect would be just too late to start a view.

BeltranBulbarellaDD: Actually I like what you said. I'm going back to the `useEffect` also because I saw other monitoring tools do the same.

lierniel (approved):
LGTM 🚀

PR #4941, opened 2026-08-12
✅ [RUM-17233] Add Shopify checkout views and actions e2e test

lierniel:
## Motivation

[RUM-17233](https://datadoghq.atlassian.net/browse/RUM-17233): cover the Shopify integration with e2e tests. The dedicated `browser-rum-shopify` bundle (#4878) explicitly scoped e2e tests out; this closes that gap by exercising the full checkout flow against a real Shopify dev store.

## Changes

* Added e2e test framework support for Shopify: `buildShopifyUrl`/`unlockShopifyStorePassword` (dev store password gate), `withShopifyApp()`/`shopifySetup` (intercepts the CDN bundle request and injects RUM configuration into both the storefront's Theme Liquid snippet and the checkout's Custom Pixel sandbox), and a `getShopifyStorePassword()` secret accessor.
* Added `shopify.scenario.ts`, covering the full flow: storefront → product page → add to cart → checkout → thank-you page → back to storefront. Asserts:
  * A single session id shared across the Theme Liquid and Custom Pixel SDK instances
  * 5 ordered view events with the correct URLs (checkout/thank-you matched via the same path regex used in `browser-rum-shopify`)
  * Ordered click actions with the correct target and page for each
  * No unexpected browser console errors, filtering known store-side noise
* Runs serially with a 15s cooldown between invocations, and `test.slow()`, since this hits a live dev store that's sensitive to request volume (Cloudflare bot protection, checkout rate limiting) and the full flow runs close to the default test timeout.

## Test instructions

To run locally:
```
SHOPIFY_STORE_PASSWORD=<dev store password> yarn test:e2e -g "shopify"
```

Or trigger the according pipeline on the gitlab: https://gitlab.ddbuild.io/DataDog/browser-sdk/-/pipelines

Chromium only (`--disable-web-security` is required to bypass the dev store's CSP/CORS restrictions). Runs against the live `custom-pixel-e2e.myshopify.com` dev store, so repeated back-to-back local runs may need a few minutes' gap to avoid tripping Cloudflare's verification wall.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

[RUM-17233]: https://datadoghq.atlassian.net/browse/RUM-17233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/e2e/lib/framework/buildShopifyUrl.ts
```diff
@@ -0,0 +1,23 @@
+import type { Page } from '@playwright/test'
+import { getShopifyStorePassword } from '../../../../scripts/lib/secrets.ts'
+
+// A Datadog-owned dev store, password-protected, used only to exercise browser-rum-shopify
+// against a real storefront + checkout + Custom Pixel sandbox.
+const SHOPIFY_STORE_URL = 'https://custom-pixel-e2e.myshopify.com/'
+
+export function buildShopifyUrl(): string {
+  return SHOPIFY_STORE_URL
+}
```
BeltranBulbarellaDD: Maybe since it's a constant we don't need this?

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -343,6 +344,41 @@ export async function salesforceSetup(options: SetupOptions, servers: Servers, p
   return ''
 }
 
+// Matches the CDN URL used by the store's Theme Liquid snippet and Custom Pixel for the main
+// bundle and its dynamically-imported chunks (e.g. the session replay recorder), served from
+// `https://www.datadoghq-browser-agent.com/<site>/v<major>/[chunks/]<name->]datadog-rum-shopify.js`
+const SHOPIFY_ASSET_URL_PATTERN =
```
BeltranBulbarellaDD, line 369: Okay not related specifically to your PR but how about making a function to reuse with Salesforce? Since it's the same pattern except for `datadog-rum-shopify` vs `datadog_rum_salesforce`?

test/e2e/scenario/shopify.scenario.ts
```diff
@@ -0,0 +1,206 @@
...
+test.describe(() => {
+  // The live dev store's Cloudflare bot protection can trip a verification wall under request volume,
+  // even from rapid consecutive solo runs. Force this file's test(s) to run one at a time.
+  test.describe.configure({ mode: 'serial' })
+
+  // Cooldown so back-to-back invocations (manual loops, CI retries) can't hammer the store fast
+  // enough to trip Cloudflare's wall or Shopify's checkout rate limit
+  test.beforeAll(async () => {
+    await new Promise((resolve) => setTimeout(resolve, 15_000))
+  })
```
BeltranBulbarellaDD, line 58: I think we can remove this. Since for now we only have 1 test and it's scheduled.

test/e2e/scenario/shopify.scenario.ts
```diff
@@ -0,0 +1,206 @@
...
+          const response = await fetch('/cart.js')
+          // Treat a transient store-side 5xx as "not there yet" so expect.poll retries instead of
+          // failing on a JSON parse error.
+          if (!response.ok) {
+            return 0
+          }
+          const cart = (await response.json()) as { item_count: number }
+          return cart.item_count
+        })
+      await expect.poll(getCartItemCount, { timeout: 10000 }).toBeGreaterThan(0)
```
BeltranBulbarellaDD: I think we can remove the timeout. At least I did locally and we are good.

test/e2e/scenario/shopify.scenario.ts
```diff
@@ -0,0 +1,206 @@
...
+        .getByRole('textbox', { name: 'Expiration date (MM / YY)' })
+        .fill('12 / 99')
+      await page
+        .frameLocator('iframe[name^="card-fields-verification_value-"]')
+        .getByRole('textbox', { name: 'Security code' })
+        .fill('123')
+
+      // Card fields sync back to the parent form via an async postMessage; give it a moment before
+      // clicking "Pay now" so the click doesn't land before the form is submittable.
+      await page.waitForTimeout(500)
```
BeltranBulbarellaDD, line 143: By default the page.getByRole wait's 30s. So I don't think this is necessary.

test/e2e/scenario/shopify.scenario.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      await page.getByRole('button', { name: 'Pay now' }).click()
+
+      await expect(page.getByRole('heading', { name: /thank you/i })).toBeVisible({ timeout: 30000 })
```
BeltranBulbarellaDD: Same here.

BeltranBulbarellaDD: (I tested locally and it passed)

test/e2e/scenario/shopify.scenario.ts
```diff
@@ -0,0 +1,206 @@
...
+      // The Theme Liquid and Custom Pixel SDK instances share one browser session.
+      const sessionIds = new Set([
+        ...intakeRegistry.rumViewEvents.map((e) => e.session.id),
+        ...intakeRegistry.rumActionEvents.map((e) => e.session.id),
+      ])
+      expect(sessionIds.size).toBe(1)
+
+      const orderedViews = [...intakeRegistry.rumViewEvents]
+        .sort((a, b) => a.date - b.date)
+        .filter((event, index, events) => events.findIndex((e) => e.view.id === event.view.id) === index)
```
BeltranBulbarellaDD, line 167: All of this is because we navigate back. Do we care about this assert? To go back to the page? Since we know the SDK worked correctly until now?

BeltranBulbarellaDD (commented):
LGTM! Just some minor comments.

test/e2e/lib/framework/pageSetups.ts
```diff
@@ -343,6 +344,41 @@ export async function salesforceSetup(options: SetupOptions, servers: Servers, p
...
+  /datadoghq-browser-agent\.com\/[^/]+\/v\d+\/(chunks\/)?([\w-]*datadog-rum-shopify\.js)(?:[?#].*)?$/
+
+// Shopify apps don't serve a locally-generated page body; this factory only intercepts the
+// bootstrap script request and injects the RUM configuration read by the store's Theme Liquid
+// snippet and Custom Pixel.
+export async function shopifySetup(options: SetupOptions, servers: Servers, page: Page): Promise<string> {
```
BenoitZugmeyer, line 372: **suggestion:** those setup functions are templates to render an HTML page the e2e test will load. It seems like you are doing very different things here. What about moving this logic in `shopify.scenario.ts`?

test/e2e/scenario/shopify.scenario.ts
lierniel: I'm not sure I got it properly, but we want to verify that the session stays the same after transition storefront page -> checkout page and vice verca. Here I'm sorting view events by time and filter them by uniq id to confirm that they came in correct order with right url

test/e2e/scenario/shopify.scenario.ts
```diff
@@ -0,0 +1,206 @@
...
+  test.describe.configure({ mode: 'serial' })
+
+  // Cooldown so back-to-back invocations (manual loops, CI retries) can't hammer the store fast
+  // enough to trip Cloudflare's wall or Shopify's checkout rate limit
+  test.beforeAll(async () => {
+    await new Promise((resolve) => setTimeout(resolve, 15_000))
+  })
+
+  // Flow: add to cart -> checkout -> thank you -> back to storefront
+  createTest('shopify checkout views and actions')
```
BenoitZugmeyer, line 61: **nitpick:** this test is huge. Is there any way to split it, so it's easier to maintain and understand what's failing?

test/e2e/scenario/shopify.scenario.ts
lierniel: Actually we still could have a few retries happening sequentially even with scheduled setup. I would keep it just in case, I believe it doesn't hurt to have anyway

test/e2e/lib/framework/shopify.ts
BenoitZugmeyer, line 1: **suggestion:** this file is not just about shopify urls. Maybe rename it to `shopify.ts` or `shopifyUtils.ts`?

test/e2e/scenario/shopify.scenario.ts
lierniel: This timeout is only for polling `/cart.js` with the interval ([playwright docs](https://playwright.dev/docs/test-assertions#expectpoll)). We're waiting for the amount of the items in cart in the response to be greater than 0 in order to avoid clicking on the "Checkout" button before items is actually landed to the cart. It might bypass without this polling a few times, but in a higher volumes of runs it could hit. 
I was running it in the loop of 40 sequential runs when I hit it and after this fix it doesn't happen anymore, therefore I'd rather keep it - but let me know if you disagree and want to discuss!

test/e2e/scenario/shopify.scenario.ts
lierniel: Yeah, but the thing is: the `Pay now` button is available immediately so page.getByRole will be executed right away without waiting for the element to appear (because it's already on the page from the start). But before clicking on `Pay now` I want to make sure that card fields form (which is on separate iframe) is synced with the parent form on the top frame. 
Without that it might happen that regardless of the card fields was already filled, clicking on the `Pay now` button fails because card details are missing (the parent form haven't received the update from card fields form).
That's why I added this `waitForTimeout(500)` - it's enough time for the card details form to finish syncing with parent form and now clicking on the `Pay now` button never fails

test/e2e/scenario/shopify.scenario.ts
BeltranBulbarellaDD: Oh okay, makes sense, it's just that with the first promise and then the timeout seemed redundant at least in 1/2 runs I tried.

test/e2e/lib/framework/pageSetups.ts
lierniel: I was following the example of the Salesforce, which has similar setup pattern

> It seems like you are doing very different things here

Actually not that different - the part of rendering an HTML page for e2e test is injecting the local sdk version and passing the init options for it - and this is exactly what I'm doing here. This method doesn't render the html - it's rendered by predefined Shopify dev-store - but I still need to override the sdk bundle path and pass the init config for it. Therefore I believe it's the right place to do this, it would look overly cumbersome in `shopify.scenario.ts` - but let me know if you disagree, I'm open to the discussion :)

test/e2e/scenario/shopify.scenario.ts
BeltranBulbarellaDD: Oh yeah vice verca I understand. Well it makes sense.

test/e2e/scenario/shopify.scenario.ts
lierniel: Yeah, but the reason behind putting everything in one test is the Claudflare bot wall we were discussing on Wednesday. I wanted to avoid parallel runs as much as possible - so I've created one big test to cover everything. It's true that with this approach it's harder to maintain and understand what is failing, but it's a trade of for avoiding multiple runs from happening at the same time to bypass Claudflare.
However I believe we could soften the assertions with [expect.soft()](https://playwright.dev/docs/test-assertions#soft-assertions) so test would keep running even if some assertions fail to better understand that's really broken, wdyt?

test/e2e/scenario/shopify.scenario.ts
lierniel: Oh, indeed, the timeout option may be redundant since it defaults to 30s - I'll remove it 👍

BenoitZugmeyer (approved):

BeltranBulbarellaDD (approved):

PR #4946, opened 2026-08-13
🐛 Fix Shopify Custom Pixel actions being marked as background events, flatten ui_extension_errored context

lierniel:
## Motivation

[RUM-18056](https://datadoghq.atlassian.net/browse/RUM-18056)

Two related fixes to the Shopify Custom Pixel integration:

1. Clicks tracked via `startAction`/`stopAction` in the Custom Pixel sandbox were always being reported with `view.in_foreground: false`. The pixel runs in a sandboxed iframe with no focusable UI, so `document.hasFocus()` always returns `false` there, which `pageStateHistory.ts` interprets as the page being backgrounded — even when the customer is actively on the checkout page.
2. Addresses [review feedback on #4878](https://github.com/DataDog/browser-sdk/pull/4878#discussion_r3622767715): the `ui_extension_errored` error context nested an `extension: { name, target, type, appName }` object whose `type` field read as the same concept as the SDK's own error-name-derived `type`, but is actually an unrelated Shopify `UiExtensionErrorType` enum (e.g. `EXTENSION_USAGE_ERROR`).

## Changes

- `patchSandboxedIframeApis.ts`: shims `document.hasFocus()` to always return `true` inside the Custom Pixel sandbox, alongside the existing `cookieStore`/`navigator.locks`/`contentWindow` shims for that same sandboxed context.
- `shopifyBindings.ts`: flattens the `ui_extension_errored` context and renames `type` to `extensionErrorType` to disambiguate it from the SDK's own error type. Also forwards `appId` and `appVersion`, which Shopify includes on the error payload but weren't previously captured.
- Updated `shopifyBindings.spec.ts` for the new flattened shape.

## Test instructions

- `yarn test:unit --spec packages/browser-rum-shopify/src/domain/shopifyBindings.spec.ts` passes

## Checklist

- [x] Tested locally
- [ ] Tested on staging

[RUM-18056]: https://datadoghq.atlassian.net/browse/RUM-18056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/browser-rum-shopify/src/domain/shopifyBindings.ts
+++ b/packages/browser-rum-shopify/src/domain/shopifyBindings.ts
@@ -11,7 +11,9 @@ export interface ErrorData {
   extensionName?: string
   extensionTarget?: string
   type?: string
+  appId?: string
   appName?: string
+  appVersion?: string
 }
 
 // Matches /checkouts/*, /checkout, including locale-prefixed paths.
@@ -54,12 +56,12 @@ export function initShopifyBindings(rumPublicApi: RumPublicApi, analytics: Shopi
     const err = new Error(error?.message)
     err.stack = error?.trace
     rumPublicApi.addError(err, {
-      extension: {
-        name: error?.extensionName,
-        target: error?.extensionTarget,
-        type: error?.type,
-        appName: error?.appName,
-      },
+      extensionName: error?.extensionName,
+      extensionTarget: error?.extensionTarget,
+      extensionErrorType: error?.type,
+      appId: error?.appId,
+      appName: error?.appName,
+      appVersion: error?.appVersion,
     })
   })
 }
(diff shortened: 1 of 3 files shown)


packages/browser-rum-shopify/src/domain/shopifyBindings.ts
```diff
@@ -54,12 +56,12 @@ export function initShopifyBindings(rumPublicApi: RumPublicApi, analytics: Shopi
     const err = new Error(error?.message)
     err.stack = error?.trace
     rumPublicApi.addError(err, {
-      extension: {
-        name: error?.extensionName,
-        target: error?.extensionTarget,
-        type: error?.type,
-        appName: error?.appName,
-      },
+      extensionName: error?.extensionName,
```
BenoitZugmeyer, line 59: **nitpick:** usually we use snake_case for attributes

BenoitZugmeyer (approved):

PR #4947, opened 2026-08-14
⚗️ Add canvas recording init configuration

BeltranBulbarellaDD:
## Motivation

Prepare the Browser SDK configuration surface for Session Replay canvas recording while keeping the capability behind an experimental feature flag. Grouping canvas recording settings in an object leaves room for future controls such as maximum image size or compression type without adding more top-level initialization parameters. Keeping `enable` separate from the related settings also allows canvas recording to be toggled without losing values such as the configured maximum FPS.

Design references:

- [One Pager: Session Replay Canvas Support](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/6882561949/One+Pager+Session+Replay+Canvas+Support)
- [Browser SDK Canvas Support](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/7082737787/Browser+SDK+Canvas+Support)

## Changes

- Add the hidden `sessionReplayCanvasRecording` RUM initialization object with this shape:

  ```ts
  sessionReplayCanvasRecording?: {
    enable: boolean
    maxFramesPerSecond?: number
  }
  ```

- Require `enable` so canvas recording can be explicitly enabled or disabled while preserving related settings.
- Default `maxFramesPerSecond` to 1 FPS and validate it at runtime as a finite number in the inclusive range `[0, 5]`. Setting it to `0` disables canvas frame recording.
- Require both `sessionReplayCanvasRecording.enable: true` and the `session_replay_record_canvas` experimental feature flag for recording to be effective.
- Keep canvas recording disabled when the configuration object is omitted or the experimental feature is unavailable.
- Add a reusable finite `number` field to the js-core configuration schema, with optional inclusive `min` and `max` bounds.
- Update the js-core API report and add focused unit coverage for defaults, custom frame rates, enabled and disabled states, feature gating, missing required fields, and invalid option shapes.

## Test instructions

- `yarn test:unit --spec packages/js-core/src/entries/configuration.spec.ts --spec packages/browser-rum-core/src/domain/configuration/configuration.spec.ts`
- `yarn typecheck`
- `yarn api:check`
- ESLint and Prettier checks on the changed source and test files.

## Checklist

- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser-rum-core/src/domain/configuration/configuration.spec.ts
+++ b/packages/browser-rum-core/src/domain/configuration/configuration.spec.ts
@@ -332,18 +332,16 @@ describe('validateAndBuildRumConfiguration', () => {
     it('is disabled by default', () => {
       const configuration = validateAndBuildRumConfiguration(DEFAULT_INIT_CONFIGURATION)!
 
-      expect(configuration.enableSessionReplayCanvasRecording).toBeFalse()
-      expect(configuration.sessionReplayMaxCanvasFps).toBe(0)
+      expect(configuration.enableSessionReplayCanvasRecording).toBeUndefined()
     })
 
     it('stays disabled when requested without the experimental feature', () => {
       const configuration = validateAndBuildRumConfiguration({
         ...DEFAULT_INIT_CONFIGURATION,
-        enableSessionReplayCanvasRecording: true,
+        enableSessionReplayCanvasRecording: {},
       })!
 
-      expect(configuration.enableSessionReplayCanvasRecording).toBeFalse()
-      expect(configuration.sessionReplayMaxCanvasFps).toBe(0)
+      expect(configuration.enableSessionReplayCanvasRecording).toBeUndefined()
     })
 
     describe('when the experimental feature is enabled', () => {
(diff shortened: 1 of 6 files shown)


BeltranBulbarellaDD:
@codex review

packages/browser-rum-core/src/domain/configuration/configuration.spec.ts
```diff
@@ -887,7 +940,12 @@ describe('serializeRumConfiguration', () => {
           ? 'track_long_task' // We forgot the s, keeping this for backward compatibility
           : // The following options are not reported as telemetry. Please avoid adding more of them.
             // `remoteConfiguration` is covered by the legacy `remote_configuration_id` field.
-            Key extends 'applicationId' | 'subdomain' | 'remoteConfiguration'
+            Key extends
+                | 'applicationId'
+                | 'subdomain'
+                | 'remoteConfiguration'
+                | 'enableSessionReplayCanvasRecording'
+                | 'sessionReplayMaxCanvasFPS'
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Better to add it to `rum-events-format`. Out of scope for this PR.

packages/browser-rum-core/src/domain/configuration/configuration.ts
```diff
@@ -394,6 +413,12 @@ export const RUM_SCHEMA = {
   enablePrivacyForActionName: { type: 'boolean', default: true },
   propagateTraceBaggage: { type: 'boolean', default: true },
   startSessionReplayRecordingManually: { type: 'boolean', default: false, strict: false },
+  enableSessionReplayCanvasRecording: {
+    type: 'schema',
+    schema: {
+      maxFramesPerSecond: { type: 'number', min: 0, max: 5, default: 1 },
+    },
+  },
```
sethfowler-datadog: Nit: if we keep the current structure, I think it'd make sense for `enableSessionReplayCanvasRecording` to be able to accept a boolean as well as an option, with the meaning that `true` means to enable the feature and use all the default settings.

That said, another option would be to use a structure like this:
```json
sessionReplayCanvasRecording: {
  enable: <boolean>,
  maxFramesPerSecond?: <number in [0, 5]>
}
```

I think this might end up feeling a bit nicer, although it's for a minor reason: if this configuration structure eventually gets mapped to an in-app UI, it's nice if the customer can enable or disable session replay canvas recording without "losing" the values of related settings like the max FPS.

I'm cool with whatever approach feels best to you!

BeltranBulbarellaDD: Ah nice! Like the other way would have been an empty array, that way they would have defaults but yeah this is totally nicer!

sethfowler-datadog (approved):

PR #4949, opened 2026-08-14
⚗️ Add canvas change detection for Session Replay

BeltranBulbarellaDD:
## Motivation

Session Replay canvas capture can be expensive. Before capturing a canvas, we need to know whether its bitmap or relevant DOM state changed so unchanged canvases can be skipped and bandwidth can be reduced.

Depends on #4947.

Design references:

- [One Pager: Session Replay Canvas Support](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/6882561949/One+Pager+Session+Replay+Canvas+Support)
- [Browser SDK Canvas Support](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/7082737787/Browser+SDK+Canvas+Support)

## Changes

- Add a `WeakSet`-backed canvas manager that treats new canvases as dirty and tracks canvases that have already been marked clean without retaining detached elements.
- Enable canvas dirty tracking only when canvas recording is enabled with a positive maximum frame rate.
- Instrument bitmap-changing 2D methods, including `fillRect`, `drawImage`, `fill`, `stroke`, `putImageData`, `clearRect`, and related methods.
- Use the existing `MutationObserver` to detect canvas `width`/`height` changes across property and attribute mutation paths, including reinsertion and inserted subtrees.
- Preserve `attributeNamespace` so namespaced attributes such as `x:width` are not mistaken for the real canvas `width` content attribute.
- Keep the canvas manager synchronized for mutations delivered through the observer and for records still queued when mutations are flushed.
- Add unit coverage for dirty-state behavior, 2D instrumentation, mutation-observer detection, namespaced attributes, inserted canvases, shadow roots, and recorder activation.
- Add an end-to-end scenario covering newly inserted canvases, drawing changes, resize changes, and non-drawing operations.

Screenshot 2026-08-19 at 13 25 47

## Scope

This PR adds the dirty-canvas prefilter and change detection. It does not yet implement periodic `toBlob()` capture, image downscaling, hash comparison, or sending canvas images. Those pieces can use the manager to capture only canvases currently considered dirty and mark them clean after processing.

The E2E scenario uses a test-only probe of the recorder's internal `WeakSet` because canvas image capture and request assertions are not part of this change yet.

## Validation

```bash
yarn test:unit --spec packages/browser-rum/src/domain/record/canvas/canvasManager.spec.ts --spec packages/browser-rum/src/domain/record/trackers/trackCanvas.spec.ts --spec packages/browser-rum/src/domain/record/trackers/trackMutation.spec.ts --spec packages/browser-rum/src/domain/record/record.spec.ts
yarn typecheck
```

- Targeted unit tests: 98 passed.
- Type checking: passed.
- Playwright coverage was added in `test/e2e/scenario/recorder/canvas.scenario.ts`; it was not run in this validation pass.


BeltranBulbarellaDD:
@codex review

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,56 @@
...
+          onPostCall(() => markCanvasDirty(context.canvas))
+        })
+      )
+    })
+  }
+
+  if (typeof HTMLCanvasElement !== 'undefined') {
+    instrumentationStoppers.push(
+      instrumentSetter(HTMLCanvasElement.prototype, 'width', markCanvasDirty),
+      instrumentSetter(HTMLCanvasElement.prototype, 'height', markCanvasDirty)
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Allrighty better watch for the different resize actions.

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,76 @@
...
+  'drawFocusIfNeeded',
+  'reset',
+]
+
+const CANVAS_SIZE_ATTRIBUTES = ['width', 'height']
+
+export function trackCanvas2DMutations(markCanvasDirty: MarkCanvasDirty): Tracker {
+  const instrumentationStoppers: Tracker[] = []
+
+  if (typeof CanvasRenderingContext2D !== 'undefined') {
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: We are focusing only on 2D

test/e2e/scenario/recorder/canvas.scenario.ts
```diff
@@ -0,0 +1,113 @@
...
+
+      window.canvasDirtyProbe = {
+        dirtyCalls,
+        originalFillRect: CanvasRenderingContext2D.prototype.fillRect,
+      }
+
+      WeakSet.prototype.delete = function (value) {
+        if (value instanceof HTMLCanvasElement) {
+          window.canvasDirtyProbe.cleanCanvases = this
+          dirtyCalls.push(value.id)
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Yeah it's just a mock.

packages/browser-rum-core/src/browser/domMutationObservable.ts
```diff
@@ -12,6 +12,8 @@ export interface RumAttributesMutationRecord {
   target: Element
   oldValue: string | null
   attributeName: string
+  // Distinguishes canvas content attributes from namespaced attributes with the same local name.
```
sethfowler-datadog: Nit: this interface matches the underlying web platform API, so we're not actually doing anything specific to `<canvas>` support at this layer; for that reason, I'd leave out this comment.

packages/browser-rum/src/domain/record/canvas/canvasManager.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { createCanvasManager } from './canvasManager'
+
+describe('createCanvasManager', () => {
```
sethfowler-datadog: Nit: these tests feel more like "CanvasManager" tests than "createCanvasManager" tests to me.

packages/browser-rum/src/domain/record/canvas/canvasManager.ts
```diff
@@ -0,0 +1,35 @@
+export interface CanvasManager {
+  clearDirtyCanvases: () => void
+  getDirtyCanvases: () => HTMLCanvasElement[]
+  isCanvasDirty: (canvas: HTMLCanvasElement) => boolean
+  markCanvasClean: (canvas: HTMLCanvasElement) => void
+  markCanvasDirty: (canvas: HTMLCanvasElement) => void
+}
+
+export function createCanvasManager(): CanvasManager {
+  const dirtyCanvases = new Set<HTMLCanvasElement>()
```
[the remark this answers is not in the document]
sethfowler-datadog: This is accurate, though the headline ("avoid strongly retaining...") points in the wrong direction, I think. You need to iterate the contents of this set, it seems, so strongly retaining is necessary. However, you should have a way to remove canvases from the set of dirty canvas, and you should perform this removal in `trackMutation.ts` when canvas elements are removed from the DOM.

packages/browser-rum/src/domain/record/trackers/trackMutation.ts
```diff
@@ -25,16 +26,19 @@ export function trackMutation(
...
-    serializeMutations(
-      timeStampNow(),
-      mutations.concat(observer.takeRecords() as RumMutationRecord[]),
-      emitRecord,
-      emitStats,
-      scope
-    )
+    // Process mutations still queued in the observer so canvas changes are tracked before flushing.
+    const queuedMutations = observer.takeRecords() as RumMutationRecord[]
+    markCanvasDirtyFromMutationRecords(queuedMutations, scope.canvasManager)
```
sethfowler-datadog: We shouldn't make this change here; handle the necessary invalidations inside `serializeMutations.ts`. I know, I know, it has "serialize" in the name, so you'd think that code is just about generating output, but the name is historical since the recording code is effectively in the middle of a massive refactor; pretend the name is `processMutations.ts`. (And, indeed, you'll notice that `serializeMutations()` is just a wrapper function that calls a function called `processMutations()` immediately!)

packages/browser-rum/src/domain/record/trackers/trackMutation.ts
```diff
@@ -25,16 +26,19 @@ export function trackMutation(
...
   })
 
-  const observer = new MutationObserver(monitor(mutationBatch.addMutations))
+  const observer = new MutationObserver(
+    monitor((mutations) => {
+      // Keep the canvas manager in sync with DOM mutations so unchanged canvases can be skipped.
+      markCanvasDirtyFromMutationRecords(mutations, scope.canvasManager)
+      mutationBatch.addMutations(mutations)
+    })
+  )
```
sethfowler-datadog: Same here; handling this here diverges from the layering that's intended in this code. Please handle this stuff inside `serializeMutations.ts`.

(Just FYI, you're also processing the same mutation records twice here; the records you handle here are the same records that get passed to the `createMutationBatch()` callback later.)

packages/browser-rum/src/domain/record/record.ts
```diff
@@ -51,11 +54,26 @@ export function record(options: RecordOptions): RecordAPI {
...
+    canvasManager
+  )
 
   const { stop: stopFullSnapshots } = startFullSnapshots(lifeCycle, processRecord, emitStats, flushMutations, scope)
 
+  // Seed all connected canvases after the initial full snapshot.
+  if (canvasManager) {
+    markCanvasAndDescendantsDirty(document, canvasManager)
+  }
+
```
sethfowler-datadog: We shouldn't do this here. Despite the fact that "serialize" is in the name (again, pretend it's "process"), we should be handling this in the serialization code -- in this case, `serializeNode.ts` is the right place to do this. Take a look at how we handle e.g. `shadowRootsController.addShadowRoot()` for inspiration.

packages/browser-rum/src/domain/record/recordingScope.ts
```diff
@@ -21,12 +22,14 @@ export interface RecordingScope {
   shadowRootsController: ShadowRootsController
   stringIds: StringIds
   styleSheetIds: StyleSheetIds
+  canvasManager: CanvasManager | undefined
```
sethfowler-datadog: Nit: please maintain alphabetical order.

Also, it'd be preferable if `canvasManager` was not optional; there's no harm in creating it all the time.

More generally, let's minimize the degree to which code outside of `trackCanvas.ts` needs to think about whether canvas recording is enabled. It's better if most code does the same thing in every configuration; this reproduces overall complexity and the amount of variations we need to test. It'd be fine even to register every canvas with the canvas tracker unconditionally, if we end up needing that functionality. Ideally enabling or disabling canvas recording should only control:
* Whether we instrument the canvas methods to track dirtiness.
* Whether we generate snapshots of the canvas content.

Those two things are worth making optional since they have a real cost, and they're also things that can be controlled internally by `trackCanvas.ts`, so other code doesn't need to test multiple code paths.

packages/browser-rum/src/domain/record/record.ts
```diff
@@ -76,6 +94,10 @@ export function record(options: RecordOptions): RecordAPI {
     trackViewEnd(lifeCycle, processRecord, flushMutations),
   ]
 
+  if (canvasManager) {
+    trackers.push(trackCanvas2DMutations(canvasManager.markCanvasDirty))
+  }
+
```
sethfowler-datadog: It's better to just create the tracker all the time. (See my longer comment about minimizing the differences between the "canvas recording enabled" and "canvas recording disabled" configurations.)

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,100 @@
...
+      isCanvasSizeAttribute(mutation.attributeName)
+    ) {
+      canvasManager.markCanvasDirty(mutation.target)
+    }
+
+    if (mutation.type === 'childList') {
+      for (let index = 0; index < mutation.addedNodes.length; index += 1) {
+        markCanvasAndDescendantsDirty(mutation.addedNodes[index], canvasManager)
+      }
+    }
```
sethfowler-datadog: Let's chat about this one. I don't think we need to do this (canvas fallback content is not rendered, after all!) but it's possible I'm missing something.

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,100 @@
...
+): void {
+  if (!canvasManager) {
+    return
+  }
+
+  for (const mutation of mutations) {
+    if (
+      mutation.type === 'attributes' &&
+      isCanvasElement(mutation.target) &&
+      (mutation.attributeNamespace === null || mutation.attributeNamespace === undefined) &&
```
sethfowler-datadog: I think you can safely just write `!mutation.attributeNamespace`, since the empty string is not a valid XML namespace prefix.

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,100 @@
...
+    stop: () => instrumentationStoppers.forEach((stopper) => stopper.stop()),
+  }
+}
+
+/**
+ * Canvas dimensions can be changed through several APIs, including the `width` and `height`
+ * properties and their corresponding attributes. The mutation observer consolidates these paths
+ * into attribute mutation records so that the canvas can be marked dirty consistently.
+ */
+export function markCanvasDirtyFromMutationRecords(
```
sethfowler-datadog: As I mentioned in another comment, you'll probably want to integrate this code into the processing logic in `serializeMutations.ts`. It should be fine to mark canvases as dirty even if canvas recording is not enabled; don't worry about making it conditional.

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,100 @@
...
+  'strokeText',
+  'drawImage',
+  'putImageData',
+  'drawFocusIfNeeded',
+  'reset',
+]
+
+const CANVAS_SIZE_ATTRIBUTES = ['width', 'height']
+
+export function trackCanvas2DMutations(markCanvasDirty: MarkCanvasDirty): Tracker {
```
sethfowler-datadog: Two nits:

1. I'd leave "mutations" out of the name and just call this function `trackCanvas`. Or maybe `trackCanvasContent`, since it's really the content drawn _within_ the canvas that we're ultimately interested in, but it's up to you; just doing some classic bikeshedding here. 🙂 

2. You should accept a `RecordingScope` as a parameter, and get the `CanvasManager` from there. You'll find later in the PR stack that you need a `RecordingScope` to emit records.

packages/browser-rum/src/domain/record/canvas/canvasManager.ts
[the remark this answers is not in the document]
BeltranBulbarellaDD: Yeah I think I get what you mean but I think that's scope for the next PR. Something like [this](https://github.com/DataDog/browser-sdk/pull/4955/changes#diff-270e538ada0140d752f5f20474f84eb10faa9f0e6f1919036bf928d3c29ef093R46-R51)?

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
BeltranBulbarellaDD: Don't you like the `2D` in the name? I think it makes it implicit.

BeltranBulbarellaDD: Yeah, scope is way better.

packages/browser-rum/src/domain/record/serialization/serializeNode.ts
```diff
@@ -138,6 +139,10 @@ function serializeElementNode(
   const domAttributes = Object.entries(serializeDOMAttributes(element, privacyLevel, transaction))
   transaction.addNode(insertionPoint, encodedElementName(element), ...domAttributes)
 
+  if (isCanvasElement(element)) {
+    transaction.scope.canvasManager.markCanvasDirty(element)
+  }
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: This is on purpose. [See](https://github.com/DataDog/browser-sdk/pull/4949#discussion_r3802406695)

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const instrumentationStoppers: Tracker[] = []
+
+  const { markCanvasDirty } = scope.canvasManager
+
+  if (typeof CanvasRenderingContext2D !== 'undefined') {
+    CANVAS_2D_DRAWING_METHODS.forEach((method) => {
+      instrumentationStoppers.push(
+        instrumentMethod(CanvasRenderingContext2D.prototype, method, ({ target: context, onPostCall }) => {
+          onPostCall(() => markCanvasDirty(context.canvas))
```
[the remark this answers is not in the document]
BeltranBulbarellaDD: Seems like too edge of a case

packages/browser-rum/src/domain/record/serialization/serializeMutations.ts
```diff
@@ -26,6 +27,7 @@ export function serializeMutations(
   emitStats: EmitStatsCallback,
   scope: RecordingScope
 ): void {
+  markCanvasDirtyFromMutationRecords(mutations, scope.canvasManager)
```
sethfowler-datadog: We want a deeper integration than this. What I'd do is to remove `markCanvasDirtyFromMutationRecords` totally, and instead handle the work you need to perform in `processAttributeMutations()`. (Probably around the point in the loop where you see `if (attributeName === 'value')`.)

Also, given that we currently don't capture any information about attribute namespaces in the recording, I'd recommend simply dropping those changes in this PR. The recorder generally does not work correctly with XHTML documents today, and I don't think we should try to solve that problem as part of these changes; it's a separate issue. I'm telling you this as part of this comment so that you don't worry about the fact that you don't have access to the XML namespace of the mutated attributes in `processAttributeMutations()` -- that's OK for now.

packages/browser-rum/src/domain/record/serialization/serializeMutations.ts
```diff
@@ -106,11 +108,16 @@ function processRemovedNodes(nodes: Set<Node>, transaction: SerializationTransac
 
   for (const node of nodes) {
     const nodeId = nodeIds.get(node)
-    if (nodeId === undefined) {
-      continue // This node wasn't serialized.
-    }
```
sethfowler-datadog, line 111: Please don't remove these lines; we should never be tracking any information about nodes that don't have node ids. You probably felt the need to do this because in `trackCanvas.ts` you are unconditionally adding canvases to the dirty set when they are mutated; the right approach is to only do that if the canvas element has a node id already, and to additional mark canvases as unconditionally dirty when they are first serialized (i.e., when they are first assigned a node id).

packages/browser-rum/src/domain/record/trackers/trackCanvasContent.ts
```diff
@@ -0,0 +1,59 @@
...
+  'stroke',
+  'fillText',
+  'strokeText',
+  'drawImage',
+  'putImageData',
+  'drawFocusIfNeeded',
+  'reset',
+]
+
+export function trackCanvasContent(scope: RecordingScope): Tracker {
```
sethfowler-datadog, line 21: Nit: I'd suggest renaming the file to `trackCanvasContent.ts` as well, to align with the function you're exporting.

packages/browser-rum/src/domain/record/trackers/trackCanvas.ts
```diff
@@ -0,0 +1,59 @@
...
+    scope.configuration.sessionReplayCanvasRecording.maxFramesPerSecond === 0
+  ) {
+    return { stop: noop }
+  }
+
+  const instrumentationStoppers: Tracker[] = []
+
+  const { markCanvasDirty } = scope.canvasManager
+
+  if (typeof CanvasRenderingContext2D !== 'undefined') {
```
sethfowler-datadog: Nit: we may as well make this check above, where we check `scope.configuration.sessionReplayCanvasRecording?.enable` and so forth; if `CanvasRenderingContext2D` does not exist, there's no reason to start up the tracker.

packages/browser-rum/src/domain/record/record.ts
```diff
@@ -76,6 +84,8 @@ export function record(options: RecordOptions): RecordAPI {
     trackViewEnd(lifeCycle, processRecord, flushMutations),
   ]
 
+  trackers.push(trackCanvasContent(scope))
```
sethfowler-datadog: Nit: please move this up into the `const trackers: Tracker[] = [` list above.

packages/browser-rum/src/domain/record/recordingScope.ts
```diff
@@ -26,7 +28,8 @@ export interface RecordingScope {
 export function createRecordingScope(
   configuration: RumConfiguration,
   elementsScrollPositions: ElementsScrollPositions,
-  shadowRootsController: ShadowRootsController
+  shadowRootsController: ShadowRootsController,
+  canvasManager: CanvasManager
```
sethfowler-datadog: Nit: please alphabetize the parameters to this function as well. (No benefit in adding the new parameter last, since all call sites have to change anyway.)

packages/browser-rum/src/domain/record/serialization/serializeMutations.ts
```diff
@@ -115,6 +117,10 @@ function processRemovedNodes(nodes: Set<Node>, transaction: SerializationTransac
         transaction.scope.shadowRootsController.removeShadowRoot(node.shadowRoot)
       }
 
+      if (isCanvasElement(node)) {
+        transaction.scope.canvasManager.markCanvasClean(node)
+      }
```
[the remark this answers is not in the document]
sethfowler-datadog: I think this is either stale or wrong; it looks to me like `trackCanvasContent` correctly checks that canvases are assigned a node id before marking them as dirty.

packages/browser-rum/src/domain/record/trackers/trackCanvasContent.ts
```diff
@@ -0,0 +1,60 @@
...
+  'strokeRect',
+  'fill',
+  'stroke',
+  'fillText',
+  'strokeText',
+  'drawImage',
+  'putImageData',
+  'drawFocusIfNeeded',
+  'reset',
+]
```
sethfowler-datadog: Nit:

```suggestion
const CANVAS_2D_DRAWING_METHODS = [
  'clearRect',
  'fillRect',
  'strokeRect',
  'fill',
  'stroke',
  'fillText',
  'strokeText',
  'drawImage',
  'putImageData',
  'drawFocusIfNeeded',
  'reset',
] as const

type Canvas2DDrawingMethod = (typeof CANVAS_2D_DRAWING_METHODS)[number]
```

This produces the same result, but avoids the need to list the method names twice.

sethfowler-datadog (approved):
Looks great! Thanks for your patience through all the discussion. One nit below; otherwise, I think this is ready to merge!

PR #4965, opened 2026-08-20
✨ Feature Flags tab — all Datadog sites, with overrides scoped per site

kellyw1806:
Adds the remaining Datadog sites to the Flags tab, and fixes a bug that expansion exposed.

---

## The bug

The wrapper stores flag overrides in the inspected page's `localStorage` under a single key with **no Datadog site in it**. `localStorage` is scoped per browser origin, not per Datadog site — so staging and US1 share one bucket on the same page.

Set an override on staging → reconnect to US1 → it's still there and **still applying**. If the same flag key exists in both environments with the same type (normal for a flag promoted through environments), nothing looks wrong at all: the row renders as an ordinary US1 override, and the page quietly behaves according to a value you set in staging.

## The fix

The wrapper can't be changed, so the extension keeps its own per-site stores and projects the connected site's into the key the wrapper reads:

```
dd.dd_flag.overrides.datad0g.com     ← staging's overrides
dd.dd_flag.overrides.datadoghq.com   ← US1's overrides
dd.dd_flag.overrides                 ← projection of the site you're connected to (what the wrapper reads)
```

Staging's copy is parked where the wrapper can't see it. Switch to US1, reload, and the page uses US1's real values; switch back, reload, and your staging override returns. Nothing is deleted, and the wrapper needs no change.

Three properties worth calling out, because they're where the design decisions live:

**Signing in stays inert.** The projection is compared before it's written and skipped when nothing would change. Reconnect to the site you used last time and nothing is written and nothing is prompted — a reload is only ever demanded when the page genuinely is running another site's values.

**Pre-scoping overrides are dropped, not migrated.** The site store is the only source, so anything left in the wrapper's key by an older build is cleared on first connect. The tab shipped days ago, so there's no override history worth carrying forward — and skipping the migration removes the adoption gate along with everything guarding it.

**A failed projection is surfaced, not assumed.** If the write fails, the page may still be applying another site's overrides while the panel shows this site's. That now raises an error instead of passing silently.

## Also in this PR

**Site picker** — adds US3, US5, EU1, AP1, AP2 (previously US1 + staging only). FED/GovCloud stays out: separate deployment, client isn't registered there.

**Clear all before signing in.** The signed-out notice warned that overrides were active but gave no way to act on them. It now has its own Clear all with a confirm step, surfaced failures, and a reload prompt (the page keeps applying them until it rebuilds its provider).

Signed out there's no site to scope to, so the notice reports every stored override and Clear all wipes all of them. Reporting only the wrapper's key would hide an override the page is still applying: a switch repoints that key before the reload that picks it up, so switching and then disconnecting without reloading would leave the notice empty while the old site's override was still in effect. Clearing only that key has the mirror problem — the rest would come back on reconnect, as if the button hadn't worked. The confirmation says it covers every site.

**Row warnings, split by severity.** The cross-environment warning is gone — an override in a site's store is that site's by construction. What remains:

| Signal | Treatment |
|---|---|
| Stored type disagrees with the flag | Red row — the wrapper rejects it at resolve time, so it genuinely won't apply |
| No active flag holds the key | Dimmed note — archived or deleted since the override was set; the override still resolves fine |

## Known limits

- **Overrides written outside the extension.** The site store is the only source, so anything written straight to the wrapper's key — by hand in the console, or by an older build — is overwritten on the next sync. Documented on `syncSiteOverrides`.
- **Switching sites while a write is in flight.** The provider remounts on a site change, so a write dispatched just before the switch can land after the new site's projection and restore the old site's overrides until the next sync. Each eval is atomic, so the stores themselves stay consistent. Accepted and documented, alongside the existing navigation-race note in that file.
- **The manual override form can still create a type mismatch.** It validates against the currently loaded catalog page only, so a flag on another page can be overridden with the wrong type. Pre-existing; now the main way to produce the red row. The proper fix is validating against an exact-key lookup, which makes Apply asynchronous — left for follow-up.

## Testing

- `yarn typecheck`, `yarn lint`, `yarn format` clean
- 94 unit tests passing. The storage tests run the real eval code against a real `localStorage`, so they exercise the projection and clear paths rather than mocks.

**Verified manually, both directions.** Worth running both if you're reviewing this — the first shows the leak is closed, the second shows nothing was destroyed closing it. Use a flag with a visible effect (dark mode works well) that exists in both environments.

*Leak is closed:*

1. Connect to **Staging**, override the flag on, reload → page picks it up.
2. Disconnect → the signed-out notice reports the stored override and offers to clear it.
3. Connect to **US1** → the reload banner appears.
4. Reload → the staging override is gone and the page uses US1's real value.

*Nothing was destroyed:*

5. Switch back to **Staging** → reload banner again.
6. Reload → the staging override is back and applying.

Step 6 is the one that matters. Steps 1–4 alone can't tell a parked override from a deleted one; only the round trip shows the store survived the switch.

### DIFF
--- a/developer-extension/src/panel/components/tabs/flagsTab/connectScreen.tsx
+++ b/developer-extension/src/panel/components/tabs/flagsTab/connectScreen.tsx
@@ -1,5 +1,6 @@
 import { Alert, Badge, Box, Button, Center, Group, Select, Stack, Text } from '@mantine/core'
-import React from 'react'
+import React, { useEffect, useRef, useState } from 'react'
+import { toErrorMessage } from '../../../../common/toErrorMessage'
 import { useSettings } from '../../../hooks/useSettings'
 import type { FlagAuthState } from './useFlagAuth'
 import { useInspectedPageOverrides } from './useInspectedPageOverrides'
@@ -41,16 +42,63 @@ export function ConnectScreen({ auth }: { auth: FlagAuthState }) {
 }
 
 /**
- * Surfaces overrides already stored on the inspected page while signed out — otherwise this screen is
- * all that renders, so an override left from an earlier session keeps affecting the page with nothing
- * to explain it. Informational only; everything that mutates overrides lives on the connected tab.
+ * Surfaces overrides already stored on the inspected page while signed out, with a Clear all so they
+ * can be wiped without signing in just for that.
  *
  * Mounted only while disconnected, so its navigation listeners never run alongside the connected
  * tab's own instance of this hook.
  */
 function DisconnectedOverridesNotice() {
-  const { status, overrides } = useInspectedPageOverrides()
+  const { status, overrides, clearAll, reloadPage } = useInspectedPageOverrides()
+  const [phase, setPhase] = useState<'idle' | 'confirming' | 'clearing' | 'cleared'>('idle')
+  const [error, setError] = useState<string | null>(null)
+  // Bumped on navigation so a clear that settles afterwards can't report on the page it left.
+  const generation = useRef(0)
   const count = Object.keys(overrides).length
+
+  // A navigation brings a fresh set of overrides: a pending confirmation would clear those instead,
+  // and a previous "cleared" no longer describes the page.
+  useEffect(() => {
+    if (status === 'loading') {
+      generation.current += 1
+      setPhase('idle')
+      setError(null)
+    }
+  }, [status])
+
+  const handleClearAll = () => {
+    const started = generation.current
+    setPhase('clearing')
+    setError(null)
+    void clearAll()
+      .then(() => {
+        if (started === generation.current) {
+          setPhase('cleared')
+        }
+      })
+      .catch((err: unknown) => {
+        if (started === generation.current) {
+          setError(toErrorMessage(err))
+          setPhase('idle')
+        }
+      })
+  }
+
+  // Gated on the page agreeing it has none: a same-document SPA navigation fires no webNavigation
+  // event, so without this the banner could outlive the state it describes.
+  if (phase === 'cleared' && status === 'ready' && count === 0) {
+    // The page keeps applying them until it reloads and rebuilds its provider.
+    return (
+      <Alert color="green" w="100%" title="Overrides cleared">
+        <Group gap="xs">
+          <Text size="xs">Reload the page to stop applying them.</Text>
+          <Button size="compact-xs" variant="light" color="green" onClick={reloadPage}>
+            Reload page
+          </Button>
+        </Group>
+      </Alert>
+    )
+  }
 
   if (status !== 'ready' || count === 0) {
     return null
@@ -66,8 +114,39 @@ function DisconnectedOverridesNotice() {
       title={`${count} override${count === 1 ? '' : 's'} active on this page`}
     >
       <Text size="xs">
-        These are stored in the page and keep applying while you are signed out. Sign in to view and remove them.
+        These are stored in the page and keep applying while you are signed out. Sign in to review and remove them
+        individually.
       </Text>
+      <Group gap="xs" mt="xs">
+        {phase === 'confirming' || phase === 'clearing' ? (
+          <>
+            <Text size="xs" fw={600}>
+              Clear all {count} override{count === 1 ? '' : 's'} on this page?
+            </Text>
+            <Button size="compact-xs" color="red" onClick={handleClearAll} loading={phase === 'clearing'}>
+              Clear
+            </Button>
+            <Button
+              size="compact-xs"
+              variant="subtle"
+              color="gray"
+              onClick={() => setPhase('idle')}
+              disabled={phase === 'clearing'}
+            >
+              Cancel
+            </Button>
+          </>
+        ) : (
+          <Button size="compact-xs" variant="light" color="red" onClick={() => setPhase('confirming')}>
+            Clear all
+          </Button>
+        )}
+      </Group>
+      {error && (
+        <Text c="red" size="xs" mt="xs">
+          Could not clear the overrides: {error}
+        </Text>
+      )}
     </Alert>
   )
 }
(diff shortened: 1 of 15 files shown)


developer-extension/src/panel/components/tabs/flagsTab/connectScreen.tsx
```diff
@@ -41,17 +42,64 @@ export function ConnectScreen({ auth }: { auth: FlagAuthState }) {
 }
 
 /**
- * Surfaces overrides already stored on the inspected page while signed out — otherwise this screen is
- * all that renders, so an override left from an earlier session keeps affecting the page with nothing
- * to explain it. Informational only; everything that mutates overrides lives on the connected tab.
+ * Surfaces overrides already stored on the inspected page while signed out, with a Clear all so they
+ * can be wiped without signing in just for that.
```
BeltranBulbarellaDD:
```suggestion
 * can be wiped without signing in.
```

developer-extension/src/panel/components/tabs/flagsTab/connectScreen.tsx
```diff
@@ -41,17 +42,64 @@ export function ConnectScreen({ auth }: { auth: FlagAuthState }) {
...
+
+  // Gated on the page agreeing it has none: a same-document SPA navigation fires no webNavigation
+  // event, so without this the banner could outlive the state it describes.
+  if (phase === 'cleared' && status === 'ready' && count === 0) {
+    // The page keeps applying them until it reloads and rebuilds its provider.
+    return (
+      <Alert color="green" w="100%" title="Overrides cleared">
+        <Group gap="xs">
+          <Text size="xs">Reload the page to stop applying them.</Text>
+          <Button size="compact-xs" variant="light" color="green" onClick={reloadPage}>
```
BeltranBulbarellaDD:
```suggestion
          <Text size="xs">Please reload the page to stop applying them.</Text>
          <Button size="compact-xs" variant="light" color="green" onClick={reloadPage}>
```

developer-extension/src/panel/components/tabs/flagsTab/flagCatalogList.tsx
```diff
@@ -141,6 +146,13 @@ function FlagRow({
         </Text>
         <FlagKey value={flag.key} />
         {flag.description && <FlagDescription description={flag.description} />}
+        {override && unusable && (
+          <Text size="xs" c="red" fw={600}>
+            {flag.unresolved
+              ? "No active flag with this key on this site — it's from a different Datadog environment, or archived here. Clear it."
+              : `Type mismatch: stored as ${flagTypeLabel(override.type)}, this flag is ${flagTypeLabel(flag.type)}. Clear it.`}
```
BeltranBulbarellaDD:
```suggestion
              ? "No active flag with this key on this site — it's from a different Datadog environment, or archived here. Please clear it."
              : `Type mismatch: stored as ${flagTypeLabel(override.type)}, this flag is ${flagTypeLabel(flag.type)}. Clear it.`}
```

developer-extension/src/panel/components/tabs/flagsTab/flagTypes.ts
```diff
@@ -64,6 +64,19 @@ export function parseTypedString(type: Exclude<FlagType, 'BOOLEAN'>, raw: string
...
+ * left by another Datadog site — though a flag archived here, or a manual override applied with the
+ * wrong type, land here too, and a key that exists identically on both sites is missed entirely.
+ */
+export function isOverrideUnusable(
+  flag: { type: FlagType; unresolved?: boolean },
+  override: { type: FlagType } | undefined
+): boolean {
+  return override !== undefined && (flag.unresolved === true || override.type !== flag.type)
+}
+
```
BeltranBulbarellaDD: Maybe we can add the site in the localstorage override. That way we only show the ones for the current site.

BeltranBulbarellaDD (commented):
Small nits. But let's see if it would be nicer to just filter by site in the local storage. It's not the full solution but I think would simplify a little the code here. So we fetch only the ones on the logged in site for example. 
All the rest looks good! I'll take a closer look after the changes.

developer-extension/src/panel/components/tabs/flagsTab/flagsTab.tsx
```diff
@@ -64,6 +66,37 @@ function ConnectedFlagsTab({ auth }: { auth: FlagAuthState }) {
       }
     >
       <Box px="md" py="sm">
+        {siteSwitchNeedsReload && (
+          <>
+            <Alert color="orange" title="Reload to apply this site's overrides">
+              <Group justify="space-between" wrap="nowrap">
+                <Text size="sm">
+                  Which overrides apply changed when you switched sites, and the page hasn&apos;t reloaded since.
```
BeltranBulbarellaDD: This message is a little bit tough to understand.

developer-extension/src/panel/components/tabs/flagsTab/inspectedPageFlags.ts
```diff
@@ -91,34 +112,104 @@ export async function readFlagState(): Promise<FlagState | null> {
...
+ * Null means the write failed — callers must surface that, since the page then keeps applying
+ * whichever site's overrides it already had.
+ */
+export async function syncSiteOverrides(site: string): Promise<{ changed: boolean; overrides: FlagOverrides } | null> {
+  const storeKey = siteOverridesKey(site)
+  try {
+    const result = (await evalInWindow(`
+      ${EVAL_HELPERS}
+      const projection = parse(localStorage.getItem(${JSON.stringify(OVERRIDES_KEY)})) || {}
+      let siteOverrides = parse(localStorage.getItem(${JSON.stringify(storeKey)}))
```
BeltranBulbarellaDD: So I know maybe some customers are using it but we released it only a few days ago so it's almost like they are previewing the feature but it could change. Would we be able to not have to account for users who have previous overrides? We can just clean their local storage flags and start over? So we don't keep legacy code.

BeltranBulbarellaDD (commented):
Works great!

developer-extension/src/panel/components/tabs/flagsTab/inspectedPageFlags.ts
kellyw1806: Yep, going to do that! good idea

BeltranBulbarellaDD (approved):
LGTM!

